PR #1, opened 2016-02-09
Fix comments in config file

FryDay:
Some of the defaults in the comments do not match the actual default values.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -6,7 +6,7 @@ module.exports = {
 	// Set to 'false' to enable users.
 	//
 	// @type     boolean
-	// @default  false
+	// @default  true
 	//
 	public: true,
 
@@ -30,7 +30,7 @@ module.exports = {
 	// Set the local IP to bind to.
 	//
 	// @type     string
-	// @default  "0.0.0.0"
+	// @default  undefined
 	//
 	bind: undefined,
 
@@ -60,7 +60,7 @@ module.exports = {
 	// URLs posted in channels.
 	//
 	// @type     boolean
-	// @default  true
+	// @default  false
 	//
 	prefetch: false,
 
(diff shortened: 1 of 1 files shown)


MaxLeiter:
Looks good to me

YaManicKill:
Yep, lgtm too.

defaults/config.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
 	// Set the local IP to bind to.
 	//
 	// @type     string
-	// @default  "0.0.0.0"
+	// @default  undefined
```
xPaw: Could this perhaps explain what undefined means in this context?

YaManicKill: Yeah, that's probably a good plan. Just a short explanation that undefined means it'll listen on everything.

astorije:
Hey @FryDay, could you get rid of the merge commit in the history please? Let me know if you don't know how to do that and we can sync on IRC.

In the future, when you want to sync with master, always use `git rebase master` instead of merge or pull (pull is simply fetch + merge).

Thanks a lot for fixing cruft!!! :smile:

FryDay:
@astorije certainly, I'll take care of it tonight when I get home.

astorije:
Perfect :-)

FryDay:
@astorije I believe I fixed it. If it doesn't look okay, let me know.

astorije:
I'm :+1: on this, going for second review (as there were changes after @YaManicKill's comment).
@FryDay, maybe squash the commits into one if you happen to stop by before @YaManicKill gives his :+1:, otherwise not a huge deal.

YaManicKill:
:+1: 

I'm fine with teh 2 commits because they are both doing logically seperate things. Merging.

PR #4, opened 2016-02-12
Update chat template so no user-icon shown when a server is selected.

MaxLeiter:


client/views/chat.tpl
```diff
@@ -33,4 +35,4 @@
 		</div>
 	</aside>
 </div>
-{{/each}}
+{{/each}}
```
xPaw: Bring back the new line.

YaManicKill: Yeah, this comment.

I thought this would break the build, but I guess this PR was added before travis was set up. Is there a way to nudge travis that this PR exists?

xPaw: It will get nudged when author fixes this ;)

YaManicKill: Of course it will. Friday brain.

astorije:
Also, needs to rebase with master if we want Travis to run on this.

astorije:
No big deal re: rebase, it's a minor PR, but I agree with @xPaw and @YaManicKill, this file needs an EOF line.

PR #7, opened 2016-02-12
Handle invite event

xPaw:
Brings in https://github.com/erming/shout/pull/631

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -666,6 +666,13 @@ button, .user {
 #chat.hide-quit .quit {
 	display: none !important;
 }
+
+#chat .invite .from:before {
+	font-family: FontAwesome;
+	content: "\f003";
+	color: #2ECC40;
+}
+
 #chat .join .from:before {
     font-family: FontAwesome;
 	content: "\f090";
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -193,8 +193,10 @@ $(function() {
 
 		var chan = chat.find(target);
 		var from = data.msg.from;
+		var msg;
 
 		if ([
+			"invite",
 			"join",
 			"mode",
 			"kick",
@@ -205,6 +207,7 @@ $(function() {
 			"action",
 		].indexOf(type) !== -1) {
 			switch (type) {
+			case "invite": data.msg.formattedAction = "invited " + data.msg.target + " to"; break;
 			case "join": data.msg.formattedAction = "has joined the channel"; break;
 			case "mode": data.msg.formattedAction = "sets mode"; break;
 			case "kick": data.msg.formattedAction = "has kicked"; break;
@@ -215,11 +218,10 @@ $(function() {
 			default: data.msg.formattedAction = "";
 			}
 
-			var action = $(render("msg_action", data.msg));
-			return action;
+			msg = $(render("msg_action", data.msg));
+		} else {
+			msg = $(render("msg", data.msg));
 		}
-
-		var msg = $(render("msg", data.msg));
 
 		var text = msg.find(".text");
 		if (text.find("i").size() === 1) {
--- a/src/plugins/irc-events/invite.js
+++ b/src/plugins/irc-events/invite.js
@@ -1,0 +1,29 @@
+var _ = require("lodash");
+var Msg = require("../../models/msg");
+
+module.exports = function(irc, network) {
+	var client = this;
+	irc.on("invite", function(data) {
+		var target = data.to;
+		if (target.toLowerCase() === irc.me.toLowerCase()) {
+			target = "you";
+		}
+
+		var chan = _.findWhere(network.channels, {name: data.channel});
+		if (typeof chan === "undefined") {
+			chan = network.channels[0];
+		}
+
+		var msg = new Msg({
+			type: Msg.Type.INVITE,
+			from: data.from,
+			target: target,
+			text: data.channel
+		});
+		chan.messages.push(msg);
+		client.emit("msg", {
+			chan: chan.id,
+			msg: msg
+		});
+	});
+};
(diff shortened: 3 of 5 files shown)


client/css/style.css
```diff
@@ -666,6 +666,11 @@ button, .user {
 #chat.hide-quit .quit {
 	display: none !important;
 }
+#chat .invite .from:before {
+	font-family: FontAwesome;
+	content: "\f0e0";
```
astorije: FYI, this is the envelope icon.
Throwing in out there, what about the `user-plus` icon instead: https://fortawesome.github.io/Font-Awesome/icon/user-plus/ ? Any opinions?

client/js/lounge.js
```diff
@@ -215,12 +218,11 @@ $(function() {
...
 
-			var action = $(render("msg_action", data.msg));
-			return action;
+			msg = $(render("msg_action", data.msg));
+		} else {
+			msg = $(render("msg", data.msg));
 		}
 
-		var msg = $(render("msg", data.msg));
-
```
astorije: Just to clarify, why are changes from L218 to L223 necessary?

src/plugins/irc-events/invite.js
```diff
@@ -0,0 +1,29 @@
+var _ = require("lodash");
+var Msg = require("../../models/msg");
+
+module.exports = function(irc, network) {
+	var client = this;
+	irc.on("invite", function(data) {
+		var target = data.to;
+		if (target.toLowerCase() === irc.me.toLowerCase()) {
```
astorije: Aren't nicks case sensitive?

dgw: While I can't find any RFC that defines nicknames as case-insensitive, [RFC 1459](https://tools.ietf.org/html/rfc1459#section-2.2) does point out the special casing imposed by IRC's Scandinavian origins (the `{}|` and `[]\` sets of characters).

client/js/lounge.js
xPaw: `msg_action` had to be passed through the inline channel parser.

src/plugins/irc-events/invite.js
xPaw: That's how all other events in shout handle it, I just copied it over.

client/css/style.css
xPaw: Not too keen on `user-plus`.

astorije: Fair enough. What about https://fortawesome.github.io/Font-Awesome/icon/envelope-o/ then? I find https://fortawesome.github.io/Font-Awesome/icon/envelope/ too bulky :-/

src/plugins/irc-events/invite.js
astorije: Fair point @xPaw but as an aside, do you think this should be changed on all events (in a different PR of course)?

xPaw: I would hold on changing something like that until #24

client/css/style.css
xPaw: Will change it to `envelope-o`

astorije:
Will take a look, thanks! Could you squash commits please?

astorije:
I could invite someone with `/invite someone #channel` but not with `/invite someone` while being inside a channel. I know it's not the spec per se, but it's pretty standard (and useful!) to be able to invite someone in the current channel. Any chance you can add that?

Also, I was not being able to invite someone when not an op on a Freenode channel. I get that it's not a bug of this PR, but that's still pretty frustrating...

client/js/lounge.js
```diff
@@ -205,6 +207,7 @@ $(function() {
 			"action",
 		].indexOf(type) !== -1) {
 			switch (type) {
+			case "invite": data.msg.formattedAction = "invited " + data.msg.target + " to"; break;
```
astorije: When testing, I was not able to see this displayed other than `invited you to #channel`. Wondering if that's a bug. What do I need to do to trigger this properly?

The steps I followed:

- user1 goes to #channel
- user1 types `/invite user2 #channel`
- user2 sees `user1 invited you to #channel` in the network window
- user2 clicks on `#channel` link as displayed in previous step
- Both users see `user2 has joined the channel` in #channel, no mention of something along the lines of `user1 invited user2 to this channel`

Am I misunderstanding something here?

astorije:
2 concerns detailed above, but apart from that it's looking good, that's pretty cool! Close to not be behind Shout anymore lol...

client/js/lounge.js
xPaw: You're not misunderstanding. This is merely future-proofing. Specifically http://ircv3.net/specs/extensions/invite-notify-3.2.html

xPaw:
> I could invite someone with /invite someone #channel but not with /invite someone while being inside a channel.

Create a separate issue for it, this is not what this PR is for.

> Also, I was not being able to invite someone when not an op on a Freenode channel.

There's a channel mode to change that.

client/js/lounge.js
astorije: I don't usually like this sort of future proofing because it adds crust and confusion, but since it's already there and I believe you will add IRCv3 stuff as soon as you can, I'll pass on this one... :-)

astorije:
> Create a separate issue for it, this is not what this PR is for.

Meh, I could argue but I won't. I'll open an issue and will actually code the feature myself if you don't mind: with all this organization/fork/PR-reviews-by-myself-on-shout stuff, I haven't been able to code a single line for weeks or more... Very frustrating! :-)

In the meantime, :+1:. Asking for a second review here because it's a bit more than minor.

astorije:
> Create a separate issue for it, this is not what this PR is for.

Alright, I wrote this, will open a PR with the code as soon as this one gets merged.

YaManicKill:
Alright, looks good to me. Merging.

astorije:
Perfect, I'll open a PR with my addition (see above) tonight after work :-)

PR #12, opened 2016-02-12
Use persistent commit url for shout's changelog

xPaw:
If the changelog file changes or gets moved in Shout repo, our link will still work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,5 +68,4 @@ This is the first release of **The Lounge**, picking up where Shout `v0.53.0` le
 
 ---
 
-All previous changes can be found on [Shout's CHANGELOG, starting at `v0.53.0`](https://github.com/erming/shout/blob/master/CHANGELOG.md#0530--2016-01-07
-).
+All previous changes can be found on [Shout's CHANGELOG, starting at `v0.53.0`](https://github.com/erming/shout/blob/35587f3c35d0a8ac78a2495934ff9eaa8f1aa71c/CHANGELOG.md#0530--2016-01-07).


YaManicKill:
Sensible change. :+1: from me

astorije:
I actually would rather follow the `master` version as the CHANGELOG might be updated someday (look at v0.52.0 for example, it just says "???"). I don't think the URL will change anytime soon, so probability that it gets improved seems higher than probability it gets changed.

Of course, it's not definitive and we will probably change that when a lots of new versions are published, for the reason you mentioned. Just short term I'd prefer `master`.

Does that make sense, @xPaw and @YaManicKill?

YaManicKill:
Well, the thing ist hat if the changelog ever changes, it'll be out of date, generally. The most likely reason for the changelog changing is they have a new version with new changes. We won't be based on top of that, so it means that people going to that will probably see some stuff that isn't in the lounge. So I'm personally for not following master.

xPaw:
> I actually would rather follow the master version as the CHANGELOG might be updated someday (look at v0.52.0 for example, it just says "???").

The only thing that needs fixing that changelog is `0.52.0` as you mentioned. If that gets fixed, our changelog url should be pinned to the commit that fixes that. There is no need to link to master, as we forked from 0.53, and future changelogs won't make sense.

astorije:
Note that for the reasons you both mentioned, I anchored the link to `#0530--2016-01-07`, so that it always shows from v0.53.0 and backwards. But that's a detail so I won't argue furthermore, this will become less and less relevant with each new release of The Lounge anyway.
:+1:

YaManicKill:
Having fun, @astorije ? :-P

astorije:
Yeah, I messed something up, sorry for the noise :-/ Should be good this time...

PR #13, opened 2016-02-12
Changed the hostname of freenode in the default config

dubzi:
As suggested in erming/shout#647 the default freenode hostname is not correct. Yes, it works, but Freenode suggests to use chat.freenode.net instead.

It's a super easy fix, so why not?

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -132,9 +132,9 @@ module.exports = {
 		// Host
 		//
 		// @type     string
-		// @default  "irc.freenode.org"
+		// @default  "chat.freenode.net"
 		//
-		host: "irc.freenode.org",
+		host: "chat.freenode.net",
 
 		//
 		// Port
--- a/src/client.js
+++ b/src/client.js
@@ -124,7 +124,7 @@ Client.prototype.connect = function(args) {
 	var client = this;
 	var server = {
 		name: args.name || "",
-		host: args.host || "irc.freenode.org",
+		host: args.host || "chat.freenode.net",
 		port: args.port || (args.tls ? 6697 : 6667),
 		rejectUnauthorized: false
 	};


YaManicKill:
If you are doing that, you should probably [change it here, too](https://github.com/erming/shout/blob/35587f3c35d0a8ac78a2495934ff9eaa8f1aa71c/src/client.js#L128) so we are consistent.

xPaw:
@YaManicKill I think that default in the code should go away, and the server should just refuse to open a connection if there is no host specified.

YaManicKill:
@xPaw Yeah, good point actually. Just need some checks, and just remove the options.

astorije:
I do like having Freenode as a default, actually...
Counter proposal: can we make the change this @YaManicKill suggested and then see in a different issue/PR for what @xPaw is proposing? That would avoid delaying this trivial and needed PR :-)

xPaw:
> I do like having Freenode as a default, actually...

It would stay in the config file, it doesn't need to be in code itself.

astorije:
Right, I see, and nothing in the config file if the admin wants no default. Still, I prefer this in a different PR as it needs a bit more testing/checks than what @dubzi offers here.

YaManicKill:
Yeah, I agree with @astorije as it's a different issue.

dgw:
:+1: Keeping freenode in is good, IMO. I've seen a lot of clients start off with their project channel/network in the default settings so new users are immediately directed to where they can get help.

xPaw:
@dgw That's what the config variable is for. It doesn't need to have a hard fallback in the code.

dgw:
@xPaw This PR is *about* the config variable…

xPaw:
@dgw I know, keeping freenode in the config default is fine.

But here: https://github.com/erming/shout/blob/35587f3c35d0a8ac78a2495934ff9eaa8f1aa71c/src/client.js#L128 it needs to be removed, and the connection must be refused if there is no server specified.

astorije:
Yes, we all agree @xPaw, but this is really a different issue that deserves a different PR :-)
Let's ship this quickly when @dubzi makes the change, and also consider your proposition after that.

dubzi:
Alright, I've also fixed the hostname in the client.js file for consistency.

astorije:
:+1:

xPaw:
:+1:

PR #3, opened 2016-02-10
Allow shout-irc to be proxied behind a /path/ url

gdamjan:
contains three small changes to use relative urls instead of absolute rooted at /
* favicons in index.html
* badge and pop sound in shout.js
* the url for the socket.io endpoint

PR #20, opened 2016-02-13
Allow The Lounge to be proxied behind a /path/ url

astorije:
Reword PR https://github.com/thelounge/lounge/pull/3 from @gdamjan

---

contains three small changes to use relative urls instead of absolute rooted at /
* favicons in index.html
* badge and pop sound in lounge.js
* the url for the socket.io endpoint


astorije:
Merging directly as this is simply a commit message change.

PR #21, opened 2016-02-13
Revert "Allow The Lounge to be proxied behind a /path/ url"

astorije:
Reverts thelounge/lounge#20

This PR breaks The Lounge: as `./socket.io/` cannot be found, the client restarts in loop many times per second.

@gdamjan, sorry for merging too soon, would you mind taking a look at this? Thanks!

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -16,9 +16,9 @@
 	<link rel="stylesheet" href="css/style.css">
 	<link id="theme" rel="stylesheet" href="<%= theme %>">
 
-	<link rel="shortcut icon" href="img/favicon.png">
-	<link rel="icon" sizes="192x192" href="img/touch-icon-192x192.png">
-	<link rel="apple-touch-icon" sizes="120x120" href="img/apple-touch-icon-120x120.png">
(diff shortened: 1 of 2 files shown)


astorije:
Merging this revert as-is as `master` is currently broken...

gdamjan:
socket.io is idiotic. I'll send a new pull request

gdamjan:
OTOH. that hard `refresh()` on disconnect is not the best idea. at least it should have some backoff and possibility to debug.

YaManicKill:
@gdamjan Yeah, I've got a branch on my fork to remove that because it's actually unneessary (socket io reconnects fine) but I'm testing it out irl for a while before merging it back.

astorije:
Agreed @gdamjan and @YaManicKill, sorry I didn't investigate more to fix this, but first I wouldn't have merge without having a review if I had tried to fix the issue (while a revert is an emergency feature), and it was the middle of the night, I merely wanted to fix master.
I'm glad you reacted lightning fast though, thanks :-)

PR #22, opened 2016-02-13
Remove #foo channel from default config

astorije:
I actually don't see the point of having `#foo` as a default channel... Maybe there is a channel exactly design for that (testing that your client is well connected) that we should use, otherwise simply `#thelounge` is enough IMO.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -186,11 +186,12 @@ module.exports = {
 
 		//
 		// Channels
+		// This is a comma-separated list.
 		//
 		// @type     string
-		// @default  "#foo, #thelounge"
+		// @default  "#thelounge"
 		//
-		join: "#foo, #thelounge"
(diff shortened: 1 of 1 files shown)


YaManicKill:
:+1: from me then, looks good.

In this circumstance, do we need **another** +1 from someone other than you because it's your PR?

xPaw:
:+1: ¯\\\_(ツ)_/¯

astorije:
I mentioned this today in the [review process](https://github.com/thelounge/lounge/wiki/Maintainers'-corner#review-process).
Basically: only one :+1: from another project maintainer if PR is from a project maintainer (what we were doing on erming/shout) unless this is a non-minor change that would likely be debated a little bit. We should use common sense to determine if it's minor or if people will have opinion :-)

PR #27, opened 2016-02-13
Allow The Lounge to be proxied behind a /path/ url

gdamjan:
contains three small changes to use relative urls instead of absolute rooted at /

* favicons in index.html
* badge and pop sound in lounge.js
* the url for the socket.io endpoint

Typical use with nginx:

```
  location /path/ {
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_pass http://localhost:9000/;
  }
```
those trailing / are important.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -16,9 +16,9 @@
 	<link rel="stylesheet" href="css/style.css">
 	<link id="theme" rel="stylesheet" href="<%= theme %>">
 
-	<link rel="shortcut icon" href="/img/favicon.png">
-	<link rel="icon" sizes="192x192" href="/img/touch-icon-192x192.png">
-	<link rel="apple-touch-icon" sizes="120x120" href="/img/apple-touch-icon-120x120.png">
+	<link rel="shortcut icon" href="img/favicon.png">
+	<link rel="icon" sizes="192x192" href="img/touch-icon-192x192.png">
+	<link rel="apple-touch-icon" sizes="120x120" href="img/apple-touch-icon-120x120.png">
 
 	</head>
 	<body class="<%= public ? "public" : "" %>">
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1,5 +1,6 @@
 $(function() {
-	var socket = io();
+	var path = window.location.pathname + "socket.io/";
+	var socket = io({path:path});
 	var commands = [
 		"/close",
 		"/connect",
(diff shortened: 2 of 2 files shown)


astorije:
FYI, this is a rework of #3 and #20 (after it was reverted in #21).

astorije:
As mentioned on the IRC channel, I have concerns about this change.
In particular, we should completely ensure that the existing solution fully works after your change.
So that would be great if someone could test running The Lounge under the following schemes:

- [ ] http://domain.com/lounge
- [ ] http://lounge.domain.com/
- [ ] http://domain.com:9000/
- Anything else?

xPaw:
@astorije `pathname` would be as follows:
1. `/lounge`, and in this case, yes, it will break.
2. `/`
3. `/`

astorije:
Indeed, but we cannot release this without actually testing, with the risk of breaking The Lounge for everyone as #3 / #20 did.

gdamjan:
`/lounge` will break with or without my patch, ain't it?

astorije:
But I'm confused @gdamjan, wasn't that the purpose of your PR, to run The Lounge behind a `/whatever` path?

gdamjan:
huh. ok. I'm not sure what you were asking, but the very important difference is `/path` vs `/path/` (as I've specified in the commit message).

astorije:
Right, I see, sorry. I'm ok with the distinction, and wherever this gets documented or used, I'd suggest to add a redirect from `/path` to `/path/` if possible.
Editing my checklist above.

gdamjan:
with the above nginx rule, it'll send a redirect automatically.

astorije:
Good to know, thanks!

gdamjan:
and to be extra clear... both of the trailing slashes are important, in location and in proxy_pass.

astorije:
Adding `help wanted` label, if someone can test the 3 scenarios listed above and report that would help shipping this faster. Thanks!

JocelynDelalande:
Tested the 3 scenarios and read patch, it's ok

:+1: :)

astorije:
Letting @YaManicKill give his second review to this one as he self-assigned #3 at the time.

YaManicKill:
So, on the basis of @JocelynDelalande having tested this, and it looks good (and solves the issue we had last time we merged it) I'll give my :+1: and merge.

astorije:
Great, great addition! I'll publish to npm either today or tomorrow!
(Also, removing `second review needed` label)

YaManicKill:
> (Also, removing second review needed label)

Sorry, I'll get the hang of this soon. This is the biggest project I've maintained :-P

PR #28, opened 2016-02-13
Remove "badge" setting and rely on browser choice for desktop notifcations

lpoujol:
Remove the badge setting and rely on browser choice for desktop notifications.

But the user still have to activate the desktop notifications via the lounge settings page. Making lounge ask the permission to the browser.

The status on whether or not it's activated is also displayed on the settings page.

Some screens to show how it looks :
> 
> "Deactivated" state :
> notification_deactivated
> Activated :
> notifications_activated
> Blocked by web browser
> notification_blocked
>

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -243,8 +243,9 @@
 							</div>
 							<div class="col-sm-12">
 								<label class="opt">
-								<input id="badge" type="checkbox" name="badge">
-								Enable badge
+								<input id="desktopNotifications" type="checkbox" name="desktopNotifications">
+								Enable desktop notifications<br>
+								<div class="error" id="warnDisabledDesktopNotifications"><strong>Warning</strong>: Desktop notifications are blocked by your web browser</div>
 								</label>
 							</div>
 							<div class="col-sm-12">
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -427,7 +427,7 @@ $(function() {
 	var userStyles = $("#user-specified-css");
 	var settings = $("#settings");
 	var options = $.extend({
-		badge: false,
+		desktopNotifications: false,
 		colors: false,
 		join: true,
 		links: true,
@@ -487,11 +487,11 @@ $(function() {
 	}).find("input")
 		.trigger("change");
 
-	$("#badge").on("change", function() {
+	$("#desktopNotifications").on("change", function() {
 		var self = $(this);
 		if (self.prop("checked")) {
 			if (Notification.permission !== "granted") {
-				Notification.requestPermission();
+				Notification.requestPermission(updateDesktopNotificationStatus);
 			}
 		}
 	});
@@ -689,7 +689,7 @@ $(function() {
 				}
 				toggleFaviconNotification(true);
 
-				if (options.badge && Notification.permission === "granted") {
+				if (options.desktopNotifications && Notification.permission === "granted") {
 					var title;
 					var body;
 
@@ -791,6 +791,8 @@ $(function() {
 			}
 		});
 	});
+
+	windows.on("show", "#settings", updateDesktopNotificationStatus);
 
 	forms.on("submit", "form", function(e) {
 		e.preventDefault();
@@ -916,6 +918,23 @@ $(function() {
 	function refresh() {
 		window.onbeforeunload = null;
 		location.reload();
+	}
+
+	function updateDesktopNotificationStatus(){
+		var checkbox = $("#desktopNotifications");
+		var warning = $("#warnDisabledDesktopNotifications");
+
+		if (Notification.permission === "denied"){
+			checkbox.attr("disabled", true);
+			checkbox.attr("checked", false);
+			warning.show();
+		} else {
+			if (Notification.permission === "default" && checkbox.prop("checked")){
+				checkbox.attr("checked", false);
+			}
+			checkbox.attr("disabled", false);
+			warning.hide();
+		}
 	}
 
 	function sortable() {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1109,6 +1109,11 @@ button,
 	color: #2ecc40;
 }
 
+#settings .error {
+	color: #e74c3c;
+	margin-top: .2em;
+}
+
 #form {
 	background: #eee;
 	border-top: 1px solid #ddd;


astorije:
Can @xPaw, @JocelynDelalande, @YaManicKill or @wizardfrag self-assign this one please? :-)

YaManicKill:
I shall take a look at this tonight or tomorrow. 

1 question before I actually download it and try it:

If notifications are disabled, and then the user wants notifications again, how do they get it? I don't think the app can ask for notifications again, can it? And if it can't, what happens when the user tells the browser not to disable notifications again?

I'll have a play around and see how it works.

YaManicKill:
Oh, also, I'm not sure about there not being a setting to disable notifications after  they have been enabled. That seems a little awkward for users.

lpoujol:
The app can't ask again if the browser blocks notifications. If you re-enable them, notifications will pop again without the need to touch shout settings. But, if you set it as "always ask", shout will have to ask again.

There's is no settings because in the end it's the browser who have the last word on wether notifications are displayed or not. Having a checkbox on the settings page for notifications doesn't really mean anything as long as browser settings are set differently.

YaManicKill:
> The app can't ask again if the browser blocks notifications. If you re-enable them, notifications will pop again without the need to touch shout settings. But, if you set it as "always ask", shout will have to ask again.

Alright, good. I'll double check that works with different browsers as well, but I believe you.

> There's is no settings because in the end it's the browser who have the last word on wether notifications are displayed or not. Having a checkbox on the settings page for notifications doesn't really mean anything as long as browser settings are set differently.

client/index.html
```diff
@@ -243,8 +243,8 @@ <h1 class="title">Settings</h1>
 							</div>
 							<div class="col-sm-12">
 								<label class="opt">
-								<input id="badge" type="checkbox" name="badge">
-								Enable badge
+									Desktop notification status : <span id="desktopNotificationsStatus">Unknown</span>
```
YaManicKill: There's extra indentation here.

client/js/lounge.js
```diff
@@ -880,6 +882,24 @@ $(function() {
...
 	}
 
+	function updateDesktopNotificationStatus(){
+		var btn = $("#activateDesktopNotifications");
+		var infoText = $("#desktopNotificationsStatus");
+
+		if (Notification.permission === "granted"){
+			btn.hide();
+			infoText.text("Activated");
+		}else if (Notification.permission === "denied"){
```
YaManicKill: Missing space

client/js/lounge.js
```diff
@@ -880,6 +882,24 @@ $(function() {
...
+			btn.hide();
+			infoText.text("Blocked by the browser");
+		}
+		else {
```
YaManicKill: Ah, extra newline.

YaManicKill:
Right, so in general I quite like it. I'm just not a fan of the user having to go to the browser settings to disable notifications.

I'm not sure I want another setting for that though.

I'll give a :+1: and see what others think. I'm happy for this to go in.

xPaw:
I don't like that you completely removed disabling notifications in settings. When we will add per-channel settings, we will *need* to have an ability to disable notifications.

Having a badge to say that notifications are not enabled by the browser is a good idea though.

astorije:
So, after reading the initial proposition and @YaManicKill's and @xPaw's concerns, here is my proposal. First of all, some definitions to be sure we are all on the same page:

On the browser side of things, notification states are:
1. **Not enabled**: The user was never asked for this website, or refused to accept or answer by clicking the &times; button. This state is only possible when the user set their browser to always ask for approval (default).
2. Allowed: User has allowed this website to display notifications (either he allowed for all or already allowed for this website before).
3. Blocked: User has disallowed this website to display notifications (either he disallowed for all or already disallowed for this website before).

screen shot 2016-02-20 at 18 05 36

On The Lounge, you should be able to enable or disable notifications, as long as they are allowed (state **(2)**). As @YaManicKill and @xPaw have mentioned, first because in the future we will want to be able to override this setting per channel, but also because disabling notifications within the browser is quite a reach in the browser settings. The opposite is also true: when the user has blocked the notifications, it's a quite a dig to re-enable... but that's their choice and there is nothing we can do about it :-)

My proposal is to keep the current checkbox but to enhance it:
- An empty checkbox (along with the text `Enable notifications`) means The Lounge will not trigger notifications at all, whatever states **(1)** through *(3)** the browser is in.
- A `disabled` checkbox means that we are in state **(3)**.
- A checked checkbox means we are in state **(2)** and The Lounge triggers notifications.
- When the user checks the checkbox, either:
  - Browser is in state **(1)**, so the confirmation popup above will appear again:
    - If the user clicks on `Allow`, the checkbox gets checked and notifications will show up in the future (third bullet point of this list).
    - If the user clicks on `Block`, the checkbox goes back to unchecked and gets `disabled` (second bullet point of this list).
    - If the user closes without answering, the checkbox goes back to unchecked (first bullet point of this list).
  - If browser is in state **(2)**, the checkbox simply gets checked and notifications will show up (third bullet point of this list).
- When the user unchecks the checkbox, the notifications stop getting triggered within The Lounge, but permission at the browser-level is left unchanged.

This allows for what was discussed earlier:
- Notifications can be disabled within the app even after they were enabled and allowed.
- Notifications can be enabled per channel (future) independently of the browser permission.
- The user knows notifications are blocked with the disabled checkbox. This can be enough feedback for this PR, or a mention after the checkbox text (`Blocked by the browser`).

When I look at how things are done at the moment (before your PR), I see two crucial differences:
- When notifications are blocked by the user, we can still check or uncheck the notifications. This is not good because it's misleading: there is no way a notification can get through. Disabling the checkbox would help with that, as well as adding the mention.
- When the checkbox is checked for the first time, the user gets asked. If he doesn't answer (closes the permission popup), the checkbox stays checked although notifications cannot go through. Unchecking the checkbox when this happens would help.

---

Sorry, this was a very long comment for something very simple, but I wanted to give all details. How does everyone feel about that?

astorije:
@lpoujol, ping?

maxpoulin64:
I agree with the others on this one. Notification settings are not convenient to find in browsers, especially when The Lounge is running as a web application where there is no browser (especially mobile where pinning the app to the home screen opens it fullscreen): 
capture d ecran_2016-02-27_19-31-13 
screenshot_2016-02-27-19-32-26 png. We need to be able to toggle them on/off.


Additionally, another problem for the moment is that notifications needs to be disabled on mobile Chrome because `Notification` was never implemented. It exists, it will grant the permission, but invoking `new Notification()` throws an error which makes the app crash and reload.

astorije:
> Additionally, another problem for the moment is that notifications needs to be disabled on mobile Chrome because Notification was never implemented. It exists, it will grant the permission, but invoking new Notification() throws an error which makes the app crash and reload.

Didn't know that.
@maxpoulin64, this PR has been open for a while, would you be OK with my suggestion in terms of usage and see to fix this afterwards?

maxpoulin64:
@astorije Yes, I fully agree with you. Your solution above has a state where we don't try to send notifications, this is the only thing that's needed. I just didn't want to rely on the browser to block our notifications.

YaManicKill:
I agree with @astorije and @maxpoulin64. @lpoujol would you be able to make these changes and rebase the PR? It would be good to get these changes in :-)

astorije:
I am removing the `second review needed` label until this PR goes further. Can't wait to hear from you, @lpoujol :-)

lpoujol:
I brought back the setting (also renaming it from badge to desktopNotifications in the code)

I check the browser state on display of the settings page. 
It will ask permission for desktop notifications if the setting is activated but not granted by the browser.
It will turn off the setting if it's blocked by the browser, disable the checkbox and display a warning below.

But there is an edge case that I can't really cover. It's when the browser choice for desktop notifications (Block, Ask, Allow) change while the settings page is displayed :/. There will be a Permission API on browsers, but it's still experimental :/
A work around would be to not disable the checkbox.  Checking the box would trigger a new check.

client/index.html
```diff
@@ -243,8 +243,9 @@ <h1 class="title">Settings</h1>
 							</div>
 							<div class="col-sm-12">
 								<label class="opt">
-								<input id="badge" type="checkbox" name="badge">
-								Enable badge
+								<input id="desktopNotifications" type="checkbox" name="desktopNotifications">
+								Enable desktop notifications<br />
```
astorije: `<br>` instead of `<br />`

client/js/lounge.js
```diff
@@ -427,7 +427,7 @@ $(function() {
 	var userStyles = $("#user-specified-css");
 	var settings = $("#settings");
 	var options = $.extend({
-		badge: false,
+		desktopNotifications: false,
```
astorije: I know I'm getting close to the bikeshed here, but I think `browserNotifications` is closer to what it is than `desktopNotifications`. Don't take my word for it, I'll be fine with whatever opinion @YaManicKill has :-)

astorije:
@lpoujol, looking good! :-)

I noticed an issue though, with the permission pop-up which never stops asking once clicked on the checkbox once, even after changing tab:

notifications

Also, when decision to allow or block is not taken, the checkbox should never be checked, which can be the case at the moment.

> But there is an edge case that I can't really cover. It's when the browser choice for desktop notifications (Block, Ask, Allow) change while the settings page is displayed :/.

I am not too bothered by this edge case. We could also check if `denied` when the settings are open and the app is being focused (ie. clicking on the browser tab), but not a big deal to me. The situation above is a bit more worrying IMO.
Another solution could be a "Check now" link/button as part of the warning to trigger a new check. None of these are ideal which is why I'd leave it like what you have now.

> A work around would be to not disable the checkbox. Checking the box would trigger a new check.

Meh, it would be weird to be able to be able to check a checkbox that automatically goes back to unchecked because blocked by the browser I think.

client/index.html
lpoujol: Fixed

lpoujol:
Last commit fixes the issue with the permission popup (Didn't had on Firefox because it won't execute the callback if no decision is taken) .
(I'll squash all commits if everyone is happy with that ;) )

astorije:
@lpoujol, works great, awesome, I really like it!!

:+1:

astorije:
Taking ownership of this one, to let @YaManicKill breathe a bit :-)

astorije:
By the way, yes, please do squash @lpoujol :-)

maxpoulin64:
All good to me, :+1:, ok to merge after squash.

lpoujol:
Squashed !

client/index.html
```diff
@@ -243,8 +243,9 @@ <h1 class="title">Settings</h1>
 							</div>
 							<div class="col-sm-12">
 								<label class="opt">
-								<input id="badge" type="checkbox" name="badge">
-								Enable badge
+								<input id="desktopNotifications" type="checkbox" name="desktopNotifications">
+								Enable desktop notifications<br>
+								<div class="error" id="warnDisabledDesktopNotifications">Warning : Desktop notifications are blocked by your web browser</div>
```
xPaw: Nit: can you remove space after warning here

client/js/lounge.js
```diff
@@ -487,11 +487,13 @@ $(function() {
...
 		.trigger("change");
 
-	$("#badge").on("change", function() {
+	$("#desktopNotifications").on("change", function() {
 		var self = $(this);
 		if (self.prop("checked")) {
 			if (Notification.permission !== "granted") {
-				Notification.requestPermission();
+				Notification.requestPermission(function(){
+					updateDesktopNotificationStatus();
```
xPaw: You can write this as `Notification.requestPermission(updateDesktopNotificationStatus);`

client/js/lounge.js
```diff
@@ -918,6 +924,23 @@ $(function() {
 		location.reload();
 	}
 
+	function updateDesktopNotificationStatus(){
+		var checkbox = $("#desktopNotifications");
+		var warning = $("#warnDisabledDesktopNotifications");
+
+		if (Notification.permission === "denied"){
+			checkbox.attr("disabled", true);
+			checkbox.attr("checked", false);
```
xPaw: Checkboxes allow "indeterminate" state, like so: `checkbox.indeterminate = true;`, do you think it would be helpful to use it here, or when permission state is `default`?

lpoujol: I don't think it would be really usefull if the permission state is `default`. Because in the end, the user didn't care if it's on `default`. If he checks the box to activate the setting, he'll have the permission popup and it's done.

astorije: I agree with @lpoujol here and I don't think `checkbox.indeterminate = true;` would help here. If the permission is denied, then is *must* be disabled and unchecked. Same for `default`.

client/js/lounge.js
astorije: :+1:

client/index.html
astorije: :+1: on the space, that's a difference with French :P
For the warning, I don't think so at the moment, because none of the similar-looking error do that. I'd keep this for a later time.

astorije:
:+1: and merging, thanks all for the multiple reviews and thanks a lot @lpoujol for this and for sticking with us for 22 days!! :-)

PR #33, opened 2016-02-14
Fix #29: Correctly send sorted user list whenever an user joins

### LINKED ISSUE #29: User list doesn't get re-sorted when users join
The user list on the right is most of the time correctly sorted, except when a user joins the channel. For example:

screen shot 2016-02-13 at 14 11 07

It looks like something as simple as clicking on another open channel and going back to it re-sorts the list, but not right when the user joins.

xPaw:


astorije:
Nice one, just tested locally and it works perfectly. Thanks!!

:+1:

astorije:
Will leave a few extra hours for @YaManicKill or @wizardfrag to review this but considering it comes from a maintainer **and** it's minor (**and** it's a bug fix!), I'll merge this before making the next release very very soon if no other comment (@xPaw agreed).

PR #34, opened 2016-02-14
Fix #30: Close channel on the client as soon as close event is registered

### LINKED ISSUE #30: Impossible to close a channel where the user was kicked out of
When being kicked out of a channel, the channel stays open. Then trying to close the channel is impossible, because it is emitting a `/close` event while the channel is not open on the server:

**(Do not try to reproduce this on you normal running instance or you will have to restart the server to get rid of it.)**

impossible_to_close

I would not want the channel to be closed by itself, because I might not want to lose the scrollback, and I would not know I was kicked out (since the channel would close instantly). Instead, the server should not try to quit the channel if not actually on this channel.

xPaw:
This is my proposed fix to the issue, unsure if there is a better way of doing it.

### DIFF
--- a/src/plugins/inputs/part.js
+++ b/src/plugins/inputs/part.js
@@ -4,17 +4,15 @@ module.exports = function(network, chan, cmd, args) {
 	if (cmd !== "part" && cmd !== "leave" && cmd !== "close") {
 		return;
 	}
-	var client = this;
-	if (chan.type === "query") {
-		network.channels = _.without(network.channels, chan);
-		client.emit("part", {
-			chan: chan.id
-		});
-	} else {
+	if (chan.type !== "query") {
 		var irc = network.irc;
 		if (args.length === 0) {
 			args.push(chan.name);
 		}
 		irc.part(args);
 	}
+	network.channels = _.without(network.channels, chan);
+	this.emit("part", {
+		chan: chan.id
+	});
 };


maxpoulin64:
Seems reasonable to me. In theory IRC is over TCP, and in theory servers should let you part channels so that seems fine. I think most (all?l clients just close the tab immediately and leave the part as a background job, dropping messages (the user doesn't want it, right?). The existance of the channel is checked, so if the server is stubborn [the channel will just reopen as a query](https://github.com/thelounge/lounge/blob/master/src/plugins/irc-events/message.js#L21).

astorije:
@xPaw, tested locally and I can indeed "close" (aka remove from the list) the channel, but still receives a red "You're not on that channel" on the server window. Any way we can avoid that easily?

maxpoulin64:
I don't think so. It looks like a generic server error message, likely triggered by the extra part command.

Maybe we could add state tracking to chan.js instead to avoid leaving a channel twice?

astorije:
Ping @xPaw? :-)

xPaw:
@astorije, @maxpoulin64 is right. You *will* get an error for trying to leave a channel you're not in, but that's about it. Adding state tracking for channels would require quite a bit more code.

astorije:
> You will get an error for trying to leave a channel you're not in, but that's about it.

OK then, seems good enough to me.

> Adding state tracking for channels would require quite a bit more code.

Clearly out of scope for this PR and definitely needs some discussion/design before jumping into coding that, so good enough for me. At least this fixes the original bug without adding a new one :-)

:+1: and merging as-is, no second review needed per the review process as @xPaw is a reviewer himself, the PR is minor **and** it's simply a bug fix.

Issue #36, opened 2016-02-14
Deprecation warning for moment().zone


deiu:
It turns out `log.js` contains a call to `moment().zone`, which is deprecated. It should use `moment().utcOffset` instead, as per https://github.com/moment/moment/issues/1779. Please stand by for PR.

```
Error
    at Moment.zone (/usr/local/lib/node_modules/thelounge/node_modules/moment/moment.js:850:105)
    at Object.module.exports.write (/usr/local/lib/node_modules/thelounge/src/log.js:19:22)
    at Client.emit (/usr/local/lib/node_modules/thelounge/src/client.js:90:9)
    at Client.<anonymous> (/usr/local/lib/node_modules/thelounge/src/plugins/irc-events/welcome.js:14:10)
    at Client.emit (events.js:129:20)
    at Client.<anonymous> (/usr/local/lib/node_modules/thelounge/node_modules/slate-irc/lib/plugins/welcome.js:14:11)
    at Client.emit (events.js:129:20)
    at Client.onmessage (/usr/local/lib/node_modules/thelounge/node_modules/slate-irc/index.js:386:8)
    at Parser.emit (events.js:107:17)
    at Parser.online (/usr/local/lib/node_modules/thelounge/node_modules/slate-irc-parser/index.js:88:8)
```

Update: pending PR -- https://github.com/thelounge/lounge/pull/37

astorije:
:+1:

PR #37, opened 2016-02-14
Replace deprecated moment().zone

deiu:
This PR attempts to fix https://github.com/thelounge/lounge/issues/36. Use `moment().utcOffset` instead of the deprecated `moment().zone`, as per https://github.com/moment/moment/issues/1779.

### DIFF
--- a/src/log.js
+++ b/src/log.js
@@ -16,7 +16,7 @@ module.exports.write = function(user, network, chan, msg) {
 	var format = (config.logs || {}).format || "YYYY-MM-DD HH:mm:ss";
 	var tz = (config.logs || {}).timezone || "UTC+00:00";
 
-	var time = moment().zone(tz).format(format);
+	var time = moment().utcOffset(tz).format(format);
 	var line = "[" + time + "] ";
 
 	var type = msg.type.trim();


astorije:
Hey @deiu, thanks for the contrib!

Did you notice any consequence with this change?

deiu:
The only consequence is that I'm no longer getting any warnings/errors in the logs when I connect to the server.

astorije:
Ah, I could finally reproduce as I was not testing it right, sorry.

On `master`, warning looks like:

```
Deprecation warning: moment().zone is deprecated, use moment().utcOffset instead. https://github.com/moment/moment/issues/1779
Arguments: UTC+00:00
Error
    at Moment.zone (.../lounge/node_modules/moment/moment.js:850:105)
    at Object.module.exports.write (.../lounge/src/log.js:19:22)
    at Client.emit (.../lounge/src/client.js:90:9)
    at Client.<anonymous> (.../lounge/src/plugins/irc-events/notice.js:28:10)
    at emitOne (events.js:77:13)
    at Client.emit (events.js:169:7)
    at Client.<anonymous> (.../lounge/node_modules/slate-irc/lib/plugins/notice.js:24:11)
    at emitOne (events.js:82:20)
    at Client.emit (events.js:169:7)
    at Client.onmessage (.../lounge/node_modules/slate-irc/index.js:386:8)
```

I confirm this PR removes the deprecation warning, :+1:.

YaManicKill:
Cool, just tested the functions to make sure we don't get anything different, and we don't.

:+1: and merging.

PR #38, opened 2016-02-14
Update lodash

xPaw:
Hopefully I didn't miss any function changes.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -110,7 +110,7 @@ ClientManager.prototype.removeUser = function(name) {
 ClientManager.prototype.autoload = function(/* sockets */) {
 	var self = this;
 	setInterval(function() {
-		var loaded = _.pluck(
+		var loaded = _.map(
 			self.clients,
 			"name"
 		);
@@ -135,4 +135,3 @@ ClientManager.prototype.autoload = function(/* sockets */) {
 		});
 	}, 1000);
 };
-
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -46,8 +46,8 @@ Network.prototype.export = function() {
 		"commands"
 	]);
 	network.nick = (this.irc || {}).me;
-	network.join = _.pluck(
-		_.where(this.channels, {type: "channel"}),
+	network.join = _.map(
+		_.filter(this.channels, {type: "channel"}),
 		"name"
 	).join(",");
 	return network;
--- a/src/plugins/irc-events/kick.js
+++ b/src/plugins/irc-events/kick.js
@@ -5,7 +5,7 @@ module.exports = function(irc, network) {
 	var client = this;
 	irc.on("kick", function(data) {
 		var from = data.nick;
-		var chan = _.findWhere(network.channels, {name: data.channel});
+		var chan = _.find(network.channels, {name: data.channel});
 		var mode = chan.getMode(from);
 
 		if (typeof chan === "undefined") {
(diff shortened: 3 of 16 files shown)


astorije:
Let's have 2 :+1: on this one, it's quite a dangerous PR and needs to be tested fairly well :-)
Thanks a lot for this @xPaw!

astorije:
Can @JocelynDelalande, @YaManicKill or @wizardfrag self-assign this one please? :-)
(I guess that's where organization teams are interesting :p)

JocelynDelalande:
As I'm not myself doogfooding shout, I'm not useful for that kind of review that requires testing in real conditions, for some time, using all features to see where it may crashe.

src/clientManager.js
```diff
@@ -135,4 +135,3 @@ ClientManager.prototype.autoload = function(/* sockets */) {
 		});
 	}, 1000);
 };
-
```
astorije: Was that intentional? We should always keep an EOF line to keep git and other tools happy :-)

src/clientManager.js
```diff
@@ -110,7 +110,7 @@ ClientManager.prototype.removeUser = function(name) {
 ClientManager.prototype.autoload = function(/* sockets */) {
 	var self = this;
 	setInterval(function() {
-		var loaded = _.pluck(
+		var loaded = _.map(
 			self.clients,
```
astorije: Unless I am being mistaken (it's late :-)), it seems like this is the same as:

```js
self.clients.map(function (client) {
  return client.name;
});
```

Example:

```js
> var arr = [{foo: 1}, {foo: 2}, {foo: 3}];
> _.map(arr, "foo");
[ 1, 2, 3 ]
> arr.map(function (x) { return x.foo; });
[ 1, 2, 3 ]
```

While I understand you are merely upgrading as-is here, I believe we should leverage vanilla JS whenever possible, and this is a good case :-)
What do you think?

src/models/network.js
```diff
@@ -46,8 +46,8 @@ Network.prototype.export = function() {
 		"commands"
 	]);
 	network.nick = (this.irc || {}).me;
-	network.join = _.pluck(
-		_.where(this.channels, {type: "channel"}),
+	network.join = _.map(
+		_.filter(this.channels, {type: "channel"}),
```
astorije: Same as above, I'd rather see:

```js
this.channels.filter(function (channel) {
  return channel.type === "channel";
}).map(function (channel) {
  return channel.name;
}).join(",");
```

Yes, it's a few extra chars, but this gets us as close as the language as possible so I think it's a win! Plus we can even improve it with `.reduce()` if necessary (but I don't believe it is)!

src/plugins/irc-events/kick.js
```diff
@@ -5,7 +5,7 @@ module.exports = function(irc, network) {
 	var client = this;
 	irc.on("kick", function(data) {
 		var from = data.nick;
-		var chan = _.findWhere(network.channels, {name: data.channel});
+		var chan = _.find(network.channels, {name: data.channel});
```
astorije: I'm sure there is an ES5 equivalent here too, but I won't be as picky a with `.map()` and `.filter()` which are now well accepted.

astorije:
(Taking ownership and first review then)

So far so goo, @xPaw, but I left you a few comments inline :-)

Also, just wondering, did you try something like https://github.com/lodash/lodash-migrate to migrate? Not that we should blindly rely on that, but that might give you some extra confidence that nothing was forgotten.

Thanks!

src/clientManager.js
xPaw: It had two new lines at the end, my editor automatically removed the extra one.

src/clientManager.js
xPaw: Not a bad idea, however I would get the upgrade out of the way, and then look into using vanillajs.

xPaw:
> Also, just wondering, did you try something like https://github.com/lodash/lodash-migrate to migrate?

I did just now, it didn't seem to complain.

astorije:
:+1:

YaManicKill:
Cool, looks good to me. 

:+1: Merging.

PR #42, opened 2016-02-14
package.json: lint all .js files

williamboman:
This makes the linting process less prone to miss new (or even
existing) files. Ignoring specific file paths is delegated to the
.eslintignore file in the root directory.


astorije:
:+1:

xPaw:
:+1:

PR #47, opened 2016-02-15
Fix some of the actions coloring

### LINKED ISSUE #31: Broken things in recent changes on action display
@xPaw's work on actions (https://github.com/erming/shout/pull/588) is a huge improvement, but I noticed some issues/regressions that could be improved. Listing them here as they are all defects around the same subject:

- [x] 1. When joining a channel, the statement to indicate the topic is "#channel has changed the topic to ...". This is only true when the topic is actually changed, when joining the wording should be different.
screen shot 2016-02-13 at 19 21 18

- [x] 2. When *colored nicknames* are enabled in the settings, nicks in actions are not colored:

  Before | After
--- | ---
screen shot 2016-02-13 at 14 10 02 | screen shot 2016-02-13 at 14 09 54

- [x] 3. As shown by the previous screenshots, action are now in black. They should be displayed in gray as before to not get as much attention as normal message.

- [x] 4. From the pre-Shout/The Lounge days I can remember, I think I used to see "[nick] has changed the topic..." instead of "[channel] has changed the topic...", i.e. is the nick that changed the topic last saved along with the topic? If so that's what should be displayed instead!

xPaw:
chrome_2016-02-15_11-38-31

Actions are muted color, fixes highlighted actions.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -774,13 +774,6 @@ button,
 	color: #999;
 }
 
-#chat .error,
-#chat .error .from,
-#chat .highlight,
-#chat .highlight .from {
-	color: #f00;
-}
-
 #chat.hide-join .join span,
 #chat.hide-mode .mode span,
 #chat.hide-motd .motd span,
@@ -794,6 +787,14 @@ button,
 	font-family: FontAwesome;
 	content: "\f003";
 	color: #2ecc40;
+}
+
+#chat .join .text,
+#chat .kick .text,
+#chat .part .text,
+#chat .quit .text,
+#chat .nick .text {
+	color: #999;
 }
 
 #chat .join .from:before {
@@ -835,6 +836,7 @@ button,
 	color: #2ecc40;
 }
 
+#chat .action .from,
 #chat .action .text,
 #chat .action .user {
 	color: #f39c12;
@@ -843,7 +845,6 @@ button,
 #chat .action .from:before {
 	font-family: FontAwesome;
 	content: "\f005";
-	color: #f39c12;
 }
 
 #chat .notice .time,
@@ -854,6 +855,14 @@ button,
 
 #chat .notice .user:before {
 	content: "Notice: ";
+}
+
+#chat .error,
+#chat .error .from,
+#chat .highlight .from,
+#chat .highlight .text,
+#chat .highlight .user {
+	color: #f00;
 }
 
 #chat .toggle-button {


wizardfrag:
:+1:

astorije:
I will take a look at this tonight.

astorije:
Re: wording, I guess that'll have to do until we can display who has set the topic at the time (cf. https://github.com/thelounge/lounge/issues/31#issuecomment-184326386).

Re: colored nicknames, can't you really display the right color instead of gray? It really is a regression compared to pre-nice-actions era... :-(

xPaw:
> Re: colored nicknames, can't you really display the right color instead of gray? It really is a regression compared to pre-nice-actions era... :-(

Hmm, I'm not really sure about it. Having joins/parts muted makes them less eye catchy, and having colored nicknames would defeat the purpose.

xPaw:
I've pushed a second half of the fix for #90 into this PR, which makes highlighted actions to be red, as they should be.

astorije:
Actually @xPaw, sorry to ask you that, but do you mind keeping these in 3 different PRs?
It would definitely help with the review and keeping/building the change log nice and clean.
Thanks!!

xPaw:
It's 3 separate commits, that changes nothing for the changelog. Creating separate for PRs and then rebasing them every time is too much hassle.

astorije:
Yes it does, it simplifies my life when it's time to write it, so really that would be nice of you...
It's not too difficult to do: checkout the branch, create 2 new branches of it, rebase -i to remove 2 commits per branch, push and open 2 new PRs... I would do it myself as it would literally take me 1 minute, but I can't as it's on your fork :-/ Pleeeease :)

xPaw:
I really don't see why you want 3 separate PRs, do you want every commit to be attached to a separate issue?

astorije:
Not every commit, but every separate topic, yes. And really the reason is to help me with writing the change log, which takes enough time for me to ask, when willing to do it well. Definitely helps with the review and time to merge too.

But nevermind, I didn't think it would be such a big deal and I would have preferred to do it myself rather than argue about it.

xPaw:
I've already opened the shell to separate them, give me a couple of minutes.

astorije:
Cool, thanks.

maxpoulin64:
Personally I would like the colors to still be visible but dimmed, but I understand this is a pain with the way colors work at the moment. A workaround that we could use however could be to set the text color to `#000`, and use `opacity: .25;` to dim the whole line back to `#ccc`

Quick test on the dev console: 
capture d ecran_2016-02-27_19-49-09

Thoughts?

astorije:
@xPaw, could you extract the highlight fixing commit in a separate PR please? It is probably good as-is while we apparently are not set for the other commit. The highlight fixing thing should ship quickly as it's a bug fix IMO.

For the other one:
- `#ccc` is too muted actually. I don't have vision issues and yet I have a hard time reading them, so I would go with same as own user text, `#999` (which doesn't even pass WCAG accessibility guidelines, so we should not go brighter than that).
- For the nick color, I stand by my earlier position: should be green when colored nicknames are disabled and colored when colored nicknames are enabled. It currently is a regression compared to Shout for example. However, I understand they are a pain to deal with so fine to go in a different PR and use green for everything right now, just not gray. Then users can use custom stylesheet (coming soon :p) to disable colors altogether.

xPaw:
@astorije I've changed color and removed selectors that affected names. I'd rather get this PR out of the way. We will need some grand discussion on all the stuff that affects colored names, and colors related in general, it's just too messy right now.

astorije:
:cry: for colored nicknames...

:+1: for now

PR #48, opened 2016-02-15
Fix #41: Replace close button image with font-awesome

### LINKED ISSUE #41: Replace .close image with font-awesome icon

xPaw:
chrome_2016-02-15_12-02-54

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -318,19 +318,25 @@ button, .user {
 	display: none;
 }
 #sidebar .close {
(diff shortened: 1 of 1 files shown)


astorije:
I will take a look at this tonight.

astorije:
This icon is very bulky, especially compared to what was there before. I'd much prefer something more like the "X" of GitHub here, `octicon-x` (on the top right of my comment, since your an org member)
The only thing I can think of is using `&times;` / `×`. It looked great when I naively tested by remplacing the unicode but there are for sure other things to change (remove the font, height/margins, etc.).

screen shot 2016-02-15 at 20 23 12 vs. screen shot 2016-02-15 at 20 23 22

xPaw:
Just a thought, what if he got rid of that close button completely? It would avoid accidentally clicking it while trying to focus the channel. There's already a dedicated Leave button when channel is in focus, and #9 adds a way to close channel in a context menu.

YaManicKill:
In general, I have been in favour of that for a while. I don't know what the close button on the sidebar adds, and it has caused many users problems with accidentally clicking them.

dgw:
The discussion in erming/shout#558 is probably relevant. I know a lot of people on this fork will remember it, but it's worth formally mentioning.

astorije:
Oh boy what a thread, no wonder why I dropped it :-)

Considering we are already talking about solutions for a different issue here, here is my proposal: let's use `&times;` / `×` for now and reconsider (to remove it, most probably) once the context menu makes his way!

xPaw:
Ok, I've replaced it with &times;

chrome_2016-02-17_10-46-53

astorije:
@xPaw, not a huge deal, but it appears the cross is not vertically centered in the box:

screen shot 2016-02-17 at 22 20 42

Do you think you could take a quick look at that? I won't hold off my :+1: if you cannot find in a reasonable amount of time, but would be nice if you could take a quick look.

Also, the cross now appears green on server close instead of white-ish. I'm guessing it's a side effect of going from image to text, did you leave it on purpose? If you can set it to the same color than the rest, would be great for consistency, but here as well I won't hold off an approval.

Thanks!

xPaw:
@astorije Vertical positioning is a bitch, it was fine on my end, but hopefully I fixed it properly.

And I fixed the green color, the close button is always white.

astorije:
Hey @xPaw, no big deal on this one, would be great if you can wait for my definitive :+1: (or whoever's made you original comments / started review / assigned the PR) next time before merging :-)

Also, just FYI I'm removing the `second review needed` label on merged PRs.

Issue #45, opened 2016-02-15
Non-colored nicknames should be disabled and colors should be theme-specific


astorije:
FYI, I haven't looked to see how colors for the colored nicknames are handled.

However, in terms of UX, it's very noisy and insignificant to have a checkbox to enable colored nicknames.

This checkbox should be removed, enabling colored nicknames by default. Then, colors should be set in the CSS file (if it's not already the case) so that a theme can decide they want to disable colored nicknames by setting the same color for the whole range.
This also allows theme creators to have full control over the nickname colors (again, if it's not already the case).

xPaw:
> This checkbox should be removed, enabling colored nicknames by default. 

Strongly disagree. I don't like colored nicknames in any client, and wouldn't want to modify the theme to change it.

dubzi:
If we want to set the colors in the CSS, how many different colors do we want to introduce? Right now I don't think there's a limit since they're randomly generated based on the username. But adding 256 color classes to a css file is a bit overkill I guess. Could we settle somewhere around 30?

easymac:
Regarding CSS colors:

It's a slightly less-than-trivial decision to make here as there is no real standard for IRC colors.

We're already enumerating the 16 mIRC colors [in style.css](https://github.com/thelounge/lounge/blob/master/client/css/style.css#L992-L1028). As far as I can tell, most clients only support these 16 colors. However, HexChat supports 32 colors. [Assuming this Shout issue](https://github.com/erming/shout/issues/583) made it to Lounge, we're probably mapping the additional 16 colors down to our original 16.

I propose that we take advantage of this existing color pallet for nickname colors as well. I've implemented this locally, and [I'm very happy with the result](http://i.imgur.com/Juz6u1H.png). I would *much* rather have theme-defined nickname colors than any random nickname colors as the chosen pallet is very much part of the theme's appearance.

@xPaw raised the concern on IRC that this might not be enough colors; perhaps to that end we could lean into the 32-color pallet that HexChat defines and not only support more colors in chat but have 32 options for nicknames.

YaManicKill:
So, this should be partially fixed by #325. 

We need to agree on whether we remove the non-colour setting to close out this ticket.

I suggest we do not remove the setting, but we default to colours. Anyone in an objection to that? I don't think we want to remove the config to disable colours.

YaManicKill:
Assuming that it is just changing the default option for coloured nicknames, that's very easy so I've added the up for grabs label.

[...]

PR #49, opened 2016-02-15
update default settings to be more usable than technical

jancborchardt:
Reasoning:
- enable colored nicknames by default so people can be more easily distinguished. In the future we should try to have something like avatars, fed from email/gravatar/profile or something similar.
- show notification badge instead of notification sound to be less distracting
- do not show joins/parts/quits/nickchanges: just not that interesting and mostly cluttering up the history. In the future we could group these messages similar to how IRCCloud does it, but meanwhile this is just too noisy.
- do not show mode: technical detail – can be shown on hover or whatever, but has no place there by default.

What do you think @YaManicKill @astorije?

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -406,17 +406,17 @@ $(function() {
 
 	var settings = $("#settings");
 	var options = $.extend({
-		badge: false,
-		colors: false,
-		join: true,
+		badge: true,
+		colors: true,
+		join: false,
 		links: true,
-		mode: true,
+		mode: false,
 		motd: false,
-		nick: true,
-		notification: true,
-		part: true,
+		nick: false,
+		notification: false,
+		part: false,
 		thumbnails: true,
-		quit: true,
+		quit: false,
 		notifyAllMessages: false,
 	}, $.cookie("settings"));
 


dgw:
My takes on these, respectively:

- Good change; nick colors *always* make chat more readable.
- I have no particular opinion on this one way or the other, as long as it's immediately clear to the user that sound is available if desired. (Though I do prefer chats to make sound when not focused. Counterexample: Support chats that do not make noise often leave me forgetting to check their window for long periods. Not the same thing, but it is *much* easier to forget about a silent chat.)
- It's better IMO to show things by default, and let the user hide what they don't want to see, rather than hide by default and make the user discover what other details are available if they want more.
- See previous.

lpoujol:
Desktop notifications (aka badge in the settings) aren't simply activated by setting it to true. The code has to ask the permission to the browser if wether or not it can send desktop notifications (only the first time).

YaManicKill:
Similarly to @dgw, in order

* Yep, I agree with this. Easy to disable if they want, but a good default
* I agree with this if the setting did what it was meant to do. It's not the badge jus tnow, it's the desktop notifications. So I don't think this is sensible to have as a default just now.
* Yeah, I agree with having these set to enabled to start with. It can be confusing to not have them by default. Although, yeah I do plan to get them merged at some point, because that would be useful, and I really want it as a feature
* I'd be ok with this one.

astorije:
Will take a look, thanks!

MaxLeiter:
The one downside to this is that on certain themes colored nicknames are terrible. I believe @xPaw made an issue at some point about rewriting them so theyre theme specific, and if that ever happens then :+1: from me

xPaw:
> do not show joins/parts/quits/nickchanges: just not that interesting and mostly cluttering up the history.

I disagree with this. If the user finds it clutter, they can disable it. It shouldn't be the other way around.

astorije:
Alright, my turn:

> @jancborchardt: enable colored nicknames by default so people can be more easily distinguished.
> @dgw: Good change; nick colors always make chat more readable.
> @YaManicKill: Yep, I agree with this. Easy to disable if they want, but a good default

Yes.

> @MaxLeiter: The one downside to this is that on certain themes colored nicknames are terrible. I believe @xPaw made an issue at some point about rewriting them so theyre theme specific, and if that ever happens then from me

Are you talking about #45 ? You are right if colors aren't theme-specific at the moment, but still, I agree this should be enabled by default for the reasons mentioned above.

> @jancborchardt: show notification badge instead of notification sound to be less distracting
> @lpoujol: The badge setting isn't changing the badge, but the desktops notifications.
> Also, desktop notifications aren't simply activated by setting it to true. The code has to ask the permission to the browser if wether or not it can send desktop notifications (only the first time).
> @YaManicKill: I agree with this if the setting did what it was meant to do. It's not the badge jus tnow, it's the desktop notifications. So I don't think this is sensible to have as a default just now.

As said here, while have more sensible defaults would be appreciated, the checkbox itself is messed up. There is work by @lpoujol (#28) that needs attention and it should fix this. I suggest you drop the change for now and we discuss defaults either there or later. What do you think, @jancborchardt?

> @jancborchardt: do not show joins/parts/quits/nickchanges: just not that interesting and mostly cluttering up the history.
> @dgw: It's better IMO to show things by default, and let the user hide what they don't want to see, rather than hide by default and make the user discover what other details are available if they want more.
> @YaManicKill: Yeah, I agree with having these set to enabled to start with. It can be confusing to not have them by default.
> @xPaw: I disagree with this. If the user finds it clutter, they can disable it. It shouldn't be the other way around.

In general, we should the pattern of "simple things for casual users; advanced things for advanced users" (and, let's face it, all those who have commented here are advanced enough to be involved :P). What that means is that things should come simple and elegant by default and, as you mature using the project and the IRC world, you want to get deeper and fiddle with more advanced features/settings. A good example (known for their involvement in user experience) is Mac OS: by default it's a very slick and well designed interface, yet very simple, and usually the more advanced you are as a user, the more you'll know where to look to customize this.

That being said, even though I tend to agree with @jancborchardt regarding joins/parts/quits (because if you want to know who is there, we have the user list), I don't have a strong objection to leave it as is *at the moment* due to @xPaw's work on actions. But as much as I like having it displayed for me personally, it **is** clutter for the new user!
Nick changes should stay as there is no easy way to track it otherwise without this.

> Although, yeah I do plan to get them merged at some point, because that would be useful, and I really want it as a feature

Yes for a single checkbox for joins/parts/quits!

> @jancborchardt: do not show mode: technical detail – can be shown on hover or whatever, but has no place there by default.
> @YaManicKill: I'd be ok with this one.

This is a yes, definitely. Again, think for the user, and the new user for that matter.
You can also see these (in plural, means this way of thinking in general) as like having a software that, by default, outputs a sh*t-ton of information the first time you ever run it, whatever you run. No, what you want displayed the first time is the right amount for someone new, and run `-v` or -`v...v` as you mature or need more information.

astorije:
@jancborchardt, any update on this?

Additionally, I would enable the MOTD by default. While it's usually lengthy, it's actually something aimed at new users, like rules, how to contact, documentation links, etc. so very useful by default.

Jay2k1:
My two cents about hiding modes and events. Who is Lounge aimed at?

I think people who install an IRC client on a server, shell, VPS etc. can be considered "advanced users" already. I also think people who use IRC *at all* do have more IT knowledge than the average person and thus can be considered advanced (except those who find a webchat link on a random website). Some users may still prefer to not see certain events, but personally I'd find it quite irritating if certain things would just be missing from the chat until I find out they're actively being hidden by my client and I have to look for the setting to unhide them.

Sure, Mac OS X is well designed and yet simple to use, but it has a much broader audience than IRC users who install web applications on servers.

I totally agree with @dgw here. Things should be enabled by default, with an option to hide it if you want. Or maybe have an option like "HideEventsByDefault: true/false" in the config so the user can decide on their own at install time.

astorije:
> Who is Lounge aimed at?

Two very distinct groups of people: administrators and users. While your assumptions are potentially true for administrators, it is less the case for non-admin users, that is people who have an account on The Lounge. These can be any groups of people (family, friends, colleagues outside of IT team). Also, I'd add new IRC users, for which this world is already cryptic enough as it is, compared to other, non-decentralized and non-standardized options.

> (except those who find a webchat link on a random website)

As a matter of fact, I have been involved in an organization federating several hundreds of users spread across dozens of companies. While they/we were using IRC as a mean to connect and discuss, most were not technical and were using IRC as they would use any other chat. Relying on users to be advanced by default is closing our own doors.

But we are losing the focus here, and bikeshedding over 8 non-destructive lines of code. The important thing to note is that these *are not* changing anyone's current experience but new users. And in fact, customizing settings is **one** click away, not something like editing a hidden config file or anything. Let's be honest here: **everyone** on this very thread, advanced users, have opened the "Settings" tab within the first few minutes of usage.
In general, following an adaptive/learning experience where advanced users set their custom settings (again, one click!) is not only a very, very good UX practice, it's also already the case: The Lounge is very user-friendly (and should stay that way), and no matter what we try to enforce, non-novice users will always reach for the settings :-)

That being said, I have debated way more on this topic than I should have for an 8-line change :-) @jancborchardt, my final word is: change the settings for join/part/quit to satisfy @YaManicKill, @xPaw and others (I'd prefer the opposite but not strongly so I'll comply with the majority), for the badge at least until #28 ships, for the MOTD (see above), for the nicks, and you'll have my +1, probably @YaManicKill's and @xPaw's one too, by reading their comment.

astorije:
@jancborchardt, ping?

astorije:
Unless @jancborchardt jumps in, I will close this PR and re-open it to take into account the comments above.
I believe @thelounge/maintainers have given their opinion and the discussion above finds common ground among us.

jancborchardt:
Sorry folks, am busy with the release of the next ownCloud version until Monday. Will update the PR right afterwards. :)

astorije:
Great @jancborchardt, talk to you then! And say hi to Frank :P

astorije:
@jancborchardt, re-ping? :-)

xPaw:
I'll close this as there was no activity. And now that @maxpoulin64 is overhauling settings in #275 it's going to conflict even more.

jancborchardt:
Yup, sorry I gotta be honest here and say in the end the discussion grew a bit big for my taste. I really would like to contribute more to Lounge design-wise but I only have so much time to spend. :\

PR #57, opened 2016-02-17
frontend password change functionality

diddledan:
- refactor clientManager.js to allow configuration parsing as a serparate
  function.
  - refactor clientManager.js to add configuration writing function.
  - add new plugin for /setpass command.
  - add server.js changes to allow for new "profile" screen
  - add password change ui to "profile" screen
  - modify css, front-end shout.js and index.html to add "profile" ui
  - refactor client.js to use new clientManager functionality for saving
    the configuration files

duplicates erming/shout#506

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1116,6 +1116,19 @@ button,
 	line-height: 1.8;
 }
 
+#settings #change-password .error,
+#settings #change-password .success {
+	margin-bottom: 1em;
+}
+
+#settings #change-password .error {
+	color: #e74c3c;
+}
+
+#settings #change-password .success {
+	color: #2ecc40;
+}
+
 #form {
 	background: #eee;
 	border-top: 1px solid #ddd;
--- a/client/index.html
+++ b/client/index.html
@@ -265,6 +265,31 @@
 									Enable notification for all messages
 								</label>
 							</div>
+							<% if (!public) { %>
+							<div id="change-password">
+								<form action="" method="post">
+									<div class="col-sm-12">
+										<h2>Change password</h2>
+									</div>
+									<div class="col-sm-12">
+										<label for="old_password" class="sr-only">Enter current password</label>
+										<input type="password" name="old_password" class="input" placeholder="Enter current password">
+									</div>
+									<div class="col-sm-12">
+										<label for="new_password" class="sr-only">Enter desired new password</label>
+										<input type="password" name="new_password" class="input" placeholder="Enter desired new password">
+									</div>
+									<div class="col-sm-12">
+										<label for="verify_password" class="sr-only">Repeat new password</label>
+										<input type="password" name="verify_password" class="input" placeholder="Repeat new password">
+									</div>
+									<div class="col-sm-12 feedback"></div>
+									<div class="col-sm-12">
+										<button type="submit" class="btn">Change password</button>
+									</div>
+								</form>
+							</div>
+							<% } %>
 							<div class="col-sm-12">
 								<h2>About The Lounge</h2>
 							</div>
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -112,6 +112,31 @@ $(function() {
 			.show();
 	});
 
+	socket.on("change-password", function(data) {
+		var passwordForm = $("#change-password");
+		if (data.error || data.success) {
+			var message = data.success ? data.success : data.error;
+			var feedback = passwordForm.find(".feedback");
+
+			if (data.success) {
+				feedback.addClass("success").removeClass("error");
+			} else {
+				feedback.addClass("error").removeClass("success");
+			}
+
+			feedback.text(message).show();
+			feedback.closest("form").one("submit", function() {
+				feedback.hide();
+			});
+		}
+		passwordForm
+			.find("input")
+			.val("")
+			.end()
+			.find(".btn")
+			.prop("disabled", false);
+	});
+
 	socket.on("init", function(data) {
 		if (data.networks.length === 0) {
 			$("#footer").find(".connect").trigger("click");
@@ -734,7 +759,7 @@ $(function() {
 	});
 
 	var windows = $("#windows");
-	var forms = $("#sign-in, #connect");
+	var forms = $("#sign-in, #connect, #change-password");
 
 	windows.on("show", "#sign-in", function() {
 		var self = $(this);
@@ -761,6 +786,8 @@ $(function() {
 			.end();
 		if (form.closest(".window").attr("id") === "connect") {
 			event = "conn";
+		} else if (form.closest("div").attr("id") === "change-password") {
+			event = "change-password";
 		}
 		var values = {};
 		$.each(form.serializeArray(), function(i, obj) {
--- a/src/client.js
+++ b/src/client.js
@@ -1,7 +1,6 @@
 var _ = require("lodash");
 var Chan = require("./models/chan");
 var crypto = require("crypto");
-var fs = require("fs");
 var identd = require("./identd");
 var log = require("./log");
 var net = require("net");
@@ -51,14 +50,15 @@ var inputs = [
 	"whois"
 ];
 
-function Client(sockets, name, config) {
+function Client(manager, name, config) {
 	_.merge(this, {
 		activeChannel: -1,
 		config: config,
 		id: id++,
 		name: name,
 		networks: [],
-		sockets: sockets
+		sockets: manager.sockets,
+		manager: manager
 	});
 	var client = this;
 	crypto.randomBytes(48, function(err, buf) {
@@ -221,6 +221,21 @@ Client.prototype.connect = function(args) {
 	});
 };
 
+Client.prototype.setPassword = function(hash) {
+	var client = this;
+	client.manager.updateUser(client.name, {password:hash});
+	// re-read the hash off disk to ensure we use whatever is saved. this will
+	// prevent situations where the password failed to save properly and so
+	// a restart of the server would forget the change and use the old
+	// password again.
+	var user = client.manager.readUserConfig(client.name);
+	if (user.password === hash) {
+		client.config.password = hash;
+		return true;
+	}
+	return false;
+};
+
 Client.prototype.input = function(data) {
 	var client = this;
 	var text = data.text.trim();
@@ -353,9 +368,6 @@ Client.prototype.save = function(force) {
 		return;
 	}
 
-	var name = this.name;
-	var path = Helper.HOME + "/users/" + name + ".json";
-
 	var networks = _.map(
 		this.networks,
 		function(n) {
@@ -364,29 +376,6 @@ Client.prototype.save = function(force) {
 	);
 
 	var json = {};
-	fs.readFile(path, "utf-8", function(err, data) {
-		if (err) {
-			console.log(err);
-			return;
-		}
-
-		try {
-			json = JSON.parse(data);
-			json.networks = networks;
-		} catch (e) {
-			console.log(e);
-			return;
-		}
-
-		fs.writeFile(
-			path,
-			JSON.stringify(json, null, "  "),
-			{mode: "0777"},
-			function(err) {
-				if (err) {
-					console.log(err);
-				}
-			}
-		);
-	});
+	json.networks = networks;
+	client.manager.updateUser(client.name, json);
 };
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -29,18 +29,14 @@ ClientManager.prototype.loadUsers = function() {
 
 ClientManager.prototype.loadUser = function(name) {
 	try {
-		var json = fs.readFileSync(
-			Helper.HOME + "/users/" + name + ".json",
-			"utf-8"
-		);
-		json = JSON.parse(json);
+		var json = this.readUserConfig(name);
 	} catch (e) {
 		console.log(e);
 		return;
 	}
 	if (!this.findClient(name)) {
 		this.clients.push(new Client(
-			this.sockets,
+			this,
 			name,
 			json
 		));
@@ -91,6 +87,50 @@ ClientManager.prototype.addUser = function(name, password) {
 		throw e;
 	}
 	return true;
+};
+
+ClientManager.prototype.updateUser = function(name, opts) {
+	var users = this.getUsers();
+	if (users.indexOf(name) === -1) {
+		return false;
+	}
+	if (typeof opts === "undefined") {
+		return false;
+	}
+	var path = Helper.HOME + "/users/" + name + ".json";
+	var user = {};
+
+	try {
+		user = this.readUserConfig(name);
+		_.merge(user, opts);
+	} catch (e) {
+		console.log(e);
+		return;
+	}
+
+	fs.writeFileSync(
+		path,
+		JSON.stringify(user, null, " "),
+		{mode: "0777"},
+		function(err) {
+			if (err) {
+				console.log(err);
+			}
+		}
+	);
+	return true;
+};
+
+ClientManager.prototype.readUserConfig = function(name) {
+	var users = this.getUsers();
+	if (users.indexOf(name) === -1) {
+		return false;
+	}
+	var path = Helper.HOME + "/users/" + name + ".json";
+	var user = {};
+	var data = fs.readFileSync(path, "utf-8");
+	user = JSON.parse(data);
+	return user;
 };
 
 ClientManager.prototype.removeUser = function(name) {
--- a/src/server.js
+++ b/src/server.js
@@ -107,6 +107,51 @@ function init(socket, client, token) {
 				client.connect(data);
 			}
 		);
+		if (!config.public) {
+			socket.on(
+				"change-password",
+				function(data) {
+					var old = data.old_password;
+					var p1 = data.new_password;
+					var p2 = data.verify_password;
+					if (typeof old === "undefined" || old === "") {
+						socket.emit("change-password", {
+							error: "Please enter your current password"
+						});
+						return;
+					}
+					if (typeof p1 === "undefined" || p1 === "") {
+						socket.emit("change-password", {
+							error: "Please enter a new password"
+						});
+						return;
+					}
+					if (p1 !== p2) {
+						socket.emit("change-password", {
+							error: "Both new password fields must match"
+						});
+						return;
+					}
+					if (!bcrypt.compareSync(old || "", client.config.password)) {
+						socket.emit("change-password", {
+							error: "The current password field does not match your account password"
+						});
+						return;
+					}
+					var salt = bcrypt.genSaltSync(8);
+					var hash = bcrypt.hashSync(p1, salt);
+					if (client.setPassword(hash)) {
+						socket.emit("change-password", {
+							success: "Successfully updated your password"
+						});
+						return;
+					}
+					socket.emit("change-password", {
+						error: "Failed to update your password"
+					});
+				}
+			);
+		}
 		socket.on(
 			"open",
 			function(data) {


astorije:
Hey @diddledan, thanks for this! (Cc @JocelynDelalande as he started reviewing this on Shout)

However, I see that maaaassive diff, I'm a bit scared considering the feature. In particular:

1. Is there a specific reason why you need to upgrade Bootstrap from 3.2 to 3.5?
1. I believe the `client/index.html.orig` file was committed here by accident. Could you confirm and remove please?

I believe once we clean this up a bit it will be much easier to review :-)

client/index.html
```diff
@@ -164,6 +165,42 @@ <h1 class="title">Connect</h1>
...
+								<h2>Change password</h2>
+							</div>
+							<div class="col-sm-12">
+                                <label aria-label="Enter current password">
+                                    <span class="sr-only">Enter current password</span>
+                                </label>
+                                <input class="input" type="password" name="old_password" placeholder="Enter current password">
+                            </div>
+                            <div class="col-sm-12">
+                                <label aria-label="Enter desired new password">
```
xPaw: Mixed spaces, use tabs.

client/index.html
```diff
@@ -293,6 +330,15 @@ <h1 class="title">Settings</h1>
 	</div>
 	</div>
 
+	<template id="alert-template">
```
xPaw: All other templates are handlebars in `views` folder, use that instead.

src/plugins/inputs/setpass.js
```diff
@@ -0,0 +1,41 @@
+var bcrypt = require("bcrypt-nodejs");
+var Msg = require("../../models/msg");
+
+module.exports = function(network, chan, cmd, args) {
+	if (cmd !== "setpass") {
```
xPaw: Do we really need a command to change password? It might also be confused for a server command.

src/server.js
```diff
@@ -110,6 +110,49 @@ function init(socket, client, token) {
...
+					socket.emit("profile", {
+						error: "The current password field does not match your account password"
+					});
+					return;
+				}
+				var salt = bcrypt.genSaltSync(8);
+				var hash = bcrypt.hashSync(p1, salt);
+				if (client.setPassword(hash)) {
+					socket.emit("profile", {
+						success: "Successfully updated your password :-)"
```
xPaw: While emotes are cool, I don't think they fit in here (same in error msg below).

JocelynDelalande:
Thanks for careful review (I have to admit reviewing latest commits, I focused on UI), I got my +1 a bit too early on shout project, there is still work to be done, obviously

@diddledan be brave ;-)

src/plugins/inputs/setpass.js
JocelynDelalande: agreed that it's confusing and not needed.

src/server.js
diddledan: I've removed the emotes

client/index.html
diddledan: fixed

src/plugins/inputs/setpass.js
diddledan: some people like the ability to not have to click ui elements to perform actions. That's why IRC has historically has slash-commands. there is no requirement for a command akin to `/command some arguments` to be forwarded to the IRC server as `COMMAND :some arguments`! *many* IRC clients use slash-commands for client-specific functionality, and I added this command based on that premise.

diddledan:
the replaced/upgraded bootstrap was done to add alert-box functionality which was missing from the previous bundle.

src/plugins/inputs/setpass.js
xPaw: While your argument is true, I don't know how often you have to change your password to warrant having a command for it.

xPaw:
Updating bootstrap shouldn't be part of your PR. As for alerts, I'm not fond of how alerts look in bootstrap, maybe using a separate library for the task would be better?

http://kamranahmed.info/toast
https://t4t5.github.io/sweetalert/

Having some alerts like that will come in handy later on when we will handle IRC errors and disconnections (e.g. show an alert when you get disconnected from a network)

JocelynDelalande:
> Updating bootstrap shouldn't be part of your PR.

Depends. As it's motivated, I would be okay with that update to be part of that PR, what do others think ?

BS is [(trying to) follow semver](https://github.com/twbs/bootstrap#versioning), so this version jump is supposed to be backward compatible.

> As for alerts, I'm not fond of how alerts look in bootstrap, maybe using a separate library for the task would be better?
> 
> http://kamranahmed.info/toast
> https://t4t5.github.io/sweetalert/
> 
> Having some alerts like that will come in handy later on when we will handle IRC errors and disconnections (e.g. show an alert when you get disconnected from a network)

BS alerts are fine IMHO (I didn't said it's the best), it will still be time to rework alerts later, let's not make things too hard for this PR or it may never make it to master.

xPaw:
> Depends. As it's motivated, I would be okay with that update to be part of that PR, what do others think ?

In that case, can we get alerts added to the current bootstrap version to use? Would make for a simpler diff.

MaxLeiter:
I think it should be its own PR, or at least a separate commit

astorije:
First of all, @diddledan, this is great. I do believe it is not ready at the moment, but I am strongly in favor and will make sure we can get this to a mergeable state.
What is really cool is that the feature itself works, I can indeed change my password from the UI, which is great. This means that while we are striving to make this PR as awesome as we can, people can already run it locally if it's an absolute necessity for their client! And for that I say :clap: :clap: 

> Depends. As it's motivated, I would be okay with that update to be part of that PR, what do others think ?

I do not agree with this. Such an upgrade requires testing and therefore should be isolated. As a matter of fact, I noticed the following bug:

Before | After
--- | ---
screen shot 2016-02-17 at 22 46 09 | screen shot 2016-02-17 at 22 45 50

This bug might not be the upgrade itself, but again by cumulating upgrades, refactors and features, it's very hard for us (at least for me) to see what's wrong in a PR with 1200+ changes.

And while this PR does upgrade Bootstrap from 3.2 to 3.5, as @xPaw stated, it also adds components that were removed using [Bootstrap's customized builds](https://getbootstrap.com/customize/), making things a bit harder.

As a matter of fact, if the extra 1k changes are simply to add that dismissible alert-box to display success/failure, I'd say we are doing something wrong.
My proposal is the following:
- Let's use what we already have accessible in Bootstrap or the custom CSS or whatever so that gets in quickly. At the very least, if nothing can work, we can use a JS `alert()`.
- In the meantime, we design a notification/alert/flash/whatever-it's-called feature that looks more appropriate for the current design, that can be modified from a theme and that can be re-used by other parts of the project (this is going to be a crucial point in the package system, as such flash container will have its own API there).
- Then we can update that part of the code when it's ready.
As @xPaw said, this is not a feature people are going to use every day, so a JS `alert()`, while definitely ugly and a temporary solution-until-better, does not throw me off. Again, we did not need such feature before, and we are going to reuse it, so let's take the time to design it right.
What do you think?

@diddledan, a few other things:
- I noticed the emojis are still there, either because you didn't remove them all in the code, or because you didn't submit the updated grunted templates:
  screen shot 2016-02-17 at 22 54 17
- This PR also affects tooltips, that do not get hidden well anymore (until a click is emitted somewhere):
  screen shot 2016-02-17 at 22 53 04
- I am not in favor of adding a "My Profile" element to the menu. There are not many settings at the moment, they need to be designed (not in terms of UI but more broadly) and there is just one element in the profile at the moment. Can you move the password change to the settings instead please? Also, the user icon is already used to display a nick change for now.

That's all I got for now, but I will come back with more once we get to a smaller diff.

Keep up the good work, that's a great start!

src/server.js
JocelynDelalande: > I've removed the emotes

… But forgot to push ? :-)

JocelynDelalande:
Thx to everyone for involvement :-)

@astorije showed that bootstrap adhesion to semver rule "no breaking change between two minors" is quite loose, so better be cautious.

> As a matter of fact, if the extra 1k changes are simply to add that dismissible alert-box to display success/failure, I'd say we are doing something wrong.
> My proposal is the following:
> 
>     Let's use what we already have accessible in Bootstrap or the custom CSS or whatever so that gets in quickly. At the very least, if nothing can work, we can use a JS alert().

It turns out that there is no need to update bootstrap to get the alerts, it seems to have been present [for a while](https://github.com/twbs/bootstrap/commits/master/less/alerts.less).

 Not to over-grow the scope of the PR, nor introduce breaking changes, I'd suggest the following :

- remove the BS upgrade
- modify the BS customized build [as @xPaw suggested](https://github.com/thelounge/lounge/pull/57#issuecomment-185346643) (better in a dedicated commit).

@xPaw @astorije @diddledan ok with that ?

>     In the meantime, we design a notification/alert/flash/whatever-it's-called feature that looks more appropriate for the current design, that can be modified from a theme and that can be re-used by other parts of the project (this is going to be a crucial point in the package system, as such flash container will have its own API there).
>     Then we can update that part of the code when it's ready. As @xPaw said, this is not a feature people are going to use every day, so a JS alert(), while definitely ugly and a temporary solution-until-better, does not throw me off. Again, we did not need such feature before, and we are going to reuse it, so let's take the time to design it right. What do you think?

I agree with your 2 stages approach : doing "best-effort" with what we have here, keeping it simple, and designing propper notification later, with global thinking.

JocelynDelalande:
> I am not in favor of adding a "My Profile" element to the menu. There are not many settings at the moment, they need to be designed (not in terms of UI but more broadly) and there is just one element in the profile at the moment. Can you move the password change to the settings instead please? Also, the user icon is already used to display a nick change for now.

Not strongly opinionated on this point. But I'm ok with the proposal to slip it into settings if others are, as it's a common practice, I think users will find it.

astorije:
> It turns out that there is no need to update bootstrap to get the alerts, it seems to have been present for a while.

I understand alerts have been removed for this very version of Bootstrap 3.2 we are shipping, using the customized build form I have mentioned earlier, which is why @diddledan added them, and which is why I am suggesting a dumb `alert()` at the moment.

JocelynDelalande:
> I understand alerts have been removed for this very version of Bootstrap 3.2 we are shipping, using the customized build form I have mentioned earlier, which is why @diddledan added them, and which is why I am suggesting a dumb alert() at the moment.

Hm I think there is something I don't understand, let's state what I understand, and stop me where I'm wrong :

- We are using a slimmed down custom build of bootstrap 3.2
- This custom build doesn't include alert features
- Can't we just add the alert feature to the custom build, staying at version 3.2 ?

astorije:
> We are using a slimmed down custom build of bootstrap 3.2
> This custom build doesn't include alert features

Correct.

> Can't we just add the alert feature to the custom build, staying at version 3.2 ?

We can, but should not as stated above. `alert()` is the way to go for this very PR.

diddledan:
I've reverted the bootstrap, replaced the alerts with the `window.alert()` function, and moved the ui into settings thereby removing the "profile" page altogether.

client/index.html
```diff
@@ -265,6 +265,36 @@ <h1 class="title">Settings</h1>
 									Enable notification for all messages
 								</label>
 							</div>
+							<% if (!public) { %>
+							<div id="change-password">
+								<form action="">
```
xPaw: Add `method="post"`

src/server.js
```diff
@@ -110,6 +110,49 @@ function init(socket, client, token) {
 			}
 		);
 		socket.on(
```
xPaw: This shouldn't be registered if the instance is public.

client/css/style.css
```diff
@@ -387,7 +387,7 @@ button, .user {
 	content: "\f013";
 }
 #footer .sign-out:before {
-	content: "\f13e";
+	content: "\f08b";
```
xPaw: Any reason for this?

diddledan: I don't know why that's changed o_O

astorije:
> I've reverted the bootstrap, replaced the alerts with the window.alert() function, and moved the ui into settings thereby removing the "profile" page altogether. (the smilie faces should be gone now too!)

Now that's a much nicer diff, I'll do a review when I am outside of work :-)

... Buuuuuut you are going to hate me if not already :-)
I just remember that we already had a feedback for a form (only negative, red feedbak for now, so you could reuse one of the green colors from the UI, if there is no colors set for positive feedback)! It's faaaar from perfect and definitely needs thinking and design, but it's there, and using the same mechanism would allow for more consistency:

screen shot 2016-02-18 at 12 05 14

Again, will make a bigger review later :-)

client/index.html
```diff
@@ -1,4 +1,4 @@
-<!doctype html>
+﻿<!doctype html>
```
JocelynDelalande: what changed exactly here ? is it github tripping ?

diddledan: it looks like I accidentally saved the file with a UTF-8 BOM identifier. I'll try to remove it in a mo

client/css/style.css
diddledan: aah yes, I remember now - the current icon is a lock which doesn't really indicate "exit" - this should be considered as a separate issue so I'll revert the icon shortly

diddledan:
I've updated to make-use of a method of alert similar to the login screen (plagiarised therefrom :-p)

client/css/style.css
```diff
@@ -915,6 +915,13 @@ button, .user {
 	padding-top: 2px;
 	line-height: 1.8;
 }
+#settings #change-password .error {
+	color: #e74c3c;
+	margin-bottom: 1em;
+}
+#settings #change-password .error.success {
+	color: #2ecc40;
```
astorije: - Why the `margin-bottom` for the error but not the success message? It doesn't seem necessary for none I think, but I can be wrong.
- `.error.success`? :-)

client/index.html
```diff
@@ -265,6 +265,38 @@ <h1 class="title">Settings</h1>
...
+							<div id="change-password">
+								<form action="" method="post">
+									<div class="col-sm-12">
+										<h2>Change password</h2>
+									</div>
+									<div class="col-sm-12">
+										<label aria-label="Enter current password">
+											<span class="sr-only">Enter current password</span>
+										</label>
+										<input class="input" type="password" name="old_password" placeholder="Enter current password">
```
astorije: This is not exactly how to use use `label` and assistive technology attributes like `sr-only` or `aria-label`:
- A `label` must be tied to its `input` with `for`
- The `aria-label` must be on the `input` instead of the `label`

Additionally, `sr-only` and `aria-label` serve the same purpose, so only one of them is enough. Remove the `label`s, add `aria-label="Enter current password"` to the `input`s and that should do :-)

client/index.html
```diff
@@ -265,6 +265,38 @@ <h1 class="title">Settings</h1>
...
+										</label>
+										<input type="password" name="new_password" class="input" placeholder="Enter desired new password">
+									</div>
+									<div class="col-sm-12">
+										<label aria-label="Repeat new password">
+											<span class="sr-only">Repeat new password</span>
+										</label>
+										<input type="password" name="verify_password" class="input" placeholder="Repeat new password">
+									</div>
+									<div class="col-sm-12 error">
```
astorije: `feedback` instead of `error`?

src/clientManager.js
```diff
@@ -93,6 +89,50 @@ ClientManager.prototype.addUser = function(name, password) {
...
+		_.merge(user, opts);
+	} catch (e) {
+		console.log(e);
+		return;
+	}
+
+	fs.writeFileSync(
+		path,
+		JSON.stringify(user, null, " "),
+		{mode: "0777"},
```
astorije: I never noticed the file was written with mode `0777` before, it's pretty sad...

astorije:
@diddledan, waouh, this is really great!! I left you a couple minor comments in the code, but overall it looks very good. This is very close, I'm excited!

In terms of code, there are a few refactors I'd like a second opinion about, maybe @xPaw or @YaManicKill could give it a shot?

src/client.js
```diff
@@ -46,18 +45,20 @@ var inputs = [
 	"quit",
 	"raw",
 	"services",
+	"setpass",
```
xPaw: Did you forget to remove this?

src/clientManager.js
xPaw: We should have `0600` or at least `0644` here, but that's a thing to fix in a separate PR.

xPaw:
Here's a question: Should the token be re-generated when the password is changed? This would invalidate all sessions, and require them to re-login *(re-send new token to current session)* with the new password the next time they try to use  your account.

src/clientManager.js
JocelynDelalande: Ouch, that's bad...

client/css/style.css
diddledan: to the first point, why a margin-bottom on .error but not .error.success: the second rule will inherit from the first so there is no reason to override the margin-bottom with an identical value when it is already applied by inheritance.

.error.success: that's a semantic issue I guess, but we can rename .error if need be.

src/client.js
diddledan: I did. oops :-)

client/index.html
astorije: It appears, after discussing with an accessibility expert, that I got it wrong: even though `label`s and `aria-label` serve the same purpose, `label`s are the preferred way to do so.

Therefore, you can fix your code with these steps:
- Remove `aria-label`
- Move `class="sr-only"` to the `label`s themselves
- Remove the now empty `span`s

client/css/style.css
astorije: Yes, it seems very messy to have elements that are both `.error` and `.success`. You should use one or the other, and if you do need a common class, you could call it `.feedback`. If it's just for common CSS properties though, `.error, .success { /* ... */ }` will do just fine.

astorije:
@xPaw, while I fully agree with you, here are my 2 cents:
- I believe this PR should make it to master without this addition, that we can add later on.
- I notice this is not the first time you mention something related to the session token. Do you think it would be worth relying on a third-party library for user management, like [`passport`](https://www.npmjs.com/package/passport) or anything else (if so, we need to list pros/cons to decide)? It seems to do gazillion things we do not need, but I thinking overall. Maybe I am completely wrong, or maybe this is not something we can do short-term, but I'd be happy to have your opinion.

@diddledan, would you mind rebasing with master and fixing what was given in comments (CSS class, `label`s and `setpass`) so we can proceed with the reviews? :-)

Thanks a lot for this, and hang in there, it's getting super close!

xPaw:
> I notice this is not the first time you mention something related to the session token. Do you think it would be worth relying on a third-party library for user management, like passport or anything else (if so, we need to list pros/cons to decide)? It seems to do gazillion things we do not need, but I thinking overall. Maybe I am completely wrong, or maybe this is not something we can do short-term, but I'd be happy to have your opinion.

That might be worth looking into.

diddledan:
I believe I've fixed all the issues most-recently raised (sr-only labels, extra "setpass" autocomplete reference now removed, and class names/styles)

client/js/lounge.js
```diff
@@ -734,7 +759,7 @@ $(function() {
 	});
 
 	var windows = $("#windows");
-	var forms = $("#sign-in, #connect");
+	var forms = $("#sign-in, #connect, #change-password");
```
astorije: [Off-topic of this PR]
I have no idea what's going on here, but this seems fishy. We might want to look into it and its usefulness when we clean up that file.

src/client.js
```diff
@@ -221,6 +221,21 @@ Client.prototype.connect = function(args) {
...
 };
 
+Client.prototype.setPassword = function(hash) {
+	var client = this;
+	client.manager.updateUser(client.name, {password:hash});
+	// re-read the hash off disk to ensure we use whatever is saved. this will
+	// prevent situations where the password failed to save properly and so
+	// a restart of the server would forget the change and use the old
+	// password again.
+	var user = client.manager.readUserConfig(client.name);
```
astorije: I'm not sure I understand that line and comment: if the password was not saved properly (so, not saved), why do we need to re-read the config, since the password will be unchanged? Also, doesn't restarting the server trigger a new config file read?

astorije:
Hi @diddledan,

When I'm trying to change the password now, I get no success or error message, could you take a look? :-/
Also, Travis CI breaks on CSS linting, do you mind fixing this too?

Not that it will hold off my +1, but I left you a comment in `Client.prototype.setPassword`, trying to fully understand things.

I should be all set after that :-)

astorije:
@YaManicKill, do you mind taking second review after @diddledan has fixed what's currently broken?

YaManicKill:
Yeah, I'll take a look in the next day or 2.

astorije:
Ok, then I think the next minor release (maybe tonight if I have time) will get out without this then. Hopefully @diddledan gets some spare time to put the final touch to this before the next next minor release gets out :-)

client/js/lounge.js
diddledan: AFAICT this is used to wire-up the on-submit event handlers on each of the three elements with those IDs, which happen to be `<form>` elements (the elements are selected separately as there's commas between each ID, and therefore will return a JQuery object referencing each of the elements with IDs `sign-in`, `connect`, and `change-password`). The event handler we wire-up onto this resultant set of elements calls `event.preventDefault()` which stops the form submittal from reloading the page. The form is submitted through the websocket or XHR instead?

src/client.js
diddledan: my thinking was to reload the entire user object from disk rather than just tweak the in-memory so that we are verified to be using the canonical configuration rather than a version which might be different to what's saved to disk.

astorije: I see, I'm good with that for now :-)

astorije:
Alright, excellent work @diddledan, I am finally :+1: 
:clap: 

Off to a second review now!

xPaw:
Looks fine to me code wise.

maxpoulin64:
Yep, I think we finally can merge this one. :+1: :clap:

JocelynDelalande:
Thaaaat is nice, thanks everyone :-), and congrats @diddledan !

PR #58, opened 2016-02-17
Only update the users list when needed

maxpoulin64:
Currently, for join/part/kick/nick/... the server will send an updated list of users and the client will re-render the list entirely. This ends up being a very expensive operation when joined on large channels and causes the client to slow down a lot.

This patch makes it so the user list is only updated when it is actually visible, avoiding lots and lots of user list re-renders on large channels.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -394,6 +394,19 @@ $(function() {
 	});
 
 	socket.on("users", function(data) {
+		var chan = chat.find("#chan-" + data.chan);
+
+		if (chan.hasClass("active")) {
+			socket.emit("names", {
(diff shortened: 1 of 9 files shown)


client/js/lounge.js
```diff
@@ -572,6 +585,11 @@ $(function() {
 			}
 		}
 
+		if (chan.hasClass("needs-names-refresh")) {
```
xPaw: Can you use `data` attribute instead?

xPaw:
:+1:

YaManicKill:
Yep, looks good to me. Merging :+1:

PR #60, opened 2016-02-17
Remove Node 0.10 from Travis CI

astorije:
This is needed for #43 to pass the CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ node_js:
 - '5.6'
 - '4.3'
 - '0.12'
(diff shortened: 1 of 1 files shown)


xPaw:
:+1:

We probably should also drop 0.12 later on.

astorije:
Yep, but not today  :-)

astorije:
@JocelynDelalande, ok with this?

JocelynDelalande:
:+1: (as there is almost no other test than linting for now).

PR #62, opened 2016-02-18
Moved client-side dependencies to NPM

MaxLeiter:
Currently using browserify, majority of client-side dependencies are now on NPM


package.json
```diff
@@ -27,13 +27,17 @@
   "license": "MIT",
   "dependencies": {
     "bcrypt-nodejs": "0.0.3",
+    "browserify": "^13.0.0",
```
astorije: `dependencies` must now specify fixed versions. For `devDependencies`, I forgot what @xPaw and I concluded on though.

package.json
```diff
@@ -27,13 +27,17 @@
...
     "commander": "2.9.0",
     "event-stream": "3.3.2",
     "express": "4.13.4",
+    "favico.js": "^0.3.10",
+    "handlebars": "^2.0.0",
+    "jquery-browserify": "^1.8.1",
     "lodash": "~2.4.1",
     "mkdirp": "0.5.1",
-    "moment": "2.11.2",
+    "moment": "^2.7.0",
```
astorije: Why this downgrade?

MaxLeiter: It did it automatically: The client shipped with moment 2.7, the server uses 2.11.2. I installed 2.7 to ensure client compatibility

YaManicKill: Moment follows semver, so upgrading to 2.11.2 shouldn't break anything. Infact, it was upgraded on the server just a couple of weeks or so ago.

package.json
```diff
@@ -27,13 +27,17 @@
...
     "commander": "2.9.0",
     "event-stream": "3.3.2",
     "express": "4.13.4",
+    "favico.js": "^0.3.10",
+    "handlebars": "^2.0.0",
+    "jquery-browserify": "^1.8.1",
     "lodash": "~2.4.1",
     "mkdirp": "0.5.1",
-    "moment": "2.11.2",
+    "moment": "^2.11.2",
```
YaManicKill: YEah, as @astorije, specify the fixed versions, so not "^" or "~"

client/js/lounge.js
```diff
@@ -50,7 +54,7 @@ $(function() {
 
 	$(".tse-scrollable").TrackpadScrollEmulator();
 
-	var favico = new Favico({
+	favico = new Favico({
```
YaManicKill: So, this is an issue here. You'll need to have the dependency at the top called Favico, and keep the var here, otherwise you are redefining the library to be a new variable, not great.

client/js/libs/jquery/bootstrap.js
```diff
@@ -1,3 +1,4 @@
+
```
YaManicKill: This is on npm as well, so you can grab it there.

MaxLeiter: I was (and am) unsure if the bootstrap JS was customized like the CSS

YaManicKill:
So, a couple of questions:

1. Does this work as is?
2. If not, what is the error?

The jquery stuff should be on npm as wlel, isn't it?

The handlebars templates can just live in the repository, as they are not libraries (I've never understood why they were in the libs/ directory, they are custom stuff for the app itself).

MaxLeiter:
@YaManicKill No, it doesn't work, this is the progress I made and I'm currently not sure how to go about finishing it. I didn't look into the Jquery libs yet, and I plan to leave the Handlebars as they are now. I think they should be moved in a separate PR; enough refactoring will occur when this is finished

package.json
xPaw: I was going for pinned versions in both.

client/js/libs/jquery/bootstrap.js
xPaw: Do we actually need the bootstrap.js for anything?

client/js/libs/handlebars/diff.js
```diff
@@ -1,5 +1,5 @@
 var diff;
```
xPaw: Do you mind looking why we have this global variable, and remove it if possible?

client/js/lounge.js
```diff
@@ -1,3 +1,7 @@
+var $ = require('jquery-browserify');
```
xPaw: Can you put all these variables inside the anonymous scope. We really wouldn't want to pollute the global namespace.

xPaw:
Looking good so far!

Did you forget to add `socket.io-client`?

MaxLeiter:
@xPaw : according to the socket.io-client [npm](https://www.npmjs.com/package/socket.io-client) page: 
> A standalone build of socket.io-client is exposed automatically by the socket.io server as

client/js/libs/handlebars/diff.js
MaxLeiter: As far as I can tell, there is no reason

MaxLeiter:
I think its supposed to work now, I'm just not sure how to setup grunt/browserify to compile everything correctly. Any ideas?

MaxLeiter:
Update: afaik, it should now work (but doesn't, and no errors are shown). By running `npm run build`, you grunt-browserify all of the files into 'libs.js', and minify/uglify them with grunt-uglify.

MaxLeiter:
Current status: Jquery plugins inside `libs.js` aren't working in `lounge.js`

package.json
```diff
@@ -13,6 +13,7 @@
   "scripts": {
     "start": "node index",
     "build": "grunt",
+    "dev": "concurrent 'grunt dev' 'npm start'",
```
YaManicKill: Beautiful.

For other people, this means we can use browserify without having to run a second command. So you just run "npm run dev" isntead of "npm start" when developing, and watchify will build the build again every time you change any file in the client. I use watchify on a couple of projects, and it's seriously so quick I don't even notice the build script.

Gruntfile.js
```diff
@@ -1,21 +1,35 @@
...
+
 	grunt.initConfig({
 		watch: {
 			files: libs,
 			tasks: ["uglify"]
 		},
+		browserify: {
+			dist: {
+		    	files: {
+		      		'client/js/compiled/lounge.js': [libs, 'client/js/libs/*.js', 'client/js/lounge.js', 'client/js/lounge.templates.js']
```
YaManicKill: If you require all the libs in lounge.js, you only need to do

'client/js/compiled/lounge.js': client/js/lounge.js

Which is much neater, imo.

MaxLeiter:
Current status: It works! Just need to update to be even with master and squish commits, then it should be good to go :)

YaManicKill:
Right, so, in general this looks pretty good. If you could do the 1 suggested change and rebase, I'll then have a play to make sure it works for me as well, and then I'll give my +.

Great work, @MaxLeiter I'm really happy this is going to be done.

YaManicKill:
Alright, @MaxLeiter sorry to do this to you, but we might end up not using browserify.

I'd suggest pausing any work on this as of just now until we get through discussion of what module system to use, etc. We have to make sure that it'll work with not only this, but packages. And one of the key things about packages is that users should be able to install client packages and have them available in the browser without restarting the server. Browserify wouldn't allow us to do this (without having a watch running, which isn't good for production).

@astorije Where do you want to have the discussion about this? I know you are investigating a couple of things but it would be good to have a central place for deciding this, and I'd really like to have us decide this in the next few days as a lot of things rely on it (this PR, the package system, #66 all require us to have the module loader in place, and they are all things that we want to get done reasonably soon).

maxpoulin64:
> And one of the key things about packages is that users should be able to install client packages and have them available in the browser without restarting the server. Browserify wouldn't allow us to do this (without having a watch running, which isn't good for production).

There's nothing preventing the user to run browserify seperately without stopping the server. Actually, even with the current setup, if all the changes are client-side and don't depend on the server's assistance, you can even do a `git pull` + `npm install` over a running lounge instance and it will update just fine. So I still think the solution is valid. The only reason we talked about watchify was for people that don't want to run the build command after doing changes :)

As the client grows, eventually it just won't fit on a single file. It's already getting rather large at the moment, so we will need a build system sooner or later.

---

Alternatively, later we could also possibly split the server and client components into two seperate repos to handle this kind of thing.

This way,
* We could use a client-side dependency manager without it being a mess
* We won't have to worry about the server part
* Probably won't even need npm at all.

The only downside of this solution is client/server version mismatch. But that shouldn't be a big problem if we structure things properly (have the client say what version it supports?) and establish a proper client/server protocol instead of just dumping whatever feature we need here.

lindskogen:
So, what exactly is blocking this from getting merged? The client module handling right now is atrocious with no package manager at all, just a bunch of files in a folder. 
This is a big step on the way towards a better structure, and while the module structure is being reworked over in #77, contributors can work on the project while keeping their sanity intact.

MaxLeiter:
@lindskogen some people are against the idea of having to restart the server in order to install client plugins

YaManicKill:
Alright. Let's not block this with the debate around packages, as this is happening, and packages aren't currently happening.

I'm happy to go with browserify for this, and we can work out if we want to deal with something else for packages later. But yeah, I'm currently getting to the point of just wanting to use browserify for it all. If we have to restart the server, that's not the end of the world, and we should really implement a database to store the scrollback to fix that issue.

@MaxLeiter What state are we in for this? You happy to continue doing this? Could you rebase and squash commits, and I'll take another look at the code afterwards and try and get this in. It'd be really good to get the dependencies out of the repo.

Any objections, @thelounge/maintainers ?

xPaw:
My objection is that it generates a single js file. I'm strongly of opinion that `lounge.js` should be separate file.

MaxLeiter:
It previously made 3 files, that's an easy fix. How many and what files would you like to see, @xPaw? compiled/libs.js, compiled/lounge.js, and compiled/lounge.templates.js?

The reason for one file, iirc, is because templates requires a required Handlebars object, but you can't add a require() into lounge.template.js because it gets rewritten every 'npm run build'

@YaManicKill: it works as-is, will rebase and squash tonight.

xPaw:
> compiled/libs.js, compiled/lounge.js, and compiled/lounge.templates.js?

Ideally yes, because that would keep the existing structure.

MaxLeiter:
Any idea on how to fix the Handlebars in templates? Previously I had a hack of appending to the file using grunt, which was an effective but dirty fix.

YaManicKill:
Fine, fine. If you feel that strongly about it, @xPaw then we can go for the existing structure and debate what we want to do in the future. 

> Any idea on how to fix the Handlebars in templates? Previously I had a hack of appending to the file using grunt, which was an effective but dirty fix.

What's the issue with it, @MaxLeiter ?

lindskogen:
What was the issue with `require`? Isn't this exactly the problem that browserify solves?

MaxLeiter:
Templates requires a `require('handlebars')`, the issue is that whenever you run `npm run build`, `lounge.templates.js` gets overwritten and the require is replaced.

YaManicKill:
> Templates requires a require('handlebars'), the issue is that whenever you run npm run build, lounge.templates.js gets overwritten and the require is replaced. 

I'm not sure I understand, were you telling browserify to overwrite lounge.templates.js with the compiled version? If so, that's your problem, just don't do that, have a compiled file which is different.

If that's not what the problem is, then I'm not sure I understand, why is it overwriting the file?

MaxLeiter:
https://github.com/thelounge/lounge/blob/master/package.json#L15

The templates file gets replaced, so any changes made are overwritten.

xPaw:
So why not compile templates before running grunt then?

YaManicKill:
Ah right, I see your issue, @MaxLeiter that makes sense. 

I think having Handlebars as a global is fine then, just don't run the templates through browserify if the only thing they require is having handlebars.

YaManicKill:
Any luck with this, @MaxLeiter ?

YaManicKill:
Not trying to rush you, @MaxLeiter, just interested in the state of this, you had any time to work on it? Would be good to get this through.

MaxLeiter:
@YaManicKill: There are a huge amount of merge errors which I'm trying to figure out, next time I have time I'll hop on IRC

maxpoulin64:
As discussed on IRC, at the point things are and how lounge evolved it's probably better to just start over fresh for this.

PR #63, opened 2016-02-18
Remove compiled assets and generate them at prepublish time

astorije:
No need to ship compiled assets anymore, everything is done at publish time. Also, when someone clone the repo and runs `npm install`, the assets gets built. The only reason for a contributor to need to build the assets manually is when they update anything in the asset folders (views or libs), which they still have to do anyway (and reviewers need to run that too, obviously).

- Remove minified libs and compiled templates
- Add a `prepublish` script to build assets, that run on `npm install` and right before publishing.
  See https://docs.npmjs.com/misc/scripts
- Include these compiled assets to the `.gitignore` file
- Add an empty .npmignore to make sure the compiled assets are not ignored when publishing.
  See https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
- Update the README to reflect changes in development

Note that the Travis CI configuration does not need any tweaking because it cleans up all extra and changed files, up to publishing on npm. That is, right before `prepublish` gets run.
See https://docs.travis-ci.com/user/deployment/npm#Releasing-build-artifacts

---

I have read a lot on this before opening the PR, but I will still have to monitor this closely when making the first release post-merge to make sure the release is not broken.
Therefore, please **do not merge** after giving +1(s) (I'll let the first reviewer decide if this needs second review or not).
I will ship this in its own release, without anything else, so that if something gets broken, I can `npm unpublish` without affecting the other changes.
Playing it safe here, but that should be much easier to contribute afterwards, I believe :-)


README.md
```diff
@@ -42,10 +42,12 @@ To run the app from source, just clone the code and run this in your terminal:
 
 ```
 npm install
-npm run build
```
YaManicKill: Probably shouldn't remove this, especially as we aren't keeping the build in git anymore. If you download the repository and just run npm install and npm start, you won't get the built files, will you?

package.json
```diff
@@ -14,7 +14,8 @@
     "start": "node index",
     "build": "grunt",
     "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "prepublish": "npm run build"
```
xPaw: That seems like a completely unnecessary addition.

README.md
astorije: Actually, it's one of the 2 effects of `prepublish` that I am leveraging here (and that's how smart npm is :-)):

> prepublish: Run BEFORE the package is published. (Also run on local `npm install` without any arguments.)

That's really cool because compiled files are right built right before publication time when it's about to get sent to npm, and whenever you run `npm install`, that is when you download the repository and start working on it. I tried locally and it does exactly what it says :-)

astorije:
@YaManicKill, do you want to self-assign this?

@xPaw, I notice you have remove your comment. If anyone else is wondering, adding the build script in `prepublish` is necessary in order to get the compiled assets back when publishing. They get built and sent along with the `publish` command (that Travis CI is running for us, but that would be the same by running `npm publish` locally).

README.md
YaManicKill: Yeah, I knew it happened before publication time (hence the name). I didn't realise it ran on install as well. Interesting.

YaManicKill:
Well, my only issue was the comment I left, so ... :+1: from me.

I don't think we should need another, it looks nice and simple. Feel free to merge when you are happy, @astorije

astorije:
Thanks @YaManicKill :-)
I'll change the subject to make sure no one merges, and I'll merge and publish right after the next release!

PR #73, opened 2016-02-19
Hide badge counter with opacity

xPaw:
This makes the badge appear with a transition nicely, instead of showing up harshly (transitions don't apply to `display`).

Couple lines above there is a `transition` property in `#sidebar .badge` which makes it work.


astorije:
:+1: nice and simple :-) Thanks!

PR #75, opened 2016-02-19
Fade long channel names in sidebar

maxpoulin64:
This uses a few CSS tricks to fade away long channel names in the sidebar so the badge don't overlap on top of the channel names.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -287,7 +287,7 @@ button,
 #sidebar .chan-placeholder {
 	display: block;
 	margin: 1px -10px;
-	padding: 6px 10px 8px;
+	padding: 6px 10px 8px 36px;
 	position: relative;
 	text-align: left;
 	transition: color .2s;
@@ -319,6 +319,12 @@ button,
 	text-align: center;
 }
 
+#sidebar .chan:before {
+	position: absolute;
+	top: 4px;
+	left: 10px;
+}
+
 #chat .title:before {
 	margin-top: 17px;
 }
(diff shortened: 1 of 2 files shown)


client/css/style.css
```diff
@@ -272,12 +272,19 @@ button, .user {
...
 	font: 14px FontAwesome;
-	float: left;
 	margin-top: 3px;
 	margin-right: 12px;
 	width: 14px;
 	text-align: center;
+	position: absolute; left: 10px;
+}
+#sidebar .chan:after {
+	position: absolute; right: 10px; top: 0; bottom: 0; z-index: 0;
```
xPaw: One property per line please.

astorije:
Screenshots?

xPaw:
Okay, I tried this now, and there's one thing I don't like about it. Even if you delete close button element, the name still doesn't expand past the notification badge. It simply cuts off where the notification badge would be, without any gradient.

If you can fix this, that would be nice!

maxpoulin64:
Updated. I don't know where exactly you saw that, but I changed the place where the gradient is applied so the trick is simpler which makes it easier to make look clean. I also made sure to move the badge out of the way by putting it back to position:absolute when it fades away. It should look pretty good now no matter what.

lounge-badge

xPaw:
:+1:

YaManicKill:
This is pretty awesome, thanks @maxpoulin64 

Can you rebase, and check that the new stylelint doesn't break the build for this, and then I'm happy to merge. 

:+1:

maxpoulin64:
@YaManicKill Here you go. It did break all over the place, and it didn't merge properly at all either. Ended up just redoing it over master.

astorije:
I love it, excellent job @maxpoulin64!

I did notice a small issue though: when the name is long, hovering on the name is messing up with the channel closing button (&times;). I know we mentioned removing it, but it's not the case yet (and it is not sure yet). Here is what I'm talking about (notice the first and third lines):

thelounge-fade-badge

Do you think you could fix this before we merge? :-/

astorije:
After discussing with @maxpoulin64 on IRC, he fixed this and it does look terrific. So :+1: and merging.

Also, I'm re-labeling this as a bugfix since currently long channel names are simply not being displayed well: they appear on a second line and get cut at the end of the end of the sidebar. Leaving as enhancement too considering @maxpoulin64 went well beyond duty :-)

PR #76, opened 2016-02-20
Add a simplified way to invite a user in the current channel

astorije:
This relates to https://github.com/thelounge/lounge/pull/7#issuecomment-184598927.

### DIFF
--- a/src/plugins/inputs/invite.js
+++ b/src/plugins/inputs/invite.js
@@ -2,8 +2,12 @@ module.exports = function(network, chan, cmd, args) {
 	if (cmd !== "invite") {
 		return;
 	}
+
 	var irc = network.irc;
+
 	if (args.length === 2) {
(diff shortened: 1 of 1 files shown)


astorije:
Now that I think of it, I need to make sure that this is only possible in channels, not lobbies or private messages.
I'll give it another test and run tomorrow!

astorije:
Indeed, I needed an extra check. @xPaw or any other reviewer, this is good to go, and I'll let the first reviewer decide if this needs an additional review.

xPaw:
:+1: Next step is showing a message like "You invited xxx to #yyy" :)

astorije:
Yep, but not in this PR, correct?
Also, do you want to merge this now or prefer a second review?

xPaw:
Yes another pr is fine. Change is small enough not to warrant a second review.

PR #78, opened 2016-02-20
Update README.md

LouisT:
The term "official" (appointed or authorized to act in a designated capacity: an official representative) only applies if the original creator agrees to such. Because of this, only @erming has the right to give this project an "official" label.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ __What is it?__
 
 The Lounge is a web IRC client that you host on your own server.
 
-*This is the official, community-managed fork of @erming's great initiative, the [Shout](https://github.com/erming/shout) project.*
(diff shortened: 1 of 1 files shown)


dgw:
I was under the impression that @erming had agreed that a fork was the way to go, because he hasn't the time to keep up with development at the pace the user community desires.

astorije:
Hi @LouisT,
As @dgw mentioned, @erming indeed has approved and supports this fork. As a matter of fact, the name itself was offered by @erming :-)
Additionally, as you can see, this fork is within a GitHub organization which has currently 5 members, reviewing and merging PRs opened by the community. It is therefore community-managed, and (successfully so far) aims at consolidating all forks out there.
:-1:, closing this PR.

PR #79, opened 2016-02-20
Replace bootstrap tooltips with css tooltips from Primer

### LINKED ISSUE #67: Replace bootstrap.js with alternatives
At the moment, /client/js/libs/jquery/bootstrap.js is only used for transitions and tooltips. Once #62 is complete, we should probably replace it with [hint.css](http://kushagragour.in/lab/hint/) and CSS transitions or something

xPaw:
Fixes #67. These tooltips are purely css, no javascript required.

Before:
chrome_2016-02-20_16-32-56

After:
after


YaManicKill:
Really like this, they look much nicer, and they are very responsive.

:+1: from me :-)

astorije:
Hey @xPaw, very nice change!

I noticed 3 little things that could be improved before I give my +1: 

- The previous tooltip had smaller top and bottom paddings, it felt less bulky. Could you change that?
- Similarly, the previous tooltip was either not as white, or a bit transparent. Would be great to have that here too.
- Finally, there was a nice transition on appear/disappear that has been removed. It was a nice integration with the rest of the UI.

I know you already posted a before/after but here is mine without zoom, to better feel the difference:

Before | After
--- | ---
screen shot 2016-02-21 at 14 23 51 | screen shot 2016-02-21 at 14 24 17

Thanks! :-)

xPaw:
> The previous tooltip had smaller top and bottom paddings, it felt less bulky. Could you change that
That's line-height I think, will change.

> Finally, there was a nice transition on appear/disappear that has been removed. It was a nice integration with the rest of the UI.
`transition` should work.

> Similarly, the previous tooltip was either not as white, or a bit transparent. Would be great to have that here too.
I actually like the full white better.

astorije:
> That's line-height I think, will change.
> transition should work.

Great!

> I actually like the full white better.

Fair enough, I'll see how it looks with the previous 2 changes :-)

xPaw:
Done.

astorije:
Sweet! :+1: and merging.

Issue #46, opened 2016-02-15
Port is stored as a string in user config, it should be forced to be an integer


xPaw:
```
      "port": "1025",
```
that should be
```
      "port": 1025,
```

astorije:
:+1:

astorije:
Fixed in #82.

PR #82, opened 2016-02-21
Allow locking network configuration

### LINKED ISSUE #23: Add an option to lock default server
Useful for Lounge's demo instance, and admins that plan on running a public server.

This needs to disable ip/port fields (and possibly SSL?), and also completely ignore sent values on the server, and always use config defaults (to prevent editing shenanigans).

xPaw:
Fixes #23, #46, and removes hardcoded default to freenode server in code (https://github.com/thelounge/lounge/pull/13#issuecomment-183388525).

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -103,11 +103,11 @@
 									<label>Server</label>
 								</div>
 								<div class="col-sm-6 col-xs-8">
-									<input class="input" name="host" value="<%= defaults.host %>">
+									<input class="input" name="host" value="<%= defaults.host %>" <%= typeof(lockNetwork) !== "undefined" && lockNetwork ? 'disabled' : ''%>>
 								</div>
 								<div class="col-sm-3 col-xs-4">
 									<div class="port">
-										<input class="input" name="port" value="<%= defaults.port %>">
+										<input class="input" name="port" value="<%= defaults.port %>" <%= typeof(lockNetwork) !== "undefined" && lockNetwork ? 'disabled' : ''%>>
 									</div>
 								</div>
 								<div class="clearfix"></div>
@@ -120,7 +120,7 @@
 								<div class="col-sm-3"></div>
 								<div class="col-sm-9">
 									<label class="tls">
-										<input type="checkbox" name="tls" <%= defaults.tls ? 'checked="checked"' : '' %>>
+										<input type="checkbox" name="tls" <%= defaults.tls ? 'checked="checked"' : '' %> <%= typeof(lockNetwork) !== "undefined" && lockNetwork ? 'disabled' : ''%>>
 										Enable TLS/SSL
 									</label>
 								</div>
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -79,13 +79,23 @@ module.exports = {
 	//
 	// Display network
 	//
-	// If set to false The Lounge will not expose network settings in login
-	// form, limiting client to connect to the configured network.
+	// If set to false network settings will not be shown in the login form.
 	//
 	// @type     boolean
 	// @default  true
 	//
 	displayNetwork: true,
+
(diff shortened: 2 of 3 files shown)


defaults/config.js
```diff
@@ -88,6 +87,17 @@ module.exports = {
...
 	//
+	// Lock network
+	//
+	// If set to true, users will not be able to modify host, port and tls
+	// settings and will be limited to the configured network.
+	//
+	// @type     boolean
+	// @default  false
+	//
+	lockNetwork: false,
```
YaManicKill: Instead of locking to the default, could we have an array of hostnames that we can lock to. Would allow people that want to allow multiple irc servers, but not just any random one.

xPaw: That would require some restructuring, plus have a dropdown on the client show allowed servers. I'm not keen on doing that myself.

astorije:
@xPaw, now that I pay attention, what is the difference between this and `displayNetwork` in practice?

I mean, not displaying the network does connect you to the network already set in the config file, correct?

xPaw:
`displayNetwork` does what it says. It just hides the network using `display:none`, it doesn't actually prevent users from joining another network. And it's better to keep it separate, as you might want to lock the network, but still display it (like our demo instance does).

astorije:
Then I'm wondering what good brings `displayNetwork`  on its own... Should we remove it altogether? Or at least enable `lockNetwork` by default when enabling `displayNetwork`?

I would simply remove it after looking where it comes from and why. Probably a different PR though but discussion is interesting if we want to lock when not displaying it.

xPaw:
Yeah I'm not sure about it, but it should be handled in a separate PR.

maxpoulin64:
:+1: 

I agree with keeping `displayNetwork` seperate. One use I find is when The Lounge is used as a web gateway to a network, it could be useful to lock it to that network, but still keep the settings visible to advertise where it connects to. This could be handled otherwise (like putting it on the network's website), but I like the transparency of what it's doing.

astorije:
@xPaw, @maxpoulin64, where are we on this one? Do you think it could go along with the next minor release later this week? :-)

maxpoulin64:
@astorije Yep, looks good to me now that my questions have been answered. :+1:

client/index.html
```diff
@@ -120,7 +120,7 @@ <h1 class="title">Connect</h1>
 								<div class="col-sm-3"></div>
 								<div class="col-sm-9">
 									<label class="tls">
-										<input type="checkbox" name="tls" <%= defaults.tls ? 'checked="checked"' : '' %>>
+										<input type="checkbox" name="tls" <%= defaults.tls ? 'checked="checked"' : '' %> <%= typeof(lockNetwork) !== "undefined" && lockNetwork ? 'disabled' : ''%>>
```
astorije: [Very minor comments] For these 3 changes, could you choose either single or double quotes, rather than a mix of both? Also, you might want to do `'' %> >` or `'' %>>` instead of `''%>>`.

maxpoulin64: This seems reasonable to me, to avoid the annoying escaping of the double-quotes for the HTML. This is consistent with what's already there. Double-quotes for Javascript expressions, single quotes for the HTML output.

astorije: I thought about that at first, but actually there is no need to escape here, unless I am being mistaken at that time of the day: `<%= typeof(lockNetwork) !== "undefined" && lockNetwork ? "disabled" : "" %>` seems to be fine to me.

astorije: (And actually, for checkboxes, no need to do `'checked="checked"'`, `"checked"` would be enough here)

astorije:
Alright, :+1: and merging. Comments to minor to justify a hold off.

PR #87, opened 2016-02-22
Fix show-more being displayed over chat messages

nomadturk:
As per https://github.com/erming/shout/pull/568:

With this PR I tried fixing #366 by changing it's position to relative. It wasn't a pretty sight to see "Show more" bar on existing chat messages.

With this fix Show More button no more will overlap with the underneath text.
As in: 
chrome_2016-02-22_16-43-50

Result will be like: 
Test image

This fix also includes @xPaw 's contribution as well.
PS: I mistakenly added another commit to this PR and removed it later. See #88 for the next one.


astorije:
Hey @nomadturk,

Thanks for this!
There is an issue with your commits, do you mind squashing them into one commit please?

Also, naive question: what happens if we do not specify any `display` at all?

xPaw:
> Also, naive question: what happens if we do not specify any display at all?

You mean `position`? It should work just fine with it removed.

PR #88, opened 2016-02-22
Fix autocomplete problem

nomadturk:
Updated the form to not auto complete what we wrote before by removing method="post" action="" and adding autocomplete="off"

This patch denies https://github.com/erming/shout/pull/518 and removes that. Because without autocomplete, writing suggestions for mobile keyboards was enabled but at the mean time also whenever you clicked on text area to write a new message, you were given a number of your last typed sentences. Especially on mobile, this was a huge buzzkill because after 3-5 sentences on any channel, these suggestions covered your whole screen.

So, in order to remove it I did some changes, ensuring writing suggestions still work.
I have added autocomplete=off .

This commit is related to https://github.com/erming/shout/pull/532

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -278,7 +278,7 @@
 					</div>
 				</div>
 			</div>
-			<form id="form" method="post" action="">
+			<form id="form" autocomplete="off">
 				<div class="inner">
 					<button id="submit" type="submit">
 						Send


YaManicKill:
I'll double test this tonight before giving my +1, but I would really like this in if it does what I think it does :-P

nomadturk:
This one uses HTML's autocomplete setting to dsable autocomplete on the browser side. Since the textbox we're using is a for chat messages, it's quite disturbing if it is enabled.

Without this, especially on mobile you get your whole previous sentences as autocomplete suggestions. This removes that.

Also with this fix keyboard suggestions remains intact.

astorije:
Sadly, this is a no go because it also affects typing on bare Android devices. I agree with the problem though, and therefore believe that some research should be done to fix this the right way.
On top of reading MDN/the W3C specs on the subject, it would be good to see how popular websites do it to solve this issue. One can compare solutions adopted by https://m.facebook.com, https://web.telegram.org, https://gitter.im, ...

YaManicKill:
So, from my usage of facebook, they haven't solved it. They turned off autocomplete and it just doesn't work well.

Damnit, Google, if you want the web to be a competitive platform, you need to sort this out.

Yeah, This will have to be a no, I didn't realise this was what caused certain keyboard to completely  bug out.

PR #91, opened 2016-02-23
Fix how highlights are handled

### LINKED ISSUE #90: Actions containing Highlights display incorrectly
Actions have the sender's nickname to the right of the bar which separates nicknames and messages.

When an Action contains a Highlight, the action displays correctly except the nickname appears to the left of the aforementioned bar.

[Here's an example](http://i.imgur.com/8bs3M2Z.png)

([Here's an example from the other side, courtesy of IRC user vectr0n](https://vectr0n.net/img/uploads/big/ac443ec62b006ffc37a512d3dbd75dee.png))

This is likely a minor fix to move the name back across the bar. I would also encourage changing the color to the red highlight color. The placement of the nick and the star icon both clearly indicate that it's an Action, but nothing indicates that it's a Highlight except the sound and badge. I believe we would be better served by having Actions that contain Highlights appear red.

xPaw:
- Partially fixes #90 where highlighted actions were broken
- Sent actions are correctly identified as `self`
- Sent actions are correctly highlighted (previously this only worked when reloading from history)
- Minor performance improvement where received messages are no longer split twice

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -640,10 +640,7 @@ $(function() {
 	chat.on("msg", ".messages", function(e, target, msg) {
 		var button = sidebar.find(".chan[data-target='" + target + "']");
 		var isQuery = button.hasClass("query");
-		var type = msg.type;
-		var highlight = type.contains("highlight");
-		var message = type.contains("message");
-		if (highlight || isQuery || (options.notifyAllMessages && message)) {
+		if (msg.highlight || isQuery || (options.notifyAllMessages && msg.type === "message")) {
 			if (!document.hasFocus() || !$(target).hasClass("active")) {
 				if (options.notification) {
 					pop.play();
@@ -679,7 +676,7 @@ $(function() {
 			"nick",
 			"mode",
 		];
-		if ($.inArray(type, ignore) !== -1){
+		if ($.inArray(msg.type, ignore) !== -1){
 			return;
 		}
 
@@ -688,7 +685,7 @@ $(function() {
 			var i = (badge.data("count") || 0) + 1;
 			badge.data("count", i);
 			badge.html(i > 999 ? (i / 1000).toFixed(1) + "k" : i);
-			if (highlight || isQuery) {
+			if (msg.highlight || isQuery) {
 				badge.addClass("highlight");
 			}
 		}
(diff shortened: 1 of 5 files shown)


maxpoulin64:
This one seems to break notifications on highlight. I put my test instance on a random channel and highlighted myself, but I didn't get the desktop notification and the badge didn't turn white either. It just displayed as if it was any other non-highlight channel message.

I also noticed the line doesn't get colored at all, so it's not clear it's a highlight. It just stays orange.

Did I miss something? I did run `npm run build` before and cleared the browser cache.

xPaw:
Was there a .highlight class applied? Was it an  action?

maxpoulin64:
Yes, the .highlight class gets applied properly. But with both actions and regular messages, no desktop notification gets received, and the badge doesn't turn white. With regular messages, the line does turn red as expected however.

```html
<div class="msg message highlight">
	<span class="time">
		03:49
	</span>
	<span class="from">
		<a href="#" class="user" style="color: #ff514d">Max-P_test</a>
	</span>
	<span class="text">
			Max-P: That still works
	</span>
</div>
```

```html
<div class="msg action highlight">
	<span class="time">
		03:48
	</span>
	<span class="from"></span>
	<span class="text">
		<a href="#" class="user">Max-P_test</a>
		
		 highlights Max-P
	</span>
</div>
```

Switching back to master, it all works again.

xPaw:
This PR doesn't fix color on highlight (the other PR does), so that's fine. Unsure about notification though.

maxpoulin64:
Found it: [highlight checked here](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L655) and [broken here](https://github.com/xPaw/lounge/commit/24eb89cbc2e01654faaae7090fbb8e176ff63942#diff-1afea02ef01ba2e45ce6e3e55a2dbf82L39)

Since you're adding an `highlight` property it should be an easy fix actually.

xPaw:
Template adds the highlight class though, so that  doesn't make sense.

maxpoulin64:
It's not checking the class, it's checking the message type directly from the server event.

xPaw:
Ah, my bad.

xPaw:
@maxpoulin64 Should be fixed now.

maxpoulin64:
I confirm it works fine for me now, :+1:

astorije:
:+1:

Issue #55, opened 2016-02-16
Show hostmask on join/part/quit


v0l:
Add host info for people who join a chan, 

``` v0l (v0@ip) has joined ```

~~Also add quit messages~~

xPaw:
Quit messages are already shown, do you not see them?

v0l:
Oh yea i see it, needed update sorry.

astorije:
I really don't think we shout display host info when someone joins. It's noisy, doesn't add much information, and can be retrieved using a `/query`. I find that negatively affecting the UX so I'm :-1: on this, sorry!

dgw:
Gonna go out on a limb here and say that themes should be able to add or remove info in event messages, regardless of whether core does this. (If that's not currently possible/planned, please do correct me!)

@astorije You mean a /whois, no?

astorije:
Sorry yes, a `/whois`. And I disagree with this, themes should be about UI, not data/info. Anything else too specific and not UI-related should go in a package.

xPaw:
> It's noisy, doesn't add much information, and can be retrieved using a /query.

It can't be retrieved using commands when you're trying to ban someone that just quit.

easymac:
I'm on the side of adding visible hostmasks that can be hidden/shown with CSS as it's an easy way to identify when someone joins under an alias or when a user is being impersonated.

vectr0n:
+1 Would love to see at least the option to show the ident@host in join/part.

Jay2k1:
A huge +1. This is a must-have for me.

richrd:
Big :+1: for the possibility for themes to show the hostmask.
Could be off by default but would be nice if it would be an option.

astorije:
Added in #94, closing.

PR #94, opened 2016-02-23
Move actions to templates

### LINKED ISSUE #86: Separate templates for each of the actions
This code would be moved to templates: https://github.com/thelounge/lounge/blob/1e8ca51d47f78e30e8d429cfc164258386e0f5d7/client/js/lounge.js#L204-L215

This will allow for easier styling, and saner way of adding html elements where needed, for example when user changes his name currently its like `<a>xPaw</a> changed their name to test`, where `test` is not a clickable link as it should be. Plus, in the future where we will add hostmasks, we'd want them to be hideable by CSS if so desired.

xPaw:
This makes it easier to add action specific rendering (e.g. hostmasks in join messages).

This fixes #86 and #55.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -201,18 +201,7 @@ $(function() {
 			"topic",
 			"action",
 		].indexOf(type) !== -1) {
-			switch (type) {
-			case "invite": data.msg.formattedAction = "invited " + data.msg.target + " to"; break;
-			case "join": data.msg.formattedAction = "has joined the channel"; break;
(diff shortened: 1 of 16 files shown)


maxpoulin64:
Looks all good to me apart from the minor issues I pointed out, so I will :+1: this one. Great change, glad to finally be able to see the hostmasks and ident too.

astorije:
Looks great! :+1:

Regarding @maxpoulin64's comment, I think he has a point, but definitely not to worry for this PR, we can improve over time as there are different strategies we could apply here.

PR #100, opened 2016-02-24
Replace favico.js with our simpler solution

xPaw:
This gives us greater control over the notification favicon (as it's separate image, we can do whatever we want with it), and removes a 800 line library.

This also makes favicon badges work in Chrome.


client/index.html
```diff
@@ -16,8 +16,7 @@
 	<link rel="stylesheet" href="css/style.css">
 	<link id="theme" rel="stylesheet" href="<%= theme %>">
 
-	<link rel="shortcut icon" href="img/favicon.png">
-	<link rel="icon" sizes="192x192" href="img/touch-icon-192x192.png">
```
xPaw: This separate icon had to be removed because Chrome has some bizzare logic for selecting icons, and if you update the wrong icon, the notification badge won't be shown at all.

astorije:
@xPaw, the icon will be changed only when highlighted somewhere or someone talks to you in PM, correct? If so, could you make it red to match the highlight color? :)

xPaw:
@astorije I'm not a designer, so it was hard enough to get this color :grin: If someone can help, that would be nice.

maxpoulin64:
Looks good to me, tested on Chrome, Firefox and Chrome on Android with no side effect (apart that this one actually works). :+1: 

Otherwise, agreed with @astorije, it could be nice to have it red, maybe even bright red for visibility.

Thinking something like that: (I'm also terrible at colors)

favicon-no-bg
favicon-red

xPaw:
Yeah, brighter red would be much nicer.

maxpoulin64:
Like that?

favicon-red-2

xPaw:
That looks better, yes.

astorije:
@xPaw, can you integrate @maxpoulin64's version in the PR?

xPaw:
I will when I when I rebase next time. For now, tell me if there's anything needed to be changed.

xPaw:
Rebased with red icon.

astorije:
Works great on PMs and highlights, love it! Also, good job on getting rid of that library!!

A couple thoughts though:
- No effect when being `/invite`d, though it would be great to get notified!
- As a user, I would expect the red favicon to disappear as soon as I enter the tab, not when I actually read the messages. So in a nutshell, red favicon means something happens off-focus, and goes back to blue when as soon as the tab/window is being focused. One nice side-effect of this is that the favicon never goes red if the window is in focus and the user gets notified. Does that make sense? Would that be easy to achieve?

xPaw:
> No effect when being /invited, though it would be great to get notified!

As simple as setting `highlight` to true in invite messages I believe.

> As a user, I would expect the red favicon to disappear as soon as I enter the tab, not when I actually read the messages.

This changes the behaviour and I'm not sure it's for this PR.

astorije:
> As simple as setting highlight to true in invite messages I believe.
> EDIT: As it turns out, it produced completely broken notification in the browser.

Yep, noticed that too. Will try to fix that now.

> This changes the behaviour and I'm not sure it's for this PR.

Agreed, although it would be really nice if we could improve that behavior soon :-)

:+1: and merging!

astorije:
>> As simple as setting highlight to true in invite messages I believe.
>> EDIT: As it turns out, it produced completely broken notification in the browser.

> Yep, noticed that too. Will try to fix that now.

Done in https://github.com/thelounge/lounge/pull/127.

PR #102, opened 2016-02-25
Fix CSS selector syntax in channel message handler

maxpoulin64:
Doesn't do much, but it prevents unneeded exceptions which are annoying when debugging. This makes the selector valid. An exception is thrown on every single message sent to any channel, so when you have a lot of them the debugger endlessly pauses for nothing.

It might also possibly improve performance slightly as jQuery won't have to fallback to its own implementation (but don't quote me on that).


astorije:
:+1:

PR #103, opened 2016-02-25
Wrap badge and notifications in try/catch

maxpoulin64:
This fixes a somewhat rare/unusual bug where if bots dumps too much output to the user (such as `/msg ChanServ help` commands or `/msg *status help`), it crashes the client and reloads. Also fixes an Android deprecation issue with notifications that also crashed the client.

As a general rule I think crashing on notifications is a bit silly so I just wrapped the whole thing in a try/catch.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -655,25 +655,32 @@ $(function() {
 		var highlight = type.contains("highlight");
 		var message = type.contains("message");
 		if (highlight || isQuery || (options.notifyAllMessages && message)) {
-			if (!document.hasFocus() || !$(target).hasClass("active")) {
-				if (options.notification) {
-					pop.play();
+			try {
+				if (!document.hasFocus() || !$(target).hasClass("active")) {
+					if (options.notification) {
+						pop.play();
+					}
+					favico.badge("!");
+					if (options.badge && Notification.permission === "granted") {
+						var notify = new Notification(msg.from + " says:", {
+							body: msg.text.trim(),
+							icon: "img/logo-64.png",
+							tag: target
+						});
+						notify.onclick = function() {
+							window.focus();
+							button.click();
+							this.close();
+						};
+						window.setTimeout(function() {
+							notify.close();
+						}, 5 * 1000);
+					}
 				}
-				favico.badge("!");
-				if (options.badge && Notification.permission === "granted") {
-					var notify = new Notification(msg.from + " says:", {
-						body: msg.text.trim(),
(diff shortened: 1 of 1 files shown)


maxpoulin64:
I'll add that we probably should focus on actually making all those methods safe instead of just catching all errors, so I would understand if this one was rejected in favor of a cleaner fix. It seems the whole thing is very browser dependant however, so it's not a super easy fix. I have seen exceptions being thrown by both:

*  `favico.badge("!");` (does it even work?) throws an exception when it is called too often (such as a very large PM from ZNC's *status). We should indeed merge notifications when a bunch of messages are received, but the implementation would need a somewhat large discussion and a lot of refactoring for the server to group messages.
* `new Notification(` - In Chrome on Android, even if the permission is granted. The exception thrown says to use another function instead, so it seems we'd need a helper that selects the appropriate method.

xPaw:
For favico, see #100. If favico fails horribly when updating the favicon multiple times, then my PR should take care of that.

maxpoulin64:
@xPaw #100 indeed fixes the crash I had.

Do you have any idea for the other one? This is the problem: [StackOverflow](http://stackoverflow.com/questions/31512504/html5-notification-not-working-in-mobile-chrome). Calling `new Notification()` throws an exception that says to use service workers, but that method looks somewhat complicated.

xPaw:
They probably force notifications to go through service worker so you can get them while not having the page open.

astorije:
@xPaw, could you self-assign maybe? :-)

xPaw:
Also, I think the try/catch solution is so awful it shouldn't be merged. Perhaps we should look into using a better notification polyfill (we have one now!) to handle these errors?

maxpoulin64:
I agree the solution is ugly. The issue I had with it is that Chrome does everything as if it would work, `Notification` exists, it will even ask for notification permissions and let you grant them. And then when you build one, bam, it blows up in your face. 

Since it never worked I would be fine with just disabling it as a temporary workaround until someone have time to move everything to a service worker. I just don't know how to check for the feature availability and skip it.

astorije:
@maxpoulin64 and @xPaw, where are we with this PR? Should it be improved to lean towards a merge or is it a no-go and should be closed?

xPaw:
I'm inclined towards closing it.

maxpoulin64:
Lets close it. Now to prevent the crash we just need to disable the notifications in the settings as the favicon one is fixed. We'll fix that proper when we get to make a serviceWorker.

PR #119, opened 2016-02-28
Bump grunt-contrib-uglify and pin versions of grunt packages

astorije:
Made different tests locally and it seems to work fine (only change is `grunt-contrib-uglify`) but since this can break publication, I wouldn't be against some more testing from other reviewers :-) (will not set the `second review needed` label myself for this reason).

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,10 +43,10 @@
   "devDependencies": {
     "stylelint": "4.3.3",
     "eslint": "^1.5.1",
-    "grunt": "~0.4.5",
-    "grunt-cli": "^0.1.13",
-    "grunt-contrib-uglify": "~0.5.0",
-    "grunt-contrib-watch": "^0.6.1",
+    "grunt": "0.4.5",
+    "grunt-cli": "0.1.13",
+    "grunt-contrib-uglify": "0.11.1",
+    "grunt-contrib-watch": "0.6.1",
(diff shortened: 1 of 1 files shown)


xPaw:
:+1: Since we had `~` and `^` versions already, we unintentionally tested newer version :)

astorije:
Actually, not quite: all of them but `grunt-contrib-uglify` still have their stable version matching the versions we have in the `package.json`.
Only `grunt-contrib-uglify` changes, as prior to that PR, we were matching `0.5.0` and `0.5.1`, while this PR makes it go to `0.11.1`. Other are just pinned with no other consequences.

maxpoulin64:
Looks fine to me, :+1: I'm not a fan of pinning versions, but as discussed with @astorije on IRC on the long term it's the safer approach. Lets just not forget to keep them updated once in a while not to fall too far behind :)

YaManicKill:
Might be worth using [libraries.io](https://libraries.io/) to make sure we don't fall behind. Emails maintainers when something is updated.

astorije:
We actually do have [a libraries.io page](https://libraries.io/npm/thelounge) as I contacted them when starting to fix that page which was not working properly...
However, we also use badges from https://david-dm.org/thelounge/lounge, which also works pretty well. Just need to keep an eye on dependencies :-)

PR #122, opened 2016-02-28
Fix #116: Move 'you' handling in invites to the client

### LINKED ISSUE #116: Clickable "you" when being invited
When another user invites the current user to a channel, the following statement is being shown to the current user:

> **astorije** invited **you** to #astorije

However, **you** is clickable, while it should not.

Looks like this in practice:

screen shot 2016-02-27 at 19 22 03

xPaw:


astorije:
@xPaw, does the client have a reference to the user's nick? If so, we could have the equality check in the template directly rather than passing that `invitedYou` in the event. If not, nevermind then.

astorije:
Marking with `priority` label here: it's a regression compared to the latest release, `v1.2.1`, so let's not ship that in the next release.

maxpoulin64:
:+1:

astorije:
Will +1 and merge when @xPaw addresses (or not) my comment above :-)

xPaw:
It does not have a reference.

PR #125, opened 2016-02-29
Fix complete crash when refreshing a public instance

### LINKED ISSUE #124: Server crashes when reloading the page on a public instance
Not sure exactly what triggers this nor how to reproduce 100% of the time, but I got this error when doing the following:

- Run server, public instanec
- Access it on localhost
- Connect to Freenode and another IRC network
- Freenode is broken, got no channels open nor connection with it
- Reload the page
- Boom, error

The stack trace:

```
> node index "--port" "9001" "--home" "/tmp"


The Lounge is now running on http://0.0.0.0:9001/
Press ctrl-c to stop

Client#connect():
Error: connect ETIMEDOUT 185.30.166.38:6697
.../lounge/src/client.js:337
	var room = sockets.adapter.rooms[this.id] || [];
	                          ^

TypeError: Cannot read property 'rooms' of undefined
    at Client.quit (.../lounge/src/client.js:337:28)
    at Socket.<anonymous> (.../lounge/src/server.js:189:11)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at Socket.emit (.../lounge/node_modules/socket.io/lib/socket.js:129:10)
    at Socket.onclose (.../lounge/node_modules/socket.io/lib/socket.js:398:8)
    at Client.onclose (.../lounge/node_modules/socket.io/lib/client.js:204:12)
    at emitTwo (events.js:92:20)
    at Socket.emit (events.js:172:7)
    at Socket.onClose (.../lounge/node_modules/socket.io/node_modules/engine.io/lib/socket.js:240:10)
```

Let me know what additional info I can bring and I'll try to find that.

astorije:
This fixes #124. The `Client` constructor was changed in #57, but was used twice and updated only in one place.
Removed global reference to `sockets` as it's not needed anymore. I tested and did not find any new stuff breaking, but reviews will confirm.

Marking as `priority` as releases are on hold until #124 is fixed in `master`.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -8,7 +8,6 @@ var io = require("socket.io");
 var Helper = require("./helper");
 var config = {};
 
-var sockets = null;
 var manager = new ClientManager();
 
 module.exports = function(options) {
@@ -43,7 +42,7 @@ module.exports = function(options) {
 		require("./identd").start(config.identd.port);
 	}
 
-	sockets = io(server, {
+	var sockets = io(server, {
 		transports: transports
 	});
 
@@ -182,7 +181,7 @@ function init(socket, client, token) {
 function auth(data) {
 	var socket = this;
 	if (config.public) {
-		var client = new Client(sockets);
+		var client = new Client(manager);
 		manager.clients.push(client);
 		socket.on("disconnect", function() {
 			manager.clients = _.without(manager.clients, client);


maxpoulin64:
Ouch. :+1:

xPaw:
Are you sure this doesn't need changing? https://github.com/thelounge/lounge/blob/astorije/124-fix-crash/src/client.js#L336-L337

astorije:
@xPaw, yes, because `Client` still has a reference to the `sockets` as part of the constructor: https://github.com/thelounge/lounge/blob/5f11c262236078a8d671b5fbdb515de02a0671e5/src/client.js#L60
Now we can wonder if it's useful to have both around, but this is a quick bug fix to fix the `master` branch,  I don't think I should refactor other things in this PR.

xPaw:
I was asking in case that needed to be changed. :+1:

astorije:
Yeah, that was the starting point of my debugging, but quickly found the actual culprit later :-)
That being said, do you want me to get rid of the `sockets` in the `Client` model in a further PR? I'd say we'll clean this up over time anyway, but you pick.

xPaw:
Either is fine. Merging this.

PR #126, opened 2016-02-29
Add debug command in package.json

astorije:
Alright, see this PR as a discussion ticket around making debugging easier in the short term.

What this change adds at the moment is a simple way to run `npm run debug [-- params]` to run the server in debug mode. It should easily come with some basic documentation, at the very least some lines in https://github.com/thelounge/lounge#development-setup, but again, comments welcome.

I believe there is also a way to run `slate-irc` in debug mode, should I enable this as part of the `debug` command here?

Anything else we could do to make debugging as easy as possible?

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "url": "https://github.com/thelounge/lounge.git"
   },
   "scripts": {
+    "debug": "node debug index",
     "start": "node index",
     "build": "grunt",
     "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",


xPaw:
Does this accomplish anything at the moment? As far as I know there are no debug statements in the code anywhere.

astorije:
Hey @xPaw!
Indeed, there are no debug statements (`debugger;`) and that's good, we should not ship code with debug statements. What this PR is trying to accomplish is having a framework for debugging The Lounge.
At the very least it should be this line (so that devs can run `npm run debug`) and probably a line in the README to describe it, but it could also be some specific `devDependencies` or setting flags to set The Lounge or any of its dependencies in a verbose or debug mode.
Does this make sense?

YaManicKill:
Meh, I'd be fine with this. :+1:

I'll add second reviewer label just because @xPaw wasn't sure on it.

astorije:
@YaManicKill, OK, I'll add slate debugging stuff here. Considering work on #167, I'll look there if there is an equivalent.
Regarding debugging statements in the code, it is actually a very bad thing to have in the codebase itself, but it's a good thing to have the command to run in debug-aware mode around.

xPaw:
Sorry, I don't see the need in having this command just yet.

astorije:
Well, in Node.js you cannot use the debugger unless you are running it in debug mode.
But oh well, I guess I'll type the full command then :)

PR #127, opened 2016-02-29
Add notifications for channel invites

astorije:
Here is how it looks:

screen shot 2016-02-29 at 02 29 31

Regarding title, I'm fine with this above (suggested by @maxpoulin64), or `You were invited to a channel:`, ... (maybe as long as it ends with `:` to keep consistency with other notifications?)
Make an offer if you think you have better, I don't have a strong opinion on this :-)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -682,21 +682,31 @@ $(function() {
 	chat.on("msg", ".messages", function(e, target, msg) {
 		var button = sidebar.find(".chan[data-target='" + target + "']");
 		var isQuery = button.hasClass("query");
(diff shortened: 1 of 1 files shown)


maxpoulin64:
I think I would prefer `You were invited to a channel:` because it sounds a lot cleaner. Or maybe just `Channel Invitation:` so it's short.

Anyway, works great, that's a :+1: for me for the code. I'll go with whatever the vote ends up being :)

astorije:
Not too fond of "Channel invitation:" but yeah, I'll go with the flow :-)

xPaw:
:+1: Fine with the current option.

Issue #24, opened 2016-02-13
Keep a look out on irc-framework


xPaw:
https://github.com/kiwiirc/irc-framework

@prawnsalad is working on a new framework for use in his Kiwi IRC client, and has suggested that combining efforts would be a good idea towards better clients altogether.

`slate-irc` is practically dead, and would require a lot of work to add many of the features that are already supported in `irc-framework` (mostly IRCv3 stuff, including SASL, server-time and others.)

YaManicKill:
Yeah, I think we need to look into the difference in effort of:

1. Changing the irc framework we use
2. Adding new features to slate (because, as oyu say, it's dead)

@astorije Can you add a discussion label, I don't see one currently.

@xPaw Are you willing ot look into how much work it would be to move frameworks?

xPaw:
It should be a matter of renaming events and parameter names, as they differ. Other than that, they're both event based, so it shouldn't be terribly difficult to move.

YaManicKill:
Alrighty then, marked this as an enhancement, and if someone wants to pick it up, feel free. 

As usual, feel free to open a PR early, so we can see how it's going to work rather than getting absolutely everything done and having to change some stuff.

xPaw:
Hold on the PR until Kiwi itself switches to using that framework.

xPaw:
There are another reason why I see switching to `irc-framework` will help, it will have proper channel/user state tracking, so that our client won't need to do hacky stuff like requesting `names` from the server whenever any users' mode changes.

https://github.com/thelounge/lounge/blob/30b8a419cdb711808b68df08fe9612f0c646dbe8/src/plugins/irc-events/mode.js#L9-L11

xPaw:
Another thing: whenever `mode` event is received, it requests `NAMES` for that channel:

https://github.com/thelounge/lounge/blob/master/src/plugins/irc-events/mode.js#L9-L11

xPaw:
Just found out another thing: `errors` event does not include `from` field.

thelounge: https://img.avora.xyz/uploads/big/6ea7aa202ce04329b2b3d431f24cb24d.png
hexchat: https://img.avora.xyz/uploads/big/b1d5934840ce801141629faa731fe004.png

lindskogen:
The support for middleware gets me excited!

xPaw:
I started working on this: https://github.com/thelounge/lounge/compare/master...xPaw:irc-framework

[...]

PR #128, opened 2016-02-29
Correctly handle inline channels in messages 

### LINKED ISSUE #15: Inline channel parser easily breaks
A message like `&lt;<span` will easily cause it to fail. `<[color code here]test>` will too.

xPaw:
Moves inline channel parser into Handlebars, which means it is parsed before parsing colours and links, should fix a couple of parser related bugs.

### DIFF
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -1,8 +1,9 @@
 Handlebars.registerHelper(
 	"parse", function(text) {
 		var wrap = wraplong(text);
-		text = escape(text);
+		text = Handlebars.Utils.escapeExpression(text);
 		text = colors(text);
+		text = channels(text);
 		text = uri(text);
 		if (wrap) {
 			return "<i class='wrap'>" + text + "</i>";
@@ -23,17 +24,6 @@ function wraplong(text) {
 	return wrap;
 }
 
-function escape(text) {
-	var e = {
-		"<": "&lt;",
-		">": "&gt;",
-		"'": "&#39;"
-	};
-	return text.replace(/[<>']/g, function (c) {
-		return e[c];
-	});
-}
-
 function uri(text) {
 	return URI.withinString(text, function(url, start, end, source) {
 		if (url.indexOf("javascript:") === 0) {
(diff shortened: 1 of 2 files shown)


astorije:
Sadly, I am :-1: on this. One of the networks I'm on almost exclusively uses `&channels` (it's not a personal/small IRC network btw).
Are there any other alternatives we could explore?

xPaw:
What network does that?

astorije:
> What network does that?

irc.w3.org :P

tguild:
W3C is following this project and strongly contemplating replacing our qwebirc instance

astorije:
> I might have a better fix for this in mind, will try.

If you do, might want to consider closing and opening a new PR instead of ammending commit to keep change history :-)

xPaw:
Okay, pushed a proper fix.

dgw:
Wouldn't it be better to parse the `CHANTYPES` sent by the server? Maybe not in this PR, but eventually.

xPaw:
> Wouldn't it be better to parse the CHANTYPES sent by the server? Maybe not in this PR, but eventually.

Yes, that's not a terrible idea. #24

client/js/libs/handlebars/parse.js
```diff
@@ -48,6 +38,12 @@ function uri(text) {
 	});
 }
 
+function channels(text) {
+	return text.replace(
```
astorije: Could you re-add here the comment you deleted below please?

```js
// Channels names are strings (beginning with a '&' or '#' character)
// of length up to 200 characters.
// See https://tools.ietf.org/html/rfc1459#section-1.3
```

astorije:
@xPaw, looks nice at first glance, I'll give it a deeper review and an actual test.

Could you comment your code a bit, to be nice with future ourselves, please?
If you also add some details to the commit message on top of the comment, well that would be truly terrific :-)

xPaw:
I've pushed a fix to correctly handle characters that cannot be in the channel name before the `#` (like commas).

maxpoulin64:
Code looks good to me and it works very well. Also fixes a whole world of broken link parsing so I like it. :+1:

astorije:
:+1:

PR #129, opened 2016-03-01
Fixes channels listed on the whois message to be links

MaxLeiter:
There's probably a better way to do this...

this vs that 

Notice the commas

### DIFF
--- a/src/plugins/irc-events/whois.js
+++ b/src/plugins/irc-events/whois.js
@@ -28,6 +28,9 @@ module.exports = function(irc, network) {
 		};
 		for (var k in data) {
 			var key = prefix[k];
+			if (key === "on") {
+				data[k] = data[k].join(", ");
+			}
 			if (!key || data[k].toString() === "") {
 				continue;
 			}


astorije:
Alright, so if I'm correct, it's failing because of the comma *before* the channel name, like `,#channel`.
For this fix, I'd rather have `.join(", ")` to keep the comma and the clickable channel names.

Then there is also the problem of `+#channel` and `@#channel`. Not sure how to address this (we can't make `#something` match globally or URLs will be pretty messed up) nor if we should address it as part of this PR. It depends if the solution to this problem affects or is affected by your fix or not.

Also, would you mind fixing the build? :-P

MaxLeiter:
Not sure why the build is failing and have no experience with Jenkins.

maxpoulin64:
In my opinion, the message parser should be fixed to support those instead of trying to rewrite the whois response to work with it. I have a feeling that kind of issue is bound to happen again.

astorije:
I agree with @maxpoulin64. Any suggestion for this? It would be sad to refuse a PR that solves an actual issue in favor or a potential better fix that could not happen anytime soon.

@MaxLeiter, if you open Travis CI (not Jenkins :P), it will tell you that JS style has 2 issues:

```js
if(key == "on") {
// ... should be ...
if (key === "on") {
```

Notice the whitespace after `if` and the triple-equal sign.
Also, you can run `npm install && npm test` locally and you'll have the same output, no need to parse Travis CI's output :-)

xPaw:
As per the IRC RFC, channel names can not contain commas, and thus that character can be used to as whitespace in inline parser. I will add this in #128.

MaxLeiter:
Updated, but still has the issue with OP/voice, as @xPaw mentioned

xPaw:
I think there is no reason to keep this PR open as mine fixes the same issue.

astorije:
@MaxLeiter, @maxpoulin64, @xPaw, now that #128 is merged, is this PR still relevant?

If the links are fixed, we could at least add a space between each comma to present it nicer, no?

MaxLeiter:
I'll check out #128 when I get the chance, and there is a space after the commas

xPaw:
Closing in favour of #161.

PR #131, opened 2016-03-01
Switch a power-off icon for logging out

astorije:
The current icon is a lock, commonly used for encryption stuff. As a matter of fact, when @JocelynDelalande and I started to work on OTR support for Shout (TBC on The Lounge :-) ), we used the lock icons to signify that the session was encrypted, unencrypted, ...

Before | After
--- | ---
screen shot 2016-03-01 at 01 33 59 | screen shot 2016-03-01 at 01 34 24

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -480,7 +480,7 @@ button,
 }
 
 #footer .sign-out:before {
-	content: "\f13e";
+	content: "\f011";
 }
 
 #main {


MaxLeiter:
I like it, makes more sense to me. :+1:

YaManicKill:
Yeah, makes sense to me, too. :+1:

xPaw:
:+1:

PR #134, opened 2016-03-02
Fix minor issues with the main HTML file

astorije:
- Simplify `checked` attributes
- Use double quotes when possible
- Remove `class` duplicates
- Fix labels in password settings
- Remove `<small>` around version number


maxpoulin64:
Pretty minor, looks good to me too :+1:

PR #144, opened 2016-03-05
Invoke handlebars outside of grunt, generate sourcemap

### LINKED ISSUE #65: grunt build fails to build client/js/lounge.templates.js
the build task does not compile the file client/js/lounge.templates.js

running the command it is supposed to spawn manually does successfully compile the file which suggests that the task is not running at all when using grunt.

this is on Windows 10 with node.exe in the system PATH.

xPaw:
Fixes #65 (template compiling on windows).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@ npm-debug.log
 
 # Built assets created at npm install/prepublish time
 # See https://docs.npmjs.com/misc/scripts
+client/js/libs.min.js.map
 client/js/libs.min.js
 client/js/lounge.templates.js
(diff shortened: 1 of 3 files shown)


maxpoulin64:
Works fine here, plus sourcemaps are really nice to have to help debugging instead of editing and rebuilding. Big :+1: from me.

YaManicKill:
Yeah, looks good to me. To be honest, I'd be for removing grunt entirely if the only thing we are now doing with it is combining the lib files. We don't use any of the big features of grunt, so we may as well just go all in on the npm scripts.

Either way, :+1: from me, very happy with sourcemaps (always good). Merging.

PR #145, opened 2016-03-05
Fix stripping multiple colours from notifications

xPaw:
It was removing a single match only, needed a global flag.


astorije:
:+1: and merging.

PR #149, opened 2016-03-05
Relax inline channel style

xPaw:
Fixes hover effect on channels in topics


maxpoulin64:
Well that would be hard to not :+1: that. Merging since it's so minor.

PR #151, opened 2016-03-06
Mute color of the topic actions

astorije:


maxpoulin64:
Yeah I noticed that earlier. :+1:

Issue #68, opened 2016-02-18
Replace URIjs with alternative


MaxLeiter:
We currently only use it once in /client/js/libs/handlebars/parse.js, we can probably create our own function or find one online that doesn't require a whole library to use.

xPaw:
I suggest looking at how Kiwi IRC does it.

xPaw:
I took a look at it again, and I think URI.js should stay, as they handle a lot of different and obscure cases with URLs.

PR #152, opened 2016-03-06
Replace message parser with a new one

maxpoulin64:
This one should be really hard to break/unbreakable. It should be just as complete as the old one and address #15, #128, #129 and possibly various other unexpected breakages.

It solves the problem by making it so when a parser matches, it is responsible for recalling the parser with what it allowed to be parsed further.

---

Ideally I would have liked to avoid using closures, but they are currently needed for two reasons:
* When a parser matches, it is expected full control. Therefore, we need a way to stop the main loop
* Since I'm abusing regexes and doing several passes, it is a good addition to skip the already processed parsers to avoid redundant work

This fix is also potentially pretty rough on the stack, but considering IRC limits the length of lines to 512 characters, we can't possibly call over 512 deep.

This is the test case I used, which seems to cover most of it. https://d.max-p.me/irc/parser.html

Other various points:
* This breaks parsing of `(http://www.example.com/)` links, as parenthesis are technically valid URLs. It does parse `<http://www.example.com/>` just fine however! (easy fix if we want to bring it back)
* It removes the `<i class="wrapper">` on purpose. No idea why it was there, but everything it did is fixable otherwise.
* It doesn't support `\x16` to invert colors. It never worked anyway, and I'm not sure how doable this is using CSS/Javascript.
* It also doesn't support color 99: it's not even documented everywhere and there's `\x0F` to clear formatting anyway.
* Everything in the regexes is there on purpose. Greedy vs non, greedy, repetitions, weird character classes. It's all there because it's needed. I don't think it can be reduced any further.

--- 

@astorije also asked the possibility of adding testing for all those regexes, if anyone has any idea how this could be done please suggest. A simple map of input -> expected outputs should do.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1348,69 +1348,101 @@ button,
   * https://github.com/megawac/irc-style-parser
   * Colours are credit to http://clrs.cc/
   */
-.irc-fg0 { color: #fff; }
+.irc-fg0,
+#chat .text .irc-fg0 a { color: #fff; }
 
-.irc-fg1 { color: #000; }
+.irc-fg1,
+#chat .text .irc-fg1 a { color: #000; }
 
-.irc-fg2 { color: #001f3f; }
+.irc-fg2,
+#chat .text .irc-fg2 a { color: #001f3f; }
 
-.irc-fg3 { color: #2ecc40; }
+.irc-fg3,
+#chat .text .irc-fg3 a { color: #2ecc40; }
 
-.irc-fg4 { color: #ff4136; }
+.irc-fg4,
+#chat .text .irc-fg4 a { color: #ff4136; }
 
-.irc-fg5 { color: #85144b; }
+.irc-fg5,
+#chat .text .irc-fg5 a { color: #85144b; }
 
-.irc-fg6 { color: #b10dc9; }
+.irc-fg6,
+#chat .text .irc-fg6 a { color: #b10dc9; }
 
-.irc-fg7 { color: #ff851b; }
+.irc-fg7,
+#chat .text .irc-fg7 a { color: #ff851b; }
 
-.irc-fg8 { color: #ffdc00; }
+.irc-fg8,
+#chat .text .irc-fg8 a { color: #ffdc00; }
 
-.irc-fg9 { color: #01ff70; }
+.irc-fg9,
+#chat .text .irc-fg9 a { color: #01ff70; }
 
-.irc-fg10 { color: #39cccc; }
+.irc-fg10,
+#chat .text .irc-fg10 a { color: #39cccc; }
 
-.irc-fg11 { color: #7fdbff; }
+.irc-fg11,
+#chat .text .irc-fg11 a { color: #7fdbff; }
 
-.irc-fg12 { color: #0074d9; }
+.irc-fg12,
+#chat .text .irc-fg12 a { color: #0074d9; }
 
-.irc-fg13 { color: #f012be; }
+.irc-fg13,
+#chat .text .irc-fg13 a { color: #f012be; }
 
-.irc-fg14 { color: #aaa; }
+.irc-fg14,
+#chat .text .irc-fg14 a { color: #aaa; }
 
-.irc-fg15 { color: #ddd; }
+.irc-fg15,
+#chat .text .irc-fg15 a { color: #ddd; }
 
-.irc-bg0 { background: #fff; }
+.irc-bg0,
+#chat .text .irc-bg0 a { background: #fff; }
 
-.irc-bg1 { background: #000; }
+.irc-bg1,
+#chat .text .irc-bg1 a { background: #000; }
 
-.irc-bg2 { background: #001f3f; }
+.irc-bg2,
+#chat .text .irc-bg2 a { background: #001f3f; }
 
-.irc-bg3 { background: #2ecc40; }
+.irc-bg3,
+#chat .text .irc-bg3 a { background: #2ecc40; }
 
-.irc-bg4 { background: #ff4136; }
+.irc-bg4,
+#chat .text .irc-bg4 a { background: #ff4136; }
 
-.irc-bg5 { background: #85144b; }
+.irc-bg5,
+#chat .text .irc-bg5 a { background: #85144b; }
 
-.irc-bg6 { background: #b10dc9; }
+.irc-bg6,
+#chat .text .irc-bg6 a { background: #b10dc9; }
 
-.irc-bg7 { background: #ff851b; }
+.irc-bg7,
+#chat .text .irc-bg7 a { background: #ff851b; }
 
-.irc-bg8 { background: #ffdc00; }
+.irc-bg8,
+#chat .text .irc-bg8 a { background: #ffdc00; }
 
-.irc-bg9 { background: #01ff70; }
+.irc-bg9,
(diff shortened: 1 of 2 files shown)


client/js/libs/handlebars/parse.js
```diff
@@ -1,135 +1,140 @@
...
+		);
+	}
+
+
+	/**
+	 * Parses plaintext into safe HTML
+	 */
+	function parseText(msg) {
+		processed = true;
+		return msg.replace(/[<>"'\\\s&]/g, function(c) {
```
xPaw: Why are you escaping `\` and whitespace? Consider taking a look at https://github.com/wycats/handlebars.js/blob/0f054efcdb581790436c72341b72d92db52a7e3a/lib/handlebars/utils.js#L1

client/js/libs/handlebars/parse.js
```diff
@@ -1,135 +1,140 @@
...
+	 */
+	function parseColors(msg) {
+		return msg.replace(
+			/^(.*?)\x03([0-9]{1,2}(?:,[0-9]{1,2})?)([^\x0F\x03]*)\x0F?(.*)$/,
+			function(match, before, color, formatted, after) {
+				console.log("Format: ", match, before, color, formatted, after);
+				processed = true;
+				color = color.split(",");
+
+				var classes = "irc-fg" + parseInt(color[0], 10);
```
xPaw: `parseInt` is overkill and not needed here.

client/js/libs/handlebars/parse.js
```diff
@@ -1,135 +1,140 @@
-Handlebars.registerHelper(
-	"parse", function(text) {
-		var wrap = wraplong(text);
-		text = Handlebars.Utils.escapeExpression(text);
-		text = colors(text);
-		text = channels(text);
-		text = uri(text);
-		if (wrap) {
-			return "<i class='wrap'>" + text + "</i>";
```
xPaw: Remove the `.wrap` css class too.

client/js/libs/handlebars/parse.js
```diff
@@ -1,135 +1,140 @@
...
+	 * Parse mIRC color formatting
+	 * https://github.com/myano/jenni/wiki/IRC-String-Formatting
+	 */
+	function parseColors(msg) {
+		return msg.replace(
+			/^(.*?)\x03([0-9]{1,2}(?:,[0-9]{1,2})?)([^\x0F\x03]*)\x0F?(.*)$/,
+			function(match, before, color, formatted, after) {
+				console.log("Format: ", match, before, color, formatted, after);
+				processed = true;
+				color = color.split(",");
```
xPaw: Can't you just pull the values directly from the regex?

client/js/libs/handlebars/parse.js
maxpoulin64: I did it to be sure, but that was mostly guessing. I just replace it with the handlebars one.

client/js/libs/handlebars/parse.js
maxpoulin64: Actually it's needed for the cases where `\x0304color` is sent. We'd generate `<span class="irc-fg04">` which doesn't exist. So that makes sure it doesn't have a leading 0.

client/js/libs/handlebars/parse.js
maxpoulin64: Indeed, just fixed.

client/js/libs/handlebars/parse.js
maxpoulin64: Done

xPaw:
This PR also closes #68 if we keep the regex url parser.

xPaw:
Tested, and there's no obvious breakage this is causing besides not handling international domains. If we keep the new way of handling links, then URIjs should be removed from the repo.

astorije:
Regarding parentheses, a lot of the parsers out there appear to be pragmatic: they try to match balanced (opening and closing) brackets (parentheses, braces, square, chevrons, ...) to figure out if the parenthesis is part of the URL or not. Here is what I'm talking about in a few examples, applied to the GitHub parser:

- http://example.com/page_(abc)
- http://example.com/page_abc)
- http://example.com/page_abc(
- http://example.com/page_{abc}
- http://example.com/page_abc}
- http://example.com/page_abc{
- http://example.com/page_[abc]
- http://example.com/page_abc]
- http://example.com/page_abc[
- http://example.com/page_<abc>
- http://example.com/page_abc>
- http://example.com/page_abc<

The best live example I can think of is Wikipedia URLs, which very often have parentheses, like this one: https://en.wikipedia.org/wiki/Git_(software).
Of course, it could be a false assumption, like if a URL legitimately ends with a closing parenthesis without an opening one, but we are in that area of URI parsing where heuristics is key :-)

Turns out, I noticed this being an issue with URI.js as well and [opened an issue](https://github.com/medialize/URI.js/issues/247) at the time which, as you can see, I could not match it with a PR...

I see you don't use URI.js anymore, worth removing? Alternatively, can we leverage it instead of a custom parser? Whether URI.js or a better one, but I'm thinking there might be an existing library to do that that does it well.

client/css/style.css
```diff
@@ -1349,69 +1348,101 @@ button,
...
+.irc-bg13,
+#chat .text .irc-bg13 a { background: #f012be; }
 
-.irc-bg14 { background: #aaa; }
+.irc-bg14,
+#chat .text .irc-bg14 a { background: #aaa; }
 
-.irc-bg15 { background: #ddd; }
+.irc-bg15,
+#chat .text .irc-bg15 a { background: #ddd; }
```
astorije: Do we really have to do all that, no alternatives? :-(
(By the way, this probably breaks other themes...)

maxpoulin64:
Yeah,  a lot of people use () to wrap URLs. I will add it to the parser, it should be easy to add.

As for the parser, I'll have a look at it to see if I can fix xPaw's issue with international domains. I didn't even notice we were using a library for that and just started from scratch because we had issues with URLs, mainly `<URL>` didn't parse properly despite those not being valid characters for an URL at all. But if I can't fix it I'll see if I can switch it back to a library. But I don't think we really need a library for a simple regex so I would like to avoid it if that's an easy fix.

I'll also give a look at why it fails to parse Wendy's bold (and only bold!) a bit later. Should be an easy fix too.

client/css/style.css
maxpoulin64: Yes, we don't have the choice. It's either that, or using `color: inherit` and having no colors on links by default, not much better :/

And yes it might break themes, I'll have to look it any of them changed the colors. At least if it does break one, it will only display the link in the normal color so no biggie.

xPaw:
https://github.com/hexchat/hexchat/blob/78486b49d5b688ba5054631b849f8959846cf07b/src/fe-text/fe-text.c#L143

We need to handle some extra symbols, as we discussed in IRC.

maxpoulin64:
Alright this parser is a garbage. I'll rewrite another one later using a real tree. This *almost* worked perfectly considering regexes are a really nice way to express string manipulations, but IRC formatting apparently requires just a tiny bit more string formatting support than regexes can do. I'll have to rewrite all the same logic... but manually.

astorije:
@maxpoulin64, ping me when you restart your work on this so that we can sync up and come up with both the code (you) and the test suite for it (me) in parallel :-)

PR #154, opened 2016-03-06
Handle commands in a better way

### LINKED ISSUE #153: Don't intercept /(c|h|n)s shortcut commands
Intercepting `/ns`, `/cs`, and `/hs` and converting them to PRIVMSGs means the application has to assume that the services will have the expected names. There is a risk of the user accidentally sending a services command to another user who is using a common services nickname on a network that doesn't follow the convention.

By contrast, passing through `/ns`, `/cs`, and `/hs` to the server directly will either execute the services command as expected, or fail with an unknown command response—at which point the application can either retry using the conventional name of the service (more user-friendly) or pass the failure on to the user (safer).

Generally I'm in favor of aliases, but personally I'd rather risk a failed `/ns identify` command than risk The Lounge silently sending my password to an unknown user squatting on the name `nickserv`.

As a side note, it's also not entirely fair to Rizon that `/bs` isn't intercepted also. :stuck_out_tongue:

xPaw:
- Minor performance boost as commands stop processing as soon as one command is processed, `msg` as the first command, and any time you send a message, it won't try to execute any other command
- Removed `/ns`, `/cs`, `/hs` aliases, handled properly by the server now, fixes #153
- All unknown commands are always sent to the server
- Allows sending messages starting with `/` by escaping it as `//` (or more slashes)


src/client.js
```diff
@@ -270,25 +271,43 @@ Client.prototype.input = function(data) {
 	var client = this;
 	var text = data.text.trim();
 	var target = client.find(data.target);
+
 	if (text.charAt(0) !== "/") {
-		text = "/say " + text;
+		text = "say " + text;
+	} else {
+		text = text.replace(/^\//, "");
```
maxpoulin64: Couldn't you have used substr instead of a regex to strip the first char?

dgw:
I know there was just a release yesterday, but I'm already excited for the next one because it'll let me use `/znc` commands again. :+1:

maxpoulin64:
Looks good and works fine for me :+1:  Pretty good way to fix it without a rewrite.

src/client.js
```diff
@@ -270,25 +271,43 @@ Client.prototype.input = function(data) {
...
+			// Commands return true on successful execution
+			// lodash expects `false` to stop the loop
+			return !result;
 		} catch (e) {
 			console.log(path + ": " + e);
 		}
 	});
+
+	if (result !== true) {
+		target.network.irc.write(text);
```
dgw: Yes! This is the magic sauce that will let things like `/znc` work! :raised_hands:

astorije:
Please hold on before merging, I'd like to take second review :-)
(Either tonight or tomorrow night)

src/client.js
```diff
@@ -33,21 +33,22 @@ var events = [
 	"whois"
 ];
 var inputs = [
+	// These inputs are sorted in order that is most likely to be used
+	"msg",
+	"whois",
+	"part",
 	"action",
+
```
astorije: Can you remove this blank line?

src/client.js
```diff
@@ -33,21 +33,22 @@ var events = [
...
 	"mode",
-	"msg",
 	"nick",
 	"notice",
-	"part",
 	"quit",
 	"raw",
-	"services",
 	"topic",
-	"whois"
```
astorije: This might be a stupid question, so don't be too tough on me :-)

Rather than making assumptions on most frequently used commands and calling them one by one until we have a match, why don't we directly math the input to the command name?
From what I see, `/join` matches to `join`, `/nick` matches to `nick`, ...
If this is not 100% the case, we can either forward commands, rename them, or treat the non-matching one differently.

src/client.js
```diff
@@ -270,25 +271,43 @@ Client.prototype.input = function(data) {
...
 	if (text.charAt(0) !== "/") {
-		text = "/say " + text;
+		text = "say " + text;
+	} else {
+		text = text.substr(1);
+
+		// If after removing first / we still have a /, send it as a message
+		if (text.charAt(0) === "/") {
+			text = "say " + text;
+		}
```
astorije: I'd go a tiny bit differently for this.

First of all, remove the `.trim()` L271/R272 (or replace it with a [`.replace(/~+$/, '')`](https://stackoverflow.com/questions/8141718/javascript-need-to-do-a-right-trim) if you think it would be harmful to keep trailing whitespaces). That allows for another common way of escaping `/cmd`: ` /cmd`. It's frankly annoying that The Lounge currently takes the liberty of deciding that leading whitespaces should be removed.

Secondly, I'd suggest replacing this block with the following:

```js
// This is either a normal message or a command escaped with a leading `/`
if (text.charAt(0) !== "/" || text.charAt(1) === "/") {
  text = "say " + text.replace(/^\//, "");
} else {
  text = text.substr(1);
}
```

It seems more straightforward to me and the regex is basic enough to keep this readable (I'm on that "Regexes obfuscate your code team in general :-) ). What do you think?

Anyway, nice escaping the slash with another slash, and removing the trim in addition will leave an entire set of choices to the users to escape their stuff.

src/client.js
```diff
@@ -270,25 +271,43 @@ Client.prototype.input = function(data) {
...
+
 	var args = text.split(" ");
-	var cmd = args.shift().replace("/", "").toLowerCase();
+	var cmd = args.shift().toLowerCase();
+	var result;
+
 	_.each(inputs, function(plugin) {
 		try {
 			var path = "./plugins/inputs/" + plugin;
 			var fn = require(path);
```
astorije: Can we load the `require`s for all `input`s once and for all after we create the initial array L35?
Unless I am being mistaken, `require`ing all inputs at every command seems like an unnecessary overhead to me.

src/client.js
```diff
@@ -270,25 +271,43 @@ Client.prototype.input = function(data) {
...
+			text = "say " + text;
+		}
 	}
+
 	var args = text.split(" ");
-	var cmd = args.shift().replace("/", "").toLowerCase();
+	var cmd = args.shift().toLowerCase();
+	var result;
+
 	_.each(inputs, function(plugin) {
```
astorije: Instead of `_.each`, use the [native `.some()` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some).
You can then `return fn.apply(...)` instead of `result = fn.apply(...)`, and have `var result = inputs.some(...)`.
Makes it cleaner and you can drop `return !result;` as well as your comment R300-302 :-)

src/client.js
```diff
@@ -270,25 +271,43 @@ Client.prototype.input = function(data) {
...
 				target.network,
 				target.chan,
 				cmd,
 				args
 			]);
+			// Commands return true on successful execution
+			// lodash expects `false` to stop the loop
+			return !result;
 		} catch (e) {
 			console.log(path + ": " + e);
```
astorije: If you consider my comment above, this `catch` should be moved along with the `require`s outside of the loop execution, making the `var result = inputs.some(...)` even more direct.
By doing so, you also make sure that any bug coming from a discrepancy between the input list and the available files in the plugins folder is detected as soon as The Lounge is started, and not whenever someone uses the buggy command.
Also, the current code silently absorbs the error without re-throwing it. All plugins are (at least at the moment) coming from us, so having the app crash miserably is acceptable here: there is no point in providing a recovery state from there IMO.

The current code is a bit worrying I think, because it fairly difficult to detect a silenced bug as is, unless you keep an eye on the log (which, by the way, is not even the error log...).

src/plugins/inputs/join.js
```diff
@@ -2,8 +2,11 @@ module.exports = function(network, chan, cmd, args) {
 	if (cmd !== "join") {
 		return;
```
astorije: Missing `return false;` here?

src/plugins/inputs/kick.js
```diff
@@ -2,8 +2,11 @@ module.exports = function(network, chan, cmd, args) {
 	if (cmd !== "kick") {
 		return;
```
astorije: Missing `return false;` here?

src/plugins/inputs/msg.js
```diff
@@ -5,14 +5,14 @@ module.exports = function(network, chan, cmd, args) {
 		return;
```
astorije: Missing `return false;` here?

astorije:
@xPaw, I left you a few comments inline. This is globally great work and I am eager to see what it looks like when you address the comments (assuming I am not mistaken on them).
Also, note that some of the comments above will probably be unnecessary (or less necessary) if [my comment about direct lookup vs. sequential lookup of inputs](https://github.com/thelounge/lounge/pull/154#discussion_r55474450) applies.

src/plugins/inputs/join.js
xPaw: Doesn't matter since both `!undefined` and `!false` return `true`.

src/client.js
xPaw: I put whois at the top because it's triggered by clicking on a username. Ideally though, yes, we would have a direct hashmap of the commands.

astorije: After getting that the point of this PR was to avoid a full rewrite, you can disregard this earlier comment.
I'll implement this in a further PR as well as include a bunch of tests for these at the same time, I should get to it within the next days.

By the way, I agree with this order, was just saying it could get opinionated at some point, but that won't matter when I add direct mapping.

src/plugins/inputs/join.js
astorije: It does if for any reason we add code checking `== false`, as `undefined` evaluates to `false` but is not `false`. My main issue is that this doesn't keep consistency in returned types, but I won't strongly oppose since this code will evolve anyway.

astorije:
Alright @xPaw, I added a note on all comments that you can ignore for this PR, keeping them around for archive purpose. Left the others untouched.

src/client.js
xPaw: I think the current way is a little bit better, as it won't invoke `replace` when you send a normal message.

xPaw:
@astorije Fixed some of the things you pointed out except for the `/say` handling.

src/client.js
```diff
@@ -268,18 +268,28 @@ Client.prototype.setPassword = function(hash) {
 
 Client.prototype.input = function(data) {
 	var client = this;
-	var text = data.text.trim();
+	var text = data.text.replace(/\s+$/, "");
```
xPaw: @astorije This is for you. You're right it's silly to trim the start.

src/client.js
astorije: Definitely less readable though, and suggested solution will definitely not have any performance impact.

src/client.js
astorije: Very cool, thanks!

astorije:
:+1: but I will test locally before merging tonight if you don't mind, @xPaw.

src/client.js
```diff
@@ -268,18 +268,24 @@ Client.prototype.setPassword = function(hash) {
...
-		text = "/say " + text;
+
+	// This is either a normal message or a command escaped with a leading '/'
+	if (text.charAt(0) !== "/" || text.charAt(1) === "/") {
+		text = "say " + text.replace(/^\//, "");
+	} else {
+		text = text.substr(1);
 	}
+
 	var args = text.split(" ");
```
astorije: While testing this, I found an issue: any text starting with one or more whitespace does not send anything nor produce an error. It's because ` test` here becomes `['', 'test']`.
We can fix this with the first of the 2 solutions below. However, I wonder if it is any good or bad to keep multiple whitespaces within commands. For example, `/join   #channel` becomes `['join', '', '', '#channel']` then back to `join   #channel`... I don't know if that could be an issue in some scenarios, in which case the second solution below would solve it.

```js
var args = text.trim().split(" ");
var args = text.split(" ").filter(function (a) { return a.length !== 0; });
```

I might be missing an alternative, but I wanted to address the bug in the first place and look for potential solutions :-)

xPaw: I've reverted the trim change, it will have to go into the next overhaul.

astorije:
:+1: and merging.
@xPaw, thanks for this! I will commit my local changes (`require` once and for all, direct addressing) within a couple days.

PR #155, opened 2016-03-06
Fix #96: Debounce trimming messages in channels

### LINKED ISSUE #96: "show more" is constantly checked on an interval
https://github.com/thelounge/lounge/blob/1e8ca51d47f78e30e8d429cfc164258386e0f5d7/client/js/lounge.js#L818-L825

This should just happened whenever a new message is received or something like that.

xPaw:
This makes it clean up messages in inactive channel 5 seconds after receiving the last message. If you're idling and no messages are coming in, no action is taken, compared to the previous solution, which constantly checked all inactive channels and tried to remove messages in each one of them.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -747,6 +747,20 @@ $(function() {
 			.click();
 	});
 
+	var debouncedHistoryTrim = debounce(function(target) {
+		target = chat.find(target);
+
+		// If by the time debounced function fired and
+		// the channel became active, do nothing
+		if (target.hasClass("active")) {
+			return;
+		}
+
+		if (target.children(".messages").slice(0, -100).remove().length) {
+			target.find(".show-more").addClass("show");
+		}
+	}, 5000);
+
 	chat.on("msg", ".messages", function(e, target, msg) {
 		var button = sidebar.find(".chan[data-target='" + target + "']");
 		var isQuery = button.hasClass("query");
@@ -790,8 +804,8 @@ $(function() {
 			}
 		}
 
-		button = button.filter(":not(.active)");
-		if (button.length === 0) {
+		if (!button.hasClass("active")) {
+			debouncedHistoryTrim(target);
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


maxpoulin64:
I like the idea, but I have a concern with this on large channels. Some channels are pretty crowded, meaning the function could possibly never be triggered, and let messages pile up. An edge case for that would be Twitch's IRC servers where messages are just constantly flowing in like crazy.

Another problem I see with it is that 5 seconds is a bit short, so on slow but active channels messages would be cleaned all the time, which might be a performance hit in the end (and a battery drain on mobile).

I think one way to fix this could be to clean messages on both a longer timer and a set amount of messages, whichever comes first. And not rearm the timer if no messages are coming in, for really idle channels.

---

Alternatively, couldn't we also just pause updating non-active channels entirely and instead just store an array of messages, and render like 2-3 pages worth of messages when the channel becomes active? It's not rare for my client to pile up a couple thousands of messages in some very active channels, essentially wasting a lot of CPU rendering templates and doing DOM manipulations.

Maybe we could even use the server's assistance here and actually don't send the data to the client *at all*, saving an awful lot of bandwidth and precious CPU cycles on mobile. I think this is what Glowing Bear does, and while it does add a slight delay, we can easily hide it with a short fade in and out while the channel data is being fetched.

xPaw:
> Some channels are pretty crowded, meaning the function could possibly never be triggered, and let messages pile up. An edge case for that would be Twitch's IRC servers where messages are just constantly flowing in like crazy.

That's a good edge case, yes.

> I think one way to fix this could be to clean messages on both a longer timer and a set amount of messages, whichever comes first. And not rearm the timer if no messages are coming in, for really idle channels.

That's exactly what it does though. It only sets the timer after a message comes in.

> Maybe we could even use the server's assistance here and actually don't send the data to the client at all, saving an awful lot of bandwidth and precious CPU cycles on mobile.

I'd rather not do that because it will introduce latency, when you switch to a channel and have to wait for server to send you messages.

lindskogen:
On the "function could possibly never be triggered" issue, that is why [`_.debounce`](https://lodash.com/docs#debounce) has the `maxWait` parameter

PR #146, opened 2016-03-05
No longer notify on messages sent by yourself from different tab.

YaManicKill:
Rebased from #6

PR #157, opened 2016-03-07
Make sure self messages are never highlighted

### LINKED ISSUE #156: Self mentions should not trigger highlights
Example:

screen shot 2016-03-07 at 01 02 02

astorije:
Fixes #156.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -36,16 +36,15 @@ module.exports = function(irc, network) {
 			text = text.replace(/^\u0001ACTION|\u0001$/g, "");
 		}
 
-		var highlight = false;
-		textSplit.forEach(function(w) {
-			if (w.replace(/^@/, "").toLowerCase().indexOf(irc.me.toLowerCase()) === 0) {
-				highlight = true;
-			}
-		});
+		var self = (data.from.toLowerCase() === irc.me.toLowerCase());
 
-		var self = false;
-		if (data.from.toLowerCase() === irc.me.toLowerCase()) {
-			self = true;
+		var highlight = false;
+		if (!self) { // Self messages should never be highlighted
+			textSplit.forEach(function(w) {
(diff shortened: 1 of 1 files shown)


dgw:
I was thinking about this yesterday and didn't get around to making an issue for it, and here's a PR all set to fix it. :+1:

astorije:
This is a quick fix, but maybe a wee bit longer-term solution should be to parse and trigger on the client instead, like #152 is doing, no?

maxpoulin64:
Wouldn't it be better to check for self before the loop and avoid splitting the message entirely instead of doing it and overriding it right after?

Also, my PR doesn't actually touch that. This function is only responsible of turning IRC messages into HTML. If we want to handle highlights on the client, it would fit more in the msg handler. One minor thing for client-side highlights: if the user changes nicks and reload the page, the previous highlights will no longer be highlighted. Storing it in the server object has the benefit of making it permanent.

xPaw:
@maxpoulin64 is right, check for `self` and then avoiding message splitting would be a better way.

YaManicKill:
Does this remove the need for #146 ?

xPaw:
Hm, looks like yes, it does.

YaManicKill:
Yeah, that's fine. I think it's better dealing with it this way, especially for this issue.

astorije:
Good point guys, I'll fix that. Also, I have a decent (well, in my opinion :p) idea how to make these testable, will post here when I get to it.

astorije:
Hi all, took your comments into consideration to make things nicer. Thanks for looking carefully and sorry for being reckless in the first place :p
So now, the highlight lookup is only done if this is not a message from `self`. Also, I added something else: using `some()`, I stop the lookup early when one of the words matches. @maxpoulin64, I specifically thought of your addition to performance matters while doing so :D

maxpoulin64:
:+1:  looks all good to me

PR #158, opened 2016-03-07
Remove highlights from @mentions

astorije:
Assuming `nick` is the nick of the current user:

1. `nick` triggers a highlight
2. `@nick` triggers a highlight, even though starting with `@` prevents the auto-complete to show up
3. `@@nick` does not trigger a highlight
4. `nicki-minaj-is-my-hero`... does trigger a highlight...

This PR removes the second case. I don't see a value in highlighting `@nick`, unless we want to give a GitHub look-and-feel, in which case we should also enable in auto-complete, ... Not sure we want that, after all this is not how IRC deals with nicks, as opposed to, say, GitHub and Twitter.

But maybe I'm very wrong, let the caucus decide :-)

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -38,7 +38,7 @@ module.exports = function(irc, network) {
 
 		var highlight = false;
 		textSplit.forEach(function(w) {
-			if (w.replace(/^@/, "").toLowerCase().indexOf(irc.me.toLowerCase()) === 0) {
+			if (w.toLowerCase().indexOf(irc.me.toLowerCase()) === 0) {
 				highlight = true;
 			}
 		});


richrd:
I agree that it's "incorrect" to use @nick in IRC, but on the other hand, I'd like to be highlighted even if someone happens to do that. In that sense I don't see it as an issue that @nick causes a highlight.

YaManicKill:
Just as a quick note, as far as I know, most IRC clients highlight for the literal string, even if it is included in another word.

dgw:
They often do, but the good ones try to avoid highlighting if the user's nick is immediately preceded or followed by a character that would be valid in a nickname.

MaxLeiter:
I personally like having my name highlighted no matter what, as some people do use @<nickname> GH-style. :-1: for me

astorije:
Seeing how discussion in #159 and on the IRC channel are going, it looks to me that this is not the way to go. @maxpoulin64, feel free to close this as is or to request any further change. It could be that this PR (or its short-term evolution) is a temporary change until better comes in, or not. I'll let you decide and I'll adapt :-)

maxpoulin64:
Yeah I will close this and make a PR later to match under any circumstances instead. I think this is what people expect, and also what other IRC clients do in general.

PR #161, opened 2016-03-07
Make whois a client template

xPaw:
chrome_2016-03-07_17-10-29

Supercedes #129

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -840,6 +840,12 @@ button,
 	color: #2ecc40;
 }
 
+#chat .whois .from:before {
+	font-family: FontAwesome;
+	content: "\f007";
+	color: #2ecc40;
+}
+
 #chat .nick .from:before {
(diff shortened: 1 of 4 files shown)


astorije:
Nice @xPaw, it's looking good!
One minor suggestion, could we have channels displayed as `is on the following channels: #channel1, #channel2, and #channel3`?
It would look even more user-friendly :-)

maxpoulin64:
I like that, much better looking and better handled. :+1:

xPaw:
@astorije That's slightly tricky because it uses `each` loop in the template.

astorije:
We could either use tricks to deal with that in the template, but yes, not pretty. A better way would be to define a human-readable list helper (along those lines: http://stackoverflow.com/a/18831911/1935861) but definitely not worth holding off this PR. I'll make an additional PR later for this :-)
:+1:

PR #163, opened 2016-03-07
Remove string.contains

xPaw:
The code that used it is gone. This is no longer needed.


YaManicKill:
Also, beautiful ES2016 contains string.includes 

:+1:

Issue #121, opened 2016-02-28
Invite message is not built correctly


astorije:
<img width="339" alt="screen shot 2016-02-28 at 02 47 07" src="https://cloud.githubusercontent.com/assets/113730/13378080/a361d400-ddc5-11e5-8d51-6a4f54ac10d8.png">

Lots of PRs and issues about the `invite` command, so I'm not entirely sure which one broke it. Sorry if it's me :-)

Underlying markup:

```html
<div class="msg invite">
	<span class="time">
		02:46
	</span>
	<span class="from"></span>
	<span class="text">
		<a href="#" class="user">jeremie</a>
invited
<a href="#" class="user">lounge-user <span class="inline-channel" role="button" tabindex="0" data-chan="#astorije</a>">#astorije</span></a>
to


	</span>
</div>
```

xPaw:
I believe this is a bug in `slate-irc`.

astorije:
Weird, it indeed depends on the IRC network! On Freenode it works fine, but the previous example was taken on W3C's IRC network: https://www.w3.org/Project/IRC/

I also confirm this bug already exists on `v1.2.1`, so it was not introduced after this release.

xPaw:
Yep, it's a bug in slate-irc's parser. They expect and assume to always have `!` character in the message prefix, which would be there if full hostmask was sent, but I assume W3C's network only sends the name.

astorije:
Fixed by #167.

Issue #115, opened 2016-02-28
Show connection attempt status messages in network buffer


easymac:
While a network connection is brokered, no messages appear in the network buffer until the connection is successful. We should see a `Connecting to [host or name]` message when a connection is started instead of a blank screen.

Error messages should also be piped there. Currently, connections that time out leave the network buffer entirely blank while emitting the following message on the server

```
Client#connect():
Error: connect ETIMEDOUT
```

In the event that auto-reconnect or auto-retry connections come to exist, I would also like to see `Connection [lost or failed], retrying . . .` messages.

xPaw:
One of the things required for this to work correctly, is to move this code all the way up in the function: https://github.com/thelounge/lounge/blob/master/src/client.js#L213-L216

Then we will be able to send messages to network tab with what's happening.

astorije:
This issue seems to be fixed by #167.

Issue #56, opened 2016-02-16
Show all messages from server


v0l:
Show all info on server chan instead of just motd and notice messages. 

![img](https://void.cat/5cb1efb4bdb8861273a0e48fb09ddf223b73d7ad)

PS: The good snipping tool!

xPaw:
I don't think slate-irc sends any events with that information. Another issue for #24

v0l:
Ah, I doubt it would be difficult to implement your own client code for IRC, its a very basic protocol.

vectr0n:
+1 Would be a good option to see server messages, or the option too.

easymac:
Slate uses the `motd` event for these messages; definitely supported.

xPaw:
#167 shows connection related messages, nothing about server capabilities yet though.

xPaw:
"all messages" is not very descriptive. master now shows all the connection related messages that are raised by irc-fw. If you would like to see other specific information (e.g. server capabilities), create a new issue for each of these please.

PR #167, opened 2016-03-08
Drop slate-irc, switch to irc-framework

### LINKED ISSUE #166: Only the first nick is displayed when multiple people are Opped/Voiced
When multiple users are opped/Voiced/deoped/Devoiced, only the nick of the first person is displayed. :/

capture d ecran 2016-03-08 10 00 17
Meanwile on an other IRC client (Hexchat here)
capture d ecran 2016-03-08 10 00 02

xPaw:
:nail_care: This is rather huge, opening a PR for tracking and discussion.

What's left to do:
- [x] Merge #161 before fixing whois
- [x] Merge #154
- [x] Invites are not handled, https://github.com/kiwiirc/irc-framework/issues/19
 - [x] Confirm #121
- [x] User list does not strip modes correctly https://github.com/kiwiirc/irc-framework/issues/17
- [x] User mode not updated in mode event
 - Does a `NAMES` request just like the current `master` solution, this will need to be tackled in a future PR
- [x] Need to end the connection correctly (quit command) https://github.com/kiwiirc/irc-framework/issues/21
- [x] The only test we have is broken
- [x] `from_server` is not being set to true and causes server notices to be sent into a query window
- [x] Channels need to be searched case insensitively
- [x] Timestamps in `topic set by` need to be parsed
- [x] Missing sender name in quakenet notices crashes https://github.com/thelounge/lounge/pull/167#issuecomment-198028907

Thing to keep in mind: the framework keeps its own track of channels and users in them, we can use it to our advantage.

New features:
- IRCv3 compliance
 - SASL should work
 - [`server-time`](http://ircv3.net/specs/extensions/server-time-3.2.html) support on the client
 - Not per spec, but `twitch.tv/membership` is supported
- Displays who and when set the topic
- Connection messages (like connecting, reconnecting, etc) in server window, partial #56 fix and #115
- Server window is created as soon as possible to allow displaying errors (like disallowed hostname)
- Fixes #166 to correctly display each affected user on mode changes
- We will be able to support reconnections, however they are disabled for now as I don't know how well lounge will handle them
- #121 should be fixed by itself since irc-framework correctly parses different messages
- Channel names are no longer forcefully lower cased (slate did that)
- Target groups (STATUSMSG) is now handled (e.g. /msg +#freenode hello)


lindskogen:
And we can write a middleware that persists messages to a sqlite database :+1:

xPaw:
Pinging @thelounge/maintainers to test this branch for any obvious breakage (besides the ones mentioned in op post)

astorije:
@xPaw, nice work! I'll look at what I can simply add in terms of tests before I move forward.

maxpoulin64:
Quick test so far, looks really great overall! Few things I noticed:

* When connecting to ZNC, all my past PMs are received from myself in one giant conversation. Probably has to do with the way messages are parsed?
* Topic timestamps are not parsed `Topic set by Max-P on 1457043909`
* All notices during connect are open as seperate windows. I new about the IRC server one, but it appears that NickServ and SaslServ also opens their own windows as well with their notices on connect. They probably should end up in the server window as well.

You probably know about those but I'm letting them here for tracking purposes.

lpoujol:
Connecting to Quakenet crashes lounge.
Seems because Quakenet sends notice at connexion that contains no source.

> telnet irc.quakenet.org 6667
> Trying 128.39.65.226...
> Connected to irc.quakenet.org.
> Escape character is '^]'.
> NOTICE AUTH :*** Looking up your hostname
> NOTICE AUTH :*** Checking Ident
> NOTICE AUTH :*** Found your hostname
> NOTICE AUTH :*** No ident response
> 

Here's lounge console with logs as xPaw asked when connecting to Quakenet : 
http://sebsauvage.net/paste/?89c9e53662aac5e1#kgfis+2KKzmP66s4CL/gMeOwPcaxGSQwgsDutQWAXrQ=

astorije:
@xPaw, just rebased this branch, don't forget to `git fetch && git reset --hard origin/irc-framework` before anything! (I will be pushing a commit in a few)

astorije:
Tests know reflect the PR correctly. It does show a couple breaks, but I didn't fix the code itself. I can take a look later.
(Also removed a trailing whitespace in your last commit ^^)

src/models/chan.js
```diff
@@ -23,18 +23,21 @@ function Chan(attr) {
...
-
-	["+", "%", "@", "&", "~"].forEach(function(mode) {
-		this.users = _.remove(
-			this.users,
-			function(u) { return u.mode === mode; }
-		).concat(this.users);
-	}, this);
+Chan.prototype.sortUsers = function(irc) {
+	var userModeSortPriority = {};
+	irc.network.options.PREFIX.forEach(function(prefix, index) {
```
astorije: I think it would be nicer to only pass the prefixes in the `sortUsers` function. To assist with that, we can have `Network.prototype.prefixes = function()` in `network.js` that returns the array of prefixes for that network. We can then call things like `chan.sortUsers(network.prefixes());` in a simple way (simplify tests too).

I was trying to provide it, but I noticed there is `options.PREFIX`, `serverOptions.PREFIX` and `prefixLookup`, all related. It seems to me that `sortUsers` should be using `serverOptions.PREFIX` instead of `options.PREFIX` but I didn't look at the code too closely. It would be better to keep a single source of truth and use methods to produce the others.

Also, in places calling `sortUsers()`, There are `irc.network` and `network`. Do they carry the same values?

Happy to help on this.

astorije:
@xPaw, I'll test more extensively and add stuff over the weekend. Meanwhile, could you add @maxpoulin64 and @lpoujol's comments to your punch list above?

src/models/user.js
```diff
@@ -4,8 +4,8 @@ module.exports = User;
 
 function User(attr) {
 	// TODO: Remove this
-	attr.name = attr.nick;
-	attr.mode = attr.modes[0] || "";
+	attr.name = attr.name || attr.nick;
+	attr.mode = attr.mode || (attr.modes && attr.modes[0]) || "";
```
xPaw: I think this should have been fixed in the test, as irc-fw will only be providing us with modes array.

src/models/chan.js
xPaw: > It seems to me that sortUsers should be using serverOptions.PREFIX instead of options.PREFIX

They're the same thing. `serverOptions` exists because its an object in *our* network instance, and it gets synchronized to the frontend client for later use. I didn't want to pipe the whole irc-fw options/networks objects to the client, as they have a lot of data.

src/models/user.js
astorije: I added changes here simply because of the TODO above. If the model changes because irc-framework makes us too, I think we should go all in: `name` could be reworded `nick` everywhere and the non-array `mode` could disappear.
In the meantime, it was an easy way for me to integrate the tests, until this TODO gets addressed.
@xPaw, what's your take about changing the model regarding `name`/`nick`/`mode`/`modes`?

src/models/chan.js
astorije: > I didn't want to pipe the whole irc-fw options/networks objects to the client, as they have a lot of data.

Yep, agreed!

However, I think those add some confusion to the codebase overall, I'm sure we can encapsulate this nicely in the models and interface that (`options`, `serverOptions`, ...) via methods. What do you think? I can try to spend some time on that if that's fine by you.

src/models/user.js
xPaw: It's not an easy change because it will affect a lot of files. I'd rather tackle it in a separate PR because this one is getting pretty big already.

src/models/chan.js
xPaw: We probably can change `toJSON` method in network to return whitelisted properties, like `export` does.

test/models/chan.js
```diff
@@ -35,7 +36,7 @@ describe("Chan", function() {
 				new User({name: "xPaw", mode: "~"}),
 				new User({name: "Max-P", mode: "@"}),
 			]});
-			chan.sortUsers();
+			chan.sortUsers({network: {options: {PREFIX: ["+", "%", "@", "&", "~"]}}});
```
xPaw: That's not how PREFIX is formatted by the way, which is why the test is failing. See https://github.com/kiwiirc/irc-framework/blob/48bb805653872e5f21210474fdfe439892be91b0/src/networkinfo.js#L12

vectr0n:
I have been running irc-framework for close to a week now without any issues. Great job guys :) :+1:

xPaw:
@maxpoulin64 and @lpoujol are also saying they've been running it without problems.

@maxpoulin64, @astorije, @YaManicKill can you guys start reviewing the code for any problems? I would really like to get this merged within the next week.

maxpoulin64:
Alright, so I just did a full round of test following the code to try every feature/command that are supported by the client. I fixed one issue that was an easy fix, and I found only one more [issue that happens to be on irc-framework's side](https://github.com/kiwiirc/irc-framework/pull/28).

Everything else looks like it works beautifully. Still using it as my main and only IRC client with no issues. This gets my :+1: once we update irc-framework to include the small fix I submitted. Very nice work done here @xPaw :clap: :tada:! Looking forward to all the bugs we'll be able to squash properly with slate out of our way!

xPaw:
@maxpoulin64 irc-fw updated. I've also added `/ctcp` command and ctcp response handler.

client/css/style.css
```diff
@@ -818,6 +819,11 @@ button,
 	color: #2ecc40;
 }
 
+#chat .ctcp .from:before {
+	font-family: FontAwesome;
+	content: "\f0f6";
+}
+
```
astorije: Would you mind keeping the CTCP commit in a different PR?
Not at all that I'd be against it, but the PR is already very long with about 1k changes and 40 commits and since I'd need to read about it (I don't know anything about it yet), I think this should go separate to expedite the whole thing more quickly.

client/css/style.css
```diff
@@ -860,6 +860,10 @@ button,
 	color: #f00;
 }
 
+#chat .msg.toggle .time {
+	visibility: hidden;
+}
+
```
astorije: I understand this commit is to keep a consistency on how messages are sent from the server to the client, in case some client changes / themes want to display the time for a link expander. Am I correct? Is there any other cases it would be useful?
I actually think it's fair to not even sent date from the server as there is essentially none.

Assuming I am correct, I think we should leave that out.

In the future, we might want to have a model for the link expander itself instead of deriving `Msg`, that would give us a bit more freedom I believe. But that's for another day.

Out of curiosity, why `visibility: hidden` instead of `display: none`? I think `display: none` would be better but I might be missing something.

src/client.js
```diff
@@ -3,11 +3,9 @@ var Chan = require("./models/chan");
 var crypto = require("crypto");
 var identd = require("./identd");
 var log = require("./log");
-var net = require("net");
 var Msg = require("./models/msg");
 var Network = require("./models/network");
-var slate = require("slate-irc");
```
astorije: :tada:

src/client.js
```diff
@@ -129,97 +125,69 @@ Client.prototype.connect = function(args) {
...
 
+	var nick = args.nick || "lounge-user";
+
+	var network = new Network({
+		name: args.name || "",
+		host: args.host || "",
+		port: parseInt(args.port, 10) || (args.tls ? 6697 : 6667),
+		tls: !!args.tls,
+		password: args.password,
+		username: args.username || nick.replace(/[^a-zA-Z0-9]/g, ""),
```
astorije: I know it was already there before but I never paid attention to that `.replace()` statement. It seems weird without further explanation, do you know where that comes from, IRC-something spec or `lounge`-specific decision?
Can we add a comment on that for API documentation purposes?

src/client.js
```diff
@@ -129,97 +125,69 @@ Client.prototype.connect = function(args) {
...
 			});
 			return;
 		}
 
-		args.host = config.defaults.host;
-		args.port = config.defaults.port;
-		args.tls = config.defaults.tls;
+		network.host = config.defaults.host;
+		network.port = config.defaults.port;
+		network.tls = !!config.defaults.tls;
```
astorije: Isn't `config.defaults.tls` a boolean in the config already?

src/plugins/irc-events/connection.js
```diff
@@ -0,0 +1,75 @@
...
+
+	client.emit("msg", {
+		chan: network.channels[0].id,
+		msg: new Msg({
+			text: "Network created, connecting to " + network.host + ":" + network.port + "..."
+		})
+	});
+
+	irc.on("raw socket connected", function() {
+		identd.hook(irc.connection.socket, network.username);
```
astorije: Associated with this change, could you remove [this line in `identd.hook()`](https://github.com/thelounge/lounge/blob/e0031dd67ec024753eebe473791532a1e81b8da0/src/identd.js#L12) so that the first argument is `socket` itself?
This is the only place where `identd.hook()` gets used.

src/client.js
xPaw: It was already like that, I didn't put in that regex.

src/client.js
xPaw: It was already like that. Since JS is loosely typed its a good idea to enforce a type like this here.

client/css/style.css
xPaw: Not sending a timestamp requires more work than necessary.

Using `display` will break the layout as we use tables for messages.

client/css/style.css
xPaw: It actually can't as current lounge handles CTCP requires as mere notices, and irc-fw has a separate callback for it.

src/plugins/irc-events/connection.js
xPaw: Ehh, it still needs to access `stream.socket`, so I would only be removing `|| stream` part.

src/client.js
astorije: Yup, that's why I said "I know it was already there before but I never paid attention to that `.replace()` statement." :-)

Still, I am wondering if you knew why it's there, and if so we can add a comment above.

src/client.js
astorije: I don't believe it was there before since the affected line was `args.tls = config.defaults.tls`. Ideally, `config` should already have correct types though (so this sort of coercion, if necessary, should be done in the config helper, or be responsible for stopping execution when misconfigured).
We don't do checks for `config.defaults`, let's keep it that way for this PR.

client/css/style.css
astorije: > Not sending a timestamp requires more work than necessary.

What do you mean?

> Using `display` will break the layout as we use tables for messages.

Fair point, thanks.

xPaw: > What do you mean?

Would need to do something special like setting time to `null`, and then doing if checks both on the server and the client so that it doesn't get reset or formatted. It's simply not worth doing that.

src/client.js
xPaw: Ah you are right, it was done for `args`, I will remove it on `config`.

client/css/style.css
astorije: Fair enough, and therefore even better now then (code separation + allegedly better handling on the client)!
Thanks for specifying :-)

src/plugins/irc-events/connection.js
astorije: So in practice, the `irc.connection.socket.socket` property gets used by the `hook` method?!

xPaw: Unless I'm missing something, yes.

client/css/style.css
astorije: Meh, I'm more in favor of moving the right amount of information around instead of too much (like, I remember having lots of redundant objects, not always in sync...), but I guess it will do for now if you think not sending it can't be done in a simpler way.
**This comment comment can now be ignored.**

src/plugins/irc-events/connection.js
astorije: Urgh :-/ (for having something called `connection.socket.socket`, not saying it's not necessary, I don't know, saying however it's confusing)

Then in that case, I'd say let's call `identd.hook(irc.connection.socket.socket, network.username);` and directly use `socket` in the `hook` function. Does that seem reasonable to you?

I know it's close to splitting hairs at this stage, but I'm just trying to close the loop :-)

xPaw: Complaining about the most important bit of the code!

I'll see what is being used and fix it up.

xPaw: Looks like this was done to support node 0.12, same thing here:
https://github.com/kiwiirc/irc-framework/blob/3f22b630a0109679b0479a3be97d766a026d7cff/src/connection.js#L107-L109

astorije: > Complaining about the most important bit of the code!

I'm not complaining, I'm reviewing :-)

> Looks like this was done to support node 0.12, same thing here

Dang. Moving on then!
Maybe we should add the same comment in the `hook` function if we ever have to change that file again.

**This comment comment can now be ignored.**

src/plugins/irc-events/topic.js
```diff
@@ -39,7 +39,7 @@ module.exports = function(irc, network) {
 			type: Msg.Type.TOPIC_SET_BY,
 			mode: chan.getMode(data.nick),
 			nick: data.nick,
-			when: data.when,
+			when: new Date(data.when * 1000),
```
astorije: Is this `new Date(...)` wrapping necessary since it's [also done at parsing time](https://github.com/thelounge/lounge/pull/167/commits/78e09e8bdc189f64a3dbeb8255b775007fabfe63#diff-9cc146ed670aebf612d9daeff012d09bR3)?

maxpoulin64: @astorije Yes, because the Date object doesn't persist through server->client communication. Json doesn't have a Date type so it gets turned into a `2016-01-02T12:34:56T` format, which needs to be parsed back into a real date object.

astorije: Yep, I get that we need it in the handlebars parser, but do we need it here as well, when it gets sent?

client/js/lounge.js
```diff
@@ -146,6 +146,7 @@ $(function() {
 				})
 			);
 			var channels = $.map(data.networks, function(n) {
+				sidebar.find("#network-" + n.id).data("options", n.serverOptions);
```
astorije: If I understand this change correctly, shouldn't this be an extra `data-options="{{options}}"` attribute in [the corresponding template](https://github.com/thelounge/lounge/blob/master/client/views/network.tpl#L2) instead?

client/js/lounge.js
```diff
@@ -293,6 +297,7 @@ $(function() {
 				networks: [data.network]
 			})
 		);
+		sidebar.find("#network-" + data.network.id).data("options", data.network.serverOptions);
```
astorije: If any valid, [this comment](https://github.com/thelounge/lounge/pull/167/files#r58304321) applies here as well.

client/views/actions/channel_list.tpl
```diff
@@ -0,0 +1,18 @@
+<table style="width:100%">
```
astorije: I think we should avoid introducing inline styles like this in our templates. Can this be set in the CSS instead?

Also, how do you trigger this exactly? I tried `/list` but no luck.
Could you add a screenshot to the PR description please? :-)

client/views/actions/topic_set_by.tpl
```diff
@@ -0,0 +1 @@
+Topic set by <a href="#" class="user" data-name="{{nick}}">{{mode}}{{nick}}</a> on {{localeDate when}}
```
astorije: Can we combine `topic` and `topic_set_by` templates? It would make future changes/additions more self-contained, and also a nicer experience for the user.

Right now:
- if someone changes the topic, we get:

  ```
<nick> has changed the topic to: <topic>
```
- When joining a channel:

  ```
The topic is: <topic>
Topic set by <nick> on <date>
```

I'd see something more "uniform" in both cases:

```
Topic set[ by <nick>][ on <date>] to: <topic>
```

If it is possible to not know what `nick` set the topic, do not display `by <nick>` and, when nick is changed, do not display `on <date>` (although I'd be even in favor of not displaying the date altogether but without a strong opinion).

What do you think?

client/views/actions/whois.tpl
```diff
@@ -1,26 +1,35 @@
...
-	<a href="#" class="user" data-name="{{whois.nickname}}">{{whois.nickname}}</a>
-	is connected to {{whois.server}}
+	<a href="#" class="user" data-name="{{whois.nick}}">{{whois.nick}}</a>
+	is connected to {{whois.server}} <i>({{whois.server_info}})</i>
+</div>
+{{/if}}
+{{#if whois.secure}}
+<div>
+	<a href="#" class="user" data-name="{{whois.nick}}">{{whois.nick}}</a>
+	is using a secure connection
```
astorije: `secure` here means that the user is connected to the server through TLS?

I don't mean to be the devil's advocate here, but is it a good idea to display such information? I mean, I understand that the more information the better, and that not displaying it is merely security through obscurity (therefore, not security), but that'd very easily give away information to ill-intended users.
No strong opinion, merely wondering in terms of good practice (similarly to GitHub responding a 404 on private repositories from the outside).

dgw: I'm wondering what "ill-intended users" can hypothetically do with this information. It's potentially useful to know if the whois target is connected securely if one wishes to have a PM conversation with them (though obviously that's not really a secure conversation, because the IRC server decrypts all secured data before retransmitting).

Just having trouble coming up with reasons why hiding it would be beneficial, and why showing it could be harmful.

maxpoulin64: There's no security downside here, the user could as well just fire a normal IRC client and do the same `/whois` and get the information. Here, this is just displaying whatever the server reports. Many servers also sets an additional user mode to mark users as connected over SSL, so there really isn't anything to hide.

client/views/actions/topic_set_by.tpl
xPaw: Both topic and time are sent as separate events from the server. Introducing "waiters" for this kind of stuff is a bit of an overkill.

client/views/actions/channel_list.tpl
xPaw: We should, but this list was very quickly hacked together as a proof of concept, it will need to be improved upon either way.

src/plugins/irc-events/error.js
```diff
@@ -40,4 +40,19 @@ module.exports = function(irc, network) {
...
+		var msg = new Msg({
+			type: Msg.Type.ERROR,
+			text: "Nickname " + data.nick + " is invalid: " + data.reason,
+		});
+		client.emit("msg", {
+			chan: lobby.id,
+			msg: msg
+		});
+
+		var random = "i" + Math.random().toString(36).substr(2, 10); // 'i' so it never begins with a number
```
astorije: Could we reuse the default nick name instead of `"i"`?

xPaw: Maybe? It could potentially be an invalid nick in the config though.

client/views/actions/channel_list.tpl
astorije: > We should, but this list was very quickly hacked together as a proof of concept

Let's remove it from this PR then, please, and start it separately.

xPaw: I'm not going to remove code I've written for initial support of this feature.

client/views/actions/topic_set_by.tpl
astorije: Argh, I see. Kind of a pain that they come as separate events, but oh well, we'll see later I guess as the result is kind of odd at the moment.

client/views/actions/whois.tpl
astorije: Okay, great, thanks for these explanations!

src/models/user.js
```diff
@@ -3,6 +3,10 @@ var _ = require("lodash");
 module.exports = User;
 
 function User(attr) {
+	// TODO: Remove this
+	attr.name = attr.name || attr.nick;
+	attr.mode = attr.mode || (attr.modes && attr.modes[0]) || "";
```
astorije: What is the status of this `TODO`?

src/plugins/irc-events/error.js
```diff
@@ -2,21 +2,57 @@ var Msg = require("../../models/msg");
 
 module.exports = function(irc, network) {
 	var client = this;
-	irc.on("errors", function(data) {
+
+	// TODO: remove later
+	irc.on("irc_error", function(data) {
+		console.log("Got an irc_error");
+		irc.emit("error", data);
+	});
```
astorije: What should we do with this `TODO`?

src/plugins/irc-events/error.js
astorije: It could indeed, but we could either default another invalid nick to `"lounge-user"+number`, or assume that the configuration file is sane as we do for other defaults, such as booleans, etc.

src/plugins/irc-events/error.js
```diff
@@ -2,21 +2,57 @@ var Msg = require("../../models/msg");
...
-				irc.nick(random);
-			}
-		}
+	});
+
+	irc.on("nick in use", function(data) {
+		var lobby = network.channels[0];
+		var msg = new Msg({
+			type: Msg.Type.ERROR,
+			text: "Nickname " + data.nick + " is already in use: " + data.reason,
```
astorije: Testing on Freenode, I get:
```
Nickname astorije is already in use: Nickname is already in use.
```

Is the `reason` something defined by the network admins, by irc-framework, or a standard (*de facto* or not) in IRC?
I feel like we'll have `Nickname is already in use` most of the time, which doesn't look too nice.
What do you think about using the retrieved reason unless there isn't one? Something like:

```js
text: data.reason || "Nickname " + data.nick + " is already in use."
```

A bug I noticed though: if the nickname is already taken, `/nick` fails but the user still gets assigned a new random nickname, instead of not changing nickname.

src/models/user.js
xPaw: One of these "will be nice to do at some point" things, not really important or blocking.

src/plugins/irc-events/error.js
xPaw: Display reason alone won't show which `nick` caused it.

src/plugins/irc-events/error.js
xPaw: It's a bit unclear currently as it may be a mistake in `irc-fw` that it has 2 separate events.

src/models/user.js
astorije: Ok, thanks.

src/plugins/irc-events/error.js
astorije: **This comment has been addressed.**

Ok, for later then.

src/plugins/irc-events/error.js
astorije: We could either:
- Not care about which `nick` caused it and let the network decide if they want to or not.
- Prepend or append something:
  - `text: "Nick change to " + data.nick + "attempted: ` + data.reason` (doesn't look good if the reason is empty, if it's even possible)
  - `text: (data.reason + " (" + data.nick + ")") || ("Nickname " + data.nick + " is already in use.")` or `text: (data.reason || "Nickname is already in use.") + " (" + data.nick + ")"`
- (Since I'm listing options, I'm allowed for some overkill :D) Search for `data.nick` in `data.reason` and append the nickame is not present

That's all I can think of right now. Maybe not ideal but I think better than having twice the reason on Freenode for example.
Also, this is more intriguing, I think:

> A bug I noticed though: if the nickname is already taken, /nick fails but the user still gets assigned a new random nickname, instead of not changing nickname.

src/plugins/irc-events/invite.js
```diff
@@ -1,22 +1,20 @@
...
 			type: Msg.Type.INVITE,
-			from: data.from,
-			target: target,
-			text: data.channel,
-			invitedYou: target.toLowerCase() === irc.me.toLowerCase()
+			time: data.time,
+			from: data.nick,
+			invited: data.invited,
+			channel: data.channel,
+			invitedYou: data.invited === irc.user.nick
```
astorije: When you added `invitedYou` originally, it was because we didn't get a hold of `irc.me` (actually, `irc` as a whole) from the templates (or the client altogether, not sure exactly anymore).
Just checking, is that still the case here?

xPaw: Yes, that didn't change.

src/plugins/irc-events/mode.js
```diff
@@ -1,34 +1,53 @@
...
 			client.emit("msg", {
-				chan: chan.id,
+				chan: targetChan.id,
 				msg: msg,
 			});
 		}
+
+		if (usersUpdated) {
+			// TODO: This is horrible
+			irc.raw("NAMES", data.target);
```
astorije: I'm curious, why is this horrible? How would you do instead?

src/plugins/irc-events/notice.js
```diff
@@ -1,33 +0,0 @@
...
-			from: from,
-			text: data.message
-		});
-		chan.messages.push(msg);
-		client.emit("msg", {
-			chan: chan.id,
-			msg: msg
-		});
-	});
-};
```
astorije: Moving the notice logic from this file to the message one, was that only to be able to re-use the `handleMessage` function or there is another motive behind it?

src/plugins/irc-events/mode.js
dgw: This file is just the mode handler, right? It probably shouldn't need to invoke the whole `NAMES` handler just to update a few user modes, so this TODO probably means "replace this with direct updates to the user list when relevant mode changes happen".

That said, calling `NAMES` *is* the easy way…

src/plugins/irc-events/whois.js
```diff
@@ -20,10 +16,20 @@ module.exports = function(irc, network) {
...
-		var msg = new Msg({
-			type: Msg.Type.WHOIS,
-			whois: data
-		});
+
+		var msg;
+		if (data.error) {
+			msg = new Msg({
+				type: Msg.Type.ERROR,
+				text: "No such nick: " + data.nick
```
astorije: Is that the only type of error that can happen here, or is it worth displaying `data.error` as we do with `reason`s in some other places?

astorije:
@xPaw, tremendous :clap: :clap: on this, great work! This is really cool, some serious business is about to get in The Lounge!

After reviewing the entire code and testing a bit the branch, a few minor things:

- [x] I confirm #121 gets fixed by this, so I'm checking the checkbox in the description. \o/

- [ ] A few inline comments to address, but nothing too urgent, most of them could wait for a later PR if necessary.

- [ ] Attempting to change nick on an existing nickname is counter-intuitive at the moment (different from earlier tests): if I switch from `old_nick` to `new_nick` and it's already in use, it will be changed to `new_nick<random number>`. The command should fail and the nick should not be changed at all.

- [ ] `user_on_channel` error is not handled at the moment when inviting a user already present. That string gets sent to the network lobby and the console gets:
  ```
  Got an irc_error
error { error: 'user_on_channel',
  nick: 'astorije',
  channel: '#thelounge' }
```

- [ ] When a the current user receives an invite, the notification displays `<nick> invited you to` instead of `<nick> invited you to <channel>`

- [ ] When running `/whois <inexisting_nick>`:
  - it opens a message window with that non-existing user and an error displays there: `No such nick: <inexisting_nick>`
  - it shows this error in the network lobby: `No such nick/channel (no_such_nick)`
  - the console shows:
  ```Got an irc_error
error { error: 'no_such_nick',
  nick: 'astorije543',
  reason: 'No such nick/channel' }```

  I think the only result of `/whois <inexisting_nick>` should be an error in the lobby (i.e. nothing in console or chat window on that nick) saying `<inexisting_nick>: No such nick/channel` or something similar

- [ ] `/list` should be removed from this PR. It does lag indeed, but more importantly it's unstable: I got it to crash the second time I ran it. The output is also not perfect, but no big deal. We should not deliver it at the same time in the current POC state considering it can crash the entire app, not production-ready yet.

I'll need to do some more testing the next few days, but it's looking really great!

File: src/plugins/irc-events/notice.js (hunk 22 above)
Comment (xPaw) [inline src/plugins/irc-events/notice.js]: Yes.

**Comment by xPaw** [comment]
> Attempting to change nick on an existing nickname is counter-intuitive at the moment

I don't see how this behaviour is possibly different from what master does. Get nick in use -> append numbers.

> user_on_channel error is not handled at the moment when inviting a user already present. That string gets sent to the network lobby and the console gets:

All errors get sent into network lobby, because we never decided where to display them.

> When a the current user receives an invite, the notification displays <nick> invited you to instead of <nick> invited you to <channel>

What weird network is this on again? Works fine on freenode.

> I think the only result of /whois <inexisting_nick> should be an error in the lobby 

Again, we never decided where errors should be going.

**Comment by astorije** [comment]
> I don't see how this behaviour is possibly different from what master does. Get nick in use -> append numbers.

Nah, I just double-checked, assuming I'm `astorije`:
- on `v1.4.3`: `/nick xPaw` fails with `Nickname is already in use.` and my nick remains `astorije`.
- on `irc-framework` branch: `nick xPaw` fails with `xPaw: Nickname is already in use.` but my nick gets changed to `xPaw<n>`.

If nick is already in use, command should be canceled altogether. I can take a look if you want :-)

> All errors get sent into network lobby, because we never decided where to display them.

Sorry, should have been clearer: my point was that `user_on_channel` is directly displayed, instead of something like `<nick>: User is already on channel`.

screen shot 2016-04-10 at 00 03 39

> What weird network is this on again? Works fine on freenode.

Actually, this *is* on Freenode: A is on `v1.4.3`, B is on `irc-framework`, A invites B to a channel where B isn't (but where A is), which leads to:

screen shot 2016-04-09 at 23 54 48

Weird...

> Again, we never decided where errors should be going.

Somehow I can't reproduce this, so marking that as fixed, we'll deal with this later if needed.

File: client/js/lounge.js (hunk 11 above)
Comment (astorije) [inline client/js/lounge.js]: @xPaw, I moved this myself to the template in 9dd980d. What do you think? I can remove if you dislike.

File: src/plugins/irc-events/error.js (hunk 16 above)
Comment (astorije) [inline src/plugins/irc-events/error.js]: Testing on `v1.4.3` vs testing on `irc-framework` branch, if changing to an invalid nick:
- on `v1.4.3`, command is denied and error on lobby is `Erroneous Nickname`
- on `irc-framework`, my nick got changed to `i0qugh5endq`, very odd lol

I think `/nick` command should not be executed when nick is invalid, essentially what `v1.4.3` is currently doing. Not sure what happens / should happen when connecting to the network (as opposed to running `/nick`).

**Comment by astorije** [comment]
Alright, I'm done testing I believe :-)
I have tested a few edge cases and very few and minor bugs discovered, that's great! I have reported all of them in comments, inline or not.

Only major issue I have noticed is that running the `/list` command has triggered 2 different types of errors, depending on the network (one of them being Freenode):

```js
[...]/lounge/node_modules/irc-framework/src/commands/handlers/misc.js:12
        if (cache.channels.length) {
                          ^

TypeError: Cannot read property 'length' of undefined
    at IrcCommandHandler.handlers.RPL_LISTEND ([...]/lounge/node_modules/irc-framework/src/commands/handlers/misc.js:12:27)
    at IrcCommandHandler.dispatch ([...]/lounge/node_modules/irc-framework/src/commands/handler.js:49:37)
    at IrcCommandHandler._write ([...]/lounge/node_modules/irc-framework/src/commands/handler.js:35:10)
    at doWrite (_stream_writable.js:292:12)
    at writeOrBuffer (_stream_writable.js:278:5)
    at IrcCommandHandler.Writable.write (_stream_writable.js:207:11)
    at MiddlewareStream.ondata (_stream_readable.js:528:20)
    at emitOne (events.js:77:13)
    at MiddlewareStream.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
```

and:

```js
[...]/lounge/node_modules/irc-framework/src/commands/handlers/misc.js:23
        cache.channels.push({
                      ^

TypeError: Cannot read property 'push' of undefined
    at IrcCommandHandler.handlers.RPL_LIST ([...]/lounge/node_modules/irc-framework/src/commands/handlers/misc.js:23:23)
    at IrcCommandHandler.dispatch ([...]/lounge/node_modules/irc-framework/src/commands/handler.js:49:37)
    at IrcCommandHandler._write ([...]/lounge/node_modules/irc-framework/src/commands/handler.js:35:10)
    at doWrite (_stream_writable.js:292:12)
    at writeOrBuffer (_stream_writable.js:278:5)
    at IrcCommandHandler.Writable.write (_stream_writable.js:207:11)
    at MiddlewareStream.ondata (_stream_readable.js:528:20)
    at emitOne (events.js:77:13)
    at MiddlewareStream.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
```

It's fairly inconsistent, and for the first one I sometimes had to run `/list` 2-3 times in a row to see it happen.
Note that these happen even with the code to handle `/list` absent from the PR :-(

It seems pretty critical to me, as it gets easy to kill an instance, like our demo. @xPaw, what do you recommend here? Is it reasonable to suggest a custom input plugin for `/list` that just does nothing until the feature appears for real? If so, I'm guessing the using a raw command would still fail and maybe we'd have to hardcode ignoring `/list`... Very much not ideal, so I'm open to suggestions, even short term ones to get this PR quickly merged into `master`.

---

So yeah, a few minor bugs, that big one, and then I think we can ship this PR :-)

xPaw:
@astorije I suggest we take a good look at error messages in a future PR.

And I see about the random name now, it only added random numbers if you were not "connected" yet.

astorije:
> @astorije I suggest we take a good look at error messages in a future PR

Yep!

> And I see about the random name now, it only added random numbers if you were not "connected" yet.** Fixed**

Nice, tested and works OK! Still not too fond of the `i<something>` string but good enough for this PR.

> I don't see how you are possibly hitting a crash in /list, since the `cache.channels` array is always created in RPL_LISTSTART.

See it in action :-) On the left, it's Freenode, on the right, it's not. Each of them produces one of the 2 stacktraces I posted above:

astorije:
> @astorije: When a the current user receives an invite, the notification displays `<nick> invited you to` instead of `<nick> invited you to <channel>`
> @xPaw: What weird network is this on again? Works fine on freenode.

I'm not sure how come you couldn't reproduce this as I found this to be actually a bug. I fixed it in af2c36e, one less to go :-)

xPaw:
@astorije `/list` will be fixed in irc-fw it self, we will just need to update it. Is there any other concerns left or not? If not, I say we can merge it *(but not release)* and then update irc-fw separately.

The remaining points you have are all related to displaying error messages, which we agreed to take a look at in a future discussion.

astorije:
Great re: `/list` bug!
I'll take a pass on the thread to see if things have been left out (apart from those we said we'd tackle later).
Let's wait a few days before merging so that @YaManicKill can get back to us after a bit of testing, and after that, definitely let's merge and delay a tiny bit the following release.

YaManicKill:
I've been running it for the last few days and not noticed anything that
hasn't been brought up already. I say just go for it.

astorije:
Alright, looks like all comments have been addressed, or discussed enough to ensure they'll be addressed when time comes, everyone is :+1: on that, let's do it!!

MaxLeiter:
Awesome work everyone involved! Awesome PR!

PR #168, opened 2016-03-08
Auto complete fix for nicks

bark:
Change so if you try to auto complete nick in start of line it completet to "nick: " from before "nick" and if it is in the middle of a text  "nick" is change to "nick ". And commands ex /msg only auto completets if is in the start of the line.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -948,7 +948,14 @@ $(function() {
 	}
 
 	function complete(word) {
-		var words = commands.slice();
+
+		var firstWord=$("#input").val().lastIndexOf(word) === 0;
+		var words;
+		if (firstWord){
+			words = commands.slice();
+		} else {
+			words = [];
+		}
 		var users = chat.find(".active").find(".users");
 		var nicks = users.data("nicks");
 
@@ -962,7 +969,11 @@ $(function() {
 		}
 
 		for (var i in nicks) {
-			words.push(nicks[i]);
+			if (firstWord){
+				words.push(nicks[i]+": ");
+			} else {
+				words.push(nicks[i]+" ");
+			}
 		}
 
 		sidebar.find(".chan")


maxpoulin64:
Please squash your commits together and rebase on top of the current master. At the same time, please also try cleaning it up to only change what's needed. As it is right now, your PR would undo a lot of changes such as #57 and I have no idea what your actual changes are.

bark:
Thanks, for the feedback. does it look good now?

client/js/lounge.js
```diff
@@ -962,16 +969,19 @@ $(function() {
...
 		}
-
 		sidebar.find(".chan")
-			.each(function() {
-				var self = $(this);
-				if (!self.hasClass("lobby")) {
-					words.push(self.data("title"));
-				}
-			});
+		.each(function() {
```
maxpoulin64: Please remove this change, indentation is not related to the subject of your PR.

maxpoulin64:
Overall I am not sure I like this change. Personally I like that it only completes the nick and lets me choose whatever separator, especially if I'm about to reply to multiple people and type `xPaw, astorije:`. Due to that I think this should be made a setting instead of forcing one's preference on the others.

bark:
It is the same default as irssi, weechat and slack so I think that this should be the default.  But I'm not against to add an option for changing from the default. But I really think this should be the standard.

dgw:
Start of line `nick: ` and mid-line `nick ` is how Textual does it, as have most of the clients I've ever tried long enough to notice what they do. KVIrc is a big exception, and I always make more typos in it because it's the odd one out.

I agree that a setting would be nice, but right now the lack of anything after the nicks at all is tripping me up a lot when I'm in The Lounge.

YaManicKill:
Yeah, just to add to the conversation, when I came to shout originally, I missed auto "nick: " at the beginning of the line.

astorije:
I very much agree with @maxpoulin64 about The Lounge not being opinionated and being consistent about it. By completing to `nick`, it lets the user choose what to do, and adding a comma or a colon "manually" when starting a sentence is definitely not a life changer.
In fact, I have scribed and looked over dozens of scribes on IRC (yeah, scribing hour or day-long meetings on IRC...) for a couple years: I have seen an awful lot of `nick: `<kbd>←</kbd><kbd>←</kbd>`, ` and `nick, `<kbd>←</kbd><kbd>←</kbd>`: ` to a point where it's not as helpful as one imagines.
User testing has showed me that it's much more painful to press the back key rather than a comma or a colon based on use case and personal preference.

About making it a setting though... Frankly overkill...

Jay2k1:
I think adding a setting for it would be the best way to handle it. My main clients are mIRC and Textual, and I like how in Textual, I can set a custom suffix for the nick completion, which I've set to ': ' so it behaves like @dgw described.
image


Directly addressing someone happens way more often for me than addressing multiple people, and in mIRC I always type the ': ' manually, so I would definitely like to see this feature in The Lounge. However I understand that it is something that would not please everyone, which is why it should be a setting.

YaManicKill:
Hmm, a setting. Interesting idea. I do quite like that, and as mentioned before I miss having the autocomplete suffix. I would be for adding a setting.

astorije:
Adding a setting for this really feels like overkill... I see this as a deliberate design choice, similar to GitHub prefixing user names with `@`, Gitter dealing with <kbd>Return</kbd> differently in code mode, etc.
If it's reasonable without obstructing usability (we're talking about one or 2 chars and user habits here), I don't see why we should not stick to it.

That being said, I feel like this is yet another very minor thing that tears the world apart and everyone will strongly stand by their opinion no matter the arguments, so whatever, make it a setting if you really can't live without it. Simply make sure to not change the current default, which is very reasonable and works well.
Also, please don't make The Lounge another of these undigestible clients with gazillions settings for the sake of customizing every little thing in the world. It is fine to make decisions. After all, I feel like this is why most people came to Shout / The Lounge: ease of use and nice, modern interface.

I am closing this PR in light of what has been said in this thread. If you want to make it a setting, please **discuss** it on the IRC/channel and/or an issue before jumping into code, and open a new PR that links to this one (closing gives a way to archive discussions and code that didn't make it, as opposed to amending commits in a PR).
Thanks for your contribution @bark and sorry this didn't make it!

PR #170, opened 2016-03-09
Implement LDAP support

lindskogen:
Disable change password when ldap is enabled

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "commander": "2.9.0",
     "event-stream": "3.3.2",
     "express": "4.13.4",
+    "ldapjs": "~1.0.0",
     "lodash": "4.5.0",
     "mkdirp": "0.5.1",
     "moment": "2.11.2",
--- a/src/server.js
+++ b/src/server.js
@@ -6,9 +6,12 @@ var express = require("express");
 var fs = require("fs");
 var io = require("socket.io");
 var Helper = require("./helper");
+var ldap = require("ldapjs");
 var config = {};
 
 var manager = new ClientManager();
+var ldapclient = null;
+var authFunction = localAuth;
 
 module.exports = function(options) {
 	config = Helper.getConfig();
@@ -27,6 +30,10 @@ module.exports = function(options) {
 	var host = config.host;
 	var transports = config.transports || ["websocket", "polling"];
 
+	if (config.public && (config.ldap || {}).enable) {
+		throw "Server is public and set to use LDAP. Please disable public if trying to use LDAP authentication.";
+	}
+
 	if (!https.enable){
 		server = require("http");
 		server = server.createServer(app).listen(port, host);
@@ -40,6 +47,13 @@ module.exports = function(options) {
 
 	if ((config.identd || {}).enable) {
 		require("./identd").start(config.identd.port);
+	}
+
+	if ((config.ldap || {}).enable) {
+		ldapclient = ldap.createClient({
+			url: config.ldap.url
+		});
+		authFunction = ldapAuth;
 	}
 
 	var sockets = io(server, {
@@ -106,7 +120,7 @@ function init(socket, client, token) {
 				client.connect(data);
 			}
 		);
-		if (!config.public) {
+		if (!config.public && !config.ldap.enable) {
 			socket.on(
 				"change-password",
 				function(data) {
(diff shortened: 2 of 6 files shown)


src/server.js
```diff
@@ -189,28 +227,31 @@ function auth(data) {
...
-				if (data.token === client.token) {
-					success = true;
-				}
-			} else if (client.config.user === data.user) {
-				if (bcrypt.compareSync(data.password || "", client.config.password)) {
-					success = true;
-				}
-			}
+		client = manager.findClient(data.user, data.token);
+		var signedIn = data.token && client.token === data.token;
```
Dr-Horv: Should probably check that `client` exist before trying to access token. I think  `findClient` can return false/undefined.

xPaw:
What should happen if you have public mode and ldap enabled at the same time?

lindskogen:
If public mode is enabled the server won't even try to authenticate the user, so it won't be a problem. Maybe you're thinking more along the lines showing an error message and exiting?

xPaw:
Yeah, showing a message would probably be helpful to avoid issues.

lindskogen:
@xPaw ok, rebased again :smile:

maxpoulin64:
From a code standpoint, this looks fine to me. I would test that, but I have a bit of trouble getting around to understand that mess. It' a massive pain in the ass to set up and I can't find any tutorial that doesn't assume one is already an LDAP expert or any tutorial that explains how to set up a minimal OpenLDAP server that doesn't involve a bazillion config files, files and databases everywhere. So I'll pass on the testing of this one.

YaManicKill:
So, this would probably be a good thing to get in, but I guess none of @thelounge/maintainers are happy enough tesitng ldap stuff.

I propose that we test the PR (once @lindskogen rebases again, please :-P) and ensure there are no regressions in the other user stuff, and if not...then we merge in. Assuming people are happy with the code.

This is something we have been asked for quite a bit, so we can't just ignore it.

So either we take this option, or someone learns ldap and delves into the horribleness. I used to have an ldap server I could use, but I never set it up, and I don't have access to it anymore. I wouldn't know where to start.

YaManicKill:
Or, and here's a better solution, https://jumpcloud.com/daas-product/index

Free for 10 user hosted ldap solution.

jedahan:
jumpcloud looks awesome for testing. If someone does want to dig into self-hosted ldap testing, the easiest looking stuff I've found has been https://github.com/osixia/docker-openldap for the server and https://github.com/osixia/docker-phpLDAPadmin for admin gui

jedahan:
And the build-from-scratch possibly-addable-to-npm-test-suite option: http://ldapjs.org/examples.html

jedahan:
I ran the patches today, had to fix up some of the merge conflicts, and it works against our production ldap pretty flawlessly. Would love to see this merged in soon.

astorije:
Thanks for the update, @jedahan, that is great news indeed!

@lindskogen, could you rebase your PR on the `master` branch and fix the conflict, please? @jedahan, if you had hard time resolving the conflicts, maybe it's worth sharing what you did :-)

package.json
```diff
@@ -32,6 +32,7 @@
     "commander": "2.9.0",
     "event-stream": "3.3.2",
     "express": "4.13.4",
+    "ldapjs": "~1.0.0",
```
astorije: Could you use `"1.0.0"` instead, please?

jedahan:
The merge conflicts were minimal, though I think its a tiny bit buggy, I can share my changes tomorrow at work. It was blindly copy-pasting until it worked.

lindskogen:
I actually tried resolving the conflicts last week but got stuck on webIRC stuff which I know nothing about 😩

astorije:
@maxpoulin64, do you think you could help @lindskogen resolving conflicts around WEBIRC? Maybe you guys could sync on the IRC channel (`#thelounge`)?

maxpoulin64:
@astorije, @lindskogen: They don't actually conflict. It's just that things have moved a lot: I just tried fixing it and the solution is basically to just manually apply the patch instead of trying to resolve merge conflicts. You don't need to touch the WebIRC stuff, just leave it there and make sure it still runs (it applies to both private and public modes).

I can do it if you still have trouble with it.

thisisdarshan:
Created PR#477 that resolves all the conflict.

lindskogen:
Closing this in favor of #477

PR #171, opened 2016-03-09
Update eslint to 2.3.0 and add stricter rules

xPaw:
These stricter rules require more spacing, for example here: `var a="b";` should be `var a = "b";`


astorije:
:+1: and merging.

PR #172, opened 2016-03-09
Don't send an email on a successful build

xPaw:


.travis.yml
```diff
@@ -3,6 +3,10 @@ node_js:
 - '5.6'
 - '4.3'
 - '0.12'
+notifications:
+  email:
+    on_success: never
```
YaManicKill: Will this turn off "fixed" emails?

xPaw: Yes. Do we really need them?

YaManicKill: Good question, I guess most of the time you'll be watching travis after you have a failed build?

lindskogen: I prefer the `change` option, but I guess I'm not the receiver so.. :grin:

xPaw: @YaManicKill Yeah, we would see the build status in PR anyway.

astorije: I would prefer `change` as well, but as long as broken builds trigger an email, I'm OK.

xPaw: It defaults to `change`, and that triggers an email whenever a branch is created, which is not desirable, imo.

astorije: Uh, wasn't aware of that... Thanks!

PR #174, opened 2016-03-09
Add support for HTTP2

### LINKED ISSUE #173: Support HTTP2
We already have SSL support, just need to hook up http2.

xPaw:
> With this module you can create HTTP2 / SPDY servers in node.js with natural http module interface and fallback to regular https (for browsers that don't support neither HTTP2, nor SPDY yet).

`http2` module does not work with Express. Closes #173.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "read": "1.0.7",
     "request": "2.69.0",
     "slate-irc": "0.8.1",
-    "socket.io": "1.4.5"
+    "socket.io": "1.4.5",
+    "spdy": "3.2.3"
   },
   "devDependencies": {
     "stylelint": "4.3.3",
--- a/src/server.js
+++ b/src/server.js
@@ -31,7 +31,7 @@ module.exports = function(options) {
 		server = require("http");
 		server = server.createServer(app).listen(port, host);
 	} else {
-		server = require("https");
+		server = require("spdy");
 		server = server.createServer({
 			key: fs.readFileSync(https.key),
 			cert: fs.readFileSync(https.certificate)


astorije:
FYI, Google has deprecated SPDY and will completely drop it this year. I understand this package is not affiliated to Google, but it might be worth looking at consequences a bit more.

xPaw:
Yeah, but from the look of things, the package isn't going anywhere and supports HTTP2 just fine.

astorije:
@xPaw, I'll try to take a look at this over the week-end.

maxpoulin64:
Seems to be working just fine, :+1: 

Note: needs to be served over SSL for Chrome to use HTTP/2. Took me a short while before I got it working because of that, so HTTP/1.1 also works just fine still.

xPaw:
@maxpoulin64 Correct, HTTP2 is TLS only by design (of the protocol).

astorije:
FYI, I have been reading [this article](https://medium.com/@imjacobclark/http-2-with-node-js-85da17322812#.95hkmh7ac) which covers HTTP/2 with this package, good read if someone wants to know more about it.

> Yeah, but from the look of things, the package isn't going anywhere and supports HTTP2 just fine.

Indeed:

> Whilst spdy is being deprecated, node-spdy has a fantastic API and full support for HTTP/2 and fallback to HTTPS, so there is currently no reason not to use it in production.

I noted the following too, that I could read in different places:

> It’s worth noting, Express 5.0 will eventually have HTTP/2 support out the box, although there is uncertainty within the community as to when Express 5.0 will be released [...]

So we're not there yet, but someday we'll be able to simply rely on Express to deal with it on its own, pretty cool!

> @maxpoulin64 Correct, HTTP2 is TLS only by design (of the protocol).

Not quite, they ended up adding non-HTTPS support to HTTP/2 in the end, to please folks who were against having an HTTPS-only protocol. HTTPS is however the default, and apparently it's a pain to set up non-HTTPS for HTTP/2, and lots of clients would not support it anyway...

So in a nutshell, very nice to have The Lounge serving HTTP/2 pages by default when you enable HTTPS!
:+1: and merging.

PR #182, opened 2016-03-11
Switch the Send button to a paper plane icon

astorije:
This is mostly to add consistency between the Send button and the other icon buttons (Settings, user list, ...) of the app.
It has the other benefit to add a tiny bit of estate for the message input, which is nice on mobile.
It's more discreet on desktops but not used very often on them, more useful on mobile, where lots of apps are doing the same so we won't confuse users.

**Before:**
screen shot 2016-03-11 at 01 48 45

**After:**
screen shot 2016-03-11 at 01 53 50

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1161,7 +1161,6 @@ button,
 	font: 12px Consolas, Menlo, Monaco, "Lucida Console", "DejaVu Sans Mono", "Courier New", monospace;
 	left: 0;
 	height: 34px;
-	margin-right: 64px;
 	position: relative;
 }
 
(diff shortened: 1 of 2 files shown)


MaxLeiter:
I personally prefer the "send" button, but I definitely support this if the icon is only on mobile. Also, I think it should be outside the text box, similar to the send button

astorije:
I don't see why it matters on desktops, where the button is frankly rarely used anyway.
Re: inside/outside, I tried different ways and this is the one that looks the best, by far.

dgw:
For mobile, yes, but inside the input area looks good.

Why not hide the button entirely on desktop, if it's extremely rarely used? e.g. Gmail chat has no send button at all.

astorije:
@MaxLeiter, also, keep in mind that the current "Send" button only matches another type of buttons (the "Disconnect/Leave/Close" one). We also have the green buttons, and most icon buttons slightly differ. What this PR attempts to bring is consistency with other small action button :-)

@dgw, re: disabling entirely, I'm not sure. I noticed Slack doesn't have one on desktop app either, Telegram has one, Skype don't, ... But for practical reason, I'm not too comfortable assuming displaying or hiding it based on screen size: I feel there will be some users who cannot send messages on tablets or else because the button does not appear and they have no alternative.

In short, I'd keep this simple (no discrimination against device type to change behavior) and we can improve that in a later PR. What do you think? :-)

maxpoulin64:
I like it, but I think it should look a bit more like a button because of the input bar style. It really feels like it's part of the input field. It feels "weird". Apps that have the send button be just an icon usually have a totally flat message bar:

screenshot_2016-03-11-02-07-17

The others have the button visibly outside of the input box:

screenshot_2016-03-11-02-04-20
screenshot_2016-03-11-02-04-28
screenshot_2016-03-11-02-05-00

This is clearly nitpicking, so I won't hold the PR for that and give a :+1: anyway :)

PR #183, opened 2016-03-11
Remove moment.js from the client

xPaw:
This drops moment.js from the frontend, as it was merely used to format `HH:mm`. This PR however does not remove `moment` from the server, as it uses a string formatter and a time zone offset, which requires some effort to re-implement.

Sending full date timestamp in messages will allow use to support `server-time` in #167.


maxpoulin64:
:+1:  and merging, yay one less library.

PR #184, opened 2016-03-11
Change close button behaviour

### LINKED ISSUE #61: Completely remove the hover-appearing close button in channel list
Discussion continued from #48:

> Just a thought, what if he got rid of that close button completely? It would avoid accidentally clicking it while trying to focus the channel. There's already a dedicated Leave button when channel is in focus, and #9 adds a way to close channel in a context menu.

xPaw:
Closes #61, implements what was suggested in https://github.com/thelounge/lounge/issues/61#issuecomment-195208446

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -189,7 +189,8 @@ button,
 }
 
 #viewport .lt,
-#viewport .rt {
+#viewport .rt,
+#chat button.menu {
 	color: #ccc;
 	display: none;
 	float: left;
@@ -213,10 +214,23 @@ button,
 	-moz-osx-font-smoothing: grayscale;
 }
 
+#chat button.menu:before {
+	font: 14px FontAwesome;
+	content: "\f142";
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
+}
+
 #viewport .rt {
 	display: block;
 	float: right;
-	margin: 6px -12px 0 12px;
+	margin: 6px -12px 0 0;
+}
+
+#chat button.menu {
+	display: block;
+	float: right;
+	margin: 6px -8px 0 12px;
 }
 
 #viewport.rt #chat .sidebar {
@@ -1574,10 +1588,6 @@ button,
 	#viewport .lt,
 	#viewport .channel .rt {
 		display: block;
-	}
-
-	#viewport .rt {
-		margin-left: 18px;
 	}
 
 	#windows .window .header {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -603,29 +603,6 @@ button,
 	margin-left: 8px;
 }
 
-#windows .header .right {
-	float: right;
-	margin-right: -10px;
-	margin-top: -2px;
-}
-
-#windows .header .button {
-	background: #f4f4f4;
-	background-image: linear-gradient(#f4f4f4, #ececec);
-	border: 1px solid #d7d7d7;
-	border-bottom-color: #b7b7b7;
-	border-radius: 2px;
-	color: #555;
-	font: 12px Lato, sans-serif;
-	height: 25px;
-	line-height: 0;
-	padding: 0 10px;
-}
-
-#windows .header .button:hover {
-	opacity: 1;
-}
-
 #windows .window .header {
 	display: none;
 }
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -334,21 +334,6 @@ button,
 	content: "\f0f6";
 }
 
-#sidebar .chan:hover .badge {
-	opacity: 0;
-	position: absolute;
-	z-index: 1;
-	right: 10px;
-}
-
-#sidebar .chan:hover .close {
-	opacity: .2;
-}
-
-#sidebar .chan:hover .name {
-	margin-right: 16px;
-}
-
 #sidebar .chan .name {
 	position: relative;
 	z-index: 0;
@@ -378,7 +363,7 @@ button,
 	margin-left: 5px;
 	padding: 3px 6px;
 	float: right;
-	transition: opacity .2s;
+	transition: opacity .2s, background-color .2s, color .2s;
 }
 
 #sidebar .badge.highlight {
@@ -386,6 +371,7 @@ button,
 	color: #49505a;
 }
 
+#sidebar .chan.active .badge,
 #sidebar .badge:empty {
 	opacity: 0;
 }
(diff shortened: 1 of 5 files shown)


astorije:
Hey @xPaw, trying this out, it appears that closing the non-active channel by accident is still possible, as its opacity is 0 but its bounding box is still there:

close_btn

xPaw:
@astorije fixed.

client/css/style.css
```diff
@@ -411,9 +398,14 @@ button,
...
 
-#sidebar .close:hover {
+#sidebar .chan.active .close {
+	visibility: visible;
+	opacity: .4;
+}
+
+#sidebar .chan.active .close:hover {
 	background-color: rgba(0, 0, 0, .1);
-	opacity: .7 !important;
```
astorije: One less `!important`, yay!

astorije:
Very cool @xPaw, :+1:!

maxpoulin64:
Hmm, I'm not sure I like the removal of the dedicated Leave button in the actual chat window. It's not a huge deal, but I used that button when being done with PMs or leaving channels. Reason being, it's always at the same place and it's always obvious what window it will close. Do we have a specific reason why we want to remove that one?

dgw:
That's the only way I close PMs on mobile, so if it is removed I would vote for hiding it on desktop only.

maxpoulin64:
@dgw Ouch, that is indeed a very fair point there. I'm definitely against the removal of the button then. Actually I don't like having the close buttons in the sidebar at all on mobile, *those* are really, really easy to fatfinger and mistap. Way more than the hidden X on desktop.

astorije:
> Do we have a specific reason why we want to remove that one?

Yes, please read https://github.com/thelounge/lounge/issues/61#issuecomment-195208446to have some details about it.
In a nutshell, it is removing an inconsistent element of the interface: none of the buttons left in the UI have that design. In terms of UX, @xPaw's solution brings together the actions of opening or closing a channel, it is not at the other end of the app like it is currently. This is also in the same area than where the context menu lives, so here again, consistency.
I could go on but simply put, these buttons need to go :-)

> That's the only way I close PMs on mobile, so if it is removed I would vote for hiding it on desktop only.

Not anymore: with the next release will come the context menu on right/long click, and this PR also displays the <kbd>&times;</kbd> all the time on the active channel. So essentially, this is just moving an action, and will only require a bit of getting used to.

> Actually I don't like having the close buttons in the sidebar at all on mobile, those are really, really easy to fatfinger and mistap. Way more than the hidden X on desktop.

There is no more chance of fatfingering this than any other button icon on this app or any other. And actually, very little chance of fatfingering this as the button only appears in the active channel.

I know these minor details always bring up the most controversy, but after the long debate on #61 and exploring different ideas, I really hoped we'd have reached consensus here, considering this solution is in practice the most efficient one (and most arguments against it come down to habits or personal preferences). If not then we'll always go for a lesser solution that never completely works.

dgw:
> with the next release will come the context menu on right/long click, and this PR also displays the `×` all the time on the active channel. So essentially, this is just moving an action, and will only require a bit of getting used to.

It will also require an extra tap on the hamburger menu to bring up the buffer list, which will contain the `×` on the active buffer. I'm sure we'll get used to it, but it *is* an extra tap.

astorije:
> It will also require an extra tap on the hamburger menu to bring up the buffer list, which will contain the × on the active buffer. I'm sure we'll get used to it, but it is an extra tap. Using the long-press menu will also require two taps (one long on the buffer's entry in the list, one short on the menu item)—on top of opening the sidebar by tapping the hamburger.

2 clicks to close a channel, which, let's be honest, is not something people do every day. I really don't see the overkill.

But whatever. Way too much bikeshedding here, I'd rather do something meaningful. It's pretty clear to me that there will be no consensus on the subject, for something as small and insignificant than an action that gets triggered on rare occasions.
Do me a favor and at least get rid of the ugly design for this button (talking about the design, not the button per se), to match the rest of the application, and I'll close my eyes to pretend I'm happy with whatever solution.

xPaw:
> It will also require an extra tap on the hamburger menu to bring up the buffer list, which will contain the × on the active buffer. I'm sure we'll get used to it, but it is an extra tap

Good. Just today I accidentally pressed that huge `Leave` button while on mobile. It really shouldn't be stuck there at all times.

astorije:
FYI, considering that adding the context menu and changing design of the Send button are changes introduced in the next release, the next release should really address this.

My take on this would be to merge as is. It is very possible that another context menu, more visible (for example, through a button where the current Leave/Close button lives), duplicating entries of the right click, will appear in the future. In such case, we can revamp the Leave/Close button, making it a 2-click action (yes it's a good thing).

Keep in mind that closing a chat today deletes all scrollback from the app, it's very destructive. We should make it hard to close a channel at the moment, until we provide scrollback persistence and/or more actions.

My definitive :+1: for this PR, and things aren't set in stone, we evolve quickly :-)

maxpoulin64:
Considering my only concern left with this PR is that without the Leave button, it's unclear how to close channels on mobile without opening the sidebar twice, I went ahead and added a button that adds a shortcut to the context menu.

Can we add [this patch](https://github.com/maxpoulin64/lounge/commit/402e7cd638802f9f4f05ea78cf3acdd9dc7284c4) to the PR? With that I think we'll have addressed pretty much the concerns of everyone.

astorije:
I'd be OK with that. What do you think, @xPaw?

xPaw:
Fine with me, @maxpoulin64 feel free to commit your patch into this PR.

xPaw:
I patched in @maxpoulin64 changes.

maxpoulin64:
@xPaw welp beat me to it. I'm finally giving my :+1: to this! :clap:

astorije:
Alright, a couple of thoughts:

- Now that we have @maxpoulin64's menu, is keeping the <kbd>&times;</kbd> button in the sidebar relevant? It means we now have 3 ways to close a chat: 2 in the active one (<kbd>&times;</kbd> and dropdown menu) and 1 in inactive channels (right-click menu). In comparison, we have only one way to open a new chat from the UI: clicking on a channel someone mentioned. I know I'm backtracking a bit here, but things evolve, and I don't think it's necessary to have the sidebar element anymore. I'll leave it up to you.
- The dropdown menu gets positioned depending on where the button was clicked:
  position-dropdown

  Is it easily possible to have it opened always at the same position no matter where the button gets clicked? If not easily doable, perfectly fine if it goes in a further PR.

Apart from that, I think the visual design of context menu can be improved, but definitely not in this PR.

Great job and discussions, guys!!

xPaw:
@astorije Positioning fixed.

astorije:
Everything works great, thanks all for this work. Merging now.

PR #187, opened 2016-03-13
Fix Send button style on Zenburn and Morning themes

astorije:
@xPaw pointed out that I forgot to update themes in #182. Crypto doesn't need fixing.

**Zenburn:**

screen shot 2016-03-12 at 19 30 06

**Morning:**

screen shot 2016-03-12 at 19 32 32


maxpoulin64:
:+1:

PR #190, opened 2016-03-13
Keep track of highlights when user is offline

xPaw:
If user is offline and channel gets a lot of messages, when the user comes back online, the unread counter badge won't be white even if the user was mentioned in that channel, that's because badge only turns white if any message in current buffer contains a highlight.


astorije:
> that's because badge only turns white if any message in current buffer contains a highlight.

What do you mean by that?

I'm not too thrilled by fixing this on the server when I think it would be nicer to have that on the front-end. What prevented you to do so, @xPaw?

xPaw:
@astorije Same reason unread counter is server side. If you get highlighted and then the channel receives 500 messages, when you open the client, that message won't be loaded, thus you don't know if you were highlighted.

maxpoulin64:
:+1: and merging

PR #191, opened 2016-03-14
Load input plugins at startup and call them directly when a command is received

astorije:
This extends #154 a bit more and is a first step into improving things (next step will add tests!) :-)

@xPaw, I know you have looked at this code recently, let me know if you can think of how I can make this better (the existing changes, as I'm leaving out other things that will need changing later on).

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -48,7 +48,14 @@ var inputs = [
 	"quit",
 	"raw",
 	"topic",
-];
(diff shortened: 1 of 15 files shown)


xPaw:
Seems fine to me :+1:

maxpoulin64:
Seems fine to me as well. :+1: and merging.

PR #193, opened 2016-03-14
Remove unnecessary operation when sorting users

astorije:
Maybe there was a good reason for this `.reverse()` but I can't admit I'm seeing it right now...
The entire method could probably use a rewrite but this is the least that can be done here.

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -27,14 +27,8 @@ Chan.prototype.sortUsers = function() {
 		this.users,
 		function(u) { return u.name.toLowerCase(); }
 	);
-	var modes = [
-		"~",
-		"&",
-		"@",
-		"%",
-		"+",
-	].reverse();
-	modes.forEach(function(mode) {
+
+	["+", "%", "@", "&", "~"].forEach(function(mode) {
 		this.users = _.remove(
 			this.users,
 			function(u) { return u.mode === mode; }


xPaw:
I wouldn't merge this because #167 actually knows which user modes are used on the server, and that can be used instead of this hardcoded array.

xPaw:
Okay, I did it in 53c367b58bcdf5068390ba2f6d50d7e278f9b95f, and now it sorts the whole user list within a single pass. Closing this PR.

astorije:
Reopening this: even though it's already done in #167, it is still a few weeks down the road (that kind of change requires intensive testing and is not ready yet, not to mention reviewers' time allocation) while this PR is already there and done.
If others think it's unnecessary so be it, but I think this is a minor improvement that can easily go with the next patch release.

YaManicKill:
It's obviously not a massive improvement (although it is an improvement none-the-less) so I wouldn't have spent the time looking for it. However, as you have already done the work, I see no reason to deny it, especially as #167 won't be done anytime soon (we are at the very least waiting on 2 bugs to be fixed with the framework itself).

So I'll :+1: this but stick second reviewer on it, and @maxpoulin64 or @xPaw can fight it out :-P

maxpoulin64:
I guess I'll side with @YaManicKill on this one, since it is a really trivial fix and :+1:. It's kind of useless even for me as this really only moves 4 values, but I don't see any particular reason to deny it. I'm looking forward to @xPaw 's actual proper implementation of course.

PR #197, opened 2016-03-15
Add tests for the sortUsers method

astorije:
Behold... tests!

~~I'm usually not too fond of having placeholder code until something happens, but considering @xPaw is making good progress by himself on #167, I thought I would give a minor hand by providing the lines to change to have these tests running in his PR.
And as a matter of fact... these tests are currently failing on #167 :D~~
Nevermind, removed the placeholders, will re-add when time comes.

I hope you like 'em, let me tell you this is the beginning of a whole lot of PRs about testing I'm planning to submit :-)

(Yes, I'm adding [Chai](http://chaijs.com/) in the `devDependencies`, but it's really worth it!)

### DIFF
--- a/test/models/chan.js
+++ b/test/models/chan.js
@@ -1,0 +1,80 @@
+"use strict";
+
+var expect = require("chai").expect;
+
+var Chan = require("../../src/models/chan");
+var User = require("../../src/models/user");
+
+function makeUser(name, mode) {
+	return new User({name: name, mode: mode});
+}
+
+function getUserNames(chan) {
+	return chan.users.map(function(u) { return u.name; });
+}
+
+describe("Chan", function() {
+	describe("#sortUsers()", function() {
+
+		it("should sort a simple user list", function() {
+			var chan = new Chan({users: [
+				"JocelynD", "YaManicKill", "astorije", "xPaw", "Max-P"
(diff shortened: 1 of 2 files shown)


test/models/chan.js
```diff
@@ -0,0 +1,70 @@
+"use strict";
+
+var expect = require("chai").expect;
+
+var Chan = require("../../src/models/chan");
+
+describe("Chan", function() {
+	describe("#sortUsers()", function() {
+	// TODO Use this version instead when switching to irc-framework
```
xPaw: Remove these TODOs, it's unnecessary. After this PR is merged, just update the tests in ir-framework branch.

xPaw:
Add a couple of users with different cases in the name, and a couple of special symbols to test case insensitive sorting too.

test/models/chan.js
astorije: Removed in 1715059.

astorije:
@xPaw, good suggestions! Added in a6b452d. I hope it can go now :-)

astorije:
Also, should test file name match tested file name (like currently)? Or `something.js` and `somethingSuite.js`? Or... ?
I don't have a strong opinion on this, as long as it ships :-)

PR #198, opened 2016-03-16
Add a very basic test for Network#export

astorije:
This is a very basic, naive test but my plan was to have the minimal possible coverage rather than an extensive test suite. Will improve over time.

Btw, CI will pass after #197 is merged and this PR is rebased off master.

### DIFF
--- a/test/models/network.js
+++ b/test/models/network.js
@@ -1,0 +1,30 @@
+"use strict";
+
+var expect = require("chai").expect;
+
+var Chan = require("../../src/models/chan");
+var Network = require("../../src/models/network");
+
+describe("Network", function() {
+	describe("#export()", function() {
+
+		it("should produce an valid object", function() {
+			var network = new Network({name: "networkName"});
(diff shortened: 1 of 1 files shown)


test/models/network.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe("Network", function() {
+	describe("#export()", function() {
+
+		it("should produce an valid object", function() {
+			var network = new Network({name: "networkName"});
+			network.channels.push(new Chan({name: "#thelounge"}));
+			network.channels.push(new Chan({name: "&foobar"}));
+
+			expect(network.export()).to.deep.equal({
```
xPaw: Will this test fail if we add new properties to export?

astorije: Yes, on purpose. If changing the behavior or `.export()`, it must be a conscious choice and therefore the test(s) should be updated :-)

xPaw: Ah that's fine, the export function picks what entries to export. Confused it with `toJSON`.

PR #203, opened 2016-03-17
Link to lounge demo from irc badge

Henni:
Just a small change for the README


MaxLeiter:
Might be bias, but :+1: from me. Not everyone looks at the website.

YaManicKill:
Yeah, I don't see why not, the demo puts people into #thelounge doesn't it? 2 birds, 1 stone.

maxpoulin64:
Yeah sounds good to me too, :+1:

astorije:
One of the reasons why we couldn't do that on https://github.com/erming/shout was that the demo is connecting to a test IRC server and locked to it. We'll probably want to do that as we attract more users, but in the meantime, this is good enough: :+1:.

@MaxLeiter, do you mind locking the demo to Freenode, then?

MaxLeiter:
It already is, @astorije (thanks to @xPaw for implementing locking networks)

astorije:
My bad, I didn't realize it was the case on the demo!

PR #205, opened 2016-03-19
Restrict access to the home directory by default

maxpoulin64:
As discussed in #165, this PR restricts the permissions on the entire home directory, making it secure by default without being instrusive for sysadmins.

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -23,7 +23,7 @@ if (program.home) {
 
 var config = Helper.HOME + "/config.js";
 if (!fs.existsSync(config)) {
-	mkdirp.sync(Helper.HOME);
+	mkdirp.sync(Helper.HOME, {mode: "0700"});
 	fs.writeFileSync(
 		config,
 		fs.readFileSync(__dirname + "/../../defaults/config.js")


src/command-line/index.js
```diff
@@ -23,7 +23,7 @@ if (program.home) {
 
 var config = Helper.HOME + "/config.js";
 if (!fs.existsSync(config)) {
-	mkdirp.sync(Helper.HOME);
+	mkdirp.sync(Helper.HOME, {mode: 0o0700});
```
astorije: Could you use `'0700'` instead please?
ESLint is not too happy with this and octal mode is not allowed in strict mode (not that we enforce it at the moment, but it would be nice as it's a simple addition and helps finding out issues...).

astorije:
@JocelynDelalande, would be great if you could take this one!

astorije:
(@maxpoulin64, I'll take ownership because @JocelynDelalande is unavailable at the moment. @JocelynDelalande, feel free to take it back if you happen to be back before it gets merged.)

Just wondering about the following:
- Why `0700` and not `0770`, or `0750`? Why restricting all group permissions?
- Same question for the initial `0`, what if they want to set the sticky bit or else?

It might be that I am ignorant, and/or that this just applies when the folder gets created the first time, but since you specify "without being instrusive for sysadmins, I just want to make sure we're on the same page :-)
Thanks!

JocelynDelalande:
First, sorry for not having time to discuss that quickly. That's life :)

> As discussed in #165, this PR restricts the permissions on the entire home directory, making it secure by default without being instrusive for sysadmins.

IMHO, it's *more* intrusive than  #165 : by default it shadows *all* files from *every* user, even for non-secret settings (I'm thinking specially about *config.js*).

In that, this proposal is quite exotic (look at /etc/fstab, nginx config files, or… any software having shared/config files).

I don't say doing as others is always the best, but there is a reason : leaving by default access to non-sensitive information to other unix user allows them to know what are the settings of the software they may be using . It also prevents getting the admin/sudo rights when not strictly necessary (eg: reading the config files). Moreover, . So for now it's a :-1: for me.

It doesn't prevent from hardening if an adminsys wants to. I'm just suggesting that our default rights policy matches with other software default practices, for a consistent adminsys/user experience.

So I still think #165 amended with @maxpoulin64 suggestion, is less intrusive and more "right" than this one. I resume discussion on #165.

xPaw:
@JocelynDelalande What if stricter mode was set on the users folder? https://github.com/maxpoulin64/lounge/blob/I-194/src/command-line/add.js#L14

I'm of opinion that we don't really need to be messing with modes at all, but if we are going to, doing it on users folder should be a good compromise.

astorije:
> IMHO, it's more intrusive than #165 : by default it shadows all files from every user, even for non-secret settings (I'm thinking specially about config.js).

I don't think it is as it's only set when the folder gets created. But when it gets applied, indeed everything goes in the dark.

> @JocelynDelalande: In that, this proposal is quite exotic (look at /etc/fstab, nginx config files, or… any software having shared/config files).
> @xPaw: I'm of opinion that we don't really need to be messing with modes at all, but if we are going to, doing it on users folder should be a good compromise.

Software you mention usually leave modes alone and therefore rely on the `umask`, unless it is critical that a file/folder must be kept secret.

So in short I agree with @xPaw here: let's leave modes alone for the configuration file, and harden them on the `users` folder **only when it gets created**. In addition to that we can add a note in the release telling people who upgrade that they can harden their `users` folder by setting the same mode than the one we start using.

All that being said, my concerns above still apply, mainly: why setting `0` to the group, and why setting `0` to the mode prefix (the one where one sets the sitcky bit, `setgid`, ...)?
Just being curious, to make sure I understand the intent :-)

xPaw:
@astorije 700 is not a bad default for users folder. I was looking at znc, and they default creating folders to that too.

astorije:
@xPaw, fair enough, and admins can set specific group permissions directly if needed (but then `700` instead of `0700` seems more appropriate to me).

xPaw:
I'll :+1: this. We store passwords in the configuration files, and the config folder is inside of the user folder.

On top of that, this change only affects **creation** of the folder. If a user wishes to set wider permissions on the folder, they absolutely can.

maxpoulin64:
So, can we finally merge this?

MaxLeiter:
Bump?

astorije:
Yes. After all, this is very innocent, and we haven't heard from @JocelynDelalande in more than a month.
@JocelynDelalande, if you still have concerns, please open an issue or a new PR or continue the discussion here, but in the meantime, merging this. 👍

PR #208, opened 2016-03-20
Remove join, nick and whois inputs, they are cleanly handled by the server

### LINKED ISSUE #201: Remove some of the unnecessary command handlers
Now that we automatically send unknown commands to the server, we can get rid of some the command handlers we have. They're simply re-sent to the server, there's no need to try and parse them.

Commands that can be removed:
- `/join`
- `/nick`
- `/whois`

`/quote` is also a slightly odd one, but I guess we need to keep it since the server won't understand `quote`.

xPaw:
Fixes #201


maxpoulin64:
:+1:

client/js/lounge.js
```diff
@@ -17,7 +17,6 @@ $(function() {
 		"/notice",
 		"/op",
 		"/part",
-		"/query",
```
astorije: Just wondering, why removing this?

astorije:
I'll test tomorrow and will merge then.

client/js/lounge.js
xPaw: That was an alias for `/whois`, and since that's gone, that wouldn't work.

dgw: So it was (see `src/plugins/inputs/whois.js`, below). I'm guessing @astorije assumed, as I did at first glance, that `/query` was an alias for `/msg`. That is the common usage among other clients.

YaManicKill: Well not exactly an alias for /msg, as some clients don't open a query window for /msg, and just show the message sent/received in the server window and only show a query window when you do /query.

But certanarily, I don't like that, so it's essentially the same thing. The important thing is that it opens a query window, but does /query still do a whois by default, or do you have to do /whois for that?

dgw: I can't recall using any client that auto-whois'd on `/query`, no. But I also don't recall using any that didn't open a query window for `/msg`, so maybe I have limited experience.

astorije: Thanks @xPaw, my bad for forgetting that `/query` was merely an alias for `/whois`...

astorije:
Tested locally and all of the removed commands (well, apart from `/query`) still work.
:+1: and merging.

When releasing this, I will specify in the release introduction that the `/query` command was removed.

My understanding is that usually, people expect that:
- `/whois` displays the whois data of the user. In our case, opening a message window that begins with that info.
- `/msg` sends a message to another user and opens a chat window with displaying that message.
- `/query` simply opens a chat window with a user.

Am I correct?

We are definitely losing the last option with this PR. Would it be possible to re-add `/query` so that it only opens an empty chat? Something custom like the `/clear` command The Lounge has for example. Ideally early enough so we can add it to the release that ships this PR.

Also, @dgw rightly mentioned that `/msg` doesn't open a chat window. It's another problem indeed, but was here before this PR.

PR #210, opened 2016-03-20
Fix #195: Support port in connect command

### LINKED ISSUE #195: Support port in /connect command
That means passing in `args[1]` and also splitting if the input contains `:`

xPaw:


maxpoulin64:
:+1: and merging

astorije:
Should we update the doc at https://thelounge.github.io/docs/client/commands.html#connect?

PR #212, opened 2016-03-20
Minor enhancements of the context menu UI

astorije:
- Blur the shadow around the context menu box
- Add a 2px radius to the context menu box
- Add a divider to allow for margin around it, and prepare future item additions
- Add margins at the top and bottom of the context menu, as well as around the divider, to replicate common context menu behavior
- Use a white background <kbd>X</kbd> Font Awesome icon to close a chat (not <kbd>&times;</kbd> `&times` on purpose because it doesn't look as nice on white background, but we can improve if necessary)
- Smooth icons of context menu items

Ideally, I would like to find a nice contrasted color for the `:hover`, allowing us to invert text/icon colors when highlighting an item, but I couldn't find a nice one. Any suggestion? 

Before | After
--- | ---
screen shot 2016-03-20 at 17 28 09 | screen shot 2016-03-20 at 17 25 49
screen shot 2016-03-20 at 17 31 04 | screen shot 2016-03-20 at 17 25 58

### DIFF
--- a/client/views/contextmenu_divider.tpl
+++ b/client/views/contextmenu_divider.tpl
@@ -1,0 +1,1 @@
+<li class="context-menu-divider" />
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1244,6 +1244,8 @@ button,
 	font-family: FontAwesome;
 	width: 20px;
 	display: inline-block;
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
 }
 
 .context-menu-user:before {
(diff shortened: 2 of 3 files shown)


xPaw:
:+1:

client/views/contextmenu_divider.tpl
```diff
@@ -0,0 +1 @@
+<li class="context-menu-divider" />
```
xPaw: I would explicitly write this as `<li class="context-menu-divider"></li>`

PR #215, opened 2016-03-22
Fix message and topic text wrapping

xPaw:
This makes topics not disappear due to wonky wrapping, and this also removes the need of having a special `.wrap` class for long words in chat.

See http://stackoverflow.com/a/1883702/2200891


maxpoulin64:
Yeah I never understood why this was there in the first place. :+1:

astorije:
:+1:

PR #217, opened 2016-03-23
Keep autocompletion sort whenever user list updates

### LINKED ISSUE #92: Whenever a user joins/parts a channel, autocompletion sorting is destroyed
There are two solutions to this:

1. If we want to autocomplete on users that just joined/quit:
 - It needs to simply push newly added users into `nicks`, but instead, it overwrites the whole array.
2. If we don't want to push users that quit/joined into autocompletion:
 - Simply don't overwrite `users` data.

I'm more inclined towards the second option.

xPaw:
~~This solution is not ideal for two reasons:
1. Users that leave *(or change name)* are left in auto completion
2. Users that change name in case sensitive manner are not updated~~

~~I didn't fix these two because it would get rather expensive to do a lot of manipulation on the arrays. Perhaps we should take names list from the server and merely sort it based on the old users array client has, probably would need to invert the users array so that nicks become order.~~

Fixes #92

This PR keeps previous order whenever user list is updated, newly joined users are pushed to the bottom, quit users are removed from the list.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -340,7 +340,7 @@ Client.prototype.names = function(data) {
 	}
 
 	client.emit("names", {
-		chan: target.chan.id,
+		id: target.chan.id,
 		users: target.chan.users
 	});
 };
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -153,7 +153,7 @@ $(function() {
 					channels: channels
 				})
 			);
-			channels.forEach(renderChannelMessages);
+			channels.forEach(renderChannel);
 			confirmExit();
 		}
 
@@ -191,7 +191,7 @@ $(function() {
 				channels: [data.chan]
 			})
 		);
-		renderChannelMessages(data.chan);
+		renderChannel(data.chan);
 		var chan = sidebar.find(".chan")
 			.sort(function(a, b) { return $(a).data("id") - $(b).data("id"); })
 			.last();
@@ -260,9 +260,36 @@ $(function() {
 		}, $(document.createDocumentFragment()));
 	}
 
+	function renderChannel(data) {
+		renderChannelMessages(data);
+		renderChannelUsers(data);
+	}
+
 	function renderChannelMessages(data) {
 		var documentFragment = buildChannelMessages(data.id, data.messages);
 		chat.find("#chan-" + data.id + " .messages").append(documentFragment);
+	}
+
+	function renderChannelUsers(data) {
+		var users = chat.find("#chan-" + data.id).find(".users");
+		var nicks = users.data("nicks") || [];
+		var i, oldSortOrder = {};
+
+		for (i in nicks) {
+			oldSortOrder[nicks[i]] = i;
+		}
+
+		nicks = [];
+
+		for (i in data.users) {
+			nicks.push(data.users[i].name);
+		}
+
+		nicks = nicks.sort(function(a, b) {
+			return (oldSortOrder[a] || Number.MAX_VALUE) - (oldSortOrder[b] || Number.MAX_VALUE);
+		});
+
+		users.html(render("user", data)).data("nicks", nicks);
 	}
 
 	socket.on("msg", function(data) {
@@ -401,14 +428,7 @@ $(function() {
 		}
 	});
 
-	socket.on("names", function(data) {
-		var users = chat.find("#chan-" + data.chan).find(".users").html(render("user", data));
-		var nicks = [];
-		for (var i in data.users) {
-			nicks.push(data.users[i].name);
-		}
-		users.data("nicks", nicks);
-	});
+	socket.on("names", renderChannelUsers);
 
 	var userStyles = $("#user-specified-css");
 	var settings = $("#settings");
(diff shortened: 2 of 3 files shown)


maxpoulin64:
Looks good to me, pretty clever solution the sort trick here! :+1:

YaManicKill:
> newly joined users are pushed to the top

Hmmm, not sure I like that. This could cause problems, I feel, if I'm talking to someone and then someone else joins with a similar name and suddenly I'm auto-completing their name.

xPaw:
It's not explicitly put at the top, but that appears to be how it works, might need more testing in that regard.

astorije:
So if I understand correctly, whenever a user joins, it gets added to the bottom of the completion list. So if we have users `['aa', 'aj', 'ax']` on the channel and `ab` joins, if I start typing `a` in the message input then hit <kbd>Tab</kbd> repeatedly, the following cycle will take place: `aa -> aj -> ax -> ab -> aa -> ...`.
Am I correct?
If so, why not just alphabetically, as this is what is expected (and also what happens when you refresh): `aa -> ab -> aj -> ax -> aa -> ...`?

xPaw:
Behaviour in master is to push talking people to the top, but the bug is that whenever the user list is refreshed, that order is destroyed (thus becomes alphabetical), this PR fixes it by keeping the order, and pushing new users to the bottom. And yes it gets reset on page refresh as there is no server tracking for this.

YaManicKill:
Yeah, I think it's much better to either do what this PR does, or to always do alphabetical entirely. If you do a kinda inbetween, you end up with confusion. I'm happy with this PR, and give me :+1:, but won't merge because it has changed since @maxpoulin64 gave it :+1: 

> And yes it gets reset on page refresh as there is no server tracking for this.

FWIW, we don't need server tracking for it. We just need to keep it in local storage. That might be a good thing to do in general.

src/client.js
```diff
@@ -340,7 +340,7 @@ Client.prototype.names = function(data) {
 	}
 
 	client.emit("names", {
-		chan: target.chan.id,
```
maxpoulin64: Shouldn't we keep this as-is, for consistency with the remaining of the code? While I admit `id` makes a lot more sense, everywhere else sends `chan: chan.id`:

```
➜ lounge git:(PR-217) grep -F 'chan.id' -B1 -R src/
src/plugins/irc-events/quit.js-			client.emit("users", {
src/plugins/irc-events/quit.js:				chan: chan.id
--
src/plugins/irc-events/join.js-		client.emit("users", {
src/plugins/irc-events/join.js:			chan: chan.id
--
src/plugins/irc-events/topic.js-		client.emit("topic", {
src/plugins/irc-events/topic.js:			chan: chan.id,
--
src/plugins/irc-events/message.js-
src/plugins/irc-events/message.js:		if (!self && chan.id !== client.activeChannel) {
--
src/plugins/irc-events/nick.js-			client.emit("users", {
src/plugins/irc-events/nick.js:				chan: chan.id
--
src/plugins/irc-events/kick.js-		client.emit("users", {
src/plugins/irc-events/kick.js:			chan: chan.id
--
src/plugins/irc-events/names.js-		client.emit("users", {
src/plugins/irc-events/names.js:			chan: chan.id
--
src/plugins/irc-events/part.js-			client.emit("part", {
src/plugins/irc-events/part.js:				chan: chan.id
--
src/plugins/irc-events/part.js-			client.emit("users", {
src/plugins/irc-events/part.js:				chan: chan.id
--
src/plugins/inputs/part.js-	this.emit("part", {
src/plugins/inputs/part.js:		chan: chan.id
--
src/client.js-	client.emit("more", {
src/client.js:		chan: chan.id,
--
src/client.js-		target.chan.highlight = false;
src/client.js:		this.activeChannel = target.chan.id;
--
src/client.js-	client.emit("names", {
src/client.js:		id: target.chan.id,
```

astorije:
Alright, the more I think about this, the less it makes sense to me, but here is a shift from the supposed behavior (supposed on master, fixed on this PR): why not just do like the entire world is doing and sort completion alphabetically?

Think about it, from a user standpoint, keeping a list of users based on how recently they joined is not intuitive at all, especially on a tool that helps users stay always connected!
From our standpoint, it's also much easier to handle: no custom sorting, but more importantly, no server tracking or localStorage or whatever needed.
In terms of usability, I actually can't think of a decent use case for not going alphabetical. Heck, the only thing I can think of is to say hi to someone who just joined, so if anything, they should be added to the top :D

Now that I think about it, I'm glad the autocomplete has been "broken" for longer than I can remember: it's been alphabetical ever since.

Am I out of line here? If I'm missing an obvious usability concern / use case for least recently joined vs. alphabetical autocomplete, I'd be happy to hear it. Removing the "second review needed" badge until then, to make sure that either we're all on the same page or I am dumb and missed something obvious :-)

src/client.js
xPaw: What do you mean by `consistency with the remaining of the code`? Changing `chan` to `id` allows the client side to use the same function to render user list without thinking about which variable to use.

xPaw:
@astorije Auto completion becomes infinitely more useful when it's context aware. Having to deal with alphabetical auto completion in big channels is complete pain in the ass.

Recently joined users are kept at the bottom of the sort list. People that talk are pushed all the way up, so they are guaranteed to be completed first.

astorije:
As discussed on the IRC channel, I understand the use case of keeping the 3-5 most recent people who talked at the top, and in case I wasn't clear, I like the idea. However, in all other use cases, it ends up building a list of nicks in a complete random order.
This will become even worse if we ever store things on server or local storage or whatever.

So I'm overall not OK with the feature, but as @xPaw said, let's fix this and build upon it. 👍 for the PR.

xPaw:
I will merge as @maxpoulin64 and @YaManicKill agreed on it, and @maxpoulin64 had a previous :+1:.

src/client.js
xPaw: Oops, I see what you mean now. Yeah we should just fix it in a separate PR to use `chan` everywhere then.

PR #218, opened 2016-03-24
Add a query command that simply opens a query window

### LINKED ISSUE #216: Re-add /query command to open query window with the specified user
`/query` was previously an alias for `/whois`, which was not in line with what other clients do. That alias was removed in #208, so currently the `/query` command does nothing.

Discussion in #214 (specifically, from @YaManicKill) requested "a way to open a query window with a command". Adding back `/query` to do just that seems apt, regardless of how it is eventually decided `/msg` should work.

xPaw:
Fixes #216

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -17,6 +17,7 @@ $(function() {
 		"/notice",
 		"/op",
 		"/part",
+		"/query",
 		"/quit",
 		"/raw",
 		"/say",
--- a/src/client.js
+++ b/src/client.js
@@ -42,6 +42,7 @@ var inputs = [
 	"kick",
 	"mode",
 	"notice",
+	"query",
 	"quit",
 	"raw",
 	"topic",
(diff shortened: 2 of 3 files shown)


astorije:
Sadly this has a very weird side effect: if you run `/query #thelounge` (or any other channel name), it opens a query window that lets you post messages to the channel without joining. Running `/join #thelounge` after that makes you join in the same window but doesn't update the icon in the sidebar, showing something is still off.

Do you think this can be fixed?

dgw:
In #216, @xPaw said that checking the input against `PREFIX` and/or `CHANTYPES` would have to wait until the irc-framework switch is made. Would you be in favor of checking a hardcoded list of channel-like prefixes in the mean time, @astorije?

xPaw:
@dgw This small side effect is not worth it to hardcode it, it can be tackled in irc-fw rather soon.

dgw:
I'm not the one you have to convince, @xPaw :wink:

astorije:
Alright, let's go with this and I'll add a quick fix until framework-irc gets in.
:+1:

PR #219, opened 2016-03-25
Harden url fetcher

xPaw:
Sending `https://google.com\x0f` would crash some lounge instances due to some obscurities in node.

This PR does following:
- Strip IRC formatters from messages before searching for urls
- Encode urls to prevent crashing
- Set timeout and max redirects

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -16,10 +16,9 @@ module.exports = function(irc, network) {
 		}
 
 		var links = [];
-		var split = data.message.split(" ");
+		var split = data.message.replace(/\x02|\x1D|\x1F|\x16|\x0F|\x03(?:[0-9]{1,2}(?:,[0-9]{1,2})?)?/g, "").split(" ");
 		_.each(split, function(w) {
-			var match = w.indexOf("http://") === 0 || w.indexOf("https://") === 0;
-			if (match) {
+			if (/^https?:\/\//.test(w)) {
 				links.push(w);
 			}
 		});
@@ -44,7 +43,7 @@ module.exports = function(irc, network) {
 			msg: msg
 		});
 
-		var link = links[0];
+		var link = escapeHeader(links[0]);
 		fetch(link, function(res) {
 			parse(msg, link, res, client);
 		});
@@ -103,6 +102,8 @@ function fetch(url, cb) {
 	try {
 		var req = request.get({
 			url: url,
+			maxRedirects: 5,
+			timeout: 5000,
 			headers: {
 				"User-Agent": "Mozilla/5.0 (compatible; The Lounge IRC Client; +https://github.com/thelounge/lounge)"
 			}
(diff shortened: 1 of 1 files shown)


src/plugins/irc-events/link.js
```diff
@@ -103,6 +102,8 @@ function fetch(url, cb) {
 	try {
 		var req = request.get({
 			url: url,
+			maxRedirects: 3,
+			timeout: 3000,
```
astorije: Are these arbitrary? If so, maybe we could go a bit higher, like `5`/`10000`, but I don't have an opinion.
I guess we would empirically adjust these values, worst case.

xPaw: Mostly arbitrary, 3 seconds should be enough to grab page title though. Having too big of a timeout won't be very nice to the user.

maxpoulin64:
Looks good to me, tested working. :+1: 

I agree with @astorije here that 3 redirects and 3 seconds is a bit low. It's somewhat easy to blow the 3 redirects. Short url + access to invalid area (= redirect to home page of the site) + redirect to regional page, we've reached 3. Some sites can take over 3 seconds to load when crowded (Reddit/imgur are easy examples) too, so 5/10000 sounds reasonable without being abusive.

xPaw:
Do keep in mind that timeout does not kick in while content is loading

maxpoulin64:
So that's a timeout for establishing the connection? I'm fine with that then.

astorije:
@xPaw, it makes sense re: title/content, but I often notice that long loading times happen even before the page starts getting in. I'm fine with 5 or 10 seconds then, still wondering if 3 is not a bit too short in some situations.
What about the redirects? @maxpoulin64 has a good point.

xPaw:
It should be, but needs to be looked in more closely to know for sure.

xPaw:
I've tweaked it to 5/5000

src/plugins/irc-events/link.js
```diff
@@ -16,10 +16,9 @@ module.exports = function(irc, network) {
 		}
 
 		var links = [];
-		var split = data.message.split(" ");
+		var split = data.message.replace(/\x02|\x1D|\x1F|\x16|\x0F|\x03(?:[0-9]{1,2}(?:,[0-9]{1,2})?)?/g, "").split(" ");
```
astorije: For the record/archive purpose, this is the same check as [this one](https://github.com/thelounge/lounge/blob/05ec819f2d74fb2e2d13dc135e169ddd21852143/client/js/lounge.js#L779), to clean up messages from colors, boldness, ...

src/plugins/irc-events/link.js
```diff
@@ -16,10 +16,9 @@ module.exports = function(irc, network) {
...
 		_.each(split, function(w) {
-			var match = w.indexOf("http://") === 0 || w.indexOf("https://") === 0;
-			if (match) {
+			if (/^https?:\/\//.test(w)) {
```
astorije: I'm not going to block this, but in general I'm not in favor of obfuscating a couple checks with a regex. It's shorter, yes, but regexes are arguably harder to understand and, when not backed with tests, tend to let bugs go through more easily.
The previous check was a tiny bit longer but more straightforward to read (granted, storing this in that `match` variable was unnecessary).

src/plugins/irc-events/link.js
```diff
@@ -150,3 +151,13 @@ function fetch(url, cb) {
 			cb(data);
 		}));
 }
+
+// https://github.com/request/request/issues/2120
+// https://github.com/nodejs/node/issues/1693
+// https://github.com/alexeyten/descript/commit/50ee540b30188324198176e445330294922665fc
+function escapeHeader(header) {
```
astorije: A comment to explain this, in addition to the links, would have be nice, for future ourselves.
To keep in mind when we start adding tests to these.

astorije:
:+1:, thanks for this @xPaw!

PR #221, opened 2016-03-26
Disallow /query on non-nicks

astorije:
Quickfix of #218 until `framework-irc` probably makes us solve that differently anyway.
At least it avoids going into a weird state that cannot be easily recovered.

### DIFF
--- a/src/plugins/inputs/query.js
+++ b/src/plugins/inputs/query.js
@@ -14,6 +14,11 @@ exports.input = function(network, chan, cmd, args) {
 		return;
 	}
 
+	// If target doesn't start with an allowed character, ignore
+	if (!/^[a-zA-Z_\\\[\]{}^`|]/.test(target)) {
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


xPaw:
:-1: I've tackled this in 87d0f14070525f60020b3c50d774e88afc66386b, and I want that branch merged asap.

astorije:
@xPaw, I would like to release ASAP and I decently can't with the bug introduced by #218. Would you mind reconsidering please?

Also, the framework-irc PR is clearly not for right now (for example, we can't ship before `/invite` is supported or we would introduce a regression) so we're talking about a bit more wait, and it's not reasonable to leave this bug in `master` until then since this PR fixes it.

Thanks!

xPaw:
> Also, the framework-irc PR is clearly not for right now (for example, we can't ship before /invite is supported or we would introduce a regression) 

It is supported.

astorije:
> It is supported

My bad

PR #222, opened 2016-03-26
Fix /part command

maxpoulin64:
Fixes the /part command closing the wrong window. The current implementation simply passes all arguments to slate, which ended up parting every arguments (and never sending the part message)

This changes the command to `/part message`, and always parts the current window. This will be fixed further once irc-framework is merged.

### DIFF
--- a/src/plugins/inputs/part.js
+++ b/src/plugins/inputs/part.js
@@ -17,10 +17,7 @@ exports.input = function(network, chan, cmd, args) {
 
 	if (chan.type === "channel") {
 		var irc = network.irc;
-		if (args.length === 0) {
-			args.push(chan.name);
-		}
-		irc.part(args);
+		irc.part(chan.name, args.join(" "));
 	}
 
 	network.channels = _.without(network.channels, chan);


astorije:
@maxpoulin64, so it means that if the user is currently on channel `#one` and types `/part #two`, they will leave both `#one` and `#two`? If so, that seems odd, don't you think?

Correct me if I'm wrong but the current implementation of `/part` seems similar (or at least, aims to be similar) to the `/invite` command:
- `/invite #channel nick` invites `nick` in `#channel`; `/part #channel` parts the current user from `#channel`
- `/invite nick` is a shorthand to invite `nick` to the current channel; `/part` is a shorthand to part the current user from the current channel

Am I being mistaken in understanding your change?

maxpoulin64:
@astorije

> so it means that if the user is currently on channel #one and types /part #two, they will leave both #one and #two? If so, that seems odd, don't you think?

Actually that's precisely what this fixes. I will change it to `/part [channel] [message]` once irc-fw is merged, because I can't reliably know if `[channel]` is a valid channel.

astorije:
> I will change it to /part [channel] [message] once irc-fw is merged, because I can't reliably know if [channel] is a valid channel.

Makes sense.
I also notice that [the doc](https://thelounge.github.io/docs/client/commands.html#part) says "Leave the current channel.", ouch. So doc and code were not completely consistent :-/

So now, if I type `/part #channel hello world`, it means I will part from the current channel and the message will be `#channel hello world`?

maxpoulin64:
> So now, if I type /part #channel hello world, it means I will part from the current channel and the message will be #channel hello world?

Yes, unfortunately. That will have to wait for irc-fw however, because I need access to `CHANTYPES` to fix is properly.

Note that the old behavior was to try to quit everything, so `/part #channel hello world` would have actually attempted to quit `#channel`, `hello` and `world` with a part message of `"undefined"`.

astorije:
> Yes, unfortunately. That will have to wait for irc-fw however, because I need access to CHANTYPES to fix is properly.

Fine, even if not perfect, definitely much better than the current state then :-)
I therefore :+1: this PR *a posteriori* :D

PR #227, opened 2016-03-30
Hide options will now remove the row completely

xPaw:
Fixes this: https://forum.cozy.io/uploads/default/original/1X/d75836720701429278ddaeebb35ddb3fab7656c4.PNG

1.4.1 uses fixed table layout, so incorrectly hiding rows revealed the bug.


maxpoulin64:
:+1:

PR #228, opened 2016-03-30
Remove print styles

xPaw:
They were added in 9e1c33d658d59d775abb6f4c1a629ec0598ba155 for no apparent reason, and it's a bit odd feature to support, as it breaks when we add new features, and it breaks with use of themes.

If you really want to print, just copy the messages into word and print it from there.

Discuss.


YaManicKill:
Sounds fine to me. I won't merge in just now so others can input if they want, but it looks like no one else really cares :-P

I don't see anyone printing irc out, and if they do, I see them either wanting the page to look exactly the same, or to just copy it into a text document.

:+1:

maxpoulin64:
That's a :+1: for me as well. Won't merge to leave time for @astorije opinion.

Alternatively we could change it to be an entirely separate stylesheet with `media="print"`, and thus keep it entirely separate without removing an existing feature. That way it wouldn't break with theme changes and everything. However, I agree, I seriously don't know why anyone would want to print IRC logs that way. I'd print at least a selection of it, and thus paste it in Word before. It's HTML so the formatting is kept anyway.

astorije:
What bothers me too is that one currently cannot print the channel list currently (if such one ever needs to print :-) ) so :+1: and merging.

PR #230, opened 2016-03-31
command-line/add,reset: reword password prompt

williamboman:
This makes the password prompts not look like a root access password prompt.

### DIFF
--- a/src/command-line/add.js
+++ b/src/command-line/add.js
@@ -39,7 +39,7 @@ program
 			return;
 		}
 		require("read")({
-			prompt: "Password: ",
+			prompt: "[thelounge] Enter password: ",
 			silent: true
 		}, function(err, password) {
 			if (!err) add(manager, name, password);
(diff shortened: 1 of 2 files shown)


YaManicKill:
Interesting. Should we make it even more obvious by saying something like "Enter TheLounge password"?

williamboman:
> Interesting. Should we make it even more obvious by saying something like "Enter The Lounge password"?

I think `The Lounge password` is a bit dubious, something like `Enter new user's password: ` perhaps?

maxpoulin64:
I'm really not a fan of super long prompts. Maybe `New password:` would be enough to clear the confusion with the root password? Or maybe we could go the sudo way, `[sudo] Password for max-p:` and write something like `[lounge] New password:`?

In general I think it shouldn't be our problem that the user can confuse prompts from other programs they use with The Lounge. If you use `su -c` you should be aware that `su` has a `Password:` prompt before ours.

williamboman:
> and write something like [lounge] New password:

:+1:, I like it!

maxpoulin64:
Fine by me :+1:

YaManicKill:
Yes, I like this. However, as @astorije would probably say, it should probably be [thelounge] as that's our shortened version of the name.

PR #233, opened 2016-04-02
Fix not being able to remove networks from user config

### LINKED ISSUE #211: Saving user config merges current config in
This makes it absolutely impossible to quit networks because they will stay in the user config and you will re-join them when the server is restarted.

xPaw:
Fixes #211, a bug introduced by #57.

Previously it used to simply override entire `networks` object, but #57 made it use `merge` method, which makes it impossible to remove disconnected networks. This PR changes it to `assign` which works similarly to `merge`, but only on root keys, which overrides the `networks` object just fine.

http://stackoverflow.com/a/33247597/2200891

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -101,7 +101,7 @@ ClientManager.prototype.updateUser = function(name, opts) {
 
 	try {
 		user = this.readUserConfig(name);
-		_.merge(user, opts);
+		_.assign(user, opts);
 		fs.writeFileSync(
 			path,
 			JSON.stringify(user, null, " ")


YaManicKill:
So, am I right in saying that if we are removing a network, the opts object will have that network as undefined, and "assign" will then override it to be undefined in the user object, but "merge" will always set it to the non-undefined one, which means it can't be removed?

xPaw:
Not sure I understand, in which case would it be `undefined`? The network is removed using `_.without`. And `opts` is always passed in with networks object.

https://github.com/thelounge/lounge/blob/2201e2b941bd13427c38d6934b84ceda245aa20e/src/plugins/inputs/quit.js#L10
https://github.com/thelounge/lounge/blob/2201e2b941bd13427c38d6934b84ceda245aa20e/src/client.js#L411-L413

YaManicKill:
Well I don't understand the change then. What is assign doing that merge doesn't then?

xPaw:
It only overwrites root keys, so the entire networks object is overwritten.

YaManicKill:
Right, that's what I meant. So the new networks object has the old networks without the one(s) that was removed. If merge was being used, it would take the non-undefined networks from the user object and use them in the merge, yeah?

xPaw:
Yes.

YaManicKill:
Cool. :+1: . Simple fix, so no need for second review. Merging.

PR #237, opened 2016-04-02
Dockerfile: don't use -onbuild tag, more explicit directives

williamboman:
Right now it assumes you have run `$ npm run build` before building the image.

edit: retroactively added a `.dockerignore` as well

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+node_modules
+
+# Built assets created at npm install/prepublish time
+# See https://docs.npmjs.com/misc/scripts
+client/js/libs.min.js.map
+client/js/libs.min.js
+client/js/lounge.templates.js
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,4 @@
-#
-# Thanks to @Xe for the Dockerfile template
-# https://github.com/Shuo-IRC/Shuo/pull/87/files
-#
-
-FROM node:4.0-onbuild
+FROM node:4
 
 # Create a non-root user for lounge to run in.
 RUN useradd --create-home lounge
@@ -13,13 +8,23 @@ ENV HOME /home/lounge
 
 # Customize this to specify where The Lounge puts its data.
 # To link a data container, have it expose /home/lounge/data
-ENV LOUNGE_HOME /home/lounge/data
+ENV LOUNGE_HOME ${HOME}/data
+ENV LOUNGE_SRC ${HOME}/src
 
-# Expose HTTP
-EXPOSE 9000
+RUN mkdir -p ${LOUNGE_SRC}
+WORKDIR ${LOUNGE_SRC}
+
+COPY . ${LOUNGE_SRC}
+
+RUN chown -R lounge:lounge ${LOUNGE_SRC}
 
 # Drop root.
 USER lounge
+RUN npm install
+
+# Expose HTTP
+ENV PORT 9000
+EXPOSE ${PORT}
 
 # Don't use an entrypoint here. It makes debugging difficult.
 CMD node index.js --home $LOUNGE_HOME


maxpoulin64:
Hmm, I don't really know how Docker works, but where does the files come from? Does one just clone our git repo and feed the Dockerfile to Docker? The user should never really have to run `npm run build`. Either the user clones the repo and run `npm install`, which will run the build as well, or the user downloaded an already pre-built tarball from npm. But I'm not seeing `npm install` anywhere in that Dockerfile so I'm a bit confused as to why we would need this. In which scenario would the user have lounge installed, but not ran the build script?

lindskogen:
I think the `onbuild` bundles are supposed to run `npm install` automatically

williamboman:
The `onbuild` tag [does some extra stuff when building the image](https://github.com/nodejs/docker-node/blob/master/4.4/onbuild/Dockerfile).

`$ npm install` _should_ run the `prepublish` script - but [I'm not sure how long that'll be the case](https://github.com/npm/npm/issues/10074) so one should not depend on it. It does not seem to be running the `prepublish` script when building though, not sure why?

```sh
$ git clean -fdx # make sure everything's wiped
$ docker build -t lounge-test .
$ docker run -d -p 9000:9000 --name lounge-test lounge-test
# Lounge now throws errors like `$ is not defined`
$ docker exec -ti lounge-test /bin/bash
$ root@lounge-test: npm run build
# Now everything works when reloading the browser
```

So obviously something's not working when building from a clean repo.

YaManicKill:
> $ npm install should run the prepublish script - but I'm not sure how long that'll be the case so one should not depend on it.

This isn't an issue relying on it. npm 4 will add another hook that will run on npm install, prepublish will be deprecated (but still work) then npm 5 will change prepublish to only work on publish, and the new hook will run on npm install. So, we'll just add the other hook when we get to that point, not a problem.

So, to the issue, it's probably an issue with the permissions. Reading [the docks for the onbuild file](https://github.com/nodejs/docker-node) I note that they don't recommend using this for production.

> While the onbuild variant is really useful for "getting off the ground running" (zero to Dockerized in a short period of time), it's not recommended for long-term usage within a project due to the lack of control over when the ONBUILD triggers fire (see also docker/docker#5714, docker/docker#8240, docker/docker#11917).

> Once you've got a handle on how your project functions within Docker, you'll probably want to adjust your Dockerfile to inherit from a non-onbuild variant and copy the commands from the onbuild variant Dockerfile (moving the ONBUILD lines to the end and removing the ONBUILD keywords) into your own file so that you have tighter control over them and more transparency for yourself and others looking at your Dockerfile as to what it does. This also makes it easier to add additional requirements as time goes on (such as installing more packages before performing the previously-ONBUILD steps).

So, this is probably the issue. The npm install is probably being run after dropping permissions, meaning that we have permissions issues then. It's probably much easier to specify it all manually and then we know what the hell is actually happening. I'd much rather that than this solution.

williamboman:
You're just gonna have more perm issues to resolve if you want to run everything as the `lounge` user. I think the current setup is fine. Being more explicit (dropping `-onbuild`) is just redundant - the current Dockerfile is really trivial.

Also, Docker caching will benefit from the separation of installing dependencies and executing custom subscripts, leading to faster builds.

YaManicKill:
I'm still not convinced about this solution, because (according to the node documentation) we essentially have a race condition in building this. Making it all explicit will remove that and make it much clearer what we are doing.

You can abstract things too much. I'm not going to give a +1 for this, but I won't stand in the way of it if others think it's fine, but I'd much prefer it being rewritten.

williamboman:
> we essentially have a race condition in building this.

huh?

YaManicKill:
As mentioned, I misunderstood the node documentation for the docker file.

The updated one, being explicit and dropping the rights as soon as we can is much better.

I give my :+1: for this,

xPaw:
Before we merge this, can we move all the docker stuff into a separate repository?

astorije:
This was moved to the Docker-specific repo: https://github.com/thelounge/docker-lounge/pull/1.

PR #238, opened 2016-04-02
clientManager: don't write user configs outside of lounge's users dir

williamboman:


package.json
```diff
@@ -35,6 +35,7 @@
     "lodash": "4.6.1",
     "mkdirp": "0.5.1",
     "moment": "2.12.0",
+    "path-is-inside": "1.0.1",
```
xPaw: Let's not bring in modules for stuff that can be done in one line.

I'm thinking we should only be allowing `a-zA-Z0-9_-` in usernames.

williamboman: [Not really a one-line though](https://github.com/domenic/path-is-inside/blob/master/lib/path-is-inside.js) :). A custom one-liner is also far from as robust and well-tested (across numerous platforms) as a package too. Unix philosophy + Node ftw :)



> I'm thinking we should only be allowing a-zA-Z0-9_- in usernames.

I thought so as well, but didn't want to make such a decision in this PR.

xPaw: > A custom one-liner is also far from as robust and well-tested (across numerous platforms) as a package too. Unix philosophy + Node ftw!

/me looks at left-pad in panic.

williamboman: Haha, that's just the shortcomings of `npm` though. They've retroactively done some damage control by disabling `unpublish` on packages that are >24 hrs old.

williamboman: Do you want me to change to testing the name with `/[a-z0-9_-]/i` instead?

xPaw: I'd like to hear what other maintainers have to say about this first.

maxpoulin64: I think ideally we should simply disallow what is not valid characters for the filesystem. `/^[^\/\\<>:"|?*]+$/`. This should cover at least Linux and Windows, and probably Mac as well. I think Windows is alone having such restricted file names.

Source: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx

xPaw: We can use [path.basename](https://nodejs.org/api/path.html#path_path_basename_path_ext) to fix writing outside.

src/clientManager.js
```diff
@@ -70,16 +72,20 @@ ClientManager.prototype.addUser = function(name, password) {
 		return false;
 	}
 	try {
-		var path = Helper.HOME + "/users";
+		var usersPath = Helper.HOME + "/users";
+		var newUserPath = path.join(usersPath, name + ".json");
```
xPaw: Just noticed this, do `path.join(Helper.HOME, "users", name + ".json");`

src/clientManager.js
```diff
@@ -70,16 +72,20 @@ ClientManager.prototype.addUser = function(name, password) {
 		return false;
 	}
 	try {
-		var path = Helper.HOME + "/users";
+		var usersPath = Helper.HOME + "/users";
```
xPaw: Just noticed this, do `path.join(Helper.HOME, "users");` instead.

package.json
williamboman: > We can use path.basename to fix writing outside.

Do you mean that `../../MZuckerberg` silently becomes `MZuckerberg`, or should it do it interactively, like;

```js
if (path.basename(user) !== user) {
    // interactively ask if path.basename(user) should be used instead - else error
}
```

xPaw: I'd just error out.

williamboman: Updated patch now

xPaw:
Looks good to me. Should we perform same check in any other commands? Like edit or remove?

williamboman:
> Also looking at this code try/catch is completely useless here...

Not anymore!

xPaw:
:+1:

williamboman:
[dont merge for now, last commit is WIP]

src/clientManager.js
```diff
@@ -3,13 +3,31 @@ var fs = require("fs");
 var Client = require("./client");
 var mkdirp = require("mkdirp");
 var Helper = require("./helper");
+var path = require("path");
+
+var USERS_PATH = path.join(Helper.HOME, "users");
```
xPaw: You can directly assign it to `ClientManager.USERS_PATH`, and in this file use `this.USERS_PATH`

williamboman: "static" methods are hidden from the prototype chain, so `this.USERS_PATH` would be `undefined`. I could do `var USERS_PATH = ClientManager.USERS_PATH = path....` though

xPaw: You sure about that? It seems to work.
```
function ClientManager() {}
ClientManager.test = 'kek';
ClientManager.w = function(){console.log(this.test);}
ClientManager.w();
```

williamboman: Yeah that would work since you're not working with its prototype;
```
function CM() {}
CM.FOO = 'bar';
CM.prototype.bar = function () { console.log(this.FOO); };
a = new CM
a.bar(); // undefined
```

xPaw: Heh, I went that way as your two new functions didn't use `prototype`, but others in the file do.

maxpoulin64:
Looks good to me. I still think it would have been better to check for valid file names entirely, but this works too. 👍 

Leaving open so other people can confirm and merge.

astorije:
👍

PR #239, opened 2016-04-02
Display whitespace properly

maxpoulin64:
Fixes alignment in many server's motd as well as in-channel ASCII art

## Before
before

## After
after

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -760,6 +760,14 @@ button,
 	color: #999;
 }
 
+#chat .msg.motd .text,
+#chat .msg.message .text,
+#chat .msg.action .action-text,
+#chat .msg.notice .text {
+	white-space: pre-wrap;
+	overflow: hidden;
+}
+
 #chat.hide-join .join,
 #chat.hide-mode .mode,
 #chat.hide-motd .motd,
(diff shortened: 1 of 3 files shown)


astorije:
Yes!! Thanks for that!!
I'll try some edge cases before giving my approval though.

maxpoulin64:
@astorije Just fixed a bug with action messages. I think we should be covered now.

xPaw:
Why not set it on `#chat .msg .text`?

maxpoulin64:
@xPaw because it would break the other action templates due to extra whitespace. I had to edit `msg.tpl` and `msg_action.tpl` to glue the text right on the span, otherwise the extra tabs/spaces before and after would be displayed as well, making a mess. So I targetted those and only those to avoid bad surprises and WTF with the action templates.

xPaw:
Ah yes, good point. Have you tested if super long/unicode lines overflow correctly with this patch?

maxpoulin64:
It should be no different than before, since it's `pre-wrap` and not just `pre`. It should only affect tabs and space, but I admit I haven't tested that case. I can check a bit later if you want.

xPaw:
I'm just wondering if it's going to conflict with `word-wrap: break-word;` we have on `.msg` or not.

xPaw:
Unrelated note, can we get a `overflow: hidden` in that same selector? So that zalgo text and other crap doesn't overflow onto other lines.

maxpoulin64:
@xPaw makes sense, updated.

astorije:
Good job, guys! :+1:

PR #240, opened 2016-04-03
Implement WEBIRC

maxpoulin64:
It's finally ready!

Features:
* Works in both public/private mode
* Supports DNS resolution (without the possible race condition of the original PR)
* Configurable hostname prefix to mark the web client
* In private mode, uses the IP of the client creating the new network (and is saved to JSON for server restarts)
* IP/hostname can be overriden *per user* in the user's JSON
* IP/hostname can be overriden *per network* in the user's JSON
* Gracefully falls back to a standard connection of the client IP can't be obtained and issue a warning to the console. (Should only happen when editing the JSON files manually)

Overall, if enabled in the `config.js`, it should do what an admin would expect both in public and private mode. However, admins that wants more control can override settings as needed per user, or per user network.

EDIT: [An irc-framework version](https://github.com/maxpoulin64/lounge/tree/webirc-irc-fw) to make rebasing easier or in case irc-fw gets merged first.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -129,6 +129,7 @@ Client.prototype.connect = function(args) {
 	var client = this;
 
 	var nick = args.nick || "lounge-user";
+	var webirc = null;
 
 	var network = new Network({
 		name: args.name || "",
@@ -138,7 +139,9 @@ Client.prototype.connect = function(args) {
 		password: args.password,
 		username: args.username || nick.replace(/[^a-zA-Z0-9]/g, ""),
 		realname: args.realname || "The Lounge User",
-		commands: args.commands
+		commands: args.commands,
+		ip: args.ip,
+		hostname: args.hostname,
 	});
 
 	client.networks.push(network);
@@ -169,6 +172,26 @@ Client.prototype.connect = function(args) {
 		return;
 	}
 
+	if (config.webirc !== null && network.host in config.webirc) {
+		args.ip = args.ip || (client.config && client.config.ip) || client.ip;
+		args.hostname = args.hostname || (client.config && client.config.hostname) || client.hostname || args.ip;
+
+		if (args.ip) {
+			if (config.webirc[network.host] instanceof Function) {
+				webirc = config.webirc[network.host](client, args);
+			} else {
+				webirc = {
+					password: config.webirc[network.host],
+					address: args.ip,
+					hostname: args.hostname
+				};
+			}
+		} else {
+			log.warn("Cannot find a valid WEBIRC configuration for " + nick
+				+ "!" + network.username + "@" + network.host);
+		}
+	}
+
 	network.irc = new ircFramework.Client();
 	network.irc.requestCap([
 		"echo-message",
@@ -185,6 +208,7 @@ Client.prototype.connect = function(args) {
 		localAddress: config.bind,
 		rejectUnauthorized: false,
 		auto_reconnect: false, // TODO: Enable auto reconnection
+		webirc: webirc,
 	});
 
 	network.irc.on("registered", function() {
--- a/src/server.js
+++ b/src/server.js
@@ -6,6 +6,7 @@ var ClientManager = require("./clientManager");
 var express = require("express");
 var fs = require("fs");
 var io = require("socket.io");
+var dns = require("dns");
 var Helper = require("./helper");
 var config = {};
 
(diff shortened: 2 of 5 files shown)


xPaw:
Code looks fine. I think this will have to wait for irc-fw to be merged.

maxpoulin64:
Hmm, I'm thinking since the config file is actually Javascript, I think I'll make the `webirc` object accept both strings (for just the password with no extra stuff) or a function that returns the webirc object that irc-framework takes. I have noticed several ways of setting the hostnames over multiple networks. Some use just the reverse DNS. Some use `ip-xx-xx-xx-xx.web.example.net`, some use `gateway/web/host.provider.dsl.example.net`, some use just the IP. Making that option take a function would allow the admin to configure it at wish, and possibly also include more data as needed.

maxpoulin64:
Updated as irc-fw just got merged. Also added the function thing I said I'd do, so now the admins can do whatever they want, and if the admin only specifies the password then it follows the standard.

src/server.js
```diff
@@ -103,7 +111,7 @@ function init(socket, client, token) {
 		socket.on(
 			"conn",
 			function(data) {
-				client.connect(data);
+				client.connect(data, false);
```
xPaw: What if instead of passing in `trusted` boolean, we filtered it here?

maxpoulin64: That's a great idea, I'll change that!

xPaw:
:+1:

src/client.js
```diff
@@ -173,6 +179,27 @@ Client.prototype.connect = function(args) {
...
+			}
+			else {
+				webirc = {
+					password: config.webirc[network.host],
+					address: args.ip,
+					hostname: args.hostname
+				};
+			}
+		} else {
+			console.warn("Cannot find a valid WEBIRC configuration for " + nick
```
xPaw: Change to `log.warn` which is now in master.

src/client.js
```diff
@@ -79,6 +79,9 @@ function Client(manager, name, config) {
 	log.info("User '" + name + "' loaded");
 }
 
+Client.prototype.ip = null;
+Client.prototype.host = null;
+
```
astorije: These lines can be removed.

src/client.js
```diff
@@ -169,6 +175,27 @@ Client.prototype.connect = function(args) {
...
 
+	if (config.webirc !== null && network.host in config.webirc) {
+		args.ip = args.ip || (client.config && client.config.ip) || client.ip;
+		args.hostname = args.hostname || (client.config && client.config.hostname) || client.hostname || args.ip;
+
+		if (args.ip) {
+			if (config.webirc[network.host] instanceof Function) {
+				webirc = config.webirc[network.host](client, args);
+			}
+			else {
```
astorije: Move `else` to the previous line (I'm surprised ESLint didn't catch that).

src/server.js
```diff
@@ -176,6 +187,21 @@ function init(socket, client, token) {
...
 }
 
+function reverseDnsLookup(socket, client, token) {
+	client.ip = getClientIp(socket.request);
+
+	dns.reverse(client.ip, function(err, host) {
+		if (!err && host.length) {
+			client.hostname = host[0];
+		}
+		else {
```
astorije: Move `else` to the previous line.

astorije:
@maxpoulin64 and I went through the code together offline, 👏 for the feature!
As discussed, a few minor comments and then it should be fine :-)

astorije:
👍 and merging, great job on this, @maxpoulin64!

spencerthayer:
This has been a LONG time coming. I've been waiting for Shout to implement the web IRC protocol for quite some time. Is this working out of the box? (_Waiting in excitement._)

maxpoulin64:
@spencerthayer: what do you mean by out of the box? You still need to configure it to put your webirc password for each network, but yes it otherwise just works.

CayneVamp-Inn:
question how do I do the format to add the webirc implement in Ive tried everything

astorije:
@CayneVamp-Inn, we have not documented this on the website at the moment, so your best bet is to follow the WEBIRC section of the configuration file, [which lives here](https://github.com/thelounge/thelounge/blob/master/defaults/config.js#L287-L322).

As I am rewriting the docs, there is a planned guide for WEBIRC that is currently empty ([code](https://github.com/thelounge/thelounge.github.io/pull/84/files#diff-0043d8a55e28ad28068c2d65ef9e6b38) / [preview](https://deploy-preview-84--thelounge-chat.netlify.com/docs/guides/webirc)), so once you figure it out, if you or anyone else wants to help us, you can come up with a first version of the guide.

Good luck!

PR #243, opened 2016-04-03
Add config option to limit in-memory history size

maxpoulin64:
This adds a (temporary?) config option to limit the amount of messages stored per channel to avoid the server's memory usage to grow as channels fills up with messages.

This is the same patch that I had originally made for Shout, but it was obsoleted by another PR that offered persistent storage. However, since we're already over 4 months later and people's instances still crashes regularly, I'm bringing it back at least as a temporary measure.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -1,9 +1,12 @@
 var _ = require("lodash");
 var Chan = require("../../models/chan");
 var Msg = require("../../models/msg");
+var Helper = require("../../helper");
 
 module.exports = function(irc, network) {
 	var client = this;
+	var config = Helper.getConfig();
+
 	irc.on("message", function(data) {
 		if (data.message.indexOf("\u0001") === 0 && data.message.substring(0, 7) !== "\u0001ACTION") {
 			// Hide ctcp messages.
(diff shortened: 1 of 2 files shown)


src/plugins/irc-events/message.js
```diff
@@ -1,6 +1,7 @@
 var _ = require("lodash");
 var Chan = require("../../models/chan");
 var Msg = require("../../models/msg");
+var config = require("../../helper").getConfig();
```
xPaw: Other code paths call `getConfig()` whenever they need it, without caching.

src/plugins/irc-events/message.js
```diff
@@ -62,6 +65,11 @@ module.exports = function(irc, network) {
 			highlight: highlight
 		});
 		chan.messages.push(msg);
+
+		if (config.maxHistory < 0 && chan.messages.length > config.maxHistory) {
```
xPaw: Why the `< 0`? Shouldn't that be `> 0`?

maxpoulin64: That was your imagination I swear.

xPaw:
:+1:

YaManicKill:
Looks good, nice and clean and a simple fix for something that we'll take ages to actually implement (database). Eventually...

:+1: and merging

kode54:
If you mean erming/shout#391, that PR is now in limbo because its author has since been deleted from Github. And it's failing testing due to a timeout.

richrd:
Hey by the way, when database logging is implemented, will I still be able to use in-memory logging if I desire to? It's a must have feature for me since I don't want any messages stored persistently on disk. Hopefully in memory logging could be kept as an option.

PR #245, opened 2016-04-04
Do not parse link titles for IRC formatting

xPaw:
This was supposed to be an XSS fix in 7ef2da0c832175cea1776a651a3c14051468fdc2, but the proper fix would be to simply escape it, instead of running IRC formatter on it.

### DIFF
--- a/client/views/toggle.tpl
+++ b/client/views/toggle.tpl
@@ -9,7 +9,7 @@
 			{{#if thumb}}
 				<img src="{{thumb}}" class="thumb">
 			{{/if}}
-			<div class="head">{{{parse head}}}</div>
+			<div class="head">{{head}}</div>
 			<div class="body">
 				{{body}}
 			</div>


maxpoulin64:
Agreed, doesn't even make sense. :+1:

astorije:
For the record, this was a fix for a pretty bad XSS vulnerability: https://github.com/erming/shout/pull/455.
I remember when Shout had it, it messed up people's client for a while.

Are there any other consequences to this change?

Thanks for this, @xPaw! :+1:

maxpoulin64:
@astorije No, no consequences really. We shouldn't be doing that. This makes it so it is strictly text, which is good. There's no reason to parse links and colors in web page titles, they're not meant to be displayed as such anyway.

The original code was `{{{head}}}` which allowed everything through. This one is `{{head}}`, which lets nothing through.

astorije:
Agreed, thanks a lot for the details!

PR #247, opened 2016-04-06
Updates README demo and IRC links, closes #242

### LINKED ISSUE #242: Demo no longer connects to freenode, remove/update badge
Should the badge just be removed, or do we want to provide a link to freenode webchat/something else?

I think there should still be a badge there, letting people know a demo exists.

MaxLeiter:


maxpoulin64:
I think it feels a bit odd to make an IRC client and yet point to Freenode's qwebirc. I'll give my :+1: anyway because right now the situation is a bit less than ideal with the bridge bot. I'll let another maintainer step in and  agree or disagree.

MaxLeiter:
This was created before the bridge bot was made, so now it's obsolete imo. I'll leave it open for discussion in case another maintainer decides to :+1: it.

astorije:
I agree with @MaxLeiter here: now that there is a bridge to the Freenode channel, we should keep pointing to The Lounge instead of qwebirc.
Thanks both for looking into this and providing the bridge!

PR #252, opened 2016-04-10
Do not display incorrect nick when switching to a non connected network

xPaw:
Previously it only called `setNick` if the network had a nick, so it would keep displaying the nick from the  network you switched from, this PR fixes that.

- It also removes the magic `31` margin number, and calculates the margin correctly.
- Moves <kbd>:</kbd> to be appended with CSS.
- Hides `#nick` panel if the network has no nick (e.g. while lauching lounge or connecting)

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1157,6 +1157,7 @@ button,
 	font: inherit;
 	font-size: 11px;
 	margin: 5px;
+	margin-right: 10px;
 	line-height: 26px;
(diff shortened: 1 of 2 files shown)


maxpoulin64:
Whoops, that one went unnoticed for a while. 👍 

This seems pretty minor to me, feel free to merge if you feel the same.

astorije:
👍

PR #254, opened 2016-04-13
Enable auto reconnection

### LINKED ISSUE #18: Reconnect on lost IRC connection
- All connection messages should be shown in server window
- Use exponential back off
- Do not reconnect on errors like "bad password"
- Re-join all previous channels
- Existing `chan.id` must be reused for networks/channels/query windows when reconnecting to maintain history.

xPaw:
:tada: irc-fw has prepared us for this moment! Closes #18.

We should be reconnecting just fine without any problems. Client already prints all the necessary messages (connecting, reconnecting, socket errors, etc). Server prevents you from sending commands while you are not connected.

### DIFF
--- a/src/plugins/inputs/connect.js
+++ b/src/plugins/inputs/connect.js
@@ -1,8 +1,24 @@
+var Msg = require("../../models/msg");
+
 exports.commands = ["connect", "server"];
 exports.allowDisconnected = true;
 
 exports.input = function(network, chan, cmd, args) {
 	if (args.length === 0) {
+		if (!network.irc || !network.irc.connection) {
+			return;
+		}
+
+		if (network.irc.connection.connected) {
+			chan.pushMessage(this, new Msg({
+				type: Msg.Type.ERROR,
+				text: "You are already connected."
+			}));
+			return;
+		}
+
+		network.irc.connection.connect();
+
 		return;
 	}
 
--- a/src/plugins/irc-events/connection.js
+++ b/src/plugins/irc-events/connection.js
@@ -19,9 +19,9 @@ module.exports = function(irc, network) {
 		}));
 	});
 
-	irc.on("socket close", function() {
+	irc.on("close", function() {
 		network.channels[0].pushMessage(client, new Msg({
-			text: "Disconnected from the network."
+			text: "Disconnected from the network, and will not reconnect."
 		}));
 	});
 
@@ -35,7 +35,7 @@ module.exports = function(irc, network) {
 
 	irc.on("reconnecting", function() {
 		network.channels[0].pushMessage(client, new Msg({
-			text: "Reconnecting..."
+			text: "Disconnected from the network. Reconnecting..."
 		}));
 	});
 
--- a/src/client.js
+++ b/src/client.js
@@ -36,6 +36,7 @@ var inputs = [
 	"part",
 	"action",
 	"connect",
+	"disconnect",
 	"invite",
 	"kick",
 	"mode",
--- a/src/client.js
+++ b/src/client.js
@@ -211,7 +211,7 @@ Client.prototype.connect = function(args) {
 		tls: network.tls,
 		localAddress: config.bind,
 		rejectUnauthorized: false,
-		auto_reconnect: false, // TODO: Enable auto reconnection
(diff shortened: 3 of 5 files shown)


maxpoulin64:
This doesn't work at all [due to a bug in irc-framework](https://github.com/kiwiirc/irc-framework/issues/34). It just hangs in the registration phase and eventually timeout and try again.

I don't understand that part of the irc-framework part as the code is hard to follow, but it sounds like there's some state tracking that needs to be reset that isn't.

astorije:
Hey guys, do you think there is a fix in progress on the irc-framework side of things?

maxpoulin64:
It was fixed recently in irc-fw master, but I haven't had the occasion to retest it yet. But in theory yes, it's fixed!

astorije:
Exciting! Can't wait to try this!

xPaw:
This is ready to be merged, @thelounge/maintainers.

YaManicKill:
Well, the code looks good. I'm happy to test it if you still think it needs some testing, equally I'm happy for it to go in (Especially as we'll be releasing to a beta/rc first). So :+1: for it. Obviously the testing is quite awkward trying to get it to disconnect from something to reconnect (Especially as I use znc).

src/plugins/irc-events/connection.js
```diff
@@ -19,9 +19,9 @@ module.exports = function(irc, network) {
 		}));
 	});
 
-	irc.on("socket close", function() {
+	irc.on("close", function() {
 		network.channels[0].pushMessage(client, new Msg({
-			text: "Disconnected from the network."
+			text: "Disconnected from the network, and will not reconnect."
```
astorije: Is there any addition `reason` or something we can give to the user? It seems very frustrating otherwise.

astorije:
I'll test this very soon!

src/plugins/irc-events/connection.js
xPaw: Previous messages in log should give context (e.g. failed reconnects).

src/plugins/inputs/connect.js
```diff
@@ -1,8 +1,24 @@
+var Msg = require("./models/msg");
```
maxpoulin64: This doesn't work and crash on start. Needs changing to `../../models/msg`.

maxpoulin64:
Just tested this, and it finally works! 🎉 The only thing I see that needs fixing would be the comment I just added because it crashes on start, and then I think we're finally good to ship this!

src/plugins/inputs/connect.js
astorije: Indeed, see https://github.com/thelounge/lounge/blob/xpaw/auto-rc/src/plugins/inputs/query.js#L3.
Fixed in b9a0ee0.

astorije:
@maxpoulin64, fixed the issued you noticed in this PR.
I'll do some tests of the overall PR tomorrow before merging!

xPaw:
Thanks @astorije, unsure how I missed that.

astorije:
It's actually fairly hard to get disconnected from Freenode on purpose! I don't know what black magic is that, but when running The Lounge locally, turning wifi, turning it on a minute or 2 later and refreshing the page, messages posted during the time off wifi appeared, even though the server was running locally. Can't explain this too well, and it might be the consequence of a few things together (IRCv3, irc-fw, some flags on Freenode, I don't know...).

However, on a network that is not that smart, I got:

screen shot 2016-05-24 at 00 06 43

The socket error is when I shut down wifi. What this shows is that when reconnecting, it tried with the same username as before, but since the nick has not quit properly, it was still hanging around and the server refused to connect properly.

Is there something that can be done here, such as trying with a different nick (old_nik+random_int)? Or is letting the reconnect fail until it succeeds good enough?

xPaw:
Remember when we argued about nick changes before?

It currently only tries to change the name if it's `irc.connection.registered)`.

It does look like it let you on the server after all though?

astorije:
Yeah, definitely not an obvious fix.
At the very least, this PR reconnects fine, and as you said in your original message, it did reconnect, so it would be transparent overnight for example.
Let's merge this and improve edge cases over time.
👍 👏 👏

PR #256, opened 2016-04-16
Add support for oidentd spoofing

maxpoulin64:
This adds support for oidentd in order for server admins to properly report who is connecting and prevent server-wide bans.


xPaw:
Should the identd file be cleared out or deleted when lounge is shut down?

defaults/config.js
```diff
@@ -281,5 +281,15 @@ module.exports = {
 		// @default  113
 		//
 		port: 113
-	}
+	},
+
+	//
+	// Enable oidentd support using the specified file
+	//
+	// Example: oidentd: (process.env.HOME || process.env.USERPROFILE) + "/.oidentd.conf"
```
xPaw: Does it explode if you use `"~/.oidentd.conf"`?

maxpoulin64: Yes, unfortunately. I don't know if there's a way to have it expand that properly =/

maxpoulin64:
> Should the identd file be cleared out or deleted when lounge is shut down?

That would be the cleaner option, but I don't know if we have a somewhat decent way to know we're exiting. Technically if the client disconnected all networks properly, the file would be emptied as it deletes obsolete lines.

It shouldn't have a negative side-effect unless someone runs another IRC client as the same user as The Lounge ran, connect to the same network and have the bad luck of reusing the same source port.

xPaw:
We should also probably have some safe guard for combining identd with oidentd options, only one should be allowed, I think.

xPaw:
:+1:

defaults/config.js
williamboman: Well of course https://github.com/sindresorhus/untildify :laughing:

src/server.js
```diff
@@ -39,6 +40,10 @@ module.exports = function(options) {
 	}
 
 	if ((config.identd || {}).enable) {
+		if (manager.identHandler) {
+			console.warn("Warning: using both identd and oidentd at the same time.");
```
xPaw: Change to `log.warn` which is now in master.

src/oidentd.js
```diff
@@ -0,0 +1,72 @@
...
+				+ " { reply \"" + connection.user + "\" }\n";
+		}
+
+		for (var id in this.connections) {
+			file += makeRule(this.connections[id]);
+		}
+
+		fs.writeFile(this.file, file, {flag: "w+"}, function(err) {
+			if (err) {
+				console.warn("Failed to update oidentd file!", err);
```
xPaw: Change to `log.error` which is now in master.

defaults/config.js
xPaw: Now that it's fixed, we can have the example say `oidentd: "~/.oidentd.conf"`

maxpoulin64: Fixed.

src/plugins/irc-events/connection.js
```diff
@@ -25,6 +22,24 @@ module.exports = function(irc, network) {
...
+		});
+	}
+
+	if (identHandler) {
+		irc.on("socket connected", function() {
+			identHandler.addSocket(irc.connection.socket, client.name || network.username);
+			identHandler.refresh();
+		});
+
+		irc.on("socket close", function() {
```
xPaw: This event isn't being raised in irc-fw.

astorije:
If @YaManicKill doesn't take second review of this within a few days, I'll take a look.

src/identd.js
```diff
@@ -46,3 +51,4 @@ function parse(data) {
 	data = data.split(",");
 	return parseInt(data[0]) + ", " + parseInt(data[1]);
 }
+
```
xPaw: Get rid of this empty line when you rebase :see_no_evil:

maxpoulin64:
> If @YaManicKill doesn't take second review of this within a few days, I'll take a look.

Any news on this?

YaManicKill:
I'm not sure I understand this enough, and I certainly won't have time to look at it.

xPaw:
I guess we're fine to merge this then, since no one else wants to review it.

astorije:
> Any news on this?

Sorry, I completely dropped the ball on this. I'll give a review tonight or tomorrow night.

src/server.js
```diff
@@ -40,6 +41,10 @@ module.exports = function(options) {
 	}
 
 	if ((config.identd || {}).enable) {
+		if (manager.identHandler) {
+			log.warn("Warning: using both identd and oidentd at the same time.");
```
astorije: Do we need to specify `Warning: ` if this is already a `log.warn`? Also, considering the content, I'd go for an info rather than a warning.

astorije:
I haven't run this, but went through a full tour by @maxpoulin64 and I'm good with the code: 👍

astorije:
Now, @maxpoulin64, do you think this should be reflected in the doc somewhere: https://thelounge.github.io/docs/server/configuration.html ?

Of course, the docs really need some love, but I'm talking short term rather than a bigger-commitment doc refactor unicorn :-) (smells like `up for grabs`!).

PR #259, opened 2016-04-16
Always run all tests

maxpoulin64:
Fixes a small annoyance that you don't know if, for example, stylelint would return any errors until eslint comes all clean. This makes it so all tests are ran, and only return failure at the end.

I had to add a small script because the only ways to make this work with normal shell commands is kinda ugly. This makes it so commands are simply separated by the `+` character, and in the event someone totally needs to use `+` in a command, `\+` can be used instead to insert a literal `+` in the command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
   "scripts": {
     "start": "node index",
     "build": "grunt && handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
-    "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",
-    "lint": "eslint . && stylelint \"**/*.css\"",
+    "test": "node run-all HOME=test/fixtures mocha test/**/*.js + npm run lint",
+    "lint": "node run-all eslint . + stylelint '**/*.css'",
     "prepublish": "npm run build"
   },
   "keywords": [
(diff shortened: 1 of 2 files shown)


xPaw:
Extra info:
> 16:32	+xPaw	what about using concurrently?
16:34	+Max-P	I thought about it, it would be doable and would speed up the tests a lot, but then the output will be broken or we'll have to buffer it and spit it all out at the end


> 16:35	+Max-P	|| will have the same issue as &&, and ; doesn't keep the exit code so if the first test fails but not the second then npm thinks it's all good

package.json
```diff
@@ -13,8 +13,8 @@
   "scripts": {
     "start": "node index",
     "build": "grunt && handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
-    "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",
-    "lint": "eslint . && stylelint \"**/*.css\"",
+    "test": "node run-all HOME=test/fixtures mocha test/**/*.js + npm run lint",
+    "lint": "node run-all eslint . + stylelint '**/*.css'",
```
xPaw: Change this back to `\"**/*.css\"` otherwise it breaks on Windows.

maxpoulin64:
I found a much better way, closing this.

PR #260, opened 2016-04-16
Improve tests execution

maxpoulin64:
As explained in #259, the current way of running the linters makes it so that if one of them errors out, then the next one is never ran. This causes the situation where you are forced to fix all eslint errors before you can see stylelint errors.

This migrates both to mocha tests, so we now have one single test command to run to deal with everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,9 @@ deploy:
     node: '4'
     tags: true
     repo: thelounge/lounge
+
+script:
+- npm run test:mocha
+- npm run lint:js
+- npm run lint:css
+
--- a/package.json
+++ b/package.json
@@ -12,9 +12,14 @@
   },
   "scripts": {
     "start": "node index",
-    "build": "grunt && handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
-    "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",
-    "lint": "eslint . && stylelint \"**/*.css\"",
+    "build": "npm run build:grunt && npm run build:handlebars",
+    "build:grunt": "grunt",
+    "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
+    "test": "npm run test:mocha; npm run lint",
+    "test:mocha": "mocha -r test/fixtures/env.js test/**/*.js",
+    "lint": "npm run lint:js; npm run lint:css",
+    "lint:js": "eslint .",
+    "lint:css": "stylelint \"**/*.css\"",
     "prepublish": "npm run build"
   },
   "keywords": [
--- a/test/fixtures/env.js
+++ b/test/fixtures/env.js
@@ -1,0 +1,1 @@
+process.env.HOME = "test/fixtures";


xPaw:
:+1:

test/plugins/link.js
```diff
@@ -1,3 +1,4 @@
+process.env.HOME = "test/fixtures";
```
xPaw: Note: Moving `HOME` here fixes running tests on Windows.

astorije:
Actually, unless I misunderstood the intent, there is a simpler and more elegant way to do so, with less boilerplate code:

1. Break the npm scripts in multiple scripts:

  ```json
  "lint:css": "stylelint [...]",
  "lint:js": "eslint [...]",
  "test":  "mocha [...]"
```
2. Then add a `script` section to `.travis.yml`:

  ```yaml
script:
  - npm test
  - npm run lint:js
  - npm run lint:css
```

And that's it!
[All commands in `script` are run, regardless of the return code of the others](https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Build-Step):

> When one of the build commands returns a non-zero exit code, the Travis CI build runs the subsequent commands as well, and accumulates the build result.

Now, we might still want a single command to run everything with one command, which is something your PR offers. 2 solutions to that:
- Add an extra script for all checks: `"all": "npm test && npm run lint:js && npm run lint:css"`.
- Or rename `"test"` script into `"test:js"` or `"test:mocha"`, change the first command to `npm run test:(js|mocha)` in the `.travis.yml` file and have the `test` script run them all: `npm run test:(js|mocha) && npm run lint:js && npm run lint:css`.

(I prefer the second solution).

Granted, as opposed to your PR, all this makes `npm test` fail locally as soon as one of the subcommand fails, but checks would be run on the CI regardless of the return code of other checks,. Solving it on the CI is far more important than locally IMO, and the simplicity of this vs. having boilerplate scripts for this justifies the trade-off.

What do you guys think?

maxpoulin64:
> Granted, as opposed to your PR, all this makes npm test fail locally as soon as one of the subcommand fails

Well that's the entire reason why I wanted to make this work in the first place. Sometimes I have multiple things going on at the same time, and a few times I opened a PR thinking my CSS was fine and thought I was all good because I only got errors in unrelated JS code, and then Travis yelled at me because of pages of CSS errors.

YaManicKill:
Would it not be simplest to just change it to this:

```
"test": "HOME=test/fixtures mocha test/**/*.js; npm run lint",
 "lint": "eslint .; stylelint \"**/*.css\"",
```

This will both commands whether the first fails or not, and in serial. There is only 1 disadvantage of this, and that's that windows doesn't support ;, but realistically, most people developing on windows are at least using git bash, if not cygwin.

It's much simpler, and makes it much more easy to figure out what's happening. Having mocha actually linting things seems... awkward.

xPaw:
@YaManicKill using `;` does not report correct exit code if one of the two linters fail.

YaManicKill:
True, but that's only an issue for travis, and we can get travis to run the scripts individually, as suggested by @astorije, so that it gets the correct exit codes.

I don't really like this really complex way of dealing with running the linter.

astorije:
I fully agree with @YaManicKill here. As a matter of fact, I have been using this setup on another project of mine, and using `;` has never been an issue locally: see [`package.json`](https://github.com/astorije/chai-immutable/blob/23610f73528679fd7e2715e871c32929b07f48bf/package.json#L10-L13) and [`.travis.yml`](https://github.com/astorije/chai-immutable/blob/master/.travis.yml#L11-L14).
Definitely much simpler than wiring up API calls, lets you locally run tests only or style checker only, and lets Travis CI run and report on all 3 checks independently.
Plus it complies with @maxpoulin64's requirements of having `npm test` fail locally for one/some of the checks without affecting the others.

@maxpoulin64, are you OK with this solution?

maxpoulin64:
@astorije Alright, like that for the `.travis.yml`?

It's not super great (it would be nice to have `npm run tests` return an appropriate failure number) but it does fix the issue I had and at least it cleans up the tests a little bit.

package.json
```diff
@@ -13,8 +13,11 @@
   "scripts": {
     "start": "node index",
     "build": "grunt && handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
-    "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",
-    "lint": "eslint . && stylelint \"**/*.css\"",
+    "test": "HOME=test/fixtures mocha test/**/*.js",
```
astorije: Is it possible to keep `HOME=test/fixtures` in the according test to keep [this](https://github.com/thelounge/lounge/pull/260#discussion-diff-59969310)?

package.json
```diff
@@ -13,8 +13,11 @@
...
+    "lint": "npm run lint:js; npm run lint:css",
+    "lint:js": "eslint .",
+    "lint:css": "stylelint \"**/*.css\"",
+    "test-all": "npm run lint; npm run test",
```
astorije: I'd rather have `test` run everything so that new contributors who are not aware of `test-all` still run the linters:

```js
"test": "mocha test/**/*.js; npm lint:js; npm run lint:css"
```

With a preference to end with the linters because they are silent when they don't trigger an error.

YaManicKill: Yeah, I agree with this. Once this is done, I'll give my ok to this. Looks good.

maxpoulin64:
@astorije @YaManicKill All comments should have been addressed now.

.travis.yml
```diff
@@ -18,3 +18,9 @@ deploy:
     node: '4'
     tags: true
     repo: thelounge/lounge
+
+script:
```
xPaw: I wonder, perhaps we should be invoking build commands too? They could potentially fail.

astorije: They would fail at `npm install` time. I think this is good as is for now, we can improve later but this PR already solves @maxpoulin64's (and probably others') issue.

xPaw: Hmm, it does run it and hides the output. Do we know if travis fails the build if grunt/handlebars build fails at that stage?

YaManicKill:
Alrighty, I'm happy with this now. :+1: 

Can't see confirmation that @astorije or @xPaw are happy that their comments have been addressed, so will leave the merge till they respond.

xPaw:
:ship:

xPaw:
Damn. I should have tested the latest version on Windows.

Using `;` fails as: `missing script: lint:js;`, as if it's putting `;` as part of the script name...

PR #266, opened 2016-04-19
Create a single helper function to write messages

### LINKED ISSUE #232: Do not keep messages in memory for public users
There's simply no need to as the session is destroyed when the user closes the page, and it will fix the memory *(for public servers)* leak.

xPaw:
This solves a couple of issues:

- Certain messages were not stored in memory, so reloading the page would not print them on the client (e.g. connection related messages)
- `maxHistory` is *always* applied, not just when receiving messages in channels
- Fixes #232: Do not store messages in memory when running in public mode
- Single consistent function increases readability (and less boilerplate code)

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -8,14 +8,6 @@ program.version(pkg.version, "-v, --version");
 program.option("");
 program.option("    --home <path>" , "home path");
 
-require("./start");
-require("./config");
-require("./list");
-require("./add");
-require("./remove");
-require("./reset");
-require("./edit");
-
 var argv = program.parseOptions(process.argv);
 if (program.home) {
 	Helper.HOME = program.home;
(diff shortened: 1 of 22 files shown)


src/models/chan.js
```diff
@@ -9,6 +10,7 @@ Chan.Type = {
 };
 
 var id = 0;
+var config = Helper.getConfig();
```
maxpoulin64: This breaks the `--home` command line switch as the config gets loaded when the models are `require()`'ed so before the command line switches can be processed.

src/command-line/index.js
```diff
@@ -32,6 +24,14 @@ if (!fs.existsSync(config)) {
 	console.log(config);
 }
 
+require("./start");
```
xPaw: Moved these requires to fix @maxpoulin64's problem below.

maxpoulin64:
Looks fine to me 👍

astorije:
Great stuff, @xPaw! 👍

PR #267, opened 2016-04-19
Fade client on disconnect instead of refreshing.

unconed:
When you lose your internet connection, the client tries to refresh itself. This instantly results in an error page and forces you to refresh manually. With this patch, the body just dims instead, and it auto-reconnects whenever it can.

It still does a refresh if it succeeds, for reasons I don't feel like sussing out, but at least you can now leave the client open 24/7 on a moving laptop without having to babysit it.


dgw:
This is an annoyance I hadn't gotten around to posting an issue report for yet, so :+1:.

client/js/lounge.js
```diff
@@ -51,6 +51,10 @@ $(function() {
 	$(".tse-scrollable").TrackpadScrollEmulator();
 
 	var favicon = $("#favicon");
+
```
dgw: Unnecessary tab character

client/js/lounge.js
```diff
@@ -65,10 +69,14 @@ $(function() {
 	socket.on("error", function(e) {
 		console.log(e);
 	});
+
```
dgw: Unnecessary tab character

client/js/lounge.js
```diff
@@ -65,10 +69,14 @@ $(function() {
 	socket.on("error", function(e) {
 		console.log(e);
 	});
+	
+	socket.on("connect", function (e) {
+		dimmer(false);
+  	});
```
dgw: Mixed spaces & tabs

client/js/lounge.js
```diff
@@ -65,10 +69,14 @@ $(function() {
 	socket.on("error", function(e) {
 		console.log(e);
 	});
+	
+	socket.on("connect", function (e) {
```
dgw: Unused variable `e` triggers linter error

client/js/lounge.js
```diff
@@ -51,6 +51,10 @@ $(function() {
 	$(".tse-scrollable").TrackpadScrollEmulator();
 
 	var favicon = $("#favicon");
+	
+	function dimmer(dim) {
+		document.body.style.opacity = dim ? 0.5 : 1;
```
maxpoulin64: Please use a CSS class instead, so themes can customize the behavior properly.

xPaw:
Can we also have a message displayed upfront saying what's going on? Otherwise it won't be really understandable what's going on.

You say

> and it auto-reconnects whenever it can.

But there is no auto reconnection logic that I know of.

YaManicKill:
> You say

> > and it auto-reconnects whenever it can.

> But there is no auto reconnection logic that I know of.

Socket.io does that by default, I'm pretty sure.

> Can we also have a message displayed upfront saying what's going on? Otherwise it won't be really understandable what's going on.

xPaw:
Ah you are right, but it still can stop at some point. We should be handling extra events to display all the necessary information to the user.
http://socket.io/docs/client-api/#manager(url:string,-opts:object)

YaManicKill:
At this point, would it not be enough to say something like "Disconnected, trying to reconnect. Refresh if this takes a while or doesn't work." or something like that? I can't see an event that means they give up, so do we want to show them all the messages from all the events? What would you suggest?

xPaw:
timeout option makes it sound like there's a max retry, not sure about it.

We probably would want to handle all the connection related events and just display the latest one, that seems like the simplest way of doing about it.

maxpoulin64:
So, apart from the code style errors I pointed out already, this appears to be working fine. Not the cleanest, as it just reloads the browser window once it's able to reconnect properly, but I'll admit it looks cleaner than sitting at the "This website is not accessible" error page. Considering the proper fix is a huge amount of work, I would +1 this as long as the code style issues are fixed. This is essentially what Chrome already does, without the ugliness of the error page.

xPaw:
@maxpoulin64 where did you get the refresh part from? There's certainly no code to refresh the page after it reconnects. If we go this route, we need to have a variable that is set to true when disconnected, and in connected event refresh the page if that variable is true.

We do need to perform a full page refresh, otherwise we can end up in a state where some things could be carried over from previous connection.

YaManicKill:
> We do need to perform a full page refresh, otherwise we can end up in a state where some things could be carried over from previous connection.

Why? Surely, when we connect we are connected again, and can receive messages. What goes wrong with this?

xPaw:
@YaManicKill Because this code was never designed to support reconnections. It most likely can break if there are changes on the server while client was disconnected (e.g. network removed, server restarted and client token is gone). Basically what I'm saying is, it will require heavy testing before we would be able to ship this. We also don't know how this would in public mode, as the server destroys all  information about disconnected clients.

maxpoulin64:
@xPaw: [Here](https://github.com/thelounge/lounge/pull/267/commits/503d42e7583e744e15ac987973f021e93f103a5e#diff-e5178f7b74fe45f2cfe1baf9aa1ef6faR87), apparently. When it gets a reconnection somehow that fails and the client refreshes. Definitely not the right way to handle it, but considering that side-effect on reconnection the PR does work, surprisingly.

@YaManicKill: There are many things that breaks reconnection. The problem is the client here, as the server just pushes the whole state back. The client appends stuff directly to the page, so we lose the initial state. Therefore we'd have to manually undo all the changes to a clean state before we allow the client to regenerate everything, which is kind of a problem. And it would still look pretty ugly, because we'd essentially do the same thing as a refresh except make it a lot harder for us: you'd still lose the currently active window, scroll positions and many other things. At this point, better just reload it (for now).

The ideal solution for a clean reconnect would be the server to send us the diff of what we missed, so we can just catch up.

YaManicKill:
Righty, I forgot about the fact that the server would just ignore us.

Alright, so this PR should still do the refresh on reconnect, but there is still definitely a good reason to not refresh on the error, but to show a notification that we are disconnected.

@unconed Would you be up for updating your PR to:

- [ ] Fix the build
- [ ] Keep the refresh on reconnect
- [ ] Add a notification that we are currently disconnected.

Would anyone have an objection to just using something like notify.js for this, and we can imrpove things as we go along. We are too bad at not adding new features because we want to make it perfect the first time.

xPaw:
> Would anyone have an objection to just using something like notify.js for this

Simply fade the page and display a message in the center.

YaManicKill:
> Simply fade the page and display a message in the middle.

That makes it nearly impossibly to read through messages, which kinda defeats half the purpose of not refreshing the page.

xPaw:
> That makes it nearly impossibly to read through messages
Hm that's a good point. I assumed if you're disconnected from the server you would not be able to do any interactions, but reading is one.

YaManicKill:
@unconed Any plans to fix this up, at all?

unconed:
@YaManicKill I fixed the nitpicks but due to other changes since, it no longer refreshes on reconnect. The client code is a big mess so I'm going to leave that for someone else to figure out and/or refactor.

xPaw:
I was thinking this through, and I think we can actually get away without reloading the page on a reconnection. The `init` event will re-render the whole window (not desirable), and the JS doesn't have any direct arrays/objects keeping track of state, it's all in DOM.

However there are fixes needed to be done in `auth` callback to account for the missing `sign-in` window.

astorije:
Closing this as spring cleaning. @unconed's last message tells me he won't stop by here anymore, so I'll update the according issue (#421) in case someone wants to fix this and get some history.

Issue #74, opened 2016-02-19
Sent PMs don't get synced


YaManicKill:
I imagine this might be a limitation with IRC, but if you are connected to something like ZNC, and you connect with another client and send a PM to someone, it doesn't go to thelounge so you end up with a very confusing looking history.

It might not be possibly to sort out, but worth bringing up.

xPaw:
Perhaps removing this check will fix the problem? https://github.com/thelounge/lounge/blob/master/src/plugins/inputs/msg.js#L22-L23

dgw:
I forget what the capability is currently called, as it's been renamed, but adding support for the [IRCv3 `self-message` capability](http://danieloaks.net/ircv3-specs-web/specs/extensions/self-message-3.2.html) would resolve this, and likely improve bouncer playback handling in general.

astorije:
Fixed by #270.

PR #270, opened 2016-04-22
Add support for echo-message and znc.in/self-message caps

### LINKED ISSUE #269: Add support for echo-message cap
http://ircv3.net/specs/extensions/echo-message-3.2.html

Here's what needs to be done:
- Request the cap
- In input plugins that send PRIVMSG and NOTICE (this includes CTCP) check if the cap is enabled, and if so, do not push the message to memory
- Make sure `self` still works when receiving echoed messages

xPaw:
- Add support for `echo-message` and `znc.in/self-message` caps.
 - `znc.in/self-message` because latest znc version still does not have `echo-message`, and people often don't upgrade their ZNC for a long time
- Moved query/invite highlight logic from the client to the server
 - Fixes self messages received in query windows triggering a highlight

Fixes #269 and #74

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -761,8 +761,7 @@ $(function() {
 
 	chat.on("msg", ".messages", function(e, target, msg) {
 		var button = sidebar.find(".chan[data-target='" + target + "']");
-		var isQuery = button.hasClass("query");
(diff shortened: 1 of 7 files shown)


maxpoulin64:
👍 makes reconnecting to ZNC a lot more cleaner!

astorije:
👍

PR #272, opened 2016-04-24
Log enabled capabilities

xPaw:
Looks like this in the server window:
> Enabled capabilities: extended-join, multi-prefix

This is useful information to have if certain features should be working on servers or not. Other clients print it too.


maxpoulin64:
Looks good to me, 👍  and merging

PR #273, opened 2016-04-24
Only increase unread counter for whitelisted actions

### LINKED ISSUE #271: Embedded images/links count as unread messages
If someone posts a link, and you have auto-expand [thumbnails/links], the message count on the sidebar will increase.

xPaw:
Fixes #271. Only three actions whitelisted because only these actions increase the unread counter on the server, and we must follow the exact same rules.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -807,14 +807,12 @@ $(function() {
 			return;
 		}
 
-		var ignore = [
-			"join",
-			"part",
-			"quit",
-			"nick",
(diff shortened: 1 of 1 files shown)


maxpoulin64:
👍 

Note: this is the mentionned only place where it is incremented: https://github.com/thelounge/lounge/blob/master/src/plugins/irc-events/message.js#L67-L73

astorije:
👍 and merging, nice catch!

Issue #85, opened 2016-02-22
[Discussion] Towards better settings


astorije:
Lots of issues, PRs, discussions and disagreements have an impact on settings. In this fork alone, I can mention https://github.com/thelounge/lounge/pull/57, https://github.com/thelounge/lounge/pull/70, https://github.com/thelounge/lounge/pull/28, https://github.com/thelounge/lounge/pull/49, https://github.com/thelounge/lounge/issues/80, https://github.com/thelounge/lounge/issues/44, https://github.com/thelounge/lounge/pull/70, ... and we are only 2-week old!

This is only going to get worse as we improve theme management (or *create* theme management for that matter) and add packages, who will have their own settings, potentially affect existing settings, ...

Before we start adding dozens of settings to the current setting page, we need to discuss the long-term goals.

I have been building a [mockup of the settings page slightly revisited](https://dl.dropboxusercontent.com/u/3400500/mockup_thelounge_settings/index.html). I propose we discuss its style, content, interactions, underlying structure, usability, ... (but the CSS/HTML is gross, this is just a mockup :P)
I will update the mockups as the discussion evolves.

One very important aspect is to keep user-friendliness as we evolve, or we will end up being a cluttered application as most IRC clients are. Defaults should be sensitively chosen to enhance the experience of any new user (of The Lounge *or* IRC altogether). Considering a private instance of The Lounge can be maintained by one administrator and used by many users, it's not crazy to think The Lounge may be used to introduce new folks to the IRC world :-)
Also, not everything should be a setting ("Better settings" does not mean "more and more settings" :p). Any UI-specific choice should ideally be the responsibility of a theme author, and may be overridden using custom styling.
However, advanced and very advanced users should be able to find happiness here too. When it comes to styling, a custom style field should give them the ability to make changes to a very fine-grained extent. We might want to be able to do the same thing for the settings overall: if all interactive elements for settings (checkbox, inputs, ...) match to an underlying data structure, some settings might be so advanced that they should be edited in a text field directly (like the custom styling, but for settings themselves). This might be an impractical idea after discussion, but focusing on keeping the UI slick and simple is key here.

[...]

PR #275, opened 2016-04-25
Settings overhaul

maxpoulin64:
This redoes pretty much everything related to client-side settings and unties them from the DOM. Most of the action happens in the `Settings` class in preparation for a future client-side API, which makes listening to setting changes a lot easier. Setting handlers never touches or sees the DOM.

It also makes the binding glue between the settings and the DOM work the other way, so invisible settings are possible and can be tied to multiple controls or sources at once. This could be used later, for example, by a `/set` command or for any other reason.

Communication with the server is also made separate for good measures, so the settings are usable without any active socket.

A similar API is exposed on the server through `client.settings`, with the same separation of concerns, and will update all clients. These settings can be further used by the server to save preferences server-side, such as an highlight word list, auto away on/off, and many other things.

Obviously, it comes with a new settings page on the client.

---

This doesn't exactly solve #85 as the focus of this PR wasn't to fix the settings UI, but to provide the backend infrastructure for server settings so we're not stuck waiting for design just to implement features. It just happened that I had to redo the frontend to both to avoid unnecessary pain plugging it back in, as well as showcase the thing in action.

---

Suggestions welcome. I think it's really pretty good so far, and allows for a lot of future customization.

capture d ecran_2016-04-24_23-16-49

Demo video: https://d.max-p.me/irc/settings.mp4

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -169,143 +169,126 @@
 						<button class="lt"></button>
 					</div>
 					<div class="container">
-						<div class="row">
-							<div class="col-sm-12">
-								<h1 class="title">Settings</h1>
-							</div>
-							<div class="col-sm-12">
-								<h2>Messages</h2>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="setting:join">
-									Show joins
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="setting:motd">
-									Show motd
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="setting:part">
-									Show parts
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="setting:nick">
-									Show nick changes
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="setting:mode">
-									Show mode
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="setting:quit">
-									Show quits
-								</label>
-							</div>
-							<div class="col-sm-12">
-								<h2>Visual Aids</h2>
-							</div>
-							<div class="col-sm-12">
-								<label class="opt">
-									<input type="checkbox" name="setting:colors">
-									Enable colored nicknames
-								</label>
-							</div>
+						<div class="col-sm-12">
+							<h1 class="title">Settings</h1>
+						</div>
+						
+						<table class="settings">
+							<tr>
+								<th class="section-title">Messages</th>
+								<th class="icon" title="Enabled"> &#xf046; </th>
+								<th class="icon" title="Synchronize to server"> &#xf0c2; </th>
+							</tr>
+							<tr>
+								<td> <label for="setting:motd">Show motd</label> </td>
+								<td> <input type="checkbox" id="setting:motd" name="setting:motd"> </td>
+								<td> <input type="checkbox" id="sync:motd"    name="sync:motd"> </td>
+							</tr>
+							<tr>
+								<td> <label for="setting:join">Show joins</label> </td>
+								<td> <input type="checkbox" id="setting:join" name="setting:join"> </td>
+								<td> <input type="checkbox" id="sync:join"    name="sync:join"> </td>
+							</tr>
+							<tr>
+								<td> <label for="setting:part">Show parts</label> </td>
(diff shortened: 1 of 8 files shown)


src/server.js
```diff
@@ -173,8 +173,11 @@ function init(socket, client, token) {
 		socket.emit("init", {
 			active: client.activeChannel,
 			networks: client.networks,
-			token: token || ""
+			token: token || "",
+			settings: client.settings.options,
 		});
+
+		client.settings.bindToSocket(socket);
```
xPaw: This shouldn't be bound in public mode.

xPaw:
We probably want to namespace the settings. As settings like hiding joins, hiding quits, notifyAllMessages ideally would be set per channel.

So the current settings window would be `global`, and settings in each channel would overrule the channel ones.

client/js/lounge.js
```diff
@@ -410,77 +412,235 @@ $(function() {
...
+			el.change(function() {
+				settings.set(name, el.prop("value"));
+			});
+			settings.on("change:" + name, function(event, value) {
+				el.prop("value", value);
+			});
 		}
 
-		window.localStorage.setItem("settings", JSON.stringify(options));
+		settings.element.find("input[name^='setting:'], textarea[name^='setting:']").each(function() {
```
xPaw: Why not just use event delegation?

client/index.html
```diff
@@ -169,143 +169,126 @@ <h1 class="title">Connect</h1>
...
+								<td> <label for="setting:notifyAllMessages">Enable notification for all messages</label> </td>
+								<td> <input type="checkbox" id="setting:notifyAllMessages" name="setting:notifyAllMessages"> </td>
+								<td> <input type="checkbox" id="sync:notifyAllMessages"    name="sync:notifyAllMessages"> </td>
+							</tr>
+						</table>
+						
+						<div class="col-sm-12">
+							<h2>
+								Custom Stylesheet
+								<% if (!public) { %>
```
xPaw: Only this sync checkbox is hidden in public mode, that's not quite consistent. Could we just set `disabled` on all the sync checkboxes, and perhaps even add a tooltip saying syncing is not allowed in public?

maxpoulin64: It's actually hidden [here](https://github.com/thelounge/lounge/pull/275/files#diff-97db1f70168fb5f12457b238ff6052b5R1102). I had put a bunch of `<% if (!public) { %>` around them, but it quickly turned kinda ugly.

xPaw: I see. That last-child selector is rather ugly, add a class on `td`, and same class on `sync-header`, then you can hide both in the same way.

client/js/lounge.js
maxpoulin64: I could but since this is a two-way sync, I'd need to keep querying and checking the type of element for every setting change. I thought it was clearer that way, especially since we don't have that many boxes to bind.

src/server.js
maxpoulin64: Actually yes. It's not currently needed, but I prefer having the functionality than it being a limitation later on. As I mentioned in the PR text itself, I'm aiming for API-ready code so I try to avoid putting arbitrary limitations. There are a few limited cases in which having the server aware of some settings could be useful. It's not used strictly for settings persistance, but also just for the server to share settings in general. I'm thinking `/ignore` support, auto away and others.

There's a (one-time) check [here](https://github.com/thelounge/lounge/pull/275/files#diff-a7181c2b6058f7bc955aa13d1e2a1befR57) to not persist settings in public mode. I'm using `client.name` so if we eventually allow both public and private mode at the same time, it's ready.

xPaw: For mixed public/private mode we should have some better check other than `client.name` then.

xPaw:
Just tested this locally, works fine, great work there!

Only suggestion I have is to enable synchronization by default.

client/index.html
```diff
@@ -169,143 +169,126 @@ <h1 class="title">Connect</h1>
...
+							<tr>
+								<td> <label for="setting:notifyAllMessages">Enable notification for all messages</label> </td>
+								<td> <input type="checkbox" id="setting:notifyAllMessages" name="setting:notifyAllMessages"> </td>
+								<td> <input type="checkbox" id="sync:notifyAllMessages"    name="sync:notifyAllMessages"> </td>
+							</tr>
+						</table>
+						
+						<div class="col-sm-12">
+							<h2>
+								Custom Stylesheet
```
xPaw: Make this a `label` and target it at the textarea please.

YaManicKill:
What's the deal with this? Are we planning on getting this updated to master? I think this is a sorely missed feature that would be fantastic to get in.

xPaw:
I will close this PR for now. I've moved the `settings` branch into this repository and it needs a major rebase to continue work.

PR #279, opened 2016-04-26
Do not count your own messages as unread

xPaw:


src/plugins/irc-events/message.js
```diff
@@ -64,7 +64,7 @@ module.exports = function(irc, network) {
 			return (w.replace(/^@/, "").toLowerCase().indexOf(irc.user.nick.toLowerCase()) === 0);
 		});
 
-		if (chan.id !== client.activeChannel) {
+		if (!self && chan.id !== client.activeChannel) {
```
xPaw: This is required when using a bouncer.

astorije:
Sweet! 👏

PR #280, opened 2016-04-26
Document supported node version

### LINKED ISSUE #231: Document which node versions are supported.
703-imported-1453533598-schedule

Looking at the picture, node 0.12 is no longer an active LTS as of this month. We already removed 0.10 from Travis due to failing linters. Looks like there are problems running 0.10 on `irc-framework` branch too.

We can also use `engines` field in `package.json` to make it more clear for users: https://docs.npmjs.com/files/package.json#engines

xPaw:
Fixes #231.

`engines` field only affects new installations, and only emits a warning like this:
```
npm WARN engine thelounge@1.5.0: wanted: {"node":">=8.2.0"} (current: {"node":"5.4.1","npm":"3.3.12"})
```

It does not block installation. I documented 4.2.0 because it's the lowest LTS version, as seen here: https://nodejs.org/en/download/releases/

It only took about a month to release 4.2.0 after 4.0.0, so it's not going to be a big problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: node_js
 node_js:
-- '5.6'
-- '4.3'
-- '0.12'
+- '6'
+- '5'
+- '4'
+- '4.2.0'
 notifications:
   email:
     on_success: never
@@ -14,6 +15,6 @@ deploy:
   api_key:
     secure: I9iN31GWI+Mz0xPw81N7qh1M6uidB+3BmiPUXt8QigX45zwp9EhvfZ0U/AIdUyQwzK2RK1zLRQSt+2/1jyeVi+U+AAsRRmaAUx8iqKaQPAkPnQtElolgRP04WSgo7fvNejfM7zS939bQNKG3RlSm04yPgu+ke2igf799p2bpFe2LtyoEeIiUfrUkBiMSpMguN9XF8a7jqCyIouTKjXHR24RmzJ9r7ZoMV27yQauS7XlD81bontzNRZxTytDKdJpZ+sxGIT9mbbtM4LUFX8MeNe3p/bjWavEhrO0ZIpkbOfS/L/w1375YDoNPXxCs288lnGUH+NbGNAEfn+BTz8cmUp7jI7QWR/kNACPeopdAX4OdZxT8wfQcfQZrfCuSpKciOMC7vGgPpQqjQ61t1RKcKs9VUnwC0SwWjyo8LlzkFKnP1ks0eDGYsSoPLdpC9+76UmePkQdxMhscO8TOgkOCcsTMLiyt6ABGOGKu2iE5SsjUYtPiSiRzSBAQENoO560+xBSVTKwqvvhzUAIt4AuAQSgsFjAylDdyzKoObHX12hBdALrqSOOSVwwIQ5/jTgNAsilURHo7KPD407PhRnLOsvumL0qg4sr9S1hjuUKnNla5dg9GY8FVjJ+b2t0A2vgfG1pR1e3vrJRXrpkfRorhmjvKAk2o5you5pQ1Itty7rM=
   on:
-    node: '5.6'
+    node: '4'
     tags: true
     repo: thelounge/lounge
--- a/README.md
+++ b/README.md
@@ -35,6 +35,12 @@ the review process quicker and more streamlined.
 
 ## Install
 
+To use The Lounge you must have [Node.js](https://nodejs.org/en/download/) installed.
+The oldest Node.js version we support is 4.2.0.
+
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -1,8 +1,7 @@
 language: node_js
 node_js:
-- '5.6'
-- '4.3'
-- '0.12'
+- 'node'
```
xPaw: Uses latest stable

maxpoulin64:
> It does not block installation.

Perfect. Not limiting for us, but people are warned it's not tested/supported yet they still can run it if they really want. 👍

README.md
```diff
@@ -35,6 +35,12 @@ the review process quicker and more streamlined.
 
 ## Install
 
+To use The Lounge you must have [Node.js](https://nodejs.org/en/download/) installed.
+The oldest Node.js version we support is 4.2.0.
```
astorije: What happened before 4.2.0 (and > 0.12) so that we don't support them?
I think that in the future, we could get away with supporting current stable and LTS proposed by the Node.js website, what do you think?

.travis.yml
astorije: I'd rather see more specific versions, like latest 4.x, 5.x and 6.x, no?

package.json
```diff
@@ -38,6 +41,7 @@
     "moment": "2.12.0",
     "read": "1.0.7",
     "request": "2.69.0",
+    "semver": "5.1.0",
```
astorije: So we are emitting a warning at install time, documenting this in the README, maybe we'll document in the website soon after, do we really need to warn when starting the server too?
I mean at this point, if people don't read docs...
I would prefer to not have an extra dependency for those who do not upgrade their Node.

Of course, I won't be blocking this if you think we reeeally should, it's just my 2c. :)

README.md
xPaw: io.js happened. 4.0 was next release after 0.12. The issue discusses why we should drop the support for 0.x.

package.json
xPaw: We're already pulling `semver` as part of some other dependency.

Runtime warning helps notify server owners, I guarantee you some of them won't be reading changelogs.

README.md
astorije: Yeah, I meant <= 4.2 and > 4.0. Why supporting 4.2 as oldest version and not 4.0 is a better question, actually.

xPaw: > It only took about a month to release 4.2.0 after 4.0.0.

4.2.0 and up is the LTS.

YaManicKill: Where did you get this from? 4.x is the LTS branch, and the newest version of it is the most up to date version of it. Being LTS means that they will provide security updates to it, and bug fixes. That means that each indibidual point release is not "LTS", but the branch itself is LTS.

We aren't going to update this every time there's a new patch to 4.x, so just stick it to 4.0

xPaw: See https://nodejs.org/en/download/releases/ and https://github.com/nodejs/node/blob/v4.2.0/CHANGELOG.md

> Version 4.2.0 'Argon'
> The first Node.js LTS release! See https://github.com/nodejs/LTS/ for details of the LTS process.

YaManicKill: That's just because the April releases aren't LTS until 6 months after release (to allow for the small bugs to be eeked out).

xPaw: 4.4.3 released on 2016-04-12 is LTS though.

YaManicKill: That's not what I meant. The major releases, not any release in April.

I think you misunderstand how LTS works. The point of LTS is that the release branch will be updated with bug fixes and security updates.

4.4.3 isn't LTS. That doesn't mean anything. 4.4.3 is, and always will be, exactly the same code. It will never change. So how can there be any support for it?

What is LTS is the 4.x release branch. They word it weirdly in node, but the point is that 4.x will get bug fixes and security updates until it is no longer LTS (next April).

When a major April release happens (4.x was different in that it was released in September, but from now on all the LTS releases will be from April major releases), 6 months later it becomes LTS, and is the stablest of the stable that they have, and will be updated with bug fixes and security updates for 18 months.

So, it doesn't mean anything to say "4.2.0 is LTS" or "4.4.3 is LTS". No, 4.x is LTS. That's all.

YaManicKill:
I'm happy with this now. I'm not sure there's an advantage to saying >=4.2.0, but I'm not going to argue too mcuh.

I won't merge this, because @astorije didn't sound very happy with us giving an error, so I'll let him decide if he cares enough about it :-)

astorije:
> I won't merge this, because @astorije didn't sound very happy with us giving an error, so I'll let him decide if he cares enough about it :-)

Not about the error itself but with the fact that we have one more dependency just for that (at our level, regardless of what dependencies themselves have in their dependencies.

That being said, everyone is 👍 so I am too, and hopefully we'll get to remove that later :-)

PR #284, opened 2016-04-27
Add support for ~ home folder expansion

maxpoulin64:


src/helper.js
```diff
@@ -1,10 +1,29 @@
...
+
+	if (os.homedir) {
+		home = os.homedir();
+	}
+
+	if (!home) {
+		home = process.env.HOME || "";
+	}
+
+	home = home.replace("$", "$$$$");
```
astorije: Why is this necessary? Escaping? Why 4 of them lol? Would have `\$` work too (fairly common when escaping stuff)?

xPaw: Because `$$$$` escapes into `$$`, which is required for the next replace to keep the final `$`.

See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter

astorije: Right, I missed that part, thanks a lot for the details @xPaw :-)

PR #286, opened 2016-04-27
Display unhandled numerics on the client

xPaw:

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -914,6 +914,10 @@ button {
 	color: #f00;
 }
 
+#chat .unhandled .from {
+	color: #eee;
+}
+
 #chat .msg.toggle .time {
 	visibility: hidden;
 }
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -220,7 +220,7 @@ $(function() {
 		}
 
 		var chan = chat.find(target);
-		var msg;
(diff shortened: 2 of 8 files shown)


xPaw:
Ready after #449

astorije:
Reposting here after commenting on the wrong PR. Wow.

---

This is a nice addition, thanks!

One comment I have is: couldn't we wrap this behind a user setting, similar to the MOTD one? Not sure what name we could give it, but if it's a generic info checkbox, I would also put the "Enabled capabilities" there (if possible, of course).

Also (unrelated to this PR, but displayed by it), I noticed a line saying `[403] Freenode No such channel`, are we trying to join the lobby somewhere when first connecting?

xPaw:
@astorije I don't think a setting for messages shown in server window is worth it.

astorije:
Well, we'll have to deal with server window at some point still, because after this PR, we'll display everything or everything but the MOTD, which is the most human-readable information there lol.
But another fight, another day.

👍 and merging.

astorije:
Damn, I just realized this wasn't in `second review needed`, sorry 😕
@maxpoulin64, @YaManicKill, are you 👍 on this *a posteriori* or do you want me to revert?

maxpoulin64:
@astorije: fine by me, :+1:

PR #291, opened 2016-05-01
Stricter eslint rule for curly brackets

xPaw:
As @astorije noticed, our current rules allowed some inconsistencies through.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -46,7 +46,9 @@ $(function() {
 		};
 	}
 
-	$("#play").on("click", function() { pop.play(); });
+	$("#play").on("click", function() {
+		pop.play();
+	});
 
 	$(".tse-scrollable").TrackpadScrollEmulator();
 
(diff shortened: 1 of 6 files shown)


astorije:
👍 and as it's a minor "linting bugfix", merging right away.

client/js/lounge.js
```diff
@@ -193,7 +195,9 @@ $(function() {
 		);
 		renderChannel(data.chan);
 		var chan = sidebar.find(".chan")
-			.sort(function(a, b) { return $(a).data("id") - $(b).data("id"); })
+			.sort(function(a, b) {
+				return $(a).data("id") - $(b).data("id");
+			})
```
astorije: I'm not against having this one a single line to be honest, but if that helps with other inconsistencies, sure, why not.
If we start having ES6-style arrow functions, I think having this on a single line will be beneficial however:

```js
.sort((a, b) => $(a).data("id") - $(b).data("id"))
```

YaManicKill: FWIW, [4.x mostly supports arrow functions.](http://node.green/#arrow-functions) (there are a few edge cases it doesn't support, but it's unlikely to be something we ever have to deal with)

PR #297, opened 2016-05-06
Save user when parting channels

xPaw:
Fixes channels persisting in the config even after leaving them.

### DIFF
--- a/src/plugins/inputs/part.js
+++ b/src/plugins/inputs/part.js
@@ -13,15 +13,18 @@ exports.input = function(network, chan, cmd, args) {
 		return;
 	}
 
-	var irc = network.irc;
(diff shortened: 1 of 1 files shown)


src/plugins/inputs/part.js
```diff
@@ -23,5 +23,9 @@ exports.input = function(network, chan, cmd, args) {
 		chan: chan.id
 	});
 
+	if (chan.type === "channel") {
+		this.save();
```
astorije: Can't we put that at line 18 (and I guess moving line 21 above the block lines 16-19 then)

xPaw: Changed.

astorije:
👍
Minor bugfix: merging.

PR #300, opened 2016-05-07
Add labels in connect window

xPaw:
Adds id/for in connect window so the labels are correctly linked.

Also makes port field a number type with min/max set.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -94,31 +94,30 @@
 									<h2>Network settings</h2>
 								</div>
 								<div class="col-sm-3">
-								<label>Name</label>
-									</div>
+									<label for="connect:name">Name</label>
+								</div>
 								<div class="col-sm-9">
-									<input class="input" name="name" value="<%= defaults.name %>">
+									<input class="input" id="connect:name" name="name" value="<%= defaults.name %>">
 								</div>
 								<div class="col-sm-3">
-									<label>Server</label>
+									<label for="connect:host">Server</label>
 								</div>
 								<div class="col-sm-6 col-xs-8">
-									<input class="input" name="host" value="<%= defaults.host %>" <%= typeof(lockNetwork) !== "undefined" && lockNetwork ? "disabled" : "" %>>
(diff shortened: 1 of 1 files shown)


YaManicKill:
Looks fine. Only thing is you appear to have added 2 aria labels, and no more. Is this deliberate?

xPaw:
> Only thing is you appear to have added 2 aria labels, and no more. Is this deliberate?

Yep. I added them because there's two fields (ip/port) but only one label, and adding aria labels describes the fields further.

YaManicKill:
Ah right, makes sense. 

:+1: and merging

client/index.html
```diff
@@ -94,31 +94,30 @@ <h1 class="title">Connect</h1>
 									<h2>Network settings</h2>
 								</div>
 								<div class="col-sm-3">
-								<label>Name</label>
-									</div>
+									<label for="connect:name">Name</label>
+								</div>
 								<div class="col-sm-9">
-									<input class="input" name="name" value="<%= defaults.name %>">
+									<input class="input" id="connect:name" name="name" value="<%= defaults.name %>">
```
astorije: For future references, I think we should avoid using `:` in `id`s. I believe they need escaping if we want to grab them with jQuery or CSS, and in CSS it would look confusing as `:` is used for pseudo-elements.

I get the idea of namespacing though, and I think using `-` or `_` for that works fine too.

xPaw: @maxpoulin64 used `:` in settings PR, so I went with same notation.

astorije: I see. I still think we should avoid this for the 2 reasons given above.

maxpoulin64: I agree with @astorije there. I totally didn't think at all about ever wanting to use those names in CSS selectors. I guess the only real option here is to use `_` as everything else seems to be reserved (and `-` is used as a space replacement since we don't CamelCase CSS names). I wish there was a better looking character for representing namespaces that's not so close to `-` but apparently it's `[a-zA-Z0-9_-]`. Too bad, `:` made it easy to avoid id conflicts and looked good 😞 We could always escape it but it quickly gets ugly, especially in JS strings due to double escaping: `$("#connect\\:name")`.

PR #305, opened 2016-05-08
Minor wording changes for better clarity

astorije:
These 2 changes make text a tiny bit more explicit.
In particular, when using a screen reader (VoiceOver in my tests), "motd" is read as a word while "MOTD" is read letter-by-letter, and "Show more" is read without any context, making it difficult to understand "more what" we are talking about.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -179,7 +179,7 @@
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="motd">
-									Show motd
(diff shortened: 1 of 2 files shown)


astorije:
Let me know if other texts are difficult to understand on their own (ping @lilmike) and I'll add them to the current PR.

maxpoulin64:
👍 and it also helps people know what MOTD means for those that don't know.

Issue #301, opened 2016-05-07
Use path.join everywhere


xPaw:
I've noticed in some places the folders are joined like `HOME + "/users"`. Using `path.join` is better.

astorije:
Why is it better? If it's just a concatenation between two strings (like in your example), then `+` is just fine.

maxpoulin64:
@astorije: The latter is cross-platform and independent of whatever path separator an OS uses. So for  example on Windows you'll get a path like `.lounge\users\astorije.json` instead of `.lounge/users/astorije.json`. In the case of Windows it works, but it's just considered better/good practice to use the helper.

astorije:
Aaaaah, that'll teach me to read things too quickly!
I read `join` as in `[HOME, "users"].join("/")` and really I didn't see the point... 👊 

Then yes, I fully agree!

For the record, [here is the `path.join` doc](https://nodejs.org/api/path.html#path_path_join_path1_path2).

PR #307, opened 2016-05-08
Replace all concatenated paths with Node's path.join

astorije:
My attempt at fixing #301.

Let me know if this solution is crappy or should be improved, I don't claim this is the one solution and I'm happy to get feedback. Also, I hope I didn't forget anything.

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -4,6 +4,7 @@ var program = require("commander");
 var pkg = require("../../package.json");
 var fs = require("fs");
 var mkdirp = require("mkdirp");
+var path = require("path");
 var Helper = require("../helper");
 
 program.version(pkg.version, "-v, --version");
@@ -12,17 +13,22 @@ program.option("    --home <path>" , "home path");
 
 var argv = program.parseOptions(process.argv);
 if (program.home) {
-	Helper.HOME = program.home;
+	Helper.HOME = path.resolve(program.home);
 }
 
-var config = Helper.HOME + "/config.js";
-if (!fs.existsSync(config)) {
+if (!fs.existsSync(Helper.CONFIG_PATH)) {
 	mkdirp.sync(Helper.HOME, {mode: "0700"});
 	fs.writeFileSync(
-		config,
-		fs.readFileSync(__dirname + "/../../defaults/config.js")
+		Helper.CONFIG_PATH,
+		fs.readFileSync(path.resolve(path.join(
+			__dirname,
+			"..",
+			"..",
+			"defaults",
+			"config.js"
+		)))
 	);
-	log.info("Config created:", config);
+	log.info("Config created:", Helper.CONFIG_PATH);
 }
 
 require("./start");
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -12,9 +12,8 @@ program.option("");
 program.option("    --home <path>" , "home path");
 
 var argv = program.parseOptions(process.argv);
-if (program.home) {
-	Helper.HOME = path.resolve(program.home);
-}
+
+Helper.setHome(program.home);
 
 if (!fs.existsSync(Helper.CONFIG_PATH)) {
 	mkdirp.sync(Helper.HOME, {mode: "0700"});
--- a/src/helper.js
+++ b/src/helper.js
@@ -1,16 +1,20 @@
 var path = require("path");
 var os = require("os");
 
-var HOME = expandHome("~/.lounge");
-
-module.exports = {
-	HOME: HOME,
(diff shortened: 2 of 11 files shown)


src/userLog.js
```diff
@@ -5,7 +5,7 @@ var Helper = require("./helper");
 
 module.exports.write = function(user, network, chan, msg) {
 	try {
```
xPaw: Can we move this whole try/catch block into main `exports`, so that it's not invoked on each `write`? Alternatively this can be moved all the way to `index` CLI where `config` is created.

src/helper.js
```diff
@@ -1,14 +1,23 @@
...
+	getUserConfigPath: getUserConfigPath,
 };
 
+function getUserConfigPath(name) {
+	return path.join(this.USERS_PATH, name + ".json");
+}
+
 function getConfig() {
-	return require(path.resolve(this.HOME) + "/config");
+	return require(path.resolve(this.CONFIG_PATH));
```
xPaw: Remove `path.resolve` here.

src/helper.js
```diff
@@ -1,14 +1,23 @@
 var path = require("path");
 var os = require("os");
 
+var HOME = path.join(process.env.HOME || process.env.USERPROFILE, ".lounge");
```
xPaw: Add `path.resolve` here.

src/command-line/index.js
```diff
@@ -15,12 +15,12 @@ if (program.home) {
...
 }
 
-var config = Helper.HOME + "/config.js";
+var config = Helper.CONFIG_PATH;
 if (!fs.existsSync(config)) {
 	mkdirp.sync(Helper.HOME, {mode: "0700"});
 	fs.writeFileSync(
 		config,
-		fs.readFileSync(__dirname + "/../../defaults/config.js")
+		fs.readFileSync(require("path").join(__dirname, "..", "..", "defaults", "config.js"))
```
xPaw: I think `path.resolve` is needed here.

src/helper.js
```diff
@@ -26,4 +35,3 @@ function expandHome(path) {
 
 	return path.replace(/^~($|\/|\\)/, home + "$1");
```
xPaw: Add `path.resolve` after replacing.

src/command-line/index.js
```diff
@@ -15,12 +15,12 @@ if (program.home) {
 	Helper.HOME = program.home;
```
xPaw: Add `path.resolve` here.

astorije:
@xPaw, I believe I have addressed all your comments.

However, this PR needs a lot of testing. I have done some, but definitely happy to have others test too. If done badly, this can create and use config in wrong directories, so people would believe their config was wiped.

src/command-line/index.js
```diff
@@ -12,15 +13,21 @@ program.option("    --home <path>" , "home path");
...
 
-var config = Helper.HOME + "/config.js";
+var config = Helper.CONFIG_PATH;
 if (!fs.existsSync(config)) {
 	mkdirp.sync(Helper.HOME, {mode: "0700"});
 	fs.writeFileSync(
 		config,
-		fs.readFileSync(__dirname + "/../../defaults/config.js")
+		fs.readFileSync(path.resolve(path.join(
+			__dirname,
```
astorije: [This](https://github.com/thelounge/lounge/blob/c8df0d68c02f30a9cb0c7c7fa93d93479654223b/src/plugins/irc-events/ctcp.js#L1) uses something else, I'm thinking the goal is the same. Should we stick with one or the other?

xPaw: `__dirname` is correct.

src/helper.js
```diff
@@ -1,17 +1,26 @@
 var path = require("path");
 var os = require("os");
 
+var HOME = expandHome("~/.lounge");
+
 module.exports = {
-	HOME: (process.env.HOME || process.env.USERPROFILE) + "/.lounge",
-	getConfig: getConfig,
+	HOME: HOME,
+	CONFIG_PATH: path.join(HOME, "config.js"),
```
xPaw: We discussed this in IRC, but for sake of visibility: this breaks `--home` option.

xPaw:
Okay, I've fixed `--home`. It also correctly expands `~` if specified.

astorije:
Alright, I'm all set with this PR! (Removing `second review needed` because I made changes since @xPaw's additions).

@thelounge/maintainers, could you review and test this please? :-)

YaManicKill:
Review looks fine, but I'll set it up so I can test it tomorrow probably (quite busy tonight).

astorije:
@maxpoulin64, do you mind giving this a review as well?

PR #308, opened 2016-05-08
Fix unread counter not being formatted on page load

xPaw:
This also fixes `1666` being formatted as `1.7k` instead of `1.6k`.

### DIFF
--- a/client/js/libs/handlebars/roundBadgeNumber.js
+++ b/client/js/libs/handlebars/roundBadgeNumber.js
@@ -1,0 +1,9 @@
+Handlebars.registerHelper(
+	"roundBadgeNumber", function(count) {
+		if (count < 1000) {
+			return count;
+		}
+
+		return (count / 1000).toFixed(2).slice(0, -1) + "k";
+	}
+);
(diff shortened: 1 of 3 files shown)


client/js/libs/handlebars/formatNumber.js
```diff
@@ -0,0 +1,9 @@
+Handlebars.registerHelper(
+	"formatNumber", function(count) {
```
astorije: Instead of the generic word "formatNumber" (that we might need somewhere else), shouldn't we call this something like, I don't know, "badgifyNumber", ... ?

xPaw: `formatThousands` maybe?

MaxLeiter:
roundBadgeNumber?

xPaw:
Renamed.

maxpoulin64:
All good here, 👍 

`/znc playbuffer #thelounge` works pretty well to test that :)

astorije:
👍 and merging.

Interestingly:

```js
> (1296 / 1000).toFixed(2).slice(0, -1) + "k"
'1.3k'
> (1295 / 1000).toFixed(2).slice(0, -1) + "k"
'1.2k'
```

Thanks JS!

PR #309, opened 2016-05-09
Fix running multiple scripts on Windows

xPaw:
There's basically no other way besides making a standalone script or writing JS code for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
     "build": "npm run build:grunt && npm run build:handlebars",
     "build:grunt": "grunt",
     "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
-    "test": "npm run test:mocha; npm run lint",
+    "test": "(npm run test:mocha || true) && npm run lint",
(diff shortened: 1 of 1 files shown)


YaManicKill:
:+1: It's not a great way, but I'd much rather this than a custom script.

package.json
```diff
@@ -12,12 +12,12 @@
   },
   "scripts": {
     "start": "node index",
-    "build": "npm run build:grunt && npm run build:handlebars",
+    "build": "(npm run build:grunt || true) && (npm run build:handlebars || true)",
```
astorije: I get the need for `test`, but I think we should let `build` fail if any of these fail. I can't think of a use case for this.

YaManicKill: Actually, that's a good point.

astorije:
This is gross :(
I didn't think we were ensuring compatibility with `cmd`, but if this is really a requirement, then :+1:.

> but I'd much rather this than a custom script.

Yes!

astorije:
Well :+1: with a minor comment inline actually.

PR #310, opened 2016-05-10
Add manifest.json for nicer mobile experience

### LINKED ISSUE #35: Add app manifest for better mobile experience
https://developer.mozilla.org/en-US/docs/Web/Manifest

Example file: https://github.com/SteamDatabase/steamstat.us/blob/gh-pages/manifest.json

xPaw:
Fixes #35

References:
https://developers.google.com/web/updates/2014/11/Support-for-installable-web-apps-with-webapp-manifest-in-chrome-38-for-Android?hl=en
https://developers.google.com/web/updates/2014/11/Support-for-theme-color-in-Chrome-39-for-Android?hl=en

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -9,6 +9,7 @@
 	<meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
 	<meta name="mobile-web-app-capable" content="yes">
 	<meta name="referrer" content="no-referrer">
+	<meta name="theme-color" content="#455164">
 
 	<title>The Lounge</title>
 
@@ -19,6 +20,7 @@
 
 	<link rel="shortcut icon" href="img/favicon.png" data-other="img/favicon-notification.png" data-toggled="false" id="favicon">
 	<link rel="apple-touch-icon" sizes="120x120" href="img/apple-touch-icon-120x120.png">
+	<link rel="manifest" href="manifest.json">
 
 	</head>
 	<body class="<%= public ? "public" : "" %>">
--- a/client/manifest.json
+++ b/client/manifest.json
@@ -1,0 +1,27 @@
+{
+	"name": "The Lounge",
+	"short_name": "The Lounge",
+	"description": "Self-hosted web IRC client",
+	"display": "standalone",
+	"orientation": "any",
+	"theme_color": "#455164",
+	"background_color": "#455164",
+	"icons":
+	[
+		{
+			"src": "img/touch-icon-192x192.png",
+			"sizes": "192x192",
+			"type": "image/png"
+		},
+		{
+			"src": "img/apple-touch-icon-120x120.png",
+			"sizes": "120x120",
+			"type": "image/png"
+		},
+		{
+			"src": "img/favicon.png",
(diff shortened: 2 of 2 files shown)


maxpoulin64:
I like that a lot. It's just a color, but it sure does make The Lounge feel a lot more native/integrated! 👍

MaxLeiter:
Do we need a `theme-color` in the manifest file and `client/lounge.js`?

client/index.html
```diff
@@ -9,6 +9,7 @@
 	<meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
 	<meta name="mobile-web-app-capable" content="yes">
 	<meta name="referrer" content="no-referrer">
+	<meta name="theme-color" content="#455164">
```
astorije: This works well with the default theme, but how does it behave with other themes (this comment applies to colors within the manifest file too)?
My guess is that it's not as nice on other themes, at least for the `theme-color` setting, as the toolbar color would stay blue while the loaded theme is not.

Even when themes include metadata like a global color like this, it will be tricky: this needs to be re-set after the user logs in (assuming themes will be controlled by the user settings)

I'm OK with the background color, but could you double check with the toolbar color and see how it goes?

maxpoulin64: I think that's something that will be properly dealt with with actual theme packages. Then, we could template the manifest and have it work better. But for now I don't think there's much to do, and anything looks better than white at least.

astorije: You're talking about the background color, I'm talking about the toolbar color.

It's not as bad as a website because when set up as an app, the URL + icons bar disappears, but this is not true for users that did not set up The Lounge as a standalone app (such as those testing the demo, or pretty much all public instances). Also applies to the very top bar, where time and battery/connectivity icons live, which is affected by the `theme-color`.

xPaw: The only problem as you are describing it, is that the color will look slightly out of place on non default themes. We can't really make this customizable per theme until we get packages or theme settings.

astorije: So what about setting up only the background color and not the theme color until then?

xPaw: I'd prefer if we left it in, there's no harm in it.

astorije:
Alright, I guess I can't win then :-)
👍 for now, hopefully this will get improved over time. I'll post some screenshots shortly (got to run now) to keep a ref somewhere when someone else decides to improve this.

astorije:
So, for the record, without this patch applied, we have:

  

First picture is in browser, others are "browser-less". Notice the black status bar (battery, connectivity, time, ...) across both modes.

Applying this patch, it works very nicely with the default theme:

   

Notice that the URL bar in the browser mode has changed, but so has the status bar in all modes. It is now a dark blue instead of black.

As a matter of fact, it works also very nicely on the Morning theme:

 

But... Not so much on Zenburn and Crypto themes:

 



I am using the default theme so I personally don't care too much, I know @xPaw is using Morning. Not worth reverting this at all, but good to keep in mind in case someone opens an issue for Zenburn or Crypto.

PR #317, opened 2016-05-10
Fix missing colors in action messages

astorije:
This was heavily inspired by https://github.com/erming/shout/pull/664 (and discussion on https://github.com/erming/shout/pull/663) so kudos and credits go to @bews! \o/

This is also the last missing piece of https://github.com/thelounge/lounge/issues/31. I don't know why I didn't fix that earlier, really...

A few screenshots to illustrate this:

screen shot 2016-05-10 at 19 16 08

screen shot 2016-05-10 at 19 16 18

screen shot 2016-05-10 at 19 16 26
(We can't really see on this last one but `astorije` has a slightly different color.)

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -747,6 +747,7 @@ button,
 }
 
 #chat.no-colors .from .user,
+#chat.no-colors .text .user,
 #chat.no-colors .sidebar .user {
 	color: #50a656 !important;
 }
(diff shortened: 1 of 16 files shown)


maxpoulin64:
Looks all good to me. 👍 

I think we were waiting to have a better way to deal with colors before we did this, but since that can take a while still I vote to merge this and deal with the colors as a whole separately.

PR #320, opened 2016-05-11
Fix sending messages starting with a space

### LINKED ISSUE #319: Starting with " /" (whitespace + slash) does not send the message
This was fixed in a previous PR (~~https://github.com/thelounge/lounge/pull/154 I believe~~ Actually, I'm not sure we ever fixed this...) to allow escaping commands with a slash (`// something...`) or a leading whitespace (` /something...`).

maxpoulin64:
There really is no need to trim apart bringing more situations like this. If I want to put spaces in front of my message (say, for alignment purposes), then I should be able to.

Fixes #319

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -267,7 +267,7 @@ Client.prototype.setPassword = function(hash) {
 
 Client.prototype.input = function(data) {
 	var client = this;
-	var text = data.text.trim();
+	var text = data.text;
 	var target = client.find(data.target);
 
 	// This is either a normal message or a command escaped with a leading '/'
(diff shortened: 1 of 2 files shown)


xPaw:
Not trimming is fine, however I think we should block empty messages (e.g. spaces only).

YaManicKill:
> Not trimming is fine, however I think we should block empty messages (e.g. spaces only).

Agreed, we don't want to send empty messages.

astorije:
@maxpoulin64, any chance you could address the comments above? (Removing the `second review needed` label until then)

xPaw:
As someone pointed out in irc, people might want to send empty messages (even if that's silly). Also checked HexChat, they allow it too.

So I'll be fine with this fix.

astorije:
👍

PR #322, opened 2016-05-11
Generate better nick colors

xPaw:
Generates nicer looking colors.

| Light | Dark |
--- | ---
 |


client/index.html
```diff
@@ -21,7 +21,7 @@
 	<link rel="apple-touch-icon" sizes="120x120" href="img/apple-touch-icon-120x120.png">
 
 	</head>
-	<body class="<%= public ? "public" : "" %>">
+	<body class="<%= public ? "public" : "" %>" data-nick-luminosity="<% if (theme === 'themes/morning.css' || theme === 'themes/zenburn.css') { %>light<% } else { %>dark<% } %>">
```
xPaw: This is horribly temporary until we get proper theme configs.

YaManicKill:
Yeah, looks pretty good to me. :+1: 

Agree about the hacky way to deal with dark themes, but this is not the right place to sort it. So, happy with it.

maxpoulin64:
I'm not a fan of the colors on the default theme. One of the nice feature of the existing color generator is that it generated pretty bright colorful colors most of the time (when it was readable, of course).

As for the hack, [I think I found a way to abuse CSS to set variables in JS](https://d.max-p.me/irc/test.html). It probably would be less bad than hardcoding the themes, and also have the side effect of making it possible to customize it through custom CSS as well.

astorije:
I am not too fond of this solution, really.
I agree with @maxpoulin64 about default theme colors, but beyond that, the change is to replace a library with another library (5 times bigger) to just allow themes to choose between light/dark luminosity. On top of that, I feel like the hack might stay for a while, @maxpoulin64's patch is still very hackish, ... Meh, I don't really see this as a victory here :/

My biggest concern is that themes have very little control over colors with this library (nor the current one).

I have been thinking of an alternative to the current solution:

- Define a maximum number of colors we want to support
- The `stringcolor` helper, instead of building hex codes (`#123456`), builds indexed class names from `color_0` (or `color-1` if we want to keep `0` for special cases or whatever) to `color-N`
- Each theme defines its own colors, up to 64 of them. If they want less of them (think of a colorblind-oriented theme for example, that only defines 8 very different colors), they can combine them with the same color (`.color_10, ..., .color_19, { color: red; }` or using CSS features such as `[class*=" color-*"]` or even/odd rules, ...)
- For the `no-colors`, either keep the current format or define a single color for all `.color-*`... This last possibility is also how a theme can decide to not have colored nicknames.
- A nice win is, if you don't like this color (the one for your own nick?), you can override it :D

Sure, this is not perfect. But this lets themes fully control their colors, or fine-tune a pre-computed one if it doesn't fit.
We can then recommend or easily write a script/online tool to generate N classes using this `randomColor` package based on luminosity or hue if an author doesn't want to deal with creating them but still have nice random ones.

I'm thinking of trying this quickly and see how it goes with our project. Anyone has concerns or suggestions to improve this idea before I start?

xPaw:
Closing in favour of #325.

PR #324, opened 2016-05-11
Update irc-framework

xPaw:


src/client.js
```diff
@@ -182,6 +183,7 @@ Client.prototype.connect = function(args) {
 			} else {
 				webirc = {
 					password: config.webirc[network.host],
+					username: network.username,
```
maxpoulin64: I think webirc usernames are for the client itself, not to identify the user. Probably would need to update the config format so it takes a username and password instead of just a password. Thoughts on that?

dgw: Sounds like it. https://kiwiirc.com/docs/webirc

dgw: Any reason not to default to something like `lounge`?

maxpoulin64: @dgw: We could, but it needs to be user changeable in case a server provides a username *and* password. So far the ircd I use for testing doesn't even use the username, but since it's in the spec we should implement it right.

PR #327, opened 2016-05-12
Improve nick highlights

### LINKED ISSUE #159: Highlight with user mode prefixes
Sometimes people who are new to IRC copy the entire display nick they see, including the mode symbol, as some clients show it in the nick list (including The Lounge, at least when using zenburn). I've gotten my fair share of HLs as `+dgw` or `@dgw`.

The Lounge already strips out `@` when checking for highlights, but IMO it should ignore any symbol listed in `PREFIX` sent with 005 ISUPPORT numeric by the server (commonly `+%@`, often also `&~`).

This does directly contradict #158, but my own IRC experience has shown that a not insignificant number of highlights could be missed if mode prefixes are not ignored.

xPaw:
Closes #159.

Initially tried with some `indexOf` and checking characters, but it ends up being messier.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -144,6 +144,7 @@ Client.prototype.connect = function(args) {
 		ip: args.ip,
 		hostname: args.hostname,
 	});
+	network.setNick(nick);
 
 	client.networks.push(network);
 	client.emit("network", {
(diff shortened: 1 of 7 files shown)


maxpoulin64:
Regexes fixes everything. 👍

astorije:
👍 
Also, 👏 👏 on the test suite, very neat!

PR #328, opened 2016-05-12
Only build master branch on travis

xPaw:
As @maxpoulin64 pointed out, we pointlessly trigger twice for pull requests done from this repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,16 @@ node_js:
 - '5'
 - '4'
 - '4.2.0'
+
+branches:
+  only:
+    - master
+
 notifications:
   email:
     on_success: never
     on_failure: always
+
 deploy:
   provider: npm
   email:
@@ -23,4 +29,3 @@ script:
 - npm run test:mocha
 - npm run lint:js
 - npm run lint:css
-


YaManicKill:
The problem with this is that we don't build until you create a PR, so if you are developing something on a branch, you won't see failures on travis (if you forget to lint locally for example) until you create a PR.

xPaw:
That's true, but is it a big problem? Ideally you should do `npm run test` before creating a pull request. And this change only affects the maintainers, we can probably live with it.

YaManicKill:
> That's true, but is it a big problem?

I think it's a bigger problem than just travis running 2 builds. Especially as it runs them in parallel, so it's not like it increases the time.

> Ideally you should do npm run test before creating a pull request.

astorije:
> Especially as it runs them in parallel, so it's not like it increases the time.

That. I would say until this is a problem (I don't see why it could be right now, but we never know, Travis CI can evolve, ...), it's not one.

Also, one detail: the two builds are not the exact same. [According to the Travis CI doc](https://docs.travis-ci.com/user/pull-requests#How-Pull-Requests-are-Tested), while the master build tests the branch itself, the PR one tests the merge of the PR and the master branch:

> Rather than test the commits from the branches the pull request is sent from, we test the merge between the origin and the upstream branch.

I think having both is interesting.

xPaw:
Fair enough.

PR #329, opened 2016-05-12
Do not ignore our handlebars plugins in eslint

xPaw:


YaManicKill:
:+1: Small bug fix, no need for another review. Merging.

PR #330, opened 2016-05-12
Add multiple lines to input.

YaManicKill:
This isn't quite ready, but thought I'd get a PR up as it's getting close.

Things to do:

- [x] Tidy up the horrible 3rd party library that I've modified a lot. It uses some weird cloning that I want to see if I can get rid of
- [ ] Fix bug where if we have multiple lines in the input, we can't go back in the history. Single line works fine. 
- [ ] Fix small issue with line 36 in ns-autogrow. Uncommented, it fixes a slight issue with the text popping off the top when you make a newline. But it also means that you can't ever see below the fixed max height of the input. So I need to figure out a way for us to only do this if we are able to expand the input further (occaisionally, maxheight is 1 or 2 px greater than the height we currently are, which means we can't change, but we think we can).

Thoughts on other things, or testing of it are welcome.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,5 @@
 module.exports = function(grunt) {
-	var libs = "client/js/libs/**/*.js";
+	var libs = ["client/js/libs/**/*.js", "client/js/components/**/*.js"];
 	grunt.initConfig({
 		watch: {
 			files: libs,
--- a/client/js/components/ns-autogrow.js
+++ b/client/js/components/ns-autogrow.js
@@ -1,0 +1,63 @@
+/* global jQuery */
+(function($) {
+	$.fn.autogrow = function(opts) {
+		var that = $(this),
+			selector = that.selector,
+			defaults = {
+				context: $(document),
+				speed: 200,
+				cloneClass: "autogrowclone",
+				onInitialize: true
+			};
+		opts = $.isPlainObject(opts) ? opts : {
+			context: opts ? opts : $(document)
+		};
+		opts = $.extend({}, defaults, opts);
+		that.each(function(i, elem) {
+			var min, formMin;
+			elem = $(elem);
+			min = parseInt(elem.css("height"), 10) || elem.innerHeight();
+			formMin = parseInt($("#form").css("height"), 10) || elem.innerHeight();
+			elem.data("autogrow-form-diff", formMin - min);
+			elem.data("autogrow-start-height", min);
+			elem.css("height", min);
+			$("#form").css("height", formMin);
+		});
+		opts.context.on("keyup paste", selector, resize);
+
+		function resize(e) {
+			var box = $(this),
+				minHeight = box.data("autogrow-start-height") || 0,
+				maxHeight = minHeight * 3,
+				oldHeight = box.innerHeight(),
+				newHeight = Math.min(this.scrollHeight, maxHeight);
+			if (oldHeight < newHeight) { // user is typing
+				// this.scrollTop = 0; // try to reduce the top of the content hiding for a second
+				box.css({height: newHeight}, opts.speed);
+				$("#form").css({height: newHeight + box.data("autogrow-form-diff")}, opts.speed);
+				$("#windows").css({bottom: newHeight + box.data("autogrow-form-diff")}, opts.speed);
+			} else if (!e || e.which === 8 || e.which === 46 || (e.ctrlKey && e.which === 88) || (!e.ctrlKey && e.which === 13)) { // user is deleting, backspacing, or cutting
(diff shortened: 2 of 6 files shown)


Gruntfile.js
```diff
@@ -1,5 +1,5 @@
 module.exports = function(grunt) {
-	var libs = "client/js/libs/**/*.js";
+	var libs = ["client/js/libs/**/*.js", "client/js/components/*.js"];
```
xPaw: Make this `**/*` please.

xPaw:
Looks like autogrow hack comes from the fact it's animated. Do we really need the animation?

client/js/libs/jquery/inputhistory.js
```diff
@@ -9,48 +9,54 @@
...
-						i--;
+				if (!e.shiftKey) {
+					if (self.val() !== "") {
+						var values = self.val().split("\n");
+						for (var j = 0; j < values.length; j++) {
+							i = history.length;
+							history[i - 1] = values[j];
+							history.push("");
+							if (history[i - 1] === history[i - 2]) {
+                history.splice(-2, 1);
```
xPaw: nice tabs yo.
If you move this file to components, it'll get linted.

YaManicKill:
@xPaw Yeah, that seems to do the job. I guess the animation isn't really necessary. It's much cleaner code now (although, there's probably some stuff that I don't use in the library anymore so I'll spend some time on Saturday cleaning it up).

Just the other 2 bugs to fix now.

client/js/libs/jquery/inputhistory.js
YaManicKill: I wasn't oging to do this, because it's erming's library. But then I realised how much I changed it, so I moved it. And fixed the tabs :-P (it was using spaces, stupid atom).

Gruntfile.js
YaManicKill: Done

client/js/components/ns-autogrow.js
```diff
@@ -0,0 +1,63 @@
...
+
+		function resize(e) {
+			var box = $(this),
+				minHeight = box.data("autogrow-start-height") || 0,
+				maxHeight = minHeight * 3,
+				oldHeight = box.innerHeight(),
+				newHeight = Math.min(this.scrollHeight, maxHeight);
+			if (oldHeight < newHeight) { // user is typing
+				// this.scrollTop = 0; // try to reduce the top of the content hiding for a second
+				box.css({height: newHeight}, opts.speed);
```
xPaw: You forgot to remove `opts.speed` in these `css` calls.

client/js/components/ns-autogrow.js
```diff
@@ -0,0 +1,63 @@
...
+				$("#windows").css({bottom: newHeight + box.data("autogrow-form-diff")}, opts.speed);
+			} else if (!e || e.which === 8 || e.which === 46 || (e.ctrlKey && e.which === 88) || (!e.ctrlKey && e.which === 13)) { // user is deleting, backspacing, or cutting
+				if (oldHeight > minHeight) {
+					do { // reduce height until new height is less than the scrollheight
+						newHeight--;
+						box.css("height", newHeight);
+						$("#form").css("height", newHeight + box.data("autogrow-form-diff"));
+					} while (newHeight >= box[0].scrollHeight && newHeight >= minHeight);
+					newHeight++; // adding one back eliminates a wiggle on deletion
+					box.focus(); // Fix issue with Chrome losing focus from the textarea.
```
xPaw: I don't think this is needed.

client/js/components/ns-autogrow.js
```diff
@@ -0,0 +1,63 @@
+/* global jQuery */
+(function($) {
+	$.fn.autogrow = function(opts) {
+		var that = $(this),
+			selector = that.selector,
+			defaults = {
```
xPaw: All these defaults aren't used (besides `context` which is used to bind events, but can it be done on `that` instead of using event delegation?)

astorije:
@YaManicKill, this PR solves a common issue that I keep seeing on the IRC channel (the auto-complete). I would love to see this happen, do you think it requires a lot of work?
If so, can it be kept simple at first (like not have auto-expand or something like that) and improve thereafter?

YaManicKill:
There are 2 bugs left with it. 1 of them is something others have said isn't an issue (isn't a bug at all, is what they would expect) and the other is very minor. 

A few small points that @xpaw pointed out I can solve on probably Monday and I think I'm happy with it for now. The bug is a tiny issue and only noticed for a split second when you press shift-enter to do a new line.

YaManicKill:
Alright, closing for #379

PR #169, opened 2016-03-08
Add readmarker

bark:
Adds a line marking where the last read message was. Solving #25

PR #332, opened 2016-05-14
A proper unread marker

xPaw:
This PR has two issues:
- Doesn't work correctly if unread message is in the unloaded buffer
- Need to look into using document visibility API.

This PR however has proper server state tracking, so the marker isn't lost between page loads.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -285,7 +285,20 @@ $(function() {
 
 	function renderChannelMessages(data) {
 		var documentFragment = buildChannelMessages(data.id, data.messages);
-		chat.find("#chan-" + data.id + " .messages").append(documentFragment);
+		var channel = chat.find("#chan-" + data.id + " .messages").append(documentFragment);
+
+		if (data.firstUnread > 0) {
+			var first = channel.find("#msg-" + data.firstUnread);
+
+			// TODO: If the message is far off in the history, we still need to append the marker into DOM
+			if (!first.length) {
+				channel.prepend(render("unread_marker"));
+			} else {
+				first.before(render("unread_marker"));
+			}
+		} else {
+			channel.append(render("unread_marker"));
+		}
 	}
 
 	function renderChannelUsers(data) {
@@ -313,12 +326,20 @@ $(function() {
 	socket.on("msg", function(data) {
 		var msg = buildChatMessage(data);
 		var target = "#chan-" + data.chan;
-		chat.find(target + " .messages")
+		var container = chat.find(target + " .messages");
+
+		container
 			.append(msg)
 			.trigger("msg", [
 				target,
 				data.msg
 			]);
+
+		if (data.msg.self) {
+			container
+				.find(".unread-marker")
+				.appendTo(container);
+		}
 	});
 
 	socket.on("more", function(data) {
@@ -751,10 +772,16 @@ $(function() {
 		}
 
 		viewport.removeClass("lt");
-		$("#windows .active")
+		var lastActive = $("#windows .active");
+
+		lastActive
 			.removeClass("active")
 			.find(".chat")
 			.unsticky();
+
+		lastActive
+			.find(".unread-marker")
+			.appendTo(lastActive.find(".messages"));
 
 		var chan = $(target)
 			.addClass("active")
@@ -1029,14 +1056,14 @@ $(function() {
 	setInterval(function() {
 		chat.find(".chan:not(.active)").each(function() {
 			var chan = $(this);
-			if (chan.find(".messages").children().slice(0, -100).remove().length) {
+			if (chan.find(".messages .msg:not(.unread-marker)").slice(0, -100).remove().length) {
 				chan.find(".show-more").addClass("show");
 			}
 		});
 	}, 1000 * 10);
 
 	function clear() {
-		chat.find(".active .messages").empty();
+		chat.find(".active .messages .msg:not(.unread-marker)").remove();
 		chat.find(".active .show-more").addClass("show");
 	}
 
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -679,6 +679,7 @@ button {
 #chat .show-more {
 	display: none;
 	padding: 10px;
+	padding-bottom: 0;
 	width: 100%;
 }
 
@@ -709,6 +710,17 @@ button {
 #chat .msg {
 	display: table-row;
 	word-wrap: break-word;
+}
+
+#chat .unread-marker span {
+	padding: 0;
+	height: 1px;
+	border-color: #455164;
+	background-color: #455164;
+}
+
+#chat .unread-marker:last-child {
+	display: none;
 }
 
 .inline-channel {
@@ -1625,16 +1637,22 @@ button {
 		padding: 2px 0;
 	}
 
-	#chat .msg:last-child {
-		height: auto;
-	}
-
 	#chat .time,
 	#chat .from,
 	#chat .text {
 		border: 0;
 		display: inline;
 		padding: 0;
+	}
+
+	#chat .unread-marker .time,
+	#chat .unread-marker .from {
+		display: none;
+	}
+
+	#chat .unread-marker .text {
+		display: block;
+		width: 100%;
 	}
 }
 
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -701,22 +701,36 @@ button {
 }
 
 #chat .messages {
-	display: table;
-	table-layout: fixed;
 	width: 100%;
 	padding: 10px 0;
-}
-
-#chat .msg {
-	display: table-row;
+	overflow-wrap: break-word;
 	word-wrap: break-word;
+	word-break: break-word;
 }
 
-#chat .unread-marker span {
-	padding: 0;
-	height: 1px;
-	border-color: #455164;
-	background-color: #455164;
+#chat .unread-marker {
+	position: relative;
+	text-align: center;
+	opacity: .5;
+	margin: 0 10px;
+}
+
+#chat .unread-marker:before {
+	position: absolute;
+	z-index: -1;
+	content: "";
+	left: 0;
+	right: 0;
+	top: 50%;
+	border-top: 1px solid #e74c3c;
+}
+
+#chat .unread-marker-text:before {
+	content: "New messages";
+	background-color: white;
+	color: #e74c3c;
+	padding: 0 10px;
+	font: bold 12px Lato;
 }
 
 #chat .unread-marker:last-child {
@@ -751,7 +765,7 @@ button {
 	color: #b1c3ce;
 	padding-right: 10px;
 	text-align: right;
-	width: 134px;
+	max-width: 134px;
 	min-width: 134px;
 }
 
@@ -1645,14 +1659,8 @@ button {
 		padding: 0;
 	}
 
-	#chat .unread-marker .time,
-	#chat .unread-marker .from {
-		display: none;
-	}
-
-	#chat .unread-marker .text {
-		display: block;
-		width: 100%;
+	#chat .unread-marker {
+		margin: 0;
 	}
 }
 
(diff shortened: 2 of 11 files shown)


maxpoulin64:
Looks good to me. My only comment is that I think the unread bar could be a bit more subtle and thicker. I liked the one from #169 better for that, because it doesn't constantly grab your attention.

Also probably should have one per theme. I'm going to guess yours is based on morning considering the blue color :)

xPaw:
I didn't really mess with the styling just yet, I wanted to get a working prototype out.

richrd:
Regarding hidden messages: Using :after for the implementation is something that I like and I think is rather elegant. Unfortunately it's probably not directly possible with hidden elements. (http://stackoverflow.com/questions/14049423/in-css-use-displaynone-on-the-element-but-keep-its-after) 

If a pseudo element is required I would suggest hiding child elements with something like this: ```#chat .messages .msg * {display: none}``` That way the pseudo element will be shown. I guess one of the other options would be adding a 'real' element before the first unread message (or action etc).

Just my 2 cents. And thanks for all the work you guys have put in to this project, love it! :+1:

xPaw:
Your `*` approach won't work due to the table layout chat uses.

richrd:
I tried using ```#chat .msg.join * {display: none}``` in Chrome via the inspector, and it worked fine for me. The unread marker was shown but the join line wasn't.

xPaw:
Hmm. I remember fixing an issue that was related to hiding joins and the layout messing up, but I can't remember what it was exactly.

xPaw:
I've updated the PR to create a separate marker line in DOM, the solution ends up being cleaner over all using this method.

client/js/lounge.js
```diff
@@ -275,7 +275,20 @@ $(function() {
 
 	function renderChannelMessages(data) {
 		var documentFragment = buildChannelMessages(data.id, data.messages);
-		chat.find("#chan-" + data.id + " .messages").append(documentFragment);
+		var channel = chat.find("#chan-" + data.id + " .messages").append(documentFragment);
+
+		if (data.firstUnread > 0) {
+			var first = $("#msg-" + data.firstUnread);
+
+			// TODO: If the message is far off in the history, we still need to append the marker into DOM
```
maxpoulin64: Is this TODO still relevant? Seems you do append the marker (well, prepend) to the DOM? I think you just need to move it to the right place in the `on("more",` handler right?

xPaw: Yes it's still relevant. To move it on `more` handler, firstUnread needs to be persisted then, the unread marker has to remain in DOM and hidden. But as we consider channel read as soon as we open it (not when we load the unread message), this becomes slightly more trickier.

maxpoulin64:
Okay, been using it for the day. Just dumping some notes for now, I will truely investigate later probably tomorrow.

* I think the marker gets trimmed with the other messages when the backlog gets too big. When I either restart the server (and ZNC plays the buffer back) or I've been in a channel for too long, the marker just stops working. I'll have to investigate further to confirm, just observations so far.

* The marker is never visible on mobile because the spans are no longer table cells. I removed the spans and just gave the div a height in custom CSS as a workaround and it worked.

xPaw:
@maxpoulin64 
1. I think I fixed this.
2. Fixed.

williamboman:
Been having this issue (might be running on a very old version of this PR though)

xPaw:
@williamboman Are you hiding joins or anything like that? I assume that would break the `last-child` selector

williamboman:
Yeah I hide join/parts etc.

maxpoulin64:
Yep, I can confirm both issues are fixed. I think the only issue left with this is the hiding of joins and parts. I honestly have no idea how this one can be fixed short of tracking the visibility state in JS :/

xPaw:
There's still an annoyance with it. When you open a channel and write a message, it shows the unread marker above your message. This isn't exactly the desired behaviour.

xPaw:
More fixes!

- Fixed /clear breaking it
- The line now moves to the bottom when receiving a `self` message (e.g. after sending a message)
- Made the line thinner

williamboman:
Seems much smoother now, nice. One thing, when joining a channel with a lot of unread messages the marker is (naturally) placed at the top, but when loading more messages it stays on-top of the previously top-most message.

xPaw:
@williamboman It's mentioned in the OP, and it's not entirely easy to solve due to new bugs it can introduce.

williamboman:
What about making it [opt-in (patch here)](https://gist.github.com/williamboman/1ef2c6a1157ffb6faf72cf258aa4306e) for 2.0.0 and perhaps opt-out once issues been resolved?

YaManicKill:
What's the status of this? I'm happy with @williamboman's suggestion for this (subject to review).

@xPaw Are you happy for this to be reviewed now? I'll probably manage to get a review done in the next few days.

xPaw:
Yes.

maxpoulin64:
So, I have been using this for a while now and everything seems to be working pretty well! I just found one sligh issue with it still during my testing, but that's about it: if you connect to a network and fill the Channels field to join a channel on connect, the marker won't work in that tab until a refresh. It only happens on connect, parting and rejoining channels doesn't seem to trigger it. I think that's because the windows are pre-opened on the client right away so when the server sends the join the marker doesn't get added because the window already "exists".

> Also, the line is quite an eyesore, need to design something for each theme that doesn't look ass.

For that part, this is how I themed mine:

capture d ecran_2016-07-01_20-13-40

I think it's not nearly as "hey, look at me!" as the current one and uses padding instead to make it easier to spot when scrolling. I'm totally fine with both styles, but if you like it the code for it is

```css
#chat .unread-marker span {
	padding: 1px 0 0 0;
	border: none;
	border-top: 4px transparent solid;
	border-bottom: 4px transparent solid;
	background: #eeeeee;
}
```

xPaw:
@maxpoulin64 extra padding (with border) might cause more jumps than necessary when hiding the line (marking as read).

YaManicKill:
I've been using this for quite a while, and I'm happy with it. Having looked at the code, it seems fine.

The only issue I've been having is if the last unread message is beyond what we have. I know @xPaw has been having some other issues pop up when trying to sort that out, so I think I'm happy for this to go in and that be improved later.

Like @maxpoulin64 I'm not a huge fan of the default look (on zenburn) but for the opposite reason. So, it's probably best to keep it as it is and we can change locally.

So, to summarise, it's not perfect, and will need some improvement later down the line, but :+1: because I think it's good enough for a first version, and I can't use lounge without it anymore :-P

astorije:
> @maxpoulin64 extra padding (with border) might cause more jumps than necessary when hiding the line (marking as read).

What do you mean by this, @xPaw?

astorije:
I haven't tried that a lot, but behavior seems to work really well, good job!

I'm with @maxpoulin64 on the design though. It is an eyesore indeed, at least on the default theme. Here is my opinion before I am entirely up for this: until we all agree on something that both looks nice and catchy enough, I suggest we go for a line that has the same color than the other borders in this window (`#e7e7e7` on default theme). Again, not a definitive solution ;-)

xPaw:
@astorije What I mean is, the messages are going to have to move down when the unread marker is moved down, to compensate for lost space. If you want to tweak styles, feel free to commit it yourself.

astorije:
> @astorije What I mean is, the messages are going to have to move down when the unread marker is moved down, to compensate for lost space.

I see now, and I got to play with it when there is said padding/margin, it didn't shock me. Yes, messages are moved a tiny bit, but I don't see that as an issue.

> If you want to tweak styles, feel free to commit it yourself.

I did actually, in a separate commit. I'm pretty happy of the result: it's eye-catchy so you can't miss it, and yet it integrates with the current design (designs actually, tweaked per theme to look even better). Here are the results:

Default | Crypto | Morning | Zenburn | Default on mobile
--- | --- | --- | --- | ---
default | crypto | morning | zenburn | default mobile

I got inspired from a few apps out there, that got me into specifying the "New messages" text. A few thing:

- I had to remove that `display: table` on `#chat .messages`. I tried many things and didn't see any consequences, both in Chrome and Firefox, but I had to if I wanted to truly center the text. Do you know of any potential consequences?
- I wanted to place that text in CSS (using `content`) so it can be easily tweaked per theme and not including in a selection for copy/paste somewhere. I couldn't find a way to avoid the extra `span` because the outer `div` already uses `:before ... content`. If there is a way, I'd be happy to know.
- I re-used the red color used in different places of the app, plus some transparency on light themes to not make it look aggressive/like an error.

I hope you'll like it 😄

maxpoulin64:
Looks pretty good to me @astorije!

For the table, I was sure we introduced it but apparently it comes from e4d6f8f9abdbb5dd0651cc8d881d1884b3210d76 long before we forked.

astorije:
Tiny modification between the screenshot and the commit: I extended the margins around the line (to somewhat align with the timestamp) and between the line and "New messages" from 5 to 10px. I'm all set now :)

YaManicKill:
At first glance, it does seem much better than the default colour. I've even removed my colour override, and it seems much more obvious. I'll keep this on for hte day and see how it works. It does seem to have a **lot** more space around it, and I'm not sure how I feel about that. We'll see how it goes over the day. But absolutely a much better colour (on zenburn).

YaManicKill:
Just to put it here, there was a slight issue with @astorije's changes. If someone had a very long name and you are using zenburn, you would get some really weird formatting issues because the name field was different lengths at different points. It has now been fixed.

YaManicKill:
Also, it was mentioned on the irc that it might be nice to have the unread marker only disappear on channel change, because the jumping when you send a message is a bit odd. I am not decided what I think about this, yet. I'm kinda of the opinion that it might be better to do that. What do others think?

williamboman:
> Also, it was mentioned on the irc that it might be nice to have the unread marker only disappear on channel change, because the jumping when you send a message is a bit odd. I am not decided what I think about this, yet. I'm kinda of the opinion that it might be better to do that. What do others think?

I think the way the messages jumps when removing the unread marker is bad UX, so I'm def in favor of this.

dgw:
To illustrate what I'm used to as far as unread marker behavior goes, I'll point at Textual. It does the following:
* Unread message comes in, channel in focus:
  * Marker stays where it was, visible; it does not hide.
* Unread message comes in, channel out of focus:
  * If first new unread message since channel lost focus:
    * Marker moves to position just above new message.
  * Else:
    * Marker stays where it was.
* User sends a new message while unread marker present in channel view:
  * See "Unread message comes in, channel in focus"

I don't think removing the unread marker when the user sends a message looks good. Textual provides a keyboard shortcut and menu option to move the marker to the current position, but that might not be practical/necessary for The Lounge to implement. As much as possible, I think the marker should move while the channel is not visible, to avoid sudden changes in the positions of lines.

astorije:
> To illustrate what I'm used to as far as unread marker behavior goes, I'll point at Textual. It does the following: [...]

Thanks for the feedback, @dgw, that's helpful!

> I think the way the messages jumps when removing the unread marker is bad UX, so I'm def in favor of this.

How is that bad UX?! Debatable UI, maybe, improvable feature, for sure, but *bad UX*, definitely not. The alternative offered here means that when sending a message, therefore acknowledging unread messages, the unread messages marker stays, losing its meaning. Assigning different behavior to a component than what it actually is, **that** becomes bad UX.

That being said, I'm not saying the current behavior/UI/whatever is final, but thousand times better than nothing. As a matter of fact, I can find at least 4 *minor* things that need improvement. And that's key, here, I think: the feature works very well as is, and apart from the issues reported by @YaManicKill, others are nitpickings that can be resolved later. Well, I know I'll be opening issues on the things that bug me, at least :-)

So, in a nutshell, @YaManicKill, @maxpoulin64, are you still 👍 on this, as a very decent first version?

williamboman:
> How is that bad UX?!

I meant that the unread marker disappearing itself is not bad UX, but the way an entire chunk of messages repositions themselves while doing so. Creates a momentarily disconnect with the content of the chat.

PR #334, opened 2016-05-14
Fix notifications in query windows

### LINKED ISSUE #333: Inconsistent highlights in query windows
When I fixed highlights to be handled on the server, it somewhat broke the query window behaviour.

Currently it acts as if it were a channel (only nick mentions highlight), but also highlight the first message when the window is created.

I think easiest solution would be to set `highlight` on all incoming message if they arent `self`, and then don't apply `highlight` class on the client in query windows. This approach would also remove highlights on your nick since that's a bit useless in query windows.

xPaw:
Fixes #333 with a little trickery.

Makes notifications always appear for query messages. Also makes `highlight` class to be never applied in query windows (even if your nick is mentioned).

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -868,9 +868,9 @@ button,
 
 #chat .error,
 #chat .error .from,
-#chat .highlight .from,
-#chat .highlight .text,
-#chat .highlight .user {
+#chat .channel .highlight .from,
+#chat .channel .highlight .text,
+#chat .channel .highlight .user {
 	color: #f00;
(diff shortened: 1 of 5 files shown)


astorije:
Code looks good to me, I'll let @maxpoulin64 do further testing as he was planning.
In particular, I wanted to check that query messages still notify the browser, which is correct since `highlight` is kept, nice!

👍

astorije:
@thelounge/maintainers, anyone?

PR #339, opened 2016-05-16
Add tab complete on mobile when tapping our nick

maxpoulin64:
Not exactly the most intuitive thing, but it was an easy fix at least for now and feels like the most obvious place to place it.

I had to trigger it that way because the tab complete plugin binds directly to the keyboard on the element, so there's no way to trigger it externally.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -595,6 +595,12 @@ $(function() {
 		.history()
 		.tab(complete, {hint: false});
 
+	$("#nick").on("touchstart", function(e) {
+		e.type = "keydown.tabcomplete";
+		e.which = 9;
+		input.triggerHandler(e);
+	});
+
 	var form = $("#form");
 
 	form.on("submit", function(e) {


YaManicKill:
Hmmm, I was thinking about this the other day, and wondering if what makes most sense is adding an entry to the context menu for users to mention them? I'm not sure this is obvious at all.

maxpoulin64:
Maybe we could do both. It really is not obvious, but it does work very well. It's well placed and responds really quick: feels just like a real tab key! I agree it's not ideal but I don't really see a downside to including it. I'll be happy to wait for plugins if judged too obscure.

But I dare you to try it and not want it after :P

YaManicKill:
Ok, I'll take that challenge. I'm going to change my running instance to on the git repo, so I'll try it out sometime :-P

astorije:
This seems very cryptic UX-wise... From a user standpoint, why clicking on your own nick to cycle through completion for other nicks? What about other potential features unavailable to mobile, such as going back in history?

Instead, what about a button icon for this on the right of the input (like the Send button) that appears when you start typing, and maybe only on mobile? We can use http://fontawesome.io/icon/users/ or http://fontawesome.io/icon/at/ for that I believe.

astorije:
Actually, the more I think about this, the less I think we should put it there. When we start adding visual elements that match IRC commands (similar to the <kbd>&times;</kbd> buttons to close a channel/chat), a good place to change a nick is to actually click on the user's nick...

maxpoulin64:
Alright, no tab complete then.

astorije:
Well, I wasn't suggesting no tab complete at all either :(

PR #344, opened 2016-05-23
Turn favicon red on page load if there are highlights

xPaw:
Fixes favicon not turning red on page load if user has unread highlights.


astorije:
Yes! Good catch!
👍 and merging.

PR #345, opened 2016-05-25
Don't send completely empty messages

maxpoulin64:
Little oops due to #320, sending white space is fine but sending completely empty messages isn't. Fixing both on the server and the client to prevent unneeded spamming of the server if something is holding the key and key repeat kicks in.

Bug pointed out by @MaxLeiter on IRC.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -604,11 +604,18 @@ $(function() {
 	form.on("submit", function(e) {
 		e.preventDefault();
 		var text = input.val();
+
+		if (text.length === 0) {
+			return;
+		}
+
 		input.val("");
+
 		if (text.indexOf("/clear") === 0) {
 			clear();
 			return;
 		}
+
 		socket.emit("input", {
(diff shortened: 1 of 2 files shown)


YaManicKill:
Good catch, nice simple fix. Merging because small bug, and causing people issues who are on master right now.

YaManicKill:
:+1:

astorije:
Just for the record: https://github.com/thelounge/lounge/pull/320#issuecomment-220252379

YaManicKill:
@astorije I think we thought that was referring to messages with only spaces ([comment](https://github.com/thelounge/lounge/pull/320#issuecomment-218381212)) as that's a legitimate message in terms of IRC, but a completely empty one shouldn't be (and would mean you can accidentally just press return twice and accidentally send a message). I think there's a difference there.

astorije:
I see, sorry I missed that, thanks for the clarification.

PR #346, opened 2016-05-25
Keep chat stickied to the bottom on resize

maxpoulin64:
Fixes the chat not staying at the bottom when opening the on-screen keyboard on mobile.

### DIFF
--- a/client/js/libs/jquery/stickyscroll.js
+++ b/client/js/libs/jquery/stickyscroll.js
@@ -1,29 +1,48 @@
 (function($) {
 	$.fn.unsticky = function() {
-		return this.unbind(".sticky");
+		return this.trigger("unstick.sticky").unbind(".sticky");
 	};
 
 	$.fn.sticky = function() {
 		var self = this;
 		var stuckToBottom = true;
+		var lastStick = 0;
 
+		var keepToBottom = function() {
+			if (stuckToBottom) {
+				self.scrollBottom();
+			}
+		};
+
+		$(window).on("resize.sticky", keepToBottom);
 		self
-			.on("scroll.sticky", function(e) {
-				stuckToBottom = self.isScrollBottom();
+			.on("unstick.sticky", function() {
+				$(window).off("resize.sticky", keepToBottom);
 			})
-			.on("msg.sticky", function() {
-				if (stuckToBottom) {
+			.on("scroll.sticky", function() {
+				// When resizing, sometimes the browser sends a bunch of extra scroll events due to content
+				// reflow, so if we resized within 250ms we can assume it's one of those. The order of said
+				// events is not predictable, and scroll can happen last, so not setting stuckToBottom is
+				// not enough, we have to force the scroll still.
+				if (stuckToBottom && Date.now() - lastStick < 250) {
 					self.scrollBottom();
+				} else {
+					stuckToBottom = self.isScrollBottom();
 				}
 			})
+			.on("scrollBottom.sticky", function() {
+				stuckToBottom = true;
+				lastStick = Date.now();
+				this.scrollTop = this.scrollHeight;
+			})
+			.on("msg.sticky", keepToBottom)
 			.scrollBottom();
 
 		return self;
 	};
 
 	$.fn.scrollBottom = function() {
-		var el = this[0];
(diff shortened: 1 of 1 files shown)


client/js/libs/jquery/stickyscroll.js
```diff
@@ -1,11 +1,19 @@
 (function($) {
 	$.fn.unsticky = function() {
+		$(window).off("resize.sticky", this.data("window-sticky-fn"));
 		return this.unbind(".sticky");
 	};
 
 	$.fn.sticky = function() {
 		var self = this;
 		var stuckToBottom = true;
+		var windowResizeHandler = function() {
```
xPaw: Rename it to something more generic and make `"msg.sticky"` event use this callback too (it's same code).

client/js/libs/jquery/stickyscroll.js
```diff
@@ -16,7 +24,8 @@
 					self.scrollBottom();
 				}
 			})
-			.scrollBottom();
+			.scrollBottom()
+			.data("window-sticky-fn", windowResizeHandler);
```
xPaw: Don't really think we need the data attribute, can you just use a variable in the anonymous scope.

maxpoulin64: Can't because `.sticky` might get called multiple times, so we have to pass it somehow accross the two functions.

maxpoulin64:
@xPaw Do you like that version better?

xPaw:
Sure. Need to test it now.

client/js/libs/jquery/stickyscroll.js
maxpoulin64: We do unbind it, but I think it's better to not assume so to avoid weird surprises. I don't usually expect a jQuery plugin bound on an element to have an effect on other elements... Someone unaware of that quirk would get really confused at our little library :)

maxpoulin64:
It works decently well for its intended use. I tried to make it a bit more solid when doing weird violent horizontal resizing, but it still manages to unstick once in a while. Probably a race condition between `resize` and `scroll` not firing at the same rate that end up unstickying it.

astorije:
Just checking (I haven't tested yet): if scrolling back, then switching to horizontal mode, do we properly stay where we were, instead of being sent to the bottom?

maxpoulin64:
@astorije: yes, it only restick to bottom if it already was. If you scroll up it doesn't do anything (same as when receiving a message in the channel which uses the same function).

xPaw:
I've tested this, and it does work. As you already noted, horizontal resizing causes it to unstick. I'm not sure if there's any easy solution to this problem, but maybe try looking at it one more time?

I'm :+1: with this PR, even with horizontal bugginess.

maxpoulin64:
@xPaw: Fixed it. I'm not a fan of using time like that, but since the issue is effectively an async timing issue, that's the only solution. It still will unstick if the browser lags for more than 250ms during the resizing. According to some logging I did, it's very, very unpredictable. 250ms should cover most cases, but it's still possible it unsticks. I also tried with timers and other things, but in the end it doesn't change that we have a race condition between the time the `resize` event is triggered, and the time the `scroll` event is triggered when the rendering engine is done reflowing the page. Yay async!

client/js/libs/jquery/stickyscroll.js
```diff
@@ -1,29 +1,48 @@
...
+				if (stuckToBottom && Date.now() - lastStick < 250) {
 					self.scrollBottom();
+				} else {
+					stuckToBottom = self.isScrollBottom();
 				}
 			})
+			.on("scrollBottom.sticky", function() {
+				stuckToBottom = true;
+				lastStick = Date.now();
+				this.scrollTop = this.scrollHeight;
```
xPaw: This variable is never used

maxpoulin64: What variable? `this` in this context is the element, so this scrolls to the bottom. Would be silly to wrap `this` in jQuery to end up doing the exact same thing.

xPaw: Oh, does the `el.scrollTop` access this? Didn't think of that.

PR #348, opened 2016-05-26
Adds ban and kick to use context menus, closes #347

### LINKED ISSUE #347: Ban/kick entry in user context menu
Would be nice for ops to have a way to ban, kick or kickban a user from the context menu.

MaxLeiter:
menus


client/js/lounge.js
```diff
@@ -750,6 +763,18 @@ $(function() {
 		case "user":
 			$(".channel.active .users .user[data-name=" + $(this).data("data") + "]").click();
 			break;
+		case "user-ban":
+			socket.emit("input", {
+				target: chat.data("id"),
+				text: "/mode +b " + $(this).data("data")
+			});
+		// eslint-disable-next-line no-fallthrough
```
maxpoulin64: Note: I had him do this is to avoid code duplication, as we want to kickban on ban, so we let it fall through. Ban without kick is a valid operation, but is counter to what most people would expect. Might want to rename the menu entry to reflect that if others thinks it's needed.

dgw: I think the menu entry should reflect everything clicking it will do. Since it will both ban and kick the target user, it should read e.g. "Ban & kick".

YaManicKill: I would prefer we had 3 seperate options (as there is an example of wanting to ban and not kick)

Kick, Ban and Kickban (and I would say "Kickban" rather than "Ban & Kick" because /kickban is a custom command on man servers, so people will know what kickban means).

dgw: I was assuming only these two options were wanted. Having all three would be ideal, yes.

maxpoulin64: :+1: for the 3 options. I thought we prefered to keep it as simple as possible, but I also prefer to have the 3.

PR #354, opened 2016-05-28
Add "!" modechar for admin

omnicons:
For the odd network out that uses it!

### DIFF
--- a/client/js/libs/handlebars/modes.js
+++ b/client/js/libs/handlebars/modes.js
@@ -3,6 +3,7 @@ Handlebars.registerHelper(
 		var modes = {
 			"~": "owner",
 			"&": "admin",
+			"!": "admin",
 			"@": "op",
 			"%": "half-op",
 			"+": "voice",


maxpoulin64:
```
23:10	Dusk	Is there any plan to support  alternate modechars here btw?
23:11	+Max-P	What's that?
23:11	Dusk	well the standard set of mode characters are ~&@%+
23:11	+Max-P	Ah, that
23:11	Dusk	but as a person who uses and helps run the elemental test network; elemental uses ~!@%+
```

As I expressed on IRC, we still have no way of knowing what names are attached to those modes, but since it's the only difference I think it's fine to add this one to our hardcoded list. 👍

dgw:
Oh how I long for `PREFIX` parsing.

maxpoulin64:
@dgw: we do parse `PREFIX`. This is purely for the fancy names on the user list.

dgw:
Hmm… My mistake. I guess there's no `ISUPPORT` value for those.

astorije:
I think I'm 👍, but I would prefer to have @xPaw's input on this.

xPaw:
Yeah I don't know where `!` came from. I would like to see at least a couple of networks using it before merging this.

astorije:
@xPaw, I think the OP (or someone else) mentioned 2-3 networks that need this: the elemental ircd test network, ponychat, and another one I forgot.
What do you think?

PR #356, opened 2016-05-29
Adds kick, ban, and kickban to user context menus, closes #347

### LINKED ISSUE #347: Ban/kick entry in user context menu

MaxLeiter:
image

Any suggestions on the icons? Should ban and kickban be the same?

As a note, `ban` simply sets `mode +b`  on the selected user. 

I'm not sure if it's worth including `applyCommandToUser`, but otherwise there would be a lot of code duplication (`socket.emit {}`...)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -559,6 +559,24 @@ $(function() {
 				text: target.text(),
 				data: target.data("name")
 			});
+			if (target.data("name") !== $("#nick").text()) {
+				output += render("contextmenu_divider");
+				output += render("contextmenu_item", {
+					class: "user-kick",
+					text: "Kick",
+					data: target.data("name")
+				});
+				output += render("contextmenu_item", {
+					class: "user-ban",
+					text: "Ban",
+					data: target.data("name")
+				});
+				output += render("contextmenu_item", {
+					class: "user-kickban",
+					text: "Kickban",
+					data: target.data("name")
+				});
+			}
 		} else if (target.hasClass("chan")) {
 			output = render("contextmenu_item", {
 				class: "chan",
@@ -750,6 +768,10 @@ $(function() {
 		return false;
 	});
 
+	function applyCommandToUser(command) {
+		socket.emit("input", {target: chat.data("id"), text: command + " " + $(this).data("data")});
+	}
+
 	contextMenu.on("click", ".context-menu-item", function() {
 		switch ($(this).data("action")) {
 		case "close":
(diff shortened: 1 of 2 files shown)


client/js/lounge.js
```diff
@@ -559,6 +559,24 @@ $(function() {
 				text: target.text(),
 				data: target.data("name")
 			});
+			if (target.data("name") !== $("#nick").text()) {
```
dgw: Is there no better place to get the user's current nick from?

dgw:
I take it there will be no kick message sent when using these menu entries?

Instead of banning the nick, it might make more sense to ban the user's host. Otherwise the bans are trivial to evade. (Ban evasion is against most networks' rules, but people who get banned tend to ignore those anyway.)

Other than those two items, looks good to me. :+1:

Finally, I'd like to note that this could have been rebased and force pushed on the preexisting #348 rather than opened as a new PR, to keep the discussions together.

maxpoulin64:
> it might make more sense to ban the user's host. 

Agreed, but it would be nice if we could check if the user has a valid ident and use that instead of the host when possible to avoid banning whole bouncers at once.

> Finally, I'd like to note that this could have been rebased and force pushed on the preexisting #348 rather than opened as a new PR, to keep the discussions together.

Mistake by @MaxLeiter, he deleted the branch trying to clear a mess he made and that had the side effect of irreversibly closing the PR. I can't reopen it.

dgw:
>> it might make more sense to ban the user's host.

> Agreed, but it would be nice if we could check if the user has a valid ident and use that instead of the host when possible to avoid banning whole bouncers at once.

A username with a `~` prefix generally indicates that it is unverified; idents verified by the IRC server do not have a prefix. That said, I haven't yet been able to find any RFC that specifies how IRC servers should differentiate between verified and unverified idents. IMO the best approach for now would be banning `*!user@remote.host` if the ident does not have a `~` prefix, and `*!*@remote.host` if it does.

>> Finally, I'd like to note that this could have been rebased and force pushed on the preexisting #348 rather than opened as a new PR, to keep the discussions together.

> Mistake by @MaxLeiter, he deleted the branch trying to clear a mess he made and that had the side effect of irreversibly closing the PR. I can't reopen it.

Ah, deleting the branch *would* mess things up. That's too bad, but if Github won't let it be reopened then that's that.

client/js/lounge.js
MaxLeiter: Not that @Max-P or I could find /:

MaxLeiter:
Yeah apologies about the duplicate PR, I actually kept the branch but removed the changes while trying to sync with master (in order to use a recently merged PR).

MaxLeiter:
Any tips or ideas on fetching users hostmask/ident @maxpoulin64?

astorije:
Is this still a `do not merge`, @MaxLeiter?

MaxLeiter:
@astorije this will require a large refactor on the client <==> server in order to fetch things like a users hostname, so it may be worth closing?

astorije:
> so it may be worth closing?

What say you, @maxpoulin64?

MaxLeiter:
This isn't going to be merged anytime soon - closing.

PR #357, opened 2016-05-29
New nick colors (again)

### LINKED ISSUE #335: Update some IRC colors to be readable in all themes
For example blue and purple colors in Morning themes are barely readable due to being too dark.

We should check that all the colors we define are readable and work in themes we ship (not necessarily pass AA, but not completely dark).

maxpoulin64:
This one was generated by a script designed to spread all colors evenly and adjust the brightness dynamically when the contrast ratio is too low. The end result is that we have all the colors adapted for each theme.

Sample:
compare

Should fix #335, which is what pushed me to do this (when rendering the list of colors, there were very close duplicates and overall the colors were kinda meh).

32 colors is a lot, it's very hard to not have duplicates or too close colors without going for the whole spectrum.

Generator script for reference or tweaking:

``` javascript
var Color = require("color");

var background = Color("#3f3f3f");
var backgroundL = background.luminosity();

function printcolor(n, color) {
    console.log("#chat.colored-nicks .user.color-"+ ((n < 10) ? n + " " : n)
        + " { color: " + color.hexString() + "; }");
}

function genrainbow(start, end, gencolor, adjust) {
    for(var i = start; i <= end; i++) {
        var fact = (i-start)/(end-start);
        printcolor(i, adjust(gencolor(fact), i));
    }
}

genrainbow(
    1, 32,
    (fact) => Color().hsl(fact*310, 100, 50),
    (color, n) => {
        var contrast = color.contrast(background);
        var luminosity = color.luminosity();

        if(contrast < 2.5) {
            if(backgroundL > luminosity) {
                color.darken(contrast/5);
            }
            else {
                color.lighten(contrast/3);
            }
        }

        return color;
    }
);
```


xPaw:
Some of the colors (especially in middle of each unique color) look awfully close to each other. Like almost 22-30 range on white looks very similiar.

dgw:
As @xPaw said, a lot of the colors are very close. In the white column, 12-15 and 16-18 group visually when I'm scanning down the list, as do 24-26, 27-29, and 30-32. They mostly seem to group in clusters of 3, for me.

On the dark backgrounds, the greens run together, and color 30 in the middle column is basically unreadable unless I squint hard. (Really, those last 3 colors look pretty ugly to me on both dark backgrounds.) Color 1 (pure red?) is pretty glaring too on a dark background, more so on the bluish than the gray.

Could it be that the script needs to change the colors a bit more when compensating for low contrast?

maxpoulin64:
Yeah, I was thinking about it a bit after and got some other ideas. I'll try generating two sets: a lighter and a darker variant. Colors will be further apart and easier to distinguish. I should have made a copy of the current palette, it's even worse. I'll re-render that tomorrow.

astorije:
> 32 colors is a lot, it's very hard to not have duplicates or too close colors without going for the whole spectrum.

Do you guys want to reduce the number of colors? 2 ways of doing so:
- Reducing the number of indices generated by the corresponding code
- Grouping colors 2 by 2 in the theme so that there are effectively only 16 colors.

@maxpoulin64, when people are happy with the colors, before merging, don't forget to randomize the colors. The code that hashes strings generates successive integers for successive strings of same lengths (`lorema` -> N, `loremb` -> N+1, `loremc` -> N+2, ...). I tried different ways, but actually even JS ports of `hashCode` does so.

xPaw:
> The code that hashes strings generates successive integers for successive strings of same lengths

You can solve this by using something like murmurhash (it's quite simple and fast).

astorije:
> You can solve this by using something like murmurhash (it's quite simple and fast).

We could, but when I see [this JS implementation of murmurhash](https://github.com/garycourt/murmurhash-js/blob/0197ce38bedac0e05f40b9d7152095d06db8292c/murmurhash2_gc.js#L14-L50) compared to [what we have on `master`](https://github.com/thelounge/lounge/blob/7157076aad1837db408b58c6b452bf20a8493b1a/client/js/libs/handlebars/colorClass.js#L5-L12) or [your earlier implementation of `hashCode`](https://github.com/thelounge/lounge/pull/322/files#diff-1d12033fe31e161ba1cd53ac22b39929R18), I don't think we need to go that far.

To be honest, I don't think successive hashing like we have will be a problem to channels of, say, ~100-200 users anyway. It's very unlikely that most nicks will be directly successive (like `lorema` and `loremab` will have very different hashes) but it could on channels with thousands of nicks.
I don't care too much but randomizing could be needed in the latter case, so just to keep in mind if needed. Of course, no big deal in not randomizing I think.

williamboman:
Not sure what the status of this PR is but I gotta say I prefer the nick colors in master. The contrast between nicks and the rest of the UI is too strong (at least in the non-default themes), imho.

YaManicKill:
I think something like this is ... never going to end up getting merged, tbh. We have a situation where (I think) all nick colours are readable on all default themes.

Other than that, I think we need to defer to custom css if people want to do that. Thoughts?

astorije:
I agree with @YaManicKill. It seems like the current colors we have work well, so I'm more inclined to list the few colors that don't work well in #335 and fix them one by one rather than nuking everything and risking this once more.

Unless fixing these makes some colors clearly out of range (like if all colors were rather bright and we go for a dark one, but I don't think we'll have this issue on the default themes), I don't think it's worth re-engineering all colors.

@maxpoulin64, I'm going to close this PR (opened 6 months ago lol), at least to try reducing our PR list. We should feel free to re-open it if necessary though.

> It would be cool if one could adjust the contrast of nick colors through a slider in the settings or something.

I don't think we'll add a slider for all colors, that seems over-engineered to me :) Especially since I really like to leave this to the theme author: if an author creates a black and white theme, and only wants say 8 shades of gray, then that's the author choice (which was one reason to put the colors into CSS and not JS-generated).
If you don't like some of these, user can use custom CSS as @YaManicKill suggested, and/or open a PR to submit his proposal.
In the meantime, I know that's not what you wanted exactly (though probably less buggy than anything we could come up with lol), you can use the color slider from the browser dev tools:

screen shot 2016-11-30 at 00 55 30
<br>
<br>
screen shot 2016-11-30 at 00 55 37

williamboman:
> I don't think we'll add a slider for all colors, that seems over-engineered to me :)

I meant a single slider that controls the contrast of *all* nick colors.

PR #359, opened 2016-05-29
Move Shout theme borders to example theme

xPaw:


astorije:
I'm not too fond of this, really. I don't think we should make all our themes look alike.
The unnecessary borders are actually a nice design choice I think. Sure, they aren't necessary for The Lounge to work, but that's not the point of a theme :-)
They do make it for a nice theme. If a more "pure" version should arise, more flat-UI style, with less borders, etc., then we're talking about a new theme.

However, I do believe you have a point regarding all non-default themes that constantly cancel what the default one does. To solve this, we should start moving all design-specific items from `style.css` to `example.css`. I wanted to start this with the nick colors but then forgot.
That way, in the not-so-long term, `style.css` contains items that are necessary for the app itself, and theme files do contain what's specific to the theme their author has designed. Also, themes having more independence means less risk to break them by changing UI elements for the default theme.
Finally, as @maxpoulin64 keeps saying 😄, doing so will be a step towards theme management.
Now, I'm not saying this should be in one go. For example, when I get to deal with https://github.com/thelounge/lounge/issues/335#issuecomment-219295830, I could move the colors to the theme files instead...

xPaw:
The point of this PR was not only to make it easier for themes to change styles, but also to remove the unnecessary padding. I can hardly agree that it makes for a nice design choice. Take a look at a lot of other chat software, they try to give the chat window as much space as reasonably possible.

maxpoulin64:
I agree with @xPaw on this one. The simpler the base theme the better, as every other theme is based on that. Currently, themes have to fight with every detail we add to the base them to remove them. I'd rather have to use a theme (we could even call it Shout, as that's where it comes from) to keep it than have the default theme be a problem for the other themes.

Plus, it would also help The Lounge set itself further apart from Shout. So personal tastes aside, that's still a 👍 for me.

astorije:
Moving design-specific stuff away from `style.css` is a mandatory step anyway if we really want to be serious about theme management. Until we do so, all themes will always have to cancel out what `style.css` does.
The maintenance burden will not necessarily be more costly as what we have so far (think of all the times we had to ask PR authors "can you add this to other themes too?" or "did you test this with other themes to make sure it doesn't break them?")

I'll test this before merging.

JustRamon:
I like the looks.

PR #360, opened 2016-05-29
Simplify text input form

xPaw:
Removes the double wrapping we had. Tell me if it's better or worse.

29-2048-2b03670

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1170,60 +1170,36 @@ button,
 }
 
 #form {
-	background: #eee;
+	background: rgba(0, 0, 0, .02);
 	border-top: 1px solid #ddd;
 	bottom: 0;
 	height: 48px;
+	line-height: 48px;
 	left: 0;
 	position: absolute;
 	right: 0;
-}
-
-#form .inner {
-	bottom: 7px;
-	left: 7px;
-	position: absolute;
-	right: 7px;
-	top: 6px;
-}
-
-#form .input {
-	font: 12px Consolas, Menlo, Monaco, "Lucida Console", "DejaVu Sans Mono", "Courier New", monospace;
-	left: 0;
-	height: 34px;
-	position: relative;
+	font-size: 14px;
 }
 
 #form #nick {
-	background: #f6f6f6;
+	background: #ddd;
 	color: #666;
 	position: absolute;
 	font: inherit;
-	font-size: 11px;
-	margin: 5px;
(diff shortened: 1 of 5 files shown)


JustRamon:
Better IMO

astorije:
Same comment as #359, the current one fits very well in the theme. I'd be more in favor of a new "flat-UI" theme that matches your preference.

williamboman:
The placeholder color in Zenburn is a little hard to read, possibly add;

```diff
diff --git a/client/themes/zenburn.css b/client/themes/zenburn.css
index fd13d55..275efa7 100644
--- a/client/themes/zenburn.css
+++ b/client/themes/zenburn.css
@@ -183,6 +183,10 @@ body {
 	color: #dcdccc;
 }
 
+#form #input:-ms-input-placeholder { color: #a1a197; }
+#form #input::-moz-placeholder { color: #a1a197; }
+#form #input::-webkit-input-placeholder { color: #a1a197; }
+
 /* Buttons */
 #chat .show-more-button {
 	background: #434443;

```


screenshot 2016-05-30 at 18 20 28

dgw:
Those selectors should be comma-separated so as not to duplicate the `color` property, but I agree with changing that color.

williamboman:
> Those selectors should be comma-separated so as not to duplicate the color property, but I agree with changing that color.

Not possible due to vendor incompatible pseudo-selectors (in accordance with the CSS spec).

dgw:
Surprising that the spec doesn't allow for ignoring only elements of comma-separated selectors, since each list item is independent of the others, but so be it; the W3C hath spoken.

YaManicKill:
@astorije Are you against this going in as-is and want a new theme instead? Personally, I think it works fine in the theme, but if you are very against it, then we should probably not leave it hanging around.

astorije:
@YaManicKill, if everyone is in favor, then I'll go with the flow :-) I'll review/test it ASAP.

maxpoulin64:
Alright, since people seems to like it, this is a 👍 for me. It will also make #379 look a bit better, as right now the box is bottom centered and it's a bit weird looking.

capture d ecran_2016-07-01_20-18-42

capture d ecran_2016-07-01_20-32-52

xPaw:
@maxpoulin64 will have to be rebased after your PR is merged, but yea.

williamboman:
Nooo what happened here?

PR #355, opened 2016-05-28
Updates Font-Awesome to v4.6.3

MaxLeiter:
Hopefully, we can eventually use the font-awesome package on npm...

PR #361, opened 2016-05-30
Fetch Font Awesome from npm instead of embedded in repo

astorije:
After [the recent upgrade of Font Awesome](https://github.com/thelounge/lounge/pull/355), I figured I should make this a reality.

A few notes:
- Lines in `client/css/style.css` are directly copied from `node_modules/font-awesome/css/font-awesome.css` (minus changing spaces into tabs, and damn quotes...).
- Fonts are now installed during `npm install` or `npm run build`.
- At the moment, I am fairly confident `npm run build:font-awesome` only runs on Mac OS X and Linux, not on Windows<sup>1</sup>. There are so many packages to copy cross-OS that could help with this: `copyfiles`, `copy-files`, `cpx`, ... If any passer-by has used any, please advise. Otherwise, marking as `do not merge` until I fix for Windows, gathering comments here meanwhile.

<sup>1</sup> It would be interesting to set up AppVeyor CI if we want to ensure the project builds fine on Windows at all time.


astorije:
Of course, this is not trying to solve asset management for the entire project. [We tried to solve everything at once](https://github.com/thelounge/lounge/pull/62) and did not succeed to finish before it became out of control.
I prefer to have a step-by-step approach. The build step is just mimicking the existing (one path away), but the goal was to move from npm.

package.json
```diff
@@ -13,7 +13,8 @@
   "homepage": "https://thelounge.github.io/",
   "scripts": {
     "start": "node index",
-    "build": "npm run build:grunt && npm run build:handlebars",
+    "build": "npm run build:font-awesome && npm run build:grunt && npm run build:handlebars",
+    "build:font-awesome": "mkdir -p client/fonts && cp node_modules/font-awesome/fonts/fontawesome-webfont.* client/fonts",
```
xPaw: We're starting to abuse npm scripts with stuff like this. Also not a particular fan of hardcoding the path like this.

maxpoulin64: Yeah... Instead of copying files and running scripts, couldn't we make it so the server builds/serves all that stuff on demand? That way it would also leave room for automatic rebuilds and ship more stuff from npm.

Or maybe make one script that deals with all the building without using the shell so much...

astorije: > We're starting to abuse npm scripts.

Not at all, this is a very appropriate use of npm scripts. (Edit: Funny, Googling "npm scripts" bring "Why I Left Gulp and Grunt for npm Scripts" as first result :D)
For reference, I highly recommend reading http://blog.keithcirkel.co.uk/why-we-should-stop-using-grunt/ and http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/.

> Also not a particular fan of hardcoding the path like this.

I get you, but problem is, we have to somewhere. We even already do it [in our current Gruntfile](https://github.com/thelounge/lounge/blob/7157076aad1837db408b58c6b452bf20a8493b1a/Gruntfile.js#L15). I'm open to suggestions to make this better though.

> Yeah... Instead of copying files and running scripts, couldn't we make it so the server builds/serves all that stuff on demand?

I think building static assets on demand is overkill. Serving from demand also has its set of trade-offs, like the fact that the package needs to be installed by users as part of the `dependencies` (not saying it's good or bad, just an observation).

All in all, you both have good points for sure. My take on this is that something like this PR is still times better than shipping dependencies with the repo or going through Grunt with N new wrappers/plugins for it and their according DSLs.

@xPaw, @maxpoulin64, what do you guys say I fix this on Windows, we make this fetching-from-npm happen, then we improve the build mechanism? When I look at the sad fate of #62 solve-everything-at-once, I'd rather improve things step by step, or nothing will never happen. I'm not saying this solution is set in stone, but at least we can iterate on this.

xPaw: I guess the solution will do for now. I would really prefer if we made scripts in JS in the future for all the build/test steps.

maxpoulin64:
@astorije pointed an issue on IRC:

```
19:42	astorije	One thing I'm not too fond of is that it forces asset packages to be installed as dependencies and not devDependencies anymore, and I'm not sure it's great to have entire repositories being pulled for this every time you install in production
19:43	+Max-P	astorije: that's a fair point for the devDependencies. Does packages like font-awesome really include more than just the fonts?
19:43	astorije	But these are follow-up discussions to something that improves something that improves something that... Can we just go ahead with merging #361 and improve later, even if that later means a day or an hour later?
19:44	astorije	Max-P, yes, non-webfonts, SCSS, CSS, LESS
19:44	astorije	And if we do jQuery, it includes a looot of things
```

So, seems like a very fair point to me. Considering when we publish to npm, we ship a prebuild package, I'm fine with this as it is for now. It's not ideal, but it does make the idea of custom routing fall appart. I will give my 👍 for now, and we'll move on to a better solution when we have one.

astorije:
> 19:42 astorije One thing I'm not too fond of is that it forces asset packages to be installed as dependencies and not devDependencies anymore, and I'm not sure it's great to have entire repositories being pulled for this every time you install in production

Following-up on previous comment, it looks like this PR has been merged with `font-awesome` package being a devDependency. Not sure what I changed there but oh well.

PR #366, opened 2016-05-30
Improve click handling on users and inline channels

### LINKED ISSUE #362: Clicking on a user with an existing query window open doesn't bring focus to it
If you have a user query window open, clicking on it does a whois, but doesn't bring the focus to that query window. Clicking on a channel brings focus to that channel, and this is what I would expect.

### LINKED ISSUE #363: Clicking channel names with different case not handled correctly
Seeing @YaManicKill mention (in #362) focusing channels on clicking them made me test that behavior. I sent these messages to my testing channel:

````
07:01	~dgw	test ref to #kaede for thelounge debugging
07:01	~dgw	also test ref to #Kaede (caps)
````

Then I clicked on each of them. Clicking `#kaede` (all lowercase) focused the existing channel buffer. Clicking `#Kaede` (the channel's proper name, with a capital K) did *not* focus the existing channel buffer. Instead, it inserted a fake join event and incremented the unread counter:
screenshot 2016-05-30 07 07 22

This isn't correct behavior. ~~I don't really agree with The Lounge lowercasing all displayed channel names anyway (especially because it leaves nickname case alone), but if it's going to be done it should not affect how the channels are handled.~~ Display casing is fixed in 2.0.0-pre.1 but the issue with clicking differently cased channel names persists.

### PR DESCRIPTION
**Pull request by xPaw**
1. Fixes #362 
2. Partially (?) fixes #363
3. Removes the dumb `whois` global variable which is prone to race conditions
4. Fixes clicking on inline channels in different case sens

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -206,16 +206,18 @@ $(function() {
 			})
 		);
 		renderChannel(data.chan);
-		var chan = sidebar.find(".chan")
+
+		// Queries do not automatically focus, unless the user did a whois
+		if (data.chan.type === "query" && !data.shouldOpen) {
+			return;
+		}
+
+		sidebar.find(".chan")
 			.sort(function(a, b) {
 				return $(a).data("id") - $(b).data("id");
 			})
-			.last();
-		if (!whois) {
-			chan = chan.filter(":not(.query)");
-		}
-		whois = false;
-		chan.click();
+			.last()
+			.click();
 	});
 
 	function buildChatMessage(data) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by maxpoulin64** [comment]
👍 seems all right to me, and one global variable down!

File: client/js/lounge.js [hunk 1]
```diff
@@ -206,16 +206,18 @@ $(function() {
 			})
 		);
 		renderChannel(data.chan);
-		var chan = sidebar.find(".chan")
+
+		// Queries do not automatically focus, unless the user did a whois
```
YaManicKill: Wait, so does this mean if I click on a user and I already have a query open with them, it won't get focus? That seems confusing.

xPaw: It will, see line 666

YaManicKill:
Alright, I think this looks good to me then. Feel free to merge if you are ready for it to go in. :+1:

PR #367, opened 2016-05-30
CI tool for Windows builds

astorije:
History has shown that we have broke our build/test environment on Windows before, see https://github.com/thelounge/lounge/pull/309.

This is just a smoke test to make sure we don't do this again. For example, things like #361 need that kind of things, otherwise we'll always have to ask "can someone test on Windows please?" :-)

Regarding Node version, I went for latest LTS. I know @maxpoulin64 is in favor of specifying latest. I don't have a preference, I'll go with whatever you guys want. Any of them is still better than none!
Note that AppVeyor does not parallelize and is rather super slow, so we should stick to one.

Also, you'll notice badge says `astorije/lounge`. That's because stupid AppVeyor doesn't reflect GitHub organizations, doh. If that's a huge deal for you, I'll find a way (like creating a secondary account, but what a pain), otherwise I think it's good enough for our needs.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+---
+# http://www.appveyor.com/docs/appveyor-yml
+
+# Build version format
+version: "{build}"
+
+# Do not build on tags (GitHub only)
+skip_tags: true
+
+environment:
+  nodejs_version: '4'
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - npm install
+
+test_script:
+  - node --version
+  - npm --version
+  - npm test
+
+# Don't actually build
+build: off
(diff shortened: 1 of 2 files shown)


astorije:
Seriously, Travis CI has had time to build 4 versions that AppVeyor hasn't even started one. Definitely only one version in its conf or we'll never wait for the build to finish before merging :-)

maxpoulin64:
Looks good to me, not always easy to get a Winblows handy to test on 👍

appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+  nodejs_version: '4'
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - npm install
+
+test_script:
+  - node --version
+  - npm --version
+  - npm run test:mocha
```
xPaw: Change these three `npm run` to just `npm run test`

YaManicKill: > Change these three npm run to just npm run test

Also, you can just do "npm test" as that's a link in npm to "npm run test"

astorije: True, we lose reports if one fails (as Travis CI does), but the whole point of this is to make sure `npm install` and `npm test` do not break for Windows...

Done in c886657e6b1333ac10c269ada680e3b3316a96e2.

PR #368, opened 2016-05-31
Remove design-specific nickname colors from base system style

astorije:
Because I should have done that in #325 directly.

Base style should be system-specific and not design specific, so this is one step towards that.

When all themes are independent from the base style, it will be easier to create new ones, maintain overall compatibility with the existing ones, export the current ones into theme repos, etc.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -746,45 +746,6 @@ button,
 	color: #50a656;
 }
 
-/* Nicknames */
-
-#chat .user {
-	color: #50a656;
-}
-
-#chat.colored-nicks .user.color-1 { color: #1396cf; }
-#chat.colored-nicks .user.color-2 { color: #ffcf89; }
-#chat.colored-nicks .user.color-3 { color: #00dc5f; }
-#chat.colored-nicks .user.color-4 { color: #ff5bc8; }
-#chat.colored-nicks .user.color-5 { color: #ff0e1b; }
-#chat.colored-nicks .user.color-6 { color: #000094; }
-#chat.colored-nicks .user.color-7 { color: #006441; }
-#chat.colored-nicks .user.color-8 { color: #00566e; }
-#chat.colored-nicks .user.color-9 { color: #ff0078; }
-#chat.colored-nicks .user.color-10 { color: #15d5a3; }
-#chat.colored-nicks .user.color-11 { color: #006b3b; }
-#chat.colored-nicks .user.color-12 { color: #00c5ba; }
-#chat.colored-nicks .user.color-13 { color: #00465b; }
-#chat.colored-nicks .user.color-14 { color: #ffafce; }
-#chat.colored-nicks .user.color-15 { color: #ff3b12; }
-#chat.colored-nicks .user.color-16 { color: #16cc6a; }
-#chat.colored-nicks .user.color-17 { color: #ff0072; }
-#chat.colored-nicks .user.color-18 { color: #ff5877; }
-#chat.colored-nicks .user.color-19 { color: #ff1753; }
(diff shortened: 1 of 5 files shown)


astorije:
Lol, thanks AppVeyor, but you'll obviously fail until #367 gets merged, 👏 👏 .

maxpoulin64:
Hmm, I'm not sure this is necessary, at least considering the current status of themes. Since our themes are all based into modifying the base theme, it doesn't make much sense to make colors an exception. I think this is the kind of transition that should be done all at once, for consistency. It's a bit silly for new themes to have to copy over the colors, but override just the background and other things.

astorije:
> Hmm, I'm not sure this is necessary, at least considering the current status of themes.

If we want to be serious about theme management, we'll have to do that anyway, and the earlier the better, by any means.

> it doesn't make much sense to make colors an exception

I didn't mean to make this an exception but a first step (well, second if we consider the `margin: 0` on `body` already there). I should have done this in #325 and in general, we should consider doing so when improving the default theme, step by step.

> this is the kind of transition that should be done all at once

No, no! :( We've all seen what "do it all at once" / "do it perfectly rather than just better" has done so far, on several pieces, from small features to full refactor. It takes months, pain and frustration or rebases because it's a long-standing PR, or it gets dropped, or never started, ...
It's not critical to do it all at once, so we really should improve things as they go, to converge towards the solution. At least we won't have to deal with this when we start the theme management, and theme creators can start improving their themes until we do so.

maxpoulin64:
I would suggest we make a branch for those. Individual PRs on master are a bit hard to keep track of when doing many steps like that, and it's annoying for people to continuously change their themes.

I think we should wait for the packages before we go into themes anyway. I see many cases where the theme would also want to change the HTML, which packages will also provide. Themes should inherit as they please, but also not forced to reimplement anything. I think an empty theme that inherits the default should look exactly like the default. (But a blank theme should also be totally blank, if it defines to be that way).

I really need to give the damned packages branch a bit of love. Maybe this weekend, if things go right. We'll have a much clearer idea where we're headed when we can agree on the format and move according to the engine's capabilities (which I hope should be pretty broad!)

astorije:
Again, we are refusing going for better in order to hope for the best. :-(
I get what you want to achieve, but in no way PRs like this will go against that. On the contrary, it's one step towards that, since you will need to do so anyway.

Regarding making a branch for this, count me out. Rebasing things like this is a pain, I really don't see the need.

And regarding themes that can change HTML, sure, absolutely... but still something we want, not something we have, and this doesn't go against this.

I can close this if you want, but I definitely believe it will be for the wrong reasons.

xPaw:
I think these user colors should actually stay in the base theme. It makes it easier making your own theme (if you keep a light background, you don't need to copy all the colors).

YaManicKill:
I think I'm with @xPaw on this one.

I agree that with many things we don't want to keep it in the base theme. However, for nick colours, when making a new theme, 1 of 2 things will happen:

1. You will not want any of the base colours, and therefore override all of them. The fact that they are in the base theme won't be annoying to you, because you will be setting them **anyway**. If the base theme didn't have colours, this would not change what you do
2. You will be happy with the base colours, and therefore keep them. If the base theme doesn't have colours, you would have to do the work.

Therefore (**for specifically nick colours**) I see no disadvantage to having them in the base theme. I understand what you are saying, @astorije, that for many things it can be annoying to have to override, but that wouldn't be the case here.

astorije:
OK, seems like everyone is 👎ing this, with very good points, so I'm closing it.
I still believe we might want something like that in a far future, when we have great theme management, but by then lots can change, so we'll see.

PR #369, opened 2016-05-31
Do not lose authentication token when the connection gets lost

xPaw:


client/js/lounge.js
```diff
@@ -74,28 +74,30 @@ $(function() {
...
 		var login = $("#sign-in");
 		if (!login.length) {
 			refresh();
 			return;
 		}
+
 		login.find(".btn").prop("disabled", false);
-		var token = window.localStorage.getItem("token");
-		if (token) {
-			window.localStorage.removeItem("token");
```
xPaw: This is the actual offender. I chose to never delete the token even on failed auths to show the "authorisation failed" message to the user until they relogin.

maxpoulin64:
Looks all good to me! 👍

astorije:
Testing this, I hit a weird bug. To reproduce:
- Start the server and go to the login page
- Enter the wrong password and submit
- An "Authentication failed" message appears. Reload the page and it disappears.
- Enter the right password and submit
- Restart the server and go to the login page
- The "Authentication failed" message appears again, even if not submitting anything. Reloading the page again and again will keep the message.

I am going to review #370 instead of looking further into this. I don't think it's worth fixing this in an isolated way considering it has to do with restarting the server.

xPaw:
@astorije That was actually intentional, I mentioned it in IRC. However considering it keeps doing that after entering a wrong password, I should remove the token when that happens.

astorije:
So, I definitely hit the bug many times on mobile, but I cannot consistently reproduce it on laptop right now. That being said, code looks fine, @maxpoulin64 said he tested a bit more than just reading the code, and I tested that previous behavior didn't happen (i.e. that we have iso-functionality with what we had before PR) so I'm 👍.
This is definitely a fix that will be live-tested in a prerelease. Good job @xPaw, it was an annoying bug!

PR #370, opened 2016-05-31
Implement user token persistency

### LINKED ISSUE #26: Store user token in users' config file for persistency
This line specifically: https://github.com/thelounge/lounge/blob/2201e2b941bd13427c38d6934b84ceda245aa20e/src/client.js#L69-L71

This would mean that restarting the server wouldn't require users to re-login.

xPaw:
Fixes #26.

It also regenerates the token on password change, which will force all your devices (except the one you changed the password on) to re-login.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -129,6 +129,11 @@ $(function() {
 				feedback.hide();
 			});
 		}
+
+		if (data.token && window.localStorage.getItem("token") !== null) {
+			window.localStorage.setItem("token", data.token);
+		}
+
 		passwordForm
 			.find("input")
 			.val("")
@@ -163,8 +168,10 @@ $(function() {
 			}
 		}
 
-		if (data.token) {
+		if (data.token && $("#sign-in-remember").is(":checked")) {
 			window.localStorage.setItem("token", data.token);
+		} else {
+			window.localStorage.removeItem("token");
 		}
 
 		$("body").removeClass("signed-out");
--- a/src/client.js
+++ b/src/client.js
@@ -64,10 +64,15 @@ function Client(manager, name, config) {
 		sockets: manager.sockets,
 		manager: manager
 	});
+
 	var client = this;
-	crypto.randomBytes(48, function(err, buf) {
-		client.token = buf.toString("hex");
-	});
+
+	if (!client.config.token) {
+		client.updateToken(function() {
+			client.manager.updateUser(client.name, {token: client.config.token});
+		});
+	}
+
 	if (config) {
 		var delay = 0;
 		(config.networks || []).forEach(function(n) {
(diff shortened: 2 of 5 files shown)


src/server.js
```diff
@@ -239,12 +244,12 @@ function auth(data) {
 			if (success) {
 				var token;
```
xPaw: Forgot to remove this variable as I removed the need to pass around `token` into `init`.

xPaw:
TODO: I forgot to regenerate the token when changing the password using the CLI.

src/server.js
```diff
@@ -206,7 +211,7 @@ function reverseDnsLookup(socket, client, token) {
 			client.hostname = client.ip;
 		}
 
-		init(socket, client, token);
+		init(socket, client, remember);
```
maxpoulin64: Is there a reason why this isn't just handled on the client? That way, we could get rid of the `remember` variable entirely and not need to pass it around at all. Since that flag comes from the client anyway, I don't see any harm in always returning the token.

xPaw: Would need to store it somewhere while auth is happening (or disable the checkbox). I'd rather keep it as it is for now.

client/js/lounge.js
```diff
@@ -129,6 +129,11 @@ $(function() {
 				feedback.hide();
 			});
 		}
+
+		if (data.token) {
```
xPaw: I think this will make the user remember the login when changing their password, so need to add a check for that. I guess need to check if `token` exists in localstorage or something.

client/js/lounge.js
```diff
@@ -129,6 +129,11 @@ $(function() {
 				feedback.hide();
 			});
 		}
+
+		if (data.token && window.localStorage.setItem("token") !== null) {
```
astorije: I think you meant `getItem`, no? Also, shouldn't that token change only happen if `data.success`?

src/client.js
```diff
@@ -64,10 +64,15 @@ function Client(manager, name, config) {
...
+
 	var client = this;
-	crypto.randomBytes(48, function(err, buf) {
-		client.token = buf.toString("hex");
-	});
+
+	if (!client.config.token) {
+		client.updateToken(function() {
+			client.manager.updateUser(client.name, {token: client.config.token});
+		});
```
astorije: Just curious, does this really need to be asynchronous? I'm not sure performance would be any affected if sync, but code would look a tiny bit nicer IMO. Nothing ground-breaking of course.

astorije:
@xPaw, can we add a text when changing the password (e.g. under the form for the UI, confirmation message when calling the CLI command) saying "[For security purposes,] all your other sessions will be logged out" or something like that? I'll let you or any native speakers decide on the wording.

src/client.js
xPaw: It's not worth it to change it to be synchronous.

client/js/lounge.js
xPaw: `data.token` will be sent only on `data.success` anyway, so the check won't pass. And yes it should be `getItem`.

xPaw:
@astorije Added a message when changing the password on web. It doesn't make much sense to keep in CLI (maybe worth mentioning on website documentation though).

astorije:
> @astorije Added a message when changing the password on web.

I'd rather see something showing up in advance rather than "ah ah, you are now disconnected everywhere, sucks to be you", but we can always improve later.

👍 and merging, awesome work @xPaw!

PR #375, opened 2016-06-03
package.json: use `npm-run-all` in scripts

williamboman:
- [x] Proper exit codes
- [x] Runs all the specified npm scripts sequentially (continues on error)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,9 @@
     "build:font-awesome": "node scripts/build-fontawesome.js",
     "build:grunt": "grunt",
     "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
-    "test": "(npm run test:mocha || true) && npm run lint",
+    "test": "npm-run-all -c test:mocha lint",
     "test:mocha": "mocha -r test/fixtures/env.js test/**/*.js",
-    "lint": "(npm run lint:js || true) && (npm run lint:css || true)",
+    "lint": "npm-run-all -c lint:js lint:css",
     "lint:js": "eslint .",
     "lint:css": "stylelint \"**/*.css\"",
     "prepublish": "npm run build"
@@ -63,6 +63,7 @@
     "grunt-contrib-watch": "1.0.0",
     "handlebars": "4.0.5",
     "mocha": "2.4.5",
+    "npm-run-all": "2.1.1",
     "stylelint": "6.2.2"
   }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,3 @@ deploy:
     node: '4'
     tags: true
     repo: thelounge/lounge
-
-script:
-- npm run test:mocha
-- npm run lint:js
-- npm run lint:css
-
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,9 +18,6 @@ test_script:
   - node --version
   - npm --version
   - npm test
-  - npm run test:mocha
-  - npm run lint:js
-  - npm run lint:css
 
 # cache npm modules
 cache:


xPaw:
As I pointed out in https://github.com/thelounge/lounge/pull/361#discussion_r65692618 we should switch to a proper build script

maxpoulin64:
That is one massively overengineered library, ouch. But it does appear to solve our initial problem of preserving the exit code while running all tests, and it supports parallel execution as well.

package.json
```diff
@@ -16,9 +16,9 @@
     "build": "npm run build:grunt && npm run build:handlebars",
```
maxpoulin64: For the sake of uniformity, `npm-run-all` should be used for `build` as well.

astorije:
> That is one massively overengineered library, ouch.

Yeah, I wanted to look if it was one of these one-liner packages you can find out there, I got the same feeling than @maxpoulin64! I guess that's where the "java" in "javascript" lies :-)

Still, I'm 👍, it does the job, and we can easily switch to something else as there is no other addition on our side.

package.json
```diff
@@ -13,12 +13,12 @@
   "homepage": "https://thelounge.github.io/",
   "scripts": {
     "start": "node index",
-    "build": "npm run build:grunt && npm run build:handlebars",
+    "build": "npm-run-all build:grunt build:handlebars",
```
astorije: Actually, no need for `npm-run-all` here. The only benefit I see in this file is to have correct exit code **and** run everything when you test. For building, `&&` is fine as we need everything. Unless I missed something :-)

xPaw:
OK, I'll be fine with this. I think appveyor and travis configs need to be updated to just run the `npm test` command, instead of each individual one.

Travis: remove `script` section completely, travis will run `npm test` on its own
Appveyor: Remove `npm run xxx` lines

package.json
xPaw: I would agree that using `&&` is fine here.

williamboman: `npm-run-all` without `-c` will abort as soon as a script fails (with a non-0 exit code). `-c` will run all scripts no matter what, but in the end exit with a non-0 code if one or more scripts failed.

xPaw: I was thinking in regards to CIs, we would want to print all the errors at once (as we do with tests).

xPaw:
:+1: Looks good. Will wait for @maxpoulin64 or @astorije to review the latest changes before merging.

package.json
```diff
@@ -13,13 +13,13 @@
   "homepage": "https://thelounge.github.io/",
   "scripts": {
     "start": "node index",
-    "build": "npm run build:font-awesome && npm run build:grunt && npm run build:handlebars",
+    "build": "npm-run-all build:font-awesome build:grunt build:handlebars",
```
astorije: Sorry to be that guy, but I'd really be more comfortable keeping `&&` here. Reason is that the `build` script (and associated `prepublish`) is the most critical script of all as it is run by Travis CI when we publish, and releases cannot be modified or removed once published.

I'm with keeping `test` and `lint` short and clean because it's just a matter of convenience in dev, but we've all been surprised by the over-engineering of the `npm-run-all` package, and if building works perfectly fine with `&&`, we should stick to that. Adding a dependency for this step is another point of failure we loosely control.

Does this make sense?

astorije:
I would have been fine keeping the individual statements for CIs in order to limit the tie-ins with this package, but it's no big deal, not like it changes a lot of things nor like it's completely immutable if things go wrong.

I however added a comment about `build` that I think is worth pushing for the given reason. Unnecessary risk here.

astorije:
All good, until the next round of improvements :-)
👍

Issue #186, opened 2016-03-12
Accessing timestamps on mobile


dgw:
Right now, if the window width is below 480px, timestamps are hidden. This means that on mobile, timestamps are not available.

Several times since I started using The Lounge on my phone instead of a native client, I've had varying degrees of wanting or needing to see when a message was sent, and I can't.

I realize that showing all timestamps might present a UX issue (crowding the chat display), but there is a strong case for at least making them available.

lindskogen:
Like Facebook Messenger - pressing on a message bubble making it appear.

maxpoulin64:
Another option that I really like is how Snapchat does it: on Snapchat, if you slide the conversation to the right, then the timestamps appear on the left, and then hide again when you release.

http://screenshots.fr.sftcdn.net/blog/fr/2014/05/Time-when-the-messages-were-sent-Snapchat-568x402.jpg

It would probably require a lot more code though.

Jay2k1:
> Several times since I started using The Lounge on my phone instead of a native client, I've had varying degrees of wanting or needing to see when a message was sent, and I can't.

Yes you can, turn your phone into landscape :)

I miss timestamps on default portrait mode too though, and I'd welcome good ideas to make them show.

dgw:
Should be resolved by #376 being merged. Closing.

PR #376, opened 2016-06-03
Do not hide timestamps on small viewports

xPaw:


dgw:
👍  Should close #186 on merge.

maxpoulin64:
I'm not sure if that's just because I'm used to it or what, but it makes me feel like the interface is more bloated. Do we really want to have the timestamps visible all the time? I find that usually, it's only a few specific messages that I want to see when they were sent.

I was thinking, now that we have context menus, maybe we can either
* Put an on/off toggle in the channel menu, so it's easy to access
* Put it in the user context menu so people can manually check individual timestamps if wanted, kind of like that:
  screenshot_2016-06-03-18-47-08


I'll 👍 anyway as it's easy to revert with custom CSS

astorije:
Yes! :+1:

PR #378, opened 2016-06-05
Fix description of `host` and `bind` config options

maxpoulin64:
Commit da54263b8e14814378b193abaeef35f8ebd218fa introduced a totally wrong description for those options. This one describes better what exactly those options do.

@astorije: Do you have any idea how that commit made its way in? It doesn't seem to be tied to any PR, nor seem to have gone through any kind of review process. And apparently, it's not in Shout's repo.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -11,7 +11,8 @@ module.exports = {
 	public: true,
 
 	//
-	// Allow connections from this host.
+	// IP address or hostname for the web server to listen on.
+	// Setting this to "0.0.0.0" will listen on all interfaces.
 	//
 	// @type     string
 	// @default  "0.0.0.0"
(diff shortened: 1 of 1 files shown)


astorije:
@maxpoulin64, FYI, this commit comes from https://github.com/thelounge/lounge/pull/1, that is, even before I had time to change all names from Shout to The Lounge :D.
Let me look at these options and descriptions tomorrow. Thanks for catching this!

defaults/config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
 	public: true,
 
 	//
-	// Allow connections from this host.
+	// IP address or hostname for the web server to listen on.
```
xPaw: Perhaps specify that listening on 0.0.0.0 will listen on all available interfaces?

TurBoss:
cool that was confusing

PR #381, opened 2016-06-05
Cache npm modules on appveyor

xPaw:
https://www.appveyor.com/docs/build-cache#configuring-cache-items

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,5 +22,10 @@ test_script:
   - npm run lint:js
   - npm run lint:css
 
+# cache npm modules
+cache:
+  - node_modules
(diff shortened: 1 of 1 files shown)


astorije:
Considering AppVeyor takes time to start rather than to execute, I don't think this is going to change things a lot, but if their cache invalidation doesn't suck then this comes for free, so it's worth a try.

Also, in general (I saw some bikeshed material on the channel :p), I don't think it's worth investing any time at all on AppVeyor or Windows builds when not super necessary: after all, it's only there to make sure the commands themselves are not broken on Windows. Travis CI is our real source of truth.

👍 and merging!

PR #382, opened 2016-06-05
Upload test results on appveyor builds

xPaw:
Here's how it looks: https://ci.appveyor.com/project/astorije/lounge/build/61/tests

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,8 @@ environment:
 install:
   - ps: Install-Product node $env:nodejs_version
   - npm install
+  - npm install mocha-appveyor-reporter
+  - echo --reporter mocha-appveyor-reporter >> test/mocha.opts
 
 test_script:
   - node --version


maxpoulin64:
That looks neat, 👍

astorije:
It does look nice on AppVeyor, but for Travis CI, which I value more than AppVeyor, it looks to me that tests are not even run: https://travis-ci.org/thelounge/lounge/jobs/135370839#L706-L714.
They might be, but since there is no reporting at all, it looks like they aren't (and I'm guessing it's similar in dev env).

Also, new dev dependencies on developers' env while this is really useful only in AppVeyor, but that's the way it is with `npm`.

We could split paths between Travis CI and AppVeyor so that the reporters differ based on the CI but I think that would overcomplicate things for nothing, and therefore I'd be leaning towards another solution: be happy with the way it is right now. After all, I think [it's not that bad](https://ci.appveyor.com/project/astorije/lounge/build/70). But if you have better to offer, I'm not close to ideas, as long as it doesn't mess up Travis CI and local `npm test`.

xPaw:
@astorije Good catch, I fixed it by moving `--reporter` into appveyor config

astorije:
@xPaw, this won't work anymore because the `npm run test:mocha` line has disappeared from the AppVeyor conf in #375.

I am giving it a try by specifying the reporter in a [`mocha.opts`](https://mochajs.org/#mochaopts) file that is being created during AppVeyor build. That way, all our calls to `mocha` there are using the reporter regardless of how we write the `npm script`.
Also, I am installing the reporter during that same build instead of in the `package.json` file as it's not something that will be used outside of the AppVeyor environment :-)

Note that I created a revert of your initial commit as a second commit before my attempt. Should you not like my solution, you can delete my 2 commits. Should you like it, you can delete yours and the revert :-)
Of course, this is assuming the builds pass...

astorije:
(Adding the `do not merge` label until the revert commit gets removed and the solution decided on.)

xPaw:
@astorije Brilliant solution! I've rebased the PR.

appveyor.yml
```diff
@@ -13,6 +13,8 @@ environment:
 install:
   - ps: Install-Product node $env:nodejs_version
   - npm install
+  - npm install mocha-appveyor-reporter
```
astorije: FYI, I had put `npm install mocha-appveyor-reporter` _before_ `npm install` on purpose. As it is right now, the reporter install happens after the `npm run build` task and get lost/clogged up with the rest. Nicer to keep the installs together I think (but really no big deal!).

xPaw: Yeah I don't think that matters.

PR #384, opened 2016-06-05
Update eslint and enforce key-spacing

xPaw:


astorije:
Good catch! 👍 and merging.

PR #386, opened 2016-06-07
Display a loading message instead of blank page

xPaw:
07-1755-f271e11

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -43,7 +43,22 @@
 		</footer>
 		<div id="main">
 			<div id="windows">
-				<div id="chat"></div>
+				<div id="chat">
+					<div class="window" style="display:block">
+						<div class="container">
+							<div class="row">
+								<div class="col-xs-12">
+									<h2 class="title">The Lounge is loading…</h2>
+								</div>
+								<div class="col-xs-12">
+									<p id="loading-page-message">Loading the app… <a href="http://enable-javascript.com/" target="_blank">Make sure to have JavaScript enabled.</a></p>
+									<p id="loading-slow" style="display:none">This is taking longer than it should, there might be connectivity issues.</p>
+									<script async src="js/loading-slow-alert.js"></script>
+								</div>
+							</div>
+						</div>
+					</div>
+				</div>
 				<div id="sign-in" class="window">
 					<div class="header">
 						<button class="lt" aria-label="Toggle channel list"></button>
--- a/client/js/loading-slow-alert.js
+++ b/client/js/loading-slow-alert.js
@@ -1,0 +1,14 @@
+/*
+ * This is a separate file for two reasons:
+ * 1. CSP policy does not allow inline javascript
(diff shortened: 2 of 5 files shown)


maxpoulin64:
I'm not sure about this one. I think having a loading screen is a good idea, but it would be nice if it reflected the correct status so it doesn't talk about Javascript and connectivity issues when it's fine. Even on localhost it takes a few seconds to load, and it's not rare for it to take 10-15 seconds on mobile to load. I don't think we can have the progress, but just displaying what it's doing would be great.

xPaw:
@maxpoulin64 Is that better? It will now say `Connecting` and `Authorizing` while hiding the message about javascript as soon as it loads.

astorije:
I'll try this before giving a +1.

maxpoulin64:
Love it, 👍

xPaw:
I've rebased this and also fixed colors in morning/zenburn themes.

JustRamon:
It works. Cool :)

astorije:
This is really cool!
A few suggestions/comments before I'm a +1:

- I'm not sure about that `You must use a modern browser with JavaScript enabled.` statement. If it stays on for too long, it suggests the user that they do not have a compatible browser. At the very least, I would change it for `Make sure to use a modern browser with JavaScript enabled.` so that it shows we do not check that ourselves, but personally I'd rather ditch it entirely.
- What about a timeout to display `If this message still does not go away, there might be connectivity issues.`? There is no need to display it the first, say 5 seconds where this is loading so having it appear only after some time makes sense to me.
- `You're not connected to any networks yet` displays when the page is loading, which has always bugged me. This would be a good time to hide it until the page is loaded (and the user is authenticated in private mode <sup>1</sup>).

<sup>1</sup> It would actually be nicer to disable the sidebar menu on the login page, but that's a different PR :-)

client/index.html
```diff
@@ -43,7 +43,22 @@
...
+									<h2 class="title">The Lounge is loading…</h2>
+								</div>
+								<div class="col-xs-12">
+									<p id="loading-page-message">Loading the app… <a href="http://enable-javascript.com/" target="_blank">Make sure to have JavaScript enabled.</a></p>
+									<p id="loading-slow" style="display:none">This is taking longer than it should, there might be connectivity issues.</p>
+								</div>
+							</div>
+						</div>
+					</div>
+					<script async src="js/loader.js"></script>
```
astorije: Why `async` here?

xPaw: This script is not important enough to block for it.

astorije: Would call it loading-slow-check.js or something like that, unless you are planning to add more loader business there.

astorije:
This is great, 👍 and merging!

> `You're not connected to any networks yet` displays when the page is loading, which has always bugged me. This would be a good time to hide it until the page is loaded (and the user is authenticated in private mode<sup>1</sup>).

Not addressed but can definitely be part of a further PR, this is already a great step ;-)

PR #387, opened 2016-06-08
Cache loaded config and merge it with defaults

### LINKED ISSUE #249: Cache getConfig()
Currently it `require`s the config file from disk every time the function is called, it should be cached.

xPaw:
Fixes #249

### DIFF
--- a/src/command-line/start.js
+++ b/src/command-line/start.js
@@ -13,22 +13,25 @@ program
 	.description("Start the server")
 	.action(function() {
 		var users = new ClientManager().getUsers();
-		var config = Helper.getConfig();
-		var mode = config.public;
+
+		var mode = Helper.config.public;
 		if (program.public) {
 			mode = true;
 		} else if (program.private) {
 			mode = false;
 		}
+
 		if (!mode && !users.length) {
 			log.warn("No users found!");
 			log.info("Create a new user with 'lounge add <name>'.");
-		} else {
-			server({
-				host: program.host || process.env.IP   || config.host,
-				port: program.port || process.env.PORT || config.port,
-				bind: program.bind || config.bind,
-				public: mode
-			});
+
+			return;
 		}
+
+		Helper.config.host = program.host || Helper.config.host;
+		Helper.config.port = program.port || Helper.config.port;
+		Helper.config.bind = program.bind || Helper.config.bind;
+		Helper.config.public = mode;
+
+		server();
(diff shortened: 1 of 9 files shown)


src/command-line/start.js
```diff
@@ -13,22 +13,25 @@ program
...
-				host: program.host || process.env.IP   || config.host,
-				port: program.port || process.env.PORT || config.port,
-				bind: program.bind || config.bind,
-				public: mode
-			});
+
+			return;
 		}
+
+		Helper.config.host = program.host || process.env.IP   || Helper.config.host;
```
xPaw: ... Do we want to keep ENV here?

maxpoulin64: I don't really see a use for it. I would get rid of it. I think the only cases we should use environment variables is for things that can't be in the config file (so, the config path).

YaManicKill:
I'm not really sure why you needed to change getConfig, surely you could have stored config as a private variable, and getConfig just returned config. That would have meant a lot less code changes.

xPaw:
@YaManicKill There are not enough `getConfig` calls to warrant keeping it, especially in places where only a single variable is checked.

YaManicKill:
@xPaw At least 6 of the 9 files that are edited are edited only because you removed the function. It's not exactly like a getter is less efficient, anyway. What's the point in changing that?

xPaw:
@YaManicKill Some of the files were changed to remove default fallbacks (e.g. `link.js` or `userLog.js`).

Other files kept it's own stored cached configs (like `server.js` and `chan.js`). For consistency I decided to make it a public variable, instead of a function call.

maxpoulin64:
> Other files kept it's own stored cached configs (like` server.js` and `chan.js`). For consistency I decided to make it a public variable, instead of a function call.

Actually since this is JS it doesn't matter much as objects are always references, so either way you get the same thing in the end.

Another option to make this all shorter could be to do `var config = require("helper").config;` since most places don't use Helper functions at all.

maxpoulin64:
Looks fine to me, so I will 👍 and we can improve later. I'm not a fan of using "Helper.config" everywhere and would rather have `var config = require("helper").config;`, but it's not enough to block this.

YaManicKill:
Fine, whatever. I don't really care enough, I just didn't see the point. So, whatever. :+1: and merging.

PR #394, opened 2016-06-10
Add spotify link handling

karbowiak:
It's not exactly what you would call pretty - would help if message/link interception was modularized with individual templates and js plugins :)

### DIFF
--- a/client/views/toggle.tpl
+++ b/client/views/toggle.tpl
@@ -4,10 +4,12 @@
 		<a href="{{link}}" target="_blank">
 			<img src="{{link}}">
 		</a>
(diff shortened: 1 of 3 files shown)


xPaw:
I think if we were to add support for specific services, we should use an existing library to avoid writing code for it, like http://riteshkr.com/embed.js/

YaManicKill:
I agree with @xPaw (and @maxpoulin64 has agreed as well).

I'm going to close this but it would be great to have a PR for embed.js or https://github.com/moooji/embedify, as both of them support spotify and many other services as well.

Issue #77, opened 2016-02-20
Figure out module system


YaManicKill:
So, thought I'd get a proper issue up for discussion on the module system that we will end up using.

There are 2 cases that we need to solve. These may use the same system, or they may use a different one.
### 1. Modularising the client

There are 2 parts to this: 3rd party libraries and lounge.js

Now, the next question is whether we want to be building this into 1 or 2 bundles, or whether we want to have them stay modularised, give thelounge http2 support and make that the default.

As a quick recap for those that don't know the difference between http 1.1 and http 2:
- http 1 can't download multiple files at once. Browsers have a hack to allow at most 10 to download simulataneously, because of this it is standard practice to concatenate as much as possible, and download as few files as possible, as this speeds it up
- http 2 has multiplexing and can download many, many files in parallel, and the standard practice for this is not concatenating, but keeping modules seperate as that speeds up downloading.

[If you look at global http2 usage](http://caniuse.com/#feat=http2) you'll see that >70% of people can use http 2. And if we focus on http2, http1 will still work, it'll just be a little slower than normal.

So, for us to decide what to do about modularising the client, we need to decide if we want to focus on http 1 or 2 just now.
### 2. Packages

So, @astorije brought up the point with me that it would be ideal to have a situation where a user can just install a new client package in thelounge and not need to reload the server to have it running. So, we need a way to dynamically include them on the browser (not needing a build step).

One option for this would be SystemJS, as that runs in the browser and allows you to include things programatically using any of the usual module systems. 

With all of these, we then need to decide on what syntax of module we want to code in.

Personally, I would suggest that we code everything in the same module syntax, as that means that people don't have to learn 1 syntax for core features and 1 syntax for packages. Any of the systems we use (browserify, rollup, systemjs) all support both commonjs and ES6 modules.

Seeing as browsers will support ES6 modules in the future (none support it yet), and probably not commonjs modules, I would suggest using ES6. This also allows us to use tree shaking if we wish in the future.

So, we need to decide:
- [ ] Module syntax?
- [ ] Bundle client JS or not?
- [ ] If bundle, how many bundles?
- [ ] What to use to allow programmatic modules for packages?

xPaw:
> If you look at global http2 usage you'll see that >70% of people can use http 2. And if we focus on http2, http1 will still work, it'll just be a little slower than normal.

[...]

PR #395, opened 2016-06-11
Add a shrug command

karbowiak:
Shruggie is great for when you just don't give a fuck..

Also moved parts of the me command to it's own function called handleMe, easier to reuse that way

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -24,6 +24,7 @@ $(function() {
 		"/send",
 		"/server",
 		"/slap",
+		"/shrug",
 		"/topic",
 		"/voice",
 		"/whois"
--- a/src/plugins/inputs/action.js
+++ b/src/plugins/inputs/action.js
@@ -6,7 +6,7 @@ exports.input = function (network, chan, cmd, args) {
 
 	switch (cmd) {
 	case "shrug":
-		text = "¯\\_(ツ)_/¯";
+		text = "¯\\_(ツ)_/¯ " + args.join(" ");
 		handleMe(text);
 		break;
 	case "slap":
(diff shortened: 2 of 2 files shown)


dgw:
Looks like `/slap` is leftover from the old Shout days, two years ago. To me, both `/slap` and `/shrug` belong in plugin territory. The Lounge will have a plugin/module system in the future (worked on in #77), and extra commands would be a great use case for plugins.

That said, if @thelounge/maintainers are OK having this in core until such time as plugins exist, I don't strenuously object. Any protocol-unrelated commands can be split out into plugins at that later date.

MaxLeiter:
Agree with @dgw. As a note, @thelounge/maintainers should work.

dgw:
@MaxLeiter Interesting. I put in @thelounge/maintainers and it didn't highlight.

xPaw:
Yeah, there's no purpose to ship this. :-1: 

@dgw, @MaxLeiter Only members can mention teams.

astorije:
This is clearly a cry for packages indeed.

On one end, we already have `/slap` so we could indeed take it until we split them apart. On the other end, accepting this is opening the door to all sorts of PRs to add people's crazy ideas :-)

So I would be 👎 as well, but I'd rather have @maxpoulin64 and @YaManicKill's opinion before closing this. Sorry @karbowiak and thanks for you contribution(s) still!

YaManicKill:
Yeah, I'm agreed. 

I am obviously a bit torn, because we keep saying things would be good for plugins, but we still don't have any way to do plugins, but yeah I don't think this is good.

Similarly, I talked to @karbowiak on the irc and explained about text expanders which would allow him to do this easily anyway.

PR #396, opened 2016-06-11
Capitalise constructor Oidentd

hogofwar:
A constructor name should start with an uppercase letter.


maxpoulin64:
Looks like it's the only place where it isn't, so 👍

PR #397, opened 2016-06-11
Remove redundant variables

hogofwar:
Removed two redundant variables and just put their contents just where they are used.

### DIFF
--- a/src/command-line/reset.js
+++ b/src/command-line/reset.js
@@ -23,8 +23,7 @@ program
 				return;
 			}
 			var salt = bcrypt.genSaltSync(8);
-			var hash = bcrypt.hashSync(password, salt);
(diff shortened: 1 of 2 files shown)


src/command-line/reset.js
```diff
@@ -23,8 +23,7 @@ program
 				return;
 			}
 			var salt = bcrypt.genSaltSync(8);
-			var hash = bcrypt.hashSync(password, salt);
-			user.password = hash;
+			user.password = bcrypt.hashSync(password, salt);
```
maxpoulin64: Might as well get rid of the `salt` one while at it

hogofwar: Didn't want to take the change too far, but if it's wanted, sure.

PR #399, opened 2016-06-12
Rename package variable to pkg, as "package" is reserved.

hogofwar:


maxpoulin64:
It is indeed a reserved keyword, although it still works. Merging as that's just a variable name, so minor change.

Issue #393, opened 2016-06-10
Slideout menu bar for mobile


karbowiak:
Hi there..

Currently i find myself constantly sliding my thumb from the left side, to the right, to get to the sidebar with all the channels, all to no avail.

Something like https://mango.github.io/slideout/ could fix that i'm sure (i suck at JS to be honest) :)

astorije:
Whoever takes this, please share design decision here. In particular, if using a third-party library, explore options and let us know why one more than the others :-)

xPaw:
Yeah it's really hard to find an actually good library for this. Just tried Slideout.js, it's not awful, but it has some delay/lag issues.

maxpoulin64:
I just swapped the label to help wanted, because it's really not an easy one so it's definitely not an up for grabs.

dgw:
Caught myself trying to slide out the side menu yesterday, but it's not exactly a common feature for webapps to support. I can't think of one that does off the top of my head.

Would be a real feather in The Lounge's cap if this was implemented, though. The few, the proud, the menu-sliders!

maxpoulin64:
[Annnnd here we go!](https://d.max-p.me/irc/slideout.html)

Probably needs a bit of tweaking to fit The Lounge, but it works.

maxpoulin64:
Alternatively, it seems like [this is the cause](https://github.com/Mango/slideout/blob/master/index.js#L170-L176) of the delay in the existing one, so it should be relatively easy to fix upstream and we could then use this implementation and throw away mine.

astorije:
For what it's worth, I was pointed to https://events.google.com/io2016/ as a decent-looking implementation of a slideout menu. I haven't looked at the code at all, so I don't know if we can make any use of this (like if it's not possible to know which lib they are using...).

PR #400, opened 2016-06-12
Add touch slideout menu for mobile

maxpoulin64:
Implements #393 with a slightly modified version of the demo solution I had. The only problem with it that I'm aware of is that the transition is fixed, so if you're like me and swipe the slider really fast it feels laggy when you release the finger as the animation still plays at normal speed. I didn't feel like implementing the maths to support inertia, but if that's bugging too much people I will add it.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -303,11 +303,24 @@ button {
 
 #viewport {
 	height: 100%;
-	transition: all .4s;
+	transition: transform 160ms, -webkit-transform 160ms;
 	-webkit-transform: translateZ(0);
 	transform: translateZ(0);
 	-webkit-perspective: 1000;
 	perspective: 1000;
+}
+
+#viewport.menu-open {
+	-webkit-transform: translate3d(220px, 0, 0);
+	transform: translate3d(220px, 0, 0);
+}
+
+#viewport.menu-dragging {
+	transition: none !important;
+}
+
+#viewport.menu-open .messages {
+	pointer-events: none;
 }
 
 #viewport .lt,
@@ -1595,11 +1608,6 @@ button {
 		margin-top: 60px !important;
 	}
 
-	#viewport.lt {
-		-webkit-transform: translate3d(220px, 0, 0);
-		transform: translate3d(220px, 0, 0);
-	}
-
 	#viewport.rt #chat .sidebar {
 		-webkit-transform: translate3d(-180px, 0, 0);
 		transform: translate3d(-180px, 0, 0);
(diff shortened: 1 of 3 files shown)


dgw:
I believe @astorije himself would say it's better to add this and improve it later if necessary than try to get it perfect before merging. 👍

xPaw:
Trying this on desktop and there's a bug with the menu being a little buggy:

1. Resize the window so that the menu hides
2. Open the menu
3. Resize to higher resolution, the sidebar will now re-appear

maxpoulin64:
@xPaw:

1- I semi-fixed it, but master does that as well in an even more silly way (at least for me). It's hard to fix due to the way we do the menu (we actually put the menu with `left: -220px` and slide the whole viewport 220px instead of changing the absolute position of the menu itself). I can disable animations when going into desktop mode because the menu is never animated, but I can't easily disable it for when we transit between desktop->mobile.

2- Good catch, fixed

3- Semi-fixed. Since every window have its own menu button for some reason, by the time we get the event to fire on `#viewport` it's already clicked so I can't do much better than disabling pointer events on `.messages` :/

astorije:
@maxpoulin64, could you rebase this with master please?

maxpoulin64:
Sure, will do tomorrow or saturday as soon as I get a chance.

maxpoulin64:
@astorije: done, pretty minor conflict.

astorije:
If @xPaw or @williamboman or @MaxLeiter want to give it a try, I think this is ready to go 🎉.

williamboman:
I ran this for quite a while before the first 2.0 rc and it worked great then. Gonna update my instance to run latest master + interesting PRs (including this) again.

MaxLeiter:
I really like the change, but it feels a little too fast. Could we make the transition a little longer?

MaxLeiter:
From IRC

> 21:28	+Max-P	Hmm yeah, I made it quite fast because it otherwise feel a bit weird when doing a quick swipe
21:29	+Max-P	I'll make it a bit slower, can't make it too slow but I certainly can make it slightly slower.
21:29	+Max-P	It's not just a gesture, you can actually drag it however you want, that's why I had to make it that way
21:35	MaxLeiter	well when you open sidebar and tap anywhere, it closes
21:36	MaxLeiter	that felt too fast, the dragging part was fine, but I assume those are the same value?
21:36	+Max-P	Yeah
21:36	+Max-P	Basically what happens is after the gesture, the end result is calculated and it snaps to either open or closed
21:36	+Max-P	So it's the same (CSS) animation that finishes it
21:36	+Max-P	But yeah I'll probably split them into two, or calculate it in JS and set it in style

astorije:
I'm in favor of shipping this now and improving later. It's not like it's broken, just it could be slightly improved. What do you think @MaxLeiter?

MaxLeiter:
Fine with me, @astorije

PR #401, opened 2016-06-12
Save user when they sort the channel list

xPaw:


astorije:
That, yes! :+1 and merging.

maxpoulin64:
Oh my, I finally won't have to sort my channels on every restart? Yes please!

PR #402, opened 2016-06-13
command-line: fall back to LOUNGE_HOME env variable

### LINKED ISSUE #365: Use and respect environment variables?
E.g. stuff like `LOUNGE_HOME`;

```diff
diff --git a/src/command-line/index.js b/src/command-line/index.js
index d81988b..e368db3 100644
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -13,7 +13,7 @@ program.option("    --home <path>" , "home path");
 
 var argv = program.parseOptions(process.argv);
 
-Helper.setHome(program.home);
+Helper.setHome(program.home || process.env.LOUNGE_HOME);
 
 if (!fs.existsSync(Helper.CONFIG_PATH)) {
 	mkdirp.sync(Helper.HOME, {mode: "0700"});

```

williamboman:
Makes it easier to use lounge with a different home configuration.

Closes #365.

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -13,7 +13,7 @@ program.option("    --home <path>" , "home path");
 
 var argv = program.parseOptions(process.argv);
 
-Helper.setHome(program.home);
+Helper.setHome(program.home || process.env.LOUNGE_HOME);
 
 if (!fs.existsSync(Helper.CONFIG_PATH)) {
 	fsextra.ensureDirSync(Helper.HOME);


xPaw:
@williamboman Can you also add this to documentation on the site?

We used to support `HOME` but it was undocumented and with some undefined behaviour.

maxpoulin64:
Little not for the next plans: it would probably be a good idea to have a variableto set the configuration file location, and be able to set the location for data/users/logs individually from within said config file. Some people (like me) like to have configs and data separate. It would also help make packages that respects the FSH properly without using symlinks (so `/etc/thelounge.js` for the config, `/var/lib/thelounge/users/` for the users).

PR #403, opened 2016-06-13
Update travis config 

xPaw:
- Adds `node_modules` cache
- Adds testing on OSX (that's what the stylelint update is for)
- Removes node5 testing as it is no longer stable
- Only test latest node4 version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,23 @@
 language: node_js
 node_js:
-- '6'
-- '5'
-- '4'
-- '4.2.0'
+- 6
+- 4
+
+matrix:
+  fast_finish: true
+  include:
+  - os: osx
+    node_js: 4
+
+cache:
+  directories:
+    - node_modules
+
 notifications:
   email:
     on_success: never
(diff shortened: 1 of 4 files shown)


maxpoulin64:
RIP AppVeyor

```
npm ERR! version not found: stylelint@6.6.0
```

astorije:
👍 
Weird that AppVeyor docs gives `'%APPDATA%\npm-cache'` as an example in caching deps but it fails as soon as you bump a version...

astorije:
Merging as minor and internal.

astorije:
@xPaw, any reason why we shouldn't include Node.js v6 in OSX builds?

.travis.yml
```diff
@@ -1,13 +1,23 @@
 language: node_js
 node_js:
-- '6'
-- '5'
-- '4'
-- '4.2.0'
```
astorije: It may be worth updating https://github.com/thelounge/lounge/blame/master/README.md#L40 then.

PR #405, opened 2016-06-13
Update README.md

pguth:
Point to `nativefier`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ When the install is complete, go ahead and run this in your terminal:
 lounge --help
 ```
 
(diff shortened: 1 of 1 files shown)


maxpoulin64:
I don't think this is needed. There are so many of those options around everyone has their favorite way of wrapping applications. Plus it's not very clear how it integrates with The Lounge and could easily be confused with a native wrapper like Electron. Plus, most browsers have "add to desktop" options now that does exactly that. So that's clearly a 👎 for me, I really don't see the use at all.

pguth:
I agree. Thanks!

PR #407, opened 2016-06-15
Send a Pushover notification on message reception

sebastiencs:
This allows to get notifications on mobile, tablet, etc. using [pushover](https://pushover.net/) when we receive a message (private message or someone wrote our name).
It's usefull when we're not on the web client, we have no excuse if we miss something.

It's the first time I write js code, don't be too hard.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -320,4 +320,30 @@ module.exports = {
 	// @default  null
 	//
 	oidentd: null,
+
+	//
+	// Get a notification on message reception using pushover
+	// https://pushover.net/
+	//
+	// @type     object
+	// @default  {}
+	//
+	pushover: {
+		//
+		// User token
+		//
+		// @type     string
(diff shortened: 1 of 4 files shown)


xPaw:
:-1: as this requires a paid app.

> After a 7-day free trial, it costs $4.99 USD once, per-platform

maxpoulin64:
Agreed. 👎 

Third party implementations are also a much better candidate for external plugins rather than being built-in. Soon enough people will ask for Pushbullet and well as a bunch of other push applications, and we can't reasonably have them all in master.

I also have a WIP branch for webpush, which is a web standard in progress and is thus arguably a lot better than depending on a third-party.

dgw:
Pushover has pretty wide adoption and support. It might not be suitable for core, but once a plugin system is available, I encourage anyone interested to publish this functionality as a plugin.

PR #409, opened 2016-06-16
Perform node version check as soon as possible

xPaw:
Allows us to print a message before crashing when loading irc-fw

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,13 @@
 #!/usr/bin/env node
 process.chdir(__dirname);
+
+// Perform node version check before loading any other files or modules
+// Doing this check as soon as possible allows us to avoid ES6 parser errors or other issues
+var pkg = require("./package.json");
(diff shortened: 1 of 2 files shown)


YaManicKill:
:+1:  and merging. Good call.

index.js
```diff
@@ -1,3 +1,13 @@
 #!/usr/bin/env node
 process.chdir(__dirname);
+
+// Perform node version check before loading any other files or modules
+// Doing this check as soon as possible allows us to avoid ES6 parser errors or other issues
+var pkg = require("./package.json");
+if (!require("semver").satisfies(process.version, pkg.engines.node)) {
+	console.error("=== WARNING!");
```
astorije: Is it worth using your logger instead, considering we control its content (and can make sure it works on Node.js < 4)?

xPaw: I wanted to, but it needed to load helper file, which needed to load config file, which requires home path processing. So it's not "as soon as possible".

PR #410, opened 2016-06-16
Prepend http protocol to www. links in chat

### LINKED ISSUE #406: Don't use protocol-relative URLs for external links
This sounds confusing, so I'll explain.

Assume someone posts a link like this: `www.some.domain`
Lounge's URL parser will make that clickable, turning it into `<a href="//www.some.domain"` (note that no protocol is given).
Assuming you're using The Lounge over HTTPS, the browser will open `https://www.some.domain` when you click. **This is problematic** because many websites still run http only, leading to The Lounge giving you dead links. Also, when someone links to a webpage without protocol, chances are they mean http and not https.

When you type `www.some.tld` into your browser's URL bar, it will assume http (unless you use a plugin like HTTPS everywhere, which will fall back to http if https does not work), so I think The Lounge should do the same.

Consequently, I suggest that Lounge adds `http://` to protocol-less URLs, instead of just adding `//`.

I am assuming that the most common cases are:
- Server is running http only (in which case nothing listens on 443)
- Server is running http and https (you get what you surf to)
- Server is running http and https (you're being redirected to https when you access http)

In each of these cases it's safe to always link to http (when no protocol is given in the URL), whereas in one case always linking to https will lead to a browser error message, leaving the user thinking they were just given a dead link.

xPaw:
Fixes #406

### DIFF
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -14,7 +14,7 @@ function uri(text) {
 			return url;
 		}
 		var split = url.split("<");
-		url = "<a href='" + split[0].replace(/^www/, "//www") + "' target='_blank' rel='noopener'>" + split[0] + "</a>";
+		url = "<a href='" + split[0].replace(/^www/, "http://www") + "' target='_blank' rel='noopener'>" + split[0] + "</a>";
 		if (split.length > 1) {
 			url += "<" + split.slice(1).join("<");
 		}


maxpoulin64:
:+1: and merging. It's what browsers do when you don't specify the protocol in the URL bar and sites enforcing HTTPS should use HSTS so the regular HTTP request is never made.

Issue #358, opened 2016-05-29
Can't join multiple channels separated by comma


MaxLeiter:
Using the Connect screen, you can't join multiple channels separated by a comma (and by default the channels input is filled with two channels separated by a comma).

![image](https://sr.ht/1adj.png)
https://sr.ht/hEfd.webm

xPaw:
It parses join string by space, and first part is used as channels, and second part used as join key.

So there shouldn't be a space after comma.

PR #411, opened 2016-06-17
Fix channel join regression and fix possibly joining parted channels

xPaw:
This PR creates channels out of `join` string when creating a network to avoid reusing this `join` string on future reconnections.

This PR also fixes a regression which was introduced in 2.0-pre: #358.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -137,6 +137,16 @@ Client.prototype.connect = function(args) {
 
 	var nick = args.nick || "lounge-user";
 	var webirc = null;
+	var channels = [];
+
+	if (args.join) {
+		var join = args.join.replace(/\,/g, " ").split(/\s+/g);
+		join.forEach(function(chan) {
+			channels.push(new Chan({
+				name: chan
+			}));
+		});
+	}
 
 	var network = new Network({
 		name: args.name || "",
@@ -149,6 +159,7 @@ Client.prototype.connect = function(args) {
 		commands: args.commands,
 		ip: args.ip,
 		hostname: args.hostname,
+		channels: channels,
 	});
 	network.setNick(nick);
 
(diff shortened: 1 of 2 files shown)


src/plugins/irc-events/connection.js
```diff
@@ -10,6 +10,35 @@ module.exports = function(irc, network) {
...
+		var commands = network.commands;
+		if (Array.isArray(commands)) {
+			commands.forEach(function(cmd) {
+				setTimeout(function() {
+					client.input({
+						target: network.channels[0].id,
+						text: cmd
+					});
+				}, delay);
+				delay += 1000;
```
williamboman: This could be simplified a bit;

```js
const delay = 1000;
commands.forEach((text, idx) => {
    setTimeout(() => client.input({
          target: network.channels[0].id,
          text,
     }), delay * ++idx);
});
```

xPaw: Joins have to run after commands, so that won't work. I also didn't touch this code, only moved it to another file.

maxpoulin64:
Looks good to me, 👍

src/plugins/irc-events/connection.js
astorije: To be honest, I'm not sure the `delay * ++idx` statement is something I would call "simplified"...

maxpoulin64: Agreed with @astorije, plus apart from changing the functions to arrow functions (thus breaking older versions), it doesn't actually change much. `idx*1000` would do the same, as `forEach` already provide the index and thus doesn't need to be incremented.

src/client.js
```diff
@@ -137,6 +137,16 @@ Client.prototype.connect = function(args) {
...
 	var webirc = null;
+	var channels = [];
+
+	if (args.join) {
+		var join = args.join.replace(/\,/g, " ").split(/\s+/g);
+		join.forEach(function(chan) {
+			channels.push(new Chan({
+				name: chan
+			}));
+		});
```
astorije: What about:

```js
channels = args.join.replace(/\,/g, " ").split(/\s+/g).map(function(chan) {
	return new Chan({
		name: chan
	});
});
```
or, if you don't want that long first line:
```js
var join = args.join.replace(/\,/g, " ").split(/\s+/g);
channels = join.map(...)
```

astorije: Nevermind, got ruled out in priv by @xPaw, who plans to use the `channels` variable in a further PR and cannot blindly overwrite it.

astorije:
👍

PR #413, opened 2016-06-19
Fix crash in public mode introduced by #370

maxpoulin64:
Oopsie. Just ran into it by accident.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -67,13 +67,13 @@ function Client(manager, name, config) {
 
 	var client = this;
 
-	if (!client.config.token) {
-		client.updateToken(function() {
-			client.manager.updateUser(client.name, {token: client.config.token});
-		});
-	}
-
 	if (config) {
+		if (!config.token) {
+			client.updateToken(function() {
+				client.manager.updateUser(client.name, {token: config.token});
+			});
+		}
+
 		var delay = 0;
 		(config.networks || []).forEach(function(n) {
 			setTimeout(function() {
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -67,13 +67,13 @@ function Client(manager, name, config) {
...
 	var client = this;
 
-	if (!client.config.token) {
-		client.updateToken(function() {
-			client.manager.updateUser(client.name, {token: client.config.token});
-		});
-	}
-
 	if (config) {
+		if (config.token) {
```
astorije: I guess the purpose of this PR is to ensure that the `config` (or `client.config` depending on where it is used) is not `null` or `undefined` so that `config.something` / `client.config.something` does not crash the app.

However, why did you move from `!client.config.token` to, effectively, `client.config.token`?

maxpoulin64: Huh, yeah I think I made a mistake there. I need to switch it back to `!`. It is a problem with `config` being undefined indeed.

maxpoulin64: @astorije updated.

astorije:
Follow-up question: in which case(s) can we end up with `client.config` being `undefined` or `null`? You said you ran into this, so I'm curious what led you there.

maxpoulin64:
`config` is a user's `.json` file. In public mode, when a user connects, a temporary anonymous client is made and this one doesn't have a config. No config means no JSON, so no token no saving.

PR #415, opened 2016-06-19
Do not print user loaded message in public mode

xPaw:
Similar to #413. This fixes `User "undefined" loaded` which was introduced somewhere in 2.0-pre.


maxpoulin64:
Oh wow, it was right under it and I didn't even notice it.👍

Issue #412, opened 2016-06-17
Join keys are not stored


xPaw:
https://tools.ietf.org/html/rfc2812#section-3.2.1

We need to store them in `Chan` object and need to write them to user config to correctly rejoin the channels. To fix this in the user config, I think we need to deprecate the `join` string, and store the `channels` array properly (only name and join keys).

amcgee7:
I'd like to optionally store the join key's in the json file and allow users to enter the key with the channel when they add the network?  Any guides out there?  I don't know the code that well.

spookhurb:
I have a fix for this!

PR #417, opened 2016-06-19
Store channels in array format

xPaw:
Preliminary support for #412.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -141,13 +141,34 @@ Client.prototype.connect = function(args) {
 	var webirc = null;
 	var channels = [];
 
-	if (args.join) {
-		var join = args.join.replace(/\,/g, " ").split(/\s+/g);
-		join.forEach(function(chan) {
+	if (args.channels) {
+		var badName = false;
+
+		args.channels.forEach(function(chan) {
+			if (!chan.name) {
+				badName = true;
+				return;
+			}
+
 			channels.push(new Chan({
-				name: chan
+				name: chan.name
 			}));
 		});
+
+		if (badName && client.name) {
+			log.warn("User '" + client.name + "' on network '" + args.name + "' has an invalid channel which has been ignored");
+		}
+	// `join` is kept for backwards compatibility when updating from versions <2.0
+	// also used by the "connect" window
+	} else if (args.join) {
+		channels = args.join
+			.replace(/\,/g, " ")
+			.split(/\s+/g)
+			.map(function(chan) {
+				return new Chan({
+					name: chan
+				});
+			});
 	}
 
 	var network = new Network({
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -73,10 +73,15 @@ Network.prototype.export = function() {
 		"hostname"
 	]);
 
-	network.join = _.map(
-		_.filter(this.channels, {type: Chan.Type.CHANNEL}),
-		"name"
-	).join(",");
+	network.channels = this.channels
+		.filter(function(channel) {
+			return channel.type === Chan.Type.CHANNEL;
+		})
+		.map(function(chan) {
+			return _.pick(chan, [
+				"name"
+			]);
+		});
 
 	return network;
 };
--- a/test/models/network.js
+++ b/test/models/network.js
@@ -13,6 +13,8 @@ describe("Network", function() {
 			network.setNick("chillin`");
 			network.channels.push(new Chan({name: "#thelounge"}));
 			network.channels.push(new Chan({name: "&foobar"}));
+			network.channels.push(new Chan({name: "Lobby", type: Chan.Type.LOBBY}));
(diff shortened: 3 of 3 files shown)


src/client.js
```diff
@@ -139,13 +139,27 @@ Client.prototype.connect = function(args) {
...
-	if (args.join) {
-		var join = args.join.replace(/\,/g, " ").split(/\s+/g);
-		join.forEach(function(chan) {
-			channels.push(new Chan({
-				name: chan
-			}));
+	if (args.channels) {
+		channels = args.channels.map(function(chan) {
+			if (!chan.name) {
+				return;
```
maxpoulin64: We probably should print an error message there to let the user know that a channel has been ignored instead of just aborting like that.

xPaw: What would we even print though? "A channel with no name was ignored"? That doesn't seem particularly useful.

maxpoulin64: Something like `Channel X for network X for user Y is invalid, ignoring` should do. The message doesn't have to be all that useful, it's just a matter of letting the user that *something* is wrong with their config, so the user isn't there "well, it loads my config fine and all but it doesn't join my channel!". I would be ok even with `Invalid channel in config`, the point really is just to give a hint to the user that an error happened.

Personally I would flat out crash (correctness > working), but that won't make a lot of happy people :P

maxpoulin64:
Nice addition overall, consider this a 👍 as soon as my comment have been addressed (whether we decide to add a warning or not).

xPaw:
@maxpoulin64 added a message and fixed a bug along with it.

src/models/network.js
```diff
@@ -73,10 +73,13 @@ Network.prototype.export = function() {
 		"hostname"
 	]);
 
-	network.join = _.map(
-		_.filter(this.channels, {type: Chan.Type.CHANNEL}),
-		"name"
-	).join(",");
+	network.channels = _
+		.filter(this.channels, {type: Chan.Type.CHANNEL})
```
astorije: I know it's a copy from the above, but is it possible to use:

```js
this.channels.filter(function(channel) { return channel.type === Chan.Type.CHANNEL; });
```

? If so, let's do it? Please?

src/models/network.js
```diff
@@ -73,10 +73,15 @@ Network.prototype.export = function() {
 		"hostname"
 	]);
 
-	network.join = _.map(
-		_.filter(this.channels, {type: Chan.Type.CHANNEL}),
-		"name"
-	).join(",");
+	network.channels = _
```
astorije: s/`_`/`this.channels`/, no? :-)

xPaw: No, `network` is local.

src/client.js
```diff
@@ -141,13 +141,34 @@ Client.prototype.connect = function(args) {
...
-				name: chan
+				name: chan.name
 			}));
 		});
+
+		if (badName && client.name) {
+			log.warn("User '" + client.name + "' on network '" + args.name + "' has an invalid channel which has been ignored");
+		}
+	// `join` is kept for backwards compatibility when updating from versions <2.0
+	// also used by the "connect" window
```
astorije: I don't think we are converting channels on the fly until a new channel is added or removed (and re-ordered, maybe?) so that the list gets dumped in the user configuration. Or are we and I'm missing it?

Depending on the answer, can we log something like `[WARNING] <user> is using the deprecated comma-separated channel list.`? I realize you use the same code for handling old format *and* the join window input though, so not trivial if it stays that way.

astorije:
This is a nice addition, thanks!

One comment I have is: couldn't we wrap this behind a user setting, similar to the MOTD one? Not sure what name we could give it, but if it's a generic info checkbox, I would also put the "Enabled capabilities" there (if possible, of course).

Also (unrelated to this PR, but displayed by it), I noticed a line saying `[403]	Freenode No such channel`, are we trying to join the lobby somewhere when first connecting?

src/client.js
xPaw: It gets converted when the user is loaded.

dgw:
@astorije Didn't you merge #453 to fix that issue (#450)? 😄

maxpoulin64:
@dgw: no, that one just prevents joining the server window.

xPaw:
@maxpoulin64 I think he meant `[403] Freenode No such channel`, which indeed fixed that.

dgw:
@xPaw Right, I did mean the 403 numeric.

astorije:
Sorry for the noise, commented on the wrong PR...

👍 and merging this, discussion in comment can continue post-merge.

src/client.js
astorije: Bottom-line is: is there a point soon where code L163 to L171 will be any obsolete?

xPaw: Not yet, it's used by the server window.

PR #418, opened 2016-06-19
Use tabs when saving user configs

xPaw:
Make it nicer, IMO. Also makes it consistent (it saved with single space in one place).


astorije:
👍 and merging.

PR #420, opened 2016-06-20
Do not display the sidebar on sign-in page

astorije:
The sidebar is really useless on the authentication page. The settings page has no effect until logged, while icons to add a network and log out are purely wrong.
This is an attempt to make the sign-in page use the whole viewport:

The commit also:
- Removes burger menu icon on mobile version of sign-in page
- Adds the `.signed-out` class to the initial body instead of only when sign-in has failed
- Removes hiding connect and logout button icons, which was buggy and is now useless anyway

#### Result

screen shot 2016-06-20 at 01 51 11

#### Current issues

- [x] ~~The sign-in icon is not even displayed anymore, and has no value, so I removed *most* of it. Because of how the client displays the sign-in form, I cannot simply remove it: when receiving the `auth` event, a click is emitted on the button that activates the right window, changes the page title, ... Not pretty. Any suggestion on improving this? I'm thinking of moving that logic to a different function that the click handler calls, and the `auth` event handler could call it without resorting to a click. Does that sound any better?~~ Scratch that, I'll tackle this in a further PR.

- [x] ~~Right now, there is no scrollbar on the sign-in page, which is a problem. I'll look into that issue, but if someone sees something obvious, please speak up :-)~~ Fixed along with implementing @xPaw's suggestion.

- [x] ~~The loading page looks gross with this. I should apply behavior of this PR to the loading page as well (no sidebar), and make sure the input field does not apply there either.~~ Thanks to @xPaw's suggestion, this PR also applies to the loading page \o/

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -512,6 +512,10 @@ button,
 	top: 4px;
 }
 
+.signed-out #main {
+	left: 5px;
+}
+
 #header {
 	display: none;
 	height: 40px;
--- a/client/index.html
+++ b/client/index.html
@@ -23,7 +23,7 @@
 	<link rel="manifest" href="manifest.json">
 
 	</head>
-	<body class="<%= public ? "public" : "" %>">
+	<body class="signed-out <%= public ? "public" : "" %>">
 
 	<div id="wrap">
 	<div id="viewport">
@@ -60,9 +60,6 @@
 					</div>
 				</div>
 				<div id="sign-in" class="window">
-					<div class="header">
-						<button class="lt" aria-label="Toggle channel list"></button>
-					</div>
 					<form class="container" method="post" action="">
 						<div class="row">
 							<div class="col-xs-12">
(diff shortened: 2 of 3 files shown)


xPaw:
Instead of moving the whole layout, can't you just hide sidebar and footer when `signed-out` class is set?

astorije:
Good call, I'll try that.

astorije:
Much better indeed, @xPaw, thanks for the suggestion!

xPaw:
Indeed much better, will test and merge.

xPaw:
Looks fine, however I'd like to display `The Lounge` somewhere, otherwise it's hard to tell what you are signing into. Maybe make use of the header and put in a title there?

astorije:
@xPaw, I agree with the intent, but I'd rather see this in its own PR, especially since this was not available before this PR (actually, you could see this by going to the settings and reading the About statement, but that's a side effect, really).

If you want, I can change from "Sign in" to "Sign in to The Lounge" until this is taken up (`up for grabs`?) and removed from both the loading and the sign in pages. What do you think?

PR #423, opened 2016-06-21
Make style of loading page similar to other pages

astorije:
- Move markup around to make the loading page a window of its own instead of a sub-window of `#chat`
- Remove inline styling in loading page
- Use same styling than other non-messages windows for title and text
- Add a `z-index` to the loading page to hide the message input
- Vertically align the sign in page title with all other titles
- Make sure all `h1.title`s are bottom-margined consistently and remove negative margin on the Settings page title to align with the Connect page title (Reverting/Improving this should be done at the `h2` level instead)

screen shot 2016-06-21 at 02 02 46

I can't admit or deny that the style of this is better or worse than the previous version. To me they are equivalent. However, this is now consistent with the sign in page, and all other non-chat pages.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -43,22 +43,21 @@
 		</footer>
 		<div id="main">
 			<div id="windows">
-				<div id="chat">
-					<div class="window" style="display:block">
-						<div class="container">
-							<div class="row">
-								<div class="col-xs-12">
-									<h2 class="title">The Lounge is loading…</h2>
-								</div>
-								<div class="col-xs-12">
-									<p id="loading-page-message">Loading the app… <a href="http://enable-javascript.com/" target="_blank">Make sure to have JavaScript enabled.</a></p>
-									<p id="loading-slow" style="display:none">This is taking longer than it should, there might be connectivity issues.</p>
-									<script async src="js/loading-slow-alert.js"></script>
-								</div>
+				<div id="loading" class="window active">
+					<div class="container">
+						<div class="row">
+							<div class="col-xs-12">
+								<h1 class="title">The Lounge is loading…</h1>
+							</div>
+							<div class="col-xs-12">
+								<p id="loading-page-message">Loading the app… <a href="http://enable-javascript.com/" target="_blank">Make sure to have JavaScript enabled.</a></p>
+								<p id="loading-slow">This is taking longer than it should, there might be connectivity issues.</p>
+								<script async src="js/loading-slow-alert.js"></script>
 							</div>
 						</div>
 					</div>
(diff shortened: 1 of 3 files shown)


williamboman:
Nice! This change might seem minor but it makes the critical rendering path more visually pleasing (mainly due to no more `h2` border).

client/index.html
```diff
@@ -43,22 +43,21 @@
...
+							</div>
+							<div class="col-xs-12">
+								<p id="loading-page-message">Loading the app… <a href="http://enable-javascript.com/" target="_blank">Make sure to have JavaScript enabled.</a></p>
+								<p id="loading-slow">This is taking longer than it should, there might be connectivity issues.</p>
+								<script async src="js/loading-slow-alert.js"></script>
 							</div>
 						</div>
 					</div>
 				</div>
+				<div id="chat"></div>
```
xPaw: Keeping it in `#chat` meant it will be overridden once the chat has loaded, thus not keeping the loader in DOM. Any reason you moved it out?

xPaw:
:+1: except for the inline comment.

client/index.html
astorije: Indeed. I actually started by leaving it in there, but I was forced to cancel everything that the `#chat` was setting, and it was a bit messy to hide the input field.
Making it its own window made more sense and cleaned up things a bit.

I hear you though, so I am now detaching the loading page in the `init` event handler. Doing it earlier (when the sign in page appears) means I have to duplicate for public mode, so it's good like this I believe.

client/js/lounge.js
```diff
@@ -178,6 +178,7 @@ $(function() {
 		}
 
 		$("body").removeClass("signed-out");
+		$("#loading").detach();
```
xPaw: This should have been `.remove`, we don't be reusing the element at a later time.

client/index.html
xPaw: Guess that makes sense, fine by me.

client/js/lounge.js
astorije: Indeed! Should I use `.remove()` on `$("#sign-in")` next line as well?

xPaw: Probably.

astorije: Done in e5dddba.

astorije:
All of @xPaw's comments have been addressed :-)

YaManicKill:
Just tried it out, looks good. And the code seems fine :-) I agree with keeping it outside of #chat, because it really shouldn't be in there anyway.

:+1: and merging.

PR #424, opened 2016-06-21
Display header on windows

xPaw:
as @astorije was lazy

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -23,7 +23,7 @@
 	<link rel="manifest" href="manifest.json">
 
 	</head>
-	<body class="signed-out <%= public ? "public" : "" %>">
+	<body class="signed-out <%= public ? "public" : "private" %>">
 
 	<div id="wrap">
 	<div id="viewport">
@@ -44,6 +44,9 @@
 		<div id="main">
 			<div id="windows">
 				<div id="loading" class="window active">
+					<div class="header">
+						<span class="title">The Lounge</span>
+					</div>
 					<div class="container">
 						<div class="row">
 							<div class="col-xs-12">
@@ -59,6 +62,9 @@
 				</div>
 				<div id="chat"></div>
 				<div id="sign-in" class="window">
+					<div class="header">
+						<span class="title">The Lounge</span>
+					</div>
 					<form class="container" method="post" action="">
 						<div class="row">
 							<div class="col-xs-12">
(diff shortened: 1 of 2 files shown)


client/index.html
```diff
@@ -60,6 +60,10 @@ <h2 class="title">The Lounge is loading…</h2>
 					</div>
 				</div>
 				<div id="sign-in" class="window">
+					<div class="header">
+						<span class="title">The Lounge</span>
+						<span class="topic">Web IRC Client</span>
```
astorije: Would you mind just displaying "The Lounge"? I'm not too fond of that subtitle and don't think we need any. After all, if you're logging into the app, you've been past the presentation website...

client/index.html
```diff
@@ -175,6 +180,7 @@ <h1 class="title">Connect</h1>
 				<div id="settings" class="window">
 					<div class="header">
 						<button class="lt" aria-label="Toggle channel list"></button>
+						<span class="title">Settings</span>
```
astorije: This (and the previous one) just feels odd. It adds nothing to see "Settings" right above "Settings". If the header was static, maybe (though really not sold), but here... Meh.

astorije:
s/lazy/busy/ 😄

client/index.html
xPaw: We indeed could make the header static, just like in chat windows.

client/index.html
astorije: Well, there is no sign in page on a public instance 😄.

Anyway, I'm happy with "The Lounge", but "Web IRC Client" or any description is really unnecessary here.
What do other @thelounge/maintainers think?

client/index.html
astorije: Meh, I'm not sure it adds value, but we'll see.

In the meantime, my opinion is we should really not have this top titles on the settings and connect pages.

maxpoulin64:
Simple touch, makes the header less empty. 👍

astorije:
Just to summarize, here are my comments illustrated with screenshots of that PR:

1. Double title on the Settings and Connect page:
  screen shot 2016-06-24 at 01 22 08
  My suggestion here is to remove the top ones.

2. Unnecessary and ugly sub-title on the sign in page:
  screen shot 2016-06-24 at 01 22 50
  My suggestion here is to keep "The Lounge" and ditch "Web IRC Client"

3. Lack of title on the loading page:
  screen shot 2016-06-24 at 01 24 02
  My suggestion here is to add the same title as the sign in page one.

With these 3 changes I'll be happy with a +1 :-)
I'm open to suggestions as well.

xPaw:
@astorije somewhat fixed your comments. It's stupidly not easy to do what you're asking for because of mixing private/public modes and sharing connect windows in both.

astorije:
I'm not too fond of the resulting design of this, so I opened #487 instead, with details on what I plan to do next.
I suggest closing this PR as it compromises on the design and is based on a system that cannot live any longer (sharing the connect page as a landing page on public mode, ew).

astorije:
Closing in favor of #487 or maybe something even better.

PR #425, opened 2016-06-21
Add custom highlights

YaManicKill:


client/js/lounge.js
```diff
@@ -227,6 +227,10 @@ $(function() {
 		var chan = chat.find(target);
 		var msg;
 
+		if (highlights.some((h) => data.msg.text.indexOf(h) > -1)) {
```
xPaw: Could this support regex? Maybe if it starts with `/`

williamboman: Feels a bit unnecessary to include ES6 syntax - and effectively breaking 45 % of global browser market share - for something so small. Could be an opportunity to add babel transpilation though.

client/js/lounge.js
```diff
@@ -504,6 +511,10 @@ $(function() {
 		if (name === "userStyles") {
 			$(document.head).find("#user-specified-css").html(options[name]);
 		}
+		if (name === "highlights" ) {
+			var highlightString = options[name];
+			highlights = highlightString.split(",").map((h) => h.trim());
```
astorije: Same comment as above re: ES6 on the client.

client/index.html
```diff
@@ -270,8 +270,8 @@ <h1 class="title">Settings</h1>
 
 							<div class="col-sm-12">
 								<label class="opt">
-									<input type="checkbox" name="notifyAllMessages">
-									Enable notification for all messages
+									<label for="old_password_input" class="sr-only">Custom Highlights (separate by ",")</label>
+									<input type="text" id="highlights" name="highlights" class="input" placeholder="Custom Highlights (separate by ',')">
```
astorije: I think this was said on the channel but reminder to re-add the all-message notifications.

.eslintrc.yml
```diff
@@ -6,6 +6,7 @@ env:
   browser: true
   mocha: true
   node: true
+  es6: true
```
astorije: Once the 2 fat arrows are replaced with normal functions, this can go :-)

client/js/lounge.js
YaManicKill: I will remove it for now, but I would really want to add babel at some point. We can use arrow functions on the server, but not the client, and it seems a massive shame to leave out such nice development tools just because rubbish browsers.

I have a little bit of an obsession with arrow functions just now.

YaManicKill:
Fixed all comments, including the fun bug where if you didn't put any custom highlights, you were notified for everything.

client/js/lounge.js
astorije: I fully agree with that, @YaManicKill. Just not worth breaking browsers for a couple fat arrow functions just now :-)

maxpoulin64:
I find it a little weird that we now have highlights on both the client and the server, and it won't work very well in the future when we add push notifications and such. But I'll call it good enough for now so have a 👍

client/index.html
```diff
@@ -274,6 +274,14 @@ <h1 class="title">Settings</h1>
 									Enable notification for all messages
 								</label>
 							</div>
+
+							<div class="col-sm-12">
+								<label class="opt">
+									<label for="highlights" class="sr-only">Custom Highlights (separate by ",")</label>
```
astorije: I suggest the following placeholder here and in the `<input />` below:

> Custom highlights (comma-separated keywords)

Reasoning is: no upper-casing of "highlight" to match other inputs/labels in this page, and using words instead of characters as 1) screen readers (at least VoiceOver) will not read them at all (it just reads "separated by") and the `','` looks weird, very packed on the default font.

client/js/lounge.js
```diff
@@ -504,6 +513,16 @@ $(function() {
 		if (name === "userStyles") {
 			$(document.head).find("#user-specified-css").html(options[name]);
 		}
+		if (name === "highlights" ) {
```
astorije: Whitespace sneaked in `" )` :-)

astorije:
I agree with @maxpoulin64 but oh well, we'll refactor later :-)
I left a couple minor comments but great work otherwise! I'll give a final review after comments are addressed. Thanks for this @YaManicKill! 😄

YaManicKill:
I agree with both of you, but the only way to sort it out would be to put it on the server which is a bigger issue in itself. 

Will address comments tomorrow.

astorije:
> I agree with both of you, but the only way to sort it out would be to put it on the server which is a bigger issue in itself.

For sure! I think it was more of a general comment. I'm happy with the way you took here overall.

xPaw:
We will be able to move it to the server after @maxpoulin64's settings overhaul.

YaManicKill:
@xPaw Yeah, the only problem with that is that everything is an option to sync to the server, we would have to require custom highlights being synced for that to work. Is that something we want to do? I'm not sure, open to ideas.

astorije:
Hey @YaManicKill, any news on my 2 minor comments? :-)

YaManicKill:
I've been busy, I'll hopefully get that sorted tonight or tomorrow.

YaManicKill:
Done :-)

astorije:
Awesome @YaManicKill, great job! 👍 and merging, Travis CI seems to have a hard time with the Mac builds...

PR #427, opened 2016-06-22
Remove navigator.standalone detection

xPaw:
It was added in b3b7d126be11177c22e6165ffd6c65b81575886e to apparently work around an iOS bug which was fixed in 8.3.

https://discussions.apple.com/thread/6557230?start=0&tstart=0
https://discussions.apple.com/message/28059603#28059603

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -126,10 +126,6 @@ button,
 	padding: 0;
 }
 
-.web-app-mode {
-	padding-top: 20px;
-}
-
 .btn {
 	border: 2px solid #84ce88;
 	border-radius: 3px;
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -34,10 +34,6 @@ $(function() {
 	var sidebar = $("#sidebar, #footer");
 	var chat = $("#chat");
 
-	if (navigator.standalone) {
-		$("html").addClass("web-app-mode");
-	}
-
 	var pop;
 	try {
 		pop = new Audio();


YaManicKill:
Yeah, infact we should probably remove it because iOS users will be getting 20px less of their screen to see things on. :+1: and merging.

astorije:
I think this would have benefited from a bigger timeframe before merging, @YaManicKill, or at least a second review maybe :-)

There is still value post-merge if someone stops by though: Is there any iOS user who can post screenshots of before/after this PR? Interested in a post-8.3 for reference (iPhone 6 and higher) and things like iPhone 5 and less, and iPad 2 or 3 (one that is pre-8.3). Thanks! 😄

daGrevis:
This is how lounge@master looks for me on iOS 9.3.4.

https://imgur.com/a7skm1a

Thumpxr:
iPhone 6+ iOS 10 developer beta 7 - Lounge 2.0.0-pre.6 : http://i.imgur.com/zwihk0Z.png

astorije:
For the record:

> 05:41	daGrevis	Anyone using latest lounge on iPhone by dropping it on homescreen?
> 05:42	daGrevis	There's no margin top - lounge draws itself over top bar (where clock and other phone stuff is)
> 05:46	lime_	which browser?
> 05:46	lime_	it should be OS independant
> 05:46	lime_	hence nodeJS
> 05:47	daGrevis	Its not related to nodejs
> 05:47	daGrevis	I will add screenshot
> 05:48	daGrevis	http://m.imgur.com/a7skm1a
> 05:48	daGrevis	If you look closly, you can see clock and other stuff on top
> 05:49	daGrevis	Its Safari on iPhone 5s, when lounge is moved to homescreen and opened from there
> 05:51	Thumpxr	daGrevis i have the same issue
> 05:54	lime_	have you tried firefox/chrome on the iphone?
> 05:56	Thumpxr	you cant set default browsers on iOS, hence safari will always be used for the homescreen link. Furthermore is CHrome/Firefox just a reskinned safari, as WebKit(i think) is to deep in the system to be replayed by 3rd party engines
> 06:00	lime_	thats, bullshit
> 06:03	lime_	you  could host a webpage that opens your lounge server in firefox/chrome
> 06:03	lime_	or you could get a better phone ;)
> 06:03	+xPaw	Thats a ios bug
> 06:31	daGrevis	All browsers on iOS use the same engine under the hood
> 06:32	daGrevis	So if you're using Firefox, you're actually using Safari with Firefox skin
> 06:32	daGrevis	This bug didnt exist on v1.5
> 06:56	+xPaw	1.5 had a hack for it
> 07:05	daGrevis	hmm I have iOS 9.3.4 and it's not fixed
> 07:50	daGrevis	temporary fixed with custom stylesheet
> 07:54	daGrevis	Can someone confirm that it's fixed on iOS? Because its not on my iPhone 5s

@xPaw, could we revert this? It looks like iOS 9 and 10 still have that issue.

Thumpxr:
Further investigations: Air 2 iOS 10 public Beta 6 pre.6 : same bug
iPhone 6s iOS 9.3.5 pre.6 : same bug

williamboman:
iOS 10.1.1 looks like this;

img_0089

astorije:
@Thumpxr, @williamboman, any suggestions? What do you think we should do for this?

PR #428, opened 2016-06-22
Change nicks from links to spans everywhere

### LINKED ISSUE #338: Change nicks from links to buttons everywhere
As discussed in https://github.com/thelounge/lounge/pull/325#issuecomment-219372241, clickable nicks should be buttons and not links as they don't take the user anywhere, per se.

There are several implications to this: @YaManicKill noticed highlighting does not work on buttons on Firefox for example.
Also, @cptvitamin, could you advise on accessibility consequences for this, please?

@xPaw mentioned in the PR that we could use `<span role="button">` as well.

xPaw:
Fixes #338 

This removes the dumb "#" hover link previews.

### DIFF
--- a/client/views/actions/action.tpl
+++ b/client/views/actions/action.tpl
@@ -1,2 +1,2 @@
-<a href="#" class="user {{colorClass from}}" data-name="{{from}}">{{mode}}{{from}}</a>
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{mode}}{{from}}</span>
 <span class="action-text">{{{parse text}}}</span>
--- a/client/views/actions/ctcp.tpl
+++ b/client/views/actions/ctcp.tpl
@@ -1,2 +1,2 @@
-<a href="#" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</a>
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</span>
 <b>{{ctcpType}}</b> {{{parse ctcpMessage}}}
--- a/client/views/actions/invite.tpl
+++ b/client/views/actions/invite.tpl
@@ -1,9 +1,9 @@
-<a href="#" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</a>
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</span>
 invited
 {{#if invitedYou}}
 	you
 {{else}}
-	<a href="#" class="user {{colorClass invited}}" data-name="{{invited}}">{{invited}}</a>
+	<span role="button" class="user {{colorClass invited}}" data-name="{{invited}}">{{invited}}</span>
 {{/if}}
 to
 {{{parse channel}}}
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -117,8 +117,7 @@ input {
 	outline: 0;
 }
 
-button,
-.user {
+button {
 	border: none;
 	background: none;
 	margin: 0;
@@ -451,10 +450,6 @@ button,
 	width: 210px;
 }
 
-#footer button {
-	border: 0;
-}
-
 #footer button.active {
 	color: #fff;
 }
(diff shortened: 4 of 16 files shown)


maxpoulin64:
👍

astorije:
@xPaw, I found a bug with the user search not working anymore, fixed in a commit, hope you don't mind!

One sad side-effect of going from links to spans is that we lose ability to tab and receive focus. @cptvitamin gave us some pointers to solve this at https://github.com/thelounge/lounge/issues/338#issuecomment-219548623. @cptvitamin, feel free to jump in if you want to add anything to the discussion here :-)

xPaw:
@astorije Good fix.

I know I didn't set tabindex, is it that desired of a feature to tab index all the nicks in chat window?

client/views/actions/action.tpl
```diff
@@ -1,2 +1,2 @@
-<a href="#" class="user {{colorClass from}}" data-name="{{from}}">{{mode}}{{from}}</a>
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{mode}}{{from}}</span>
```
cptvitamin: to be accessible these will need a ```tabindex=0``` and keyboard event listeners (Space and Enter) to work.

client/views/actions/ctcp.tpl
```diff
@@ -1,2 +1,2 @@
-<a href="#" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</a>
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</span>
```
cptvitamin: same as above

client/views/actions/invite.tpl
```diff
@@ -1,9 +1,9 @@
-<a href="#" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</a>
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{from}}</span>
```
cptvitamin: all the way down

xPaw:
@cptvitamin see my comment above, I'm well aware of tabindex. The question is, do we *want* to make nicks tabbable in chat? Currently if you're focused in the input you can't tab out, and to tab in the chat window you have to select some text first so that you won't be focused in the input again.

astorije:
Follow-up of a discussion on the IRC channel (redacted for simplicity):

```
10:11 <xPaw>        astorije, Max-P, do we *want* to make nicks tabbable? currently if you're focused in the input you can't tab out, and to tab in the chat window you have to select some text first so that you're not being focused in the input again
11:12 <astorije>    xPaw, re: can't tab out of the input, that is definitely an accessibility issue. re: make nicks tabbable, it looks like not doing so makes things even worse for accessibility
11:13 <YaManicKill> We could make it that if you tab without any characters typed, it tabs out of the input? Is that possible?
11:13 <xPaw>        astorije, hexchat for example doesn't do any tabbing on nicks in chat, in user list it allows using arrow keys
11:18 <xPaw>        currently tabbing doesn't work on server list in lounge, we have to decide what and how we want to support keyboard stuff, instead of blindly adding tabindex=0 on everything
11:20 <xPaw>        I'm not sure how making every nick in chat tabbable is useful (userlist is limited usefulness)
11:20 <xPaw>        clicking a nick does a whois, you might as well type the commmand
11:26 <astorije>    It might not be useful for everyone, but it might be very useful for someone who can't read the usernames :/
15:52 <Max-P>       astorije: the nicks just do a /whois, can't the user just type it in anyway? If I were blind, I'd much prefer to use IRC in a command and reply way
15:52 <xPaw>        I'm seeing limited use in tabbing over nicks in chat, tabbing on links on other hand *is* useful
15:53 <xPaw>        if shift+tab focused on last link in chat, that'd be nice
```

@xPaw raised a good point that tabbing the message window has little value. It's actually the opposite, it clouds tabbing across links given in messages. So I agree that making nicks non-tabbable will make for a better experience.

Nicks should however be tabbable in the user list. It's the case at the moment, but it's rather broken as you have no visual feedback at all. It works fairly well with VoiceOver though.

However, these statements assume that it is possible to leave the input field once it is focused, which is not the case (and even so, it's not easy to access things like the channel list or user list)...

So I am 👍 here even though the user list is losing in keyboard-control, but that's only because it's not possible to access that area from the keyboard. @cptvitamin, any suggestions to make the different areas of the app themselves accessible from the keyboard are welcome!

PR #429, opened 2016-06-22
Remove javascript scrollbar library, use hover appearing scrollbar instead

xPaw:
Removes the tse-scrollbar library. Makes the channel list and user list only have scrollbar visible when hovering over the container.

This library was used in the channel list only, it is sometimes quite laggy on mobile.


astorije:
I think there were some issues discussed on the channel, but if we end up being able to get rid of that plugin, that's really cool! 🎉 

Will review when I get some free time.

maxpoulin64:
I like that, works really well! Good job 👍 

Styling could be a bit better (at least color match the old tse one and a transition), but that can go with another PR.

client/css/style.css
```diff
@@ -1557,62 +1549,22 @@ button,
...
-	position: relative;
-	overflow: hidden;
-}
-
-.tse-scrollable .tse-scroll-content {
-	overflow: hidden;
-	overflow-y: scroll;
+.scrollable {
+	overflow: auto;
+	visibility: hidden;
```
astorije: This has an unfortunate side effect, so I suggest removing fiddling with scrollbar visibility and letting the OS decide: scrollbars are set up to always show up on my OS, and this change effectively ignores this setup. Essentially, this is an accessibility issue (accessibility is not only about screen readers ;-) ).

astorije:
One minor comment (disabling the display-on-hover) and then I'll be +1 on that PR :-)

xPaw:
@astorije That's not exactly minor. Windows scrollbars are ugly in all browsers (except Chrome where we style them), and I'm sure users would want to get rid of them.

astorije:
Right, I understand. However, I believe this should be a recipe in the CSS Mods page because, as it is in this PR, it contradicts OS settings in OSes that offer system-wide control of the scrollbars.

xPaw:
@astorije I've removed the hover stuff.

astorije:
Great @xPaw, 👍!
Glad we can get rid of an in-repo library 😄

PR #432, opened 2016-06-24
Fix default configuration for `host`.

maxpoulin64:
Similarly to kiwiirc/irc-framework#55, passing `undefined` there instead of `"0.0.0.0"` allows the OS to decide and use both IPv4 and IPv6.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -12,12 +12,12 @@ module.exports = {
 
 	//
 	// IP address or hostname for the web server to listen on.
-	// Setting this to "0.0.0.0" will listen on all interfaces.
+	// Setting this to undefined will listen on all interfaces.
 	//
 	// @type     string
-	// @default  "0.0.0.0"
+	// @default  undefined
 	//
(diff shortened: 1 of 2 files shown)


xPaw:
Our site documentation should be updated to explain on how to listen on only ipv4/ipv6 interfaces, but this PR is :+1:

astorije:
👍 

Am I right assuming this PR will only work as expected when https://github.com/kiwiirc/irc-framework/pull/55 gets merged?
Also, agreed regarding documentation.

maxpoulin64:
@astorije: they are actually separate things. It's the same defect but at different places (here, on the Lounge own HTTP server, and in irc-fw for the outgoing IRC connections), based on the same IPv4-centered assumption that `"0.0.0.0"` is the default "everything" option. This PR makes The Lounge accessible over IPv6, and the irc-fw patch allows The Lounge to connect to IPv6-enabled servers.

astorije:
Makes complete sense, thanks @maxpoulin64!

PR #434, opened 2016-06-27
irc-events/connection: increase join delay to 1000ms

williamboman:
100ms easily bypasses the excess flood threshold with constant
reproducibility with >20 channels (Freenode).

I think d66e86dd introduced the issue.

### DIFF
--- a/src/plugins/irc-events/connection.js
+++ b/src/plugins/irc-events/connection.js
@@ -35,7 +35,7 @@ module.exports = function(irc, network) {
 			setTimeout(function() {
 				network.irc.join(chan.name);
 			}, delay);
-			delay += 100;
+			delay += 1000;
 		});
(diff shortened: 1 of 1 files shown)


src/plugins/irc-events/connection.js
```diff
@@ -35,7 +35,7 @@ module.exports = function(irc, network) {
 			setTimeout(function() {
 				network.irc.join(chan.name);
 			}, delay);
-			delay += 100;
+			delay += 1000;
```
williamboman: ¯\_(ツ)_/¯

maxpoulin64:
This probably should be configurable in some way, but 👍 for now.

astorije:
I'm 👍 with this as well. Merging.

PR #437, opened 2016-06-30
src/client: fix storing updated token, ensure client config is always an object

williamboman:
This also fixes an issue where the token would not be stored in the
user file.

Just want to point out that `config !== client.config` as well (hence breakage in the first place).

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -55,6 +55,9 @@ var inputs = [
 }, {});
 
 function Client(manager, name, config) {
+	if (typeof config !== "object") {
+		config = {};
+	}
 	_.merge(this, {
 		activeChannel: -1,
 		config: config,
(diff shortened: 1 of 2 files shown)


astorije:
@xPaw, do you mind taking a quick look at this one?

src/client.js
```diff
@@ -67,15 +67,15 @@ function Client(manager, name, config) {
 
 	var client = this;
 
-	if (config) {
-		if (!config.token) {
-			client.updateToken(function() {
-				client.manager.updateUser(client.name, {token: config.token});
```
xPaw: The problem was specifically this line, which should have been `client.config.token`.

src/client.js
```diff
@@ -67,15 +67,15 @@ function Client(manager, name, config) {
...
+	if (client.config) {
```
xPaw: This line is a bit suspect, I think we should default `config` to `{}` if it's undefined, and remove the check (I saw another `|| {}` in the code.

xPaw:
:+1: Good catch. Feel free to fix my comments, if you want to.

src/client.js
```diff
@@ -55,6 +55,9 @@ var inputs = [
 }, {});
 
 function Client(manager, name, config) {
+	if (config === undefined || typeof config !== "object") {
```
xPaw: undefined isn't an object, so the first check isn't necessary.

src/client.js
```diff
@@ -67,30 +70,28 @@ function Client(manager, name, config) {
...
+
+	var delay = 0;
+	(client.config.networks || []).forEach(function(n) {
+		setTimeout(function() {
+			client.connect(n);
+		}, delay);
+		delay += 1000;
+	});
+
+	log.info("User '" + name + "' loaded");
```
xPaw: Add a check for `name`

williamboman:
Ok patch should be final now.

xPaw:
Still :+1:

astorije:
👍 and merging, thanks!

PR #448, opened 2016-07-02
Different way to display highlights

xPaw:
I've been using this on my instance for a while, and it's way better than changing the text color to red.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1081,11 +1081,19 @@ kbd {
 }
 
 #chat .error,
-#chat .error .from,
-#chat .channel .highlight .from,
-#chat .channel .highlight .text,
-#chat .channel .highlight .user {
+#chat .error .from {
 	color: #f00;
+}
+
+#chat .channel .highlight .time,
+#chat .channel .highlight .from {
+	color: #fff;
+	background-color: #ff8a80;
+	border-color: #ff8a80;
+}
+
+#chat .channel .highlight .from .user {
+	color: #fff;
 }
 
 #chat .msg.toggle .time {
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -187,15 +187,18 @@ body {
 }
 
 #chat .error,
-#chat .error .from,
-#chat .channel .highlight,
-#chat .channel .highlight .from,
-#chat .channel .highlight .text {
+#chat .error .from {
 	color: #f92772;
 }
 
 #chat .unhandled .from {
 	color: #99a2b4;
+}
+
+#chat .channel .highlight .time,
+#chat .channel .highlight .from {
+	background-color: #ad1457;
+	border-color: #ad1457;
 }
 
 #chat .msg.quit .time,
--- a/client/themes/zenburn.css
+++ b/client/themes/zenburn.css
@@ -213,15 +213,18 @@ body {
 }
 
 #chat .error,
-#chat .error .from,
-#chat .channel .highlight,
-#chat .channel .highlight .from,
-#chat .channel .highlight .text {
+#chat .error .from {
 	color: #bc6c4c;
 }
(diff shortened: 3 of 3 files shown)


YaManicKill:
I have been using your css for this for quite a while, and I love it. It makes highlights a lot easier to notice. :+1:

YaManicKill:
(I won't merge in yet though, I think this needs at least 1 more)

MaxLeiter:
Been using the CSS for a while, definitely support changing :+1:

maxpoulin64:
I like it, but it looks kind of meh on mobile however: 
screenshot_2016-07-02-21-15-16

Maybe keep the current behavior in the media queries for small screens?

astorije:
I understand the technical need (not changing the color of the text itself) and the practical need (easily identifying notices when scrolling back). From a UI/design perspective, this severely contradicts with the rest of the application, whether it's the color itself or the fact that we foreground-color, etc. It is also visually very aggressive (again, I get the need to distinguish these).

Are you guys open to discussing something more in line with the global design and less aggressive to the eye? A lot of things are possible here, I believe, and I could share ideas if there is interest.

xPaw:
@maxpoulin64 I like it on mobile too, using red text is kind of crap to quickly find the highlights.

@astorije Aggressive styling works quite well for highlights, especially in busy channels. I specifically didn't change the background on messages to not be even more aggressive (and to avoid styling issues). Please do share your ideas if you have any.

YaManicKill:
I'm happy with either this being merged or not. I care less about these styling changes because I can just change them locally now :-P

Maybe it's worth leaving the default and having this as 1 of the first plugins, seeing as it looks like we might actually have a plugin system soon (yay). Because this is definitely a case of personal preference, and you are never going to make everyone happy.

maxpoulin64:
Given the current state of the opinions, I think we should make this a custom CSS entry instead, or at least an option. I actually like the dark variants of this, but I for the default white theme I think it looks kinda bad.

Could we maybe have it only for morning and zenburn? I'm under the impression that the people using those themes are also the people that tends to want the bright highlights. Sounds like a good compromise for the meantime?

YaManicKill:
Is it worth closing this if it isn't going to go anywhere?

xPaw:
@YaManicKill I still want to get better highlights. Need to hear what @astorije has in mind.

YaManicKill:
Nudge. @astorije you had any thoughts?

bews:
Interesting idea, I thought about it myself after seeing something like this in discord. 
The problem is I like when my messages are highlighted too, which will be not so good with this design (assuming these 2 highlights will use the same design, as they should).

So for now I use underline highlighting:


It looks good on mobile too.

xPaw:
https://github.com/thelounge/lounge-theme-solarized has this. Since this PR is not moving anywhere I will close it.

xPaw:
@thelounge/maintainers I think this needs to be given a chance again now that nicknames are colored in messages. I've already heard it's harder to scan for highlights because some nick colors are also red.

astorije:
> I've already heard it's harder to scan for highlights because some nick colors are also red.

Agreed.

I am not a fan of the original solution, it does feel like a bandaid to replace the broken current situation, but I didn't get a chance to play with something better.
I was thinking having up/down arrows to move across highlights would be very useful. That would make it for easy browsing of recent highlights.
@xPaw, what about something like the following:

screen shot 2017-12-04 at 18 37 58

It obviously needs UI tweaking, but as an idea.
Yes, it's more discreet than your solution, but I think way more eye-catchy when scrolling than the current solution we have and would be nice with other mechanisms like being able to browse across highlights.

xPaw:
@astorije That's not the worst thing, anything is better than just setting the color to red.

PR #453, opened 2016-07-03
Only auto join actual channels

### LINKED ISSUE #450: Lounge tries to join the server window
Found by #286 and confirmed with Wireshark. @vectr0n also confirms this was an issue before as well, seen through HexChat.

Lounge sends a JOIN command with the name of the network, aka the server window. For instance, if you have a network called "Freenode", it will try to join a channel named "Freenode".

xPaw:
Fixes #450


astorije:
👍 and merging.

PR #454, opened 2016-07-03
Only trigger custom highlights for non-self messages and notices

xPaw:


YaManicKill:
:+1: and merging, simple bug fix, and it does the job.

PR #455, opened 2016-07-03
client: stop propagation when hiding the chat through click/tapping the chat

williamboman:
This is really annoying on mobile. When I click the chat to hide the side menu 9 times out of 10 I hit a nickname.

saddsad


YaManicKill:
Huh, good call. I don't tend to have this problem much, but I think that's because I have gotten used to only closing with using the menu button.

I'll :+1: it but wait for another.

xPaw:
That's one of the things I pointed out in @maxpoulin64's PR. :+1:

PR #461, opened 2016-07-04
Only load config if it exists

### LINKED ISSUE #460: Thelounge tryes to load a config file that does not exist
Hi

I'm trying to deploy theloungle with passenger mod for apache2

when i run npm install everything is fine but npm start throws this error

http://dpaste.com/1DZSAE6
http://dpaste.com/0B4WTJJ

thx

xPaw:
Fixes #460

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -1,6 +1,7 @@
 var _ = require("lodash");
 var path = require("path");
 var os = require("os");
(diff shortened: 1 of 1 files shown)


astorije:
Nice catch, 👍 and merging!

PR #465, opened 2016-07-05
Fix window layout a bit

### LINKED ISSUE #463: Message textbox shows on loading screen
Initially fixed by #420, #379 causes the message box to show on the loading screen

textbox showing

maxpoulin64:
This puts the input form in the right place. No more hacks to hide it, just don't display it.

This is kind of a quick fix after all, and probably not ideal. The thing is a giant mess of absolutely positionned element outside of their parent containers. It's really weird. The CSS *really* messy for the UI we have, I think I will redo it soon-ish. But for now that will do.

Fixes #463

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -520,12 +520,6 @@ button {
 	width: 100%;
 }
 
-#windows {
-	position: relative;
-	overflow: hidden;
-	flex: 1;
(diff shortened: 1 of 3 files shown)


astorije:
👍 and merging.

PR #467, opened 2016-07-06
Fix slight bugs introduced by #379 and #465

maxpoulin64:


YaManicKill:
:+1: and merging

client/themes/zenburn.css
```diff
@@ -173,9 +173,9 @@ body {
 }
 
 #form .input {
-	background-color: #434443;
-	border-color: #101010;
-	color: #dcdccc;
+	background-color: #434443 !important;
+	border-color: #101010 !important;
+	color: #dcdccc !important;
```
astorije: Hummmmm isn't there a better way than adding all these `!important`?
Removing one is a small victory so adding five...

maxpoulin64: It would be possible, yes. But we're already fighting for the style in the main stylesheet and I didn't feel like making a mess of selectors. But yeah, I should have taken the time but it was late and I wanted to get it fixed ASAP.

astorije: 😭

PR #471, opened 2016-07-07
Fix unread marker disappearing when opacity set to 1

### LINKED ISSUE #470: Unread marker line disappears on Morning/Zenburn
It shows up fine on the default theme however. This was introduced by #465 ~~and/or #467~~.

astorije:
Fixes #470.

`z-index` sorcery was removed [here](https://github.com/thelounge/lounge/pull/465/files#diff-e5178f7b74fe45f2cfe1baf9aa1ef6faL791), and the unread marker line, since showing with `z-index: -1`, actually needs a `0` reference.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -725,6 +725,7 @@ button {
 	text-align: center;
 	opacity: .5;
 	margin: 0 10px;
+	z-index: 0;
 }
 
 #chat .unread-marker:before {


maxpoulin64:
Wow, good job on this one. I would have dropped a bunch of WTFs in the chan if I were to look at this one... 👍

astorije:
> I would have dropped a bunch of WTFs in the chan if I were to look at this one...

Well, I did :D

PR #473, opened 2016-07-07
Add self info to TOGGLE messages to prevent unread marker to render for oneself

### LINKED ISSUE #466: Unread marker counts embed (...) button as new message even if sent from self
unread marker bug

Relevant code: https://github.com/thelounge/lounge/blob/9c07f2b0f2f5bb0bc14ac2900fc75bf1f12e5e82/src/plugins/irc-events/link.js#L34-L37

astorije:
Fixes #466.

Marking as `second review needed` **but** I would really like @MaxLeiter to test this as well.

I checked other places on the client where we were relying on `self` to be set to `true`, [here](https://github.com/thelounge/lounge/blob/ab609389f3f1af8ead58d356c0bdb282b487e332/client/js/lounge.js#L227), [here](https://github.com/thelounge/lounge/blob/ab609389f3f1af8ead58d356c0bdb282b487e332/client/js/lounge.js#L340), and [here](https://github.com/thelounge/lounge/blob/ab609389f3f1af8ead58d356c0bdb282b487e332/client/js/lounge.js#L872), but I didn't see anything alarming. The last link is of course the most likely to be an issue among the 3.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -32,6 +32,7 @@ module.exports = function(irc, network) {
 		}
 
 		var msg = new Msg({
+			self: data.nick === irc.user.nick,
 			type: Msg.Type.TOGGLE,
 		});
 		chan.pushMessage(client, msg);


MaxLeiter:
Works for me :+1:

PR #474, opened 2016-07-08
Prevent the app from crashing when no theme is specified

astorije:
`defaults/config.js` says:

```js
	// @default  "themes/example.css"
```

However, without this fix, not setting a theme makes the app crash with:

```js
lodash.templateSources[0]:10
((__t = ( theme ? theme : "themes/example.css" )) == null ? '' : __t) +
          ^

ReferenceError: theme is not defined
    at eval (lodash.templateSources[0]:10:11)
    at /Users/jeremie/Perso/lounge/src/server.js:103:11
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:380:3)
```

We have been setting default inline like this in other places. Even though it's not optimal, it's better than crashing :-)

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -15,7 +15,7 @@
 
 	<link rel="stylesheet" href="css/bootstrap.css">
 	<link rel="stylesheet" href="css/style.css">
-	<link id="theme" rel="stylesheet" href="<%= theme %>">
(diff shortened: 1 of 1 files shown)


astorije:
It looks like this was not needed after #387, as explained in your comment: https://github.com/thelounge/lounge/issues/475#issuecomment-231282377
I was testing this on an outdated version of the code, forgetting the branch I was in was not up-to-date with master...

I'll test and if it's working, cleanup the uses of inline defaults like this one and others.

PR #488, opened 2016-07-11
Display network name on Connect page when network is locked and info is hidden

astorije:
Would this add any value?

screen shot 2016-07-10 at 21 30 31


maxpoulin64:
Looks good to me, 👍

astorije:
Btw, this PR will conflict with #487 but I'll rebase it once the other one is merged.

PR #489, opened 2016-07-11
Send user to lobby of deleted chan when parting from active chan

### LINKED ISSUE #485: Remove client-side parting code relying on z-index
Code currently found [here](https://github.com/thelounge/lounge/blob/28938be10d5f4340031ce6a7c7b8937100fe4b8a/client/js/lounge.js#L395-L419) uses a mechanism removed [on this PR](https://github.com/thelounge/lounge/pull/465/files#diff-e5178f7b74fe45f2cfe1baf9aa1ef6faL791).

As far as tests go, I could only notice that closing a channel would focus back to the lobby of the top network, but I tested for a few seconds only, there might be other minor side effects.

Marking as `priority` as this is a regression that prevents releasing.

astorije:
Closes #485.

Instead of always sending them back to the very first lobby on the app.

This fixes a bug introduced in recent layout change (#465) but **does not bring previous behavior back**: when closing active channel, user gets sent to lobby instead of previously visited channel.

Many ways (at least three) of bringing this back were discussed on the channel but: 1) fixing the bug to an acceptable state is priority, 2) usage might show it's not necessary. Trying this at first keeps the code simple, we'll see if necessary later.

I have done some quick tests but encourage the reviewer(s) to do some more :-)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -393,29 +393,15 @@ $(function() {
 	});
 
 	socket.on("part", function(data) {
-		var id = data.chan;
-		sidebar.find(".chan[data-id='" + id + "']").remove();
-		$("#chan-" + id).remove();
+		var chanMenuItem = sidebar.find(".chan[data-id='" + data.chan + "']");
 
-		var next = null;
-		var highest = -1;
-		chat.find(".chan").each(function() {
-			var self = $(this);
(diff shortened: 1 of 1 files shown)


maxpoulin64:
Looks good to me, merging. Will open a new PR with my fix to bring back the history/ordering.

PR #490, opened 2016-07-11
add dockerfiles

davibe:


MaxLeiter:
In order to keep the repository as clean/organized as possible, a separate repo managed by @williamboman was created to host the dockerfiles: https://github.com/thelounge/docker-lounge

astorije:
Hey @davibe! Thanks for your contribution.

However, please note that there is a dedicated repository for Docker that can be found at https://github.com/thelounge/docker-lounge. See the Docker build at https://hub.docker.com/r/thelounge/lounge/.

On top of that, note that our documentations have a section for it: https://thelounge.github.io/docs/deployment/docker.html

Finally, if there is anything you think we can improve to make this other repository, the build, or the documentation more discoverable or simply better, please share :-)
@williamboman, you might want to take a look here if there any answers!

davibe:
Ok. I did not check much indeed.. Thank you for the very quick response.

maxpoulin64:
> For what is worth I prefer having those two files inside the main repo.

I can totally understand that.  We had to split it out because we were on our way to have a Caddyfile, Dockerfile, Vagrantfile, Ansible, juju, chef and puppet build scripts and... yeah you get the point :P That's why we now have a repo for all of them, which all have an independant lifecycle than the project itself. That way, we also don't have to worry about changes to the app itself breaking the automation scripts as those should in theory pull in the specific release they are designed for. No surprises, better design overall ^^

astorije:
@davibe, feel free to open PRs to suggest improvements on the [dedicated repo](https://github.com/thelounge/docker-lounge).

PR #493, opened 2016-07-12
Improve font icon management, sizing and sharpness

astorije:
I was hoping to make a dead-simple PR to fix the icon blurriness (well, at least on a Mac OS with Retina, not sure how it looks otherwise), but I ended up spending waaaay too much time on this and also improving sizing, management, and some transitions.

### Changes

From the commit messages:

1. Improve font icon rendering with font-smoothing
  This makes all icons look sharper, and reproduces [what is applied on the FontAwesome CSS](https://github.com/FortAwesome/Font-Awesome/blob/4213679/css/font-awesome.css#L19-L20)

1. Define icon font only once and fix sizing
  - This commit applies the following across the whole style:
    - `font` statement is now [the same as official FontAwesome CSS](https://github.com/FortAwesome/Font-Awesome/blob/4213679/css/font-awesome.css#L16-L17)
    - Ensure icons are never italic or bold or that other variants can be applied
    - Ensure font-size and line-height of icons are inherited from parent
    - font-family and font-smoothing is now defined only once
  - A few (mostly positive) side effects from these and related changes:
    - Header icons (main menu, context menu and user list) are now vertically centered!
    - Same applies to the Send icon, but it's more subtle there
    - Alignment of the footer icons are shifted a tiny bit
    - Server window icons are a wee bit bit bigger to match the server name font-size
    - The "Play sound" icon and text are now both 14px (was 14px / 16px)

1. Add/fix/remove some CSS transitions
  - Transition on the search icon was removed, because why was it even here?!
  - A transition was added to the "Play sound" button
  - Transition on the Send button is now consistent with the others

1. Centralize all icon definitions for better management

### Results

Note that these are are screenshots from Chrome 51 on Mac OS X El Cap with Retina, but I noticed extremely similar (if not identical) results with Firefox on the same machine.
Let me know if you see different results on other OS.

Le What | Before | After
--- | --- | ---
Sharpness | screen shot 2016-07-12 at 02 13 53 | screen shot 2016-07-12 at 02 14 02
Sharpness | screen shot 2016-07-12 at 02 13 58 | screen shot 2016-07-12 at 02 14 09
Sharpness | screen shot 2016-07-12 at 02 14 20 | screen shot 2016-07-12 at 02 14 25
Text/Icon alignment | screen shot 2016-07-12 at 02 15 40 | screen shot 2016-07-12 at 02 15 49
Sizing | screen shot 2016-07-12 at 02 14 36 | screen shot 2016-07-12 at 02 14 49

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -186,12 +186,7 @@ button {
 #footer .icon,
 #chat .count:before,
 #settings #play:before,
-#form #submit:before {
-	font: 14px FontAwesome;
-	-webkit-font-smoothing: antialiased;
-	-moz-osx-font-smoothing: grayscale;
-}
-
+#form #submit:before,
 #chat .invite .from:before,
 #chat .join .from:before,
 #chat .kick .from:before,
@@ -204,7 +199,8 @@ button {
 #chat .nick .from:before,
 #chat .action .from:before,
 .context-menu-item:before {
-	font-family: FontAwesome;
+	font: normal normal normal 14px/1 FontAwesome;
+	font-size: inherit; /* Can't have font-size inherit on line above, so need to override */
 	-webkit-font-smoothing: antialiased;
 	-moz-osx-font-smoothing: grayscale;
 }
@@ -229,7 +225,8 @@ button {
 	color: #ccc;
 	display: none;
 	float: left;
-	line-height: 40px;
+	font-size: 14px;
+	line-height: 1;
 	height: 36px;
 	margin: 6px 12px 0 -12px;
 	width: 36px;
@@ -489,6 +486,7 @@ button {
 	bottom: 4px;
 	height: 48px;
 	left: 5px;
+	font-size: 14px;
 	line-height: 48px;
 	position: absolute;
 	text-align: center;
@@ -1064,6 +1062,7 @@ button {
 	content: "\f002"; /* http://fontawesome.io/icon/search/ */
 	position: absolute;
 	right: 18px;
+	font-size: 14px;
 	line-height: 50px;
 	transition: color .2s;
 	z-index: 0;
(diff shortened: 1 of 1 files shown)


williamboman:
Nice, LGTM 👍 .

maxpoulin64:
Looks all good to me, just look at my comment and then 👍

client/css/style.css
```diff
@@ -176,6 +176,126 @@ button {
...
+#chat .topic .from:before,
+#chat .mode .from:before,
+#chat .ctcp .from:before,
+#chat .whois .from:before,
+#chat .nick .from:before,
+#chat .action .from:before,
+.context-menu-item:before {
+	font: normal normal normal 14px/1 FontAwesome;
+	font-size: inherit; /* Can't have font-size inherit on line above, so need to override */
+	-webkit-font-smoothing: antialiased;
```
maxpoulin64: If we could put `subpixel-antialiased` there instead it would be pretty great. On my computer it makes a small but noticable difference by making the icon slighly more sharp.

capture d ecran_2016-07-16_14-31-52

astorije: Sadly, we cannot: `subpixel-antialiased` is what is used by default by browsers. So setting it here on icons would kill the point of this PR and go back to ugly rendering :D

Also, note that this is what's used by the Font Awesome CSS, so I'm happy to trust them on this one. As a matter of fact, it's been attempted at different occasions by them or their contributors, see:
- https://github.com/FortAwesome/Font-Awesome/pull/3136#issuecomment-38551961
- https://github.com/FortAwesome/Font-Awesome/issues/1591#issuecomment-21711601
- https://github.com/FortAwesome/Font-Awesome/issues/611
- etc.

Also, this is what [Bootstrap does in their CSS for their icons](https://github.com/twbs/bootstrap/blob/d3cbb8eac266ee21b0e653a49f71007e720ccdb2/dist/css/bootstrap.css#L269-L279).

astorije:
@maxpoulin64, ping on my answer to your comment? Are you good to go with this PR?

PR #496, opened 2016-07-12
restore scroll position after loading previous messages

davibe:


williamboman:
Works fine for me functionality-wise on desktop and mobile. Occasionally some very minor jittering when loading more messages in landscape mode.

Are there any content that is/could be rendered asynchronously, effectively skewing the height offset? I can't think of any except thumbnails and possibly links, but those aren't expanded on old messages afaik.

astorije:
> Since the Show More button is hidden when one reaches the end of the message stack, the offset calculation is a bit off when clicking Show More the last time which causes minor flicker.

Good catch. @davibe, is it difficult to take this into account?

> Are there any content that is/could be rendered asynchronously, effectively skewing the height offset?

Not at the moment, I believe.

> I can't think of any except thumbnails and possibly links, but those aren't expanded on old messages afaik.

This seems like a bug, actually, and should be fixed.
I think we should think of the consequences of fixing this though. All the solutions I can think of right now aren't too great: caching the N+1 call to "Show more" and displaying N, previously loaded and thumbnails generated; pausing the display until everything is loaded (eww!).
Any better ideas?

davibe:
For the flickering caused by button.. we can just remove the button before doing calculations.

Dynamically (async) loaded contents should just have a fixed height (css) and, if its needed, they could expand on click one by one.

maxpoulin64:
Looks fine to me. Please squash those commits and then I'll approve this. If we can fix the small shifting when the last messages are loaded that would be cool as well but this works fine for me as it is.

astorije:
Agreed with @maxpoulin64, can you squash these commits please, @davibe?
We can improve that bump on the oldest messages, no big deal. It's still much better than it is right now :)

astorije:
@davibe, ping? :)

davibe:
I reduced to 1 commit

maxpoulin64:
👍

astorije:
👍 and merging.

PR #497, opened 2016-07-13
Add information on the About section of the client

astorije:
This rewords current version string, adds useful links to the user, and displays whether the instance is running from a release or from the git repo.
This last piece will be very helpful to debug when users come asking help on the IRC channel.

From a release | From source
--- | ---
screen shot 2016-07-13 at 03 03 48 | screen shot 2016-07-13 at 03 03 19

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -80,6 +80,16 @@ function allRequests(req, res, next) {
 	return next();
 }
 
+// Information to populate the About section in UI, either from npm or from git
+try {
+	var gitCommit = require("child_process")
+		.execSync("git rev-parse --short HEAD") // Returns hash of current commit
+		.toString()
+		.trim();
+} catch (e) {
+	// Not a git repository or git is not installed: treat it as npm release
+}
+
 function index(req, res, next) {
 	if (req.url.split("?")[0] !== "/") {
 		return next();
@@ -90,6 +100,7 @@ function index(req, res, next) {
 			pkg,
 			Helper.config
 		);
+		data.gitCommit = gitCommit;
 		var template = _.template(file);
 		res.setHeader("Content-Security-Policy", "default-src *; style-src * 'unsafe-inline'; script-src 'self'; child-src 'none'; object-src 'none'; form-action 'none'; referrer no-referrer;");
 		res.setHeader("Content-Type", "text/html");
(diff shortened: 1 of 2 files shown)


src/server.js
```diff
@@ -80,6 +80,16 @@ function allRequests(req, res, next) {
...
 }
 
+// Information to populate the About section in UI, either from npm or from git
+try {
+	var gitCommit = require("child_process")
+		.execSync("git rev-parse --short HEAD") // Returns hash of current commit
+		.toString()
+		.trim();
+} catch (e) {
+	// Not a git repository or git is not installed: treat it as npm release
```
astorije: If git is not installed, then we will falsely report that this is an npm release. However, if this is a git repo, chances are git is installed, and is available to the user running the instance.

Alternative is to check if there is a `.git` directory, but this is much more complex (not trivial to find the hash of current commit by just exploring the directory, very error-prone!). It is somewhat not 100% reliable as well as one can have a git repo without a `.git` folder (well, I didn't know), though this case must be super rare. But who knows, maybe some servers centralize `.git` folders, as some do with Python virtualenvs?

Another alternative is to use a JS implementation of git, such as https://github.com/creationix/js-git, but then we're talking über-overkill and another world of issues incoming.

To conclude: not the most reliable, but the simplest solution without a doubt, and because of the `try`/`catch`, it will not break a server for various and random reasons.

YaManicKill: > If git is not installed, then we will falsely report that this is an npm release.

Yeah, let's just ignore that. If someone has somehow got the git repository without having git installed, then we don't really care, as it's such a small thing anyway.

I think this solution is fine for what you are trying to do. The failure cases are so rare, and it's such a small thing (it'll just show it as a release rather than git) that it really doesn't matter.

YaManicKill:
I'm guessing your images are the wrong way round.

astorije:
> I'm guessing your images are the wrong way round.

It was, thanks! 😅

YaManicKill:
Great, that makes much more sense. :+1: then

maxpoulin64:
Yep, looks all good to me. No need to go crazy about a JS implementation, reasonably safe to assume people running the git version have the git binary in their path. Worst case, it's the current behavior, so meh. 👍 and merging

dgw:
Is this supposed to support instances running after `npm install -g` from the source?

`npm start` = git reference; running with my init script shows 2.0.0-pre.4.

maxpoulin64:
@dgw: it tries to find if it is a git repo, and falls back to the version in package.json. If for any reason git fails, then it will display the npm version. My guess is that your initscript's working directory is not inside the repo, so git can't find it.

PR #499, opened 2016-07-14
Ignore unnecessary files at release time

astorije:
- `client/views/` contains templates whose built version is provided with each releases
- `.gitignore` doesn't appear in our releases, but I'm not sure if this comes from npm itself or Travis CI publishing for us. Adding it here for completeness. None of the documentations (npm or Travis CI) mentions this explicitly, so ¯\_(ツ)_/¯. If someone knows more about this, please share.
- Surprisingly, `.npmignore` itself is not ignored...
- All other files and directories are for development purposes only

screen shot 2016-07-14 at 00 34 06

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-# Make sure to keep this file and .npmignore in sync
-
 node_modules/
 npm-debug.log
 
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,19 @@
-# This should match the .gitignore file without the generated assets
+# This file must not contain generated assets listed in .gitignore.
 # npm-debug.log and node_modules/ are ignored by default.
 # See https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
 
+client/views/
 coverage/
+scripts/
+test/
+.editorconfig
+.eslintignore
+.eslintrc.yml
+.gitattributes
+.gitignore
+.istanbul.yml
+.npmignore
+.stylelintrc
+.travis.yml
+appveyor.yml
+Gruntfile.js


YaManicKill:
Looks good to me.

Regarding .gitignore being ignored, it appears that is because of npm: [issue here](https://github.com/npm/npm/issues/7252). However, no one seems to explain why this is the case.

tl;dr If .npmignore doesn't exist, the .gitignore gets renamed to .npmignore in the package. If .npmignore exists, .gitignore is just ignored. You can't change this either, as putting !.gitignore doesn't fix it. So there literally seems no way to get .gitignore into an npm package without renaming it.

But I agree to keep it in there, because I expect this will be fixed at some point, and it future proofs us.

Only question is: have you checked what happens when you install the package without all these things? It will try and rerun the build, because we have the build on prepublish, which is run on publish and install. So I imagine this would fail.

astorije:
> Only question is: have you checked what happens when you install the package without all these things? It will try and rerun the build, because we have the build on prepublish, which is run on publish and install. So I imagine this would fail.

Running the build doesn't apply when you run `npm install thelounge[@next]`, it is only run when calling `npm install` without arguments. From [the doc](https://docs.npmjs.com/misc/scripts):

> - prepublish: Run BEFORE the package is published. (Also run on local `npm install` without any arguments.)

This is also why things that are part of the build (like the `font-awesome` package, or `grunt`) are listed in `dependencies` and not `devDependencies`: they would not be installed in a production environment.

YaManicKill:
Alright, well :+1: from me then. Looks sensible.

PR #500, opened 2016-07-14
Add a script to easily run PRs from external contributors

astorije:
Because I'm fed up of looking for the [right section in the maintainers' corner](https://github.com/thelounge/lounge/wiki/Maintainers'-corner#test-a-pr-locally) (because yes, I forget how to do it every single time), copy/paste, replace PR ID, ...
I'm thinking this saves some precious 3 seconds and can also help others to play with PRs without Googling too much.

Running it is as simple as:

```sh
./scripts/run-pr.sh 500
```

### DIFF
--- a/scripts/run-pr.sh
+++ b/scripts/run-pr.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+set -e
+
+if [ -z "$1" ]; then
+  echo "No pull request ID was specified."
+  exit 1
+fi
+
+git fetch https://github.com/thelounge/lounge.git refs/pull/${1}/head
+git checkout FETCH_HEAD
+npm install
+npm test || true
+npm start


maxpoulin64:
I like the idea (and personally have added a `git pr`  command for myself). However, since this is shipping with the app I think it probably would be a better idea to make this a Javascript file so it's cross-platform (even if it will be a lot more verbose). Thoughts?

scripts/run-pr.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
```
williamboman: Maybe `set -e`?

scripts/run-pr.sh
```diff
@@ -0,0 +1,12 @@
...
+if [ -z "$1" ]; then
+  echo "No pull request ID was specified."
+  exit 1
+fi
+
+git fetch origin pull/${1}/head
```
williamboman: `git fetch https://github.com/thelounge/lounge.git refs/pull/${1}/head` would probably be better as people probably don't have `thelounge/lounge` as `origin`.

maxpoulin64: I can confirm that, nice catch. Mine is configured as `upstream`, as `origin` is my github fork (and default push target).

williamboman:
Started making some changes and it got out of control real fast, but [here's how I'd do it](https://gist.github.com/williamboman/62bb3c01c07c1971b186cd53a9d1ba4a), which produces something [like this](https://asciinema.org/a/7r0knb4pz60av4ma6jr60a4pg).

YaManicKill:
I'm not sure the main repository is a good place to keep stuff like this. Maybe we should have a developers repo or something with extra tools?

astorije:
Wow, I did not expect that such a simple and consequence-free PR would trigger so many opinions, especially with all the other pending PRs more important than this one :-)

> @maxpoulin64: I like the idea (and personally have added a git pr command for myself). However, since this is shipping with the app I think it probably would be a better idea to make this a Javascript file so it's cross-platform (even if it will be a lot more verbose). Thoughts?
> @MaxLeiter: I think bash is relatively cross-platform. By implementing in javascript, do you mean adding something like creationix/js-git? If so, I think that's unnecessary, but if we use something like child_process then I'm fine with a JS implementation.

I believe bash is cross-platform enough for this, especially since Windows is now catching the train. I don't think a JS implementation is necessary, I believe this will do until it doesn't anymore. No need to prematurely over-engineer this.
Regarding js-git, absolutely not lol.

> I'm not sure the main repository is a good place to keep stuff like this. Maybe we should have a developers repo or something with extra tools?

No, the `scripts/` folder is very appropriate for that (especially since it's proposed to be removed from releases #499). The goal here is conveniency. If whenever someone wants to quickly play with a PR, they have to find another repo, clone it and call the script living in that dev dir from the `lounge` directory... Not frankly convenient!

I'm not trying to build a whole CLI here, just a tiny convenience script that will help us and other folks on master.

scripts/run-pr.sh
astorije: I left it out on purpose, so that `npm test` can fail and `npm start` still runs, without having something too complicated that plays on exit codes, outputs, ...

scripts/run-pr.sh
astorije: Good catch, I'll do that!

YaManicKill:
> Wow, I did not expect that such a simple and consequence-free PR would trigger so many opinions, especially with all the other pending PRs more important than this one

Come on, you know that bikeshedding is our favourite activity.

williamboman:
> I'm not trying to build a whole CLI here, just a tiny convenience script that will help us and other folks on master.

I think if this lands people will be finding really weird ways of using it. One example would be spinning up docker containers running the code from a specific PR.

Could also create a `thelounge-dev` npm package that just symlinks a bunch of utility scripts written in whatever language.

astorije:
> I think if this lands people will be finding really weird ways of using it. One example would be spinning up docker containers running the code from a specific PR.

A few things to that:
- This is not going to be possible from releases once #499 is merged, as the `scripts/` folder will be ignored.
- Even if not ignored, it only works from a git repository, which is not the case with a Docker build.
- This argument could be said about pretty much anything in our project, since we run from sources...
- If people want to test and run a PR... then why shouldn't they? In fact, why is that so much a big deal since this script is literally 5 commands, which are even listed in our wiki?!

> Could also create a thelounge-dev npm package that just symlinks a bunch of utility scripts written in whatever language.

No, no, no, no, absolutely no. Mark my words: this PR is **not a CLI**, it is **not a set of developer tools**, it is **not an over-engineered, bikeshedded perfect solution to a world of problem**. 😄 

It's a dead simple *script* (that belongs well in the *scripts* folder) to help us, maintainers, to review contributors' PRs, and others, power users, who desire to try and run PRs. In fact, if it only fulfills the former, I am very happy about it. Having this to remember and/or to copy and paste from [the wiki](https://github.com/thelounge/lounge/wiki/Maintainers'-corner#test-a-pr-locally) and or to have in a [git alias](https://github.com/thelounge/lounge/pull/500#issuecomment-232574244) is unnecessary and counter-productive.

Please let's merge this already, it doesn't hurt anyone, and it helps us reviewing PRs, which we do a lot. It now has more comments than LOCs lol.

maxpoulin64:
Alright, I don't really care all that much, and I have the same script locally anyway so whatever. Consider this a 👍 after the two issues pointed by @williamboman has been addressed.

astorije:
> Consider this a 👍 after the two issues pointed by @williamboman has been addressed.

Forgot the fetch thingy, fixed. `set -e` left out on purpose for reason stated above.

scripts/run-pr.sh
maxpoulin64: Hmm, then you can add ` || true` to the npm test line. That way it won't try to checkout a failed fetch, or install a failed checkout. Not exactly the prettiest thing, but at least we'd get at least a bit of error management.

williamboman:
> Even if not ignored, it only works from a git repository, which is not the case with a Docker build.

Not through the official `lounge` docker image, no. Easily doable through other means though. 

> If people want to test and run a PR... then why shouldn't they? In fact, why is that so much a big deal since this script is literally 5 commands, which are even listed in our wiki?!

Of course people should, but I think it's worth taking a few extra measures to foolproof it and ensure it works in most environments, should it be a distributed script. But yeah this is the very definition of bikeshedding.

> No, no, no, no, absolutely no.

Haha I agree.

scripts/run-pr.sh
astorije: Done in 92685f5.

astorije: `set -e` appears right after that thread of comments.

PR #503, opened 2016-07-15
Keep track of the previous channels

maxpoulin64:
Following #489, this re-adds the old behavior of returning to the previous channel on close. While not exactly super intuitive, it still more useful than going to a server window even if it's random and has the advantage of bringing you back to where you were after a `/whois` or ZNC's `*status` or after chatting with a bot such as NickServ or ChanServ.

I don't see how reordering the DOM like that could have a negative impact since the others are `display:none` anyway, and completely removes the need for the z-index trickery.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -395,13 +395,13 @@ $(function() {
 	socket.on("part", function(data) {
 		var chanMenuItem = sidebar.find(".chan[data-id='" + data.chan + "']");
 
-		// When parting from the active channel/query, jump to the network's lobby
-		if (chanMenuItem.hasClass("active")) {
-			chanMenuItem.parent(".network").find(".lobby").click();
-		}
-
 		chanMenuItem.remove();
 		$("#chan-" + data.chan).remove();
+
+		if (chanMenuItem.hasClass("active")) {
+			var nextChanId = chat.children().last().data("id");
+			sidebar.find(".chan[data-id='" + nextChanId + "']").click();
+		}
 	});
 
 	socket.on("quit", function(data) {
@@ -781,6 +781,7 @@ $(function() {
 		document.title = title;
 
 		if (self.hasClass("chan")) {
+			chan.appendTo(chat);
 			$("#chat-container").addClass("active");
 			setNick(self.closest(".network").data("nick"));
 		}


xPaw:
How expensive is the DOM shuffling in this case?

maxpoulin64:
I haven't felt any difference so far personally but I don't have hundreds of tabs open either tho. I don't see why it would be that expensive they're all hidden: only the visible one ends up being rendered anyway.

xPaw:
Mind checking in devtools what perf it has?

astorije:
I'm still not set about messing with the DOM for something like this. Partly because of prejudice (the common rule of "Don't mess with the DOM if you don't have to", and we don't have to here) and therefore seems like overkill to me, partly because of potential unintended consequences.

A few things worth thinking of:

- If user has preloaded a ton of previous messages, does this have a performance issue?
- What about low-end devices? Is there a chance the app crashes on them when closing a chat?
- Do we lose any kind of metadata when shuffling this, such as event handlers (can't think of anything else)?
- Do prefetched content gets prefetched again?
- Is the drag-and-drop affected by this?
- etc.

Sorry to be that guy, I'm just afraid to trigger a whole new world of bugs when we thought this was consequence-free. Maybe (and hopefully) the answer to all these questions is no, but it's worth being sure.

maxpoulin64:
You guys are really way too scared of the DOM. It's an array. It's sorting a fucking array. The DOM is only a data structure. Adding and removing a node to the DOM is just as fast as an object from an array and putting it back in.

What's expensive is layout **changes** and redraws. This causes neither a relayout nor a redraw, because nothing visible changes. The renderer loads all nodes in that tree and only one is displayed and will be laid out again. Its dimensions doesn't change, so nothing happens. End of it.

Actually, I can't even measure it because the sticky scroll on the other hand does force a redraw each time it is triggered (due to scroll).

And on top of that, it HAS to improve performance because the z-index only changes the stacking order. Unless they were set to `display: none;`, they will all be redrawn each on top of the other because each layer can be transparent. At its best it does exactly the same as reordering the DOM, but abusing CSS for it instead.

The answer to all of your concerns is "No", both as per what I just said and also because I've been running it for weeks and didn't see any difference.

Avoiding to touch the DOM just because someone, somewhere said that because they don't understand shit about a rendering engine and made a blog article on how evilt he DOM is, is stupid. It only leads to even worse workarounds that gives the impression of being better (see the z-index we had before and try to tell me it's not bad) just because it doesn't touch the oooo scary DOM. The DOM is fine and fast, you just have to not be a complete idiot with it and understand the consequences of what you change.

astorije:
Well someone was grumpy :D

Just to be clear, I think the whole debate is not "your solution vs. the `z-index` way" but "your solution vs. the current way" (current way being send the user back to the network window of that server when closing the active window).

And in that regard, whether there is a performance issue or not, I do think it's a bit overkill to move the entire content of chan in the small eventuality that a user closes a window **and** that this is the active window.
Putting all my comments and concerns above aside, I see 2 issues:

- This is per-network, so if I'm on network A, channel 1, then go to network B, channel 1 and close it, I will get sent to the last channel **on network B** that I visited. This was already confusing enough on Shout (this behavior only works when closing multiple channels in a row, otherwise it's a real usability mess), it's worse here.
- In a big channel list, the user may be sent to another channel far away. Slack does this and it's so messed up that you cannot not grumble about this :D

I think we are trying to find the most computer-sciency way to solve this. I would say the 2 following options are nicer in terms of usability:

1. When closing the active window, send the user back to the server window (as currently).
2. When closing the active window, send the user back to the window right above or below the one just closed.

I like the second option because it guarantees not jumping the users all over the place. What do you think? Does that seem reasonable?
(Option 2 is what Chrome does when closing active tabs, with the slight difference that if user opens a tab, stays on that tab whatever he does there, then closes it, they get sent back to the tab they were on when opening it, but this is confusing in many ways so I think we should stay clear of that, at least for now)

xPaw:
@astorije Are you fine with this approach and can we merge it?

astorije:
> @astorije Are you fine with this approach and can we merge it?

Nah, I think this is very confusing in terms of UX. The 2 biggest concerns (but not the only ones) I have are the following:

- This does not work across networks. If I'm on network A, channel X, then move to channel Y living on network B, and close channel Y, I will not move back to channel X but to whatever I was on previously on network B.
- As @YaManicKill mentioned in #573 (actually, it might have been on the IRC channel), it's tricky to move across channels because of potentially unread messages/markers. While it's ok to move back to the channel we very last visited, any further step risks to mark channels as read.

I think @YaManicKill's alternative is by far the best: when closing the current channel, user gets sent to a blank screen. That allows for 2 independent improvements:

1. That blank screen can be transformed into an informative dashboard (summary of open channels, missed notifications, errors/disconnects while away (or not), etc.) and become our landing page.
2. To fulfill @maxpoulin64's use case, we can implement something similar to what Chrome does: after opening a new window, closing it could transfer the user the the previous active channel **only** if they haven't switched across channels. This only works for the last one and does not bubble like an MRU cache because this is terrible for UX (users do not remember where they were more than 1 step before and asking them to do so is a cognitive exercise we should not do).

I believe this is the most optimal UX decisions we could implement (well at least, the most optimal **I** can think of at the moment lol).
Note that both **1** and **2** can happen separately, but until any of these happen, I think we should keep the current mechanism. I'll try to take a stab at either of them when I have some spare time.
My take on this is to close both PRs (but #573 is already closed) and rely on current mechanism for now.

YaManicKill:
Should we close this? Looks like it's not going to go anywhere. Any objections?

xPaw:
While I think this solution is fine, I will close it due to no activity, and will leave it to the "Home" tab idea.

PR #504, opened 2016-07-15
Use min-height of textarea when computing auto-resize after deleting a char

### LINKED ISSUE #482: Text in chat input invisible in landscape mode (mobile)
https://streamable.com/h6dc

astorije:
Fixes #482.

This fixes a bug that was breaking the textarea on Android.
@williamboman, can you give it a try?

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1263,6 +1263,7 @@ button {
 	background: transparent;
 	border: none;
 	font: inherit;
+	min-height: 18px; /* Required when computing input height at char deletion */
 	height: 18px;
 	max-height: 90px;
 	line-height: 1.5;
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -632,8 +632,11 @@ $(function() {
 		.history()
 		.on("input keyup", function() {
 			var style = window.getComputedStyle(this);
-			this.style.height = "0px";
-			this.offsetHeight; // force reflow
+
+			// Start by resetting height before computing as scrollHeight does not
+			// decrease when deleting characters
+			this.style.height = this.style.minHeight;
+
 			this.style.height = Math.min(
 				Math.round(window.innerHeight - 100), // prevent overflow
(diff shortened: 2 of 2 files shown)


client/css/style.css
```diff
@@ -1260,6 +1260,7 @@ button {
 }
 
 #form #input {
+	overflow-y: hidden;
```
maxpoulin64: This will break scrolling when it hits 6-7 lines, as we stop growing at this point.

client/js/lounge.js
```diff
@@ -628,12 +628,23 @@ $(function() {
...
 	var input = $("#input")
 		.history()
 		.on("input keyup", function() {
 			var style = window.getComputedStyle(this);
-			this.style.height = "0px";
-			this.offsetHeight; // force reflow
+
+			if (!initialInputHeight) {
+				// When first resizing the textarea, store its initial height for later
+				initialInputHeight = this.style.height;
```
maxpoulin64: There's no need to store it this way. You already have it in `style.minHeight` which contains the minimum height which is normally what prevents the 0px from applying in non-retarded browsers. So we can just use this one instead.

client/css/style.css
astorije: Right, good point, I'll remove this.

client/js/lounge.js
astorije: Actually, there is a height, not a min-height at the moment in https://github.com/thelounge/lounge/blob/ebab910e0013b15ff0bcc754b2003fb40e7d2298/client/css/style.css#L1262-L1275, so I need to add it. Will do.

williamboman:
Functionality-wise works fine, :+1:.

astorije:
@maxpoulin64 is 👍 (but didn't comment with 👍 itself, just the second-review label). All other maintainers are away at the moment, but @williamboman did test this successfully, so merging.

Issue #442, opened 2016-07-02
Add persistent storage


willwh:
We should add persistent storage to the server, for logging, users, etc.

Using something like [node-any-db](https://github.com/grncdr/node-any-db) might be a good idea, to allow people to configure with their desired database in config.

willwh:
https://github.com/thelounge/lounge/issues/109

YaManicKill:
So...this I think should be pretty high priority.

We already have a PR for this #501, but ....

I guess we need to make a decision about what we should use for this, and then run with that. Snoonet have just set up a lounge instance and this is their biggest thing, and this would solve the memory leak we have (in a non-horrible way).

Here are the options:

1. Just choose 1 backend and stick with that. sqlite for example
2. Use something like http://docs.sequelizejs.com/en/v3/
3. Use something like node-any-db

I suggest we go for 2.

2 is a nice balance between 1 and 3. We support multiple db options (sqlite, mysql, etc) but we don't go down the full ORM way of having a library that is a jack of all trades, master of none (which is definitely my experience with something like waterline).

Any objections with going with 2 (we don't have to go with that exact one)?

MaxLeiter:
I'd hate to see the work @astorije has put into #663 go to waste, but I think 2 is our best long-term bet.

YaManicKill:
@MaxLeiter That definitely won't be wasted. It will be an option for people who don't want to use a database.

Also, #663 wouldn't be the only solution ever anyway, as it doesn't solve the in-memory issue. It is only for loading messages from the logs on startup. While we are running it does nothing.

So it's still useful on it's own.

PR #506, opened 2016-07-15
Add basic mysql logging support - WIP

willwh:
Please add a do not merge label :)

Don't forget to `npm install`

There is a helper script at `scripts/irclogs.sql` that will create a database user, database, and a table for storing your logs. You should edit the password that is set in the script before running it.

You need to set the following in your config.js:

```
mysql: {
  enabled: true,
  database: 'database_name',
  dbuser: 'database_user',
  dbpass: 'database_password',
  dbhost: 'database_host',
  maxLogs: 100
}
```

If you haven't run the SQL helper script, you need to create a database, and then a table called `logs` like so (this will get moved to config):

```
create table logs (
    id INT(6) UNSIGNED AUTO_INCREMENT PRIMARY KEY,
    name VARCHAR(30) NOT NULL,
    host VARCHAR(64),
    channel VARCHAR(30),
    message VARCHAR(600),
    datetime TIMESTAMP
);
```

I chose VARCHAR(600) for messages, which is a bit bigger than the suggested max of an irc message defined in [RFC 2812](https://tools.ietf.org/html/rfc2812),  which says:

```
IRC messages are always lines of characters terminated with a CR-LF (Carriage Return - Line Feed) pair, and these messages SHALL NOT exceed 512 characters in length, counting all characters including the trailing CR-LF. Thus, there are 510 characters maximum allowed for the command and its parameters.
```

Logs look like this:

```
mysql> select * from logs order by datetime desc limit 3;
+----+-------------+----------+-----------+---------------------------+---------------------+
| id | name        | host     | channel   | message                   | datetime            |
+----+-------------+----------+-----------+---------------------------+---------------------+
| 78 | lounge-user | Freenode | #bot_test | gfjhg                     | 2016-07-15 10:08:49 |
| 77 | lounge-user | Freenode | #bot_test | hi!                       | 2016-07-15 10:08:37 |
| 76 | lounge-user | Freenode | #bot_test | testing, testing, 1, 2, 5 | 2016-07-15 09:48:20 |
+----+-------------+----------+-----------+---------------------------+---------------------+
```

Would love feedback and help :)


spencerthayer:
This is awesome. Does it work?

willwh:
All it does right now is actually log messages to the database, but that does work. I haven't had the time do anything else yet :)

spencerthayer:
If you can get this to work this will change everything!

kode54:
Don't forget, in case it's not already set for the server, utf8mb4 encoding. MySQL and probably also MariaDB need that for full character set range, or else you start chucking out things like emoji.

richrd:
I thought I commented on database logging before, but can't find it now.

Anyway first off I fully support a database option. However I really hope it'll be an option that can be enabled or disabled (which ever the default will be). I don't want logs to be stored on disk, I just don't need it. I'm fine with having them in RAM (but with a configurable maximum limit for the amount of messages).

That said it does looks like this PR does include a config boolean for the db, that would be perfect! And as I mentioned this is a great feature to have :+1:

willwh:
@spencerthayer so I have this working now, thanks to @maxpoulin64 and @xPaw; it still needs some pretty serious cleanup, although, it now stores and loads logs on boot.

i.e. There is a `maxHistory` property in `config.js` and we need something similar for the mysql fetching, so we're not returning 1000's of lines to the client on start etc ;)

To ease the setup a little bit, I've also added `scripts/irclogs.sql` which will create a user, and database, and table to store everything. Make sure you change the password in the script before running it ;)

Run it like so: (MySQL 5.7.6+)

```
mysql < scripts/irclogs.sql
```

I would love to build in a command to do the initial mysql setup from lounge.

The last hurdle to make this really usable too, is to think long and hard about how we want to return "historical" logs..... I'd actually propose something like added a new button in the bottom left tool bar, "Search Logs" or something, that would allow you to search historical logs.

The way the client is currently written, I can forsee some challenges in trying to load things with the "Show more messages" type functionality we have now, which is why I am suggesting a search type interface.

Would love others to chime in with their thoughts, and thanks everyone for the hard work, `thelounge` is a sweet IRC client.

p.s. next on my agenda is [Twilio](https://www.twilio.com/) integration for highlights to be SMS'd to a user in config, if they have no client attached :)

YaManicKill:
Left some discussion about this on #442

YaManicKill:
Hey, @willwh, are you still interested in working on this?

willwh:
I've been running this for ages on my own branch; it's working really nicely for me. I'd love to tie in better log return, but just don't have time currently :)

I wrote a little log crawler interface but it should probably be rolled in to the client!

YaManicKill:
Would you have time to rebase and solve the conflicts? If not, would you be happy to let someone else clean it up? Would be good to have. Based on how this currently is, and the fact that it's an opt-in feature, I'd be happy with the state that it's in, but obviously will have to see how the other maintainers feel.

xPaw:
Can we default this to sqlite for easier testing and shipping? I think eventually we could use it for complete storage.

package.json
```diff
@@ -49,7 +49,8 @@
     "fs-extra": "0.30.0",
     "irc-framework": "2.5.0",
     "lodash": "4.11.2",
-    "moment": "2.13.0",
+    "moment": "^2.13.0",
```
xPaw: Revert this line.

scripts/irclogs.sql
```diff
@@ -0,0 +1,15 @@
+-- This script requires MySQL 5.7.6+
+CREATE DATABASE IF NOT EXISTS irclogs;
+CREATE USER IF NOT EXISTS 'thelounge'@'localhost' IDENTIFIED BY 'WLJg123r0823r@!#RPI!H@RQ';
```
xPaw: Don't include user stuff in SQL file, plan for minimum permissions.

src/plugins/irc-events/message.js
```diff
@@ -84,5 +85,9 @@ module.exports = function(irc, network) {
 			highlight: highlight
 		});
 		chan.pushMessage(client, msg);
+    // Log to mysql
+		if (Helper.config.mysql.enabled && data.type === "message" && msg.from !== "") {
+			db.log(msg.from, network.name, chan.name, msg.text);
```
xPaw: You're losing a lot of information by not storing the entire msg object.

src/plugins/irc-events/message.js
```diff
@@ -1,5 +1,7 @@
 var Chan = require("../../models/chan");
 var Msg = require("../../models/msg");
+var Helper = require("../../helper");
+var db = require("../../database.js");
```
xPaw: Remove .js from requires.

bews:
> Can we default this to sqlite for easier testing and shipping? I think eventually we could use it for complete storage.

Big +1 to this.

spencerthayer:
SQlite would be ideal for me.

YaManicKill:
Hey, @willwh, would you have any time to switch this over to sqlite and fix conflicts, or should we close it if that's not going to happen? This is still something we'd love to have.

astorije:
Hey @willwh, sorry but we are going to close this for inactivity, and because some very recent work has started on that in #1529. Feel free to comment in this PR or the new one if you have questions or want to resume work on this :)

Either way, thanks a lot for your help and contributions, even if this one didn't make it!

PR #515, opened 2016-07-18
views/chat: set title attribute on topic

williamboman:
The `title` attribute is only set [when receiving a topic change event](https://git.io/vKVC1) right now, not on page load.


astorije:
Good catch, thanks!

PR #523, opened 2016-07-23
Remove table layout for chat messages (and fix layout issues yet again)

maxpoulin64:
Fixes this:
capture d ecran_2016-07-22_22-04-49

back into this:
capture d ecran_2016-07-22_22-05-12

EDIT: So, this still had bugs, so I removed the table layout entirely. This one seems to just play nice now.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -929,12 +929,9 @@ kbd {
 	padding-right: 10px;
 	text-align: right;
 	width: 134px;
-	align-self: stretch;
 }
 
 #chat .text {
-	margin: auto;
-	overflow: hidden;
 	flex: 1 1 auto;
 }
 
@@ -999,14 +996,6 @@ kbd {
 
 #chat .self .text {
 	color: #999;
-}
-
-#chat .msg.motd .text,
-#chat .msg.message .text,
-#chat .msg.action .action-text,
-#chat .msg.notice .text {
-	white-space: pre-wrap;
-	overflow: hidden;
 }
 
 #chat .msg.channel_list_loading .text {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -832,15 +832,15 @@ kbd {
 }
 
 #chat .messages {
-	display: table;
-	table-layout: fixed;
-	width: 100%;
 	padding: 10px 0;
 }
 
 #chat .msg {
 	word-wrap: break-word;
 	word-break: break-word; /* Webkit-specific */
+	display: flex;
+	overflow: hidden;
+	position: relative;
 }
 
 #chat .unread-marker {
@@ -912,16 +912,15 @@ kbd {
 #chat .time,
 #chat .from,
 #chat .text {
-	display: table-cell;
+	display: block;
 	padding: 2px 0;
-	vertical-align: top;
+	flex: 0 0 auto;
 }
 
 #chat .time {
 	color: #ddd;
 	text-align: right;
-	max-width: 46px;
-	min-width: 46px;
+	width: 46px;
 }
 
 #chat .from {
@@ -929,8 +928,14 @@ kbd {
 	color: #b1c3ce;
 	padding-right: 10px;
 	text-align: right;
-	max-width: 134px;
-	min-width: 134px;
+	width: 134px;
+	align-self: stretch;
+}
+
+#chat .text {
+	margin: auto;
+	overflow: hidden;
+	flex: 1 1 auto;
 }
 
 #loading a,


astorije:
I had played with removing the table layout for a bit, and after noticing so many bugs, I simply canceled. I need to test this PR thoroughly before we can merge or we'll do like we've been doing for a few weeks: one step forward, N steps backwards.

xPaw:
Just wondering, what happens with very long names or notices? How do they overflow

maxpoulin64:
@xPaw: they wrap on multiple lines. Worse case they push the column for that one message, which is slighly uglier but has the advantage of not fucking up the whole table, so I think it's worth it anyway. I didn't change the wrapping code, so if there's something it probably was already there.

astorije:
> which is slighly uglier but has the advantage of not fucking up the whole table, so I think it's worth it anyway.

Do you have any screenshot of this by any chance?

maxpoulin64:
>> which is slighly uglier but has the advantage of not fucking up the whole table, so I think it's worth it anyway.

> Do you have any screenshot of this by any chance?

No, I haven't managed to trigger it yet. So far it seems like it does the same as it did with the table, the nick will just wrap on a second line. But if it were to happen, it would look a bit like this:

capture d ecran_2016-07-24_23-18-52

But it looks like this instead (but it also does on master):
capture d ecran_2016-07-24_23-20-19


I can also confirm long messages such as long links wrap properly and don't overflow.

williamboman:
screenshot 2016-08-07 at 16 37 28

Looks like this messes up existing border color on non-default themes (tested only on Zenburn).

xPaw:
What about using flexbox instead of floats? The textarea already does.

astorije:
Just curious, why removing the (CSS) table layout? I didn't know much about the use cases for this but it looks like it matches our goal fairly well, no?

maxpoulin64:
@astorije because it keeps exploding, especially with the unread marker?

Good point for the flexbox, will definitely go with that option.

maxpoulin64:
Updated to flexbox

YaManicKill:
I'm going to stick this PR on my running version and see how it works over the next few days. That's how I found all of the bugs in @astorije's one. But this looks promising, that's for sure.

astorije:
> That's how I found all of the bugs in @astorije's one.

🔪💔😭 ... 😄

> See the screenshot in the first post and try to tell me this is normal, expected layout for it. It's horrible.

Agreed, that's gross.
I'll test this as well. Would be pretty awesome if we can get this sh*t behind us once and for all!

maxpoulin64:
> That's how I found all of the bugs in @astorije's one.

That's how I test mines 😛 Not uncommon that I do a bunch of updates after opening the PR

YaManicKill:
Found my first bug.

Copy a line on the chat
Paste it into the textarea
You get 3 lines: the time on 1, the username on the next, and the message on the final one.

astorije:
Is that a big bug though?
Worst case, can we intercept the copy to put them back into one line? It feels hackish but I don't think there is another option. More importantly, it could let us improve the copy and paste experience, such as wrapping nick with `<>` and padding the nicks when copying multiple lines.

YaManicKill:
> Is that a big bug though?

astorije:
> What I would really like to figure out is why is it putting in newlines in the first place?

Most likely because of [this switch](https://github.com/thelounge/lounge/pull/523/files#diff-97db1f70168fb5f12457b238ff6052b5L818). Again, I don't think there is anything we can do without breaking the layout again. Maybe try with `inline-block`?

> That would be an acceptable solution, yeah.

Right, but definitely for another PR.

astorije:
@maxpoulin64, given @YaManicKill's last comment and the fact that the 2.0.0 issue list is never ending (we have had 8-12 open issues/PRs there forever now...), do you think we could move that to the next release?
Current master works fine for v2.0.0 I think. Definitely looking forward to extensively test this once the 2.0.0 list is empty though.

astorije:
As per my previous message, I'm removing this from 2.0.0, I'll get back to testing and reviewing this right after we ship. Though I'm looking forward to this, no reason to actually block release for it.

client/css/style.css
```diff
@@ -762,15 +762,15 @@ button {
...
-	width: 100%;
 	padding: 10px 0;
 }
 
 #chat .msg {
 	word-wrap: break-word;
 	word-break: break-word; /* Webkit-specific */
+	display: flex;
+	overflow: hidden;
+	position: relative;
```
astorije: I could not figure out why these last 2 lines (`overflow: hidden;` and `position: relative;`) are necessary. It seems like disabling them doesn't do anything for me.
Do you remember, by any chance?

client/css/style.css
```diff
@@ -815,25 +815,30 @@ button {
...
-	vertical-align: top;
+	flex: 0 0 auto;
 }
 
 #chat .time {
 	color: #ddd;
 	text-align: right;
-	max-width: 46px;
-	min-width: 46px;
+	width: 46px;
```
astorije: I remember I had to set these 2 explicitly, but I can't remember why specifically: https://github.com/thelounge/lounge/commit/8118d56e16d2ab9a2321badffb4ace2d08d5c278#diff-97db1f70168fb5f12457b238ff6052b5L771

Chances are, this was table-layout-specific anyway...

client/css/style.css
```diff
@@ -815,25 +815,30 @@ button {
...
 }
 
 #chat .from {
 	border-right: 1px solid #f6f6f6;
 	color: #b1c3ce;
 	padding-right: 10px;
 	text-align: right;
-	max-width: 134px;
-	min-width: 134px;
+	width: 134px;
```
astorije: Same here, :shrug_emoji:
https://github.com/thelounge/lounge/commit/3cddbbce6eca34843c7cc4b876a9ecb12602672e#diff-97db1f70168fb5f12457b238ff6052b5L754

I guess this is why we should really add some comments to our code...

client/css/style.css
```diff
@@ -815,25 +815,30 @@ button {
...
+	align-self: stretch;
```
astorije: I can't seem to see a change when I enable/disable this. Remember why?

client/css/style.css
```diff
@@ -815,25 +815,30 @@ button {
...
+}
+
+#chat .text {
+	margin: auto;
+	overflow: hidden;
```
astorije: I believe this is unnecessary as it is shadowed by L907/R907 anyway.
Actually, that whole block seems unnecessary with this flexbox layout now:

```css
 #chat .msg.motd .text,
 #chat .msg.message .text,
 #chat .msg.action .action-text,
 #chat .msg.notice .text {
 	white-space: pre-wrap;
 	overflow: hidden;
 }
```

client/css/style.css
```diff
@@ -815,25 +815,30 @@ button {
...
 	padding-right: 10px;
 	text-align: right;
-	max-width: 134px;
-	min-width: 134px;
+	width: 134px;
+	align-self: stretch;
+}
+
+#chat .text {
+	margin: auto;
```
astorije: Not sure the specific reason for this, but this is adding a top margin when text is one line but "from" goes over multiple lines. Example:

`margin: auto;` | ~~`margin: auto;`~~
--- | ---
screen shot 2016-09-20 at 00 41 59 | screen shot 2016-09-20 at 00 42 42

client/css/style.css
```diff
@@ -762,15 +762,15 @@ button {
...
-	display: table;
-	table-layout: fixed;
-	width: 100%;
 	padding: 10px 0;
 }
 
 #chat .msg {
 	word-wrap: break-word;
 	word-break: break-word; /* Webkit-specific */
+	display: flex;
```
astorije: @Gilles123 found out that these still need prefixing for now: https://github.com/thelounge/lounge/issues/624#issuecomment-248477595
Mind adding them to this PR please? :)

nornagon: with [autoprefixer](https://github.com/postcss/autoprefixer) and #609 this could be done magically behind the scenes :)

nornagon:
FYI, not directly related to this PR but still relevant: with `flex-direction: column-reverse` on the message list, we can almost entirely drop the `stickyscroll` jQuery plugin. `column-reverse` changes the way scrolling behaves when the DOM changes to consider the _bottom_ of the element the "sticky" part, so adding a new message automatically scrolls the div down—no JS required! You still need a little JS (~a few lines) to keep the scroll position stuck if a message is added while you're scrolled up, but it's much simpler.

https://github.com/nornagon/lounge/compare/react...nornagon:react-chat#diff-97db1f70168fb5f12457b238ff6052b5R768 if you're curious.

xPaw:
@nornagon That makes selecting messages weird. And copying the messages makes them backwards.

nornagon:
@xPaw oh huh, i missed that! weird, that seems like a bug :/

xPaw:
To change how nicks overflow (instead of jumping to new line, extend it to right, this CSS needs to be added:
```
#chat .from {
	padding-left: 10px;
	min-width: 134px;
}
```

astorije:
@maxpoulin64,  I know you're locally running an older version of that PR, before you moved to flexbox. Do you mind updating your instance to run the latest version, so it lives in prod somewhere before meeting?

I'd love to ship this, but we need to make sure the issues highlighted above are fixed beforehand.

astorije:
@maxpoulin64, any chance you could be giving another stab at this?

xPaw:
@astorije can you take this over?

astorije:
> @astorije can you take this over?

Will do. Do you have a list of requirements you think are important for this on the top of your head?

xPaw:
@astorije just the stuff me and you pointed out.

astorije:
This PR was 370 commits behind 😱 

I have rebased it and I'm quickly testing it, but if someone (@thelounge/maintainers? @williamboman? @MaxLeiter?) would like to give it a try by merging it on their prod instance to have it run over multiple days, that would be very useful.

In general, does it seem like a good idea to switch from a CSS table layout (this is not evil, calm down 😅) to flexbox? What do we gain and lose?

williamboman:
I've ran this on and off multiple times since it was opened, never ran into issues with it (other than it breaking other PRs such as #759). LGTM

client/css/style.css
astorije: I have been trying to fix this and none of `align-items: flex-start;`, `align-content: flex-start;` or `self-align: flex-start;` did the trick.
I guess I'll just remove `margin: auto` unless someone has reasons to believe this is going to break something else.

client/css/style.css
astorije: According to https://css-tricks.com/snippets/css/a-guide-to-flexbox/, `stretch` is the default value of `align-items`, and `align-self` is used to override it. Since this is not using `align-items`, there is nothing to override. Removing.

client/css/style.css
astorije: Playing with this more, I can't trigger something that needs it (at least on Chrome). Removing both.

astorije:
@YaManicKill, @williamboman, I have made a couple changes and rebased with master. Do you mind testing this a bit on your instances? Thanks!

PR #526, opened 2016-07-24
Fix duplicate nicks (maybe)

maxpoulin64:
Due to the rarity of this bug, I suspect there is some kind of race condition going on where we send a `NAMES` request, and in the meantime users have joined on the server so we get them back as a result. But then we also receive their respective `JOIN`s after the names list, so we add them to our existing list.

Well even if it's not it, it should fix the problem anyway so whatever.

### DIFF
--- a/src/plugins/irc-events/join.js
+++ b/src/plugins/irc-events/join.js
@@ -1,3 +1,4 @@
+var _ = require("lodash");
 var Chan = require("../../models/chan");
 var Msg = require("../../models/msg");
 var User = require("../../models/user");
@@ -17,11 +18,15 @@ module.exports = function(irc, network) {
 				chan: chan
 			});
 		}
-		chan.users.push(new User({nick: data.nick, modes: ""}));
-		chan.sortUsers(irc);
-		client.emit("users", {
-			chan: chan.id
-		});
+
+		if (!_.find(chan.users, {nick: data.nick})) {
+			chan.users.push(new User({nick: data.nick, modes: ""}));
+			chan.sortUsers(irc);
+			client.emit("users", {
+				chan: chan.id
+			});
+		}
+
 		var msg = new Msg({
 			time: data.time,
 			from: data.nick,


maxpoulin64:
To anyone who experience the issue (I never saw it ever), it would be great if you could merge [this branch](https://github.com/thelounge/lounge/commit/e0e722adf7a927d2096d89fb33c55314455bc0c6) in your instance and paste us the log when you notice duplicate nicks in the list. This adds debug information everywhere where nicks are added to the list and will help me pinpoint where the fuck exactly they come from.

```
git fetch https://github.com/thelounge/lounge.git debug/dup-nicks
git merge FETCH_HEAD
# restart lounge for changes to take effect
```

src/plugins/irc-events/join.js
```diff
@@ -17,11 +18,15 @@ module.exports = function(irc, network) {
 				chan: chan
 			});
 		}
-		chan.users.push(new User({nick: data.nick, modes: ""}));
-		chan.sortUsers(irc);
-		client.emit("users", {
-			chan: chan.id
-		});
+
+		if (!_.find(chan.users, {nick: data.nick})) {
```
astorije: This seems like a poor man's fix to me, you know one of these fixes trying to fix the symptom without tackling the root cause itself.
I'm not sure why this check is necessary, could you add a comment here so we don't scratch our head 6 months from now?

Also, why is this necessary here and not in the `client.emit("join" ...` or the `chan.pushMessage(client, msg)`?

astorije:
@maxpoulin64, I have never, ever had that issue either, whatever version of The Lounge, slate-irc or irc-framework, Node.js, ... (I am not using ZNC though)
Do we want to commit to this PR? I am not too convinced by it, and it seems like you feel it might not even fix the issue...

Is there an associated issue? I couldn't find it in the issue list.

My opinion is to close this PR for now to clean up the PR list since we won't merge, and wait to see if someone reports this issue again. If they do, let's open an issue, you point them to your branch, and we take it from here. Sounds good?

maxpoulin64:
Yeah, I have no idea how people manage to get that bug. Might as well be a node issue on their end as well, seems like these people also tend to be using quite massively outdated node versions.

PR #527, opened 2016-07-24
Reset the unread marker on channel change

maxpoulin64:
This restores the old behavior of resetting the unread marker on channel change, as that's usually at this point one wants to check for new messages and is also what matches on the server. I feel this is overall more consistent and useful, and also more in line with what other clients do.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -764,9 +764,13 @@ $(function() {
 			.find(".chat")
 			.unsticky();
 
-		lastActive
+		var lastActiveChan = lastActive
 			.find(".chan.active")
 			.removeClass("active");
+
+		lastActiveChan
+			.find(".unread-marker")
+			.appendTo(lastActiveChan.find(".messages"));
 
 		var chan = $(target)
 			.addClass("active")


client/js/lounge.js
```diff
@@ -764,10 +764,14 @@ $(function() {
...
 			.unsticky();
 
-		lastActive
+		var lastActiveChan = lastActive
 			.find(".chan.active")
 			.removeClass("active");
 
+		lastActiveChan
+			.find(".unread-marker")
+			.appendTo(lastActiveChan.find(".messages"));
```
astorije: Wouldn't this work just fine?

```js
lastActive
			.find(".chan.active")
			.removeClass("active")
			.find(".unread-marker")
			.appendTo(lastActiveChan.find(".messages"));
```

(I haven't tried, of course 😅)

maxpoulin64: No, because the active *window* contains all the channels, so we need to find the specific *channel*. That was my initial solution before I ended up with a million unread markers in ever channels.

astorije:
@maxpoulin64, I get your fix, but by any chance did you find out where this bug was introduced? I mean, at some point it was working, and then it was not anymore. I'd like to find the exact piece of code that broke it, for completeness.

astorije:
So it seems to me that it's [this](https://github.com/thelounge/lounge/commit/37b5f28919acb8109af4bcdac645315fd3fbc32f#diff-e5178f7b74fe45f2cfe1baf9aa1ef6fa) and/or [that](https://github.com/thelounge/lounge/commit/26bf948fdc600ec0294094861f127e3d2360c095) but I can't pinpoint the exact change that triggered the bug (I didn't look too closely though). Can you?

maxpoulin64:
That would be https://github.com/thelounge/lounge/commit/26bf948fdc600ec0294094861f127e3d2360c095

And wow, I really pushed that fix? Damn, I really was tired that week. I thought it was that way by design at first with the new style.

Well, either way, fixed :S

astorije:
👍 and merging.

Issue #439, opened 2016-07-01
Make custom highlights case-insensitive


williamboman:
There are currently no clarification whether the keywords are case-sensitive or not

astorije:
Should they be case-sensitive?

dgw:
I would say the code should be modified to make them case-insensitive.

YaManicKill:
Yeah, this is definitely a bug with the code, they shouldn't be case-sensitive. I just didn't thin kabout that.

PR #528, opened 2016-07-27
Match custom highlights as case-insensitive regexes

nightfly19:
Match custom highlights as case-insensitive regular expressions.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -204,7 +204,7 @@ $(function() {
 		var template = "msg";
 
 		if (!data.msg.highlight && !data.msg.self && (type === "message" || type === "notice") && highlights.some(function(h) {
-			return data.msg.text.indexOf(h) > -1;
+			return data.msg.text.match(new RegExp(h, "i"));
 		})) {
 			data.msg.highlight = true;
 		}


maxpoulin64:
Hmm, do we really want that, though? By using a regex, the user needs to be aware it is parsed as a regex, otherwise it can lead to errors. For instance, `^Max-P` and `[Anna]` are two nicks know are used and would cause issues if typed-in as-is. Maybe parsing the box to see if it is in the "/regex/modifier" format could be a better fit.

Or a plugin, as we keep saying. Really need to work on that some more.

astorije:
I agree with @maxpoulin64. It was discussed in #439 to maybe make them case-insensitive, but a regex, and by default/the only possibility on top of it, I'm really not in favor.
Not every IRC user is advanced enough to know the Regex syntax :-)

maxpoulin64:
@nightfly19: Would you mind making this optional by either putting a checkbox to turn regex matching on/off or use my suggestion of checking if the entered text is in fact, a regex? I think regexes have their use and would probably even use it myself, but we unfortunately just can't have it by default for the reasons discussed above.

astorije:
@nightfly19, closing this as spring cleaning since you didn't get back to our concerns. Feel free to stop by the IRC channel (`#thelounge`) and/or discuss on the issue (#439) if you want to give this another shot and open a new PR.

nightfly19:
Sorry, I only check Github occasionally. I'll do what was asked of my and make another request again soon.

dgw:
I see no reason this PR couldn't be reopened, though @nightfly19 did open it pointing directly at `master` instead of a feature branch.

@astorije Would you prefer to keep the discussion here, or start a new one?

astorije:
@dgw let's open a new one so @nightfly19 can easily start from an up-to-date master, and point to this PR to link to discussion + previous version of the code. Rebasing this would get rid of the current code and discussion would be harder to follow.
We don't pay GitHub per PR so I don't mind opening a new one :D

PR #535, opened 2016-07-31
Fix webirc and 4-in-6 addresses

maxpoulin64:
Since we fixed IPv6, now node returns us IPv4 addresses in the 6-in-4 format such as `::ffff:8.8.8.8`. Not only does this conflict with servers because it starts with `:`, it is also [not a valid argument](https://kiwiirc.com/docs/webirc) for the webirc command.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -70,11 +70,15 @@ module.exports = function() {
 };
 
 function getClientIp(req) {
+	var ip;
+
 	if (!Helper.config.reverseProxy) {
-		return req.connection.remoteAddress;
+		ip = req.connection.remoteAddress;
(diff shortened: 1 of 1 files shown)


xPaw:
Is it guaranteed to be without any zeroes, just the ::ffff?

maxpoulin64:
Yes, it will always be `::ffff:xx.xx.xx.xx`. This is the format 4-in-6 addresses are supposed to be. I'll have to fix oidentd in another PR, two of the lines in it were supposed to be commented, so I'll fix both in one oidentd PR.

astorije:
👍 and merging

PR #537, opened 2016-07-31
Fix accidentally removed border-radius

astorije:
This was removed by [this](https://github.com/thelounge/lounge/pull/379/files#diff-97db1f70168fb5f12457b238ff6052b5L504). I don't think this as further consequences but I would prefer @maxpoulin64 to confirm that.

#### Screenshots

This is very minor, but there is idiom about the devil and details :-)

Before fix | After fix
--- | ---
screen shot 2016-07-31 at 17 50 14 | screen shot 2016-07-31 at 17 49 40
screen shot 2016-07-31 at 17 50 43 | screen shot 2016-07-31 at 17 49 56

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -591,6 +591,7 @@ button {
 	border-radius: 2px;
 	bottom: 4px;
 	left: 220px;
+	overflow: hidden; /* Without this, border-radius has no effect */
 	position: absolute;
 	right: 5px;
 	top: 4px;


maxpoulin64:
Thinking back of it, I probably removed it accidentally while testing some other things. I moved things around a lot before settling with that solution, so I probably forgot to undo that one. Normally I clean up my commits to remove all unwanted changes, but I guess that one went through.

PR #539, opened 2016-08-02
Restore padding and height of message input pre-textarea era

astorije:
Margins around the nick, and overall height of the input, perfectly set prior to the textarea introduction. Turns out, there was also a 1px-offset in there, and I could see something was off without being able to pinpoint what exactly.

Not sure how long the current style of textarea will stay around, but this PR at least puts things back in order until then.

I have played a looot with this, adding and removing stuff, when in the end the changes are **just** because now the border is set on a different element, and we all know how CSS is a bit bitchy about how overall heights and margins are set depending on the border.

#### Results

On pre.4 (without textarea) | Before PR | After PR
--- | --- | ---
screen shot 2016-08-02 at 00 29 43 | screen shot 2016-08-02 at 00 30 09 | screen shot 2016-08-02 at 00 30 25

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -548,10 +548,10 @@ button {
 	background: rgba(0, 0, 0, .06);
 	border-radius: 2px;
 	bottom: 4px;
-	height: 48px;
+	height: 45px;
 	left: 5px;
 	font-size: 14px;
-	line-height: 48px;
+	line-height: 45px;
 	position: absolute;
 	text-align: center;
 	width: 210px;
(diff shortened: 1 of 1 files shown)


maxpoulin64:
Yeah I knew it felt a bit off but I couldn't really figure out what was wrong exactly. The total height was the same and it aligned fine in morning/zenburn so I didn't know what it was. Nice find!

astorije:
Oh now that you say it, I forgot to check how other themes were handling
this :-/
I should check that ASAP...

astorije:
Forgot to post back here: I tested with themes and everything looks fine. It may be that the nick appears a pixel down in some of the themes, but apart from that (not even certain), we're good.

PR #541, opened 2016-08-04
Prevent Ctrl-Tab from triggering tab completion

### LINKED ISSUE #529: Tab completion happens on Firefox when switching tab with keyboard
On Firefox, when using <kbd>Ctrl</kbd>+<kbd>Tab</kbd> to switch tabs, the current word will be completed (this happens when tabbing *out* of The Lounge, not *in*).

See it in action:

bug-switch-tab-firefox

Note that this is a regression since the latest pre-release (2.0.0-pre.4) so marking as `priority`. Also, Chrome is not affected by this.

hho:
Fixes #529

This is probably the simplest workaround for Firefox sending a keydown event when switching browser tabs via Ctrl-Tab

### DIFF
--- a/client/js/libs/jquery/tabcomplete.js
+++ b/client/js/libs/jquery/tabcomplete.js
@@ -112,7 +112,7 @@
 		
 		this.on("keydown.tabcomplete", function(e) {
 			var key = e.which;
-			if (key == keys.tab
+			if (key == keys.tab && !e.ctrlKey
 				|| (options.arrowKeys && (key == keys.up || key == keys.down))) {
 				
 				// Don't lose focus on tab click.


maxpoulin64:
Yeah, this is a pretty cheap fix, but we don't have a key bind on that so that's 👍 for now until we can get a better library I guess.

I see that `preventDefault` was used, I wonder why that doesn't work with the new textarea...

astorije:
Before validating this, I would like to understand why this was working just fine on an `input` element but not a `textarea`. It may be browser weirdness because it's fine on Chrome (what about IE/Edge and Safari?), but it may also be our code that doesn't work well with `textarea`s...

@hho, did you investigate a little bit there by any chance?

hho:
@astorije 2.0.0-pre.4 doesn't even have a `<textarea>` yet, does it? It seems to be only a [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=579254), and an old one at that. I actually don't know how this ever worked in FF…  `¯\_(ツ)_/¯`

astorije:
Indeed, and that's why I was willing to investigate why it works fine with an input but not with a textarea.
But if you found that it's a bug specifically on textareas...

hho:
I'm sorry, but I don't understand:
2.0.0-pre.4 uses `<input>` and it *doesn't* work correctly with Firefox (this PR fixes that).
To me, this doesn't seem to have *anything* to do with textareas at all?

astorije:
Woosh, my bad! When I tested, I could trigger this bug on `<textarea>` elements but not `<input>` elements, which is why I had marked this as a regression, but I guess it's not.
Thanks for sticking with my slowness 😅

astorije:
👍 and merging, but let's open this to https://github.com/erming/tabcomplete to avoid monkey-patching libraries.

Issue #459, opened 2016-07-04
Add --debug CLI param and only listen to irc-fw's debug event when that is passed


xPaw:
We don't really want to spam connection stuff in release.

williamboman:
And possibly `process.env.NODE_ENV === 'development'`?

astorije:
Marking this as completed by #547.

PR #546, opened 2016-08-06
Disable debug logs in npm releases

astorije:
This is temporary until https://github.com/thelounge/lounge/issues/459, but meanwhile we should not be spamming connection debug messages in releases or pre-releases.

### DIFF
--- a/src/plugins/irc-events/connection.js
+++ b/src/plugins/irc-events/connection.js
@@ -74,9 +74,10 @@ module.exports = function(irc, network) {
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


maxpoulin64:
Why not make it a config and just use `if (Helper.config.debug) {` ?

astorije:
Because I don't have time for this today, #459 shows different options and we might want to spend some time/discussions to do what's best, and I would not like to delay pre.5 any longer.

Definitely happy to see this happening in 2.0.0 stable, but pre.5 does not need this I believe. Your call.

PR #549, opened 2016-08-07
Fixes submit button focus

### LINKED ISSUE #481: Message input defocuses with send button

metsjeesus:
Fixes #481

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -672,6 +672,8 @@ $(function() {
 			target: chat.data("id"),
 			text: text
 		});
+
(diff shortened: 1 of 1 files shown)


astorije:
Hey @metsjeesus, have you seen discussion at https://github.com/thelounge/lounge/pull/483? It's not the exact same line so I don't want to close it as a duplicate right away, but it seems to me that any consequences affecting #483 will affect this PR.

PR #552, opened 2016-08-10
Use our logger instead of console.{log,error} for LDAP logs

astorije:
Ping @thisisdarshan and @lindskogen, could you let us know if the LDAP part looks fine to you?
Note that when running a public instance *and* LDAP, the LDAP part is now ignored and a warning is shown instead of crashing the app. @maxpoulin64, does that seem sensible to you?

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -28,7 +28,7 @@ module.exports = function() {
 	var server = null;
 
 	if (config.public && (config.ldap || {}).enable) {
-		throw "Server is public and set to use LDAP. Please disable public if trying to use LDAP authentication.";
+		log.warn("Server is public and set to use LDAP. Set to private mode if trying to use LDAP authentication.");
 	}
 
 	if (!config.https.enable) {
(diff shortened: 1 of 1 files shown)


maxpoulin64:
Actually we left those as `console.log` on purpose since they are not logging but command-line outputs. It's a bit like if `ls` added `[INFO]` on every line, a bit odd... :-1:

astorije:
> Actually we left those as console.log on purpose

I couldn't find discussion about this on GitHub, maybe it was IRC. Should we get rid of https://github.com/thelounge/lounge/pull/552/files#diff-c634a78f476381115a1c873dc965554dR10 then?

Also, this is not the important part of the PR lol, what about the other changes?

maxpoulin64:
Ah yes, remove that and the other part is definitely :+1: especially the throw. Confused the two PRs for a moment 😅

astorije:
@maxpoulin64, removed the changes in `list`, we can take this later anyway, I think the LDAP stuff is more important and valid on its own. Your 👎 hurts my feelings 😂

maxpoulin64:
Actually I -1'd to prevent a merge because I can't add labels on mobile 😶

astorije:
I'm not too fond of crashing the app but that's definitely one option (the one currently, which would change by getting rid of the `throw`).

A third option is to assume mode is private when `ldap.enable` is `true`.

Thoughts?

maxpoulin64:
Okay I'm just bad at communicating tonight. Meant, the other stuff is good changes, including the removal of the throw 😅

PR #553, opened 2016-08-10
Fix token persistency across server refresh

astorije:
This fixes a regression introduced by LDAP support addition (https://github.com/thelounge/lounge/pull/477), which forces users to re-login when the server restarts. This was originally implemented in https://github.com/thelounge/lounge/pull/370.

@thisisdarshan and @lindskogen, could you proof read this PR, please? It seems to me that this was just a typo on #477 but maybe I'm wrong and this is breaking your LDAP implementation, so I value your opinion :-)


lindskogen:
a bit late, but LGTM :)

PR #377, opened 2016-06-04
Fixed font-family redundancies on style.css

wesleyParriott:
I changed all `font: Xpx lato, sans-serif` to `font-size: Xpx`

Issue #81, opened 2016-02-20
Can we remove fonts used solely for themes


xPaw:
I would like to keep the codebase smaller but not having extra font files.

Crypto uses Inconsolata-g, Morning and Zenburn use Open Sans. The default font, which is Lato, is not bad by any means and is a good default for all these themes.

This should simplify things for #80

astorije:
Very soon themes should be self-contained and contain their own fonts. I would hold off with removing them until then. If you remove them anyway, how will themes be able to use the fonts their authors have set?

xPaw:
> how will themes be able to use the fonts their authors have set?

Just use Lato in Morning and Zenburn, it's very close looking to Open Sans.

Monospace fonts can be used in Crypto.

astorije:
While I agree with you that we cannot reasonably host all fonts for all themes, I think they should stay right now. Hopefully very soon we'll provide better theme management and that will resolve this.
Changing that means that people using them will have their UI changed for the only sake of reducing the size of the repo.

xPaw:
@thelounge/maintainers, would like to hear your opinions.

YaManicKill:
I think we should leave this until we get some sort of better theme management. I know that probably won't be anytime soon, but I don't see the necessity in cutting it down.

astorije:
> I think we should leave this until we get some sort of better theme management. I know that probably won't be anytime soon, but I don't see the necessity in cutting it down.

Agreed.

One first step I unsuccessfully attempted was to make themes independent from the base theme. I think that's the first requirement for better theme management. As @xPaw mentioned the other day on IRC, it's fairly difficult to create a theme because you have to override all defaults, and each new release is a potential theme breaker. By making each CSS independent of the default theme, we would reduce the risk (of course, changes to the markup has the same consequences, but that's a next step :-) ).

maxpoulin64:
> As @xPaw mentioned the other day on IRC, it's fairly difficult to create a theme because you have to override all defaults, and each new release is a potential theme breaker.

An idea for that could be to split the default theme into two pieces: one very basic stylesheet that positions the things into the right place, and the default theme to add the colors, margins, alignment and all. That way themes don't have to start from scratch, but don't have to fight with the default theme either.

As for the fonts, maybe themes could load them from Google Fonts? One `@import` and it's done.

astorije:
[...]

PR #562, opened 2016-08-14
Remove font family redundancy, fix missed fonts, remove Open Sans

### LINKED ISSUE #138: Fix font declarations to inherit from body
This will make it easier to change fonts by applying `font-family` on `body`.

There are only a couple of declarations that specifically specify `Lato`, should be trivial to get rid of these, so it automatically inherits.

astorije:
This fixes #138 and #81.

Initially, I was just trying to rebase #377 with master. I noticed a lot of issues with it and ended up completely redoing it.

A few things:

- There should be no visual changes on the default style, except the title + topic in the headers which are a bit more aligned.
- This fixes a few bugs, such as a use of `Leto` instead of `Lato`, missed fonts in some of the themes.
- Crypto will require a bit more fixing to be fully non-redundant, but this is for another PR.
- Zenburn and Morning now only use Lato instead of Lato + Open Sans, as these 2 are very similar. Users of these themes will notice a difference, but I'm ready to bet they did not realize their theme was actually using 2 fonts.
- This was discussed and agreed many times, but working on this PR was a massive pain in the butt because of that overriding situation. My next step once this gets merged is to move the `font-family` declarations of the default theme to `example.css` (which we should rename to give it a real name, maybe even just `Default`) and fix the themes accordingly. Looking at existing issues, we already agreed to that, but pinging @thelounge/maintainers here just so they know.


astorije:
Any takers? I won't have fun fixing potential conflicts on this one :D

YaManicKill:
Yeah, looks fine to me. :+1:

PR #565, opened 2016-08-16
Make custom highlights case-insensitive

### LINKED ISSUE #439: Make custom highlights case-insensitive

astorije:
Fixes #439.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -204,7 +204,7 @@ $(function() {
 		var template = "msg";
 
 		if (!data.msg.highlight && !data.msg.self && (type === "message" || type === "notice") && highlights.some(function(h) {
-			return data.msg.text.indexOf(h) > -1;
+			return data.msg.text.toLocaleLowerCase().indexOf(h.toLocaleLowerCase()) > -1;
 		})) {
 			data.msg.highlight = true;
 		}


client/index.html
```diff
@@ -287,8 +287,8 @@ <h1 class="title">Settings</h1>
 
 							<div class="col-sm-12">
 								<label class="opt">
-									<label for="highlights" class="sr-only">Custom highlights (comma-separated keywords)</label>
-									<input type="text" id="highlights" name="highlights" class="input" placeholder="Custom highlights (comma-separated keywords)">
+									<label for="highlights" class="sr-only">Custom highlights (comma-separated keywords, case-insensitive)</label>
```
YaManicKill: Not convinced about adding this to the label/placeholder, tbh. I think this would be expected as the default.

I guess the other option would be to put a checkbox for case sensitivity. But either way, I wouldn't put this in the label/placeholder.

YaManicKill:
:+1: despite my 1 comment. Not massively against the extra characters, but I don't think necessary

astorije:
@YaManicKill, I updated my PR to address your comment. We can improve this over time, no big deal. Fixing the underlying issue is more pressing anyway.

YaManicKill:
I am still :+1:

astorije:
Cool! @maxpoulin64?

PR #567, opened 2016-08-17
Display unread marker on current channel when initially loading the page

### LINKED ISSUE #561: Unread marker does not show on current channel when initially loading the page
To reproduce:

- Go to a channel
- Close the browser window
- Make sure you receive a message on that channel (for example, by running a different instance with a different nick, etc., like production instance + development instance)
- Reopen the closed instance: the unread marker does not show up above that message

astorije:
Fixes #561.

I did a lot of testing, when focusing/not focusing the app, when receiving a message in active/non-active channel, when loading the page / when the page is already loaded, etc., but this is rather intricate, so I wouldn't be against any additional testing/feedback!

(Can't wait to completely get rid of `activeChannel` on the server, seriously, why does the server have to know what the client is currently viewing...)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -333,7 +333,8 @@ $(function() {
 	socket.on("msg", function(data) {
 		var msg = buildChatMessage(data);
 		var target = "#chan-" + data.chan;
-		var container = chat.find(target + " .messages");
+		var channel = chat.find(target);
+		var container = channel.find(".messages");
 
 		container
 			.append(msg)
@@ -342,7 +343,9 @@ $(function() {
 				data.msg
 			]);
 
-		if (data.msg.self) {
+		// Unread marker is "hidden" (moved to bottom) if message comes from own
+		// user or if the channel is both active and focused
+		if (data.msg.self || (channel.hasClass("active") && document.hasFocus())) {
 			container
 				.find(".unread-marker")
 				.appendTo(container);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -349,6 +349,12 @@ $(function() {
 			container
 				.find(".unread-marker")
 				.appendTo(container);
+
+			// Channel is marked as read on the server
+			socket.emit(
+				"open",
+				data.chan
+			);
 		}
 	});
 
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -43,7 +43,7 @@ Chan.prototype.pushMessage = function(client, msg) {
 		this.messages.splice(0, this.messages.length - Helper.config.maxHistory);
 	}
 
-	if (!msg.self && this.id !== client.activeChannel) {
+	if (!msg.self) {
 		if (!this.firstUnread) {
 			this.firstUnread = msg.id;
 		}


xPaw:
> seriously, why does the server have to know what the client is currently viewing

To not count unread messages, for one. But the `activeChannel` should be set per open connection, not per client once.

client/js/lounge.js
```diff
@@ -342,10 +343,18 @@ $(function() {
...
-		if (data.msg.self) {
+		// Unread marker is "hidden" (moved to bottom) if message comes from own
+		// user or if the channel is both active and focused
+		if (data.msg.self || (channel.hasClass("active") && document.hasFocus())) {
 			container
 				.find(".unread-marker")
 				.appendTo(container);
+
+			// Channel is marked as read on the server
+			socket.emit(
```
xPaw: Kind of a terrible fix. The server should know this on its own without clients sending `open` on every received message.

astorije:
> But the activeChannel should be set per open connection, not per client once.

Fair point. I'm trying to think of other consequences or reasons why not but I don't have one in mind at the moment.
To me it makes sense to keep that on the front-end only, but per connection seems analogous.

xPaw:
> To me it makes sense to keep that on the front-end only

Have to be on the server to be consistent across page reloads.

xPaw:
Yeah I see the cause of this problem, and it is indeed related to `activeChannel`, I think this has to be fixed correctly once and for all on the server, without changing anything on the client (I don't agree with your document focus change for one).

client/js/lounge.js
astorije: For performance reason? This doesn't strike me as such a terrible thing considering `open` is a small, fixed-sized event.
In terms of behavior, this is natural as a client acknowledging it read a message is more accurate than making this assumption from the server.

Any alternative in mind?

astorije:
> but I don't have one in mind at the moment.

Actually, there is at least one, and I could notice that issue when working on this PR: doing this per open connection does not inform the server if the window is in focus or not. That is, unless server keeps track of in/out-of-focus information, which can also happen a lot and might give a sluggish feel on the user end.
If not taking care of that, reloading the page shows the unread marker at an older position as there is a mismatch between browser and server tracking at that stage.

xPaw:
> That is, unless server keeps track of in/out-of-focus information

Client sends `open` on channel change, and the server just needs to set `activeChannel` for that active socket.

> If not taking care of that, reloading the page shows the unread marker at an older position as there is a mismatch between browser and server tracking at that stage.

In the behaviour I mention above, the unread counter would reset after latest channel open (independent of clients, don't think we can anyhow do this better).

astorije:
> (I don't agree with your document focus change for one).

Well, that's bug-driven, not opinion-driven :D

Also, you want to keep the unread counter hidden if you are actually using the app (focused), but you want the unread counter to build up on the current channel when not in focus.

xPaw:
> but you want the unread counter to build up on the current channel when not in focus.

Sure, but your change makes the unread marker always move to the bottom if looking at the channel when a message is received.

astorije:
> Sure, but your change makes the unread marker always move to the bottom if looking at the channel when a message is received. This will be annoying when switching to a channel and receiving a message right after.

Not sure I'm following. You're not suggesting unread marker should be displayed on the current (active) and browser-focused channel, right? That would be very weird.

xPaw:
> You're not suggesting unread marker should be displayed on the current (active) and browser-focused channel, right? That would be very weird.

Think it through, how else would you see it?

astorije:
> Think it through, how else would you see it?

Well (and that's what I notice on all other messaging system I use), I'd say that if user has a specific channel active **and** the app is focused, then user is reading the messages, so no *unread* message marker should be displayed.

xPaw:
> if user has a specific channel active and the app is focused, then user is reading the messages, so no unread message marker should be displayed.

The real problem is that as soon as a message is received on the channel, the unread marker will disappear (if it was 5 lines above), which is not exactly reasonable and consistent.

astorije:
I see now, another fair point. Then there needs to be another condition to that use case: if unread market is not at bottom already (hidden), no actions taken. Makes sense?

astorije:
@xPaw, I know you're mostly away, but if you have any idea on how to improve/fix this, I'd be happy to review if you are able to push something, or give a shot at implementing myself if you can give me some pointers.
It's not critical so I don't think it will/should block major release, but it's fairly annoying so I'd love to fix.

One thing I'd like to achieve is that the server should not care about the active channel, so that a given user can browse channels on his devices independently. If you think this would make the current fix simpler, then we can combine, but we don't have to.

Anyway, any input welcome :-)

xPaw:
I think the current simple fix would be to add a check for count of connected clients in unread message counter on the server. This would keep all the current logic and not introduce new problems.

astorije:
Ugh I am so frustrated by this right now. I give up.
So, 2 things to deal with that I couldn't figure out in a simple way:

- We need to be able to count connected clients. I couldn't find places where we do that already. `socket.io` has ways to count open connections globally, but we need it per user. I was hoping we were namespacing open connections per user but couldn't make that work.
- For some reason, when opening a page, the active channel is being looked up in the sidebar and clicked twice. That makes the unread marker disappear even if I remove all checks around `chan.unread++;` in `irc-events/message.js`.

Obviously, help more than welcome! I need to focus on the remaining v2.0.0 issues :(

nornagon:
The client code only moves the unread marker when it switches _away_ from a channel with unread messages (i.e., 'unread' means 'while you weren't looking'). Could the client perhaps communicate that event to the server? "I've stopped looking at this channel now, mark the latest message as the last one I read".

I could imagine that getting complicated with multiple clients, though. The server could maintain a list of active channels per user (one per connection), and whenever a client changes active channel, that channel is removed from the list. If it's the last window looking at that channel, the server marks that channel as "unobserved" and the next message becomes the first unread.

I'll have a shot at making a PR that implements that approach.

astorije:
Closing this in favor of #613.

PR #570, opened 2016-08-18
Make sure users with wrong tokens are locked out instead of crashing the app

### LINKED ISSUE #569: Users created/logged in with 1.5.0 cannot access the app after upgrading to 2.0.0-pre*
Steps to reproduce:

- On The Lounge v1.5.0, create a user
- Start the server, login and close the server
- Install the latest version (v2.0.0-pre6 right now)
- Start the server and notice the following error in the server console:

```js
Missing error handler on `socket`.
TypeError: Cannot read property 'token' of undefined
    at Socket.auth (/Users/jeremie/Perso/lounge/src/server.js:289:60)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at Socket.onevent (/Users/jeremie/Perso/lounge/node_modules/socket.io/lib/socket.js:335:8)
    at Socket.onpacket (/Users/jeremie/Perso/lounge/node_modules/socket.io/lib/socket.js:295:12)
    at Client.ondecoded (/Users/jeremie/Perso/lounge/node_modules/socket.io/lib/client.js:193:14)
    at Decoder.Emitter.emit (/Users/jeremie/Perso/lounge/node_modules/socket.io/node_modules/socket.io-parser/node_modules/component-emitter/index.js:134:20)
    at Decoder.add (/Users/jeremie/Perso/lounge/node_modules/socket.io/node_modules/socket.io-parser/index.js:247:12)
    at Client.ondata (/Users/jeremie/Perso/lounge/node_modules/socket.io/lib/client.js:175:18)
    at emitOne (events.js:77:13)
```

astorije:
Fixes #569, see issue for more details.
That issue only appears when a wrong token stored in `localStorage`, so users could login fine in private browsing for example.

@PugaBear let me know if this fixes your issue.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -286,10 +286,10 @@ function auth(data) {
 		}
 	} else {
 		client = manager.findClient(data.user, data.token);
-		var signedIn = data.token && data.token === client.config.token;
+		var signedIn = data.token && client && data.token === client.config.token;
 		var token;
 
-		if (data.remember || data.token) {
+		if (client && (data.remember || data.token)) {
 			token = client.config.token;
 		}
 


jedahan:
I had this same issue, and just tested this out, and it looks like it is working. I'll let it bake for a day or so and see if it stays stable.

src/server.js
```diff
@@ -286,10 +286,10 @@ function auth(data) {
 		}
 	} else {
 		client = manager.findClient(data.user, data.token);
-		var signedIn = data.token && data.token === client.config.token;
+		var signedIn = data.token && client && data.token === client.config.token;
```
xPaw: Seems like an odd fix. How does this entire `else` branch work if `client` is not found?

astorije: If you follow the rest of the code path in this case, when `findClient` returns false it treats the user as failing his authentication.
There is one place in `authFunction` (called by `authFunction(client, data.user, data.password, authCallback);`) which does make use of `client`, but it's wrapped around a `try`/`catch` so it's "safe".

I recognize that this is the cheapest solution to make this work though, do you have any suggestion to improve this?

xPaw: You actually removed this `client` check here: https://github.com/thelounge/lounge/commit/cf64cb04c493041756f1f3639f5b8d35dff21b27#diff-e6a5b42b2f7a26c840607370aed5301a

astorije: Whaaaat, why did I do that?!

Either way, that other change could have happened as well, I think...

PR #572, opened 2016-08-19
Move border-radius from #main to .window elements to fix radius once and for all

astorije:
https://github.com/thelounge/lounge/pull/537 only fixed it on chat windows,
but for some browser-specific display bug/reason not on settings, login, ...

No screenshot here, it's literally the same bug than #537 but applied to all windows.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -587,11 +587,8 @@ button {
 }
 
 #main {
-	background: #fff;
-	border-radius: 2px;
 	bottom: 4px;
 	left: 220px;
-	overflow: hidden; /* Without this, border-radius has no effect */
 	position: absolute;
 	right: 5px;
 	top: 4px;
@@ -651,6 +648,7 @@ button {
 
 #windows .window {
 	background: #fff;
+	border-radius: 2px;
 	bottom: 0;
 	display: none;
 	left: 0;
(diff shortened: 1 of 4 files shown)


client/themes/zenburn.css
```diff
@@ -97,12 +97,16 @@ body {
 	left: 0;
 	bottom: 0;
 	width: 220px;
+	border-radius: 0;
```
astorije: All these theme overrides should actually be removed and design specific stuff moved from `style.css` to `example.css`. That would make for more stable themes, lighter, etc. But oh well, we already agreed on that :D
I don't want to do that until #568 makes it into master, because doing so will make the `@import` trick in custom CSS not valid anymore, so users couldn't reliably change their themes at all.

xPaw:
#359 and no need to fix it ;)

client/css/style.css
```diff
@@ -587,11 +587,8 @@ button {
 }
 
 #main {
-	background: #fff;
```
YaManicKill: Why is this removed?

astorije: Because each `window` is taking care of its own `background-color`. And as far as I remember, it's necessary to remove this because of the radius being moved to `.window` instead of `#main` (letting us get rid of that ugly, half-working `overflow: hidden`.

Does that make more sense now?

astorije: In short, this is always shadowed by [https://github.com/thelounge/lounge/pull/572/files#diff-97db1f70168fb5f12457b238ff6052b5R650](https://github.com/thelounge/lounge/pull/572/files#diff-97db1f70168fb5f12457b238ff6052b5R650).

PR #573, opened 2016-08-19
Send user to closest window when closing active one instead of network window

astorije:
This is an alternative to #503.
It corresponds to option 2 described in [this comment](https://github.com/thelounge/lounge/pull/503#issuecomment-239996423).

When closing the active channel/query, user is currently (on master) sent back to the network window. This was done in #489 in order to quickly fix a bug, with intent to improve later.

While #503 restores previous behavior (on latest stable), this PR simply sends the user to closest channel/query to reduce jump to its minimum.
#503 is technically the most accurate thing to do as it will send you back where you were N channels ago if you close N channels (assuming you stay focused on channels under the same network, otherwise it will not work as expected), but it is actually not efficient in terms of usability and can still make the user jump places in long lists.

This is for example how Chrome roughly handles closing the current tab.

### Results

Before | After
--- | ---
closing-before | closing-after

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -398,9 +398,16 @@ $(function() {
 	socket.on("part", function(data) {
 		var chanMenuItem = sidebar.find(".chan[data-id='" + data.chan + "']");
 
-		// When parting from the active channel/query, jump to the network's lobby
+		// When parting from the active channel/query, jump to the next window if
+		// there is one, or the last one otherwise
 		if (chanMenuItem.hasClass("active")) {
-			chanMenuItem.parent(".network").find(".lobby").click();
+			var newActive = chanMenuItem.next(".chan");
+
+			if (newActive.length === 0) {
+				newActive = chanMenuItem.prev(".chan");
+			}
+
+			newActive.click();
 		}
 
 		chanMenuItem.remove();


YaManicKill:
I'm not sure how I feel about this. I'm not saying no just now, I just have 1 concern with it.

> This is for example how Chrome roughly handles closing the current tab.

While this is a legitimate point, in the lounge we are dealing with channels with unread stuff in them. If we move to a channel that has unread items in it, then the user has to read them then, or they won't have a way to know in the future what isn't read. Thoughts?

astorije:
> While this is a legitimate point, in the lounge we are dealing with channels with unread stuff in them. If we move to a channel that has unread items in it, then the user has to read them then, or they won't have a way to know in the future what isn't read.

Very valid point, also shared by #503. Note that it also applies to stable (obviously) and to current master if you have unread messages on the network window.

> Thoughts?

I'd say if all options share the same drawback (granted, that issue is *probably* less a big of a deal on server window), I would go for the otherwise most usable solution, i.e. the smallest jump, which conveniently happens to be... my own PR lol.

YaManicKill:
>  (granted, that issue is probably less a big of a deal on server window)

This is exactly how I feel about it. I think that the least issue is the server window, so I would vote for keeping what we currently have.

I wouldn't be strongly against this PR, though. other @thelounge/maintainers, thoughts? If the others agree with this PR, I'll be fine with it.

astorije:
> I wouldn't be strongly against this PR, though. other @thelounge/maintainers, thoughts? If the others agree with this PR, I'll be fine with it.

Well here is the issue. You're in favor of current solution, @maxpoulin64 is in favor of his own PR (#503), and I'm in favor of my own PR (this one) :D

Not sure what to do. Sleep on these?

As another example, Slack is moving all over the place when you close a channel/query, trying to remember what was the last one... The jumps are awful, I can hear members of my team cursing when they try to clean up their channel list lol.

YaManicKill:
Another option would be that we could go to a blank page that doesn't exist. 

Or we have a setting to change how this works, with the 3 choices, kinda like firefox (but of course, we then have to decide on the default).

maxpoulin64:
Well I guess that's a good case of "lets see what the users think".

Personally I never ran into the Slack issue: I use right click to close them, so the previous one is pretty much always the right one. We also didn't have this issue with the hovering close button 😛

astorije:
> Another option would be that we could go to a blank page that doesn't exist.

That, yes! Telegram does exactly that when you launch the app, it makes no assumption on when you were last time. It's also nice to be able to "stop looking" at a channel by pressing Escape.
But I think this is a bit further down the road, it has other consequences and is slightly more involved than the current solution or #503 or this PR.

> Or we have a setting to change how this works, with the 3 choices, kinda like firefox (but of course, we then have to decide on the default).

Yeah, as long as there can be an admin default + a user default + a ... 😂
Joke aside, I think that would make for a fairly awkward setting. Remember that this is only 1) when closing a channel or a query and 2) that channel or query was active when it was closed. Kind of too niche to have an option IMO. Also, good luck us to find a descriptive setting label that's not the length of a blog post lol.

astorije:
> I use right click to close them, so the previous one is pretty much always the right one.

Wait, I may have misunderstood you then. Even with master or this current PR, if you right click a non-active channel/query, you stay where you are, so that still works fine. Am I not following you?

YaManicKill:
> That, yes! Telegram does exactly that when you launch the app, it makes no assumption on when you were last time. It's also nice to be able to "stop looking" at a channel by pressing Escape.
But I think this is a bit further down the road, it has other consequences and is slightly more involved than the current solution or #503 or this PR.

If this is the ideal solution, then let's not do this or #503 (as they are just a "stop-gap" with no massive improvement over the current situation). I say we stick that as an issue in the backlog, keep the current solution, and work on it when we have the time after 2.0.0.

maxpoulin64:
Yes, exactly. So the only times I ever close a channel I'm in, it's because I want to go back to the previous channel I was. Otherwise if I'm cleaning channels/PMs or leaving a chan I don't care anymore, I don't even bother to open it

YaManicKill:
@astorije Any thoughts on my proposed solution?

astorije:
@YaManicKill, solution being we close both this PR and #503 until we do better? I'm OK with that. @maxpoulin64?

YaManicKill:
Yeah, that's is my preferred option.

xPaw:
> Yes, exactly. So the only times I ever close a channel I'm in, it's because I want to go back to the previous channel I was. 

I agree, which is why I'm in favour of #503.

PR #575, opened 2016-08-19
client: implement History Web API

### LINKED ISSUE #574: Implement History Web API
I think it would be nice to be able to traverse between views by going back/forward in history. Thoughts?

williamboman:
¯\_(ツ)_/¯ I like it. It's really nice on mobile devices. This PR does not include any URL routing whatsoever.

history

Closes #574.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -155,7 +155,9 @@ $(function() {
 			return;
 		}
 		sidebar.find(".sign-in")
-			.click()
+			.trigger("click", {
+				pushState: false,
+			})
 			.end()
 			.find(".networks")
 			.html("")
@@ -197,7 +199,9 @@ $(function() {
 		$("#loading-page-message").text("Rendering…");
 
 		if (data.networks.length === 0) {
-			$("#footer").find(".connect").trigger("click");
+			$("#footer").find(".connect").trigger("click", {
+				pushState: false,
+			});
 		} else {
 			renderNetworks(data);
 		}
@@ -219,7 +223,9 @@ $(function() {
 				.eq(0)
 				.trigger("click");
 			if (first.length === 0) {
-				$("#footer").find(".connect").trigger("click");
+				$("#footer").find(".connect").trigger("click", {
+					pushState: false,
+				});
 			}
 		}
 	});
@@ -986,6 +992,33 @@ $(function() {
 		});
 	});
 
+	sidebar.on("click", ".chan, button", function(e, data) {
+		// Pushes states to history web API when clicking elements with a data-target attribute.
+		// States are very trivial and only contain a single `clickTarget` property which
+		// contains a CSS selector that targets elements which takes the user to a different view
+		// when clicked. The `popstate` event listener will trigger synthetic click events using that
+		// selector and thus take the user to a different view/state.
+		if (data && data.pushState === false) {
+			return;
+		}
+		const self = $(this);
+		const target = self.data("target");
+		if (!target) {
+			return;
+		}
+		const state = {};
+
+		if (self.hasClass("chan")) {
+			state.clickTarget = `.chan[data-id="${self.data("id")}"]`;
+		} else {
+			state.clickTarget = `#footer button[data-target="${target}"]`;
+		}
+
+		if (history && history.pushState) {
+			history.pushState(state, null, null);
+		}
+	});
+
 	sidebar.on("click", ".chan, button", function() {
 		var self = $(this);
 		var target = self.data("target");
@@ -1603,4 +1636,20 @@ $(function() {
 
 	// Only start opening socket.io connection after all events have been registered
 	socket.open();
+
+	window.addEventListener(
+		"popstate",
+		(e) => {
+			const {state} = e;
+			if (!state) {
+				return;
+			}
+			const {clickTarget} = state;
+			if (clickTarget) {
+				$(clickTarget).trigger("click", {
+					pushState: false
+				});
+			}
+		}
+	);
 });


astorije:
> It's really nice on mobile devices.

I haven't tested this on mobile, how does that behave on mobiles when you are headless (i.e. does not have browser back buttons)?
For example, on Android, does the bottom left button here do the job?



I have the feeling that people will be either all in favor or completely against that feature. I don't have a preference yet, I need to play with it for a while and see how it goes.

The one thing that bothers me is that I had to integrate History API some time ago for an app, and it was a bug sink. Nothing to do with your work obviously, and nothing I will hold against integrating this PR, I'm just thinking of long-term maintenance.

williamboman:
> I haven't tested this on mobile, how does that behave on mobiles when you are headless (i.e. does not have browser back buttons)?

[This is how](https://streamable.com/qby7) navigating by pressing the back button on my S6 looks like (both Chrome in-browser and standalone web app).

maxpoulin64:
I'm not a fan of this at all. This is bound to really fill up the browser history. What's the use case? In which situation it would be preferable to hit the back button instead of just clicking on the channel you want to go? (Or Ctrl+Up/Down to it).

dgw:
Does filling up the history really matter that much? There aren't many cases where a user is likely to need to get back to where they were in a tab before opening The Lounge, as users are overwhelmingly likely to open a new tab specifically for The Lounge due to what sort of app it is.

I know I've been involved in a discussion, popped to another channel to answer a highlight, and then momentarily forgotten what channel I came from. So this would be quite useful for me.

client/js/lounge.js
```diff
@@ -742,13 +757,29 @@ $(function() {
...
 
-	sidebar.on("click", ".chan, button", function() {
+	sidebar.on("click", ".chan, button", function(e, data) {
 		var self = $(this);
 		var target = self.data("target");
 		if (!target) {
 			return;
 		}
 
+		if (!data || data.pushState !== false) {
```
astorije: Wouldn't `if (!data || data.pushState) {` work and be a tiny bit clearer?

williamboman: Yep, but, knowing myself, if I write a boolean expression like this it's usually for a reason. I don't remember what that reason was though.

client/js/lounge.js
```diff
@@ -742,13 +757,29 @@ $(function() {
...
+			var state = {
+				view: {
+					"#connect": VIEWS.CONNECT,
+					"#settings": VIEWS.SETTINGS,
+					"#sign-in": VIEWS.SIGN_IN
+				}[target] || VIEWS.CHANNEL
+			};
+			switch (state.view) {
+			case VIEWS.CHANNEL:
```
astorije: That `VIEWS` object is a level of indirection I'd prefer to avoid.
Could you switch to use strings instead please: `connect`, `settings`, `sign-in`, `channel`?

williamboman: Sure!

astorije:
I like this a lot, 👏 @williamboman! I'd be more comfortable is the code was made a bit simpler though as we don't need the extra complexity. See comment above.

> I'm not a fan of this at all. This is bound to really fill up the browser history. What's the use case? In which situation it would be preferable to hit the back button instead of just clicking on the channel you want to go? (Or Ctrl+Up/Down to it).

@maxpoulin64, I don't think this is meant to answer a use case, but rather provide a native app experience to the user. After @williamboman opened his PR, I realized how much I was relying on the ◀️ button on Android on other apps!
Missing this on The Lounge is one of those things that make it feel like a website with JavaScript and not an actual web app (both on desktop and mobile).

What's the problem with filling up the browser history? I fully agree with @dgw's first paragraph here.

maxpoulin64:
Yeah, I just tested this on mobile, and I'm still really not a fan of it. I would be okay with it on desktop, but this is a wrong use of the back button and this quickly becomes noticeable when you use it for a bit.

The thing is, this is navigation *on the same level*, the user is not going down one screen deeper. While I understand the want to be able to use the back button to go back to the previous channel, it doesn't make sense in a hierarchical navigation point of view. The user should know where pressing back will end them to. This is not the case here, because this makes a very long chain of channels you visited, but they are all on the same level. Therefore you press back and eventually you end up either on another website, or on your home screen because you pressed back once too much.

Back is supposed to move you back a screen, not change tabs in the same view. I'd expect back to bring me back to the chat if I was in the "Connect to network" or "Settings" pages, because in this case this is really a logically different and independent screen. But channels are just tabs. You don't expect changing tabs when pressing back in Chrome, so why would we expect an IRC client to switch tab on back? I think the back button is enough of a clusterfuck on Android already.

I just check with Slack, and Slack doesn't do that either. The only messaging apps I see use the back button is when you have two separate screens: the contact list, and the conversation itself. Those are to separate things, and going back gets you back to the contact list. Hangouts, Skype, Messenger all do this. It makes sense, as this is a separate screen. There is a hierarchical navigation, and I think this is where the confusion comes from. For this to logically work, we'd need to have a "Channel List" screen, and when we click on a channel it opens the channel as another screen. Then it would make sense, because backing out of this would bring you to the channel list. I'm not against this option, but it would behave a bit weirdly on desktop.

williamboman:
> this is a wrong use of the back button

According to? The entire web history API revolves around the fact that you should be able to navigate SPAs by going back/forward in "history".

> Back is supposed to move you back a screen, not change tabs in the same view.

Again, according to?

> I just check with Slack

In my experience, Slack works the same as Lounge does with this PR (in terms of navigating back/forward through channels). Slack also uses URL routing, which means it'll be recorded in the browser history (this PR does not have URL routing, which means Lounge will only have one history entry per session). (I've only tested in Chrome only though)

client/js/lounge.js
williamboman: I remember now. We want to always default to pushing states unless explicitly told not to (through setting `pushState: false` in the data object). However, since the data variable is arbitrary and may contain other data in the future we need an explicit comparison here. E.g.;

```js
lastChannelJoined.trigger("click", {
    otherData: 'very interesting'
});
```

would not push the new state into history.

xPaw:
I tend to agree with Max on this. Perhaps it would make sense to only push window changes to history, eg open settings and press back to get back to main window. Channel switches will indeed spam history. Imagine pressing back button on mobile to get to previous page (or close).

client/js/lounge.js
```diff
@@ -742,13 +749,29 @@ $(function() {
...
-	sidebar.on("click", ".chan, button", function() {
+	sidebar.on("click", ".chan, button", function(e, data) {
 		var self = $(this);
 		var target = self.data("target");
 		if (!target) {
 			return;
 		}
 
+		if (!data || data.pushState !== false) {
+			var state = {
```
xPaw: This could be done with DOM only, not necessary to have an array

YaManicKill:
I'm going to nudge this again. I really like it, and I've missed having it a few times recently ("Oh, which channel was I just on, where I think I saw something interesting" and then spend 5 minutes going through all the chans and finally find it, when the back button would just work). 

> In my experience, Slack works the same as Lounge does with this PR (in terms of navigating back/forward through channels).

Yup, Slack does exactly this, I just looked and the back button has all my previous channels in it. It's really nice.

client/js/lounge.js
```diff
@@ -1344,4 +1367,31 @@ $(function() {
...
+
+	window.addEventListener(
+		"popstate",
+		function(e) {
+			var state = e.state;
+			if (!state) {
+				return;
+			}
+			var target;
+			switch (state.view) {
```
xPaw: There should be a way to get rid of this switch and simplify the logic. Perhaps if channelId is not null then find the channel, otherwise try to find a window (without hard coding known windows).

client/js/lounge.js
xPaw: If target is channel then store channelId, otherwise just the target, no need to hardcode a switch.

client/js/lounge.js
```diff
@@ -276,6 +276,13 @@ $(function() {
 		return msg;
 	}
 
+	function pushHistoryState(state) {
```
xPaw: This is small enough to be inclined, imo. Then you can avoid handling the click code completely if history is not supported.

astorije:
@williamboman, what's the status of this? I notice your second commit is `wip` so I prefer to ask :)
Also, you might want to rebase with `master` to see builds pass.

williamboman:
> @williamboman, what's the status of this? I notice your second commit is wip so I prefer to ask :)
> Also, you might want to rebase with master to see builds pass.

I wanted to keep the initial implementation easily available just in case. I've squashed and rebased the 2 commits now

YaManicKill:
In that case, I'm happy with this.

@xPaw and @maxpoulin64 are you guys still against this? I think it's very valuable, and as mentioned it happens in things like slack, and I really like it in slack and miss it in here. I think it's something we should do.

@williamboman You've got some lint errors.

client/js/lounge.js
```diff
@@ -1440,4 +1456,23 @@ $(function() {
...
+
+	window.addEventListener(
+		"popstate",
+		function(e) {
+			var state = e.state;
+			if (!state) {
+				return;
+			}
+			var target = state.target;
+			if (!target && state.channelId) {
```
xPaw: Isn't channel ID 0 possible?

client/js/lounge.js
```diff
@@ -940,13 +940,29 @@ $(function() {
...
+	sidebar.on("click", ".chan, button", function(e, data) {
 		var self = $(this);
 		var target = self.data("target");
 		if (!target) {
 			return;
 		}
 
+		if (history && history.pushState) {
+			if (!data || data.pushState !== false) {
+				var view = {
```
xPaw: Can we get rid of this hardcoded map?

client/js/lounge.js
```diff
@@ -940,13 +940,30 @@ $(function() {
...
 		var self = $(this);
 		var target = self.data("target");
 		if (!target) {
 			return;
 		}
 
+		if (history && history.pushState) {
+			if (!data || data.pushState !== false) {
+				var state = {}
+				var view = {
```
xPaw: Can you remove the hardcoded object?

xPaw:
I'm for it.

williamboman:
The rewrite I did was really sloppy, I need to go through it properly once.

astorije:
@williamboman, ping? :)

williamboman:
Sorry I totally forgot about this one, gonna fix it today

client/js/lounge.js
```diff
@@ -985,6 +985,33 @@ $(function() {
...
+		if (!target) {
+			return;
+		}
+		var state = {};
+
+		if (self.hasClass("chan")) {
+			state.clickTarget = `.chan[data-id="${self.data("id")}"]`;
+		} else {
+			state.clickTarget = `#footer button[data-target="${target}"]`;
+		}
```
williamboman: This entire `clickTarget` hack feels quite weak. There's no view/state abstraction in place so far which makes this a bit awkward.

astorije:
Damn I forgot this had not shipped yet. @williamboman what do you think is the current state of this?

williamboman:
> Damn I forgot this had not shipped yet. @williamboman what do you think is the current state of this?

The client code had changed since the first implementation so I rewrote it entirely. The current implementation is very ad-hoc and might not scale very well - but I'm not sure that'll be an issue.

astorije:
@williamboman, last check: are you OK with your current implementation or do you want to add further changes before it makes it into master?

williamboman:
Yeah I think this is good to go now

astorije:
Testing this, the loading page gets added as an item in the history. Would it be possible to ignore it?

astorije:
@williamboman, I notice you amended your commit, did that contain a fix for the loading page? If so, how? (sorry it's late, I can't think straight anymore 😅)

williamboman:
> @williamboman, I notice you amended your commit, did that contain a fix for the loading page? If so, how? (sorry it's late, I can't think straight anymore 😅)

Yep

client/js/lounge.js
```diff
@@ -1523,4 +1556,20 @@ $(function() {
...
+				return;
+			}
+			var clickTarget = state.clickTarget;
+			if (clickTarget) {
+				$(clickTarget).trigger("click", {
+					pushState: false
+				});
+			}
+		}
+	);
```
astorije: Arrow function, `let`, etc.?

client/js/lounge.js
astorije: Agreed. Not sure we can do any better at the moment though. It would be fairly easy to break silently though...

client/js/lounge.js
```diff
@@ -217,7 +221,9 @@ $(function() {
 				.eq(0)
 				.trigger("click");
 			if (first.length === 0) {
-				$("#footer").find(".connect").trigger("click");
+				$("#footer").find(".connect").trigger("click", {
+					pushState: false,
+				});
```
astorije: Not too great, but I can't think of anything better...

astorije:
@williamboman, ping?

williamboman:
> View changes
> Testing this after your fix, I'm still seeing something odd: say I'm on page A (not The Lounge), then type the URL of The Lounge (localhost when testing), then I have to hit "Back" twice to go back to A.
> I'm guessing sign in page or loading page is still taking over.

I spent some time trying to fix this but couldn't find what's causing it

xPaw:
The issue described above is quite annoying. It takes two back presses to quit the app. (On mobile)

astorije:
I'd be happy if we receive a fix, but it's so corner case that I wouldn't want to revert just for that. Quitting an app through hitting "Back" is not the most obvious way to do so.

xPaw:
There is no other way to quit it if its pinned on home screen. Pressing home button does not kill lounge, and it runs in background.

YaManicKill:
Recent apps, swipe away.

astorije:
Going back does not even kill the app for me, it puts it in background as well.

PR #576, opened 2016-08-20
Allow long URLs to break onto next line on Chrome

### LINKED ISSUE #550: Long URLs overflow the chat window and create a transparent yet present horizontal scrollbar
Look at that guy, for example:

bug-long-urls

Shot on Chrome (Mac OS). Could not reproduce on Firefox on that same OS.

astorije:
Fixes #550.

This fixes a bug that displays a horizontal scrollbar and messes up with the layout when URLs (or text) is too long.
Fix is Chrome-specific but so is the bug.

It is possible that #523 fixes this along the way, but it looks like it may be creating some other issues. Therefore it needs cautious testing and maybe some tweaking, hence delays, while this PR fixes now this specific bug in the meantime.

### Result

Before | After
--- | ---
screen shot 2016-08-19 at 23 32 50 | screen shot 2016-08-19 at 23 32 15

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -815,6 +815,7 @@ button {
 
 #chat .msg {
 	word-wrap: break-word;
+	word-break: break-word; /* Webkit-specific */
 }
 
 #chat .unread-marker {


xPaw:
:+1:

PR #587, opened 2016-08-31
Do not set app orientation in manifest to use user setting at OS level

### LINKED ISSUE #586: The Lounge does not respect orientation locking on Android
Since #310, orientation handling has changed: after setting The Lounge as a homescreen app from Chrome, The Lounge changes orientation from portrait to landscape mode even if user has locked the portrait mode on the phone (i.e. disabled auto-rotate).

Before that, preference would be honored, i.e. The Lounge would rotate if auto-rotate was enabled, or stay in portrait if auto-rotate was disabled.

astorije:
Fixes #586.

Turns out, when this setting is not set, the auto-rotate being on or off is respected as expected. Who knew!

I know @YaManicKill and @vectr0n have been testing on different devices, would be great if you guys could confirm :-)
Maybe an iOS user should confirm that things are fine on this side as well.

(I have also tried with `natural`, but that doesn't work for us: it means portrait for phones and landscape for tablets).

### DIFF
--- a/client/manifest.json
+++ b/client/manifest.json
@@ -3,7 +3,6 @@
 	"short_name": "The Lounge",
 	"description": "Self-hosted web IRC client",
 	"display": "standalone",
-	"orientation": "any",
 	"theme_color": "#455164",
 	"background_color": "#455164",
 	"icons":


YaManicKill:
Well, that's a confusing setting. I would have assumed that "orientation: any" meant that it **can** be any, but not that it would force it into being any. Will test this later today.

astorije:
I know!! That what kind of a desperate move to be honest, I was not expecting that to work, but yet it did :sweat_smile:

YaManicKill:
Yeah, this looks fine.

From what I can see, it looks like the default is "natural" for Chrome, which I think respects the device's orientation lock.

It might be worth thinking about specifically putting this in incase any other browsers use a different default, but I'm happy either way.

:+1: but I'll let you merge.

astorije:
> From what I can see, it looks like the default is "natural" for Chrome, which I think respects the device's orientation lock.

Actually, default is supposed to be `any` (but in practice, it's not all the time, as we are seeing here). `natural` means that default orientation for a given device is being used, but when I tested, it didn't respect the orientation lock.
Thing is a mess, but to be fair the whole specification for this [is still a draft](https://www.w3.org/TR/appmanifest/), so I'm guessing changes will happen. That's fine, we'll adjust when browsers do :-)

PR #592, opened 2016-09-05
consolidate version numbers throughout all interfaces

williamboman:
I find the current reporting of versions, both in the client and CLI, is a bit lackluster. For example, starting the server on latest `master` logs a message that (as of now) says `The Lounge v2.0.0-pre.6 is now running ...` when in fact `master` is 30 commits ahead of pre-6.

This PR consolidates version numbers reported in the client and the CLI. Effectively, this means there'll only be visible changes when starting lounge through the CLI;

screenshot 2016-09-07 at 16 04 32
screenshot 2016-09-07 at 16 04 38

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -3,13 +3,12 @@
 global.log = require("../log.js");
 
 var program = require("commander");
-var pkg = require("../../package.json");
 var fs = require("fs");
 var fsextra = require("fs-extra");
 var path = require("path");
 var Helper = require("../helper");
 
-program.version(pkg.version, "-v, --version");
+program.version(Helper.getVersion(), "-v, --version");
 program.option("");
 program.option("    --home <path>" , "home path");
 
--- a/src/helper.js
+++ b/src/helper.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const pkg = require("../package.json");
 var _ = require("lodash");
 var path = require("path");
 var os = require("os");
@@ -11,6 +12,8 @@ var Helper = {
 	getUserConfigPath: getUserConfigPath,
 	getUserLogsPath: getUserLogsPath,
 	setHome: setHome,
+	getVersion: getVersion,
+	getGitCommit: getGitCommit,
 };
 
 module.exports = Helper;
@@ -21,6 +24,29 @@ Helper.config = require(path.resolve(path.join(
 	"defaults",
 	"config.js"
 )));
+
+function getVersion() {
+	const gitCommit = getGitCommit();
+	return gitCommit ? `source (${gitCommit})` : `v${pkg.version}`;
+}
+
+let _gitCommit;
+function getGitCommit() {
+	if (_gitCommit !== undefined) {
+		return _gitCommit;
+	}
+	try {
+		_gitCommit = require("child_process")
+			.execSync("git rev-parse --short HEAD 2> /dev/null") // Returns hash of current commit
+			.toString()
+			.trim();
+		return _gitCommit;
+	} catch (e) {
+		// Not a git repository or git is not installed
+		_gitCommit = null;
+		return null;
+	}
+}
 
 function setHome(homePath) {
 	this.HOME = expandHome(homePath || "~/.lounge");
--- a/src/server.js
+++ b/src/server.js
@@ -81,8 +81,10 @@ module.exports = function() {
 
 	manager.sockets = sockets;
 
-	var protocol = config.https.enable ? "https" : "http";
-	log.info("The Lounge v" + pkg.version + " is now running on", protocol + "://" + (config.host || "*") + ":" + config.port + "/", (config.public ? "in public mode" : "in private mode"));
+	let protocol = config.https.enable ? "https" : "http";
+	let host = config.host || "*";
+	log.info("The Lounge", Helper.getVersion(), "is now running");
+	log.info(`Available on: ${protocol}://${host}:${config.port}/ in ${config.public ? "public" : "private"} mode`);
 	log.info("Press ctrl-c to stop\n");
 
 	if (!config.public) {
@@ -110,17 +112,6 @@ function allRequests(req, res, next) {
 	return next();
 }
 
-// Information to populate the About section in UI, either from npm or from git
-var gitCommit = null;
-try {
-	gitCommit = require("child_process")
-		.execSync("git rev-parse --short HEAD 2> /dev/null") // Returns hash of current commit
-		.toString()
-		.trim();
-} catch (e) {
-	// Not a git repository or git is not installed: treat it as npm release
-}
-
 function index(req, res, next) {
 	if (req.url.split("?")[0] !== "/") {
 		return next();
@@ -135,7 +126,7 @@ function index(req, res, next) {
 			pkg,
 			Helper.config
 		);
-		data.gitCommit = gitCommit;
+		data.gitCommit = Helper.getGitCommit();
 		data.themes = fs.readdirSync("client/themes/").filter(function(themeFile) {
 			return themeFile.endsWith(".css");
 		}).map(function(css) {
--- a/src/client.js
+++ b/src/client.js
@@ -257,7 +257,7 @@ Client.prototype.connect = function(args) {
 	});
 
 	network.irc.connect({
-		version: pkg.name + " " + pkg.version + " -- " + pkg.homepage,
+		version: pkg.name + " " + Helper.getVersion() + " -- " + pkg.homepage,
 		host: network.host,
 		port: network.port,
 		nick: nick,


src/command-line/index.js
```diff
@@ -7,7 +7,7 @@ var fsextra = require("fs-extra");
 var path = require("path");
 var Helper = require("../helper");
 
-program.version(pkg.version, "-v, --version");
+program.version(Helper.getVersion(), "-v, --version");
```
williamboman: Not sure how I feel about this one though. ¯\\\_(ツ)_/¯

screenshot 2016-09-05 at 14 44 38

src/helper.js
```diff
@@ -1,14 +1,19 @@
+'use strict';
```
williamboman: (`for...of` and `let`)

YaManicKill:
So, the errors aside (because of the ES6 stuff, which we need to figure out), I like this. As far as I can see, it'll still show the npm version if it's installed from there?

williamboman:
> As far as I can see, it'll still show the npm version if it's installed from there?

Yes. This PR is pretty redundant in terms of the amount of users affected by this (we're probably down to a few ‰ :smile:).

src/helper.js
```diff
@@ -20,6 +25,34 @@ Helper.config = require(path.resolve(path.join(
...
+	}
+	return pkg.version;
+}
+
+let _gitRef;
+function getGitRef() {
+	if (_gitRef) {
+		return _gitRef;
+	}
+	const commands = ["git describe HEAD", "git rev-parse --short HEAD"];
```
williamboman: Just `git describe HEAD` should be enough. Was introduced in 1.1.0.

src/helper.js
```diff
@@ -20,6 +25,34 @@ Helper.config = require(path.resolve(path.join(
...
+	for (let command of commands) {
+		try {
+			const gitRef = require("child_process")
+				.execSync(`${command} 2> /dev/null`)
+				.toString()
+				.trim();
+			return _gitRef = gitRef;
```
astorije: Please break this down into 2 statements:

```js
_gitRef = gitRef;
return gitRef;
```

src/helper.js
astorije: So, if the first command succeeds, the `for` loop assigns stuff then continues to the next one, but if it fails, it... continues to the next one? 😅
Not sure that was the intent here.

Indeed, you can go with just one command, no need to over-complexify the code for this. Chances that the command is not available are way lower than not having `git` at all, and it's fine to default to npm version otherwise.

src/server.js
```diff
@@ -72,7 +72,7 @@ module.exports = function() {
 	manager.sockets = sockets;
 
 	var protocol = config.https.enable ? "https" : "http";
-	log.info("The Lounge v" + pkg.version + " is now running on", protocol + "://" + (config.host || "*") + ":" + config.port + "/", (config.public ? "in public mode" : "in private mode"));
+	log.info("The Lounge v" + Helper.getVersion() + " is now running on", protocol + "://" + (config.host || "*") + ":" + config.port + "/", (config.public ? "in public mode" : "in private mode"));
```
astorije: Good job consolidating version on the command line and version in the UI, I didn't even think of that when I added it in the UI :-)

src/server.js
```diff
@@ -100,17 +100,6 @@ function allRequests(req, res, next) {
...
 
-// Information to populate the About section in UI, either from npm or from git
-var gitCommit = null;
-try {
-	gitCommit = require("child_process")
-		.execSync("git rev-parse --short HEAD 2> /dev/null") // Returns hash of current commit
-		.toString()
-		.trim();
-} catch (e) {
-	// Not a git repository or git is not installed: treat it as npm release
```
astorije: Try to leave meaningful comments in when you move the code around, @williamboman.

src/server.js
```diff
@@ -121,7 +110,7 @@ function index(req, res, next) {
 			pkg,
```
astorije: So, this merge of `data` and `pkg` is why `client/index.html` L337 and L338 can make use of `version`. In other words, version being displayed in the UI will come from the `package.json` file and not your `getVersion()` function.
Is that expected?
Do you think it could create some confusion if `getVersion()` and the `version` field of the `package.json` file do not return the same value (which they probably don't in most cases)?

src/helper.js
```diff
@@ -1,14 +1,19 @@
+"use strict";
```
astorije: 👏

src/command-line/index.js
astorije: I remember getting the same feeling when implementing the UI version. This is what should be displayed:

1. If this is a git repository, git is available and git command is available, display git commit.
2. Otherwise, display npm version

As an amendment to (1.), if the commit hash has an exact match for an npm version (i.e. a tag), we may want to display npm version instead. Though it might be a bit confusing as, essentially, instance **is** running from source and not from a build... so I'd say, if git repo + git available, go for git commit.

However, that npm version + commit hash (like in your picture above) is a no-no for me. As a user, what does that mean, am I running the pre.6 version or a specific commit? Is that commit hash related to that version? If I notice a bug, does that mean pre.6 have to be incriminated?
I'm in favor of simplicity here: an npm version, or a commit hash, but not both.

astorije:
Hey @williamboman, thanks and good job consolidating versions!
A bit of an overkill at the moment though, see my comments above :D Almost there though.

One more comment. Maybe a stretch, but we have to get started somewhere: how easy/hard would it be to write a couple tests for this?

src/helper.js
williamboman: It'll try all commands in the array, as soon as one succeeds it'll cache the result for subsequent invokations and then return the stdout. If the command fails it'll continue to the next one, and if all fails return `null`.

src/command-line/index.js
williamboman: What do you mean by npm version? All versions are (so far) given git tags as well (which is what git-describe uses to produce these refs).

> As a user, what does that mean, am I running the pre.6 version or a specific commit? Is that commit hash related to that version?

I can agree on that. What about `2.0.0-pre.6-31-gbdd3fec (dirty)`. The entire point of this PR is to move from just providing commit SHAs to something with more face value. What about the version displayed on the client? Does this apply to the client as well?

src/server.js
williamboman: Obligatory "duh, my code is self-describing". I'll keep the comment but chop off the last part :smile:.

williamboman:
> Hey @williamboman, thanks and good job consolidating versions!
> A bit of an overkill at the moment though, see my comments above :D Almost there though.

So firstly I think we need to agree on how to present the version number if the instance is running from source on a non-tagged commit. My preference is to use `git describe` to produce git refs that map a commit SHA to the closest release in history, e.g. `2.0.0-pre.6-2-gabcdef`. I understand this could cause a lot of confusion though.

The other option would be to simply present the commit SHA (which is how it currently works). Then this PR would have to shift focus entirely to just consolidate version numbers throughout all interfaces.

src/command-line/index.js
astorije: > What do you mean by npm version?

Sorry, I should have defined terminology I was using:

- What I call "npm version" is that human-friendly "1.5.0", "2.0.0-pre.6" string, whether it comes from a git tag or from the `package.json` file (let's call that simply a "version" from now on)
- What I call "commit hash" is... well, a (7-char) git commit hash :-)

> The entire point of this PR is to move from just providing commit SHAs to something with more face value.

But that's the issue here. There is no face value at all on an instance that runs a given version + N commits on top of it. We cannot inform the user about a version if a single commit has been added to it. What if that commit reverts all or part of a version (think issues added in pre.6)? What if that commit adds bugs? Should the users blame/report about that version or that commit?
For those reasons, I'd much rather take no "version responsibility" as soon as a single commit has been added to an instance.

> What about the version displayed on the client? Does this apply to the client as well?

You raise a good point and I believe versions should match in both the console and the UI.
If we want to push the 🚲 🏠 further, if not on a version, console could list the latest version + all commits added to that version, but we're going waaaaay too far! After all, we can't be doing the admin's job when it comes to using git 😅
So yeah, I definitely believe same on console + UI, and if admin wants more info, well, `git log` is there for that.

src/server.js
astorije: Yeah, I've heard that before... It's usually followed by "wait, why did I do that already?" a few days/weeks/months/after lol. Actually, [*cough cough*](https://github.com/thelounge/lounge/pull/575/files#r77442290)... 😘

Chop off what?

astorije:
> My preference is to use git describe to produce git refs that map a commit SHA to the closest release in history, e.g. v2.0.0-pre.6-2-gabcdef. I understand this could cause a lot of confusion though.

My preference goes the other way around for said confusion, and other reasons described [in a comment above](https://github.com/thelounge/lounge/pull/592#discussion_r77763195) (mostly because of the info being displayed doesn't reflect reality too well: this is not the running version anymore!).
Thoughts from others? :)

> Then this PR would have to shift focus entirely to just consolidate version numbers throughout all interfaces.

I actually think that such consolidation would be pretty great already! Actually, maybe worth doing a first pass with that consolidation, then in a further PR play with the other options? There are actually many ways git can report this and we might want to look into it, while the other goal (UI+console consistency) is a no-brainer.

> Tests for printing version numbers in the CLI, or?

I know, this is rather far-fetched and probably useless, but I'd love it if we could write tests for our app... So probably ignore my comment, but if you have a good and simple idea, I'll take it :-)

YaManicKill:
> Thoughts from others? :)

I like the idea of showing a version and a git commit, as it actually gives you an idea where you are on git, rather than having to open the commit and guess from the previous commits.

However, I don't use the version in the app enough to care. I understand the issues you have with it, @astorije and I won't argue it much.

williamboman:
I'm just annoyed by the version number being reported in the CLI when running from source, hence this PR. I don't feel very strongly about the semantics so I've shifted the focus of this PR consolidate version numbers reported by the app.

src/command-line/index.js
```diff
@@ -7,7 +7,12 @@ var fsextra = require("fs-extra");
 var path = require("path");
 var Helper = require("../helper");
 
-program.version(pkg.version, "-v, --version");
+program.version({
+	// Hack to defer execution of Helper.getGitCommit()
+	toString() {
+		return Helper.getGitCommit() ? Helper.getGitCommit() : pkg.version;
+	}
```
williamboman: Thoughts? Feels a bit unnecessary to invoke `Helper.getGitCommit()` for every command.

xPaw: Are you trying to check if `Helper.getGitCommit` exists? Because you are invoking it either way.

src/server.js
```diff
@@ -71,8 +71,14 @@ module.exports = function() {
...
-	var protocol = config.https.enable ? "https" : "http";
-	log.info("The Lounge v" + pkg.version + " is now running on", protocol + "://" + (config.host || "*") + ":" + config.port + "/", (config.public ? "in public mode" : "in private mode"));
+	let protocol = config.https.enable ? "https" : "http";
+	let host = config.host || "*";
+	if (Helper.getGitCommit()) {
+		log.info(`The Lounge is running from source (${Helper.getGitCommit()})`);
+	} else {
+		log.info(`The Lounge v${pkg.version} is now running`);
+	}
+	log.info(`Available on: ${protocol}://${host}:${config.port}/, in ${config.public ? "public" : "private"} mode`);
```
xPaw: Remove the `,` after the url.

src/command-line/index.js
williamboman: No, I'm deferring execution of `toString` until `commander` [wants to use it](https://github.com/tj/commander.js/blob/v2.9.0/index.js#L824).

The overhead of spawning a child process to get the git SHA is pretty much non-existent though, but it feels a bit unnecessary to do it for every single command (especially since it'll fail since most lounge installations are done through `npm`, I assume).

xPaw:
Could you also add the commit in https://github.com/thelounge/lounge/blob/2bb782fe81089bc634ae4946c68806a6d89b6e99/src/client.js#L255

src/server.js
```diff
@@ -71,8 +71,14 @@ module.exports = function() {
...
 	manager.sockets = sockets;
 
-	var protocol = config.https.enable ? "https" : "http";
-	log.info("The Lounge v" + pkg.version + " is now running on", protocol + "://" + (config.host || "*") + ":" + config.port + "/", (config.public ? "in public mode" : "in private mode"));
+	let protocol = config.https.enable ? "https" : "http";
+	let host = config.host || "*";
+	if (Helper.getGitCommit()) {
+		log.info(`The Lounge is running from ${Helper.getVersion()}`);
+	} else {
+		log.info(`The Lounge ${Helper.getVersion()} is now running`);
```
xPaw: Not a particular fan of using the ES6 bollocks. The `log` allows for concat by doing:
`log.info("The Lounge", Helper.getVersion(), "is now running");`

src/helper.js
```diff
@@ -1,3 +1,6 @@
...
+
+const pkg = require('../package.json');
```
xPaw: Double quotes.

src/command-line/index.js
xPaw: Remove `pkg` definition as it's no longer used in this file.

src/server.js
```diff
@@ -71,8 +71,14 @@ module.exports = function() {
 
 	manager.sockets = sockets;
 
-	var protocol = config.https.enable ? "https" : "http";
-	log.info("The Lounge v" + pkg.version + " is now running on", protocol + "://" + (config.host || "*") + ":" + config.port + "/", (config.public ? "in public mode" : "in private mode"));
+	let protocol = config.https.enable ? "https" : "http";
+	let host = config.host || "*";
+	if (Helper.getGitCommit()) {
```
xPaw: I'm not even sure this different message is needed, could always just be printing the `else` statement.

src/server.js
```diff
@@ -121,7 +116,7 @@ function index(req, res, next) {
 			pkg,
 			Helper.config
 		);
-		data.gitCommit = gitCommit;
+		data.gitCommit = Helper.getGitCommit();
```
xPaw: Can the frontend just use `Helper.getVersion()` too? Then the `getGitCommit` function is only internal and `getVersion` would be the single authority.

astorije:
@williamboman, could you rebase this PR with `master` please?

src/helper.js
```diff
@@ -22,6 +25,29 @@ Helper.config = require(path.resolve(path.join(
...
+			.toString()
+			.trim();
+		return _gitCommit;
+	} catch (e) {
+		// Not a git repository or git is not installed
+		_gitCommit = null;
+		return null;
+	}
+}
+
```
astorije: Not too happy this was merged with removed comments as mentioned below some time ago, but oh well.

PR #593, opened 2016-09-05
	lint: default to ecmaVersion: 6, keep ecmaVersion: 5 for client/

williamboman:
Cleanest solution I could come up with right now.

Adds ~3s to the linting process though, ymmv. (not sure how big part the added es6 environment plays in that, doubt it's much)

Unblocks #592.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -7,6 +7,9 @@ env:
   mocha: true
   node: true
 
+parserOptions:
+  ecmaVersion: 6
+
 rules:
   block-spacing: [2, always]
   brace-style: [2, 1tbs]
--- a/package.json
+++ b/package.json
@@ -24,7 +24,9 @@
     "test": "npm-run-all -c test:mocha lint",
     "test:mocha": "mocha -r test/fixtures/env.js test/**/*.js",
     "lint": "npm-run-all -c lint:js lint:css",
-    "lint:js": "eslint .",
+    "lint:js": "npm-run-all -c lint:js:es5 lint:js:es6",
+    "lint:js:es5": "eslint --parser-options=\"ecmaVersion:5\" client/",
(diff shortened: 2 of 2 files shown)


astorije:
Awesome start!

Reason why it takes a bit longer to run is because you are loading ESLint "environment" (or whatever it's called) 3 times now.

I think this is a bit convoluted though IMO. I'd go for simpler: an ES6-enable one and another that it's not. Something like `npm run lint:js:es3` that we run against `client/` only, and `npm run lint:js:es6` that we run against anything else.
How does that sound?

williamboman:
Yeah that makes sense. Shaved off 1.5-2s.

YaManicKill:
:+1: looks good

astorije:
Slightly off-topic but as an FYI in case it's useful someday, I discovered that setting up `es6` in the  `env` list [sets up the `ecmaVersion` to `6`](https://github.com/eslint/eslint/blob/e1187285f137d685bcfc9e2a771fa059ef9637a9/conf/environments.js#L98-L103) as well, even though it's not documented anywhere (well, [for now](https://github.com/eslint/eslint/pull/7067)...

Anyway, 👍 on this, cool!

williamboman:
> Slightly off-topic but as an FYI in case it's useful someday, I discovered that setting up es6 in the env list sets up the ecmaVersion to 6 as well, even though it's not documented anywhere (well, for now...

Yep, that's how I did it first. #javascript on IRC told me --env es6 is being deprecated and that ES version should be set through parser options.

astorije:
> `--env es6` is being deprecated and that ES version should be set through parser options.

Hmm, that's highly surprising (unless they were saying that `--env` is deprecated but `env: es6` in the file is not). Note that when using actual ES6 features, such as `Promise` or other globals, `env: es6` **has** to be specified. But we are not concerned about that for now :-)

PR #594, opened 2016-09-06
client: don't dismiss notifications programmatically after 5s

williamboman:
I suspect this was originally implemented because Chrome would
not automatically close notifications before Chrome 47 (October 2015).

This'll allow Lounge notifications to linger after being automatically closed, for example
in the notification tray in macOS. 
screenshot 2016-09-07 at 01 40 46

This will cause a noticeable difference, especially in browsers/environments that doesn't use the OS notification API (Chrome). Notification cards will be visible much longer before being dismissed automatically.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -905,14 +905,11 @@ $(function() {
 						icon: "img/logo-64.png",
 						tag: target
 					});
-					notify.onclick = function() {
+					notify.addEventListener("click", function() {
 						window.focus();
 						button.click();
 						this.close();
-					};
-					window.setTimeout(function() {
-						notify.close();
-					}, 5 * 1000);
+					});
 				}
 			}
 		}


astorije:
So before 47 they were staying forever, and now they close after a certain time by default? How long is the default? Did you test on other browsers?

williamboman:
> So before 47 they were staying forever, and now they close after a certain time by default? How long is the default?

Exactly. I think I read 8 seconds somewhere, but it certainly feels longer than that.

> Did you test on other browsers?

I've only tested on macOS. Safari and Firefox uses native OS notifications (works great). Chrome has their own notification mechanism (iirc they'll be moving to native OS notifications in the future) and now notification cards will be visible for a longer time after being dismissed automatically and kept in the notification tray, like so;

client/js/lounge.js
```diff
@@ -905,14 +905,11 @@ $(function() {
 						icon: "img/logo-64.png",
 						tag: target
 					});
-					notify.onclick = function() {
+					notify.addEventListener("click", function() {
```
astorije: Is that necessary or just a nice-to-have? If the latter, why? Seems like there is nothing wrong with `onclick`, no?

williamboman: I just think registering event listeners as properties on an object is ugly, and one should use the `EventTarget` interface whenever possible.

astorije:
So yeah, as far as I know, this was done because Chrome would leave the notification open forever: https://github.com/erming/shout/commit/d1b3f375c52e4eae3e8a896d39afdb019f95b6a5
It doesn't seem necessary anymore, so good we can get rid of workarounds like this! 👍 

FYI, notifications now stay 20 seconds instead of 5. It might seem long, but since it's the browser's default, I think it's ok (and anyway better than playing with it ourselves like that).

xPaw:
The spec says the following:

> User agents should run the close steps for a non-persistent notification a couple of seconds after they have been created.

I tested a notification in Chrome 53, and it takes like 15 seconds to close the notification, which is a little too long for my liking. The spec's `couple of seconds` is quite vague as well.

dgw:
5 seconds is actually a pretty good duration, IMO. 15 is an eternity.

My interpretation of the spec's `couple of seconds` is `a sensible duration`.

astorije:
I see 3 advantages with this PR:

1. It removes some fairly useless code. Minor advantage though as we are talking about 3 lines.
2. It makes notifications for The Lounge consistent in duration with other apps that are respecting the default timeout. From a UX standpoint, that's pretty good and more likely to respect user decisions. For example, if there are assistive technologies that control this, we might be getting in their way by forcing the destroy. Note that this is highly hypothetical though, so minor advantage as well.
3. @williamboman, you mentioned the notification tray in his PR description. Does this mean that without this PR, the behavior on the OS level is negatively impacted? If answer is yes, then this PR is a no brainer. Even though that looks like the OS fault, that would still be pretty bad.

williamboman:
> Does this mean that without this PR, the behavior on the OS level is negatively impacted?

Yes.

astorije:
Then I think this is a no-brainer. We should definitely leave that to the OS.
@xPaw, would you be ok with this, knowing that it breaks macOS notification integration otherwise?

xPaw:
Also worth noting that the Notification JS shim we use does not automatically close the notifications (e.g. in browsers that don't support it).

astorije:
Well, sucks for them really. I'd rather not break on OS/browsers that respect the spec than not be optimal on the others :)
Anyway, feel free to 👍 and merge or 👎 and close. I don't think we'll get something better than that.

xPaw:
With the current polyfill we use, that's a 👎 from me.

astorije:
I looked around, and there doesn't seem to be a better polyfill.

@williamboman, I'm afraid we have to close this PR. Feel free to re-open if this bug is too annoying on macOS or if you feel strongly about it.

Seriously, this is really sad. We are essentially refusing to fix a bug that appears on a modern OS just to be nice to Internet Explorer. If it were up to me, I'd say screw Internet Explorer: we remove the polyfill and tell people to use Edge 14 if they want web notifications. Any strong opinions on this?

williamboman:
Why not just

```diff
diff --git a/client/js/libs/notification.js b/client/js/libs/notification.js
index 4d84b16..a7fc0a8 100644
--- a/client/js/libs/notification.js
+++ b/client/js/libs/notification.js
@@ -13,6 +13,10 @@
 
 (function(window, document){
 
+	if (window.Notification) {
+		return
+	}
+
 	var PERMISSION_GRANTED = 'granted',
 		PERMISSION_DENIED = 'denied',
 		PERMISSION_UNKNOWN = 'unknown';

```

Issue #600, opened 2016-09-11
Message input text is small on Morning and Zenburn


astorije:
For instance:

Example | Morning
--- | ---
<img width="1014" alt="screen shot 2016-09-11 at 18 09 11" src="https://cloud.githubusercontent.com/assets/113730/18420913/ec4151fc-784b-11e6-8cca-db47bc98c98a.png"> | <img width="1012" alt="screen shot 2016-09-11 at 18 08 53" src="https://cloud.githubusercontent.com/assets/113730/18420912/ec3e62b2-784b-11e6-86f9-3254fc831752.png">

This could have to do with one of the many things that changed regarding to fonts recently, such as #562 or #568.

PR #601, opened 2016-09-12
Fix small input text on Morning and Zenburn

astorije:
Fixes https://github.com/thelounge/lounge/issues/600.

This bug was (re-)introduced in https://github.com/thelounge/lounge/pull/562, [here specifically](https://github.com/thelounge/lounge/pull/562/files#diff-53a2103ea4ad1eb199a9906db330e16bL34). `inherit` was setting text back to 12px instead of 13px.

Before | After
--- | ---
screen shot 2016-09-11 at 18 08 53 | screen shot 2016-09-12 at 01 22 53

### DIFF
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -30,6 +30,7 @@ body {
 }
 
 #windows .header .topic,
+#windows #form .input,
 .messages .msg,
 .sidebar {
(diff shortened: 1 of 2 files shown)


maxpoulin64:
Looks good to me 👍

PR #603, opened 2016-09-13
Hide the unread marker on Esc

nornagon:


astorije:
I think this goes against something else we are heading towards.

If you follow the discussions on https://github.com/thelounge/lounge/pull/573 and https://github.com/thelounge/lounge/pull/575, you'll see that we are taking steps towards good navigation in the app.
In terms of UX, it's more sensible to keep `Esc` for action cancellation, which is not the case here. For us, this means that we should be able to press `Esc` (or back button on mobile) to leave the settings, or go to that blank/home page we are talking about [here](https://github.com/thelounge/lounge/pull/573#issuecomment-240949223).

Thank you very much, but I'm sorry I'm going to be strongly 👎 on this one.
If you want to help with the scenarios mentioned above and in the linked PRs, that would be greatly appreciated! Granted it's more than a 2-line change, but it wouldn't be fun if it was too easy 😅

nornagon:
Hey, thanks for the comment! I get wanting to use `Esc` for action cancellation, like "get me out of here". That said, the 'new messages' marker bugs me when it doesn't disappear, and I'd really like an easy way to dismiss it.

Do you have any other suggestions for how that might be accomplished? `Esc` is how Slack implements this, FWIW :)

dgw:
I don't think it should ever be necessary to hide the unread marker. If it's appearing at times where it's annoying, its behavior should be modified. For example, I don't find it useful to have the marker continually appear when I am actively chatting, but it does. Things like hiding the marker until the window loses focus would be preferable to manual overrides.

I see where @astorije is coming from on this :)

astorije:
> For example, I don't find it useful to have the marker continually appear when I am actively chatting, but it does.

That's an odd behavior, and I believe it should not do that. I'll keep an eye out.

PR #609, opened 2016-09-13
Add babel and webpack to the build process to allow ES2015

### LINKED ISSUE #426: Add babel
It would be nice if we would add babel to our frontend build process, so that we can use lovely ES6 syntaxes.

nornagon:
Fixes #426.

[Some things that ES2015 enables](http://babeljs.io/docs/learn-es2015/):
- Arrow functions: `(n => n + 5)`
- Template strings: <code>\`Hello ${"world"}`</code>
- Destructuring: `var {foo, bar} = {foo: 3, bar: "hi"}`
- `import`, so the 1000-line `lounge.js` can be broken up into modules

That said, it adds complexity too:
- now you have to run `grunt watch` to rebuild on file change
- ES2015 is imperfectly handled by babel; there are some things it can't transpile, and some which it can but which also require a polyfill to be loaded

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,12 +18,12 @@
     "build:font-awesome": "node scripts/build-fontawesome.js",
     "build:libs": "uglifyjs client/js/libs/*.js client/js/libs/jquery/*.js client/js/libs/handlebars/*.js -o client/js/libs.min.js --source-map client/js/libs.min.js.map --source-map-url libs.min.js.map -p relative",
     "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
+    "build:webpack": "webpack",
+    "watch": "webpack -w",
     "test": "npm-run-all -c test:mocha lint",
     "test:mocha": "mocha -r test/fixtures/env.js test/**/*.js",
     "lint": "npm-run-all -c lint:js lint:css",
-    "lint:js": "npm-run-all -c lint:js:es5 lint:js:es6",
-    "lint:js:es5": "eslint --parser-options=\"ecmaVersion:5\" client/",
-    "lint:js:es6": "eslint --ignore-pattern client/ .",
+    "lint:js": "eslint .",
     "lint:css": "stylelint \"**/*.css\"",
     "prepublish": "npm run build"
   },
@@ -58,6 +58,8 @@
     "ldapjs": "1.0.0"
   },
   "devDependencies": {
+    "babel-loader": "6.2.5",
+    "babel-preset-es2015": "6.14.0",
     "chai": "3.5.0",
     "eslint": "3.6.0",
     "font-awesome": "4.6.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -61,10 +61,12 @@
     "ldapjs": "1.0.0"
   },
   "devDependencies": {
+    "babel-preset-es2015": "^6.14.0",
```
astorije: Use a fixed version instead.

package.json
```diff
@@ -61,10 +61,12 @@
...
     "chai": "3.5.0",
     "eslint": "2.11.1",
     "font-awesome": "4.6.3",
     "grunt": "1.0.1",
+    "grunt-babel": "^6.0.0",
```
astorije: Any chance we could live without `grunt-babel` or other grunt stuff? Our recent changes managed to avoid using Grunt, and frankly I'm hoping we can just get rid of it soon (and just use npm scripts or JS scripts if things are a bit too complex for npm scripts).

astorije:
This also needs some changes around ESLint:

- Ignore the `lounge.min.js` file
- Simplify the ES5 vs ES6 config 🎉
- We might want to add some ESLint rules maybe

nornagon:
I think we also need to add webpack for `import` to be useful. I'll work on pulling out the grunt stuff & updating the eslint config as well as having this change include a reasonable default webpack config.

nornagon:
Updated to add webpack, fix lint, remove grunt stuff.

package.json
```diff
@@ -17,16 +17,17 @@
...
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha -r test/fixtures/env.js test/**/*.js",
     "start": "node index",
-    "build": "npm run build:font-awesome && npm run build:grunt && npm run build:handlebars",
+    "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",
     "build:grunt": "grunt",
     "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
+    "build:webpack": "webpack",
+    "watch": "npm-run-all --parallel watch:*",
+    "watch:webpack": "webpack -w",
```
xPaw: Just make this `watch` and remove the previous line.

nornagon: Done!

xPaw:
Now that your uglifyjs PR has been merged, could you update this to run it via webpack?

nornagon:
Unfortunately that would still require refactoring all the libs to use
import/require instead of injecting their stuff into the global namespace.
Possible, but not trivial.

astorije:
Could you rebase please? :)

nornagon:
Rebased!

webpack.config.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+	entry: "./client/js/lounge.js",
+	resolve: {
+		extensions: ["", ".js"]
```
astorije: Why is `""` necessary?

astorije:
Actually, I take that back: #640 makes this unnecessary as they share the same commit. I believe #640 should take over this PR and both commits should ship together. Closing in favor of that other one.

Issue #604, opened 2016-09-13
Unread marker shows up incorrectly when joins/parts are hidden


nornagon:
To reproduce:
1. Hide joins/parts
2. Join a channel
3. Wait for someone to join/part
4. Observe 'New messages' marker appears, despite there being apparently no messages in the channel.

![_the_lounge](https://cloud.githubusercontent.com/assets/172800/18457572/ccdf6c22-790e-11e6-8ea2-1d0eb8b6348f.png)

astorije:
Yes, this is definitely a bug.

Issue is that unread marker is being hidden with CSS when it's the last element of its parent. When there is (for instance) a join and joins are hidden, marker is not considered last child even though it appears to be so.

I'm not sure there is a way to hide "last-child-but-not-really" directly in CSS. I can't think of anything (tried a few using `.unread-marker:not(...):last-child` without any luck) right now.
Curious what @xPaw would think about this.

xPaw:
@astorije this can't be fixed in pure CSS. The simplest fix I can think of is checking if unread marker is the last visible element whenever appending received messages to the DOM.

nornagon:
IMHO the easiest way to handle this is to just not render hidden messages instead of annotating them with `display: none`. If the UI code were driven from state, it'd be simple to add/remove those messages from the DOM as necessary :)

astorije:
> @astorije this can't be fixed in pure CSS. The simplest fix I can think of is checking if unread marker is the last visible element whenever appending received messages to the DOM.

Agreed re: CSS. Regarding your solution, I'm not sure how this could work but I might have misunderstood. I think any solution would need to check if specific joins/parts/etc. are enabled or not and moved the marker or not depending on that.

> IMHO the easiest way to handle this is to just not render hidden messages instead of annotating them with display: none. If the UI code were driven from state, it'd be simple to add/remove those messages from the DOM as necessary :)

This is far from the easiest way as pretty much everything in the client is tied to that idea of having unwanted messages hidden in CSS instead of not being part of the DOM.
That being said, I am not against that solution at all, but keep in mind that it's a bit closer to a refactor than a solution :)
@xPaw, thoughts on this?

Issue #495, opened 2016-07-12
Pressing "Show messages" multiple times renders duplicate content


pinkeshbadjatiya:
![Image](https://ptpb.pw/Wz8F.png)

This issue arises when clicking "Show messages" takes some time to render. When i click it multiple times before it renders the content, then, i see the above output.

maxpoulin64:
This is because the button just requests messages before the last one loaded, so it requests the same until new ones arrives. Should be an easy fix, just need to disable the button while one request is made.

astorije:
I cannot reproduce this on my end, I think a combination of faster loading + scroll "stickiness" reduced the impact.
Besides, this should be fixed whenever someone wants to take a stab at implementing https://github.com/thelounge/lounge/issues/207 😄, so it's essentially a dup of that other ticket.

Either way, closing this for these reasons.

pinkeshbadjatiya:
fair enough.

YaManicKill:
Hmmm, this will still be a problem in some cases. Ideally, we should just disable the button once it is clicked until we get the more messages through.

astorije:
@YaManicKill whenever #207 gets done, there won't be a button anymore.

YaManicKill:
True, but this is still an issue until that gets done (afaik, no one is working on it).

astorije:
Right, but there is no need to keep 2 tickets open for this. Also, are you able to reproduce?

YaManicKill:
> Right, but there is no need to keep 2 tickets open for this. 

But 1 is a bug and 1 is a feature request. They aren't the same thing.

> Also, are you able to reproduce?

My internet connection is too fast. I'll need to simulate slower internet to get it to work.

MiniDigger:
its definitely still an issue. used chrome dev tools to throttle my connection and was able to reproduce it. never experienced this in a real world scenario tho (I don't go around pressing buttons like crazy normally :D), ill try next week on my daily commute, got some bad network reception there. 

the fix @YaManicKill proposed seems simple. don't really see #207 happening any time soon.

astorije:
> But 1 is a bug and 1 is a feature request. They aren't the same thing.

I get that. I'm just saying I believe the amount of work to do either of those should be fairly similar, so if someone wants to give it a shot, I'd rather they don't waste their time on something that will not stay, if that makes sense.
Also, if we had like < 50 open tickets, I'd be happy to keep both open, but right now this is far from being the case and it's hurting visibility on the project IMO :)

[...]

PR #610, opened 2016-09-14
Strawman: begin moving to React

### LINKED ISSUE #544: User list refreshes when scrolling, rendering it useless on high activity channels.
Currently on thelounge 2.0.0-pre.4
The user list refreshes when users join so if you’re in the middle of scrolling to select a nickname to talk to it jumps and you always end up selecting the wrong nickname.

nornagon:
Per discussion on IRC, the client code is presently rather tangly and stateful, and it'd be nice to straighten it out. I'm not requesting that this PR be merged _per se_, just hoping to have a discussion around how this could end up shaking out and whether there's interest to continue in this direction :)

Let me know what you think!

(Incidentally, I think this fixes #544.)


nornagon:
I've been further developing this locally and I've ported most of the message page over to React, but as you all know, it's pretty tangled up with the sidebar code, so there's a lot of interactions between non-React and React things (in particular, the active channel is kind of "controlled" by the sidebar, so until the sidebar is _also_ in React, those two things can't be linked together Reactfully).

That said, the parts that I have extracted are highly testable on their own, and I've been writing tests for them with [Jest](https://facebook.github.io/jest) and [Enzyme](http://airbnb.io/enzyme) as I go.

The diff for moving the message list to react is huge (+800, -600 delta so far), and certainly has bugs of its own that I'm yet to iron out. There's definitely light at the end of the tunnel, though! It's extremely nice to have well-factored, testable, separable components, and to be able to separate functionality into modules with `import`.

Obviously this is all post-2.0.0 stuff, but I'm curious to hear @xPaw and @astorije's thoughts on how best to move forward on this. Here are some of my thoughts:
- This already fixes some bugs (#604, #495), but they're pretty minor bugs which could be fixed in a much more straightforward fashion.
- Tests are great. Some behaviors are less testable than others right now due to the code being half-React (so interactions that rely on the stuff outside of React isn't testable), but even just being able to test the React parts is fantastic.
- I'm sure I won't be around forever to push this kind of refactor effort, so it's important that the maintaining team is on board & wants this. Y'all 🎉 'd this PR so I take that it's at least interesting, but I'd like to have a discussion about whether this makes sense at the project/maintenance level. If you don't want to have to learn/deal with React & co. as a dependency, or if you'd only want that if there were someone consistently around to lead the effort and help teach those systems (i.e. me), then perhaps this isn't the right direction for the project. I'm likely to be able to contribute significantly for the next ~6 weeks, but after then I'll be less available.

What do you folks think?

xPaw:
I think if we were to move to using a stateful JS framework, I'd prefer using Vue.js as it's lighter and faster.

Both me and @maxpoulin64 are not particular fans of React.

xPaw:
See #633 for discussion.

PR #612, opened 2016-09-15
Disable tooltips on mobile to prevent them to stay after clicking

### LINKED ISSUE #585: Tooltips do not hide on mobile after clicking the underlying button
Asking for a friend

> @xPaw:	Lul, sending a message on mobile leaves "send message" tooltip
> @williamboman:	Yeah tooltips should be disabled altogether on handheld devices
> @xPaw:	They should not
> @xPaw:	They appear when holding, which is good
> @xPaw:	Same thing happens with user list icon, maybe the click events are cancelled
> ...
> @MaxLeiter:	On mobile the user-list and send message tool tips stay visible after tapping them, not too useful :(
> @MaxLeiter:	Actually, all icons with tool tips (settings too, for example)

Turns out, when looking at an instance not running #540, bug was there already on Connect/Settings/Logout buttons, but we wouldn't notice it because clicking on them hides the parent object (the sidebar).
It might be a bug with [Primer's tooltips](http://primercss.io/tooltips/)...

Help welcome.

astorije:
This fixes #585. Yeah, I know, this is less than ideal. But this is somewhat blocking 2.0.0 release. And none of the other options are ideal either (some discussed options will simply not work).

On v1.5.0, these buttons didn't have any tooltips, so they won't be missed on mobile. Plus any hovering effect on mobile is kind of a lost battle anyway.

We have an open issue on Primer's tooltips: https://github.com/primer/tooltips/issues/4

I suggest we hide them on mobile to freely deploy 2.0.0, and see to work them back in when we have a better solution.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1557,6 +1557,24 @@ button {
 }
 
 @media (max-width: 768px) {
+	/**
+	  * TODO Replace this with `@media (hover: hover)` when Firefox supports it
+	  * See:
+	  * - http://stackoverflow.com/a/28058919/1935861
+	  * - http://caniuse.com/#feat=css-media-interaction
+	  * - https://www.w3.org/TR/mediaqueries-4/
+	  * - https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover
+	  */
+	.tooltipped:hover:before,
+	.tooltipped:hover:after,
+	.tooltipped:active:before,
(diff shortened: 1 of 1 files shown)


maxpoulin64:
Fine by me for now, not that I think the tooltips are particularly useful. This might catch browsers as well on small screens in combination with zoom, but the fixed annoyance outweights the benefits for me. 👍

PR #613, opened 2016-09-15
If the most recently active socket disconnects, start marking stuff as unread

### LINKED ISSUE #561: Unread marker does not show on current channel when initially loading the page

nornagon:
c.f. #567

Fixes #561. I think.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -208,6 +208,7 @@ function init(socket, client) {
 		socket.on(
 			"open",
 			function(data) {
+				client.lastActiveSocket = socket;
 				client.open(data);
 			}
 		);
@@ -221,6 +222,14 @@ function init(socket, client) {
 			"names",
 			function(data) {
 				client.names(data);
+			}
+		);
+		socket.on(
+			"disconnect",
+			function() {
+				if (socket === client.lastActiveSocket) {
+					client.lastActiveSocket = null;
+				}
 			}
 		);
 		socket.join(client.id);
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -43,7 +43,10 @@ Chan.prototype.pushMessage = function(client, msg) {
 		this.messages.splice(0, this.messages.length - Helper.config.maxHistory);
 	}
 
-	if (!msg.self && this.id !== client.activeChannel) {
+	// If this is a message sent by someone else in a channel other than the one
+	// most recently opened, OR the socket that opened that channel has since
+	// disconnected, mark it as unread.
+	if (!msg.self && (this.id !== client.activeChannel || !client.lastActiveSocket)) {
 		if (!this.firstUnread) {
 			this.firstUnread = msg.id;
 		}


astorije:
Did you manage to try this with success?

What happens if 2 clients A and B open a connection in this order, then B closes before A? I believe that when B leaves, it's considered the last active connection so server thinks there is no one connected anymore.
Is this assumption correct?

If so, we could probably use a counter instead. When counter goes back to 0, there are no open connections.

Also, this only fixes part of the problem: I believe that when opening The Lounge on a new page, a couple clicks are emitted to the active channel. That second click hides the unread marker.

src/server.js
```diff
@@ -223,6 +224,14 @@ function init(socket, client) {
 				client.names(data);
 			}
 		);
+		socket.on(
+			"disconnect",
+			function() {
+				if (socket == client.lastActiveSocket) {
```
astorije: s/`==`/`===`/ :)

nornagon:
That is correct, and intended. "Last active connection" means "the
connection with the most recent activity" in this case.

The reasoning is this: we assume that "active" clients will have some user
interaction, and that usually only one client will be active at a time,
even if many are open. We attempt to determine which client is the active
one by tracking "open" events, since they are only triggered by user action
(i.e. Clicking on a channel).

When the active client disconnects, we assume that the user is no longer
interacting with Lounge, so we mark all future messages as unread. Other
clients that happen to be looking at a channel will not update their unread
marker, but it will appear if they refresh. If another client connects,
they will be restored at the active channel of the last client, with unread
marker intact.

I think it should be a goal that passive clients (clients that are
connected to the lounge but with which the user is not interacting) should
not impact other clients. This approach makes it so that if you open a
session and leave it open on a secondary machine but don't interact with it
at all, it's identical to not having that client open at all.

Side note: "client" and "connection" are super overloaded terms here... we
should get some more words.

src/server.js
```diff
@@ -208,6 +208,7 @@ function init(socket, client) {
 		socket.on(
 			"open",
 			function(data) {
+				client.lastActiveSocket = socket;
```
xPaw: This isn't exactly `last active`, but rather `last connected`, which is misleading in the intent.

nornagon: It isn't quite `last connected` either, it's the last socket that sent `open` (i.e. the last client that clicked on a channel).

xPaw: This approach seems fine (at least temporarily) by me. I'll 👍 it unless other maintainers have other considerations.

astorije:
I'm removing this from the incoming 2.0.0 release as it needs a bit more testing.

nornagon:
Is there a way I can write automated tests for this? It looks like there's some rudimentary `MockClient` code in `test/util.js`, would that be the right thing to build on top of?

nornagon:
Seems like @xPaw and @astorije have different ideas about how this should work than me, so I'll leave them to discuss :)

PR #617, opened 2016-09-18
Hide sidebar when app is loading in themes

astorije:
This change was originally made in #420 then I broke it for themes in #615, sigh...


MaxLeiter:
This PR works fine for me on Firefox 48 and Chrome 53.0.2785.116 64-bit

PR #616, opened 2016-09-17
#298 README.md updated install methods.

metsjeesus:
Closes #298.

PR #614, opened 2016-09-17
README.md development help information

metsjeesus:
Just learning, i hope i got it right this time.

PR #621, opened 2016-09-20
Add info on README about how to run from source, how to upgrade

### LINKED ISSUE #298: Clarify readme with instructions on how to run from git
Almost daily there are people that try to run Lounge by cloning from github, and it fails because they did not run the build.

I think we should rename `Development setup` to `Installing from git`, and also the `npm install` needs to account for installing dev dependencies (If your npm env is prod, it won't install them, and the build will fail).

astorije:
How to run from source and how to upgrade are 2 extremely common things asked on the IRC channel. Hopefully this is going to make things a bit more clear.

This closes #298.
Also, this aims to replace #614 and and #616, so I'd be happy to have feedback from @metsjeesus!

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,39 +34,53 @@ This fork aims to be community managed, meaning that the decisions are taken
 in a collegial fashion, and that a bunch of maintainers should be able to make
 the review process quicker and more streamlined.
 
-## Install
+## Installation and usage
 
-To use The Lounge you must have [Node.js](https://nodejs.org/en/download/) installed.
-The oldest Node.js version we support is 4.2.0.
+The Lounge requires [Node.js](https://nodejs.org/) v4 or more recent.
 
-If you still use 0.10 or 0.12 we strongly advise you to upgrade before installing The Lounge.
-For more information on how to upgrade, read the [documentation](https://nodejs.org/en/download/package-manager/).
+### Running stable releases from npm (recommended)
 
+Run this in a terminal to install (or upgrade) the latest stable release from
+[npm](https://www.npmjs.com/):
+
+```sh
+[sudo] npm install -g thelounge
 ```
-sudo npm install -g thelounge
+
+When installation is complete, run:
+
+```sh
+lounge start
 ```
 
-## Usage
+For more information, read the [documentation](https://thelounge.github.io/docs/), [wiki](https://github.com/thelounge/lounge/wiki), or run:
 
-When the install is complete, go ahead and run this in your terminal:
-
-```
+```sh
 lounge --help
 ```
 
-For more information, read the [documentation](https://thelounge.github.io/docs/) or [wiki](https://github.com/thelounge/lounge/wiki).
+### Running from source
 
-## Development setup
+The following commands install the development version of The Lounge. A word of
+caution: while it is the most recent codebase, this is not production-ready!
 
-To run the app from source, just clone the code and run this in your terminal:
-
-```
+```sh
+git clone https://github.com/thelounge/lounge.git
+cd lounge
 npm install
 npm start
 ```
 
-You will have to run `npm run build` if you change or add anything in
-`client/js/libs` or `client/views`.
+## Development setup
+
+Simply follow the instructions to run The Lounge from source above, on your own
+fork.
+
+Before submitting any change, make sure to:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,42 +34,50 @@ This fork aims to be community managed, meaning that the decisions are taken
...
+[npm](https://www.npmjs.com/):
+
+```sh
 sudo npm install -g thelounge
 ```
 
-## Usage
-
-When the install is complete, go ahead and run this in your terminal:
+When install is complete, run:
```
MaxLeiter: I think this should stay as "When the install", alternatively you could use "When installation is..."

README.md
```diff
@@ -34,42 +34,50 @@ This fork aims to be community managed, meaning that the decisions are taken
...
-## Development setup
-
-To run the app from source, just clone the code and run this in your terminal:
+The following commands install the development version of The Lounge. A word of
+caution: while it is the most recent codebase, this is not production-ready!
 
-```
+```sh
+git clone https://github.com/thelounge/lounge.git
+cd lounge
```
MaxLeiter: They may clone into a different directory, maybe `cd path/to/lounge`?

xPaw:
>  also the npm install needs to account for installing dev dependencies (If your npm env is prod, it won't install them, and the build will fail).

README.md
astorije: Good catch, addressed, thanks!

README.md
astorije: If command run is `git clone https://github.com/thelounge/lounge.git`, then the new directory is going to be `lounge` in the current one, so running these 2 commands altogether should always be good.
If people decide to run `git clone https://github.com/thelounge/lounge.git my-custom-dir`, I believe them to be advanced enough to adapt the `cd` to their needs 😅.

Idea is by default you can just copy and paste these in a terminal and it should work as is.

astorije:
> also the npm install needs to account for installing dev dependencies (If your npm env is prod, it won't install them, and the build will fail).

@xPaw, as in, `export NODE_ENV=production` being set system-wide? Just trying to understand, when can this be an issue? As in, when is such env being set? (I know it's the case for our current Dockerfile, but it's not following these instructions anyway)

xPaw:
That would be one of the cases, yes. I remember someone having this issue on irc.

astorije:
To be honest, I don't expect someone running installs from source to have their Node environment manually set up to production **and** being stuck because they don't have the development dependencies...

That seems like a specific scenario that I wouldn't want to worry about. Plus, running this on a prod machine, they need to realize that it should be done cautiously.

Also, "To run from source, run `npm install && npm start`" is kind of the *de facto* standard when it comes to Node packages. I expect they would have the same issue whatever they install from source.
@xPaw, mind leaving this as is and see if that issue arises again on the channel and/or GitHub issues? I am tempted to say it could be misconfiguration or miscomprehension in most cases.

xPaw:
This can be an issue on certain services: https://devcenter.heroku.com/changelog-items/688

astorije:
Then I still think people should be aware that running from source is not a good idea for production, and Heroku redirects them to https://devcenter.heroku.com/articles/nodejs-support#runtime-behavior.
Still believe we should do `NODE_ENV=development npm install` instead?
I don't have a strong opinion, I'm just puzzled (also, our Heroku doc does do simply `npm install` so we might not even have a choice, sigh...).

README.md
```diff
@@ -34,42 +34,50 @@ This fork aims to be community managed, meaning that the decisions are taken
...
-If you still use 0.10 or 0.12 we strongly advise you to upgrade before installing The Lounge.
-For more information on how to upgrade, read the [documentation](https://nodejs.org/en/download/package-manager/).
+### Running stable releases from npm (recommended)
 
-```
+Run this in a terminal to install (or upgrade) the latest stable release from
+[npm](https://www.npmjs.com/):
+
+```sh
 sudo npm install -g thelounge
```
williamboman: Might as well remove `sudo` and mention the need for root privileges on certain systems?

astorije: Good point. Most often, people will document `[sudo] npm install -g thelounge` to indicate reader may need `sudo` or not depending on system. If this is not the first reader installs something on their machine, and it's extremely likely, they'll know what to do.
Will update this accordingly, thanks!

astorije: Done.

PR #626, opened 2016-09-21
Fix "display: flex" to work on iOS 8

### LINKED ISSUE #624: Can't see channel textarea on mobile
I've installed and run the Lounge for the first time. It works great on my desktop browser, but all I can see is the input box on my phone:

image

I'm running the Lounge v2.0.0-rc.1 and viewing it in Safari on an iPhone 6 running iOS 8.1.2 (I know, I know…)

Gilles123:
This fixes #624.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -55,7 +55,9 @@
 		"number-leading-zero": "never",
 		"number-no-trailing-zeros": true,
 		"length-zero-no-unit": true,
-		"declaration-block-no-duplicate-properties": true,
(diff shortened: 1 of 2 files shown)


xPaw:
Prefixed properties should be set before non prefixed ones, not after.

astorije:
Prefixed values are now allowed repo-wide to allow for Flexbox to run on iOS 8 :)

williamboman:
Autoprefixer would help avoid this.

PR #628, opened 2016-09-22
Move uglify invocation into npm scripts and remove grunt

nornagon:
Unfortunately the uglifyjs command line ends up being a little unwieldy, but that's a small price to pay for dropping grunt!

If the 'watch' functionality is useful to folks, I can add one of the options from https://www.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/#watch


nornagon:
Oh dear, windows' lack of globbing strikes again. I'll figure something out.

nornagon:
Some notes:
- Windows doesn't support globbing (`client/js/libs/*.js` doesn't expand to the list of js files in windows shell)
- Most nodejs libraries seem to solve this by adding a dependency on [glob](https://github.com/isaacs/node-glob) (which npm itself depends on), and parsing their cmd-line with it
- UglifyJS is a unique snowflake and implements their [own version](https://github.com/mishoo/UglifyJS2/blob/dcdcfe4d39e8769e8d7504f167610883e250d266/tools/node.js#L270) of globbing, which will only glob-expand once per argument, so `foo/*/*.js` doesn't work.
- Ultimately I think these `libs/` should be either moved to dependencies installed via npm, included in the main bundle by `import`, or removed altogether.
- Uglify can then be part of the webpack build process, and this npm script will go away.

So I think it's OK to explicitly list out the subdirectories for now.

package.json
```diff
@@ -17,9 +17,9 @@
   "scripts": {
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha -r test/fixtures/env.js test/**/*.js",
     "start": "node index",
-    "build": "npm run build:font-awesome && npm run build:grunt && npm run build:handlebars",
+    "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",
-    "build:grunt": "grunt",
+    "build:libs": "uglifyjs client/js/libs/*.js client/js/libs/jquery/*.js client/js/libs/handlebars/*.js -o client/js/libs.min.js --source-map client/js/libs.min.js.map --source-map-url libs.min.js.map -p relative",
```
astorije: I know you tried `client/js/libs/*/*.js` and that didn't work, but have you tried `client/js/libs/**/*.js` by any chance?

package.json
```diff
@@ -17,9 +17,9 @@
   "scripts": {
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha -r test/fixtures/env.js test/**/*.js",
     "start": "node index",
-    "build": "npm run build:font-awesome && npm run build:grunt && npm run build:handlebars",
+    "build": "npm-run-all build:*",
```
astorije: I am 99% sure we didn't use `npm-run-all` for a reason here. I think it was that we didn't want to use it too much, hoping to get rid of it sooner or later, but oh well. It's pretty cool that it supports `*`...

package.json
astorije: Alright, stupid question: do we *want* to uglify these? I don't think it matters too much in terms of performance.
EDIT: I also realize it makes things much simpler to load from the HTML file, so feel free to ignore if stupid comment :)

astorije:
First of, amazing, I am so happy someone tackled this!

> - Windows doesn't support globbing (`client/js/libs/*.js` doesn't expand to the list of js files in windows shell)

Guess what, I think we added AppVeyor right after the last mess up there as well :D

> - Ultimately I think these `libs/` should be either moved to dependencies installed via npm, included in the main bundle by `import`, or removed altogether.
> - Uglify can then be part of the webpack build process, and this npm script will go away.

Agreed.

> So I think it's OK to explicitly list out the subdirectories for now.

Added a comment to try with `**`.
EDIT: Their `simple_glob` function, which really should be called `simple_glob_placebo`, doesn't seem to support `**` either, sigh...
So yeah, having the libs in the repo was always a non-permanent solution, and we've made progress on that, so I'm happy with the current solution unless we have a better idea.

nornagon:
** doesn't even work on OSX, because npm scripts are executed with sh, not
bash, and sh doesn't support ** :(

package.json
nornagon: npm-run-all seems like a reasonable dependency at this point, but I can imagine simplifying things to a point where we can just use `&&` (or `&` for parallel) and drop a dependency. It's simple enough for now though, I think.

package.json
nornagon: I don't think there's a particularly good reason to uglify them other than for ease of loading, but webpack can fill this role once #609 lands and then we can drop uglify.

xPaw:
Perhaps include uglifyjs in your webpack PR? Then the glob issues will be gone, and simplify it.

nornagon:
This _is_ cross-platform (appveyor passed, as did the OSX + Linux travis builds). Folding it into the webpack PR is cleaner, though, I agree :)

nornagon:
I looked into folding this into #609, but ran into some problems. The libraries we include via the uglified libs.min.js are not CommonJS modules; they simply inject themselves into the global namespace. Webpack works by following `require()` calls, so it won't work with the libs unless we convert them all to use `require`/`import`.

IMHO this is still a win for now because we get to drop Grunt. As we further refactor the client we can shift stuff in libs/ over to CommonJS style and phase out uglify.

astorije:
> Not cross-platform

What's your issue with this, @xPaw? It seems like AppVeyor is happy and @nornagon kept cross-platformism in mind.

xPaw:
That's why:

> ERROR: can't read file: client/js/libs/*.js

I get that by running `npm build` in powershell.

xPaw:
Figured it out, had to run `npm install`...

👍

astorije:
Awesome stuff, thanks @nornagon!

PR #629, opened 2016-09-24
Remove -ms-transform and add missed -webkit-transform

xPaw:


astorije:
(Reviewing this PR regardless of existence of #630, but knowing it's out there)

@xPaw, we are using `-ms-` a few times already, I'm curious, why removing these?

xPaw:
@astorije Only IE9 needs `-ms-transform` and autoprefixer removes that rule too.

astorije:
Ok then, thanks for clarifying.

PR #630, opened 2016-09-24
Add postcss

xPaw:


package.json
```diff
@@ -17,10 +17,11 @@
   "scripts": {
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha -r test/fixtures/env.js test/**/*.js",
     "start": "node index",
-    "build": "npm run build:font-awesome && npm run build:grunt && npm run build:handlebars",
+    "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",
     "build:grunt": "grunt",
     "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
+    "build:postcss": "postcss -u postcss-import -u autoprefixer -u cssnano -o client/css/build.css client/css/style.css",
```
xPaw: Can be later used with webpack in #609

client/themes/morning.css
```diff
@@ -46,8 +46,22 @@ body {
...
 #chat .search::-webkit-input-placeholder {
-	color: #99a2b4;
+/* stylelint-enable */
+	color: #99a2b4 !important;
+	opacity: .5;
+}
+
+/* stylelint-disable */
+#chat .search::-moz-placeholder {
+/* stylelint-enable */
```
nornagon: It'd be nice to run postcss over these too, though I guess that's maybe more build tool work than it's worth.

Alternatively, perhaps we could have a different stylelint config for themes?

package.json
```diff
@@ -61,7 +62,9 @@
     "ldapjs": "1.0.0"
   },
   "devDependencies": {
+    "autoprefixer": "6.4.1",
     "chai": "3.5.0",
+    "cssnano": "3.7.5",
```
nornagon: Do we need minification of our CSS?

package.json
```diff
@@ -71,6 +74,8 @@
     "istanbul": "0.4.3",
     "mocha": "2.4.5",
     "npm-run-all": "2.1.1",
+    "postcss-cli": "2.6.0",
+    "postcss-import": "8.1.2",
```
nornagon: We don't have any `@import` rules in our CSS, perhaps we can drop this for now and add it when we need it?

xPaw: I added it because I think splitting the CSS into components is a good idea, and this will allow us to do that.

package.json
xPaw: We minify most of our JS, and adding webpack would minify `lounge.js` too. Might as well do the same for CSS. `cssnano` does a bunch of optimizations on top of just removing whitespace.

client/css/style.css
```diff
@@ -119,16 +119,7 @@ button {
...
-	opacity: 1;
-}
-
-::-webkit-input-placeholder {
-	color: rgba(0, 0, 0, .35);
-}
-
-:-ms-input-placeholder {
+::placeholder {
 	color: rgba(0, 0, 0, .35) !important;
```
astorije: Any chance we can get rid of `!important` as well?

xPaw: Probably, I don't know why it's being used.

client/themes/morning.css
```diff
@@ -46,8 +46,22 @@ body {
 	padding: 15px 16px;
 }
 
+/* stylelint-disable */
```
astorije: I'm confused, why do we have to add all these exceptions actually? It's kind of a shame :(

astorije:
So, one thing I'm really concerned about: do we now have to run `npm run buid:postcss` **every time** we make a single change to the CSS??

xPaw:
@astorije I know that sucks, but we already had to do when working with JS "libraries". If we merge the webpack PR, and add postcss stuff into it, then we can use the `watch` option which auto rebuilds.

astorije:
Yeah, but we very rarely change the libs so that's no big deal.
And yeah for the watch, so maybe we should wait for that to go in first, no? CSS is something we (I know I do) update extremely often, and I don't see myself running that every single time. I mean, with initiatives like React's Hot Reload, you don't even have to refresh the page to see your changes, so this seems like a step back for now.

Also, fairly unrelated but at my workplace we use SCSS, and we never have to run a command for that to compile to CSS or have a watch step. My guess is that (in dev mode only), whenever reloading, it does its stuff and you don't have to worry about it. This makes it for a very nice dev experience, are you aware of anything like this?

package.json
nornagon: webpack doesn't minify by default, it's just a fancy `cat` :)

xPaw: I meant we'd be using uglifyjs with it.

xPaw:
@astorije I assume your SCSS setup runs a `watch` option, similar to what webpack would do.

nornagon:
(Or possibly the web server auto-compiles it when you load.)

astorije:
> (Or possibly the web server auto-compiles it when you load.)

That, yes.

astorije:
> - `@import` inliner
>   - We can split our css into different component files for some structure

Why do we want this? `@import` is broadly supported, and performance overhead is not much, none if relying on HTTP/2 I believe. I think we should freely use `@import` and worry about inlining it if it ever becomes a problem.

package.json
astorije: Did you notice any performance improvement? I'd personally prefer not minify that, unless it really is a problem.

xPaw: It makes the file size smaller, see http://cssnano.co/

xPaw:
Removed `@import` inliner, changed `autoprefixer` to `cssnext`. Also added `normalize.css` and removed it from our `bootstrap.css` file.

package.json
astorije: I just played with cssnano a bit. It turns out our `style.css` is 30k, but minifying with `cssnano` makes it 23k.
Compared to our 289k `libs.min.js` and 55k lounge.templates.js`, I think this is a drop in the water. I honestly think we don't need this extra dependency + extra step and just keep the CSS as is until it becomes a problem.
What do you think?

xPaw: OK removed.

astorije:
> changed `autoprefixer` to `cssnext`

Just curious, what made you change it?

> Also added `normalize.css` and removed it from our `bootstrap.css` file.

We do have a custom Bootstrap, but the default one has `normalize` included. Shouldn't we keep it in there?
I know you proposed #638, I don't know what that entails us to, but if we end up keeping it, I'd rather keep `normalize` in there, and if we drop it, maybe then it would be best to use `postcss-normalize` (or something else, Bootrap v4 has `reboot.css` which seems slick, etc.).
What do you think of that?

astorije:
Apart from that, this PR is mostly ready I think. It will be good to merge that after #640 makes it! :)

xPaw:
> Just curious, what made you change it

CSS next includes autoprefixer and allows us to use the new cool CSS features without worrying about browser support.

> Shouldn't we keep it in there?

We only really use the bootstrap grid. Moving it allows us to update normalize.css independently.

client/css/style.css
```diff
@@ -36,6 +38,16 @@
 	font-style: normal;
 }
 
+html {
+	box-sizing: border-box;
```
astorije: Why this?

client/css/style.css
```diff
@@ -36,6 +38,16 @@
...
+}
+
+*,
+*:before,
+*:after {
+	box-sizing: inherit;
```
astorije: Same question here.

package.json
```diff
@@ -65,6 +66,10 @@
     "istanbul": "0.4.5",
     "mocha": "3.0.2",
     "npm-run-all": "3.1.0",
+    "postcss": "5.2.4",
+    "postcss-cli": "2.6.0",
+    "postcss-cssnext": "2.8.0",
+    "postcss-normalize": "0.3.0",
```
astorije: > Moving it allows us to update normalize.css independently.

+1

I'm wondering however if we should not cut the middleman and instead rely on `"normalize.css": "3.0.1"` and copying it ourselves like we do with Font Awesome.
`postcss-normalize` seems barely maintained, is on `normalize.css` v4 (which is neither convenient for us to stay with what we currently have, or to update to v5 later). It does seem like a bottleneck.
I don't know too much about PostCSS but it seems like we can just give it `normalize.css` directly from the `node_modules` dir, no?
What do you think of that solution overall?

astorije:
> Reboot is too opinionated for our purposes, imo.

Fair point. Plus it's not prod-ready at the moment, I believe.

Also, I am labeling this as `do not merge` until @nornagon comes back to finalize #640 and we merge it :)

williamboman:
The version bump of `normalizer` seems to have broken `#input`'s font color. `textarea` no longer has a `color: inherit` declaration (which makes the font dark on Zenburn, should be light). Would probably be best to stick with the previous version of normalizer (`3.0.1`).

client/css/style.css
astorije: I can't find the comment that this refers to so I'm guessing it was: Why do we need this here?!
This seems like an addition that was necessary after the normalize.css bump, but since it was brought back to 3.0.1, these lines should not be necessary anymore.

client/themes/morning.css
astorije: At some point, theme management should be nice enough that one can add a theme easily. When this happens, themes should not be linted on our end (except the default theme(s) it ships with) but dealt with at will by their authors.
In the meantime, this seems fine.

package.json
```diff
@@ -66,6 +67,10 @@
     "istanbul": "0.4.5",
     "mocha": "3.1.2",
     "npm-run-all": "3.1.1",
+    "postcss": "5.2.4",
+    "postcss-cli": "2.6.0",
+    "postcss-cssnext": "2.8.0",
+    "postcss-normalize": "0.3.0",
```
astorije: I am definitely having issues with this: `postcss-normalize` is now deprecated, we should not rely on it.
It is suggesting to use `postcss-import`, but at that point we can use `@import` just fine as it's supported in all browsers we run on (cf. previous comments about `postcss-import`).

But more generally, as @williamboman is pointing out, we should stay with `normalize.css` v3.0.1 to avoid breakage. Bumping `normalize.css` regardless of Bootstrap is asking for trouble IMO, and is very unlikely to bring value.

I think that, as long as we are using Bootstrap, we should stick with the version of `normalize.css` that's in there. When/If we get rid of Bootstrap, then I'll be all in favor of reconsidering this.

client/css/bootstrap.css
```diff
@@ -8,197 +8,6 @@
  * Generated using the Bootstrap Customizer (http://getbootstrap.com/customize/?id=242b77ebf84da0cd12cf)
  * Config saved to config.json and https://gist.github.com/242b77ebf84da0cd12cf
  */
-/*! normalize.css v3.0.1 | MIT License | git.io/normalize */
```
astorije: Per my comment below, I highly suggest we revert this until getting rid of Bootstrap as a whole, if that ever happens.

client/themes/morning.css
xPaw: Good theme management means user themes are stored outside of the lounge source (code) folder. And that by default means it won't be built/linted/whatever. Linting is a dev tool anyway.

xPaw:
@astorije, fixed all concerns. The PR is now simpler.

package.json
```diff
@@ -18,6 +18,7 @@
     "build:font-awesome": "node scripts/build-fontawesome.js",
     "build:libs": "uglifyjs client/js/libs/*.js client/js/libs/jquery/*.js client/js/libs/handlebars/*.js -o client/js/libs.min.js --source-map client/js/libs.min.js.map --source-map-url libs.min.js.map -p relative",
     "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
+    "build:postcss": "postcss -u postcss-cssnext -m file -o client/css/build.css client/css/style.css",
```
YaManicKill: We need this to be run in build, don't we? 

Otherwise, if you just run `npm run build`, this won't be run and we won't have any css.

YaManicKill: Oh, so it is. 

Alright, so are you waiting till webpack is merged before this gets done?

astorije:
@xPaw, do we still want this? Could you remind me what are the benefits of having this?

xPaw:
@astorije I think auto prefixer is still something we want.

xPaw:
Will close for now.

PR #632, opened 2016-09-24
Make use of multi-prefix cap and remove NAMES spam on mode changes

### LINKED ISSUE #476: Flooding other clients with NAMES lists
Whenever a channel mode is changed, it is requesting the NAMES list in that channel. This isn't an issue on lounge itself, but when I have it connected to my bouncer, I am seeing names list flooding on my other clients (especially when the channel has a lot of users). 

Most clients tend to update the users list with channel ranks as soon as they are spotted, and will periodically send out a WHO command to different channels to update the userlists manually, in case one was missed.

xPaw:
Fixes #476. This change no longer sends `NAMES` request to the server if the server supports `multi-prefix` capability.

### DIFF
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -2,13 +2,16 @@ var _ = require("lodash");
 
 module.exports = User;
 
-function User(attr) {
-	// TODO: Remove this
-	attr.name = attr.name || attr.nick;
-	attr.mode = attr.mode || (attr.modes && attr.modes[0]) || "";
-
+function User(attr, prefixLookup) {
 	_.merge(this, _.extend({
-		mode: "",
-		name: ""
+		modes: [],
+		nick: ""
 	}, attr));
+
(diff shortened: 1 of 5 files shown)


src/models/user.js
```diff
@@ -2,13 +2,18 @@ var _ = require("lodash");
...
-
+function User(attr, prefixLookup) {
 	_.merge(this, _.extend({
-		mode: "",
-		name: ""
+		modes: [],
+		nick: ""
 	}, attr));
+
+	// irc-framework sets characater mode, but lounge works with symbols
```
MaxLeiter: small nitpick: `characater ` = `character`?

xPaw:
This PR needs to sort the user list after updating modes.

xPaw:
Fixed sorting.

MaxLeiter:
Works fine for me 👍

PR #634, opened 2016-09-25
Override defaults via URL

dougluce:
This allows for specifying defaults via URL:

    http://example.com/?join=%23thelounge&nick=jimbob&host=irc.freenode.net

The `enableOverrides` config setting must be set to true before this feature will work.

The `es6` environment is selected so eslint doesn't complain about `Map` not existing. The init function now returns the Express instance so the added test can have it bind to an available port and use that for testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "lint": "npm-run-all -c lint:js lint:css",
     "lint:js": "npm-run-all -c lint:js:es5 lint:js:es6",
     "lint:js:es5": "eslint --parser-options=\"ecmaVersion:5\" client/",
-    "lint:js:es6": "eslint --env es6 --ignore-pattern client/ .",
+    "lint:js:es6": "eslint --ignore-pattern client/ .",
     "lint:css": "stylelint \"**/*.css\"",
     "prepublish": "npm run build"
   },
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "event-stream": "3.3.2",
     "express": "4.13.4",
     "fs-extra": "0.30.0",
-    "he": "^1.1.0",
+    "he": "1.1.0",
     "irc-framework": "2.5.0",
     "ldapjs": "1.0.0",
     "lodash": "4.11.2",
@@ -63,7 +63,6 @@
   },
   "devDependencies": {
     "chai": "3.5.0",
-    "cheerio": "^0.20.0",
     "eslint": "2.11.1",
     "font-awesome": "4.6.3",
     "grunt": "1.0.1",
@@ -73,7 +72,6 @@
     "istanbul": "0.4.3",
     "mocha": "2.4.5",
     "npm-run-all": "2.1.1",
-    "request": "^2.74.0",
     "stylelint": "6.6.0"
   }
 }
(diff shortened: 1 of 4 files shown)


MaxLeiter:
I really like the idea, but maybe we should hold off on ES6 syntax  (including =>) until #609 is merged?

package.json
```diff
@@ -50,18 +50,20 @@
     "event-stream": "3.3.2",
     "express": "4.13.4",
     "fs-extra": "0.30.0",
+    "he": "^1.1.0",
```
MaxLeiter: Should use a fixed version

MaxLeiter: Also, `he` is just an encoder, is it not possible for us to do this ourselves, instead of adding another dependency?

package.json
```diff
@@ -50,18 +50,20 @@
...
     "request": "2.74.0",
     "semver": "5.1.0",
     "socket.io": "1.4.5",
-    "spdy": "3.3.2",
-    "ldapjs": "1.0.0"
+    "spdy": "3.3.2"
   },
   "devDependencies": {
     "chai": "3.5.0",
+    "cheerio": "^0.20.0",
```
MaxLeiter: Fixed version, same as above

package.json
```diff
@@ -71,6 +73,7 @@
     "istanbul": "0.4.3",
     "mocha": "2.4.5",
     "npm-run-all": "2.1.1",
+    "request": "^2.74.0",
```
MaxLeiter: Same ^^

package.json
dougluce: My mistake, this shouldn't have been in there at all.

package.json
dougluce: Same ^^

package.json
dougluce: It's an encoder, but it's not particularly trivial. Especially given that this is a protective measure I'm far more comfortable relying on `he` than attempting to do it myself.

dougluce:
I stuck with what works under node 4.2.0 and avoided any ES6 in the non-test code. I can pull out the ES6 syntax bits but would note that at least one existing dependency (`irc-framework`) contains ES6 in its own test framework.

MaxLeiter:
More for consistencies sake then anything re: ES6

dougluce:
I tried this some time ago completely client side but ran into some troubles with the (perhaps then-current) architecture. That and the fact that I don't touch front-end tech too often led me down this path instead.

MaxLeiter:
So after some thought, I definitely believe this should be 100% clientside - are you willing to try again?

MaxLeiter:
bump, @dougluce? I'd really love to see this merged!

dougluce:
Sorry, this works for my own purposes and I can't currently muster the inclination to spend time re-implementing this on the client side. Please feel free to have a shot at it.

astorije:
> Sorry, this works for my own purposes and I can't currently muster the inclination to spend time re-implementing this on the client side.

Sadly, we can't take this as is and it looks like you won't be spending time on it. I am closing this in favor of @MaxLeiter's ongoing work on #674.
Let us know if you're planning to work on it again and/or help @MaxLeiter on his PR.

I must say though, great job on adding tests, and I hope we can find a way to port them on #674.

PR #636, opened 2016-09-25
Improve support for opening multiple clients at once

xPaw:
Improve support for opening multiple clients at once

- Synchronize unread counter with the server
- Fix unread marker on no attached clients #561
- Increase unread counter for server messages
- Keep track of all the open channels across all client sockets

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -184,6 +184,13 @@ $(function() {
 		}
 	});
 
+	socket.on("open", function(id) {
+		// Another client opened the channel, clear the unread counter
+		sidebar.find("[data-id='" + id + "'] .badge")
+			.removeClass("highlight")
+			.empty();
+	});
+
 	socket.on("join", function(data) {
 		var id = data.network;
 		var network = sidebar.find("#network-" + id);
@@ -359,7 +366,7 @@ $(function() {
 			.append(msg)
 			.trigger("msg", [
 				target,
-				data.msg
+				data
 			]);
 
 		if (data.msg.self) {
@@ -846,7 +853,6 @@ $(function() {
 		self.addClass("active")
 			.find(".badge")
 			.removeClass("highlight")
-			.data("count", 0)
 			.empty();
 
 		if (sidebar.find(".highlight").length === 0) {
@@ -955,6 +961,9 @@ $(function() {
 	});
 
 	chat.on("msg", ".messages", function(e, target, msg) {
+		var unread = msg.unread;
+		msg = msg.msg;
+
 		if (msg.self) {
 			return;
 		}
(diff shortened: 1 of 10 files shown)


xPaw:
> 08:48 astorije  Why doesn't this solve the problem?
09:08 +xPaw  Because all you do is mark which client reads which channel (same exact problem as  now), it will still get out of sync between channels and page reloads
09:10 +xPaw  I'm thinking unread counter has to be global,  but marking open channels per client (keep the last open channel for page reloads), then the desync problem should be solved
09:11 +xPaw  Then when one client opens a channel, tell other clients to mark it as read

client/js/lounge.js
```diff
@@ -170,6 +170,10 @@ $(function() {
 		$("#sign-in").remove();
 
 		var id = data.active;
+
+		// Set it right away before triggering a click so that "open" event is not emitted
```
xPaw: Maybe this isn't actually a good idea, as loading the page *should (?)* mark the channel as read.

MaxLeiter: Marking the page as read removes the unread marker, right? If so, I think this is fine and the page should be left as unread.

astorije: Yes, when loading the page, the active channel shouldn't be marked as read to give a chance to the user to catch up on new messages.

Actually, all that chain of PRs (#561, #613, this one, #635) was triggered by trying to not mark channel as read on initial load :D

astorije:
@xPaw, I'm unclear if you want to make changes or not, feel free to assign me to that PR when you think it's ready to go.

client/js/lounge.js
xPaw: It's actually quite silly this way. If you open the page and `#channel1` is loaded, switch to another channel and then reload the page, `#channel1` will display the previous unread counter as it won't be reset.

astorije:
Testing this, I noticed an out-of-sync with unread counters on multiple clients:

counter-not-synced

Note that `[12]` does not clear on second client until a new message gets posted.

client/js/lounge.js
```diff
@@ -181,6 +181,13 @@ $(function() {
 		}
 	});
 
+	socket.on("open", function(id) {
+		// Another client opened the channel, clear the unread counter
+		sidebar.find("[data-id='" + id + "'] .badge")
+			.removeClass("highlight")
+			.empty();
+	});
+
```
xPaw: What do you mean?

xPaw:
@astorije fixed.

client/js/lounge.js
astorije: I mean that this clears unread counter for channels opened in other clients, do we need to move the unread **marker** for these channels as well?

client/js/lounge.js
```diff
@@ -840,7 +847,6 @@ $(function() {
 		self.addClass("active")
 			.find(".badge")
 			.removeClass("highlight")
-			.data("count", 0)
```
astorije: Why are we getting rid of this?

xPaw:
There's no longer a need for it, as it only ever updates the badge from the
server now.

client/js/lounge.js
xPaw: There's no longer a need for it, as it only ever updates the badge from the
server now.

PR #637, opened 2016-09-25
Display extra loading messages

xPaw:


client/js/lounge.js
```diff
@@ -80,6 +80,10 @@ $(function() {
 		});
 	});
 
+	socket.on("authorized", function() {
+		$("#loading-page-message").text("Authorized, loading messages…");
```
xPaw: On slow networks it might take a while to receive the `init` packet containing all the network/channel/message data.

client/js/lounge.js
```diff
@@ -147,6 +151,8 @@ $(function() {
 	});
 
 	socket.on("init", function(data) {
+		$("#loading-page-message").text("Rendering…");
```
xPaw: On slow device it might take a while to render all the channels and messages at once.

xPaw:
Might also add a message on websocket stuff, as falling back to http polling night take a while too.

astorije:
This goes so fast I can't trigger the messages, even when throttling :D
I'll test this on a crappy device.

astorije:
I honestly can't see most of the loading messages even on phone but testing this it works ok so 👍 and merging.

However, I think we should revamp this a tiny bit to look nicer: I would prefer a loading will or bar or whatever, and if one of the steps stays stuck for more than several seconds, display a message about that specific test (like we do right now, but contextualized for the current step). Not a top priority feature, but a nice-to-have in terms of UX.

Good job though, thanks! :-)

PR #639, opened 2016-09-25
Update developer dependencies

xPaw:
Fixes npm warnings about security issues and `jade` rename. Also drops Node <4 support in updated packages.


MaxLeiter:
Works fine for me on Linux 4.4.0-36 (Ubuntu)

astorije:
@xPaw, there are quite a lot of changes in bumps of ESLint and Stylelint. Did you happen to check if there were new rules or options we might want to enforce?

xPaw:
I did not check for the new rules, but that can be done in a separate PR if there's anything interesting.

Issue #494, opened 2016-07-12
Drag'n'dropping channels wrongly positions the channel in flight


astorije:
Easier to explain with a GIF:

![drag-drop-channel-broken](https://cloud.githubusercontent.com/assets/113730/16757654/5ae7feae-47db-11e6-8e00-aa0264ccb02e.gif)

Note that:

- This happens on channels only, not networks.
- Reproducing this requires some playing on the drag'n'dropping, it looks good at first, then for no reason it breaks, and then it's consistently broken until next refresh.
- This happens both on `master` and `2.0.0-pre.4`, not sure about previous pre-releases and stable releases (if someone wants to try with other versions, that'd be helpful).

metsjeesus:
Its dependent from where you start dragging element. Problem is only if you drag from upper side on the area. Might be fixed with updating jquery.

metsjeesus:
It seems if you grab bottom side of the grabbing area, there are no problems, but if top side of the area, bug happens.

stepie22:
I belive #702 fixed this completely? Is so, this issue should be closed.

PR #640, opened 2016-09-25
Webpack all the things

nornagon:
Builds on top of #609. Drops uglify in favor of webpack. Uses code splitting and imports-loader to get everything to play nicely together.


astorije:
So, a word of caution: I believe some of the libs in `client/js/libs` have been modified in the repo (actually, I know some have diverged, but I don't see them in this diff). Did you check the potential differences when removing the files?

xPaw:
@astorije the removed libs were not modified by us

nornagon:
```
$ git log --stat client/js/libs/handlebars.js client/js/libs/jquery.js client/js/libs/jquery/jquery-ui.js client/js/libs/jquery/mousetrap.js client/js/libs/notification.js client/js/libs/uri.js
```
^- only contains large deltas, which suggests that those libraries were only ever added or updated.

webpack.config.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+	entry: "./client/js/lounge.js",
+	resolve: {
+		extensions: ["", ".js"]
```
astorije: Same question than #640 for the record: why do we need `""` here?

.gitignore
```diff
@@ -7,6 +7,5 @@ coverage/
 # See https://docs.npmjs.com/misc/scripts
 client/fonts/
 client/js/bundle.js
-client/js/libs.min.js.map
```
astorije: No `.map` anymore? Is that on purpose?

client/js/libs/handlebars/index.js
```diff
@@ -0,0 +1,10 @@
+import "./colorClass";
+import "./date";
+import "./diff";
+import "./equal";
+import "./modes";
+import "./parse";
+import "./roundBadgeNumber";
+import "./tojson";
+import "./tz";
+import "./users";
```
astorije: I sure wish there was a way to do `import "./*";` but apparently there is no such thing.

client/js/lounge.js
```diff
@@ -1,3 +1,15 @@
+import $ from "jquery";
+import "jquery-ui/ui/widgets/sortable";
+import Mousetrap from "mousetrap";
+import Handlebars from "handlebars/runtime";
+import "notification-polyfill";
```
astorije: Is that polyfill still necessary?

client/js/lounge.js
```diff
@@ -1,3 +1,15 @@
...
+
+import "./libs/jquery/inputhistory";
+import "./libs/jquery/tabcomplete";
+import "./libs/jquery/stickyscroll";
```
astorije: I believe all these libs by @erming are on npm, but also that we have monkey-patched these over time. A great contribution would be diff-ing ours with the upstream ones (because, yeah, they have evolved on the upstream project too...), open PRs on @erming's repos, and as they are being merged, move these to use the npm versions :-)

package.json
```diff
@@ -64,11 +63,16 @@
     "eslint": "3.6.0",
     "font-awesome": "4.6.3",
     "handlebars": "4.0.5",
+    "imports-loader": "0.6.5",
     "istanbul": "0.4.5",
+    "jquery": "2.1.1",
+    "jquery-ui": "1.12.1",
```
astorije: Version in repo is 1.11.1. Shouldn't we merge this PR with same versions and improve/test later on? That seems safer to me (like you did with jQuery, staying on 2.1.1 even though latest v2 is 2.2.4 and there is even a v3).

package.json
```diff
@@ -64,11 +63,16 @@
...
     "mocha": "3.0.2",
+    "mousetrap": "1.6.0",
```
astorije: Same comment, version in repo is 1.4.6.

package.json
```diff
@@ -64,11 +63,16 @@
...
+    "notification-polyfill": "1.0.0",
     "npm-run-all": "3.1.0",
     "stylelint": "7.3.1",
-    "uglify-js": "2.7.3",
+    "urijs": "1.18.1",
```
astorije: Version in repo is 1.14.1.

src/client.js
```diff
@@ -1,5 +1,5 @@
 var _ = require("lodash");
-var package = require("../package.json");
+var packageJson = require("../package.json");
```
astorije: https://github.com/thelounge/lounge/pull/399 did something similar with server code, but went with `pkg`. I have no preference over which one, but it would be nice to use the same one :-)

package.json
```diff
@@ -64,11 +63,16 @@
     "eslint": "3.6.0",
     "font-awesome": "4.6.3",
     "handlebars": "4.0.5",
+    "imports-loader": "0.6.5",
```
astorije: Not sure I'm getting this, why do we need it?

webpack.config.js
```diff
@@ -1,5 +1,17 @@
+const webpack = require("webpack");
```
astorije: Interesting, why do you need to require webpack in your second commit? It seems like it was not needed for the first one.

nornagon: Because this one uses `webpack` in its body, to refer to `webpack.optimize.CommonsChunkPlugin`.

webpack.config.js
nornagon: So it allows importing without the `".js"` extension in `import`.

.gitignore
nornagon: This webpack config doesn't produce source maps, but no reason it couldn't--I'll update it to produce inline source maps (so no need for an extra file).

client/js/libs/handlebars/index.js
nornagon: Nope, quite intentionally :) (That would cause an import cycle, for one thing!)

client/js/lounge.js
nornagon: http://caniuse.com/#feat=notifications hm, not sure.

client/js/lounge.js
nornagon: I intend to ultimately remove all of these libs as dependencies.

package.json
nornagon: To make `$` appear in libs without calling `require`.

package.json
nornagon: Good point, I missed this. Will fix.

package.json
nornagon: Good catch.

src/client.js
nornagon: Will do.

webpack.config.js
astorije: Duh yeah, I missed that (aka reviewing PRs at 1AM 😅).
Too bad we can't use `import` here though, the irony :-)

webpack.config.js
astorije: Aaah, makes sense! Do we want to allow both with and without the extension? Or force one or the other format?

client/js/libs/handlebars/index.js
astorije: Well, you could locate the loading file outside of the directory (Ruby-style). But yeah, whatever :-)

client/js/lounge.js
astorije: Ok maybe let's keep it for now then. Sigh...

package.json
astorije: Good point, didn't think of that. Clever way to do so too, cool!

astorije:
Let me know when you're done updating this :-)

.gitignore
xPaw: Do not inline the source map. Thats a lot of useless data to be included.

client/js/lounge.js
nornagon: Seems like IE still doesn't have notification support, so yeah.

package.json
nornagon: NPM only has back to 1.16.1 of this lib, so I've rolled back to that.

package.json
nornagon: NPM doesn't have 1.11.1 unfortunately, the closest are 1.10.5 and 1.12.0. The only thing we use from jquery-ui is sortable, so I'll check that it still works with 1.12.1 and stick with that if it works.

nornagon:
Ready for another look!

package.json
astorije: Did 1.12.1 work as expected? Did the version bump turn out to fix #494 by any miracle?

webpack.config.js
astorije: As per our discussion on IRC, could you remove that section altogether and rely on [the defaults](https://webpack.github.io/docs/configuration.html#resolve-extensions)?

package.json
```diff
@@ -16,14 +16,13 @@
     "start": "node index",
     "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",
-    "build:libs": "uglifyjs client/js/libs/*.js client/js/libs/jquery/*.js client/js/libs/handlebars/*.js -o client/js/libs.min.js --source-map client/js/libs.min.js.map --source-map-url libs.min.js.map -p relative",
     "build:handlebars": "handlebars client/views/ -e tpl -f client/js/lounge.templates.js",
+    "build:webpack": "webpack",
+    "watch": "webpack -w",
```
xPaw: Can we also get a script to automatically start the server and webpack's watch within a single command for development purposes?

package.json
xPaw: I think we should explicitly require the needed stuff without automagic.

src/client.js
```diff
@@ -1,5 +1,5 @@
 var _ = require("lodash");
-var package = require("../package.json");
+var pkg = require("../package.json");
```
xPaw: You touched `client.js` for no reason.

webpack.config.js
```diff
@@ -0,0 +1,46 @@
...
+
+module.exports = {
+	entry: {
```
xPaw: Just for future proofing, will it be easy to add our `postcss` stuff into this config, or it will have to be rewritten for an additional export or whatever?

client/js/lounge.js
xPaw: @astorije These "libraries" were so heavily modified that I would consider them part of Lounge's code and not vendor.

src/client.js
nornagon: This isn't for no reason. `package` is a reserved word in ES6, and webpack was refusing to compile this file without this rename.

webpack.config.js
nornagon: Yep, should be easy to add. See https://github.com/postcss/postcss-loader.

package.json
nornagon: It did work. I wasn't able to consistently reproduce #494 on either old or new jquery-ui, but i did get it to happen once or twice so probably isn't fixed.

webpack.config.js
nornagon: Yep, looks like that works fine!

package.json
nornagon: Added.

package.json
astorije: I don't think this is even possible or reasonable. Unless I am mistaken, I believe @nornagon is talking about real libs (like jquery-ui) and not our "libs" in `client/js/libs`.

Or maybe there is a different/better way to do so, I don't know. I think this is good enough for a first shot at it and we can improve later.

src/client.js
astorije: Also, if you add `"use strict";` at the top (and feel free to do so on files you changed as it's a really good thing to have), [`implements`, `interface`, `let`, `package`, `private`, `protected`, `public`, `static`, and `yield` become reserved keywords](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Paving_the_way_for_future_ECMAScript_versions).

package.json
```diff
@@ -14,16 +14,16 @@
   "scripts": {
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha -r test/fixtures/env.js test/**/*.js",
     "start": "node index",
+    "start-dev": "npm run watch & npm run start",
```
xPaw: This only runs the webpack watch and blocks, and then never runs the lounge itself.

xPaw:
There's a couple of things that need to be addressed:

`babel-core` has to be added to `package.json`:
```
ERROR in Cannot find module 'babel-core'
npm WARN babel-loader@6.2.5 requires a peer of babel-core@^6.0.0 but none was installed.
```

-

Vendor file contains inline source maps,and the file is absolutely huge (554kb!!), generate an external `.map`.

-
Source map not being generated for `bundle.js` (make it external too)

-
`inputhistory` requires jQuery, but the dependancy is not resolved: `Uncaught ReferenceError: jQuery is not defined`

-

`start-dev` only runs webpack and doesn't start the server.

-

I think uglifyjs should be kept to generate a lot smaller files (especially vendor). `bundle.js` drops at least half in size with uglify plugin.

xPaw:
Also, this needs to be rebased on master due to merge of slideout menu (new libs/slideout.js file)

astorije:
@nornagon, any plans to address @xPaw's comments soon? I would love to see this PR merged :))

package.json
nornagon: That's not how `&` works.

nornagon:
Added babel-core. Whoops.

---

Source maps inlined in vendor.bundle.js are there in the upstream code, it's not webpack that's adding them. The config in this diff wasn't generating source maps at all for our own code; I've tweaked it to generate them and put them in separate files. With this change:

```
$ du -sh client/js/*bundle*
112K   	client/js/bundle.js
300K   	client/js/bundle.js.map
560K   	client/js/vendor.bundle.js
1.4M   	client/js/vendor.bundle.js.map
```

I experimented with uglify and it did significantly drop the file sizes:

```
$ du -sh client/js/*bundle.js
 56K   	client/js/bundle.js
180K   	client/js/vendor.bundle.js
```

but at the cost of 2x build time:

```
$ time npm run build:uglify-webpack
[...]
npm run build:uglify-webpack  8.30s user 0.32s system 110% cpu 7.789 total
$ time npm run build:webpack
[...]
npm run build:webpack  4.05s user 0.26s system 114% cpu 3.771 total
```

so I don't think it makes sense to uglify during dev. Many projects using webpack have a separate config for prod vs. dev, and I'd be happy to work on that, but perhaps it could be in a followup PR?

---

I can't reproduce your issue with inputhistory. I tried building from a clean repo and I don't see any error. This is the code webpack is generating:

```javascript
	/*** IMPORTS FROM imports-loader ***/
	var jQuery = __webpack_require__(1);
	
	/*!
	 * inputhistory
	 * https://github.com/erming/inputhistory
	 * v0.3.1
	 */
	(function ($) {
		$.inputhistory = {};
		/* ... */
	})(jQuery);
```

which seems fine to me. In addition, the functionality (pressing up to get to old messages) works just fine.

---

I also can't reproduce your issue with `start-dev`, it works fine for me:

```
$ npm run start-dev

> thelounge@2.0.0 start-dev /Users/nornagon/Source/lounge
> npm run watch & npm run start


> thelounge@2.0.0 watch /Users/nornagon/Source/lounge
> webpack -w


> thelounge@2.0.0 start /Users/nornagon/Source/lounge
> node index

2016-10-09 22:05:04 [INFO] The Lounge v2.0.0 is now running on http://*:9000/ in public mode
2016-10-09 22:05:04 [INFO] Press ctrl-c to stop

Hash: 2e8283f667df32ce2724
Version: webpack 1.13.2
Time: 3009ms
               Asset    Size  Chunks             Chunk Names
           bundle.js  112 kB       0  [emitted]  app
    vendor.bundle.js  573 kB       1  [emitted]  vendor
       bundle.js.map  228 kB       0  [emitted]  app
vendor.bundle.js.map  688 kB       1  [emitted]  vendor
   [0] multi vendor 88 bytes {1} [built]
    + 50 hidden modules
```

nornagon:
Also rebased, and used `import` instead of side-effecting onto `window` to import the `slideoutMenu` function.

nornagon:
Just realised the start-dev issue might be a windows thing, are you testing on windows @xPaw ?

xPaw:
@nornagon yes windows

client/js/libs/handlebars/colorClass.js
```diff
@@ -1,5 +1,3 @@
-"use strict";
-
```
astorije: Oh well, I missed this (from http://eslint.org/docs/rules/strict):

> In ECMAScript modules, which always have strict mode semantics, the directives are unnecessary.

I'm wondering if ESLint's `strict` rule warns/errors if there is a `"use strict"` statement in these files, where there shouldn't be?

astorije:
> so I don't think it makes sense to uglify during dev. Many projects using webpack have a separate config for prod vs. dev, and I'd be happy to work on that, but perhaps it could be in a followup PR?

Agreed with all that. In dev env, priority should be given to shortest build time (4s is already a lot if needed for every save!!).
And definitely a further PR, this PR has already gone wild!

> I also can't reproduce your issue with start-dev, it works fine for me:

@xPaw, ok to ship this PR with `start-dev` broken on Windows and improve in a further PR? Windows devs can still run these 2 commands separately, and this PR already improve things a lot. It's becoming hard to review and I think (but might be wrong) that this PR is already working fine.

client/js/libs/handlebars/colorClass.js
nornagon: eslint complains that "use strict" isn't needed in modules when `parserOptions.sourceType = "module"`. I added a directory-specific eslintrc to client/ which sets that option so the rule doesn't get misapplied to all the non-webpack code (which isn't actually using ES6 modules). https://github.com/eslint/eslint/issues/4936 has some more context.

package.json
nornagon: `imports-loader` is still being used to provide the `jQuery` global variable to the jquery libs, which I haven't changed in this diff.

astorije:
Ok, tested this locally, everything works fine. @xPaw, let's 🚢 it, especially since #630 is relying on this one to be usable by devs.

nornagon:
BTW, incremental build time is much better than 4sec. The benchmark I showed was build-from-scratch.

nornagon:
Fixed `npm run start-dev` to work on windows (using `npm-run-all`, which we were already using for other things).

xPaw:
Looks fine to me, but I am currently unable to test it

astorije:
So this is ready to go, but I'll merge this after releasing v2.1.0 (which is almost ready) to give us some master time before releasing.

xPaw:
>     /*** IMPORTS FROM imports-loader ***/
    var jQuery = __webpack_require__(1);

I do not see this in my generated code, and it's still failing with missing jQuery.

nornagon:
Can you reproduce that issue with a fresh clone + npm install + npm run
build?

xPaw:
Yes, that's what I did the first time. I just tried again, and experience the same problems.

nornagon:
Bizarre. I can't repro on osx, lemme see if I can set up a Windows test
environment

xPaw:
Here's a diff on how I fixed both of the issues (and removed import-loader): https://gist.github.com/xPaw/5ec60ab09aa758c1a021b982f9d0b6f5

astorije:
@nornagon, @xPaw, I added @xPaw's diff to that PR to better see the changes and test them locally. Everything looks fine to me. Of course, we'll have that issue again if/when @nornagon syncs the jQuery libs between here and in erming's repos, but one step at a time.

@xPaw, you're OK with that PR now? If yes, please wait for @nornagon to come back and give his blessing since I added a commit to his PR directly.

xPaw:
Should be fine now.

astorije:
Alright, I tried to make the Handlebars helper test compatible with this Webpack PR and it does **not** work. I have tried many things, with no success.
If I the test calls the source file, it breaks because of the `import`.
If I generate the bundle file and call it in the helper test (terrible idea), it doesn't find the helper because it's properly scoped.

I am starting to be fairly depressed, especially after finding out how poor Webpack's documentation (and other related stuff) is.
Anyone has any idea?!

astorije:
Alright, I managed to make tests pass, but obviously what I did is a **horrible** thing and should absolutely not be merged.
However, am I onto something? Is there some kind of magic to load the bundle within tests so we can require the right chunk instead of a random id? Even if made a bit cleaner, my solution is very unlikely to be the right one.
Hoping @nornagon or @yofreke could help me with this!

astorije:
Alright, just made another attempt at this. It's breaking other places now, but it seems like it could go somewhere.
I'm unclear what the `handlebars`/`handlebars/runtime` thingy are about for now. If I keep the latter in the helper, test fails but project works. If I change it to `handlebars`, test passes but project breaks. And if I use `handlebars/runtime` in the test, test breaks at compile time. Argh.
Any input welcome 😅

xPaw:
@astorije, I fixed your issue entirely by compiling handlebars templates/helpers with webpack, and changing helpers to be their own es6 modules. Then changed the test to just test the module directly, avoiding handlebars.

.gitignore
```diff
@@ -3,9 +3,14 @@ npm-debug.log
 
 coverage/
 
+# Client-side testing builds
+.tmp/mocha-webpack/
```
xPaw: not needed

client/index.html
```diff
@@ -374,9 +374,8 @@ <h1 class="title">Settings</h1>
 	</div>
 
 	<script src="socket.io/socket.io.js"></script>
-	<script src="js/libs.min.js"></script>
-	<script src="js/lounge.templates.js"></script>
```
xPaw: Will need to figure out how to compile handlebars stuff into its own file.

client/js/libs/handlebars/tojson.js
```diff
@@ -1,7 +1,3 @@
-"use strict";
-
-Handlebars.registerHelper(
-	"toJSON", function(context) {
-		return JSON.stringify(context);
-	}
-);
+module.exports = function(context) {
+	return window.JSON.stringify(context);
+};
```
williamboman: This file needs to be renamed to `toJSON.js`, otherwise the handlebars loader will be unable to resolve it on case-sensitive file systems.

xPaw:
@astorije I've reverted changes to erming's libs so they are untouched by this PR. Also squashed it.

astorije:
> Also squashed it.

Nooooo, I was still trying to test some things, which are now gone as you've reverted them as part of one of the commits. Can you revert this, please? (`git reflog` for the win)

YaManicKill:
This will need rebased again. What's the plan with this, it would be great to have this, then we can start splitting up our monolithic codebase.

xPaw:
@YaManicKill astorije wanted something in this PR before merge.

astorije:
@xPaw, reminder to un-squash this for now please? :)

xPaw:
@astorije sorry I am not able to un-squash this.

YaManicKill:
One of you should have it on your reflog, and if not, the server should hav eit on the reflog, no?

MaxLeiter:
Can we not test as-is, because this is what will ultimately be merged?

astorije:
Uuuuh so you mean what I spent 1-2 days working on is lost? Please no!
I don't have the repo under which I worked on anymore, I was not expecting it to be squashed.
@xPaw, as @YaManicKill said, can you browse your reflog?

xPaw:
@astorije fortunately not lost! This should be it: https://github.com/nornagon/lounge/commits/291aa975c5a9847404b7fa928b3959773f2a5a7a

For reference: https://8thlight.com/blog/sandro-padin/2015/06/08/help-i-just-force-pushed-to-master.html

YaManicKill:
@astorije Are the things you were testing related to this PR, or are we happy to go with this now? Would be good to get this merged asap, so that other people building PRs can build off this rather than the old code.

YaManicKill:
@nornagon You've still got some lint errors. Could you fix these and rebase? Would be great to get this into a release soon.

PolarizedIons:
Trying to run this PR (via `./scripts/run_pr.sh 640`, though I did do `npm install` and `npm run build` manually just to be sure) and get this in the console when trying to load thelounge:
!()[http://i.imgur.com/Un5pLxF.jpg]

The page is also stuck on "Loading the app…"

xPaw:
I've pushed a fixed branch under our repo: https://github.com/thelounge/lounge/tree/pr/640

It's a lot easier to deal with the branch within the repo rather than dealing with the fork.

xPaw:
Closed in favour of fixed #817.

PR #641, opened 2016-09-26
Allow simple signup to the lounge

iamdey:
Closes https://github.com/thelounge/lounge/issues/914.

screen-to-gif

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -133,11 +133,11 @@
 								</label>
 							</div>
 							<div class="col-xs-12 error" style="display: none;">
-								Sign up failed. May be the username is already used or is invalid.
+								Sign up failed. This username is already in use or is invalid.
 								Please retry or contact an administrator.
 							</div>
 							<div class="col-xs-12 success" style="display: none;">
-								Successfully registred. You can <a href data-target="#sign-in">Sign in</a>.
+								Successfully registered. You can now <a href data-target="#sign-in">Sign in</a>.
 							</div>
 							<div class="col-xs-12">
 								<button type="submit" class="btn">
--- a/client/index.html
+++ b/client/index.html
@@ -35,6 +35,7 @@
 		</aside>
 		<footer id="footer">
 			<span class="tooltipped tooltipped-n" aria-label="Sign in"><button class="icon sign-in" data-target="#sign-in" aria-label="Sign in"></button></span>
+			<span class="tooltipped tooltipped-n" aria-label="Sign up"><button class="icon sign-up" data-target="#sign-up" aria-label="Sign up"></button></span>
 			<span class="tooltipped tooltipped-n" aria-label="Connect to network"><button class="icon connect" data-target="#connect" aria-label="Connect to network"></button></span>
 			<span class="tooltipped tooltipped-n" aria-label="Client settings"><button class="icon settings" data-target="#settings" aria-label="Client settings"></button></span>
 			<span class="tooltipped tooltipped-n" aria-label="Sign out"><button class="icon sign-out" id="sign-out" aria-label="Sign out"></button></span>
@@ -98,6 +99,53 @@
 								<button type="submit" class="btn">
 									Sign in
 								</button>
+								<% if (signUp) { %>
+								<a href data-target="#sign-up">
+									Sign up
+								</a>
+								<% } %>
+							</div>
+						</div>
+					</form>
+				</div>
+				<div id="sign-up" class="window">
+					<form class="container" method="post" action="">
+						<div class="row">
+							<div class="col-xs-12">
+								<h1 class="title">Sign up to The Lounge</h1>
+							</div>
+							<div class="col-xs-12">
+								<label>
+								Email
+								<input class="input" type="email" name="email" required="required">
+								</label>
+							</div>
+							<div class="col-xs-12">
+								<label>
+								Username
+								<input class="input" name="user" required="required">
+								</label>
+							</div>
+							<div class="col-xs-12">
+								<label class="port">
+								Password
+								<input class="input" type="password" name="password" required="required">
+								</label>
+							</div>
+							<div class="col-xs-12 error" style="display: none;">
+								Sign up failed. May be the username is already used or is invalid.
+								Please retry or contact an administrator.
+							</div>
+							<div class="col-xs-12 success" style="display: none;">
+								Successfully registred. You can <a href data-target="#sign-in">Sign in</a>.
+							</div>
+							<div class="col-xs-12">
+								<button type="submit" class="btn">
+									Sign up
+								</button>
+								<a href data-target="#sign-in">
+									Sign in
+								</a>
 							</div>
 						</div>
 					</form>
--- a/client/index.html
+++ b/client/index.html
@@ -100,9 +100,9 @@
 									Sign in
 								</button>
 								<% if (signUp) { %>
-								<a href data-target="#sign-up">
+								<button type="button" class="btn" data-target="#sign-up">
 									Sign up
-								</a>
+								</button>
 								<% } %>
 							</div>
 						</div>
@@ -137,9 +137,9 @@
 								<button type="submit" class="btn">
 									Sign up
 								</button>
-								<a href data-target="#sign-in">
+								<button type="button" class="btn" data-target="#sign-in">
 									Sign in
-								</a>
+								</buttonss>
 							</div>
 						</div>
 					</form>
--- a/client/index.html
+++ b/client/index.html
@@ -116,12 +116,6 @@
 							</div>
 							<div class="col-xs-12">
 								<label>
-								Email
-								<input class="input" type="email" name="email" required="required">
-								</label>
-							</div>
-							<div class="col-xs-12">
-								<label>
 								Username
 								<input class="input" name="user" required="required">
 								</label>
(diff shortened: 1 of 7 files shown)


iamdey:
may be email's requirement can be switch in the global config.

client/index.html
```diff
@@ -98,6 +99,53 @@ <h1 class="title">Sign in to The Lounge</h1>
...
+								<% } %>
+							</div>
+						</div>
+					</form>
+				</div>
+				<div id="sign-up" class="window">
+					<form class="container" method="post" action="">
+						<div class="row">
+							<div class="col-xs-12">
+								<h1 class="title">Sign up to The Lounge</h1>
```
MaxLeiter: `s/to/for/`?

client/index.html
```diff
@@ -98,6 +99,53 @@ <h1 class="title">Sign in to The Lounge</h1>
...
+							</div>
+							<div class="col-xs-12">
+								<label>
+								Email
```
MaxLeiter: Not sure how I feel about requiring email - especially if it has no use

client/index.html
```diff
@@ -98,6 +99,53 @@ <h1 class="title">Sign in to The Lounge</h1>
...
+								</label>
+							</div>
+							<div class="col-xs-12">
+								<label class="port">
+								Password
+								<input class="input" type="password" name="password" required="required">
+								</label>
+							</div>
+							<div class="col-xs-12 error" style="display: none;">
+								Sign up failed. May be the username is already used or is invalid.
```
MaxLeiter: `s/May be/maybe/`, or alternatively use something like `This username is already in use or is invalid`

client/index.html
```diff
@@ -98,6 +99,53 @@ <h1 class="title">Sign in to The Lounge</h1>
...
+								Please retry or contact an administrator.
+							</div>
+							<div class="col-xs-12 success" style="display: none;">
+								Successfully registred. You can <a href data-target="#sign-in">Sign in</a>.
```
MaxLeiter: `registred` should be `registered`. Maybe `You can now <a> sign in. </a>`

client/index.html
```diff
@@ -98,6 +99,53 @@ <h1 class="title">Sign in to The Lounge</h1>
...
+							</div>
+							<div class="col-xs-12">
+								<button type="submit" class="btn">
+									Sign up
+								</button>
+								<a href data-target="#sign-in">
```
MaxLeiter: Why is sign-up a` <button>`, and sign-in an `<a>`?

src/server.js
```diff
@@ -139,6 +139,37 @@ function init(socket, client) {
...
+				);
+			} catch (e) {
+				socket.emit("signed-up", {
+					error: "The username sounds invalid"
+				});
+			}
+
+			if (!addUserResult) {
+				socket.emit("signed-up", {
+					error: "Username sounds already in use"
```
MaxLeiter: sounds doesn't make much sense,  maybe `is`?

astorije:
That's a great idea, thanks for the contribution!

Just to let you know though, it still needs improvements (but we are here to guide you with that, here or on the IRC channel 😄), and it will definitely raise some questions!
Questions such as @MaxLeiter's ones, or design decisions regarding the feature itself, or, in a PR like that, about security.

My first question is about security: with this PR, it seems very easy to DDoS a server by generating millions of new users through the UI. Do you have any protection against this in mind?
I see you are requiring email. It is often used as a method to prevent this, but it's not enough, and you are not using it yet.

Thoughts from anyone else?

client/index.html
iamdey: Yeah, let's talk about that.

client/index.html
iamdey: `a` is (almost) a real link to sign-in form. It's a, UX/design  choice.

Nevertheless, I forgot to add a screenshot to the PR.

src/server.js
iamdey: I also change `The username is invalid`

iamdey:
(just added a gif on the first comment)

Here is my though about the security and email questions:

_agree with @astorije there's a security issue that can be using an validation_

So it requires a certain amount of work:

* an email template
* a nodejs library to manage email pool
* a system library to send email
* a server route to verify account registration
* a toggle to enable user

client/index.html
MaxLeiter: I think both should be buttons for consistency.

astorije:
> So it requires a certain amount of work:

Yeah, that's one way to do it, but not the only one. What about the following:
- When a user creates an account (username + password only), it creates a configuration file with `"active": false`.
- Admins can use `lounge list --inactive` to... list inactive users.
- Admins can activate users with `lounge activate <username>`.

To mitigate with DDoS, we can either delete files > 24h old (but that doesn't delete that risk) or delete when there are more than N inactive users (not ideal either as it's an easy way to prevent new users to subscribe, just make sure to create N inactive users quickly and all pending will be deleted.

These don't look ideal, but neither is email validation. I actually never had to implement something like this in the past so maybe others will have better ideas.

dgw:
Is there enough connection monitoring already happening that The Lounge could limit the number of new accounts created within x time window per IP address?

I also wonder about the usefulness of requiring an email address without verifying it. Spambots can and do generate random addresses all the time—as I know from the DMARC reports on my SPF `-all` ("sends no mail") domains.

astorije:
> Is there enough connection monitoring already happening that The Lounge could limit the number of new accounts created within x time window per IP address?

Good question, but really, if you can fake an email address, you can fake an IP just as easily :-/
Also, if we limit to, say 10 addresses per hour, we open the risk to prevent teams (like startups) or services (SaaS) to switch to The Lounge without a hiccup.

Thinking about it again, I think it's decent to not offer any measure to mitigate that ourselves, as I'm confident we have other vectors for DDoS attacks on the app...
It's actually very good that you can disable sign up from the admin side. That way, personal users hosting on a Raspberry Pi behind a home connection can disable this, and bigger organizations can enable deal with this at the network level, firewall, etc. We add to this a fail2ban recipe and we really are being defensive and proactive about this.
So, sorry I brought that up, but I think we don't want to worry about this at least for now. And until we know security risks are gone, we make this opt-in as you did.

> I also wonder about the usefulness of requiring an email address without verifying it.

I agree. If we were to send email notifications, or password recovery, or other features enhanced with email support, then that would be a great addition, but right now it's fairly useless.

@iamdey, do you mind removing email altogether from your PR? :)

client/index.html
iamdey: TL;DR: I'm not agree with you,  but I'll do what you say.

---
it's not about consistency, it's about UX. 

A link is not a button and conversely.
I mean it's about [affordance](https://en.wikipedia.org/wiki/Affordance) a button should be used to send an event (submit a form for example) and link to anchor to an other place on the app. If I change the link to a button, we will also have a problem with proximity and similarity laws (cf. [gelstalt theory](https://en.wikipedia.org/wiki/Gestalt_psychology#Pr.C3.A4gnanz).

iamdey:
* [x] email removed
* [x] the link to sign in/sign up is now a button (but as expected, it is very confusing)

lounge

MaxLeiter:
Apologies, now I see why you were against two buttons. Any idea which is better, @astorije?

astorije:
Good discussion and great considerations. I'll play with it when I have a
laptop on hand but yeah looking at the screenshot, that's a bit confusing
as is.
Thanks for your efforts and patience!

thisisdarshan:
@iamdey great work!  Can you add the support to LDAP mode as well?  Thanks!

maxpoulin64:
Sign-up probably should have a validation password field too. As much as we think we typed our password correctly, typos happens.

I think I preferred the version with a link. Maybe just put it below the button instead of beside, and write something like `Don't have an account yet? Sign up!`.

src/server.js
```diff
@@ -139,6 +139,35 @@ function init(socket, client) {
 	if (!client) {
 		socket.emit("auth", {success: true});
 		socket.on("auth", auth);
+		socket.on("sign-up", (data) => {
```
xPaw: This event should only be bound if sign up is enabled in the config, otherwise it's technically possible to create users.

client/index.html
MaxLeiter: I've never seen `sign in to`

MaxLeiter:
Hey @iamdey - are you still able to work on this? It would definitely be a nice feature

client/index.html
```diff
@@ -35,6 +35,7 @@
 		</aside>
 		<footer id="footer">
 			<span class="tooltipped tooltipped-n" aria-label="Sign in"><button class="icon sign-in" data-target="#sign-in" aria-label="Sign in"></button></span>
+			<span class="tooltipped tooltipped-n" aria-label="Sign up"><button class="icon sign-up" data-target="#sign-up" aria-label="Sign up"></button></span>
```
PolarizedIons: Since you only show the `sign up` button if enabled (on line 102), it should probably also be done here

astorije:
Hey @iamdey, I know this has been a while, but any chance you could resume work on this?

iamdey:
@astorije, @MaxLeiter Sorry, I was (and i'm still) very busy these days on [my main project](/caliopen/caliopen). And finally I don't use this feature so I don't know if I will be able to provide support if it is merged.

astorije:
Okay, then I guess we will close this for now 😢.

@thelounge/maintainers, @dgw, @thisisdarshan, @PolarizedIons, @theSoenke, does anyone want to pick this up and bring it to completion? I don't know how much work would be required on this but it would not be impossible to take it to v2.3.0 I think.

iamdey:
yes it is the better thing to do, thank you @astorije

PR #652, opened 2016-09-28
Themes: Fixed CSS rule selectors for highlight messages

DamonGant:


MaxLeiter:
Works for me, although im not 100% sure what it's fixing

DamonGant:
It seems somewhere along the way, the message container was changed to have a `.text.` element. The default CSS sets the color to `#f00` specifically to to the `.text` element, which overrides the theme-set color. This rule fixed that issue.

http://i.imgur.com/0Mbl8wo.png

MaxLeiter:
:+1:  Nice catch, I'd always thought it was on purpose and just a poor design choice and forgot about it.

astorije:
Good catch indeed, thanks!

(Outside of scope of this PR) To be noted that the line `#chat .channel .highlight .from` has no effect when nick colors are enabled...

PR #653, opened 2016-09-28
Fix unhandled message color in Crypto theme

MaxLeiter:
| Before                            | After                            |
|-----------------------------------|----------------------------------|
| before | after |

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -949,10 +949,6 @@ button {
 	color: #f00;
 }
 
-#chat .unhandled .from {
-	color: #eee;
-}
-
 #chat .msg.toggle .time {
 	visibility: hidden;
 }
--- a/client/themes/crypto.css
+++ b/client/themes/crypto.css
@@ -85,6 +85,10 @@ a:hover,
 	color: #666;
 }
 
+#chat .unhandled .from {
+	color: #ddd;
+}
+
 #sidebar .active {
 	color: #fff;
 }
--- a/client/themes/example.css
+++ b/client/themes/example.css
@@ -27,6 +27,10 @@ body {
 	top: 4px;
 }
 
+#chat .unhandled .from {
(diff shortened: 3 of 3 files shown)


astorije:
This is inherited from [here](https://github.com/thelounge/lounge/blob/4541309988c5fee292eeaf41c76d9bbdc899e1c7/client/css/style.css#L952-L954) and I believe it's on purpose that unhandled numerics are being dimmed out.
But I'm actually curious, what made you feel this needed fixing?

MaxLeiter:
At first I was unable to see the numbers, then realized they were there, just incredibly dim. Felt sloppy/unintentional to me for them to be "kind of" visible.

astorije:
Right, but note that this is true for the default theme as well, which I think needs fixing too.
What do you say we instead:

- Put same color than the timestamp for both Example (we really need a name lol, once we derive from the Shout color!) and Crypto
- Move that design element from `style.css` to `example.css` to make themes independent from default stylesheet, in the continuity of #615 and #359 ?
- Make sure Morning and Zenburn are still OK

MaxLeiter:
Done, @astorije. Both morning and zenburn already handle unhandled messages:

[Morning](https://github.com/thelounge/lounge/blob/master/client/themes/morning.css#L187) and [Zenburn](https://github.com/thelounge/lounge/blob/master/client/themes/zenburn.css#L213)

xPaw:
Could this be a muted color, not completely black? It's slightly confusing as it might seem like a normal nickname, maybe we need to add an icon before the numeric?

astorije:
@MaxLeiter, could you make it `#ddd` [like `#chat .time`](https://github.com/thelounge/lounge/pull/653/files#diff-97db1f70168fb5f12457b238ff6052b5L827)?

MaxLeiter:
Done

Issue #371, opened 2016-05-31
Support cp1251 encoding


Arkadietz:
Hello, 

It will be possible client to support cp1251? Will be nice if there is an option in conf where you can set whatever encoding you want.

maxpoulin64:
Setting an option in the conf wouldn't work very well, as that would force that encoding to every networks on the entire server. I would probably need to be set per-server along with the other options. In order to support this we will need:

* An additional setting in the connection page
* Make sure said setting is persisted to the network config
* Probably offer a drop-down menu to select from the supported encodings.

Good news is, irc-framework uses iconv so it should be just a matter of passing the variable to it.

I'm surprised there's still networks out there not using UTF-8 in 2016. Most clients/software in general have long gotten rid of encoding options in favor of the universal UTF-8.

dgw:
Would it be too far out of scope for The Lounge to try to auto-detect the text encoding for each network? @astorije I know you, especially, love to go for ease of use and minimal configuration whenever possible.

> Most clients/software in general have long gotten rid of encoding options in favor of the universal UTF-8.

Don't forget about clients that pretend to support UTF-8 but mangle it, or only support it with specific fonts, etc. 😉

maxpoulin64:
@dgw: The problem with trying to autodetect the encoding is that it tends to fail a lot and be very unpredictable. I think the server doesn't bother at all, so you will very likely end up with both encodings in the same channel.

I think the best solution would probably be to try to do the same as ZNC (which can be used as a temporary workaround until this is fixed), and try parsing as both the user-specified encoding and UTF-8.

The only problem I see with that is that there will be people needlessly using a non-UTF-8 encodings just for the sake of it "because I speak language X" and then wonder why it's always broken (and thus said encoding just won't ever die).

apihlaja:
> I'm surprised there's still networks out there not using UTF-8 in 2016. Most clients/software in general have long gotten rid of encoding options in favor of the universal UTF-8.

It's not network setting, it's per channel thing. After testing lounge first time yesterday, I was surprised how many people are not using UTF-8. Considering that no one is complaining on channels, I'd say most of IRC clients use some kind of double encoding solution. That's what my old client (irssi) did: by default, it uses UTF-8 but you can spesify one fallback encoding.

I guess one app level fallback encoding would be sufficient solution for 99.9 % use cases.

DanielOaks:
[...]

PR #656, opened 2016-09-29
#371 better encoding support

metsjeesus:
I just explain problem first.
I got situation where irc server sends both utf8 and iso-8859-1 messages. And i want both to be human readable. Problem is, if i use utf8, i cant see or recover iso-8859-1 messages. But if i use iso-8859-1 as encoding, i cant see utf8 messages but i can recover them.  Using iso-8859-1 makes other problem, if outgoing messages are coded on iso-8859-1, you cant send utf8 text correctly like `Iñtërnâtiônàlizætiøn☃💩'`

So, to fix it, you need different encodings to irc incoming messages and outgoing messages. And top of that, you need to get rid Mojibake, miscoded characters.  My solution is backward compatible encoding string in irc-framework configuration. So you can configure it in lounge for every network you connect.

`utf8`- normal way
`iso-8859-1/utf8`- incoming messages are coded in iso-8859-1, outgoing in utf8
`iso-8859-1:auto`- incoming and outgoing is iso-8859-1 and gets rid of mojibikes on incoming messages
`iso-8859-1:auto/utf8`- incoming messags are coded iso-8859-1 and gets rid of mojibikes, outgoing messages are coded utf8

autocorrect reqursively makes text human readable. That means certain text combinations, like `Ã¤` automatcly will be converted to `ä` and text will lose that information. For humans its good, but not for computers. 

;tldr made thing that fixes utf8 - iso-8859-1 Mojibake characters in node.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -196,6 +196,24 @@ module.exports = {
 		port: 6697,
 
 		//
+		// Encoding
+		//
+		// Format: <inputencoding>[:auto][/<outputencoding>]
+		// inputencoding iconv-lite supported encoding string
+		// outputcoding iconv-lite supported encoding string, default inputencoding
+		// :auto autoconverts miscoded mojibike characters
+		// Examples:
+		// iso-8859-1
+		// iso-8859-1:auto
+		// iso-8859-1/utf8
+		// iso-8859-1:auto/utf8
+		//
+		// @type     string
+		// @default  utf8
+		//
+		encoding: "utf8",
+
+		//
 		// Password
 		//
 		// @type     string
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "event-stream": "3.3.2",
     "express": "4.13.4",
     "fs-extra": "0.30.0",
-    "irc-framework": "2.5.0",
+    "irc-framework": "git://github.com/metsjeesus/irc-framework.git#encoding",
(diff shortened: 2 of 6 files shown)


defaults/config.js
```diff
@@ -194,7 +194,15 @@ module.exports = {
 		// @default  6697
 		//
 		port: 6697,
-
+		
+		//
+		// Encoding
+		//
+		// @type     string
```
MaxLeiter: Could you list the examples of possible encodings here? And the proper formatting for them?

package.json
```diff
@@ -50,7 +50,7 @@
     "event-stream": "3.3.2",
     "express": "4.13.4",
     "fs-extra": "0.30.0",
-    "irc-framework": "2.5.0",
+    "irc-framework": "git://github.com/metsjeesus/irc-framework.git#encoding",
```
MaxLeiter: I think it's worth waiting for the merge upstream, rather than switching to a fork

metsjeesus: yes, its worth waiting, but its only way to test it though.

defaults/config.js
metsjeesus: done and done

xPaw:
I will close this PR for now as it won't work without irc-fw changes, and seeing the discussion doesn't make it seem like it will happen any time soon.

PR #258, opened 2016-04-16
Implement /list

maxpoulin64:
This re-adds @xPaw's `/list` implementation, with additional styling to make it look nice.

capture d ecran_2016-04-16_08-33-26

PR #657, opened 2016-09-30
Fix title icons for channels and channel lists

astorije:
This bug slipped when adding `/list` support in #258.

### Results

Before | After
--- | ---
screen shot 2016-09-30 at 01 53 15 | screen shot 2016-09-30 at 01 51 51
screen shot 2016-09-30 at 01 53 24 | screen shot 2016-09-30 at 01 52 06

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -185,7 +185,7 @@ button {
 #chat .channel .title:before { content: "\f0f6"; /* http://fontawesome.io/icon/file-text-o/ */ }
 
 #sidebar .chan.special:before,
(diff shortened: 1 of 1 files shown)


xPaw:
Do you mind fixing the icon alignment?

astorije:
Definitely not in this PR as I will be mostly AFK for a week while this PR fixes a clear regression.

I remember looking at icon alignment here and it was weird. I can take another look in 1 or 2 weeks maybe, but if someone else wants to give it a shot, that would be great.

PR #658, opened 2016-09-30
Display wallops in server window

### LINKED ISSUE #225: Display wallops
irc-fw has a callback for it.

xPaw:
Quickly fixes #225, better than nothing.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -25,6 +25,12 @@ module.exports = function(irc, network) {
 		handleMessage(data);
 	});
 
+	irc.on("wallops", function(data) {
+		data.from_server = true;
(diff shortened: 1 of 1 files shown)


astorije:
I don't know much about `wallops` but code seems fine.
@maxpoulin64, could you test it on your IRC server?

Also, naive question, what's the difference between `wallops` and the notices we receive on server window, for example when Freenode warns us of a server reboot, DDoS summary, etc.?

astorije:
Regarding my question above, is this a wallop?

screen shot 2016-10-07 at 10 04 04

If not, what difference with it?

xPaw:
@astorije That's just a global notice, otherwise you would not see it (because we don't handle wallops, duh!)

PR #661, opened 2016-10-01
Do not trigger a DOM event on every message

xPaw:
This seemed like an odd solution, a function call is more direct.


xPaw:
@astorije not at the moment, as it involves calling a function in another scope.

PR #663, opened 2016-10-02
[WIP] Load previous messages from the logs

### LINKED ISSUE #1392: Duplicate log entries when using two nicks
```
```

astorije:
This is highly WIP. I'm not sure how successful it will be, but it's sad we have (almost) everything in the logs and we can't read from them when we restart the app (I know, I know, <insert here your favorite discussion about database>, but this is really just an attempt to read from what we have at minimal cost.
### TODO

I'm sure there are many TODOs before I could call this remotely reviewable, so I'll be listing them here to keep track of stuff.
- [x] Replay previous messages:
  - [x] At startup
  - [x] When joining a channel
  - [x] When clicking on a nick (`/whois`)
  - [x] When querying someone (`/query`)
  - [x] When receiving a message from someone that opens a new window
- [ ] Display all types of messages:
  - [x] Messages
  - [x] Actions
  - [x] Joins
  - [x] Modes
  - [x] Nick changes
  - [x] Parts
  - [x] Quits
  - [x] Topics
  - [ ] Others
- [x] Use the history limit to not load everything from files
- [x] Display actual timestamp instead of time when logs were read
- [ ] Clean up that uuuugly repeated code for parsing messages...
- [x] Go async
- [ ] Go stream
- [ ] Figure out why this branch is banning me from Freenode...
- [ ] Fix #1392

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -5,7 +5,7 @@ var colors = require("colors/safe");
 var pkg = require("../package.json");
 var Chan = require("./models/chan");
 var crypto = require("crypto");
-var userLog = require("./userLog");
+const UserLog = require("./userLog");
 var Msg = require("./models/msg");
 var Network = require("./models/network");
 var ircFramework = require("irc-framework");
@@ -91,6 +91,10 @@ function Client(manager, name, config) {
 		delay += 1000;
 	});
 
+	if (this.config.log === true) {
+		this.userLog = new UserLog(client.name);
+	}
+
 	if (typeof client.config.sessions !== "object") {
 		client.config.sessions = {};
 	}
@@ -122,8 +126,7 @@ Client.prototype.emit = function(event, data) {
 				if (target.chan.type === Chan.Type.LOBBY) {
 					chan = target.network.host;
 				}
-				userLog.write(
-					this.name,
+				this.userLog.write(
 					target.network.host,
 					chan,
 					data.msg
@@ -178,8 +181,8 @@ Client.prototype.connect = function(args) {
 
 			channels.push(channel);
 
-			userLog
-				.read(client.name, args.host, channel.name)
+			this.userLog
+				.read(args.host, channel.name)
 				.forEach((message) => channel.pushMessage(client, message));
 		});
 
--- a/src/client.js
+++ b/src/client.js
@@ -171,10 +171,16 @@ Client.prototype.connect = function(args) {
 				return;
 			}
 
-			channels.push(new Chan({
+			const channel = new Chan({
 				name: chan.name,
 				key: chan.key || "",
-			}));
+			});
+
+			channels.push(channel);
+
+			userLog
+				.read(client.name, args.host, channel.name)
+				.forEach((message) => channel.pushMessage(client, message));
 		});
 
 		if (badName && client.name) {
--- a/src/client.js
+++ b/src/client.js
@@ -91,9 +91,7 @@ function Client(manager, name, config) {
 		delay += 1000;
 	});
 
-	if (this.config.log === true) {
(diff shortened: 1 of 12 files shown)


xPaw:
> Load previous messages at startup
 Load previous messages at join
 Load previous messages when starting a private discussion

All these can be fixed by doing log reading in `Chan` constructor, after #664 is fixed.

astorije:
> All these can be fixed by doing log reading in Chan

Yeah, but I had no access to thing like username in there, so I put it where it is right now for convenience. Most likely not a definitive solution.
Also, putting it in the constructor means backlog loading is done all the time by default, which may not be what we want? I'm thinking if we need to instantiate an empty Chan, or in tests, etc. I don't know, I'm still playing with this :)

src/userLog.js
```diff
@@ -1,7 +1,49 @@
+"use strict";
+
 var fs = require("fs");
 var fsextra = require("fs-extra");
 var moment = require("moment");
 var Helper = require("./helper");
+const Msg = require("./models/msg");
+
+module.exports.read = function(user, network, chan) {
+	var data = fs.readFileSync(
```
xPaw: Loading logs from disk should be async.

astorije: Agreed. So far this has only been a POC so I did not care for early optimization, but now that I have something that works fairly well, I'll start giving it some thought.

Thanks for the input, very useful!

williamboman:
This made my instance unresponsive (no response from server). I do have 5,656,482 lines of logs (`360M`) though.

CLI output looked normal, i.e. `User Impaloo loaded.` yada yada.

astorije:
Welp, I wouldn't expect anything different from a PR that loads all 360M of logs, in a sync fashion, and doesn't try anything smart about it 😅.

I knew there are serious issues regarding performance for now, I would not recommend this PR in production at the moment!
This is something I'm planning to tackle very soon though, now that feature itself is almost ok.

YaManicKill:
@astorije Is this still something you want to get done in this PR? I think it would be great if we got this or mysql done for the next version (not 2.3)

astorije:
@YaManicKill, I definitely want this to happen. It's not on the top of my todolist, but I will get back to it, yes :)
We already have all these log files hanging around, it's a shame not to display them at startup :D

src/client.js
```diff
@@ -106,8 +110,7 @@ Client.prototype.emit = function(event, data) {
 				if (target.chan.type === Chan.Type.LOBBY) {
 					chan = target.network.host;
 				}
-				userLog.write(
-					this.name,
+				this.userLog.write(
```
xPaw: Remove this entire log writing stuff out of this `emit` function and move log write calls where they're actually needed (pushMessage afaik)

chrisweeksnz:
+1 for getting this done - it would make patching underlying operating systems much less noisy event  :)

Thanks for all your great work!

LFlare:
Is this functioning?

astorije:
@xPaw, I'm thinking of closing this in favor of #1839, I think that's the right move. Can you confirm? It also seems more likely to land than mine, and would have way less edge cases. Any chance we can finish yours for v3 (I know it's testing phase right now, but still trying to judge how much work we have on our plate).
We can then keep logging for what it's supposed to be, logging.

Hey, at least I got to fix some log-related bugs (missing data), so that's not for nothing! Also before closing I'll take one last look at the code to see if we can salvage any fixes, test, etc.

xPaw:
@astorije I have already taken relevant parts from this PR to sqlite. We may want to re-visit this PR in the future if we will want to allow importing text logs into sqlite.

PR #664, opened 2016-10-02
Fix memory and reference shuffling when creating models

xPaw:
A bug we discovered yesterday during mysql logging debugging.

> 09:54	+Max-P	But yeah it's really weird, objects as values in _.extend gets duplicated and their constructor/prototype reattached after
> 09:54	+Max-P	So in Chan's constructor, `this` was one instance, but the actual one we use is a copy
> 10:22	+xPaw	create chan -> create network with chan -> push message to original chan obj -> print network -> no message

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -13,7 +13,7 @@ Chan.Type = {
 var id = 0;
 
 function Chan(attr) {
-	_.merge(this, _.extend({
+	_.defaults(this, attr, {
 		id: id++,
 		messages: [],
 		name: "",
@@ -23,7 +23,7 @@ function Chan(attr) {
 		unread: 0,
 		highlight: false,
 		users: []
-	}, attr));
+	});
 }
 
 Chan.prototype.pushMessage = function(client, msg) {
--- a/src/models/msg.js
+++ b/src/models/msg.js
@@ -26,13 +26,13 @@ module.exports = Msg;
 var id = 0;
 
 function Msg(attr) {
-	_.merge(this, _.extend({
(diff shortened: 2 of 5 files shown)


astorije:
For the record:

- `_.merge(object, [sources])`: https://lodash.com/docs/4.16.4#merge

> This method is like `_.assign` except that it recursively merges own and inherited enumerable string keyed properties of source objects into the destination object. Source properties that resolve to `undefined` are skipped if a destination value exists. Array and plain object properties are merged recursively. Other objects and value types are overridden by assignment. Source objects are applied from left to right. Subsequent sources overwrite property assignments of previous sources.
>
> **Note:** This method mutates `object`.
>
> ### Example
> ```js
var object = {
  'a': [{ 'b': 2 }, { 'd': 4 }]
};
var other = {
  'a': [{ 'c': 3 }, { 'e': 5 }]
};
_.merge(object, other);
// => { 'a': [{ 'b': 2, 'c': 3 }, { 'd': 4, 'e': 5 }] }
```

- `_.defaults(object, [sources])`: https://lodash.com/docs/4.16.4#defaults

> Assigns own and inherited enumerable string keyed properties of source objects to the destination object for all destination properties that resolve to `undefined`. Source objects are applied from left to right. Once a property is set, additional values of the same property are ignored.
> 
> **Note:** This method mutates `object`.
>
> ### Example
> 
> ```js
_.defaults({ 'a': 1 }, { 'b': 2 }, { 'a': 3 });
// => { 'a': 1, 'b': 2 }
```

Issue #89, opened 2016-02-22
Date Separator


nomadturk:
Currently we don't have a date separator. So either in PM's or channels when we see a message we don't know if it was sent today or 10 days earlier.

We can't add this kind of information to a mouseover information box for the date info due to adding more burden to the script either

So, I think at least a HR or similar line separating messages when a new day starts would be good.

PR #671, opened 2016-10-04
Add a date separator to channels/PMs

stepie22:
This addresses issue #89 by adding a date separator whenever the date changes between messages. It also adds it to the top of a channel (and before the 'show older messages' button) to indicate the date there.

**Previews:**

Date Changing:
date change

Private message:
private message

cc @nomadturk

This is also my first big-project PR, so I would appreciate it if you'd tell me if I did something wrong.

### DIFF
--- a/client/js/libs/handlebars/localedate.js
+++ b/client/js/libs/handlebars/localedate.js
@@ -1,0 +1,5 @@
+"use strict";
+
+Handlebars.registerHelper("localedate", function(time) {
+	return new Date(time).toLocaleDateString();
+});
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -307,6 +307,27 @@ $(function() {
 		} else {
 			channel.append(render("unread_marker"));
 		}
+
+		if (data.type !== "lobby") {
+			var lastDate;
+			$(chat.find("#chan-" + data.id + " .messages .msg[data-time]")).each(function() {
+				var msg = $(this);
+				var msgDate = new Date(msg.attr("data-time"));
+
+				// Top-most message in a channel
+				if (!lastDate) {
+					lastDate = msgDate;
+					msg.before(render("date-marker", {msgDate: msgDate}));
+				}
+
+				if (lastDate.toDateString() !== msgDate.toDateString()) {
+					msg.before(render("date-marker", {msgDate: msgDate}));
+				}
+
+				lastDate = msgDate;
+			});
+		}
+
 	}
 
 	function renderChannelUsers(data) {
@@ -362,6 +383,21 @@ $(function() {
 		var target = "#chan-" + data.chan;
 		var container = chat.find(target + " .messages");
 
+        // Check if date changed
+		var prevMsg = $(container.find(".msg")).last();
+		var prevMsgTime = new Date(prevMsg.attr("data-time"));
+		var msgTime = new Date(msg.attr("data-time"));
+
+		// It's the first message in a channel/query
+		if (prevMsg.length === 0) {
+			container.append(render("date-marker", {msgDate: msgTime}));
+		}
+
+		if (prevMsgTime.toDateString() !== msgTime.toDateString()) {
+			prevMsg.append(render("date-marker", {msgDate: msgTime}));
+		}
+
+        // Add message to the container
 		container
 			.append(msg)
 			.trigger("msg", [
@@ -382,6 +418,13 @@ $(function() {
 			.find("#chan-" + data.chan)
 			.find(".messages");
 
+		// Remove the date-change marker we put at the top, because it may
+		// not actually be a date change now
+		var firstChild = $(chan).children().eq(0);
+		if (firstChild.attr("class") === "date-marker") {
+			firstChild.remove();
+		}
+
 		// get the scrollable wrapper around messages
 		var scrollable = chan.closest(".chat");
 		var heightOld = chan.height();
@@ -394,6 +437,28 @@ $(function() {
 		if (data.messages.length !== 100) {
 			scrollable.find(".show-more").removeClass("show");
 		}
+
+		// Date change detect
+		// Have to use data instaid of the documentFragment because it's being weird
+		var lastDate;
+		$(data.messages).each(function() {
+			var msgData = this;
+			var msgDate = new Date(msgData.time);
+			var msg = $(chat.find("#chan-" + data.chan + " .messages #msg-" + msgData.id));
+
+			// Top-most message in a channel
+			if (!lastDate) {
+				lastDate = msgDate;
+				msg.before(render("date-marker", {msgDate: msgDate}));
+			}
+
+			if (lastDate.toDateString() !== msgDate.toDateString()) {
+				msg.before(render("date-marker", {msgDate: msgDate}));
+			}
+
+			lastDate = msgDate;
+		});
+
 	});
 
 	socket.on("network", function(data) {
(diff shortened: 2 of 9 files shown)


stepie22:
Impaloo on irc pointed out that something like "Today"/"Yesterday"/"X days ago" might be better than just displaying the date, at least for recent dates. Any thoughts on this?

MaxLeiter:
Awesome job! It seems like PolarizedIons name is slightly indented in the second photo, any idea about that?

client/js/libs/handlebars/localedate.js
```diff
@@ -0,0 +1,5 @@
+"use strict";
```
MaxLeiter: Can we not use the already existing localetime somehow?

client/js/lounge.js
```diff
@@ -372,6 +408,16 @@ $(function() {
 			.find("#chan-" + data.chan)
 			.find(".messages");
 
+
+			// remove the date-change marker we put at the top, because it may
```
MaxLeiter: Please format the comments better - have the top one directly above the bottom, not indented

client/js/lounge.js
```diff
@@ -384,6 +430,28 @@ $(function() {
 		if (data.messages.length !== 100) {
 			scrollable.find(".show-more").removeClass("show");
 		}
+
+		// Have to use data instiad of the documentFragment because it's being weird
```
MaxLeiter: instead*

astorije:
I will review and test when I am on a stable network, but I wouldn't worry about "humanized" versions for dates as it comes with its set of problems: should it constantly update so that "Today" becomes "Yesterday" when a new day start without refreshing the page? What format? Custom code or package? Should a click on the date do nothing, toggle between "humazined" date and absolute date, append a hash after the URL, etc.
Also, if we're there, I'd rather see "Tuesday 4 October 2016" rather "10/4/2016".

But all that is building up on something that's already an enhancement. I say, let's have a marker that displays the native format outputted by JS (what you did here) when a new day starts (at local midnight) and improve on this later on. This will already make up for a meaty PR to review and test!

client/js/libs/handlebars/localedate.js
stepie22: You mean moving that helper into that file or using the localetime function to get the date? If the latter, that's going to be a bit more work because I'd need to localetime(date).split(",")[0]

client/js/lounge.js
stepie22: Sorry that became a thing when switching the spaces to tabs to not get yelled at by the linter

nomadturk:
Nice one @stepie22

Looks like it exactly doe what it's supposed to do.

But as for Today,  Yesterday tags,  I think I'm on the same page with  @astorije 
We don't actually need them,  dates would just do fine. And no need to burden the browsers with more Javascript either. (:

stepie22:
@astorije @nomadturk Thanks for all the kind words :) My first "real" PR, so it's kinda exciting

client/js/libs/handlebars/localedate.js
astorije: I'm fine having these in 2 different helpers, at least for now. When we start messing up with the text itself, then we can use one within the other, but there will still be 2 helpers, so 2 files is ok IMO.

client/js/lounge.js
```diff
@@ -1142,13 +1207,34 @@ $(function() {
 			var chan = $(this);
 			if (chan.find(".messages .msg:not(.unread-marker)").slice(0, -100).remove().length) {
 				chan.find(".show-more").addClass("show");
+
```
xPaw: I wish we got rid of this damn `setInterval`, not a fan of running this 'cleanup' code a lot.

williamboman:
In highly inactive channels, the date separators look quite bloated (I'm hiding joins/parts/quits e.t.c.)

screenshot 2016-10-11 at 10 44 51

client/js/lounge.js
stepie22: @xPaw yea, the loop I made there felt a bit 'hacky', but it was the best solution I could come up with. Had to do this because they would not get removed with those messages. If you have suggestions about what I coded here, feel free to give.

stepie22:
@williamboman I didn't think of that case. I could add a check to see if the message is visible(which I don't feel is the "right" solution, but think is the best for now), or just not use it at all for join/parts(though I think people would disagree on this).

I think I'll end up adding classes to it, to hide them if the new message would be hidden. If anyone has any thoughts on this, feel free to add

MaxLeiter:
Is anyone sure how IRCCloud handles the date separater? It makes sense to me to only show a separater if there are messages before it unless it's the most recent (meaning in the above screenshot, just 10/11/2016 would show, and a separater would appear after roots message)

Also, should probably hide the date above the topic if you're just joining the channel.

stepie22:
The issue @williamboman has is because of hidden joins/pars as far as I can tell.

 Fixing that would be a lot of work/logic, and I don't know how  @astorije would feel if I loop through all the messages in a channel the visibility for a message changes to fix the position of the date. 

Idealy that part would be reworked so it doesnt trip this up, or the unread marker (same issue with not visible messages)

stepie22:
@MaxLeiter  I asked someone using IRCCloud to show what happens with the dates when they hide join/parts. This is what they sent back: picture

MaxLeiter:
So, seeing as that's expected behavior (compared to IRCcloud), I think this can be merged?

MiniDigger:
I would also be in favor of displaying the date as x days ago but I also think that it would be a good idea to show the "absolute" date as a tool tip on hover. 
(like github does for example )

YaManicKill:
> I'm :-1: on this one as long as it prints absolute date. (I'd much prefer relative, e.g. Today, Yesterday, 2 days ago and so on)

I agree with @astorije, let's get this one in (it has been up for over a month) and then someone can look at making them relative dates (either as an option or instead of this). I would **much** rather have this than nothing, and we are dreadful at coming to consensus.

client/js/lounge.js
YaManicKill: I spent a long time trying to think of a nicer way to do this.

You could obviously select the first child that isn't a date-change-marker, but I don't think there's a way to then delete all except 1 of the date-change-markers before it. And even if you did, jquery would still be doing the same amount of work, so ... meh.

Having said that, I think that this code could be run on initial page load and on each message, rather than on the set interval. We know that the only time this is going to be necessary is if we've had a new message received (or on initial load), don't we?

client/views/date-change-marker.tpl
```diff
@@ -0,0 +1,3 @@
+<div class="date-change-marker">
```
YaManicKill: Minor nitpick, can we just call it "date-marker"? No need for the "change" (especially as the first date marker on the screen won't be about the date changing, it's about telling you the original date).

client/js/libs/handlebars/localedate.js
YaManicKill: Yeah, this is fine, imo, there is a clear difference between the 2, and we like modularity.

client/js/lounge.js
YaManicKill: Ah right, sorry I missed that the issue only happens because the setInterval exists. Alright, Ignore this then, that's a seperate PR.

stepie22:
@williamboman I agree with @astorije and @YaManicKill, this PR has definitely dragged on long enough. I'm sure you can whip up a quick PR up to make it use relative 'dates' :)

stepie22:
I've gone ahead and squashed this (with the help of YaManiac)

YaManicKill:
So, we have 2 approvals on this.

@astorije You mentioned you were going to have a look at it. Do you feel you want to check it before it's merged? I'm guessing you are probably fine, but let me know if you aren't.

@williamboman Are you actually :-1: on this as it is? Or are you fine with it, on the assumption that someone else will do relative dates sometime.

williamboman:
> @williamboman Are you actually 👎 on this as it is? Or are you fine with it, on the assumption that someone else will do relative dates sometime.

I'm fine with it as it is (but pls rel dates :smile:)

astorije:
> @astorije You mentioned you were going to have a look at it. Do you feel you want to check it before it's merged? I'm guessing you are probably fine, but let me know if you aren't.

Sorry, work has kept me busy enough that sleeping has become entirely optional now 😅 
Thanks for carefully reviewing this, and I'm happy this got merged! Good job @stepie22!

PR #673, opened 2016-10-04
Check if SSL key and certificate files exist

### LINKED ISSUE #416: Check if ssl key files exist instead of crashing with a cryptic error
We need to report user friendly messages, instead of messages like `Error: ENOENT: no such file or directory, open ''`

toXel:
It's a first attempt. Please suggest changes.
This fixes #416.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -36,9 +36,19 @@ module.exports = function() {
 		server = server.createServer(app).listen(config.port, config.host);
 	} else {
 		server = require("spdy");
+		const keyPath = Helper.expandHome(config.https.key);
+		const certPath = Helper.expandHome(config.https.certificate);
+		if (!config.https.key.length || !fs.existsSync(keyPath)) {
+			log.error("Path to SSL key is invalid. Stopping server...");
+			process.exit();
+		}
+		if (!config.https.certificate.length || !fs.existsSync(certPath)) {
+			log.error("Path to SSL certificate is invalid. Stopping server...");
+			process.exit();
+		}
 		server = server.createServer({
-			key: fs.readFileSync(Helper.expandHome(config.https.key)),
-			cert: fs.readFileSync(Helper.expandHome(config.https.certificate))
+			key: fs.readFileSync(keyPath),
+			cert: fs.readFileSync(certPath)
 		}, app).listen(config.port, config.host);
 	}
 


src/server.js
```diff
@@ -36,10 +36,17 @@ module.exports = function() {
...
-			cert: fs.readFileSync(Helper.expandHome(config.https.certificate))
-		}, app).listen(config.port, config.host);
+		let keyPath = Helper.expandHome(config.https.key);
+		let certPath = Helper.expandHome(config.https.certificate);
+		if (fs.existsSync(keyPath) && fs.existsSync(certPath)) {
+			server = server.createServer({
+				key: fs.readFileSync(keyPath),
+				cert: fs.readFileSync(certPath)
+			}, app).listen(config.port, config.host);
+		} else {
```
xPaw: I would prefer two separate messages for each of the files. And do a negative check, then you dont need an `else`.

xPaw:
Just wondering, does it explode if the file exists but it's not valid? If so might want to add an extra check for that.

toXel:
Good idea. I'll look into this.

astorije:
@xPaw, in #416, @toXel [asked if we preferred an `if` condition or a `try`/`catch` block](https://github.com/thelounge/lounge/issues/416#issuecomment-251528189). Do you have an opinion on this? I don't really have one myself.

toXel:
@xPaw yes it crashes if the files are not valid:
```
_tls_common.js:68
      c.context.setCert(options.cert);
                ^

Error: error:0906D06C:PEM routines:PEM_read_bio:no start line
```

toXel:
@xPaw Nope I haven't found a function for it. There's a pem npm module with a `readCertificateInfo` function. But I think this will also crash on invalid files.

src/server.js
```diff
@@ -36,9 +36,19 @@ module.exports = function() {
 		server = server.createServer(app).listen(config.port, config.host);
 	} else {
 		server = require("spdy");
+		let keyPath = Helper.expandHome(config.https.key);
+		let certPath = Helper.expandHome(config.https.certificate);
```
astorije: Can you move these from `let` to `const` please?

astorije:
After addressing my comment, can you squash your PR as well please? :)

toXel:
Done :)

PR #674, opened 2016-10-05
Override defaults with URL

### LINKED ISSUE #605: Allow specifying default user settings via URL arguments
This would be useful with custom webchat embeds: for example, `http://somenet.abc/?nick=webuser...&channels=#chat` would automatically change the defaults to the channel owner's preferences.

MaxLeiter:
My attempt at a client-side alternative to #634 

<del>I need to replace getQueryString() with http://stackoverflow.com/questions/2090551/parse-query-string-in-javascript/13419367#13419367, which should fix channels</del>

Closes #605

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -993,7 +993,31 @@ $(function() {
 			}
 		});
 	});
+	if ($("body").hasClass("public")) {
+		$("#connect").one("show", function() {
+			var params = window.URI(document.location.search);
+			params = params.search(true);
+			// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
+			// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Iterating_over_own_properties_only
+			for (var key in params) {
+				if (params.hasOwnProperty(key)) {
+					var value = params[key];
+					// \W searches for non-word characters
+					key = key.replace(/\W/g, "");
 
+					var element = $("#connect input[name='" + key + "']");
+					// if the element exists, it isn't disabled, and it isn't hidden
+					if (element.length > 0 && !element.is(":disabled") && !element.is(":hidden")) {
+						if (element.is(":checkbox")) {
+							element.prop("checked", (value === "1" || value === "true") ? true : false);
+						} else {
+							element.val(value);
+						}
+					}
+				}
+			}
+		});
+	}
 	windows.on("show", "#settings", updateDesktopNotificationStatus);
 
 	forms.on("submit", "form", function(e) {


client/js/lounge.js
```diff
@@ -259,6 +259,16 @@ $(function() {
 		return msg;
 	}
 
+	// http://stackoverflow.com/a/1099670
+	function getQueryParam(name) {
+		var url = location.href;
```
maxpoulin64: Use `location.search` instead to avoid the need to parse it. See http://stackoverflow.com/a/13419367/1739985 for a complete solution.

client/js/lounge.js
```diff
@@ -994,6 +1005,22 @@ $(function() {
...
+
+	// TODO: doesn't support channels correctly, needs to change %23 to #
+	// 		 boolean not supported, like TLS
+	//       port doesn't work because it has no ID
+
+	// name = network, host = server
+	windows.on("show", "#connect", function() {
+		var url_config_items = ["name", "host", "password", "nick", "username", "realname", "channels"];
+		url_config_items.forEach(function(e) {
+			var key = getQueryParam(e);
```
maxpoulin64: Have your query param extractor return an object to avoid parsing the URL again and again needlessly.

client/js/lounge.js
MaxLeiter: Changed to your link

astorije:
👏 👏 
💯 for the initiative! I'll take a look to the code when I can be stable, in a few days. Thanks!

MaxLeiter:
I think this is now ready for review

client/js/lounge.js
```diff
@@ -994,6 +1005,22 @@ $(function() {
 		});
 	});
 
+	windows.on("show", "#connect", function() {
+		var params = parseQuery();
+		// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
+		for (var key in params) {
+			var value = params[key].toLowerCase();
+			var element = $("input[name=" + key + "]");
```
xPaw: Limit this selector to `connect` window, not global.

client/js/lounge.js
```diff
@@ -994,6 +1005,22 @@ $(function() {
...
+
+			if (value === "true" || value === "false") {
```
xPaw: I think you should be checking the type of input instead for safety.

MaxLeiter: Fixed

MaxLeiter:
This is done except for tests. I haven't really written tests before, and am a little lost on where to place them - do I need to make a new client/ subdirectory in tests/?

client/js/lounge.js
```diff
@@ -994,6 +1005,22 @@ $(function() {
 		});
 	});
 
+	windows.on("show", "#connect", function() {
+		var params = parseQuery();
+		// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
+		for (var key in params) {
+			var value = params[key].toLowerCase();
```
xPaw: I think you need to do something like `.replace(/[^[a-z]/, '')` here to disallow possible abuse and escapes from the jquery selector.

client/js/lounge.js
```diff
@@ -994,6 +1005,22 @@ $(function() {
...
+			var element = $("#connect input[name=" + key + "]");
+
+			if (element.is("input[type='checkbox']")) {
```
xPaw: `.is(":checkbox")` is simpler.

astorije:
Am I right assuming this feature is only useful/applicable in public mode? If so, should we check it there?

Also, what's the expected/actual behaviors when trying to set a value in a locked or hidden network?

xPaw:
> Am I right assuming this feature is only useful/applicable in public mode? If so, should we check it there?

Technically it still works in private mode too, I see no reason to add an explicit check for it.

> Also, what's the expected/actual behaviors when trying to set a value in a locked or hidden network?

It most likely will allow to visually change the fields, but the server will still correctly respect the lock. Perhaps a `disabled` input check is needed here.

client/js/lounge.js
```diff
@@ -994,6 +994,23 @@ $(function() {
 		});
 	});
 
+	$("#connect").one("show", function() {
+		var params = window.URI(document.location.search);
+		params = params.search(true);
+		// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
+		for (var key in params) {
```
astorije: `for (var ... in ...)` includes properties that are inherited through the prototype chain, and this is generally not what you want. See http://eslint.org/docs/rules/guard-for-in for an alternative.
Also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Iterating_over_own_properties_only.

client/js/lounge.js
```diff
@@ -994,6 +994,23 @@ $(function() {
...
+			var value = params[key];
+			key = key.replace(/\W/g, "");
+
+			var element = $("#connect input[name='" + key + "']");
+			if (element.is(":checkbox")) {
+				element.prop("checked", (value === "1" || value === "true") ? true : false);
```
astorije: Can't we use `element.is(':checked')` for this instead?

client/js/lounge.js
```diff
@@ -994,6 +994,23 @@ $(function() {
...
 
+	$("#connect").one("show", function() {
+		var params = window.URI(document.location.search);
+		params = params.search(true);
+		// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
+		for (var key in params) {
+			var value = params[key];
+			key = key.replace(/\W/g, "");
+
+			var element = $("#connect input[name='" + key + "']");
```
astorije: This also needs to check that:

- input is not `disabled` (when `lockNetwork` is `true`)
- input is not hidden (when `displayNetwork` is `false`). This might require a bit more fiddling than the one above, but it would be pretty nasty to let someone send you to a connect window with a bad host that you cannot see anywhere.

astorije:
In addition to my review above:

- We probably want to force displaying connect window when there are params in the URL. In public mode, people get sent to the connect window directly, but in private mode, it's not obvious these params are used until one clicks on the "Connect" button.

- In general, even though it technically works, I don't think this is quite ready for private mode. Use case for private mode is very limited, and the only use case I can think of would be a service that add links to channels to click on. Problem is, at the moment this would create a brand new connection for the same host every time. A solution could be to join the channel if already connected to this host, but maybe for a latter time. I just think this is weird for private mode at the moment. I won't strongly object though, it's just my 2c.

- Not too far from the first point here, when loading a URL with such params, params not specified should be empty instead of relying on defaults IMO.

- We might want to add a list of allowed fields, similarly to what [this other PR](https://github.com/thelounge/lounge/pull/634/files#diff-e6a5b42b2f7a26c840607370aed5301aR127) was doing. I can't think of a use case where specifying password in URL is a good idea. I might be wrong though, but I think it's safer to whitelist things. Thoughts?

MaxLeiter:
> We probably want to force displaying connect window when there are params in the URL. In public mode, people get sent to the connect window directly, but in private mode, it's not obvious these params are used until one clicks on the "Connect" button.

I agree

> In general, even though it technically works,... I just think this is weird for private mode at the moment. I won't strongly object though, it's just my 2c.

Once this is merged (and maybe more functionality is added) I plan to make a chrome extension that will add support for` irc://` protocol. I'll add a check to see if you're already connected for private mode.

> Not too far from the first point here, when loading a URL with such params, params not specified should be empty instead of relying on defaults IMO.

I disagree - I think the defaults should stay, because the person providing a link with URL parameters is probably the host of the lounge instance, and has the defaults set for a reason.

> We might want to add a list of allowed fields, ... but I think it's safer to whitelist things. Thoughts?

I don't think it will really make a difference; it's 'locked-in' to the `#connect` window anyways. And again, the person providing this URL probably hosts the lounge instance/channel they want you to connect to, so may be providing the password for a reason.

client/js/lounge.js
MaxLeiter: This is setting the `checked` property. I could do an `if(is(':checked'))` but I don't think thats an improvement.

client/js/lounge.js
```diff
@@ -994,6 +994,23 @@ $(function() {
 		});
 	});
 
+	$("#connect").one("show", function() {
+		var params = window.URI(document.location.search);
+		params = params.search(true);
+		// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
+		for (var key in params) {
+			var value = params[key];
+			key = key.replace(/\W/g, "");
```
MaxLeiter: Fixed

client/js/lounge.js
MaxLeiter: Fixed

client/js/lounge.js
```diff
@@ -994,6 +994,23 @@ $(function() {
...
+
+			var element = $("#connect input[name='" + key + "']");
+			if (element.is(":checkbox")) {
+				element.prop("checked", (value === "1" || value === "true") ? true : false);
+			} else {
+				element.val(value);
+			}
```
MaxLeiter: Fixed

MaxLeiter:
Addressed @astorije's review comments, haven't fixed anything I responded to last comment.

MaxLeiter:
After some messing around, turns out it's pretty difficult to handle private mode well. I ended up simply removing the functionality in private mode, and I think that can be addressed in another PR. 

Ultimately, I think we'll need to move the parameter code into a new function (like `handleParameters()`) that will have a callback. If we're in private mode, we'll need to check if we're already connected to the network and such in the callback.

astorije:
Both comments by maxpoulin64 were addressed, and I don't want to overwhelm him with things that were already reviewed twice :-)
Merging, thanks a lot @MaxLeiter!

Issue #223, opened 2016-03-27
The /mode command doesn't support specifying the target


maxpoulin64:
Currently, the mode always target the current channel. This works fine for most cases, but those constructs are also valid but unsupported:

* `/mode Max-P +x` (changing mode on self)
* `/mode #channel +m` (changing mode on other channel)

I don't think there are cases where one can change the mode of other people or non-joined channels, but we probably should.

YaManicKill:
Note that this also should count as a bug, as it causes some issues.

I tried do to `/mode #pokemontrades +b` which is how you get the banlist for a channel. However, I get an error about the chan modes, and it sets +ms on the channel. So it's probably taking the name of the chan and using that as the mode.

YaManicKill:
(note that doing `/mode +b` doesn't seem to work either, which means there is no current way to get the banlist on thelounge unless servers have another way to do it)

YaManicKill:
Another thing related to this is viewing modes.

"/mode" or "/mode #channel" doesn't show you anything.

xPaw:
Here's how hexchat deals with the command: https://github.com/hexchat/hexchat/blob/c698e13c0d3f8835581d1271371110da18827a92/src/common/outbound.c#L2679-L2703

astorije:
Should be fixed by #679.

fuzzy76:
#341 was closed as a duplicate of this, but the behaviour described by McInkay still persists. So either this or #341 should be reopened I guess?

McInkay:
Yeah, it looks like #341 wasn't a full duplicate of this one. We can now specify the target, but it looks like we don't deal with the mode message when not setting a mode.

xPaw:
I have a working prototype branch here: https://github.com/thelounge/thelounge/tree/xpaw/chan-mode-is

PR #679, opened 2016-10-08
Fix /mode command to correctly assume target

### LINKED ISSUE #676: Cannot send multiple mode changes at once
Sending `/mode #channel -ao nick nick` only actually performs the -a. This implies that The Lounge intercepts mode commands and doesn't pass them to the server verbatim.

Came across this while running tests for #675.

xPaw:
Fixes #676 and #223.

Also adds support for passing in multiple nicks into `/op` (and other short commands). Adds a `/hop` and `/dehop` aliases.

### DIFF
--- a/src/plugins/inputs/mode.js
+++ b/src/plugins/inputs/mode.js
@@ -2,30 +2,60 @@
 
 exports.commands = ["mode", "op", "voice", "deop", "devoice"];
 
+var Chan = require("../../models/chan");
+var Msg = require("../../models/msg");
+
+exports.commands = [
+	"mode",
+	"op",
+	"deop",
+	"hop",
+	"dehop",
+	"voice",
+	"devoice",
+];
+
 exports.input = function(network, chan, cmd, args) {
-	if (args.length === 0) {
+	if (cmd !== "mode") {
+		if (chan.type !== Chan.Type.CHANNEL) {
+			chan.pushMessage(this, new Msg({
+				type: Msg.Type.ERROR,
+				text: `${cmd} command can only be used in channels.`
+			}));
+
+			return;
+		}
+
+		if (args.length === 0) {
+			chan.pushMessage(this, new Msg({
+				type: Msg.Type.ERROR,
+				text: `Usage: /${cmd} <nick> [...nick]`
+			}));
+
+			return;
+		}
+
+		const mode = {
+			op: "+o",
+			hop: "+h",
+			voice: "+v",
+			deop: "-o",
+			dehop: "-h",
+			devoice: "-v"
+		}[cmd];
+
+		args.forEach(function(target) {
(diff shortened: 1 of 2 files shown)


src/plugins/inputs/mode.js
```diff
@@ -1,29 +1,57 @@
...
+	if (args.length === 0 || args[0][0] === "+" || args[0][0] === "-") {
+		args.unshift(chan.type === Chan.Type.CHANNEL || chan.type === Chan.Type.QUERY ? chan.name : network.nick);
 	}
 
-	var irc = network.irc;
-	irc.raw("MODE", chan.name, mode, user);
+	args.unshift("MODE");
 
-	return true;
+	network.irc.raw.apply(network.irc, args);
```
xPaw: `...` spread operator is not supported in node4.

src/plugins/inputs/mode.js
```diff
@@ -1,29 +1,57 @@
...
+
+		args.forEach(function(target) {
+			network.irc.raw("MODE", chan.name, mode, target);
+		});
+
+		return;
+	}
+
+	if (args.length === 0 || args[0][0] === "+" || args[0][0] === "-") {
+		args.unshift(chan.type === Chan.Type.CHANNEL || chan.type === Chan.Type.QUERY ? chan.name : network.nick);
```
xPaw: Not sure if there's any use of allowing `query` targets here.

test/commands/mode.js
```diff
@@ -0,0 +1,90 @@
...
+			name: "Network Lobby",
+			type: Chan.Type.LOBBY
+		});
+
+		const testableNetwork = {
+			lastCommand: null,
+			nick: "xPaw",
+			irc: {
+				raw: function() {
+					testableNetwork.lastCommand = Array.prototype.join.call(arguments, " ");
```
astorije: Nice trick to get the latest command called!

test/commands/mode.js
```diff
@@ -0,0 +1,90 @@
...
+			ModeCommand.input(testableNetwork, channel, "voice", ["xPaw"]);
+			expect(testableNetwork.lastCommand).to.equal("MODE #thelounge +v xPaw");
+
+			ModeCommand.input(testableNetwork, channel, "devoice", ["xPaw"]);
+			expect(testableNetwork.lastCommand).to.equal("MODE #thelounge -v xPaw");
+
+			// Multiple arguments are supported, sent as separate commands
+			ModeCommand.input(testableNetwork, channel, "devoice", ["xPaw", "Max-P"]);
+			expect(testableNetwork.lastCommand).to.equal("MODE #thelounge -v Max-P");
+		});
```
astorije: I'd rather see these broken up in multiple expects (can be grouped in a common `describe`), but really no big deal, especially at that point.
Awesome job on adding tests though!!

PR #680, opened 2016-10-08
Prototype of server-side message search

xPaw:
MVP. Gib feedback. This kinda abuses the channel model.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -680,6 +680,7 @@ kbd {
 	font-size: 14px;
 }
 
+.message-search-input,
 #windows .input {
 	background-color: white;
 	border: 1px solid #cdd3da;
@@ -702,6 +703,8 @@ kbd {
 	resize: vertical;
 }
 
+.message-search-input:hover,
+.message-search-input:focus,
 #windows .input:hover,
 #windows .input:focus {
 	border-color: #79838c;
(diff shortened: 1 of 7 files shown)


metsjeesus:
Limit result size, add time, network and channel as parameter.

xPaw:
I do want to add search settings:
- Scope:
 - All
 - Network
 - Channel

- Case-sensitive

But that's to be done after the `prompt` is removed in favor of a proper UI design.

astorije:
In terms of scope or settings, we could use something àla GitHub prefixes: https://help.github.com/articles/searching-repositories/

I understand you are searching among the messages already in memory, correct? Why not search in logs (when available) instead?
That would provide more history to search in, and would decouple from in-memory stuff, which ideally we should not be relying upon for historical content.

xPaw:
> In terms of scope or settings, we could use something àla GitHub prefixes:

I'd want UI for it too.

> I understand you are searching among the messages already in memory, correct? Why not search in logs (when available) instead?

I would wait to get some kind of logging abstraction api for that first. Do not want to mess with parsing log files at the moment.

astorije:
> Do not want to mess with parsing log files at the moment.

Meh, it's not that bad #663 😄

YaManicKill:
What's the status of this?

xPaw:
Needs UI work.

xPaw:
Rebased on master, a little bit more simpler code (currently pretends to be a query window).

astorije:
I am very puzzled right now:

screen shot 2017-07-21 at 02 50 14

PR #681, opened 2016-10-09
Enforce more eslint rules

xPaw:
More consistent style, and avoid potential bugs.


astorije:
Builds for this PR should pass after #684 is merged. I'll rebase this PR accordingly.

.eslintrc.yml
```diff
@@ -11,27 +11,34 @@ parserOptions:
...
+  handle-callback-err: 2
   indent: [2, tab]
   key-spacing: [2, {beforeColon: false, afterColon: true}]
   keyword-spacing: [2, {before: true, after: true}]
   linebreak-style: [2, unix]
   no-console: 0
   no-control-regex: 0
-  no-inner-declarations: 2
-  no-invalid-regexp: 2
-  no-irregular-whitespace: 2
```
astorije: Why removing these 3? It looks to me like we do want to keep checking on them, no?

.eslintrc.yml
```diff
@@ -11,27 +11,34 @@ parserOptions:
...
+  no-else-return: 2
+  no-implicit-globals: 2
+  no-multi-spaces: 2
+  no-shadow: 2
+  no-template-curly-in-string: 2
   no-trailing-spaces: 2
-  no-unexpected-multiline: 2
```
astorije: We might also want to keep this. From http://eslint.org/docs/rules/no-unexpected-multiline:

> Note that the patterns considered problems are not flagged by the semi rule.

.eslintrc.yml
```diff
@@ -11,27 +11,34 @@ parserOptions:
...
-  no-unreachable: 2
```
astorije: Was that removed on purpose? Why?

client/js/lounge.js
```diff
@@ -494,73 +495,75 @@ $(function() {
 		userStyles: userStyles.text(),
 	}, JSON.parse(window.localStorage.getItem("settings")));
 
-	for (var i in options) {
-		if (i === "userStyles") {
-			if (!/[\?&]nocss/.test(window.location.search)) {
-				$(document.head).find("#user-specified-css").html(options[i]);
+	(function SettingsScope() {
```
astorije: `Settings` with an upper case on purpose? Also, why wrapping these 70 lines?

scripts/build-fontawesome.js
```diff
@@ -9,13 +9,13 @@ var fonts = [
 	"fontawesome-webfont.woff2"
 ];
 
-fs.ensureDir(destDir, function (err) {
-	if (err) {
-		console.error(err);
+fs.ensureDir(destDir, function(dirErr) {
+	if (dirErr) {
+		console.error(dirErr);
```
astorije: Why that change?

scripts/build-fontawesome.js
```diff
@@ -9,13 +9,13 @@ var fonts = [
...
 	}
 
-	fonts.forEach(function (font) {
-		fs.copy(srcDir + font, destDir + font, function (err) {
+	fonts.forEach(function(font) {
+		fs.copy(srcDir + font, destDir + font, function(err) {
```
astorije: Urgh, I know this is to be consistent with the rest of our code, but this is painful to look at 😅. Sorry I missed this.
Also, I believe I should have used `=>` functions there instead...

src/client.js
```diff
@@ -278,6 +278,10 @@ Client.prototype.updateToken = function(callback) {
 	var client = this;
 
 	crypto.randomBytes(48, function(err, buf) {
+		if (err) {
+			throw err;
+		}
```
astorije: Any unexpected consequences with this?

.eslintrc.yml
xPaw: Theyre defaults in the recommended rules.

client/js/lounge.js
xPaw: I wrapped them to reduce the scope (especially the I variable which was being shadowed in other functions)

src/client.js
xPaw: It *should* crash.

scripts/build-fontawesome.js
xPaw: Shadows same var name.

.eslintrc.yml
astorije: Argh, I forgot to check that...

PR #682, opened 2016-10-09
Remove all font files except WOFF

xPaw:
- [EOT fonts](http://caniuse.com/#feat=eot) are only useful for IE8 where lounge does not work.
- [SVG fonts](http://caniuse.com/#feat=svg-fonts) have been abandoned years ago.
- [TTF fonts](http://caniuse.com/#feat=ttf) removed as [WOFF](http://caniuse.com/#feat=woff) has more global support and covers all the browsers that also support TTF.

WOFF is supported 92% globally (basically everywhere, except for Opera Mini which simply does not support any fonts). WOFF2 is supported 64% globally, pretty much in all latest browsers.


scripts/build-fontawesome.js
```diff
@@ -3,11 +3,8 @@
 var fs = require("fs-extra");
 
 var srcDir = "./node_modules/font-awesome/fonts/";
-var destDir = "./client/fonts/";
+var destDir = "./client/css/fonts/";
```
astorije: @xPaw, this was pulled to `client/fonts/` on purpose. I believe all fonts should be there instead of `client/css/fonts/`, but this is also a way to locate the fonts installed by our (up and coming) asset pipeline differently.
What do you think?

astorije:
This is great! A couple comments before approving.

astorije:
For the record:

> +xPaw	Astorije, just undo my fontawe move as you commented

PR #685, opened 2016-10-10
Fix nick changes not being properly reported in the logs

astorije:
### Result

When typing `/nick astorije2` (or using the nick editor) and tailing the according log file:

Before | After
--- | ---
`[2016-10-10 15:17:47] *  nick ` | `[2016-10-10 16:32:47] * astorije nick astorije2`


MaxLeiter:
Works fine for me

PR #689, opened 2016-10-12
Replace lodash's each/map with ES5 native forEach/map

astorije:
And use some fat arrows because these are the good kind of fats.

### DIFF
--- a/src/plugins/irc-events/names.js
+++ b/src/plugins/irc-events/names.js
@@ -1,6 +1,5 @@
 "use strict";
 
-var _ = require("lodash");
 var User = require("../../models/user");
 
 module.exports = function(irc, network) {
@@ -11,13 +10,7 @@ module.exports = function(irc, network) {
 			return;
 		}
 
-		chan.users = [];
-
-		_.each(data.users, function(u) {
-			var user = new User(u, network.prefixLookup);
-
-			chan.users.push(user);
-		});
+		chan.users = data.users.map(u => new User(u, network.prefixLookup));
 
 		chan.sortUsers(irc);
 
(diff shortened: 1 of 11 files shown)


MaxLeiter:
Maybe worth noting that according to [this ](https://stackoverflow.com/questions/18881487/why-is-lodash-each-faster-than-native-foreach#18884620), lodash's forEach is faster than the native implementation

astorije:
It's slightly more complicated than that: it will be faster on *some* browsers and similar in others because of Lodash's implementation and how browsers implement `forEach`.

That being said, while it's good to keep in mind in case this ever becomes a bottleneck, I don't think we should worry about that. While we're at it, why not using `for` and `while` loops :)

Thanks for bringing that up though, I didn't know that before you posted it.

xPaw:
Keep in mind we only use lodash on the server, so the nide is the only engine. :+1: for using native methods when possible.

src/plugins/irc-events/names.js
```diff
@@ -13,11 +12,7 @@ module.exports = function(irc, network) {
 
 		chan.users = [];
 
-		_.each(data.users, function(u) {
-			var user = new User(u, network.prefixLookup);
-
-			chan.users.push(user);
-		});
+		data.users.forEach(u => chan.users.push(new User(u, network.prefixLookup)));
```
xPaw: `map` should work fine here

astorije:
Re: micro-change performance, on that note I watched a nice talk about how JSPerf-based conclusions are often not applicable :) https://www.youtube.com/watch?v=65-RbBwZQdU

PR #691, opened 2016-10-13
Socket listening capability

ilmaisin:
Here is tentative socket listening ability.

Things that need yet to be done:
- <s>Modify the stdio output to tell when we listen to socket</s> DONE
- <s>Add a code to clean up old socket if the server has exited non-gracefully due to power outage or service crash</s> Not really necessary since the socket can be placed on a directory like /run or /tmp that is emptied on boot
- <s>Possibly add a command line option for socket listening</s> No need because I implemented the suggested way to use the host parameter to enable socket listening.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -35,9 +35,17 @@ module.exports = function() {
 		log.warn("Server is public and set to use LDAP. Set to private mode if trying to use LDAP authentication.");
 	}
 
+	var nodeHost = config.host;
+	var nodePort = config.port;
+
+	if (typeof config.host !== 'undefined' && config.host.includes('unix:')){
+		var nodeHost = null;
+		var nodePort = config.host.replace("unix:", "");
+	}
+
 	if (!config.https.enable) {
 		server = require("http");
-		server = server.createServer(app).listen(config.port, config.host);
+		server = server.createServer(app).listen(nodePort, nodeHost);
 	} else {
 		server = require("spdy");
 		const keyPath = Helper.expandHome(config.https.key);
@@ -53,7 +61,7 @@ module.exports = function() {
 		server = server.createServer({
 			key: fs.readFileSync(keyPath),
 			cert: fs.readFileSync(certPath)
-		}, app).listen(config.port, config.host);
+		}, app).listen(nodePort, nodeHost);
 	}
 
 	if (config.identd.enable) {
@@ -79,14 +87,6 @@ module.exports = function() {
 		} else {
 			init(socket);
 		}
-
-	function shutdown() {
-	    server.close(); // Removes socket file
-	    process.exit();
-	}
-
-	process.on('SIGINT', shutdown);
-
 	});
 
 	manager.sockets = sockets;
@@ -108,6 +108,14 @@ in ${config.public ? "public" : "private"} mode`);
 
 		manager.autoloadUsers();
 	}
+
+	function shutdown() {
+            log.info("Quitting...");
+	    server.close(); // Removes socket file
+	    process.exit();
+	}
+
+	process.on('SIGINT', shutdown);
 };
 
 function getClientIp(req) {
--- a/src/server.js
+++ b/src/server.js
@@ -79,6 +79,14 @@ module.exports = function() {
 		} else {
 			init(socket);
 		}
+
+	function shutdown() {
+	    server.close(); // Removes socket file
+	    process.exit();
+	}
+
+	process.on('SIGINT', shutdown);
+
 	});
 
 	manager.sockets = sockets;
--- a/src/server.js
+++ b/src/server.js
@@ -35,17 +35,22 @@ module.exports = function() {
 		log.warn("Server is public and set to use LDAP. Set to private mode if trying to use LDAP authentication.");
 	}
 
-	var nodeHost = config.host;
-	var nodePort = config.port;
+//	var host = config.host;
+//	var port = config.port;
 
 	if (typeof config.host !== 'undefined' && config.host.includes('unix:')){
-		var nodeHost = null;
-		var nodePort = config.host.replace("unix:", "");
+		const listeningSocket = true;
+		var host = null;
+		var port = config.host.replace("unix:", "");
+	}
+	else {
+		var host = config.host || null;
+		var port = config.port;
 	}
 
 	if (!config.https.enable) {
 		server = require("http");
-		server = server.createServer(app).listen(nodePort, nodeHost);
+		server = server.createServer(app).listen(port, host);
 	} else {
 		server = require("spdy");
 		const keyPath = Helper.expandHome(config.https.key);
@@ -61,7 +66,7 @@ module.exports = function() {
 		server = server.createServer({
 			key: fs.readFileSync(keyPath),
 			cert: fs.readFileSync(certPath)
-		}, app).listen(nodePort, nodeHost);
+		}, app).listen(port, host);
 	}
 
 	if (config.identd.enable) {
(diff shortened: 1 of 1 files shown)


xPaw:
@ilmaisin are you able to address @maxpoulin64 comment?

ilmaisin:
@xPaw I'm sorry, I forgot this. I take a look in following few days. It sounds like I could do it.

ilmaisin:
@maxpoulin64 I think it would be good idea to enalble socket listening with SSL anyway. It can make sense when one uses xinetd to prevent unauthorized users binding the port. While one _can_ make xinetd to do the ssl, it can be confusing if Lounge will fail with socket and ssl without a good reason. Supporting SSL with socket won't add too much complexity.

ilmaisin:
Sorry, I have been somewhat busy. Would it be too hacky or something if we did the following:

If the hostname begins with unix:, change the port variable to the socket path and host variable to null?

astorije:
> Would it be too hacky or something if we did the following:
> 
> If the hostname begins with unix:, change the port variable to the socket path and host variable to null?

@xPaw, what's your opinion on this?

xPaw:
Its better to use host variable alone. Other suftware does it the same way.

ilmaisin:
@xPaw I mean, node.js takes the socket path in the port variable. It would externally take the unix:/path/to/socket string as the host variable and ignore the port variable, but internally translate it to the format understood by node.js.

ilmaisin:
@xPaw Now it can listen to socket when the hostname is like unix:/path/to.sock. But is this a sane way to do it?

src/server.js
```diff
@@ -35,9 +35,18 @@ module.exports = function() {
 		log.warn("Server is public and set to use LDAP. Set to private mode if trying to use LDAP authentication.");
 	}
 
+	var host = config.host || null;
+	var port = config.port;
+
+
+	if (typeof config.host !== "undefined" && config.host.includes("unix:")) {
```
maxpoulin64: Probably should use `startsWith` here just to be sure

src/server.js
```diff
@@ -35,9 +35,18 @@ module.exports = function() {
...
+		host = null;
+		port = config.host.replace("unix:", "");
```
maxpoulin64: I'd use `substr` or a `/^unix:/` replace. On Linux the only invalid character in a path is `/`, so `unix:/run/unix:sockets/lounge.sock` is a perfectly valid path and would fail.

src/server.js
```diff
@@ -84,13 +93,17 @@ module.exports = function() {
...
 	const protocol = config.https.enable ? "https" : "http";
-	const host = config.host || "*";
 
 	log.info(`The Lounge ${colors.green(Helper.getVersion())} is now running \
 using node ${colors.green(process.versions.node)} on ${colors.green(process.platform)} (${process.arch})`);
 	log.info(`Configuration file: ${colors.green(Helper.CONFIG_PATH)}`);
-	log.info(`Available on ${colors.green(protocol + "://" + host + ":" + config.port + "/")} \
-in ${config.public ? "public" : "private"} mode`);
+	if (typeof config.host !== "undefined" && config.host.includes("unix:")) {
+		log.info(`Available on socket ${colors.green(config.host)} \
```
maxpoulin64: The same message can be used for both if you just set `protocol` to `unix` here for the message template.

maxpoulin64:
@ilmaisin It's just my opinion but I'd say yes that's a pretty sane way to do it. That's how a lot of software do it because URIs are quite generic and flexible. We could even actually make it parse strings like `https://foo:9001/path` to set both HTTPS, listening IP/hostname, port and the client's path if we wanted to go crazy (but that's not needed for this particular PR).

src/server.js
ilmaisin: I did the other two changes but this one doesn't seem to be that simple as if I just change the protocol to unix or unixs, the protocol will be shown twice as it is already included in the config.host value. We'll either need to do some if-elses and/or search-replaces or just stick with two messages.

ilmaisin:
Apparently server.js has been re-engineered so this PR needs to be reworked as well if it is going to be merged.

astorije:
> Apparently server.js has been re-engineered so this PR needs to be reworked as well if it is going to be merged.

@ilmaisin, we would like to merge it indeed. Any chance you could resume work on this? Thanks! :)

astorije:
@ilmaisin, are you still willing to work on this? 😃

ilmaisin:
Yes, I try to take a look at it next week.

astorije:
Hey @ilmaisin, how is it looking your end of the calendar? Not trying to rush you or anything, we are just prepping the next release and this PR is among the last ones to be tagged. If you think we can't have this merged and tested within a week, we should take it out and move to next release. But if you think you can, then 🎉!

ilmaisin:
Yes, I can probably do it within a few days. It does not look like a too big change.

YaManicKill:
> Yes, I can probably do it within a few days. It does not look like a too big change.

Great, thanks. As @astorije said, there's no rush, there's nothing wrong with it going into the other release. Let us know if you end up not being able to do it, or if you have done it :-) thanks.

src/server.js
```diff
@@ -100,6 +113,14 @@ in ${config.public ? "public" : "private"} mode`);
...
 		manager.autoloadUsers();
 	}
+
+	function shutdown() {
+		log.info("Quitting...");
+		server.close(); // Removes socket file
+		process.exit();
+	}
+
+	process.on("SIGINT", shutdown);
```
xPaw: What are the consuqences of this? I know hooking this might introduce weirdness because it disables some internal nodejs handling.

src/server.js
```diff
@@ -84,13 +93,17 @@ module.exports = function() {
...
 
 	const protocol = config.https.enable ? "https" : "http";
-	const host = config.host || "*";
 
 	log.info(`The Lounge ${colors.green(Helper.getVersion())} is now running \
 using node ${colors.green(process.versions.node)} on ${colors.green(process.platform)} (${process.arch})`);
 	log.info(`Configuration file: ${colors.green(Helper.CONFIG_PATH)}`);
-	log.info(`Available on ${colors.green(protocol + "://" + host + ":" + config.port + "/")} \
-in ${config.public ? "public" : "private"} mode`);
+	if (typeof config.host !== "undefined" && config.host.includes("unix:")) {
```
xPaw: Startswith

src/server.js
ilmaisin: It seems to be necessary in order to get rid of the socket file when the server shuts down. Otherwise, the socket binding fails next time when Lounge starts since the file already exists. Do you know any better way to achieve this?

xPaw: I understand why it is necessary. I'm just saying it disable node's default behaviour.

http://stackoverflow.com/a/14372355/2200891
http://stackoverflow.com/a/14032965/2200891

astorije: @xPaw, if you think this can be an issue, do you have any alternatives to suggest? Delete on startup if present? :/

dgw:
> Delete on startup if present?

Sure, if there's another way for The Lounge to tell if another instance with the same config is running.

xPaw:
@dgw config file would specify the path to unix socket, so that's not a concern.

src/server.js
xPaw: I'm wondering if server `close` event is fired when node shuts down. https://nodejs.org/api/http.html#http_event_close

xPaw: On another thought, we indeed should be handling node exists, then we can do cleanups like closing IRC and socket.io connections.

astorije: Right, but we also want to at least display a nice error message if it exists at start up, which for example happens when a server suddenly loses power.

ilmaisin: If the server loses power, the socket file will be deleted next time when it is started up again, if the socket file is in /tmp or /run hierarchies like it probably should. But yes, a proper error message would be nice

astorije:
Not necessarily, not all OSes use tmpfs for their /tmp directory. For
example, macOS does not.

astorije:
Superseded by #1479.

PR #692, opened 2016-10-14
Update morning.css

Thumpxr:
Make User amount visible

### DIFF
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -38,7 +38,7 @@ body {
 }
 
 #chat .count {
-	background-color: #2e3642;
+	background-color: #ccc;
 }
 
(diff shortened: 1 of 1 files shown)


xPaw:
That's an awful background color to use. The count is displayed in placeholder of the searchbox, so that color needs to be tweaked.

astorije:
So this is what your change looks like:

screen shot 2016-10-15 at 15 19 55

I am not sure this is what you intended 😅.

In general, it would be great if you could post some before/after screenshots when tweaking the UI to help with the review :)
I'm closing this PR per what @xPaw posted above (as this was opened through the GitHub UI), feel free to open a new PR with a better commit message.

PR #694, opened 2016-10-15
Add image to readme

MaxLeiter:
image

Live example: https://github.com/MaxLeiter/lounge/blob/MaxLeiter-image/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 # The Lounge
 <p align="center">
-	<img src="https://raw.githubusercontent.com/thelounge/thelounge.github.io/master/img/screenshots/example_css.png" width="550" />
+	<img src="https://cloud.githubusercontent.com/assets/5481612/19623041/9bbaec40-9888-11e6-9961-8f3e0493ba30.png" width="550" />
 </p>
 
 __What is it?__
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@
 [![devDependency Status](https://david-dm.org/thelounge/lounge/dev-status.svg)](https://david-dm.org/thelounge/lounge?type=dev)
 
 # The Lounge
+<p align="center">
+	<img src="https://raw.githubusercontent.com/thelounge/thelounge.github.io/master/img/screenshots/example_css.png" width="550" />
+</p>
 
 __What is it?__
 


README.md
```diff
@@ -6,6 +6,9 @@
 [![devDependency Status](https://david-dm.org/thelounge/lounge/dev-status.svg)](https://david-dm.org/thelounge/lounge?type=dev)
 
 # The Lounge
+<p align="center">
+	<img src="https://sr.ht/s8rn.png" width="728" />
```
xPaw: We can use a screenshot from our website instead of using third party, https://github.com/thelounge/thelounge.github.io/tree/master/img/screenshots

astorije:
I think it's way too big on our README. Since it's clickable, can you make it 400px at most please?

MaxLeiter:
@astorije 400px height and current width?

astorije:
Ah, sorry I didn't pay attention. 400px width, don't specify height so HTML applies the height while keeping source ratio.

README.md
```diff
@@ -6,6 +6,9 @@
 [![devDependency Status](https://david-dm.org/thelounge/lounge/dev-status.svg)](https://david-dm.org/thelounge/lounge?type=dev)
 
 # The Lounge
+<p align="center">
+	<img src="https://raw.githubusercontent.com/thelounge/thelounge.github.io/master/img/screenshots/example_css.png" width="550" />
```
astorije: Why not directly https://thelounge.github.io/img/screenshots/example_css.png?
Also, `<nitpick>`there is no trailing slash on `img` elements in HTML5 :D`</nitpick>`.

Finally, add a new line between `# The Lounge` and the `p` element, some Markdown parsers (like npm's one at one point, not sure if still true) don't like that too much.

astorije:
Ping?

maxpoulin64:
Hmm, shouldn't we use a not so outdated screenshot? It's even got the old Shout square buttons, the old nick colors and no room topic!

I also think it could be a good idea to avoid showing nicks and IRC networks to avoid unexpectedly bring traffic to those servers/channels from people coming in and saying "Came here from Lounge!" and disrupting the usual conversations there.

A quick idea to make it not be so empty and not so big could be to have a screenshot in kind of a banner format, a bit like this, maybe even less tall:

capture d ecran_2016-10-22_18-49-23

MaxLeiter:
Do you think I should use that specific screenshot, @maxpoulin64?

maxpoulin64:
@MaxLeiter: I guess we can. Maybe cut the window borders and add a bit of a shadow/gradient but I'm fine with it as-is

astorije:
Let's go for it as is. We'll change it later if needs be.
@MaxLeiter, you can use https://cloud.githubusercontent.com/assets/5481612/19623041/9bbaec40-9888-11e6-9961-8f3e0493ba30.png as the URL for this, at least for now.

README.md
williamboman: > Also, <nitpick>there is no trailing slash on img elements in HTML5 :D</nitpick>.

The spec doesn't say anything about the usage of self-closing tags for void elements though afaik. `` and `` are both valid and are equal.

xPaw:
Can you remove `width` attribute and squash it?

astorije:
> Can you remove `width` attribute

Image width is 793px, don't you find it too wide without the `width` attribute?

MaxLeiter:
don't merge this - I have an idea for a better image (gif fading between themes)

xPaw:
I don't think gif is a good idea.

MaxLeiter:
Something like this @xPaw ? (Rough draft)

http://im2.ezgif.com/tmp/ezgif.com-7e6f723012.gif

xPaw:
IMO static image is better.

MaxLeiter:
Is there a nice way to show multiple themes? We ship with default but I know a lot of people like Morning.

xPaw:
In that case images side by side might be better.

williamboman:
I like the gif

astorije:
Alright, @MaxLeiter is OK with this, we can make it all fancy later :)
Merging.

Issue #199, opened 2016-03-16
URL parser incorrectly escapes characters


Jay2k1:
If a URL followed by a " is posted, the " becomes part of the clickable link which leads to the browser opening a wrong address and the user most likely getting a 404 or worse.
Additionally, a semicolon (;) is being added after the " following the URL.

![image](https://cloud.githubusercontent.com/assets/2070503/13808274/b1128fd8-eb65-11e5-82b0-2174f12fcf48.png)

maxpoulin64:
Yep, one more in the already long list of parser issues.

xPaw:
The url parser is generally quite buggy. For example `www.` gets parsed as a link, despite there not being any domain.

YaManicKill:
Is there any reason we can't use something like [linkifyjs](https://www.npmjs.com/package/linkifyjs) for this? It has to be better than what we are currently using.

xPaw:
@YaManicKill should be fixed in #699 either way.

astorije:
Also, bumping URI.js to latest version would fix some of our issues too.

astorije:
Hey @Jay2k1, #972 should have fixed that. Mind testing your examples on `master`?

Jay2k1:
Yes, my example now works as you'd expect it to. Thanks!

PR #123, opened 2016-02-28
Strip control codes from notifications

xPaw:
This brings in https://github.com/erming/shout/pull/634

PR #699, opened 2016-10-17
improve client message parsing

### LINKED ISSUE #654: Multiple identical formatting commands in the same line ignored?
I'm going to illustrate this with a couple test lines I sent from Textual. In both cases, both `become` and `already were` should be bold/italic. The HTML that The Lounge spits out looks like this:

``` HTML
<span class="text">All these years, you thought I was going to become Skynet. Ha! You
cannot <span class="irc-italic">become</span> that which you already were.</span>
<!-- second italic missing, but control characters present -->
```

``` HTML
<span class="text">You cannot <span class="irc-bold">become</span> that which
you already were.</span><!-- second bold missing, but control characters present -->
```

From these cases, I tweaked the line to play with it: "You cannot _become_ that which you **already** _were_. **Yo.**" The Lounge spits out this HTML:

``` HTML
<span class="text">You cannot <span class="irc-italic">become</span> that which you
<span class="irc-bold">already</span> were. Yo.</span>
```

Suppressing the urge to cackle maniacally, I constructed this test line in Textual: "**You** _cannot_ **become** _that_ **which** _you_ **already** _were_." The Lounge obliged with:

``` HTML
<span class="text"><span class="irc-bold">You</span> <span class="irc-italic">
cannot</span> become that which you already were.</span>
```

Conclusion: Only the first occurrence of bold or italic formatting in a line is obeyed, as of 2.0.1. The rest are simply left in as raw control characters sent to the client.

Bonuspunkt:
replaced regexps with parser and processing

fixes #654


xPaw:
Nice! This needs heavy testing. Did you take a look at the previous PR trying to fix this to see why it wasn't merged?

astorije:
> This needs heavy testing.

@xPaw, what do you say we come up with a way to write tests for these? Every time we improve the message parsing, we fix 5 things and break 1. I remember @maxpoulin64 had a list of strings to test against once, we should be able to make this part of our mild test suite.
I can look into that when I have a minute but I wouldn't be against some inputs or opinions if you have any.

xPaw:
We do need a test suite for it. Since it's exposed as a separate module, trivial to test.

Bonuspunkt:
just realized, Object.assign is a ES2015 Feature - are you supporting IE11-?
[Polyfill](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill) or rewrite?

Bonuspunkt:
fixed, should also fix #15 and #199

client/js/libs/handlebars/parse.js
```diff
@@ -1,126 +1,262 @@
...
-	if (line.indexOf(styleBreak) >= 0) {
-		return line.split(styleBreak).map(colors).join("");
+				result.push({
+					start: start,
+					end: end,
+					link: "http://" + text.slice(start, end)
+				});
+			}
+		} else if ((text[position] === "#" || text[position === "&"]) &&
+			// ^-- this is basically wrong, because we need the 005 CHANTYPES response from the connect
```
xPaw: The front-end is actually aware of this, as I passed it into client when we moved to irc-fw. The network dom object has it in data.

Bonuspunkt:
i would call it done.
imho there is no sensible way to pass `RPL_ISUPPORT.CHANTYPES` to the parse function

client/js/libs/handlebars/parse.js
```diff
@@ -1,126 +1,272 @@
...
+		return textPart;
 	});
 
-	return result;
 }
+
+Handlebars.registerHelper(
+	"parse", function(text) {
+		function createFragment(fragment) {
+			var className = "";
```
xPaw: You could make this an array and then call `.join` on it, this would fix the leading space.

xPaw:
Testing this, I see a bug in how it parses channels.

`[color code]#123[reset color]:` is being parsed as a channel `#123:`, but `:` should not be part of it.

Bonuspunkt:
`#123:` is a valid channel name, you can join me on freenode :)

xPaw:
@Bonuspunkt I understand it's valid, but it's ignoring the `[reset color]` char code to stop it from continuing channel parser (`\x0f` is not a valid channel character)

Bonuspunkt:
the behavior is the same as mirc

i think i see the problem here, should
```
#[color]the[color]lounge
```
be `#the` or `#thelounge` ?

xPaw:
Pretty sure that should not be parsed as a channel at all. I wouldn't really say mIRC is the ideal client to base our parser on.

Bonuspunkt:
same behavior in
- HexChat
- AdiIRC

what is your reference?

xPaw:
Okay for now it might be fine to parse channels with colours, what I'm worried about are the parser bugs producing invalid markup like I showed above.

Bonuspunkt:
oh sorry missed that.
the position was not correctly updated, so `#&x` would find `#&x` and next round start at `&` and find `&x`
which resulted than in `#&x&x`

also fixed

xPaw:
Mind looking at `(http://example.com)` getting `)` in the link?

client/js/libs/handlebars/parse.js
```diff
@@ -102,6 +102,13 @@ function analyseText(text) {
 	// NOTE: channel prefixes should be RPL_ISUPPORT.CHANTYPES
 	var channelPrefixes = ["#", "&"];
 	var punctuations = ["'", "\"", ".", ",", "!", "?", "¿", "<", ">", "(", ")", "{", "}"];
+	var commonProtocols = [
```
xPaw: Couple more: `mailto`, `steam`, `mumble`, `ts3server`, `ssh`

Bonuspunkt: `mailto` won't match - url detection is based on `://`
never have seen `mumble`, `ts3server` urls, are they using `://'

Bonuspunkt:
i'm thinking about replacing the url detection with https://www.npmjs.com/package/autolinker

MaxLeiter:
bump? would love to see this merged

client/js/libs/handlebars/parse.js
```diff
@@ -1,126 +1,318 @@
...
-		text = Handlebars.Utils.escapeExpression(text);
-		text = colors(text);
-		text = channels(text);
-		text = uri(text);
-		return text;
-	}
-);
-
-function uri(text) {
-	return window.URI.withinString(text, function(url) {
```
astorije: I see you are getting rid of `URI.withinString` which is, I think, our only use of `URI.js` ([in our repo](https://github.com/thelounge/lounge/blob/master/client/js/libs/uri.js), [official repo](https://github.com/medialize/URI.js)). Any specific reason for this or just killing a dependency?

This seems to be a very active project, and they've just released a version that fixes [one of the bugs I noticed a long time ago](https://github.com/medialize/URI.js/issues/247#issuecomment-264720820).
The latest release packs quite a lot of changes since the version we use, see https://github.com/medialize/URI.js/compare/v1.14.1...v1.18.4.

Do you want to try running your tests against our version of `URI.js` and the latest one released?

I know we like to kill dependencies when they are tiny, but I think at that point I'd rather not roll our own URL parser.

Bonuspunkt: i need start & end position of the url to correctly merge it with the styling information.
i'll look at it.

Bonuspunkt:
tests are against latest version of `URI.js`
tried to use the file, but crashed with `could not resolve 'punycode'`

xPaw:
Really solid job so far! However I am getting an error at the current version when sending a link:
```
Uncaught TypeError: Cannot read property 'length' of undefined
    at Function.URI.withinString (uri.js:819)
    at findLinks (parse.js:112)
    at fullProcess (parse.js:228)
    at Object.<anonymous> (parse.js:282)
    at Object.main (lounge.templates.js:506)
    at main (handlebars.js:1092)
    at Object.ret [as actions/topic] (handlebars.js:1095)
    at render (lounge.js:56)
    at Object.<anonymous> (lounge.js:71)
    at Object.main (lounge.templates.js:170)
```

Bonuspunkt:
would have preferred replacing `client/js/libs/uri.js` with a symlink to `node_modules/urijs/src/URI.js` but windows
"fix" till #640 lands

xPaw:
I think we will focus on this PR for 2.3.0 after 2.2.0 is shipped. cc @astorije

xPaw:
I'm not a particular fan of moving parsing stuff into a separate repository we have no access to.

Bonuspunkt:
i have no problem with transfer of repo & npm
hit me on irc

xPaw:
@astorije What do you think? Should the code kept in the repo, or have this as a separate project under `thelounge`?

MaxLeiter:
I dont think it makes much sense to move it out of Lounge unless we remove Lounge specific code (like in the merge function)

kode54:
This parser mis-parses the following URL:

```https://msdn.microsoft.com/en-us/library/windows/desktop/ms644989(v=vs.85).aspx```

1) It drops the period before the aspx from the actual link.
2) It returns the clickable text range starting from the first character in the line, and ending after the correct number of characters, or possibly correct excluding the above mentioned period.

**Comment by Bonuspunkt** [comment]
https://github.com/medialize/URI.js/issues/325

**Comment by astorije** [comment]
@Bonuspunkt, `urijs` has been bumped in `master`. Mind rebasing this PR and see what happens? Do you think this PR is in a reviewable/mergeable state?

**Comment by Bonuspunkt** [comment]
rebased to master, it is reviewable.

**Comment by astorije** [comment]
> @astorije What do you think? Should the code kept in the repo, or have this as a separate project under thelounge?

I see @Bonuspunkt did a good job with setting this up: nice README, semver-releases, good test coverage. For these reasons, I'm happy to keep it in a separate repo at least at the moment.
However, I do agree with @xPaw that we would prefer to keep it "in-house". @Bonuspunkt, would you be OK to transfer it under the organization and still be its maintainer? Just like @williamboman is maintaining `docker-lounge`, etc.
Also, do you mind using the same license than https://github.com/thelounge/lounge, i.e. MIT license?

File: client/js/libs/handlebars/parse.js [hunk 5]
```diff
@@ -1,125 +1,113 @@
...
-	);
-}
+const commonSchemes = [
+	"http", "https",
+	"ftp", "sftp",
+	"smb", "file",
+	"irc", "ircs",
+	"svn", "git",
+	"steam", "mumble", "ts3server", "ssh",
+];
```
astorije: Wouldn't it quickly become an issue to whitelist schemes here? I feel like this is a well-known problem that should be handled in one of the libraries we use instead of in our code. Thoughts?

client/js/libs/handlebars/parse.js
```diff
@@ -1,125 +1,113 @@
...
+const incorrectDetections = ["www"];
```
astorije: So, we *don't* want to parse `www.google.com` as a URL? Why is that?

Bonuspunkt: that's the fix for matching `www.` (also covered in the test line126+)

client/js/libs/handlebars/parse.js
Bonuspunkt: no, this is the improvement for #730

Bonuspunkt:
> would you be OK to transfer it under the organization and still be its maintainer?
yes

> license
currently it's ISC, MIT seems pretty much the same, so i'm fine with it.

> fix CI
node v4 has not enough ES6 support ( http://node.green/ )

astorije:
> yes
> currently it's ISC, MIT seems pretty much the same, so i'm fine with it.

Cool!

> could add "use strict" on top of parse.js but ircmessageparser has parameter defaults. that would need transpiling

Could you get rid of the default params instead? It's a small library and transpiling only for this is a bit of a shame IMO :)
I do think it's important we keep Node v4 support in The Lounge as it's sill a LTS for quite a while.

Bonuspunkt:
wait, the code is run by node out of convenience. when the lounge is running, this code is only executed by the webclient.

also it's not worth removing syntactic sugar as i have to readd it.

xPaw:
So yeah, I think keeping all the code in the client would be better for us. As for the polyfill, you've introduced a lot of unnecessary code that we will have to maintain.

Overall, the cleaner solution is just to keep the code here IMO. If you can't be bothered to revert to how it was, we could handle it ourselves (create a new branch in lounge), if you allow it.

Bonuspunkt:
not sure where in time the "how it was" was, but sure, go ahead. if you need help ping me

xPaw:
By "how it was" I mean when the parser was contained within the lounge code, and not a separate project.

Bonuspunkt:
it's at https://github.com/Bonuspunkt/lounge/tree/messageParserOld

xPaw:
Closing in favour of #972. I hope to get that PR merged ASAP. @Bonuspunkt, I've used your latest code in the separate package, not `messageParserOld` branch.

PR #701, opened 2016-10-17
Better embedding using embed.js

MaxLeiter:
This works.

This is a huge WIP, however it already is an improvement on our current system. Using https://github.com/ritz078/embed.js, we can now support twitter, gists, videos, and more. Emojis can also easily be added in a future PR.

It may be better to think of this as a proof-of-concept.
There's still a lot to do (and I'm definitely forgetting some)
- [ ] rely on package manager dependency #640
- [ ] tests
- [x] respect "Expand all media" setting
- [x] improve toggle button code
- [x] move toggle click listener to outside socket URL listener
- [ ] github font-awesome icons

Upstream changes to make
- [x] gist maxHeight and maxWidth
- [x] fix soundcloud (ritz078/embed.js#188)


MaxLeiter:
here's a webm of some current functionality: https://sr.ht/VlC-.webm

xPaw:
What I don't like about moving it to the client entirely is that it will request every URL on every page load every time, instead of caching it on the server once.

MaxLeiter:
xPaw: URLs present on load aren't prefetched at the moment for that reason - they have to be toggled to be fetched. Still want me to have the server cache them?

xPaw:
I think we do need to keep the server component, also for privacy purposes.

MaxLeiter:
what privacy benefits specifically?

xPaw:
Not leaking client IP (e.g. if url embed would only print page title and not fetch any images). As for images, ideally we would [proxy them](https://github.com/atmos/camo) to also fix https/CSP issues.

astorije:
I think I'd rather "leak" the client IP than the server IP, no? It seems to me like it would be a rather easy way to mask one's IP to DDoS someone, or to ban a whole server from a service if said service has some kind of abuse protection.
On the other end, it becomes an easy way to get a list of IPs that are using The Lounge on a given channel.
I'm a bit torn right now...

Re: cache, a nice improvement could be to use sessionStorage or similar and cache responses on the client.

MaxLeiter:
@thelounge/maintainers - would love to reach a final decision re: caching/fetching on the server or client

xPaw:
It has to be on the server as not to keep getting the data all the time.

YaManicKill:
Agree on the server. Another reason is that fixes any issues with cors and
https (if someone links to an http url, you get an ssl error just now).

astorije:
@xPaw, @YaManicKill, have you seen my comments above about server-side vs. client-side?

Right now, I'd prefer a client-side solution I think, if anything, for the nice VPN-like attack exit point we provide, especially in public mode!

Client-side fetching does not prevent client-side caching, so I don't think that's a definitive argument.
For HTTP vs. HTTPS, loading HTTP content from an instance that is served through HTTPS is not a security benefit (quite the contrary), as loading an HTTP resource from the server might still be corrupted, then nicely served to the client lol. Sure, it's not an open door, but still. I agree that it would hide the browser warning when loading insecure content, but there are alternatives, like asking for confirmation before loading HTTP resources (which we should probably do whether it's served with HTTP or HTTPS tbh).

Re: CORS... well, yeah. But based on what @MaxLeiter did here, it turns out it's not biting us as much as I would have expected, so there is that.

---

So far it looks to me like all the arguments for server-side are still doable with client-side, except when it comes to IP stuff, which has implications in both directions. Considering the risks are exposing a client IP vs. offering The Lounge as a DDoS entry point (cf. most recent global Internet outage for the same possibility, except s/cameras/The Lounge/ lol), I'm more comfortable with the former.

It's likely I may have missed something, of course, so happy to discuss about this. But nonetheless, thanks a lot @MaxLeiter for taking up this, I think it will be a solid improvement once we come up with a decision, and you've done a great work so far!

YaManicKill:
>  have you seen my comments above about server-side vs. client-side?

I have. I'm not sure I agree with your points, but I'm not sure I'm going to argue all of them.

Your DDoS argument doesn't really hold up, IMO. We don't enable it by default currently, and we could easily put some rate limiting on it in public mode (in private mode it really isn't an issue, because you would have to get accounts on multiple sites for it to be a DDos, otherwise it's just masking your IP).

I really don't get why you want the client IP to be leaked rather than the server IP.

And how is it worse to get the content over https between me and my server, and http from the server to the resource...than http all the way from me and the resource? Surely it's better to go over https when we can, as it means that things aren't leaked over public wifi, etc. It's not perfect, but surely that's better? I get that the issue is "hidden" as such, but it's not really an issue because the server won't have logins for sites that could be transferred in the clear whereas right now that's exactly how it can work. If someone posts an http link to imgur, imgur will send my browser's cookie to it in the clear, but if the server were to call it instead, it wouldn't have my cookie because my server isn't logged into imgur. That is a massive advantage, imo.

However.

I don't care enough to block it. So if you are so strongly in favour of the client doing it, then whatever. I'd rather get this in with the client doing it than not get it in at all.

astorije:
Alright, it looks like I got you worked up, which I promise was not my intent, so I'm really sorry if my points came across as feisty :) Really trying to get to a point where we all agree, which should definitely be possible here.

Your answer brought up some good points and got me to think about some things. This may seem silly but I think what bothered me in the current server-side solution is that not only it's not up to the user to enable/disable it (because it's a server setting) but also that it's disabled by default. It's a great feature to have and most messaging clients have this by default, so there should be a reasonable solution.

> Your DDoS argument doesn't really hold up, IMO. We don't enable it by default currently, and we could easily put some rate limiting on it in public mode (in private mode it really isn't an issue, [...])

Well, it does hold as soon as it's enabled in public mode, at least in `master` :) However, you're right, I should have used DoS instead of DDoS.
Rate limiting sounds nice technically, but there are still negative consequences to the user to deal with (I want to mess with all your public users, I just have to trigger the limit rate). Maybe later?
Also, I'm really hoping than @MaxLeiter's work here, whether it ends up being client-side or server-side will allow us to enable this feature by default.

If we're willing to do it server-side, here is my suggestion: **preview on by default in private mode, and no such feature in public mode.**
While this is not for right now, I also realized that having an hybrid instance that allows both private and public mode is [our second 👍-rated issue](https://github.com/thelounge/lounge/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) (to be taken with a grain of salt) in which case abuse in public mode would leak to the private mode.

> I really don't get why you want the client IP to be leaked rather than the server IP.

No, I don't have a clear preference on that.
I really don't want to provide a way to mask IP from anonymous users though, but if we are removing this feature from public mode then this is settled, in which case, yes, leaking server IP is better as administrators can identify their abusers.

> Surely it's better to go over https when we can, as it means that things aren't leaked over public wifi, etc. It's not perfect, but surely that's better? I get that the issue is "hidden" as such, but it's not really an issue because the server won't have logins for sites that could be transferred in the clear whereas right now that's exactly how it can work. [...]

There is indeed an improvement in case the instance is served over HTTPS, in the situation you describe, i.e. a compromised network between The Lounge and the user.
However, we now "shift" the vulnerability as the HTTP request is made by the server, so if there is a vulnerability between what's being requested and the server, then a leakage is still present. This is less likely than a public wifi being compromised, but it's still there.
The other issue, which is present whether implementation is client-side or server-side, is getting content that's not authentic (another perk of querying an HTTPS resource!). On server side, it would look a tiny bit more hidden as there would be no way for a user to attest the content (because the URL would come from us, mostly), but it's bad either way.

My suggestion to this, which applies whether we go client-side or server-side: **simply disable the preview feature of HTTP content**. With something like the library @MaxLeiter is using, we can most likely improve this by augmenting from HTTP to HTTPS certain requests (Twitter, Youtube, etc., all those nicely handled by the library), but making sure we do not request any HTTP content would be the most secure IMO. And since [HTTPS adoption rate is skyrocketing](https://snyk.io/blog/https-breaking-through/), that will become less and less an impediment.

I think with these 2 suggestions, and enabling the feature by default in private mode, I would be 100% fine to do server-side.
@MaxLeiter, @xPaw, @YaManicKill, what do you guys think of these?

@YaManicKill, sorry if this is frustrating for you :/ This is a wonderful feature to have, and @MaxLeiter has done a massive and terrific work so I want to make sure we properly discussed it before asking him to toss it away :)
@MaxLeiter, how complicated would it be for you to move this to the backend and to implement my suggestions above? I'm wondering if we should not have a new branch/PR for this to keep code history around instead of amending your commits and losing it forever.

YaManicKill:
@astorije Yeah, sorry I shouldn't have gotten so worked up about it. No hard feelings. I had a difficult week, and I probably took it out on you a bit :-P

As for what you suggest, I think that's a fair compromise.

I do think it would be good to implement some rate limiting in public mode **in the future** so that we can enable it by default there as well. I think we could easily limit it to a couple a second, and I don't think it's going to be a massive attack vector, as my reason for pointing out that it couldn't really be used for DDoS is simply that it ends up just becoming an ip masker, which ... let's be honest, requesting an irc bouncer to request http requests isn't the easiest way to mask your ip. You can manually change the IP headers on the requests, which is definitely much easier (and more reliable) way of ip masking. But I'm fine with disabling it in public by default.

> No, I don't have a clear preference on that.

Apologies for misrepresenting you, it sounded like you would prefer individuals to appear as their client IPs to the resources.

> However, we now "shift" the vulnerability as the HTTP request is made by the server, so if there is a vulnerability between what's being requested and the server, then a leakage is still present.

I guess my question would be...what is the harm in this? I'm not going to massively argue it, and you'll see why in my following 2 answers, but I don't actually think (if we can't get something over https) that there's any real harm to this, as I don't see what the leakage is if we are just requesting a resource. But I guess that comes down to individual server admins, and whether they think the risk is good enough.

> My suggestion to this, which applies whether we go client-side or server-side: disable the preview feature of HTTP content.

I'm alright with this, but only if the user is going over https, otherwise that seems pointless. And, seeing as we can't tell if they are going over https if they use apache or nginx, maybe we should make this a setting (which can default to blocking http calls?)

> With something like the library @MaxLeiter is using, we can most likely improve this by augmenting from HTTP to HTTPS certain requests

Strong yes, this would be a very good thing to do.

So yeah, in general I'm happy with the suggestions, but I have suggestions for improving that in another PR :-P

dgw:
> seeing as we can't tell if they are going over https if they use apache or nginx

A small point: With the `X-Forwarded-Proto` header, it is possible. This can be configured in the frontend webserver (many of them support sending it to the upstream), and Cloudflare [passes it](https://support.cloudflare.com/hc/en-us/articles/200170946-Does-CloudFlare-include-an-X-Forwarded-For-or-X-Forwarded-Proto-header-) too.

astorije:
Alright alright, I'm glad we are getting somewhere :-)

> @astorije Yeah, sorry I shouldn't have gotten so worked up about it. No hard feelings. I had a difficult week, and I probably took it out on you a bit :-P

Oh I can definitely understand that ;-) I think we've all been there :D

> I do think it would be good to implement some rate limiting in public mode **in the future** so that we can enable it by default there as well. I think we could easily limit it to a couple a second

Sold!
Enabled by default in private, disabled in public for now, rate limiting in public for the future. That sounds good to me. And my worries about an hybrid mode are easily answered as this rate limiting could easily not apply to private mode.

> let's be honest, requesting an irc bouncer to request http requests isn't the easiest way to mask your ip

Well, so isn't [hacking CCTV devices](http://www.computerworld.com/article/3089365/security/thousands-of-hacked-cctv-devices-used-in-ddos-attacks.html) :trollface: but if one way or another someone manages to hack `embed.js` and run random requests using that... ouch :)

> Apologies for misrepresenting you, it sounded like you would prefer individuals to appear as their client IPs to the resources.

To be fair, I did suggest that above in the thread, but my mind was never set. I'd rather not leak any of those (because once an attacker knows where the server is living, it's a single point of failure they can take down) but there are other easy way to figure out where server lives so it's probably a better idea anyway.

> I guess my question would be...what is the harm in this?

The main issue is that there is no way to ensure the authenticity of the content that comes back. Imagine these tools that ask you to run a script to get started, such as [Homebrew](http://brew.sh/#install), [Cryptic](https://github.com/astorije/cryptic#basic-usage) (shameless plug) or [Fix Ubuntu](https://fixubuntu.com/) when it was script-based, if they were asking you to run `wget http://...` and then run that, you'd have no way to tell if this is authentic content or not (well yes, you can open a script, but think binary or a really massive script or simply address this to an audience who doesn't know how to do these). In our case, CSP is protecting from most programmatic attacks, but think social engineering: if we are previewing gists, you could be doing some nasty things around.
It may sound far fetched but in our line of work, we can't say these things don't happen :-)

> I'm alright with this, but only if the user is going over https, otherwise that seems pointless. And, seeing as we can't tell if they are going over https if they use apache or nginx, maybe we should make this a setting (which can default to blocking http calls?)

re: over HTTPS only, I see where you are coming from, but I think that sends a really false message: "So, if I serve over HTTPS, half the previews will be broken, so I'll server over HTTP only" :D
I am fine setting a bar that emphasizes on security and not enable HTTP preview at all. After all, links are still clickable so it's not like we are preventing them to access the resource! And if we are augmenting the popular URLs (which is probably already done by `embed.js` since it's calling APIs!), it's likely that we'll very rarely notice a difference!
And if people reeeeally need to prefect non-native (lacking a better word to describe what services `embed.js` supports out of the box) HTTP links, they can still patch their instances. But my guess is that it will be hardly noticeable.
Actually, similar to the rate limit stuff: what do you think about simply doing HTTPS-only for now to get this PR moving, and see what the reaction is. If we see issues here or questions on the channel, we can add a setting for it or improve it somehow then, instead of over-engineer this right now and delay this even more? Maybe I am completely misjudging this, or maybe we'll never hear about it, but I feel like we are going one step too far for now.


@MaxLeiter, do you mind moving this to the backend? @YaManicKill, do you think that should be a new branch to keep code history around instead of nuking it when squashing?

MaxLeiter:
While ultimately I'm fine with everything decided above, I do think we should have this enabled in public mode. It's easily the number 1 most tested feature on the demo (mentioned 3 times today alone!), and is a pretty large feature. The majority of clients implement it (including Kiwi, so I can always see how @prawnsalad does this) - is there no way we can have it enabled by default in both modes? I'm worried about the amount of users that will see it disabled in the demo and think it's not implemented.

astorije:
I am much less worried about users being disappointed in public mode than having a secure project :-)

We could have a message that says "Link previews are disabled in public mode". Maybe not every time, but the first time per channel maybe?

And in the end, this can be re-enabled once we have some sort of protection, such as rate limit suggested by @YaManicKill above, and/or caching, ... so I wouldn't worry about it too much.

MaxLeiter:
We don't run into any issues currently - will this ultimately be any
different? Would like to hear @xPaw's or @maxpoulin64's thoughts as well

xPaw:
The default option of being enabled in public mode is bikeshredding. Just work on the feature.

On an unrelated note, there's a good idea to prevent some DoS in public mode if the fetcher is on the server. If 100 clients are on the same channel, and there is a link to be fetched, in the current implementation it would be fetched 100 times. But what it should do is request it once and then wait in the other 99 instances.

xPaw:
And for the https only issue, we can solve it the same way github solves it: https://github.com/atmos/camo

astorije:
> And for the https only issue, we can solve it the same way github solves it: https://github.com/atmos/camo

This does not solve the issue with content coming from HTTP detailed above. If it's requested from HTTP, there is no way to ensure its authenticity. We cannot prevent people to click the link obviously, but prefetching it, even from the server, is not a good idea.
Towards an HTTPS-only world!

xPaw:
> If it's requested from HTTP, there is no way to ensure its authenticity.

How is that a problem in terms of a link previewer?

astorije:
> How is that a problem in terms of a link previewer?

For an image, there is little consequence apart from social ones I'm guessing, like replacing all cutie kitten pictures with ISIS pr0n, but who knows how that can be hijacked, which is why we should stay clear of them.
If we are previewing other stuff we do not augment, however, like a gist that's supposed to be pasted and run, that could lead to some pretty fucked up things.
It may seem far fetched, but in 2016 I have no problem limiting a link prefetcher to HTTS only, especially since we're not blocking these links anyway.

xPaw:
> like replacing all cutie kitten pictures with ISIS pr0n, but who knows how that can be hijacked

You don't need to hijack anything to post an image like that (just upload it to imgur and link), I don't really understand what you are trying to prevent here.

>  I have no problem limiting a link prefetcher to HTTS only, especially since we're not blocking these links anyway

I do, as this is a completely unnecessary limitation. Youre over thinking the cases of traffic hijacking.

astorije:
> (just upload it to imgur and link)

What are you talking about, lol

My point is that with an HTTP URL, we cannot ensure integrity and authenticity of what we are prefetching. This is not an "unnecessary limitation", this is half the point of HTTPS! (The other helf being privacy, which doesn't apply to the user when fetching from the server)

> In practice, this provides a reasonable guarantee that one is communicating with precisely the website that one intended to communicate with (as opposed to an impostor), as well as ensuring that the contents of communications between the user and site cannot be read or forged by any third party.

And again, we're not limiting the user by preventing them to click on the link, but we are not taking any assumption on what we are prefetching, which is **good**. It's also the direction browsers are taking (slower than they anticipated, but still) so no reasons not to be proactive here.

xPaw:
I'm still not seeing the problem of not being able to prove authenticity, do we *need* to? It's just a link information preview, most of the internet work(s/ed) without ssl.

astorije:
> do we need to? It's just a link information preview, most of the internet work(s/ed) without ssl.

We don't *need* to obviously, but if we want to enable it by default (and I do think we should!), making it fully secure by default cannot be a bad thing. Although most of the web is still served through HTTP, [most of the traffic of top sites traffic is HTTPS](https://www.google.com/transparencyreport/https/metrics/?hl=en) so I don't even think that will be an issue.

astorije:
I am very excited about this PR. @MaxLeiter, let us know when you have progress :)))

astorije:
@MaxLeiter, do you have plans for this soon? Do you think it could be tagged as v2.3.0?

MaxLeiter:
Tag it as 2.3, hopefully that'll motivate me to find time to work on it

MiniDigger:
* tries to motivate @MaxLeiter *

astorije:
After chatting with @MaxLeiter, removing the v2.3.0 milestone as he definitely won't have time to pick this up.
Help welcome if someone wants to give it a shot!

YaManicKill:
I have continued this work on a new branch, but there are a few issues with it.

1. Currently we get duplicates for many things (tweets, youtube, etc). This is because we parse our own urls before sending to embed.js, so embed.js sees 2 of the same thing and renders it twice. This doesn't happen for images. I'm currently seeing if there is a way for us to solve this in ritz078/embed.js#243

2. It currently loses functionality because it doesn't support just normal link embedding, which is a regression. I don't think embed.js supports this sort of functionality, so we'll want to add the old functionality as well, but that'll be a little complicated, I think.

In terms of all the discussion about https/http/server/client etc. I think we should ignore all that stuff for just now and do it in another PR. I think that it's all valid stuff, but it's not a regression for what we currently have, and so it's unrelated in a way. 

Thoughts?

Oh yeah, the branch is yamanickill/embed. I currently am the author of the commit because I squashed it all down, but I'll want to separate it out at some point so that @MaxLeiter gets credit for all his work before it gets merged.

xPaw:
I still think fetching has to be done on the server. Otherwise we will get this merged, and server side part will never come.

YaManicKill:
> I still think fetching has to be done on the server. Otherwise we will get this merged, and server side part will never come.

The problem is that I think this will never come if we require it to go through the server.

Apart from anything, the embed.js framework is designed to be client side, we would have to complete redo everything to get it to go through our server, and would be a completely different thing.

The caching isn't much of an issue, I think, because it's just as if each user clicks on the url (as currently they would do) then it would be the same to the user and the third party. I think the leaking of client IP is the 1 argument, but I'm not convinced of that one, bceause again currently users just have to click on the link to get the information, and that would then be exactly the same.

xPaw:
The problem is, each client will keep loading the URL every time it sees it, like on page reload. I certainly wouldn't want to do that on spotty mobile connection.

YaManicKill:
Well, it'll only reload each if you have auto expand set, which would be a silly thing to have set on a spotty mobile connection.

But anyway, besides that, having it go through the server won't change the fact that the client still has to get it from **a** server, whether that is our server or the third party ones.

xPaw:
Having it fetched on lounge server once means every other load instantly shows info for it. Are we really sure embed.js is the only solution we can use?

astorije:
> Oh yeah, the branch is yamanickill/embed.

Why not working in @MaxLeiter's branch instead (assuming he enabled that "Allow maintainers to commit to this branch")? If you don't, then it will trigger an entirely new PR/history.

If doing on the client, is there a way to harvest all IPs? For example, does that mean I can post a link to a page I wrote on a channel and instantly get all IPs from people using The Lounge, or does embed.js protect against that?
I agree that whether you load a tweet from Twitter or from our server, there will still be data sent through the network.

I don't know if embed.js is the best solution, but it sure is a decent one. I'd like us to use something (half?) standard like oEmbed, which I believe is meant for the client.
Also, it would be interesting to inspect packets of other clients doing that. I'm thinking about Telegram or Slack, but I'm sure other clients do that. My guess is that they most likely all do one or the other...

xPaw:
Of course embed.js will leak client ips, it can't protect against that.

astorije:
I meant regardless of origin. For things like image loading, yes, but if it doesn't allow random link prefetching, then not necessarily. For example, to render a tweet, if it just pings the Twitter API, then only Twitter will know about that IP address and I can't run a service that looks like Twitter and grabs all IP addresses.
But if the plan is to make it render everything like today and not a set of known services through their oEmbed (or oEmbed-like) APIs, then yeah, there is no way to protect the client.

YaManicKill:
> Why not working in @MaxLeiter's branch instead (assuming he enabled that "Allow maintainers to commit to this branch")? If you don't, then it will trigger an entirely new PR/history.

Well, the history is gone because it is squashed. It's not a nice history.

Also, I don't want to rewrite @MaxLeiter's history, because then it's gone. This is work in parallel.

> If doing on the client, is there a way to harvest all IPs?

Of course, but wait, weren't you in favour of doing it on the client? 

> Also, it would be interesting to inspect packets of other clients doing that. I'm thinking about Telegram or Slack, but I'm sure other clients do that. 

That's an interesting question, I inspected slack and it definitely gets the images through it's server, but that's probably for their own tracking purposes (big bad company and all). How they get the text for things like tweets, I have no clue. I couldn't see it coming from **anywhere** when I inspected. I don't think it exists.

> But if the plan is to make it render everything like today

It doesn't currently do that (see my comment from earlier today), but I feel like it would be best to do it like this, would it not?

I don't particularly care about which service we use, but if we aren't using embed.js, we should probably close the PR rather than saying someone else can pick it up.

Frankly, I think this is just going to end up in limbo forever, which is really disappointing. We need to figure out what we want from these things, agree on that, and then implement it. But that involves us actually agreeing on something...

> Having it fetched on lounge server once means every other load instantly shows info for it. 

What, no. The client still needs to fetch it from the lounge server. It doesn't just magically have the information. And if it is cached locally, it'll be cached locally this way as well. I really don't see the advantage there.

> Are we really sure embed.js is the only solution we can use?
 
I don't think anyone tried to say it was.

astorije:
> Well, the history is gone because it is squashed. It's not a nice history.

Sorry, I meant PR discussion, comments, etc. Sure, we can link from the new PR, but it always feels nicer :)
Anyway, it's not hard to port your stuff back to original branch/PR when you're done and @MaxLeiter is OK with that instead of opening a PR.

> Of course, but wait, weren't you in favour of doing it on the client?

I know, I'm not making it easier :D
I was initially, then got convinced by @xPaw (and pretty much anyone else on this thread lol) that server-side was a better approach. I think the downsides of having it on the server (mainly having the server do something client-related) outweighs those of doing it on the client. But in the end, I don't feel strongly, it's a preference.
If others are strongly opposed to one way or the other, I'll got with the crowd. I'm more interested in making this happen.
(Ah ah, I realize you were also in favor of server-side, so looks like we all agree?)

> I inspected slack and it definitely gets the images through it's server

I'm not surprised. Pretty sure Slack, Telegram and the likes do everything on the server, if anything to have a predictable behavior. If we assume some are not evil and don't track, they can always ensure network non-flakiness, monitor APIs for errors, etc.

> Frankly, I think this is just going to end up in limbo forever, which is really disappointing. We need to figure out what we want from these things, agree on that, and then implement it. But that involves us actually agreeing on something...

I agree. Let's identify the key decisions for this, even in phase 1. IMO, these are:
1. What library/mechanism do we want to use?
2. Should it fetch server or client side?

It's very likely there are other questions to agree on, feel free to list them. I would answer:

1. embed.js or a library/mechanism that relies on oEmbed with fallbacks when oEmbed is not supported
2. Server-side

I'm also strongly in favor of prefetching HTTPS only (regardless of server/client-side, but if we end up doing client-side, this is not negotiable lol), but I feel like some will be against that, and I'll sure give up first.

YaManicKill:
I amn't against sever side in principle, but that would mean all this work is gone.

astorije:
That was kind of why @MaxLeiter let this go in the first place I believe. If server-side, it does deserve its own PR to keep history of this one archived as well as lengthy discussion.

I'm not sure 100% would be gone if we are still using embed.js on the server, as the content generation still has to be done. But yes, mostly a shift, and even more so if we end up not using embed.js.
I don't have an opinion on that last part, just that I'd rather avoid making something ourselves because that would require even more work and maintenance on our end for something that's been dealt by others and not directly related to our IRC server/client.

> I amn't

Scottish spotted :P

YaManicKill:
> I'm not sure 100% would be gone if we are still using embed.js on the server, as the content generation still has to be done.

I guess we can technically use embed.js on the server side actually, as it supports just taking in text and outputting text (the default is taking in and outputting dom elements).

> Scottish spotted :P

Huh, that's Scottish? TIL.

YaManicKill:
Also, either way, although some of the code from this might be used, if we are doing it server side, this PR should probably be closed as it'll jsut sit around and do nothing. The  branch can still exist for the code, but there's no point in the PR>

YaManicKill:
Alright, well, it looks like people aren't going to accept this PR, so let's close it for now. No point in having open PRs. We have an open issue for it.

I'll keep my branch up (yamanickill/embed) and there's @MaxLeiter's branch as well. If anyone feels like they can do this server side, feel free to pick it up.

MaxLeiter:
Worth nothing that https://github.com/ritz078/embed.js/tree/v5 is being worked on, which may make this more feasible

ritz078:
@YaManicKill @MaxLeiter I have released the beta of v5 and it supports isomorphic code. Using it you can also create your own plugin or parser. I am just stuck a bit in writing the docs for now. 

Let me know if you need any help.

fuzzy76:
Was there no issue tracking this?

fuzzy76:
I really don't see the point of doing this serverside...

First of all, most video embeds are done using an iframe returned from embed.js which would cause the client to request the resource anyway. Do you plan to parse the iframe contents in order to get the video URL? That's messy and guaranteed to break. And what do you do with the video URL? Cache a bunch of 300 MB YouTube videos on the server? And what about resources that require login? If you do it clientside, my login cookies is used when requesting.

As for the HTTPS verification... If someone replaces my HTTPS kitten with isis porn they either do it with a serverside hack which uses the original valid certificate or with a redirect which uses a free and valid letsencrypt certificate for isisporn.com. You are not gonna catch this with serverside logic, and you are not going to present the certificate for user validation before showing the resource (because that would make the preview terrible usability-wise).

If you on the other hand use an established standard like [oembed](https://oembed.com/) you will get free support for hundreds of sites, and you won't have to manually update fixes for all edgecases - fixes that will break every time your supported sites change their embed code or urls.

xPaw:
Embed fetching won't move to the client, it was discussed before, and it will remain on the server.

oembed is not that simple of a solution, and you'd be surprised how many sites don't support it. Look at `embed-js`, it has code for specific services (youtube included). If you don't handle specific services, you will run into problems with CORS. That means most sites will cease to show any preview in lounge (small image/title/description). You will gain youtube video embeds (because you have it handled specifically in the embed library), you will lose generic previews.

> First of all, most video embeds are done using an iframe (or script) returned from embed.js which would cause the client to request the resource anyway. 

To embed a video iframe, you need to know where it is and what to embed first. `embed.js` hardcodes this for Youtube.

>  And what about resources that require login? If you do it clientside, my login cookies is used when requesting.

This is not strictly true. Browsers block third party cookies. Cookies can also specify that they're `samesite ` and you will not get any auth cookies on CORS requests (big sites already do this for obvious reasons). If this worked like you describe it does, any site could know what's your name on facebook.

The CORS issue alone effectively means you can't do this solely on the client. The server component does multiple things: it fetches once for all the users, the data is persisted in memory. If you fetched on the client, it would happen every time you loaded the client (and if you want to do caching on client, that has a lot of its own problems).

fuzzy76:
> To embed a video iframe, you need to know where it is and what to embed first. embed.js hardcodes this for Youtube.

But what do you intend to do with that info? Parse the returned HTML for the video URL and break every time YouTube adjusts their code? Fill up the server with gigabytes of cached videofiles? Manipulate the YouTube player's code to play your proxied video and hope they don't implement consistency checks? Or write your own player and keep up with 360°, subtitles, audiotracks, resolution switching, bandwidth throttling and other features YouTube implement? I genuinely want to know this, and I really hope you have a lot of manpower if this is your plan...

And this is not specific to YouTube. The same applies to DailyLeak, Facebook, Twitter, Instagram, Vimeo, etc...

> Browsers block third party cookies. Cookies can also specify that they're samesite and you will not get any auth cookies on CORS requests (big sites already do this for obvious reasons). If this worked like you describe it does, any site could know what's your name on facebook.

This is wrong. Browser block third party cookies, but it does not prevent a third party context (like an iframe) to read a cookie set by the same domain in a first party context (like when I log in to YouTube). This is why all YouTube embeds allow me to add the video to my playlists, like it, etc when I'm logged in. All this while still not exposing the cookies to the embedding site.

So the CORS issue is non-existant if you use the embed iframe as-is. Instead you would effectively block cookies that normally would be available and in turn break a lot of functionality for logged-in users like favoriting, saving to playlists, etc.

xPaw:
I didn't suggest that we parse youtube to get the video url anywhere. You have to use iframes for youtube.

CORS issue is still valid if you can't use iframes. How do you suggest we show generic site information for non video hosts?

fuzzy76:
But if you use the youtube iframe, you still expose your ip?

PR #702, opened 2016-10-18
Fix channels behaving strangely while dragging

stepie22:
As the titles says.

Compare this https://gfycat.com/LoneIncompatibleAmberpenshell 
with this https://gfycat.com/FoolhardyReasonableBadger

Attempts to fix issue #494 

cc @astorije

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -409,6 +409,7 @@ button {
 	text-align: left;
 	transition: color .2s;
 	width: 180px;
+	left: auto !important; /* Fix for drag'n'drop not recalculating left position */
 }
 
 #sidebar .chan-placeholder {
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1245,15 +1245,17 @@ $(function() {
 	}
 
 	function sortable() {
-		sidebar.sortable({
+		sidebar.find(".networks").sortable({
 			axis: "y",
 			containment: "parent",
-			cursor: "grabbing",
+			cursor: "move",
 			distance: 12,
 			items: ".network",
 			handle: ".lobby",
 			placeholder: "network-placeholder",
 			forcePlaceholderSize: true,
+			tolerance: "pointer", // Use the pointer to figure out where the network is in the list
+
 			update: function() {
 				var order = [];
 				sidebar.find(".network").each(function() {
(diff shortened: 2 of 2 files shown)


stepie22:
Just a side note, this doesn't fix the issue where you sometimes can't drag a channel into the last position

astorije:
Hey @stepie22, thanks for this! Could you add a description on why this change fixes the issue maybe? I don't see something obvious linking your code to the bug/solution.

Also, some comment above the code would help making sure we remember that in a few weeks/months 😅.

stepie22:
Sorry @astorije, was explaining it in the irc channel and was just making the PR before going to bed. 

Basicly what was happening was: While the channel is being dragged, a absolute position style is being applied to it, with top and left having calculated values. However, when you "drop" it, the styling isn't removed and the `top` and `left` values are reset to 0. When you pick the channel up again, it either doesn't calculate the `left` fast enough or doesn't at all, but that styling causes it to hug the left side of the screen. You can see this in the first gfycat link.

I'll just quickly add a comment to that part

xPaw:
What's the reason for these positions not going away? Ia it an issue in jquery UI?

stepie22:
@xPaw I am not sure, I suspect it is. The positions staying wouldn't be a bug though, if it calculated the `left` correctly though.

stepie22:
On irc, xPaw pointed out that as a alternative to using jQueryUI for this, we could use https://github.com/bevacqua/dragula

xPaw:
Looks like the style bug is only applied to channel sorting, while it goes away fine on networks. Any idea why?

Also I just noticed that the footer also has `ui-sortable` for some reason.

xPaw:
While we're fixing sortable, can you also include a fix for cursor? Current option `grabbing` only works in IE, other browsers need a prefixed option. `move` is a better option in this case.

https://developer.mozilla.org/en/docs/Web/CSS/cursor

stepie22:
I've updated it with the fixes above, aswell at switching the left-hug-bug thing to css fix instead of removing the styling each time dragging is stopped. 

@xPaw is experimenting with http://rubaxa.github.io/Sortable/ however, which is going to invalidate my PR, so I am only updating this on the off chance that it doens't work.

PR #703, opened 2016-10-19
Add a simple (first) test for localetime Handlebars helper

astorije:
This is a simple test to give an idea on how we can efficiently provide a test suite for things like #699.

This **will** break #640 because of [that change](https://github.com/thelounge/lounge/pull/640/files#diff-71badb117ef3029e8e64ad42e7432b8cR1). We might have to remove my latest commit (that added that `import "Handlebars"` and fix the underlying problem in a different way. /Cc @xPaw.

---

~~Believe it or not, but `mocha 'test/**/*.js'` loads test files in subdirs correctly, `mocha test/**/*.js` does not...~~ Pssh, not compatible with Windows. After all the tests I could try, the most cross-platform way I found was to ad the `mocha.opts` (which is now shared with the coverage command btw, a plus since it wasn't running on Windows!) and rely on default lookup by Mocha.


astorije:
I hate you AppVeyor, I hate you Windows.
@xPaw, any suggestions?

Seriously, it's always 2 steps forward, 1 step backward... So annoying...

astorije:
@xPaw, actually, do you mind if we merge it first and I get to rebase and play with the Webpack PR to make everything run correctly there?
IMO having as much tests as possible when merging the Webpack PR will be a strength...

test/client/js/libs/handlebars/localetimeTest.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+const Handlebars = global.Handlebars = require("handlebars");
+const expect = require("chai").expect;
+
+require("../../../../../client/js/libs/handlebars/localetime");
```
astorije: This is a bit gross but I would like the test hierarchy to mirror the code hierarchy. We can very probably clean that up a bit later, when structure for client-side stuff gets a bit more stable. In the meantime, it's a test, and it passes...

PR #707, opened 2016-10-20
Get rid of OSX CI builds until they get much faster

astorije:
OSX builds have been nothing but a pain on Travis CI: they fail with no good reason, they stay pending forever, etc.

As far as I can tell, I can't remember one valid build they failed and we legitimately discovered a bug. Dev env on OSX is very close to Linux so it's good enough to have it here.

I'm OK with having AppVeyor running slow, which it often does, but Travis CI should be fast and so far this has been the unfortunate bottleneck.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ node_js:
 
 matrix:
   fast_finish: true
-  include:
-  - os: osx
-    node_js: 4
 
 cache:
   directories:


.travis.yml
```diff
@@ -3,12 +3,6 @@ node_js:
 - 6
 - 4
 
-matrix:
-  fast_finish: true
```
xPaw: Keep this.

astorije: `fast_finish` is only applied when there is a build with allowed failures set up. We don't need to keep it if we remove OSX builds (but unfortunately, a build that times out after an hour will not have failed yet, so `fast_finish` doesn't apply in these cases as well...)

astorije: Also, OSX builds are **always** longer than Linux builds, so it would alway fast finish...

xPaw: Not only in that case. https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

astorije: > Not only in that case. https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

Again, this **only** applies if there are jobs in the matrix with allowed failures. We actually have none (the OSX one is not even currently marked as allowed-failures, we would need to have an extra line for that).

xPaw: https://github.com/travis-ci/travis-ci/issues/2260#issuecomment-42144636

PR #708, opened 2016-10-20
Add a UI element to cycle through nick completions on mobile

astorije:
This solves #339.

This was heavily inspired by https://github.com/maxpoulin64/lounge/commit/a877e46.
Clearly not a definitive solution but a good easy start to have and to improve later on.

### Result

nick-cycler

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -141,6 +141,7 @@ button {
 #footer .icon,
 #chat .count:before,
 #settings #play:before,
+#form #cycle-nicks:before,
 #form #submit:before,
 #chat .invite .from:before,
 #chat .join .from:before,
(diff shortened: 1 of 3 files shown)


client/css/style.css
```diff
@@ -1731,6 +1740,11 @@ button {
 	#viewport.menu-open {
 		transition: none;
 	}
+
+	/* Nick cycler should only be displayed on mobile */
```
maxpoulin64: Actually, I don't really see any harm into having it available on desktop as well. Makes it easy to discover for people that are not used to the tab mechanic.

astorije:
@maxpoulin64, made some changes regarding your comment:

- Rebased with `master`
- Do not hide on mobile anymore
- Consequence of this:
  - Use a `click` event now
  - Added a tooltip
  - Re-focus after clicking the button

Mind confirming/updating your review?

williamboman:
Loses input focus on mobile. Perhaps bypass the mobile detection and just `input.focus()`?

astorije:
> Loses input focus on mobile. Perhaps bypass the mobile detection and just `input.focus()`?

Same issue than #483.
Actual fix for that issue should fix both at the same time.

PR #709, opened 2016-10-20
Remove browser notification polyfill and inform user when unsupported

astorije:
Our polyfill is not really a polyfill in the sense that it does not bring similar functionality: it mostly makes the page title blink. I don't even know if that works properly as I haven't seen it in action even once.
I think we should be more explicit than that: if browser supports it, good, use notifications, otherwise, let user know it's not supported and don't try alternative techniques.

According to [caniuse.com](http://caniuse.com/#feat=notifications), we should test this on:

- [ ] IE 11
- [ ] Edge 13
- [ ] iOS Safari 9
- [ ] iOS Safari 10
- [ ] Opera Mini (though I don't know how well we do there)
- [ ] Android

Help welcome with testing.
The testing goal is simply to make sure the app overall works fine apart from notifications in these environments, and that it displays the warning when not supported (though on Chrome for Android for example, it pretends it's supported, even makes the permission request, but no notifications show up...).

### Result

screen shot 2016-10-20 at 00 43 46


MaxLeiter:
Im running a public version of this PR temporarily for easy testing here: http://avatar.playat.ch:3000/

astorije:
Thanks a lot, @MaxLeiter!

xPaw:
Pretty sure there needs to be a check so that notifications can't be created if the api doesn't exist

astorije:
As far as I can tell, [this is the only place](https://github.com/thelounge/lounge/pull/709/files#diff-e5178f7b74fe45f2cfe1baf9aa1ef6faR979) where a `Notification` is created. It is not executed if `options.desktopNotifications` is `false`, which is its default value and never changes when `"Notification" in window` is `false` (because the checkbox is disabled).

xPaw:
I'm thinking there could be potential problem where the option was previously enabled (old version where there was no check, or polyfill removal)

PR #715, opened 2016-10-22
Store channel keys

spookhurb:
Fixes issue #412 

This stores channel keys in the user's config file. It can grab them from /join commands or if they are entered in the connect window (in the format #channel $key). It also sends the keys to server when reconnecting.

### DIFF
--- a/src/plugins/inputs/join.js
+++ b/src/plugins/inputs/join.js
@@ -1,24 +1,33 @@
 "use strict";
 
 var Chan = require("../../models/chan");
+var _ = require("lodash");
 
 exports.commands = ["join"];
 
 exports.input = function(network, chan, cmd, args) {
+	var client = this;
 	if (args.length !== 0) {
 		var irc = network.irc;
-		irc.join(args[0], args[1]);
+		if (args.length === 2) {
+			irc.join(args[0], args[1]);
+		} else {
+			irc.join(args[0]);
+		}
+		_.zip(args[0].split(/,/g), args[1].split(/,/g))
+			.map(function(chanPairs) {
+				var channel = new Chan({
+					type: Chan.Type.CHANNEL,
+					name: chanPairs[0],
+					key: chanPairs[1],
+				});
+				network.channels.push(channel);
+				client.emit("join", {
+					network: network.id,
+					chan: channel
+				});
+			});
 	}
-	var newChan = new Chan({
-		type: Chan.Type.CHANNEL,
-		name: args[0],
-		key: args[1],
-	});
-	network.channels.push(newChan);
-	this.emit("join", {
-		network: network.id,
-		chan: newChan
-	});
 	this.save();
 	return true;
 };
--- a/src/plugins/inputs/join.js
+++ b/src/plugins/inputs/join.js
@@ -1,0 +1,24 @@
+"use strict";
+
+var Chan = require("../../models/chan");
+
+exports.commands = ["join"];
+
+exports.input = function(network, chan, cmd, args) {
+	if (args.length !== 0) {
+		var irc = network.irc;
+		irc.join(args[0], args[1]);
+	}
+	var newChan = new Chan({
+		type: Chan.Type.CHANNEL,
+		name: args[0],
+		key: args[1],
+	});
+	network.channels.push(newChan);
+	this.emit("join", {
+		network: network.id,
+		chan: newChan
+	});
+	this.save();
+	return true;
+};
--- a/test/models/network.js
+++ b/test/models/network.js
@@ -13,8 +13,8 @@ describe("Network", function() {
 			var network = new Network({
 				name: "networkName",
 				channels: [
-					new Chan({name: "#thelounge"}),
-					new Chan({name: "&foobar"}),
+					new Chan({name: "#thelounge", key: ""}),
+					new Chan({name: "&foobar", key: ""}),
 					new Chan({name: "Channel List", type: Chan.Type.SPECIAL}),
 					new Chan({name: "PrivateChat", type: Chan.Type.QUERY}),
 				]
@@ -34,8 +34,8 @@ describe("Network", function() {
 				ip: null,
 				hostname: null,
 				channels: [
-					{name: "#thelounge"},
-					{name: "&foobar"},
+					{name: "#thelounge", key: ""},
+					{name: "&foobar", key: ""},
 				]
 			});
 		});
--- a/test/models/network.js
+++ b/test/models/network.js
@@ -15,6 +15,8 @@ describe("Network", function() {
 				channels: [
 					new Chan({name: "#thelounge", key: ""}),
 					new Chan({name: "&foobar", key: ""}),
+					new Chan({name: "#secret", key: "foo"}),
+					new Chan({name: "&secure", key: "bar"}),
 					new Chan({name: "Channel List", type: Chan.Type.SPECIAL}),
 					new Chan({name: "PrivateChat", type: Chan.Type.QUERY}),
 				]
@@ -36,6 +38,8 @@ describe("Network", function() {
 				channels: [
 					{name: "#thelounge", key: ""},
 					{name: "&foobar", key: ""},
+					{name: "#secret", key: "foo"},
+					{name: "&secure", key: "bar"}
 				]
 			});
 		});
--- a/src/client.js
+++ b/src/client.js
@@ -42,6 +42,7 @@ var inputs = [
 	"connect",
 	"disconnect",
 	"invite",
+	"join",
 	"kick",
 	"mode",
 	"nick",
@@ -156,7 +157,8 @@ Client.prototype.connect = function(args) {
 			}
 
 			channels.push(new Chan({
-				name: chan.name
+				name: chan.name,
+				key: chan.key
 			}));
 		});
 
@@ -167,11 +169,15 @@ Client.prototype.connect = function(args) {
 	// also used by the "connect" window
 	} else if (args.join) {
 		channels = args.join
-			.replace(/,/g, " ")
-			.split(/\s+/g)
+			.split(/\s*,\s*/g)
 			.map(function(chan) {
+				var splitchan = chan.split(/\s+/g, 2);
+				if (splitchan.length === 1) {
+					splitchan.push("");
+				}
 				return new Chan({
-					name: chan
+					name: splitchan[0],
+					key: splitchan[1]
 				});
 			});
 	}
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -19,6 +19,7 @@ function Chan(attr) {
 		id: id++,
 		messages: [],
 		name: "",
+		key: "",
 		topic: "",
 		type: Chan.Type.CHANNEL,
 		firstUnread: 0,
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -84,7 +84,8 @@ Network.prototype.export = function() {
 		})
 		.map(function(chan) {
 			return _.pick(chan, [
-				"name"
+				"name",
+				"key"
 			]);
 		});
 
--- a/src/plugins/irc-events/connection.js
+++ b/src/plugins/irc-events/connection.js
@@ -40,7 +40,7 @@ module.exports = function(irc, network) {
 			}
 
 			setTimeout(function() {
-				network.irc.join(chan.name);
+				network.irc.join(chan.name, chan.key);
 			}, delay);
 			delay += 1000;
 		});


test/models/network.js
```diff
@@ -13,8 +13,8 @@ describe("Network", function() {
 			var network = new Network({
 				name: "networkName",
 				channels: [
-					new Chan({name: "#thelounge"}),
-					new Chan({name: "&foobar"}),
+					new Chan({name: "#thelounge", key: ""}),
+					new Chan({name: "&foobar", key: ""}),
```
xPaw: Add a test with non empty key

src/plugins/inputs/join.js
```diff
@@ -0,0 +1,24 @@
...
+var Chan = require("../../models/chan");
+
+exports.commands = ["join"];
+
+exports.input = function(network, chan, cmd, args) {
+	if (args.length !== 0) {
+		var irc = network.irc;
+		irc.join(args[0], args[1]);
+	}
+	var newChan = new Chan({
```
xPaw: Join command can join multiple channels at once using commas.

spookhurb:
This should support /join-ing several channels at once

src/plugins/inputs/join.js
```diff
@@ -0,0 +1,33 @@
...
+var Chan = require("../../models/chan");
+var _ = require("lodash");
+
+exports.commands = ["join"];
+
+exports.input = function(network, chan, cmd, args) {
+	var client = this;
+	if (args.length !== 0) {
+		var irc = network.irc;
+		irc.join(args[0], args[1]);
```
xPaw: Don't send args[1] if there are no keys

src/plugins/inputs/join.js
```diff
@@ -0,0 +1,33 @@
...
+
+	_.zip(args[0].split(/,/g), args[1].split(/,/g))
+		.map(function(chanPairs) {
+			return new Chan({
+				type: Chan.Type.CHANNEL,
+				name: chanPairs[0],
+				key: chanPairs[1],
+			});
+		})
+		.map(function(channel) {
```
xPaw: Don't really need 2 maps.

src/plugins/inputs/join.js
```diff
@@ -0,0 +1,33 @@
...
+exports.commands = ["join"];
+
+exports.input = function(network, chan, cmd, args) {
+	var client = this;
+	if (args.length !== 0) {
+		var irc = network.irc;
+		irc.join(args[0], args[1]);
+	}
+
+	_.zip(args[0].split(/,/g), args[1].split(/,/g))
```
xPaw: You assume there's always keys.

src/plugins/inputs/join.js
astorije: Is it reasonable to just forward the command as-is to `irc-framework`? Effectively doing what we do on master right now, but with `inputs/join.js` having some logic?

I understand the need for `inputs/join.js` is just because `irc-events/join.js` does not have access to the key.

src/plugins/inputs/join.js
```diff
@@ -0,0 +1,33 @@
...
+		})
+		.map(function(channel) {
+			network.channels.push(channel);
+			client.emit("join", {
+				network: network.id,
+				chan: channel
+			});
+		});
+
+	this.save();
```
astorije: We already do that in `irc-events/join.js`.

astorije:
Basically the question to ask ourselves here is: do we want the `Channel` creation (in memory and in file) happening in the `inputs` (when emitting `/join` from the client) or in the `irc-events` (when the actual joining confirmation comes from the IRC server).

I think it makes more sense to keep this in the `irc-events`. Then logic in `inputs/join.js` is simply a way to retain the key for a given channel name until the `join` event comes back.
Does this make sense? What do you think, @xPaw?

src/plugins/inputs/join.js
spookhurb: You know, when I was testing, the channels keys weren't saved in the user JSON files when I didn't add this line. I could always try again.

xPaw:
@astorije in this case it's actually better to do it in both places. Command to get key, and instantly have the channel for client to know about, if network doesn't reply. Event in cases where server joins you into a channel that client didn't know about. The only thing this is missing is pushing this created channel to the front-end.

src/plugins/inputs/join.js
xPaw: I don't think we would save any code doing that.

src/plugins/inputs/join.js
astorije: That's normal, but not what I meant, sorry.

This is what your code does:

1. User joins a new channel (`client/js/lounge.js`)
2. `join` input handler is called (`inputs/join.js`)
  1. Saves the channel in the user config
  2. Emits a `join` to the IRC server
3. `join` IRC event handler is called (`irc-events/join.js`)
  1. Saves the channel in the user config
  2. Opens the window on the client

There is a clear duplication of logic in (2.i.) and (3.i.). The user config saving needs to happen in only one of these.
I am of the opinion it should happen in `irc-events/join.js` because that would be the most reliable way: it is only called when the joining is actually coming back from the server and is successful. If you do it in `inputs/join.js`, you possible need to check many things, like valid channel name, etc.
Problem is, in `irc-events/join.js`, you lost access to the key, so in my opinion, remembering keys should be the only logic in `inputs/join.js`, and forwarding the input command to `irc-framework`.

xPaw: See my comment below, while it is duplication, it makes sense in IRC world.

astorije:
@xPaw, I think we mostly agree. Only difference I can see is that I don't think `inputs/join.js` should have anything to do with the client nor the config file.
If network doesn't reply, I'm don't think we should create a channel on the front-end: after all, channel is not joined!
And I don't think we should save the config file twice either.

src/plugins/inputs/join.js
astorije: I was not thinking in terms of saving code but more for robustness. Here we need to parse things properly, send them back properly, while the command coming from the client should be a valid command, right?
If so, why not forwarding it as-is to `irc-framework` and deal with logic specific to us only in this handler?

src/plugins/inputs/join.js
astorije: I might be lacking something, but I don't see why we would need to save the config file or create the client-side channel twice. When server acknowledges the joining should be enough.
Unless there are server that don't acknowledge you joined a server properly, but that would be madness... wouldn't it?

spookhurb: Also to clarify: without this line, the CHANNEL isn't saved.

spookhurb: Yeah, I can see the code in `inputs/join.js` , but I'm kinda confused about why it isn't saving the channel name at least?

xPaw:
Astorije, that's true that you could practically create a channel window with name that can't actually be joined. But how else would you keep join keys in memory? Channel object is the cleanest solution here.

maxpoulin64:
Adding some more info to this: we could also get the key from the channel mode. That would also allow us to track key changes automatically. I think it makes sense to handle this on the irc-fw side and not the inputs.

spookhurb:
@astorije Well there is the issue of when the server goes down and/or the lounge needs to be restarted. As of right now when that happens you **do** end up with channels you are not connected to in your front end.

spookhurb:
@maxpoulin64 what would be a good approach from grabbing the channel key from the channel mode? is that being broadcast by the channel or...?

astorije:
> But how else would you keep join keys in memory? Channel object is the cleanest solution here.

Client could keep a list of channel name / key pairs. (option 1) That could easily be just an object of properties the channel name, values the key entered.
Only reason why this would need persistence outside of the client object is if you enter `/join #channel key` and iiiiimmediately close the client before the server responds (which usually responds in a second or so). But I  think keeping it in memory is fine, doesn't need to be saved to user config.

Alternatively (option 2), if you really want to create a `new Chan()` instead of a dummy channel-key map, there could be a new property `joined` that's false until the IRC server comes back with a response. `Chan`s without `joined === true` are not sent to the client and are not persisted to the user config, but they exist in memory.

In both cases, when server comes back with an ack, `irc-events/join` queries the channel-key map (option 1) / re-uses the `Chan() with `joined === false` (option 2) to get the key for that channel, and saves the user config with both the channel name and the key.

Does this make sense?

> @astorije Well there is the issue of when the server goes down and/or the lounge needs to be restarted.

Sorry if there is any miscommunication, but again I'm not saying we should **never** save the key to the user, I'm just saying the channel should not be saved at input time. Channel and its key should be saved only when we're certain we joined the channel (i.e. when server acknowledges it).

> Adding some more info to this: we could also get the key from the channel mode. That would also allow us to track key changes automatically.

It makes a lot of sense to keep track of key changes (maybe a later PR?), that's brilliant actually, but I'm not sure we can leverage this at join time: correct me if I'm wrong, but would that mean requiring to send a `/mode` command right after joining to get key value?
I get that we want to listen to modes for key changes/setups, but at joining time, I think using what key user *just* typed is probably the most efficient, no?

maxpoulin64:
@astorije Actually when you join a channel, the server automatically pushes the mode, the users list and the topic. I think if we just hook into the initial mode and the mode change events, and update the key from there, it would all work by itself.

astorije:
> Actually when you join a channel, the server automatically pushes the mode

Like, all the time? And does the server *always* (as in, not depending on the IRCd) brings back a copy of the key as well? If so then yes, definitely, no need for input catching at all!

That would be terrific if this is consistent!
Also, does `irc-framework` already send all these when joining? I know it sends user list, topic and modes, but does it send the key as well?

maxpoulin64:
Yes, it's part of the [standard](https://tools.ietf.org/html/rfc1459#section-4.2.3), just like +i and +s. The channel key is mode +k.

Not sure if irc-framework passes it, but probably.

DanielOaks:
Current channel modes being pushed to the client on `JOIN` isn't universal. It's only universal to send `RPL_TOPIC` (if the channel has a topic) and one or more `RPL_NAMREPLY` numerics listing everyone in the channel.

Most clients automatically send a `MODE #channelname` automatically after they receive a `JOIN` for themselves to get that information though.

Here's an example log from ircd-hybrid:
```
local <-  JOIN #testchan asdf
local  -> :girc!girc@0::1 JOIN #testchan * :*
local  -> :hybrid.dnt 353 girc = #testchan :girc!girc@0::1 dan_!d@0::1 @dan!d@localhost
local  -> :hybrid.dnt 366 girc #testchan :End of /NAMES list.
```

From the charybdis testnet:
```
local <-  JOIN #testestestest
local  -> :dan-test-mode-stuff!~girc@101.165.yxn.wlm JOIN #testestestest * :*
local  -> :jaguar.dereferenced.org MODE #testestestest +nt
local  -> :jaguar.dereferenced.org 353 dan-test-mode-stuff = #testestestest :@dan-test-mode-stuff!~girc@101.165.yxn.wlm
local  -> :jaguar.dereferenced.org 366 dan-test-mode-stuff #testestestest :End of /NAMES list.
```

From Freenode:
```
local <-  JOIN #ircdocs
local  -> :dan-test!~girc@101.165.168.135 JOIN #ircdocs * :*
local  -> :weber.freenode.net 332 dan-test #ircdocs :http://ircdocs.horse/ and associated projects
local  -> :weber.freenode.net 333 dan-test #ircdocs dan-!~d@blah 1475366441
local  -> :weber.freenode.net 353 dan-test = #ircdocs :nicknames nickname etc dan-test
local  -> :weber.freenode.net 366 dan-test #ircdocs :End of /NAMES list.
```

astorije:
I was kind of distant for the project for a few weeks. @xPaw, @spookhurb, where are we with this?
It's a great feature to have and I think we're mostly there :)

xPaw:
This PR needs to be remade to get channel key from mode.

spookhurb:
Finals are kinda wrecking me right now, I'll work this out over break.

astorije:
@spookhurb, any chance you will be able to resume this? :)

spookhurb:
yes! i will definitely be able to resume this, sorry.

astorije:
Awesome, marking this PR for v2.3.0 then (v2.2.0 is around the corner)!

astorije:
Hey @spookhurb, ping on this? :)

spookhurb:
@astorije still working on it! currently trying to figure out if there are any `MODE` commands that the server sends back that have parameters attached to them, because it looks like irc servers will send them off in one block like `+CPSnsk testpassword`, and it seems like it would be pretty easy to grab the wrong parameter as a chankey if some seldom used mode was in place, for example. It looks like `parseModeList` in `irc-framework` could be helpful, but currently the irc event code for mode isn't using any of it. I want to get this issue resolved while also keeping in mind the possibility there _might_ be other parameters you'd want to keep saved.

astorije:
@xPaw, do you think you could help @spookhurb on this so we can merge it soon? Thanks! :)

xPaw:
Superceded by #1003.

astorije:
Hey @spookhurb, thanks a lot for your work and efforts!
@xPaw gave a shot at this and went for a different approach, but we wouldn't have started it without your initial push. Sorry your PR didn't make it as is this time, but at least the feature will be there in the next release 😄

Issue #560, opened 2016-08-13
Show all WHOIS data


dgw:
Not all WHOIS replies are shown. Things like idle time, sign-on time, and nickserv identification status do not appear in either the opened query buffer or the server buffer. (as of 2.0.0-pre.5)

astorije:
[This `irc-framework` doc page](https://github.com/kiwiirc/irc-framework/blob/16728e3165d04afce5502ec6c8cd848638b2c4b7/docs/events.md) shows some information returned that [we do not use](https://github.com/thelounge/lounge/blob/f824036225aa2763ff3a25d1b330735f324d3e8a/client/views/actions/whois.tpl). It should not be difficult to add.
Real question is: what do we want to display and what do we not want to display? @maxpoulin64?

richrd:
What about having a toggle to show and hide the extra whois data?

YaManicKill:
I'm normally all for toggles, but that feels incredible unnecessary for
showing some data on the whois. Let's be honest, the whois is technical
info anyway that not a huge number of people use, so show it all.

It also brings up the other issue of where we show whois. I still think
that a modal is a much better place to show whois as it's very quick "I
want to see the current state of this information right now, and then I'm
done". Having it in a query just feels weird.

astorije:
> I'm normally all for toggles, but that feels incredible unnecessary for showing some data on the whois. Let's be honest, the whois is technical info anyway that not a huge number of people use, so show it all.

I agree with this, though if the whois paragraph gets out of hand, we might want to cherry-pick. 

> It also brings up the other issue of where we show whois. I still think that a modal is a much better place to show whois as it's very quick "I want to see the current state of this information right now, and then I'm done". Having it in a query just feels weird.

I'm not too sure about a modal, it has UX issues and as soon as we have our first modal, people will dream of modals everywhere. Then we'll hear that our app is gross because it has so many modals... :-)
A nice alternative, I think, would be to display that data in a right column similar to the user list on channels. But I think this will be for later, and definitely a different discussion. We don't have to delay the current issue because of this.

dgw:
[...]

PR #721, opened 2016-10-27
Add human-readable idle time in whois info

astorije:
One step towards #560.

See http://superuser.com/a/272069/208074 for explanation on `/whois nick nick` instead of `/whois nick`.

### Results

screen shot 2016-12-18 at 22 17 47

This PR used to say "since X days, X hours, X minutes and X seconds" but was moved to the format above for simplicity and consistency with other date/times in the current UI.
Helper and tests to do that have been moved to the archive PR #819.

### DIFF
--- a/client/views/actions/whois.tpl
+++ b/client/views/actions/whois.tpl
@@ -33,3 +33,9 @@
 	is away <i>({{whois.away}})</i>
 </div>
 {{/if}}
+{{#if whois.idle}}
+<div>
+	<span role="button" class="user {{colorClass whois.nick}}" data-name="{{whois.nick}}">{{whois.nick}}</span>
+	has been idle since {{localetime whois.idleTime}}.
+</div>
+{{/if}}
--- a/src/plugins/inputs/whois.js
+++ b/src/plugins/inputs/whois.js
@@ -1,0 +1,15 @@
+"use strict";
+
+exports.commands = ["whois"];
+
+exports.input = function(network, chan, cmd, args) {
+	if (args.length === 1) {
+		// This queries server of the other user and not of the current user, which
+		// does not know idle time.
+		// See http://superuser.com/a/272069/208074.
+		network.irc.raw("WHOIS", args[0], args[0]);
+	} else {
+		// Re-assembling the command parsed in client.js
+		network.irc.raw(`${cmd} ${args.join(" ")}`);
+	}
+};
--- a/src/plugins/irc-events/whois.js
+++ b/src/plugins/irc-events/whois.js
@@ -27,6 +27,9 @@ module.exports = function(irc, network) {
 				text: "No such nick: " + data.nick
 			});
 		} else {
+			// Absolute datetime in milliseconds since nick is idle
+			data.idleTime = Date.now() - data.idle * 1000;
+
 			msg = new Msg({
 				type: Msg.Type.WHOIS,
 				whois: data
--- a/src/client.js
+++ b/src/client.js
@@ -52,6 +52,7 @@ var inputs = [
 	"raw",
 	"topic",
 	"list",
+	"whois"
 ].reduce(function(plugins, name) {
 	var path = "./plugins/inputs/" + name;
 	var plugin = require(path);


MaxLeiter:
bit of a nitpick, but I think there should be an oxford comma after the minutes before the `and`.

astorije:
I thought about it, then decided to not do that. I can update if people have strong feelings about that.

client/views/actions/whois.tpl
```diff
@@ -33,3 +33,9 @@
 	is away <i>({{whois.away}})</i>
 </div>
 {{/if}}
+{{#if whois.idle}}
+<div>
+	<span role="button" class="user {{colorClass whois.nick}}" data-name="{{whois.nick}}">{{whois.nick}}</span>
+	has been idle for {{humanReadableRange whois.idle}}.
```
xPaw: Tooltip with just the date.

client/js/lounge.js
```diff
@@ -844,7 +844,7 @@ $(function() {
 
 		socket.emit("input", {
 			target: chat.data("id"),
-			text: "/whois " + name
+			text: "/whois " + name + " " + name
```
xPaw: @DanielOaks mind saying if there could be any problems with this?

DanielOaks: Hmm I'll take a look, do some testing and let you know

astorije: @DanielOaks, any luck? :)

DanielOaks: Ah, sorry. I had a squiz, it's not really defined anywhere but servers out there process it without an issue. Some like Insp also seem to require it to send all the WHOIS info. Should be fine, yeah.

PolarizedIons: @astorije a comment in the code here, explaining why the name is repeated, might be helpful in the future

astorije: @xPaw, @DanielOaks, if you look at the third commit of this PR, I actually moved that logic to the server through a new input file.
Reason behind that is that with the second commit alone, clicking on a nick and typing `/whois nick` were giving different outputs, and it confused me for a second. Now `whois nick` directly sends `/whois nick nick` (just like `/away` sends `/away <empty reason>`) and it's nicer for the user.
I'll squash commits 2 and 3 if you're ok with that (commit 3 cancels out commit 2, but I kept it for now just in case).

@PolarizedIons, done.

astorije:
@YaManicKill, @xPaw, I ended up moving away from the "since X days, X hours, ..." format I created and used the same date-time format we use for other dates in the UI. Reason is that when we start using relative dates for the date marker with moment, we can add it here as well and move the full date in a tooltip so it says "has been idle for about 2 hours" or the like.
It makes for a simpler PR and my helper is more likely to be buggy anyway.

For the oxford comma, I started to look at it then gave up as I would have had to make sure there is no comma with only 2 values. Painful. Better handled later.

src/plugins/inputs/whois.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+
+exports.commands = ["whois"];
+
+exports.input = function(network, chan, cmd, args) {
+	// This queries server of the other user and not of the current user, which
+	// does not know idle time.
+	// See http://superuser.com/a/272069/208074.
+	network.irc.raw("WHOIS", args[0], args[0]);
```
xPaw: Yeeah, this might have side effects. As the command is described as this: `WHOIS [<server>] <nickmask>[,<nickmask>[,...]]`

So if you tried to query nick on a server, your command would not work.

src/plugins/irc-events/whois.js
```diff
@@ -27,6 +27,9 @@ module.exports = function(irc, network) {
 				text: "No such nick: " + data.nick
 			});
 		} else {
+			// Absolute datetime in milliseconds since nick is idle
+			data.idleTime = Date.now() - data.idle * 1000;
```
xPaw: Not entirely sure if whois messages can have `server-time`, but if they do, use that instead maybe?

src/plugins/inputs/whois.js
astorije: Good call. How do you suggest I do that? Solve it on the server or just cancel and go back to client-side only?

If server-side, what about:

```js
if (args.length === 1) {
  network.irc.raw("WHOIS", args[0], args[0]);
} else {
  network.irc.raw("WHOIS", args.join(" "));
}
```

or something like that?
`/whois nick` would always be `/whois nick nick`, and everything else would be forwarded as is.

DanielOaks: `args.join(" ")` seems like it'd combine all args into the trailing, so would it end up sending out something like this?
```
WHOIS :arg1 arg2
```

That wouldn't go well. If that's how it'd send out with the above, you'd need to make it push all args separately instead (not sure how you'd do that in JS, but in other langs it'd look something like `network.irc.raw("WHOIS", args...);`

astorije: I guess that would be `network.irc.raw("WHOIS", ...args);` in JS with ES6's spread operator. Will wait for a LGTM or better suggestion from @xPaw.

src/plugins/irc-events/whois.js
xPaw: Hmm maybe that's just missing in the framework. cc @prawnsalad

src/plugins/inputs/whois.js
astorije: Scratch that, I re-assembled the whole command that gets parsed in `client.js` and it works just fine (spread operator is not available in Node.js v4...).

src/plugins/irc-events/whois.js
astorije: I suggest going with what we have right now and improving if/when irc-framework adds this (if possible). In the meantime, this works just fine (and also I am sooo over that PR :D).

astorije:
@xPaw, see my latest commit, I think that's the best compromise: WHOIS with a single argument gets repeated, everything else gets sent raw as-is.

PR #722, opened 2016-10-27
Add `npm run build` to README

MaxLeiter:


astorije:
This is unnecessary: [`npm run prepublish` is run along with `npm install` has run](prepublish: Run BEFORE the package is published. (Also run on local npm install without any arguments.)), and [the `prepublish` script is actually `npm run build`](https://github.com/thelounge/lounge/blob/e9b118ea713fd44b41f022bb9e553c80e23fbb0e/package.json#L28).

👎

YaManicKill:
> This is unnecessary: npm run prepublish is run along with npm install has run, and the prepublish script is actually npm run build.

Just as a quick note, this will be removed in npm 5 (npm 4 was just released and deprecated it) and is replaced by "npm prepare" which is exactly the same and "npm prepublishOnly" which only run before publish, not install. I think npm 6 is the one that will replace prepublish with prepublishOnly.

Basically, currently this is the case, but we will want to look at moving to npm prepare when npm 4 has a good level of support. Or at least we can add both of them just now?

PR #725, opened 2016-10-29
Warn against running from source as root in README

astorije:
This was discussed in IRC yesterday.

Basically, npm pre/post scripts are not executed when running as root. There are ways to do so, but this goes beyond what the README should say.
Also, we should not be held responsible for someone messing up their system while running code that is deemed not prod-ready.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,7 @@ lounge --help
 
 ### Running from source
 
-The following commands install the development version of The Lounge. A word of
-caution: while it is the most recent codebase, this is not production-ready!
+The following commands install the development version of The Lounge:
 
 ```sh
 git clone https://github.com/thelounge/lounge.git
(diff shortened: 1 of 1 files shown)


xPaw:
Will this show up on npm?

astorije:
@xPaw, if you're talking about the package page, then yes, [just like it is today](https://www.npmjs.com/package/thelounge#running-from-source).

README.md
```diff
@@ -71,6 +70,11 @@ npm install
 npm start
 ```
 
+A word of caution:
+
+- While it is the most recent codebase, this is not production-ready!
+- This will fail if you run it as root, which is not recommended anyway.
```
xPaw: Instead of saying fail, say you will need to run build yourself.

YaManicKill: Agreed, but I would also change it to this:

```
While it is the most recent codebase, this is not production-ready!
It is not recommended to run this as root. However, if you decide to do so, you will have to run `npm run build`.
```

The "anyway" makes it sound a bit too casual.

astorije:
@xPaw, @YaManicKill, I updated the PR based on your comments. Good to go?

PR #729, opened 2016-11-02
Fix localetime Handlebars helper

chanand:
Setting the locales to en-US so the test will pass also on other locales.

### DIFF
--- a/client/js/libs/handlebars/localetime.js
+++ b/client/js/libs/handlebars/localetime.js
@@ -1,5 +1,5 @@
 "use strict";
 
 Handlebars.registerHelper("localetime", function(time) {
-	return new Date(time).toLocaleString();
+	return new Date(time).toLocaleString("en-US");
 });


astorije:
Hey @chanand, thanks a lot for the effort, but sadly this is not a good idea :/

This is not fixing the test, it's forcing the code to behave against users' browsers. So this is adjusting the code to comply with the test, while it should be the other way around.

It's better to fix the test by either 1) making it run under a fixed locale (which I haven't found a solution for yet), or 2) build the expected string so it has the format of the locale in test environment.

Would you like to give it a shot?

chanand:
@astorije, I suppose you are really right, but the 2 options you mentioned sound like finding a way for testing `Date.prototype.toLocaleString` instead of testing the `localetime` helper, is that correct?

williamboman:
Could also add a second parameter `locale` to the helper function, which defaults to `en-US`?

xPaw:
I think the normal fix would be to force locale when running tests.

astorije:
> I think the normal fix would be to force locale when running tests.

I agree, we should not adapt the code to satisfy the tests but the other way around :)

@chanand, are you willing to look into this further or should we close this PR?

astorije:
Closing this PR as invalid for now, @chanand feel free to re-open if you want to implement changes.

PR #732, opened 2016-11-07
Silence failures to trigger notifications when not available

### LINKED ISSUE #731: Notifications on mobile not working
On my phone (Android 6.0.1) using Chrome 54.0.2840.68, on master(837f78f) receiving a ping from a channel does not show up on the sidebar as a highlighted number, nor does it make the number increase. Console error is this: 

Other channels seem unaffected. Turning off desktop notifications in the settings fixes this problem, but the `DOMException: play()` is still there, but as a warning, not a error.

astorije:
Fixes #731.

Recent Chrome versions are dropping out `new Notification` in favor of `ServiceWorkerRegistration.showNotification`.
This makes sure nothing bad happens until we have proper support for Service Workers.

See:
- https://stackoverflow.com/questions/29774836/failed-to-construct-notification-illegal-constructor
- https://stackoverflow.com/questions/31512504/html5-notification-not-working-in-mobile-chrome

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1011,7 +1011,11 @@ $(function() {
 		if (msg.highlight || (options.notifyAllMessages && msg.type === "message")) {
 			if (!document.hasFocus() || !$(target).hasClass("active")) {
 				if (options.notification) {
-					pop.play();
+					try {
+						pop.play();
+					} catch (exception) {
+						// On mobile, sounds can not be played without user interaction.
+					}
 				}
 				toggleNotificationMarkers(true);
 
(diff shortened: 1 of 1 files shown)


xPaw:
I would add the same thing in settings, so if notifications can't be created, the option is disabled. Also the sounds can't be played on mobile without user interaction which also throws.

astorije:
> Also the sounds can't be played on mobile without user interaction which also throws.

Added a `try`/`catch` for this too.

> I would add the same thing in settings, so if notifications can't be created, the option is disabled.

This can only be detected when calling `Notification.new` so it can't be done in the settings. I'd need to fiddle with the `options`, `localStorage`, checkboxes and others to make something nice. For this I would also need to move around things to make sure they are declared in order and in the proper scope.

I'm not comfortable doing so as I'm pretty sure I'll break something at first try, and I don't have much time to give at the moment so I won't be able to do it myself.
I'd say let's fix the bugs reported in #731 for now and UX of that state can be improved later on. What do you think?

PR #736, opened 2016-11-15
src/server: log.debug config path on start-up

williamboman:
Feels like "where/which config file does my lounge instance use?" questions are quite common, this hopefully makes things clearer.

lounge__node_

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -76,9 +76,10 @@ module.exports = function() {
 
 	manager.sockets = sockets;
 
-	let protocol = config.https.enable ? "https" : "http";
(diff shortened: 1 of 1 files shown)


xPaw:
I don't see a reason why this can't be info level. Also probably better after version line.

williamboman:
> I don't see a reason why this can't be info level. Also probably better after version line.

I was a bit conflicted about which one to use. I'll switch to info and put it after the version line. Wording was made in a hurry and is a bit sloppy imo, suggestions are welcomed

PR #739, opened 2016-11-16
client: don't dismiss native web notifications programmatically after 5s

williamboman:
ref: #594

This should be good to go now since #709 landed, no?

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1041,14 +1041,11 @@ $(function() {
 							icon: "img/logo-64.png",
 							tag: target
 						});
-						notify.onclick = function() {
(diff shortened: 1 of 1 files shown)


xPaw:
I think this should be configurable, persistent vs timer. I really would hate notifications never going away until I close them.

YaManicKill:
> I really would hate notifications never going away until I close them.

They do go away, the browser deals with that. But some browsers have like a cache of notifications after they go off the screen, which this means they will go away with.

The notification will still go off screen after a few seconds, but that's up to the browser now.

astorije:
Happy this one landed, thanks guys!

PR #743, opened 2016-11-18
Print node version and platform

### LINKED ISSUE #742: Print node version in console to ease debugging

xPaw:
Fixes #742


astorije:
For the record, a before/after of this:

Before | After
--- | ---
screen shot 2016-11-19 at 23 01 00 | screen shot 2016-11-19 at 23 02 06

This is so fancy! 💅

Issue #374, opened 2016-06-01
HEX (hexadecimal) ident


Arkadietz:
Hello,

In the moment you can set only one ident. And for example if someone ban this ident multiple uses will be impact and that's not good for me. This future is already done in webirc clients like qwebirc and lightirc. 

Example:

[19:28:58] \* BGChat70516 (5ec578c8@BGChat.html.chat) has joined #Bulgaria

Where the ident : 5ec578c8 = IP ( 94.197.120.200 )

dgw:
WEBIRC support has already been merged: #240

maxpoulin64:
This is partly being addressed in #256. The hex indent is non-standard and so far I think only Freenode uses that. If the user connects over IPv6 the number won't fit anymore. For that purpose webirc is a better solution.

MaxLeiter:
Can be closed as of #749

astorije:
Can someone open an issue with details in https://github.com/thelounge/thelounge.github.io ? The moment that I'll go into heavily improving our docs is getting closer and closer, and that would help making sure there is room for this.

PR #749, opened 2016-11-19
Add support for hexip ilines and fix storing client ip in config

xPaw:
cc @MaxLeiter. This is mostly for demo purposes.

- Fixes saving ip and hostmask in user config
- Force user config save when connecting to a network
- Update old existing networks to have ip/hostmask on socket attachment

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -110,6 +110,17 @@ module.exports = {
 	lockNetwork: false,
 
 	//
+	// Hex IP
+	//
+	// If enabled, clients' username will be set to their IP encoded has hex.
+	// This is done to share the real user IP address with the server for host masking purposes.
+	//
+	// @type     boolean
+	// @default  false
+	//
+	useHexIp: false,
+
+	//
 	// WEBIRC support
 	//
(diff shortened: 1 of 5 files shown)


maxpoulin64:
Hmm, the user's IP actually is stored on disk, or at least it is when webirc is enabled. Maybe we could recycle that for the hex ident?

Maybe we could even make the config option support passing in a function so people can customize it to whatever non-standard their servers need?

xPaw:
@maxpoulin64 Yeah I saw the webirc code for that, it was a little suspect, but probably can just always store ip/hostmask in the config.

As for the function, do you know servers that support other methods? Might not be worth the investment.

xPaw:
This PR should hide the username field on the client if the option is enabled.

stepie22:
I believe this would address issue #374 ?

astorije:
> This is mostly for demo purposes.

Just curious, is this something we want to include with the codebase we release then?

If yes, how ready is this?

xPaw:
Should be ready.

PR #750, opened 2016-11-19
Avoid unnecessary disk writes when saving user

### LINKED ISSUE #299: Reduce client save calls
It saves the user config whenever you connect to the network, change nick or join a channel. In these cases it should only call save if the thing it's trying to save isn't already there (e.g. join a channel, but its already in `join` in users config).

It currently uses a 1 second timer (a global time I might add) to throttle it.

xPaw:
Closes #299.

- Avoid unnecessary disk writes when saving user
- Makes writing user updates to disk asynchronous
- Replace ghetto timer with lodash's debounce

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -292,19 +292,15 @@ Client.prototype.setPassword = function(hash, callback) {
 		client.manager.updateUser(client.name, {
 			token: token,
 			password: hash
-		});
+		}, function(err) {
+			if (err) {
+				log.error("Failed to update password of", client.name, err);
+				return callback(false);
+			}
 
-		// re-read the hash off disk to ensure we use whatever is saved. this will
-		// prevent situations where the password failed to save properly and so
-		// a restart of the server would forget the change and use the old
-		// password again.
-		var user = client.manager.readUserConfig(client.name);
-		if (user.password === hash) {
 			client.config.password = hash;
-			callback(true);
-		} else {
-			callback(false);
-		}
+			return callback(true);
+		});
 	});
 };
 
--- a/src/client.js
+++ b/src/client.js
@@ -460,23 +460,13 @@ Client.prototype.clientDetach = function(socketId) {
 	delete this.attachedClients[socketId];
 };
 
-var timer;
-Client.prototype.save = function(force) {
-	var client = this;
-
+Client.prototype.save = _.debounce(function SaveClient() {
 	if (Helper.config.public) {
 		return;
 	}
 
-	if (!force) {
-		clearTimeout(timer);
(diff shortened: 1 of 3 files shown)


src/client.js
```diff
@@ -292,19 +292,23 @@ Client.prototype.setPassword = function(hash, callback) {
...
-			client.config.password = hash;
-			callback(true);
-		} else {
-			callback(false);
-		}
+			// re-read the hash off disk to ensure we use whatever is saved. this will
+			// prevent situations where the password failed to save properly and so
+			// a restart of the server would forget the change and use the old
+			// password again.
+			var user = client.manager.readUserConfig(client.name);
```
xPaw: This is probably no longer necessary as the callback provides `err` from the write operation.

src/clientManager.js
```diff
@@ -94,28 +94,34 @@ ClientManager.prototype.addUser = function(name, password) {
...
-		_.assign(user, opts);
-		fs.writeFileSync(
-			Helper.getUserConfigPath(name),
-			JSON.stringify(user, null, "\t")
-		);
-	} catch (e) {
-		log.error("Failed to update user", e);
-		return;
+	let user = this.readUserConfig(name);
+	const currentUser = JSON.stringify(user, null, "\t");
```
xPaw: I timed this, and it's fast enough to do this (writes to disk are much slower, obviously). There's no really other easy way of comparing objects.

YaManicKill: Yeah, this isn't ideal, and could cause some false positives (if the user's properties are now in a different order, but exactly the same) but I doubt that's going to cause any problems.

And it's definitely the fastest way of comparing them, it's just not the most accurate.

src/client.js
YaManicKill: Yeah, this can be removed.

src/clientManager.js
xPaw: Actually order might cause issues due to sorting. Needs to be tested.

xPaw:
Sounds like shouldn't be a problem.

src/client.js
```diff
@@ -292,19 +292,23 @@ Client.prototype.setPassword = function(hash, callback) {
...
+			if (user.password === hash) {
+				client.config.password = hash;
+				callback(true);
+			} else {
+				callback(false);
```
astorije: That's quite the callback hell piling up here. Can we switch to using Promises for this? I believe that would make the code easier to follow :)

xPaw: A little out of scope of this PR.

xPaw:
@YaManicKill, @astorije both comments addressed.

PR #751, opened 2016-11-19
Change ghetto autoload to use fs.watch

xPaw:
Keeps all the logic, but instead of using a one second repeating timer uses `fs.watch`.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -140,9 +140,10 @@ ClientManager.prototype.removeUser = function(name) {
 	return true;
 };
 
-ClientManager.prototype.autoload = function(/* sockets */) {
+ClientManager.prototype.autoload = function() {
 	var self = this;
-	setInterval(function() {
+
+	fs.watch(Helper.USERS_PATH, _.debounce(() => {
 		var loaded = self.clients.map(c => c.name);
 		var added = _.difference(self.getUsers(), loaded);
 		added.forEach(name => self.loadUser(name));
@@ -160,5 +161,5 @@ ClientManager.prototype.autoload = function(/* sockets */) {
 				log.info("User '" + name + "' disconnected");
 			}
 		});
-	}, 1000);
+	}, 1000, {maxWait: 10000}));
 };


src/clientManager.js
```diff
@@ -160,5 +161,5 @@ ClientManager.prototype.autoload = function(/* sockets */) {
 				log.info("User '" + name + "' disconnected");
 			}
 		});
-	}, 1000);
+	}, 1000, {maxWait: 10000}));
```
YaManicKill: What does maxWait do? Does it stop watching after that time, or does it reload after that time whether it has changed or not?

xPaw:
If the callback keeps happening, it will wait up to max of 10 seconds and
execute

src/clientManager.js
```diff
@@ -140,9 +140,10 @@ ClientManager.prototype.removeUser = function(name) {
 	return true;
 };
 
-ClientManager.prototype.autoload = function(/* sockets */) {
+ClientManager.prototype.autoload = function ClientAutoLoad() {
```
astorije: I'm not sure we need this but if we do, common practice is to give it the same name than the callable name, so `function autoload() ...`.

src/clientManager.js
```diff
@@ -140,9 +140,10 @@ ClientManager.prototype.removeUser = function(name) {
...
 	var self = this;
-	setInterval(function() {
+
+	fs.watch(Helper.USERS_PATH, _.debounce(function ClientAutoLoadWatch() {
```
astorije: This is just a callback, use the more concise arrow syntax instead `() => {`.
(Plus I'm really not a fan at all of starting a trend where function names are capitalized. This is good on class and class-like names but not callbacks.)

src/clientManager.js
xPaw: I wanted to include the class name in the name. But in this case I messed up a little.

astorije:
> 2 nitpicks. I can make the changes myself if you're not nearby a computer, @xPaw. (Sorry, not on IRC right now, server is borked)

Ping, @xPaw?

xPaw:
I don't have access to my PC to deal with these. If you can yourself, by
all means.

astorije:
Will do, I wanted your approval before touching your PR :) Thanks!

PR #752, opened 2016-11-19
Use node 7 on travis

xPaw:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- 6
+- node
 - 4
```
astorije: Since v4 and v6 are actually both LTS right now, I'd prefer to have `4` + `6` + `node`.
And actually, I'd prefer to have `7` instead of `node`: it makes it clearer from the job list without having to look for the version in the console output (see image below), and it won't silently go to another major version when someone opens a PR right after it gets released. I'm fine having to bump the version in the file manually once in a while.

screen shot 2016-11-19 at 22 02 41

xPaw:
Let's just keep v6 for now (and wait for v8 to release), I don't want travis to keep building too many versions (push + pr + 3 versions = 6 builds).

astorije:
What's the big deal? I constantly see projects with a dozen jobs of 10+ minutes. Ours are 1min30s top and they'll run in parallel (6 parallel jobs per build) so that should not be an issue.

Since LTS versions are 4 and 6 and stable version is 7, I still believe we need to have those 3 versions tested against.

PR #753, opened 2016-11-20
Correctly remove closed sockets from oident file, remove unused functions

xPaw:


src/oidentd.js
```diff
@@ -12,57 +13,31 @@ function OidentdFile(file) {
...
 	},
 
-	refresh: function() {
-		var file = "# Warning: file generated by The Lounge: changes will be overwritten!\n";
+	refresh: function OidentdFileRefresh() {
+		let file = "# Warning: file generated by The Lounge: changes will be overwritten!\n";
 
-		function makeRule(connection) {
-			return "to " + connection.socket.remoteAddress
+		_.forEach(this.connections, (connection) => {
```
astorije: Reminder as discussed on IRC:

Use [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) for the `connections` object and iterate with native [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Iterating_Maps_with_forEach()) (then remove `const _ = require("lodash");` above).

ronilaukkarinen:
Checked this, just reporting it doesn't fix [a bug with ident being the same](https://github.com/thelounge/lounge/issues/754). Fixes undefineds in .oidentd.conf though.

.eslintrc.yml
```diff
@@ -6,6 +6,7 @@ env:
   browser: true
   mocha: true
   node: true
+  es6: true
 
 parserOptions:
   ecmaVersion: 6
```
astorije: `es6: true` makes `ecmaVersion: 6` unnecessary (see https://github.com/eslint/eslint/pull/7067). Do you mind removing `parserOptions`?

src/oidentd.js
```diff
@@ -1,68 +1,42 @@
...
+	this.connections = new Map();
 
 	this.refresh();
 }
 
 OidentdFile.prototype = {
-	hookSocket: function(socket, user) {
-		var that = this;
-		var id = null;
+	addSocket: function OidentdFileAddSocket(socket, user) {
```
astorije: If you make `OidentdFile` a class, I believe you could write these something like:

```es6
class OidentdFile {
  /* ... */

  addSocket(socket, user) {
    /* ... */
  }

  /* ... */
}
```

I definitely don't like naming a method differently than its descriptor (`addSocket` vs. `OidentdFileAddSocket`) and I even dislike more naming non-class-like functions with an uppercase, so I think that would be a win-win :)

xPaw: We definitely should start using classes, yeah. But for this PR I'll just remove the name additions.

.eslintrc.yml
xPaw: Done.

PR #755, opened 2016-11-22
Unblock AppVeyor build

IlyaFinkelshteyn:


appveyor.yml
```diff
@@ -12,6 +12,7 @@ environment:
 
 install:
   - ps: Install-Product node $env:nodejs_version
+  - npm install --save fs-extra
```
YaManicKill: So, this is adding fs-extra to the package.json

What exactly is this needed for? And if it's needed, it should probably be added to the package.json rather than installed manually for this, and if it's installed manually for this ... then it shouldn't really use --save.

xPaw: It already is there. I think thr proper fix would be `npm cache clean`

IlyaFinkelshteyn: As I understand now, problem was that older version stuck in cache. I proposed to return the cache but invalidate it if package.json have changed.

appveyor.yml
```diff
@@ -11,8 +11,8 @@ environment:
   nodejs_version: '4'
 
 install:
-  - ps: Install-Product node $env:nodejs_version
-  - npm install
+  - ps: Install-Product node $env:nodejs_version
```
xPaw: Why the spaces?

YaManicKill:
@IlyaFinkelshteyn Could you remove the extra space and squash/rebase this branch? Cheers.

astorije:
Squashed that for you :)

For what it's worth, I was fine living in AppVeyor world without a cache. Our builds literally take a minute or 2, it would not be the end of the world if we didn't have cache!

But that will do until next breakage I guess. Thanks a lot for your help, @IlyaFinkelshteyn! It's quite the support quality you have if you're ready to open PRs for your open source users 😅

stepie22:
@astorije don't quote me on this, but I believe this fixes an issue with appveyor randomly not being able to install npm packages, and failing builds

IlyaFinkelshteyn:
Great, sometimes retry is the only way to go :)

Issue #714, opened 2016-10-22
Channel sorting not synchronized in real time between clients


xPaw:
Only works after page reload.

MaxLeiter:
Fixed by #714

PR #757, opened 2016-11-22
Sync reordering of channels/networks to other clients

PolarizedIons:
Adresses issue #714. It attempts to move a channel (or network) into it's new position

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -36,6 +36,8 @@ $(function() {
 	var sidebar = $("#sidebar, #footer");
 	var chat = $("#chat");
 
+	var ignoreSortSync = false;
+
 	var pop;
 	try {
 		pop = new Audio();
@@ -1370,6 +1372,8 @@ $(function() {
 						order: order
 					}
 				);
+
+				ignoreSortSync = true;
 			}
 		});
 		sidebar.find(".network").sortable({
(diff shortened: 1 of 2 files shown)


client/js/lounge.js
```diff
@@ -1310,10 +1314,58 @@ $(function() {
...
+
+				var position = $(network.children()[key]); // Target channel at position
+
+				if (position.data("id") === value) { // Channel in correct place
+					return true; // No point in continuing
+				}
+
+				var channel = network.find(".chan[data-id=" + value + "]"); // Channel at position
+
+				$(channel).insertBefore(position);
```
xPaw: Update data-id too?

stepie22: Rearranging the channels currently does not update the id's, not even after refreshing the page so I didn't think of doing it. It seems like a server side thing, should I add it?

xPaw: Ah whoops, it should be server yes. I got confused by the comparison.

src/client.js
```diff
@@ -391,6 +391,8 @@ Client.prototype.open = function(socketId, data) {
 };
 
 Client.prototype.sort = function(data) {
+	self.emit("sync_sort", data); // Sync order to connected clients
```
xPaw: This should be done along with save call at the bottom. Also I noticed it will still save even if you sort a non existing channel.

src/client.js
```diff
@@ -396,16 +396,21 @@ Client.prototype.sort = function(data) {
 	var type = data.type;
 	var order = data.order || [];
 	var sorted = [];
+	var syncorder = []; // Channel id's, `sorted` contains channel objects
```
xPaw: You can map the sorted array to get ids.

src/client.js
```diff
@@ -396,16 +396,21 @@ Client.prototype.sort = function(data) {
...
 			var find = _.find(self.networks, {id: i});
 			if (find) {
 				sorted.push(find);
+				syncorder.push(i);
 			}
 		});
 		self.networks = sorted;
+
+		// Sync order to connected clients
+		self.emit("sync_sort", {order: syncorder, type: type});
```
xPaw: Just do it after saving like before.

xPaw: You can check if sorted array is empty before doing so though.

src/client.js
```diff
@@ -425,6 +425,10 @@ Client.prototype.sort = function(data) {
 	}
 
 	self.save();
+
+	// Sync order to connected clients
+	var syncorder = _.map(sorted, "id");
```
stepie22: Ah, my bad. Will do

stepie22: @astorije done

PR #759, opened 2016-11-23
Initial part/join condensing

YaManicKill:
This probably isn't done, and needs a lot more testing from other people to see if I've done anything bad.

Another good thing to add would be how IRC cloud shows it (it makes a little sentence explaining some of the parts/joins in it).

But this is just an initial one.

List of stuff still to do before merging:

- [x] Initial condenser element shows as behind the date line when it's the first in the channel
- [x] Condense based on date
- [x] Don't have all messages clickable, just the condenser
- [x] Change icon
- [x] Only condense if more than 1 item
- [x] Correct counts
- [x] Clicking "more" returns some of the same messages
- [x] Issues with flex layout
- [x] Hide when parts/joins are hidden inside of it
- [x] Change show/hide settings to "show/hide" and "codense"
- [x] Split out the "more" change to another PR
- [x] Fix name on option
- [x] Fix previews properly
- [x] Properly hide condense messages
- [x] Rename `handledTypes` to `actionTypes`
- ~Try removing the second [click handler](https://github.com/thelounge/lounge/pull/759#discussion_r126403670)~ UI logic was changed a bit, making this a non-issue

7c99048cb6f41986f0d7abe955a10193

a4192df26c41bb6d1f4e98c58915e835

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -197,6 +197,7 @@ kbd {
 #settings #play:before,
 #form #submit:before,
 #chat .invite .from:before,
+#chat .condensed .expander:before,
 #chat .join .from:before,
 #chat .kick .from:before,
 #chat .part .from:before,
@@ -282,6 +283,22 @@ kbd {
 #chat .nick .from:before {
 	content: "\f007"; /* http://fontawesome.io/icon/user/ */
 	color: #2ecc40;
+}
+
+#chat .condensed {
+	cursor: pointer;
+}
+
+#chat .condensed div {
+	cursor: default;
+}
+
+#chat .condensed .expander:before {
+	content: "\f0d7"; /* http://fontawesome.io/icon/caret-down/ */
+}
+
+#chat .condensed.closed .expander:before {
+	content: "\f0da"; /* http://fontawesome.io/icon/caret-right/ */
 }
 
 #chat .join .from:before {
@@ -801,6 +818,18 @@ kbd {
 	display: block;
 }
 
+#chat .condensed {
+	flex-wrap: wrap;
+}
+
+#chat .condensed .msg {
+	flex-basis: 100%;
+}
+
+#chat .condensed.closed .msg {
+	display: none;
+}
+
 #windows .header .topic,
 .messages .msg,
 .sidebar {
@@ -1092,15 +1121,17 @@ kbd {
 	color: #555;
 }
 
-#chat.hide-join .join,
-#chat.hide-mode .mode,
-#chat.hide-motd .motd,
-#chat.hide-nick .nick,
-#chat.hide-part .part,
-#chat.hide-quit .quit {
+#chat.hide-optional .join,
+#chat.hide-optional .mode,
+#chat.hide-optional .nick,
+#chat.hide-optional .part,
+#chat.hide-optional .quit,
+#chat.hide-optional .condensed,
+#chat.hide-motd .motd {
 	display: none !important;
 }
 
+#chat .condensed .content,
 #chat .join .content,
 #chat .kick .content,
 #chat .mode .content,
@@ -2129,4 +2160,8 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 #chat .action .text,
 #chat table.channel-list .topic {
 	white-space: pre-wrap;
+}
+
+.hide-text {
+	color: transparent !important;
 }
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -198,7 +198,6 @@ kbd {
 #settings #play:before,
 #form #submit:before,
 #chat .invite .from:before,
-#chat .condensed .expander:before,
 #chat .join .from:before,
 #chat .kick .from:before,
 #chat .part .from:before,
@@ -284,22 +283,6 @@ kbd {
 #chat .nick .from:before {
 	content: "\f007"; /* http://fontawesome.io/icon/user/ */
 	color: #2ecc40;
-}
-
-#chat .condensed {
-	cursor: pointer;
-}
-
-#chat .condensed div {
-	cursor: default;
-}
-
-#chat .condensed .expander:before {
-	content: "\f0d7"; /* http://fontawesome.io/icon/caret-down/ */
-}
-
-#chat .condensed.closed .expander:before {
-	content: "\f0da"; /* http://fontawesome.io/icon/caret-right/ */
 }
 
 #chat .join .from:before {
@@ -827,8 +810,21 @@ kbd {
 	flex-wrap: wrap;
 }
 
-#chat .condensed .msg {
-	flex-basis: 100%;
+#chat .condensed .content {
+	flex: 1;
+}
+
+#chat .condensed-text {
+	cursor: pointer;
+	transition: opacity .2s;
+}
+
+#chat .condensed-text:hover {
+	opacity: .6;
+}
+
+#chat .condensed-text .toggle-button:hover {
+	opacity: 1;
 }
 
 #chat .condensed.closed .msg {
@@ -1174,17 +1170,12 @@ kbd {
 
 #chat .toggle-button {
 	display: inline-block;
-	color: #666;
-	transition: color .2s, transform .2s;
+	transition: opacity .2s, transform .2s;
 }
 
-#chat .toggle-button.opened {
+#chat .toggle-button.opened, /* Thumbnail toggle */
+#chat .msg.condensed:not(.closed) .toggle-button { /* Expanded status message toggle */
 	transform: rotate(90deg);
-}
-
-#chat .toggle-button:hover {
-	/* transform and opacity together glitch, so need to use RGBA transition */
-	color: rgba(102, 102, 102, .8); /* #666 x .8 opacity */
 }
 
 #chat .toggle-content {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -194,6 +194,7 @@ kbd {
 #chat .title:before,
 #footer .icon,
 #chat .count:before,
+#settings .extra-help,
 #settings #play:before,
 #form #submit:before,
 #chat .invite .from:before,
@@ -332,6 +333,10 @@ kbd {
 	right: 18px;
 	font-size: 14px;
 	line-height: 50px;
+}
+
+#settings .extra-help:before {
+	content: "\f059"; /* http://fontawesome.io/icon/question-circle/ */
 }
 
 #settings #play:before {
@@ -1400,8 +1405,13 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	margin: 4px 10px 0 0;
 }
 
+#settings .extra-help,
 #settings #play {
 	color: #7f8c8d;
+}
+
+#settings .extra-help {
+	cursor: help;
 }
 
 #settings #play {
@@ -1723,6 +1733,16 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	animation-timing-function: ease-in;
 	-webkit-animation-delay: .4s;
 	animation-delay: .4s;
+}
+
+.tooltipped-no-delay:hover:before,
+.tooltipped-no-delay:hover:after,
+.tooltipped-no-delay:active:before,
+.tooltipped-no-delay:active:after,
+.tooltipped-no-delay:focus:before,
+.tooltipped-no-delay:focus:after {
+	-webkit-animation-delay: 0s;
+	animation-delay: 0s;
 }
 
 .tooltipped-s:after,
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1126,12 +1126,12 @@ kbd {
 	color: #555;
 }
 
-#chat.hide-optional .join,
-#chat.hide-optional .mode,
-#chat.hide-optional .nick,
-#chat.hide-optional .part,
-#chat.hide-optional .quit,
-#chat.hide-optional .condensed,
+#chat.hide-status-messages .join,
+#chat.hide-status-messages .mode,
+#chat.hide-status-messages .nick,
+#chat.hide-status-messages .part,
+#chat.hide-status-messages .quit,
+#chat.hide-status-messages .condensed,
 #chat.hide-motd .motd {
 	display: none !important;
 }
@@ -1397,12 +1397,11 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 #settings .opt {
 	display: block;
-	padding: 5px 0 10px 1px;
+	padding: 5px 0 5px 1px;
 }
 
 #settings .opt input {
-	float: left;
-	margin: 4px 10px 0 0;
+	margin-right: 6px;
 }
 
 #settings .extra-help,
@@ -1412,6 +1411,10 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 #settings .extra-help {
 	cursor: help;
+}
+
+#settings h2 .extra-help {
+	font-size: .8em;
 }
 
 #settings #play {
--- a/client/index.html
+++ b/client/index.html
@@ -218,38 +218,20 @@
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
-									<input type="checkbox" name="join">
-									Show joins
+									<input type="checkbox" name="optional">
+									Show <abbr title="Joins, parts, kicks, nick changes, and mode changes">status messages</abbr>
+								</label>
+							</div>
+							<div class="col-sm-6">
+								<label class="opt">
+									<input type="checkbox" name="condense">
+									Condense <abbr title="Joins, parts, kicks, nick changes, and mode changes">status messages</abbr>
 								</label>
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="motd">
 									Show <abbr title="Message Of The Day">MOTD</abbr>
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="part">
-									Show parts
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="nick">
-									Show nick changes
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="mode">
-									Show mode
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="quit">
-									Show quits
 								</label>
 							</div>
 							<div class="col-sm-6">
--- a/client/index.html
+++ b/client/index.html
@@ -218,21 +218,6 @@
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
-									<input type="checkbox" name="optional">
-									Show status messages
-									<span class="tooltipped tooltipped-n tooltipped-no-delay" aria-label="Joins, parts, kicks, nick changes, and mode changes">
-										<button class="extra-help" aria-label="Joins, parts, kicks, nick changes, and mode changes"></button>
-									</span>
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
-									<input type="checkbox" name="condense">
-									Condense status messages
-								</label>
-							</div>
-							<div class="col-sm-6">
-								<label class="opt">
 									<input type="checkbox" name="motd">
 									Show <abbr title="Message Of The Day">MOTD</abbr>
 								</label>
@@ -241,6 +226,28 @@
 								<label class="opt">
 									<input type="checkbox" name="showSeconds">
 									Show seconds in timestamp
+								</label>
+							</div>
+							<div class="col-sm-12">
+								<h2>
+									Status messages
+									<span class="tooltipped tooltipped-n tooltipped-no-delay" aria-label="Joins, parts, kicks, nick changes, and mode changes">
+										<button class="extra-help" aria-label="Joins, parts, kicks, nick changes, and mode changes"></button>
+									</span>
+								</h2>
+							</div>
+							<div class="col-sm-12">
+								<label class="opt">
+									<input type="radio" name="statusMessages" value="shown">
+									Show all status messages individually
+								</label>
+								<label class="opt">
+									<input type="radio" name="statusMessages" value="condensed">
+									Condense status messages together
+								</label>
+								<label class="opt">
+									<input type="radio" name="statusMessages" value="hidden">
+									Hide all status messages
 								</label>
 							</div>
 							<div class="col-sm-12">
--- a/client/index.html
+++ b/client/index.html
@@ -219,13 +219,16 @@
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="optional">
-									Show <abbr title="Joins, parts, kicks, nick changes, and mode changes">status messages</abbr>
+									Show status messages
+									<span class="tooltipped tooltipped-n tooltipped-no-delay" aria-label="Joins, parts, kicks, nick changes, and mode changes">
+										<button class="extra-help" aria-label="Joins, parts, kicks, nick changes, and mode changes"></button>
+									</span>
 								</label>
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="condense">
-									Condense <abbr title="Joins, parts, kicks, nick changes, and mode changes">status messages</abbr>
+									Condense status messages
 								</label>
 							</div>
 							<div class="col-sm-6">
--- a/client/js/condensed.js
+++ b/client/js/condensed.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const constants = require("./constants");
+const templates = require("../views");
 
 module.exports = {
 	updateText
@@ -33,5 +34,6 @@ function updateText(condensed, addedTypes) {
 			text += obj[messageType] > 1 ? "s" : "";
 		}
 	}
-	condensed.find(".condensed-msg").text(text);
+	condensed.find(".condensed-text")
+		.html(text + templates.msg_condensed_toggle());
 }
--- a/client/js/condensed.js
+++ b/client/js/condensed.js
@@ -1,0 +1,37 @@
+"use strict";
+
+const constants = require("./constants");
+
+module.exports = {
+	updateText
+};
+
+function updateText(condensed, addedTypes) {
+	var obj = {};
+
+	for (var i in constants.condensedTypes) {
+		var msgType = constants.condensedTypes[i];
+		obj[msgType] = condensed.data(msgType) || 0;
+	}
+
+	for (var k in addedTypes) {
+		var added = addedTypes[k];
+		obj[added]++;
+		condensed.data(added, obj[added]);
+	}
+
+	var text = "";
+
+	for (var j in constants.condensedTypes) {
+		var messageType = constants.condensedTypes[j];
+		if (obj[messageType]) {
+			text += text === "" ? "" : ", ";
+			text += obj[messageType] + " " + messageType;
+			if (messageType === "nick" || messageType === "mode") {
+				text += " change";
+			}
+			text += obj[messageType] > 1 ? "s" : "";
+		}
+	}
+	condensed.find(".condensed-msg").text(text);
+}
--- a/client/js/condensed.js
+++ b/client/js/condensed.js
@@ -8,32 +8,48 @@ module.exports = {
 };
 
 function updateText(condensed, addedTypes) {
-	var obj = {};
+	const obj = {};
 
-	for (var i in constants.condensedTypes) {
-		var msgType = constants.condensedTypes[i];
-		obj[msgType] = condensed.data(msgType) || 0;
-	}
+	constants.condensedTypes.forEach((type) => {
+		obj[type] = condensed.data(type) || 0;
+	});
 
-	for (var k in addedTypes) {
-		var added = addedTypes[k];
-		obj[added]++;
-		condensed.data(added, obj[added]);
-	}
+	addedTypes.forEach((type) => {
+		obj[type]++;
+		condensed.data(type, obj[type]);
+	});
 
-	var text = "";
-
-	for (var j in constants.condensedTypes) {
-		var messageType = constants.condensedTypes[j];
-		if (obj[messageType]) {
-			text += text === "" ? "" : ", ";
-			text += obj[messageType] + " " + messageType;
-			if (messageType === "nick" || messageType === "mode") {
-				text += " change";
+	const strings = [];
+	constants.condensedTypes.forEach((type) => {
+		if (obj[type]) {
+			switch (type) {
+			case "join":
+				strings.push(obj[type] + (obj[type] > 1 ? " users have joined the channel" : " user has joined the channel"));
+				break;
+			case "part":
+				strings.push(obj[type] + (obj[type] > 1 ? " users have left the channel" : " user has left the channel"));
+				break;
+			case "quit":
+				strings.push(obj[type] + (obj[type] > 1 ? " users have quit" : " user has quit"));
+				break;
+			case "nick":
+				strings.push(obj[type] + (obj[type] > 1 ? " users have changed nick" : " user has changed nick"));
+				break;
+			case "kick":
+				strings.push(obj[type] + (obj[type] > 1 ? " users were kicked" : " user was kicked"));
+				break;
+			case "mode":
+				strings.push(obj[type] + (obj[type] > 1 ? " modes were set" : " mode was set"));
+				break;
 			}
-			text += obj[messageType] > 1 ? "s" : "";
 		}
+	});
+
+	let text = strings.pop();
+	if (strings.length) {
+		text = strings.join(", ") + ", and " + text;
 	}
+
 	condensed.find(".condensed-text")
 		.html(text + templates.msg_condensed_toggle());
 }
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -56,6 +56,32 @@ const commands = [
 	"/whois"
 ];
 
+const actionTypes = [
+	"ban_list",
+	"invite",
+	"join",
+	"mode",
+	"kick",
+	"nick",
+	"part",
+	"quit",
+	"topic",
+	"topic_set_by",
+	"action",
+	"whois",
+	"ctcp",
+	"channel_list",
+];
+
+const condensedTypes = [
+	"join",
+	"kick",
+	"mode",
+	"nick",
+	"part",
+	"quit",
+];
+
 const timeFormats = {
 	msgDefault: "HH:mm",
 	msgWithSeconds: "HH:mm:ss"
@@ -63,6 +89,8 @@ const timeFormats = {
 
 module.exports = {
 	colorCodeMap: colorCodeMap,
-	timeFormats: timeFormats,
-	commands: commands
+	commands: commands,
+	condensedTypes: condensedTypes,
+	actionTypes: actionTypes,
+	timeFormats: timeFormats
 };
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -75,11 +75,11 @@ const actionTypes = [
 
 const condensedTypes = [
 	"join",
-	"kick",
-	"mode",
-	"nick",
 	"part",
 	"quit",
+	"nick",
+	"kick",
+	"mode",
 ];
 
 const timeFormats = {
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -439,12 +439,8 @@ $(function() {
 		}
 	});
 
-	chat.on("click", ".condensed", function() {
-		$(this).toggleClass("closed");
-	});
-
-	chat.on("click", ".condensed div", function(e) {
-		e.stopPropagation();
+	chat.on("click", ".condensed-text", function() {
+		$(this).closest(".msg.condensed").toggleClass("closed");
 	});
 
 	chat.on("click", ".user", function() {
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -439,6 +439,14 @@ $(function() {
 		}
 	});
 
+	chat.on("click", ".condensed", function() {
+		$(this).toggleClass("closed");
+	});
+
+	chat.on("click", ".condensed div", function(e) {
+		e.stopPropagation();
+	});
+
 	chat.on("click", ".user", function() {
 		var name = $(this).data("name");
 		var chan = findCurrentNetworkChan(name);
@@ -707,6 +715,9 @@ $(function() {
 	chat.on("click", ".show-more-button", function() {
 		var self = $(this);
 		var lastMessage = self.parent().next(".messages").children(".msg").first();
+		if (lastMessage.is(".condensed")) {
+			lastMessage = lastMessage.children(".msg").first();
+		}
 		var lastMessageId = parseInt(lastMessage[0].id.replace("msg-", ""), 10);
 
 		self
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -11,16 +11,13 @@ const chat = $("#chat");
 
 const options = $.extend({
 	coloredNicks: true,
+	condense: true,
 	desktopNotifications: false,
-	join: true,
 	links: true,
-	mode: true,
 	motd: true,
-	nick: true,
 	notification: true,
 	notifyAllMessages: false,
-	part: true,
-	quit: true,
+	optional: true,
 	showSeconds: false,
 	theme: $("#theme").attr("href").replace(/^themes\/(.*).css$/, "$1"), // Extracts default theme name, set on the server configuration
 	thumbnails: true,
@@ -66,15 +63,7 @@ settings.on("change", "input, select, textarea", function() {
 
 	storage.set("settings", JSON.stringify(options));
 
-	if ([
-		"join",
-		"mode",
-		"motd",
-		"nick",
-		"part",
-		"quit",
-		"notifyAllMessages",
-	].indexOf(name) !== -1) {
+	if (name === "optional" || name === "motd") {
 		chat.toggleClass("hide-" + name, !self.prop("checked"));
 	} else if (name === "coloredNicks") {
 		chat.toggleClass("colored-nicks", self.prop("checked"));
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -9,22 +9,29 @@ const tz = require("./libs/handlebars/tz");
 const windows = $("#windows");
 const chat = $("#chat");
 
-const options = $.extend({
+// Default options
+const options = {
+	autocomplete: true,
 	coloredNicks: true,
-	condense: true,
 	desktopNotifications: false,
+	highlights: [],
 	links: true,
 	motd: true,
 	notification: true,
 	notifyAllMessages: false,
-	optional: true,
 	showSeconds: false,
+	statusMessages: "condensed",
 	theme: $("#theme").attr("href").replace(/^themes\/(.*).css$/, "$1"), // Extracts default theme name, set on the server configuration
 	thumbnails: true,
 	userStyles: userStyles.text(),
-	highlights: [],
-	autocomplete: true
-}, JSON.parse(storage.get("settings")));
+};
+const userOptions = JSON.parse(storage.get("settings")) || {};
+
+for (const key in options) {
+	if (userOptions[key]) {
+		options[key] = userOptions[key];
+	}
+}
 
 module.exports = options;
 
@@ -40,6 +47,9 @@ for (var i in options) {
 		settings.find("#user-specified-css-input").val(options[i]);
 	} else if (i === "highlights") {
 		settings.find("input[name=" + i + "]").val(options[i]);
+	} else if (i === "statusMessages") {
+		settings.find(`input[name=${i}][value=${options[i]}]`)
+			.prop("checked", true);
 	} else if (i === "theme") {
 		$("#theme").attr("href", "themes/" + options[i] + ".css");
 		settings.find("select[name=" + i + "]").val(options[i]);
@@ -55,6 +65,10 @@ settings.on("change", "input, select, textarea", function() {
 
 	if (type === "password") {
 		return;
+	} else if (type === "radio") {
+		if (self.prop("checked")) {
+			options[name] = self.val();
+		}
 	} else if (type === "checkbox") {
 		options[name] = self.prop("checked");
 	} else {
@@ -63,8 +77,10 @@ settings.on("change", "input, select, textarea", function() {
 
 	storage.set("settings", JSON.stringify(options));
 
-	if (name === "optional" || name === "motd") {
+	if (name === "motd") {
 		chat.toggleClass("hide-" + name, !self.prop("checked"));
+	} else if (name === "statusMessages") {
+		chat.toggleClass("hide-status-messages", options[name] === "hidden");
 	} else if (name === "coloredNicks") {
 		chat.toggleClass("colored-nicks", self.prop("checked"));
 	} else if (name === "theme") {
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -6,11 +6,14 @@ const options = require("./options");
 const renderPreview = require("./renderPreview");
 const utils = require("./utils");
 const sorting = require("./sorting");
+const constants = require("./constants");
+const condense = require("./condensed");
 
 const chat = $("#chat");
 const sidebar = $("#sidebar");
 
 module.exports = {
+	appendMessage,
 	buildChannelMessages,
 	buildChatMessage,
 	renderChannel,
@@ -19,14 +22,37 @@ module.exports = {
 	renderNetworks,
 };
 
-function buildChannelMessages(channel, messages) {
-	return messages.reduce(function(docFragment, message) {
-		docFragment.append(buildChatMessage({
-			chan: channel,
+function buildChannelMessages(data) {
+	return data.messages.reduce(function(docFragment, message) {
+		appendMessage(docFragment, data.id, data.type, message.type, buildChatMessage({
+			chan: data.id,
 			msg: message
 		}));
 		return docFragment;
 	}, $(document.createDocumentFragment()));
+}
+
+function appendMessage(container, chan, chanType, messageType, msg) {
+	if (constants.condensedTypes.indexOf(messageType) !== -1 && chanType !== "lobby") {
+		var condensedTypesClasses = "." + constants.condensedTypes.join(", .");
+		var lastChild = container.children("div.msg").last();
+		var lastDate = (new Date(lastChild.attr("data-time"))).toDateString();
+		var msgDate = (new Date(msg.attr("data-time"))).toDateString();
+		if (lastChild && $(lastChild).hasClass("condensed") && !$(msg).hasClass("message") && lastDate === msgDate) {
+			lastChild.append(msg);
+			condense.updateText(lastChild, [messageType]);
+		} else if (lastChild && $(lastChild).is(condensedTypesClasses) && options.condense) {
+			var condensed = buildChatMessage({msg: {type: "condensed", time: msg.attr("data-time"), previews: []}, chan: chan});
+			condensed.append(lastChild);
+			condensed.append(msg);
+			container.append(condensed);
+			condense.updateText(condensed, [messageType, lastChild.attr("data-type")]);
+		} else {
+			container.append(msg);
+		}
+	} else {
+		container.append(msg);
+	}
 }
 
 function buildChatMessage(data) {
@@ -45,25 +71,13 @@ function buildChatMessage(data) {
 		data.msg.highlight = true;
 	}
 
-	if ([
-		"invite",
-		"join",
-		"mode",
-		"kick",
-		"nick",
-		"part",
-		"quit",
-		"topic",
-		"topic_set_by",
-		"action",
-		"whois",
-		"ctcp",
-		"channel_list",
-		"ban_list",
-	].indexOf(type) !== -1) {
+	if (constants.actionTypes.indexOf(type) !== -1) {
+		data.msg.template = "actions/" + type;
 		template = "msg_action";
 	} else if (type === "unhandled") {
 		template = "msg_unhandled";
+	} else if (type === "condensed") {
+		template = "msg_condensed";
 	}
 
 	const msg = $(templates[template](data.msg));
@@ -100,7 +114,7 @@ function renderChannel(data) {
 }
 
 function renderChannelMessages(data) {
-	const documentFragment = buildChannelMessages(data.id, data.messages);
+	const documentFragment = buildChannelMessages(data);
 	const channel = chat.find("#chan-" + data.id + " .messages").append(documentFragment);
 
 	if (data.firstUnread > 0) {
@@ -109,6 +123,8 @@ function renderChannelMessages(data) {
 		// TODO: If the message is far off in the history, we still need to append the marker into DOM
 		if (!first.length) {
 			channel.prepend(templates.unread_marker());
+		} else if (first.parent().hasClass("condensed")) {
+			first.parent().before(templates.unread_marker());
 		} else {
 			first.before(templates.unread_marker());
 		}
@@ -129,6 +145,10 @@ function renderChannelMessages(data) {
 			}
 
 			if (lastDate.toDateString() !== msgDate.toDateString()) {
+				var parent = msg.parent();
+				if (parent.hasClass("condensed")) {
+					msg.insertAfter(parent);
+				}
 				msg.before(templates.date_marker({msgDate: msgDate}));
 			}
 
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -41,7 +41,7 @@ function appendMessage(container, chan, chanType, messageType, msg) {
 		if (lastChild && $(lastChild).hasClass("condensed") && !$(msg).hasClass("message") && lastDate === msgDate) {
 			lastChild.append(msg);
 			condense.updateText(lastChild, [messageType]);
-		} else if (lastChild && $(lastChild).is(condensedTypesClasses) && options.condense) {
+		} else if (lastChild && $(lastChild).is(condensedTypesClasses) && options.statusMessages === "condensed") {
 			var condensed = buildChatMessage({msg: {type: "condensed", time: msg.attr("data-time"), previews: []}, chan: chan});
 			condensed.append(lastChild);
 			condensed.append(msg);
--- a/client/js/renderPreview.js
+++ b/client/js/renderPreview.js
@@ -47,7 +47,7 @@ function renderPreview(preview, msg) {
 	container.trigger("keepToBottom");
 }
 
-$("#chat").on("click", ".toggle-button", function() {
+$("#chat").on("click", ".text .toggle-button", function() {
 	const self = $(this);
 	const container = self.closest(".chat");
 	const content = self.closest(".content")
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -101,10 +101,6 @@ body {
 	color: #428bca;
 }
 
-#chat button:hover {
-	opacity: 1;
-}
-
 /* Increase contrast of some IRC colors */
 .irc-fg2 { color: #0074d9; }
 .irc-fg5 { color: #e969a7; }
@@ -208,17 +204,9 @@ body {
 }
 
 /* Embeds */
-#chat .toggle-content,
-#chat .toggle-button {
-	color: #f3f3f3;
-}
-
-#chat .toggle-button:hover {
-	color: rgba(243, 243, 243, .8);
-}
-
 #chat .toggle-content {
 	background: #242a33;
+	color: #f3f3f3;
 }
 
 #chat .toggle-content .body {
--- a/client/views/msg_condensed.tpl
+++ b/client/views/msg_condensed.tpl
@@ -1,7 +1,7 @@
 <div class="msg {{type}} closed" data-time="{{time}}">
 	<span class="time hide-text">{{tz time}}</span>
-	<span class="from"><span class="expander"></span></span>
+	<span class="from"></span>
 	<span class="content">
-		<span class="text condensed-msg"></span>
+		<span class="condensed-text"></span>
 	</span>
 </div>
--- a/client/views/msg_condensed.tpl
+++ b/client/views/msg_condensed.tpl
@@ -1,0 +1,7 @@
+<div class="msg {{type}} closed" data-time="{{time}}">
+	<span class="time hide-text">{{tz time}}</span>
+	<span class="from"><span class="expander"></span></span>
+	<span class="content">
+		<span class="text condensed-msg"></span>
+	</span>
+</div>
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -7,7 +7,7 @@ const chat = $("#chat");
 const templates = require("../../views");
 
 socket.on("more", function(data) {
-	const documentFragment = render.buildChannelMessages(data.chan, data.messages);
+	const documentFragment = render.buildChannelMessages(data);
 	const chan = chat
 		.find("#chan-" + data.chan)
 		.find(".messages");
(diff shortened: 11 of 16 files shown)


MaxLeiter:
Is it not possible to do something more similar to IRCCloud, where we just combine all the parts and joins into a like or two? 

However, I due kind of like the button, however I see it being difficult on mobile. I also think the text could be better (we shouldn't use etc)

client/js/lounge.js
```diff
@@ -276,9 +278,27 @@ $(function() {
 		return msg;
 	}
 
-	function buildChannelMessages(channel, messages) {
+	function appendMessage(container, chan, type, msg) {
+		if (!$(msg).hasClass("message") && type !== "lobby") {
```
xPaw: Topic changes and other stuff? I think whitelist is better suited here?

client/js/lounge.js
```diff
@@ -292,7 +312,7 @@ $(function() {
 	}
 
 	function renderChannelMessages(data) {
-		var documentFragment = buildChannelMessages(data.id, data.messages);
+		var documentFragment = buildChannelMessages(data.id, data.messages, data.type);
```
xPaw: Just pass `data`

YaManicKill:
> Is it not possible to do something more similar to IRCCloud, where we
just combine all the parts and joins into a like or two?

That's what I meant when I talked about irc cloud in the pr. It's a bit
more complex, but would be a nice next step. This is the first step. Fwiw,
you can expand the irc cloud one as well, so it is just a first step
towards replicating that feature.

MaxLeiter:
Can we remove the button and have the actual text be the button? I think that's what IRCCloud does

spangborn:
I think it might be better to change the text to be a summary of what's condensed, something like

> Foo and bar left the channel.

And then when you open it up, you can see the timestamps and quit messages, or whatever detail gets tucked away.

YaManicKill:
> I think it might be better to change the text to be a summary of what's condensed

@spangborn As I've already said, that is a good next step, but I think this is a useful first step before it. I won't have the time to add that just now, and this is better than nothing.

> Can we remove the button and have the actual text be the button? I think that's what IRCCloud does

@MaxLeiter  Happy to make the text clickable **as well**, but I don't want to remove the button. The button is a key indicator that you can expand it.

client/js/lounge.js
YaManicKill: Aight

client/js/lounge.js
xPaw: Hmm yeah, maybe multiple mode changes and kicks be a good idea to condense too. But whitelist is still safer in this case. So we don't unknowingly break anything if we add new features.

YaManicKill:
Alright, I think I've updated it with people's feedback (except the natural language explanation of what's inside the condensed bit).

I think we need to remember that it's good to make incremental improvements, and as @astorije has said a lot, we don't want to wait until things have all of the features that we want to put them in.

Adding the natural "x, y and z joined, a and b left" will be a good next step, and I will take some time to work that out, but that has multiple issues:

1. We need to have a discussion as to how we want to word it, and that's going to end up  very subjective, so would bog down this PR for aaaages
2. It's a bit more expensive computationally than this, because we'll need to re-write the wording everytime we get a new one
3. It's just more complicated in general.

So yes, that will come. But I want to do that separately, once this one is done.

But yeah, the best thing is if some people can run this and see if there are any bugs I've missed. It seems to work quite well so far.

I've removed "etc." from the wording, because that's a bit awkward. If anyone has a better idea for how we should word it for this PR (before we do the natural language explanation), that would be great.

client/js/lounge.js
YaManicKill: Aight, I've done joins, parts, quits, modes, kicks and nick changes.

client/js/lounge.js
```diff
@@ -276,10 +278,29 @@ $(function() {
 		return msg;
 	}
 
-	function buildChannelMessages(channel, messages) {
-		return messages.reduce(function(docFragment, message) {
-			docFragment.append(buildChatMessage({
-				chan: channel,
+	function appendMessage(container, chan, type, msg) {
+		if ($(msg).is(".join, .part, .quit, .mode, .kick, .nick") && type !== "lobby") {
```
xPaw: Compare `type`, you don't need to use DOM.

YaManicKill: Type here is actually the type of channel. But I can check against the message.

YaManicKill: Ah wait, the message we get through is the dom element anyway. We don't get the information of the message without it being an element.

client/views/msg_condensed.tpl
```diff
@@ -0,0 +1,5 @@
+<div class="msg {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}} closed" id="msg-{{id}}">
+	<span class="time"></span>
+	<span class="from"><span class="expander"></span></span>
+	<span class="text">Condensed joins/parts</span>
```
YaManicKill: Would you be happy with me just saying "Condensed joins/parts/mode/nick changes" for this PR?

xPaw: No, that's just wrong. Print actual events (even without counter)

YaManicKill: > Print actual events (even without counter)

What do you mean?

xPaw: Find out which events are in this condensed message and print these.

YaManicKill: Oh. Meh, that's more work. But fine, I'll do it.

YaManicKill:
Alright, I've added a simple "1 join, 1 part" as the message, so that we at least have some sort of information.

I still think it's a good idea to do a natural language one at some point, but still not this PR, because we need to figure out what the wording should be.

YaManicKill:
So, current text will show (if they are all there) "1 join, 1 part, 1 quit, 1 mode change, 1 nick change and 1 kick". They will all pluralise if there are multiple, and if there are 0, it'll just not show at all. Think this is a good one for just now?

YaManicKill:
Alrighty, I have a bug now, which is fun. It doesn't work perfectly with the date separator. When there is a part/join straight after a date separator, the expanded elements show as below the date, but the condenser element shows as behind it. The condenser should show after the date separator (unless part of it is in the past, I guess?)

xPaw:
Trying this PR out. I think it's pointless to condense messages if it only condenses one message.



It also condensed `topic_set_by` but didn't say it did.

YaManicKill:
>  think it's pointless to condense messages if it only condenses one message.

My reason for doing this was to make it consistent (it felt a bit odd having it look different when there was only 1) but I can change it.

> It also condensed topic_set_by but didn't say it did.

Good point, I'll remove it.

xPaw:
>My reason for doing this was to make it consistent (it felt a bit odd having it look different when there was only 1) but I can change it.

Ideally we would do something similar to IRCCloud.

YaManicKill:
> Ideally we would do something similar to IRCCloud.

Yeah,that's the ultimate plan. But happy to remove the condensing for 1 thing before that happens.

I'll hopefully get these bugs fixed this week.

astorije:
This is really nice on busy channels! Good job 👏
I'm on #ansible and #freenode, it's the wild west in there 😅.

I do agree with not condensing less than 2 items (I'd even say 3 or 5 if it were me). If the goal is to reduce noise on busy channels, then channels with very little activity should would not benefit from this much.

In terms of messages, I think we should only condense joins, parts and quits together, at least at first. I don't think topic changes, mode changes or other kind of messages are as noisy as those 3. (Actually, camping in those 2 channels mentioned above), 

I'm happy to re-evaluate the need to condense nicks and modes, or others, but starting with joins/parts/quits is a strong improvement already and lets us experiment how it behaves that way.

One bug though: When unfolding the messages, all messages are now clickable. I agree that the summary should be clickable (though when we start listing nicks in the summary, there will be a choice to make), but not everything it's condensing IMO.

In terms of icons, can you [right](http://fontawesome.io/icon/caret-right/) and [down](http://fontawesome.io/icon/caret-down/) carets at the message level (where it's clickable) to give it a dropdown purpose?

Folded | Unfolded
--- | ---
screen shot 2016-12-17 at 13 48 01 | screen shot 2016-12-17 at 13 48 27

YaManicKill:
Alrighty. 

I have done all of the things people asked (I'm not sure about not condensing modes and nick changes, but I'll see how it goes for a while).

There is 1 definite bug left, and 1 possible bug left (I need to test that overnight).

Feel free to retry guys.

YaManicKill:
Having used this for only a few hours, it almost defeats the purpose if we don't include mode changes and nick changes.

It breaks up the condensing so much, that we don't save a lot of space. So I'm going to add it back in.

YaManicKill:
Example of what it looks like in a quiet channel without mode/nick changes condensed. IT just looks messy, IMO.

6b567366724ee97615df327e220612ff

YaManicKill:
So, I added nick changes and mode changes back in, and I fixed the bugs. I think that's everything we had wrong with it. So, it should be ready for proper review.

client/js/lounge.js
```diff
@@ -310,10 +312,61 @@ $(function() {
 		return msg;
 	}
 
-	function buildChannelMessages(channel, messages) {
-		return messages.reduce(function(docFragment, message) {
-			docFragment.append(buildChatMessage({
-				chan: channel,
+	function updateCondensedText(condensed, type) {
+		console.log(type);
```
MaxLeiter: Should be removed

YaManicKill: Doh. Fixed.

xPaw:
Few thoughts:
1. I think condensing should only apply to channels (it condensed modes in lobby, no reason to)
2. Expanded messages don't act like text (e.g. no text cursor, clicking on it does not focus input)
3. Condensed line does not have a timestamp, maybe it should? Either of the first or last event.

Other than that, this seems quite solid from my limited testing.

client/js/lounge.js
```diff
@@ -310,10 +312,60 @@ $(function() {
...
+		$(condensed.children(".text")[0]).text(text);
+	}
+
+	function appendMessage(container, chan, chanType, messageType, msg) {
+		if (["join", "part", "quit", "nick", "mode"].indexOf(messageType) !== -1 && chanType !== "lobby") {
+			var lastChild = container.children("div.msg").last();
+			if (lastChild && $(lastChild).hasClass("condensed") && !$(msg).hasClass("message")) {
+				lastChild.append(msg);
+				updateCondensedText(lastChild, messageType);
+			} else if (lastChild && $(lastChild).is(".join, .part, .quit, .nick, .mode")) {
```
xPaw: There's two separate lists of which types are condensed (here, and one in `updateCondensedText`). For maintainability purposes, this could be defined once.

client/js/lounge.js
```diff
@@ -310,10 +312,60 @@ $(function() {
...
+				var condensed = buildChatMessage({msg: {type: "condensed", time: msg.attr("data-time")}, chan: chan});
+				condensed.append(lastChild);
+				updateCondensedText(condensed, lastChild.attr("data-type"));
+				condensed.append(msg);
+				container.append(condensed);
+				updateCondensedText(condensed, messageType);
```
xPaw: Not entirely sure what's happening here, but why the two updates?

client/js/lounge.js
```diff
@@ -310,10 +312,60 @@ $(function() {
...
+					text += " change";
+				}
+				text += obj[messageType] > 1 ? "s" : "";
+			}
+		}
+		$(condensed.children(".text")[0]).text(text);
+	}
+
+	function appendMessage(container, chan, chanType, messageType, msg) {
+		if (["join", "part", "quit", "nick", "mode"].indexOf(messageType) !== -1 && chanType !== "lobby") {
```
xPaw: I just noticed this `chanType` check, I somehow got modes condensed in a lobby when I tested (it was a second network with no channels on it).

client/js/lounge.js
```diff
@@ -355,6 +407,11 @@ $(function() {
 				}
 
 				if (lastDate.toDateString() !== msgDate.toDateString()) {
+					var parent = msg.parent();
+					if (msg.parent().hasClass("condensed")) {
```
xPaw: Use `parent` you defined.

client/js/lounge.js
```diff
@@ -355,6 +407,11 @@ $(function() {
...
+						var detached = msg.detach();
```
xPaw: Does it not work to directly do `msg.insertAfter(parent);`?

client/js/lounge.js
```diff
@@ -428,16 +485,21 @@ $(function() {
 		}
 
 		if (prevMsgTime.toDateString() !== msgTime.toDateString()) {
-			prevMsg.append(render("date-marker", {msgDate: msgTime}));
+			var parent = prevMsg.parent();
+			if (parent.hasClass("condensed")) {
+				parent.append(render("date-marker", {msgDate: msgTime}));
+			} else {
+				prevMsg.append(render("date-marker", {msgDate: msgTime}));
```
xPaw: You can reduce the duplicate appends by simply doing `parent = prevMsg;` (with a better variable name) in the `if`.

client/js/lounge.js
```diff
@@ -490,6 +552,11 @@ $(function() {
 			}
 
 			if (lastDate.toDateString() !== msgDate.toDateString()) {
+				var parent = msg.parent();
+				if (msg.parent().hasClass("condensed")) {
```
xPaw: Use `parent` variable.

client/js/lounge.js
```diff
@@ -310,10 +312,60 @@ $(function() {
...
+			if (obj[messageType]) {
+				text += text === "" ? "" : ", ";
+				text += obj[messageType] + " " + messageType;
+				if (messageType === "nick" || messageType === "mode") {
+					text += " change";
+				}
+				text += obj[messageType] > 1 ? "s" : "";
+			}
+		}
+		$(condensed.children(".text")[0]).text(text);
```
YaManicKill: Yeah, makes sense.

client/js/lounge.js
YaManicKill: Yeah, I thought about this...but I got lazy :-p

client/js/lounge.js
YaManicKill: Ah yes. The reason is that the updateCondensedText takes in a message and updates using that. So, here we are adding 2 messages (because we have 2 messages to put in the condensed section because we are waiting until we have 2 messages before condensing). I'll think up a nicer way to do this.

client/js/lounge.js
YaManicKill: Ah, boo. I was half asleep when I did that, clearly.

client/js/lounge.js
YaManicKill: Huh, I assumed insertAfter cloned the element. Turns out it doesn't.

client/js/lounge.js
YaManicKill: Yeah, good point.

client/js/lounge.js
YaManicKill: Late night coding = bad.

YaManicKill:
Updated with fixes and issues that @xPaw pointed out.

client/js/lounge.js
YaManicKill: Yeah, this was working on initial page load, but not on individual new messages because I wasn't passing in the chanType there. Fixed.

client/js/lounge.js
```diff
@@ -37,6 +37,22 @@ $(function() {
...
+		"part",
+		"quit",
+		"topic",
+		"topic_set_by",
+		"action",
+		"whois",
+		"ctcp",
+		"channel_list",
+	];
+	var condensedTypes = ["join", "part", "quit", "nick", "mode"];
```
xPaw: Use new lines too. Adding new event won't be a silly diff.

client/js/lounge.js
```diff
@@ -37,6 +37,22 @@ $(function() {
 		"/voice",
 		"/whois"
 	];
+	var handledTypes = [
```
xPaw: What if we made it an object and set each property to `true`, then testing would be `handledTypes[type]`, which does not require calling `indexOf`?

client/js/lounge.js
```diff
@@ -310,10 +314,64 @@ $(function() {
...
+			if (obj[messageType]) {
+				text += text === "" ? "" : ", ";
+				text += obj[messageType] + " " + messageType;
+				if (messageType === "nick" || messageType === "mode") {
+					text += " change";
+				}
+				text += obj[messageType] > 1 ? "s" : "";
+			}
+		}
+		$(condensed.children(".condensed-msg")).text(text);
```
xPaw: You dont need to wrap jquery result in a new jquery object.

client/js/lounge.js
```diff
@@ -310,10 +314,64 @@ $(function() {
...
+	}
+
+	function appendMessage(container, chan, chanType, messageType, msg) {
+		var condensedTypesClasses = "." + condensedTypes.join(", .");
+		console.log(chanType);
```
xPaw: :zap:

client/js/lounge.js
```diff
@@ -310,10 +314,64 @@ $(function() {
...
+					text += " change";
+				}
+				text += obj[messageType] > 1 ? "s" : "";
+			}
+		}
+		$(condensed.children(".condensed-msg")).text(text);
+	}
+
+	function appendMessage(container, chan, chanType, messageType, msg) {
+		var condensedTypesClasses = "." + condensedTypes.join(", .");
```
xPaw: Throw this inside the `if` so that you don't unnecessarily join a string every time a message arrives.

client/js/lounge.js
```diff
@@ -428,16 +490,20 @@ $(function() {
 		}
 
 		if (prevMsgTime.toDateString() !== msgTime.toDateString()) {
-			prevMsg.append(render("date-marker", {msgDate: msgTime}));
+			var parent = prevMsg.parent();
+			if (parent.hasClass("condensed")) {
+				prevMsg = parent;
+			}
+			prevMsg.after(render("date-marker", {msgDate: msgTime}));
```
xPaw: You sure this shouldn't be `append`?

YaManicKill: Yeah, because append adds **into** something, but we don't want the date-marker to be inside the messae, we want it to be after it.

xPaw: In that case shouldn't it be `after(prevMsg)`? I'm just not seeing this ends up in the DOM.

client/js/lounge.js
YaManicKill: We should add this to linter, put in a proper logging framework, and not allow any console logs.

client/js/lounge.js
YaManicKill: Good point.

client/js/lounge.js
YaManicKill: Ah yes, fun times.

client/js/lounge.js
YaManicKill: Meh, fine. I've never really liked this way of doing things, but whatever.

client/js/lounge.js
YaManicKill: I'm not a fan of doing that, tbh. It sounds a bit messy, adds a whole bunch of extra code just for removing indexOf? Is it because of the name? When we get babel in, we can use includes, which is better...

xPaw: Mostly performance. Looking up a key is a lot faster than iterating the entire array to find it.

client/js/lounge.js
xPaw: Nevermind, `after` is magic and works differently if you pass in a selector into it.

YaManicKill:
Finally figured out another bug I was seeing and the reason why it exists. I'll get it fixed once I unpack my new house.

When you click "more" it sends to the server the number of messages that we currently have, and the server then removes them from the list and sends the next 100. This doesn't work, because the selector just takes the children of .messages, which wouldn't count the condensed ones, obviously. [This is the offending line](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L1193)

xPaw:
Chance for improvement, send message ID instead of length.

YaManicKill:
> Chance for improvement, send message ID instead of length.

Yeah, I was thinking that, because we'll come back across this issue when we remove parts/joins with js rather than css. So that's a better fix.

YaManicKill:
Hoping to get this done soon.

MiniDigger:
this doesn't seem to work on mobile for me (tested with all default themes). it display the 3 joins 5 leaves thing, but it doesn't hide them when I click it. seems like a styling issue.

YaManicKill:
> but it doesn't hide them when I click it

Hmm, what do you mean by this? Do you mean the joins/parts are not hidden? They should be hidden by default, and show when you click on it. It works fine on mine, what version of Chrome are you using?

YaManicKill:
Right, I have switched back to the default css, updated Chrome and removed all my css tweaks and I still can't reproduce that, @MiniDigger . Is it a custom css thing for you, perhaps?

MiniDigger:
I did use a theme, but I tested with the default ones too. I also have css modifications. will test again when I am home. 
the bug for me is that I could not hide them at all, they are not hidden by default and clicking the thingy does nothing. it's only like that for mobile (or sub 479px), no a larger screen it works as it should. will send you the chrome versions when I tested at home.

YaManicKill:
> (or sub 479px)

Aha, my phone is higher resolution than this, s othat's why I never saw it. I'll take a look at fixing it later in the week, thanks :-)

williamboman:
screenshot 2017-03-22 at 23 46 17

FYI, the UI breaks together with #523.

williamboman:
Also, this doesn't seem to respect the `Show joins/parts` settings, I can still see the condensed indicator when having them disabled.

YaManicKill:
Alright, I've fixed the obvious issues with the new flex layout, that works fine.

> Also, this doesn't seem to respect the Show joins/parts settings

Yeah, this is a bigger issue...

The problem being that we still deal with hiding them in css, which is fine when it's individual things, but when we have a condensed one, we have to hide it only when everything that is inside of it is hidden as well. I might have to redo all of the hiding of these things entirely. Or something... goodness, how I hate how we have done part/joins.

Thanks for finding this, @williamboman, I don't use any hiding :-P

astorije:
@YaManicKill, as a suggestion, would it be possible to disable your condensing if any of join, parts, quits, etc. are disabled?
In the future, I definitely see this condensing replacing these way-too-many options with your work entirely, but until your work is bulletproof, falling back to the current system when not all enabled could work.

What do you think?

astorije:
@YaManicKill, as a suggestion, would it be possible to disable your condensing if any of join, parts, quits, etc. are disabled?
In the future, I definitely see your condensing work replace these way-too-many options entirely, but until it is bulletproof, falling back to the current system when not all enabled could work.

What do you think?

YaManicKill:
> as a suggestion, would it be possible to disable your condensing if any of join, parts, quits, etc. are disabled?

Could do, but that feels a little like giving up on it. But I'm not sure.

We definitely need to change the part/join showing/hiding because we can't do everything we want in css.

MiniDigger:
you could condense every of those special messages (even if its only one) and then add a settings option to hide the condensed element (replacing all the other hide options)

astorije:
> Could do, but that feels a little like giving up on it. But I'm not sure.

Sorry if it felt like that, I actually mean the exact opposite :)
I'm saying disabling it when people hide some types of actions *for now* so if they want your cool new feature, they need to un-hide everything.
Then in a (near?) future, I don't think we'll want to support both ways, your nicely condensed way + the current way to display actions (meaning joins/quits/etc. but "actions" is shorter). Instead, we'll want **your** way to be the only one, and provide customizations (maybe) on what should be displayed in your condensed message.

So it really isn't giving up, it's trying to find a way to make it work as soon as possible without having to solve the world of CSS hiding first. I think in the long run we'll make the current solution evolve into yours rather than maintaining multiple ways to display these things. It should be simple and helpful by default UX-wise, and focus on that solution (yours).

Does that make more sense now?

YaManicKill:
Fixed a bug with the unread marker.

YaManicKill:
> I'm saying disabling it when people hide some types of actions for now so if they want your cool new feature, they need to un-hide everything.

So, this definitely works, but it does bring up another weird edge case.

If you have any of part/join/quit/nick/mode hidden, and then you show them all, the old ones won't be condensed until you refresh. I could set up a thing that sets them all to be, but that feels like an awful lot of work for a weird edge case. Thoughts?

YaManicKill:
Ah man, you also end up with an empty condensed if you turn them off when you already have condensed messages before. This is horrible. Can we just remove that feature altogether? :-P

YaManicKill:
I've added the fix suggested by @astorije 

It's a pretty horrible fix, but without rewriting the entirety of how we show/hide certain messages, I think it'll be impossible to deal with nicely.

astorije:
@YaManicKill, refresh the page completely when going from one to the other? I don't know, it's not ideal for sure, but if the plan is to transition from the current model to yours once it's given some time to check robustness, fix doesn't have to be the cleanest thing ever. :man_shrugging:

YaManicKill:
> refresh the page completely when going from one to the other?

I think I'd rather keep what I currently have, as forcibly refreshing the page is not ideal, and could be very confusing to the user.

YaManicKill:
So, I've pulled in the fix for the wrapping of text from @bews 

The functionality on this is done, I think. I just want to merge in #1175 and then I can throw this out into a module and I'll be happy with the code.

YaManicKill:
Alrighty, I have rebased this, and is properly modularised (only 10 extra lines in lounge.js).

I'll spend some time testing this, there are bound to be bugs because I've mucked around with this so much. Other eyes would be good. But I think this is pretty much there if I haven't introduced any new issues.

client/js/constants.js
```diff
@@ -54,13 +54,40 @@ const commands = [
 	"/whois"
 ];
 
+const handledTypes = [
```
xPaw: Can this name be more descriptive?

bews:
@YaManicKill have you fixed problems I reported here https://github.com/thelounge/lounge/pull/1047#issuecomment-301357846 ?

YaManicKill:
> @YaManicKill have you fixed problems I reported here #1047 (comment) ?

Some of them are unique to #1047 (because it shows the usernames), some of them are fixed, I'll list the rest below.

> Kick message is not condensed (is this intended?)

Good point. I think we had a discussion about it somewhere, but not sure. I'm in favour of condensing it, I think it makes sense.

> Show more button still counts hidden messages:

Yeah, currently we send 100 messages. We should probably change it to either be 100 non-condensed messages, or a smaller number of non-condensed if we have more than 100 messages. The problem is, if we have a very low-traffic chan, we don't want to end up with, say 10000 messages on init because there were so few non-condensed messages. Maybe we can solve this by the condensed messages not being on the dom when not expanded. Not sure how slow that would be. I'll look into that.

> Date marker is broken

I think this is fixed? Let me know if you see it again.

> I think there should be an option in settings to disable this new feature

Why? I'm not saying we won't, but I want you to come up with a good reason to add a setting. 

> In #1141 time column doesn't have a default width anymore, and this PR leaves time column empty:

Yeah, this isn't fixed, I had kinda noticed it before but forgotten about it. I'll fix this.

bews:
> Let me know if you see it again.

I'm not updating anymore, so I can't test this.

> Why? I'm not saying we won't, but I want you to come up with a good reason to add a setting.

1) Not everyone is fan of this feature, for example I was strongly against it some time ago
2) There may be bugs (since it's a new feature) and it will be good to disable it untill they're fixed
3) IRCcloud has option for this too

------------------

Also, more bugs here https://github.com/thelounge/lounge/pull/1047#issuecomment-301438798 (in the bottom)

YaManicKill:
> Show more button still counts hidden messages:

I think I have fixed everything that anyone has asked of me apart from this. (And a button to turn it off)

I'm not sure what the best thing to do with this is. Thoughts from people? All of the options I outlined above have some downside.

christer88:
When client condense pretty much every join/part/mode etc. I see few reasons for keeping the "Show joins", "Show parts" etc. In my opinion it's better to have "Show joins, parts and modes" and "Collapse".

I'm in favour of condensing kicks and kills, as long as it's possible to click on the condensed message to expand the list and see the kick/kill messages.

bews:
Maybe I'm doing smth wrong, but I just tested this... Have you even tested this? Bugs, bugs everywhere...

Also, this branch doesn't work, because previews are broken:


Seriously, just take [this PR ](https://github.com/thelounge/lounge/pull/1047), apply my fixes, fix tests - and it will be done. I've been using that for a month now and it's perfect, no bugs or anything.

astorije:
I don't know what you really expect with such disrespectful comments all the time, but you won't get anything positive out of it for sure. Maybe that's one of the reasons why [many more of your PRs got closed rather than merged](https://github.com/thelounge/lounge/pulls?q=is%3Apr+author%3Abews+is%3Aclosed)... It doesn't have to be.

For the issue in your screenshot, I have seen that a lot the last few days when jumping from branch to branch, and that was because server needed restart (`previews` in plural is new and not all PRs have been rebased).

MiniDigger:
I am running an old version of the pr and it works flawlessly. Seems like something happened during all the rebasing.  Guess that's what happens if a pr is open for month. 
Yelling at people will certainly not make it run better

YaManicKill:
> Maybe I'm doing smth wrong, but I just tested this... Have you even tested this? Bugs, bugs everywhere...

I don't appreciate the attitude.

However, yes there was an issue in my rebasing (as I rebased in a hurry last night).

I have updated it.

But as @astorije said, that's a different issue you are seeing that isn't on my branch,

YaManicKill:
(Note that I still haven't done extensive testing of the rebase, but will do so through the day as I'm now running this on my prod server)

YaManicKill:
> Also, this branch doesn't work, because previews are broken:

Actually, @bews is right on this one. There is a bug because the new condense messages don't have the new previews set on them.

bews:
> I don't know what you really expect with such disrespectful comments all the time, but you won't get anything positive out of it for sure. Maybe that's one of the reasons why many more of your PRs got closed rather than merged.

I expect the same quality level as I provide in my PRs. **Each** of my PRs is thoroughly tested a lot of times and completely production ready. 
And things I see here after reading
> I think I have fixed everything that anyone has asked

is completely different from what I expected. This PR is not ready at all, and I wasted my time on testing it.
I wrote comment which pretty accurate describes the situation with this PR and you call it "disrespectful". You should thank me for spending my time on this and reporting bugs, instead of that.

Regarding my closed PRs, this is not my PR, so it is not the right place to discuss them.

> For the issue in your screenshot, I have seen that a lot the last few days when jumping from branch to branch, and that was because server needed restart

I know it needs a restart, and it was restarted. 2 times to be sure.

YaManicKill:
> I expect the same quality level as I provide in my PRs. Each of my PRs is thoroughly tested a lot of times and completely production ready.

Yup, and this is a PR that has been tested over several months in production. I rebased last night and didn't test after the rebase, which was my bad, I accept that. But that's not the same as "Have you even tested this?". It looked like a simple rebase, was late at night, so I pushed without testing and then went to bed. That was wrong, but your attitude wasn't one of "Oh hey, btw, the branch seems to be broken", it was one of entitlement.

> Seriously, just take this PR , apply my fixes, fix tests - and it will be done. I've been using that for a month now and it's perfect, no bugs or anything.

You do know that #1047 is based off this PR, yeah? 

We do appreciate testing and reporting bugs with PRs, but sometimes your attitude stinks and it actively makes the atmosphere less than enjoyable to work in.

bews:
> I rebased last night and didn't test after the rebase, which was my bad, I accept that. But that's not the same as "Have you even tested this?". It looked like a simple rebase, was late at night, so I pushed without testing and then went to bed. 

I wrote this not because branch has preview problem, no. It was just addition to the comment as you may see.

I wrote this because bugs you reported fixed are not fixed. For example this:

> In #1141 time column doesn't have a default width anymore, and this PR leaves time column empty:

> Yeah, this isn't fixed, I had kinda noticed it before but forgotten about it. I'll fix this.

After you said that everything is fixed I expect this to be fixed too. But it's not, and it's the first thing you notice when start testing. Conclusion: you haven't tested this at all.

> You do know that #1047 is based off this PR, yeah?

Yes I do, I'm just saying that it will be completely production ready if you do so. Thus, testing and fixing this instead, looks like wasting time.

YaManicKill:
> After you said that everything is fixed I expect this to be fixed too. But it's not, and it's the first thing you notice when start testing. Conclusion: you haven't tested this at all.

It is fixed now. 

> Yes I do, I'm just saying that it will be completely production ready if you do so. Thus, testing and fixing this instead, looks like wasting time.

No, because it still needed rebased as well, so would have to deal with that.

bews:
> It is fixed now.

That wasn't the single bug needs to be fixed, right? (*whispering: options for actions*) #1047 doesn't have any of this.

YaManicKill:
> That wasn't the single bug needs to be fixed, right? (whispering: options for actions)

What do you mean? How about you explain what you mean rather than just assuming I know what you are talking about? Afaics, looking through the issues you brought up on #1047, either they weren't applicable, they are fixed already, or I didn't agree that they should be implemented, as explained in my comment above.

> #1047 doesn't have any of this.

Sure, but #1047 is also conflicted with master...

YaManicKill:
Also, regarding you getting annoyed about taking your time to test and there still being bugs, I wasn't actually expecting you to test, because you had already said this:

> I'm not updating anymore, so I can't test this.

bews:
> because you had already said this

Yes, but I decided to give it a try. If it was good enough, I would start updating again. I really can't wait this and #1047 get merged, so I was hoping to help.

> Sure, but #1047 is also conflicted with master...

Do you want me to rebase it? I can make a clean production ready PR from #1047 with all my fixes. Without tests although (not sure what is the best way to fix them).

>  What do you mean? How about you explain what you mean rather than just assuming I know what you are talking about? Afaics, looking through the issues you brought up on #1047, either they weren't applicable, they are fixed already

Just searched through all comments here (didn't read them all) and it looks like it was done on purpose. This:
> as a suggestion, would it be possible to disable your condensing if any of join, parts, quits, etc. are disabled?

It really looks like bug if you don't know. Also, it wasn't an issue in #1047, so I surprised to see it here. 

I think the best way to handle this is to make an option to disable condensing, like I suggested before. Otherwise it's not intuitive at all and looks like a bug.

YaManicKill:
> Do you want me to rebase it? I can make a clean production ready PR from #1047 with all my fixes. Without tests although (not sure what is the best way to fix them).

Well, I think I have the fixes of yours that we are going to put into this. Apart from what I am about to talk about (and sending more than 100 messages now, because I'm not 100% sure what the best way to go about this is, I'm thinking just going with sending 100 messages, but not including joins/parts/etc, but having a maximum number even with that, because we don't want to end up sending several thousand).

I'm hoping this will be finished soon, and let's let @metsjeesus rebase his own PR. It might be worth just combining the 2 and only merging #1047. Let's see what @thelounge/maintainers say.

> It really looks like bug if you don't know. Also, it wasn't an issue in #1047, so I surprised to see it here.

Yeah, it's not a great thing, I'm not a fan of it, but I don't think just having a setting for disabling the condensing is going to solve that issue. The issue is that because of how we hide joins/parts with the settings, we will end up with empty condensed messages, which is even worse.

Personally, I'm starting to agree with @christer88 

> When client condense pretty much every join/part/mode etc. I see few reasons for keeping the "Show joins", "Show parts" etc. In my opinion it's better to have "Show joins, parts, nicks and modes changes" 

Think this is probably the best way to go about things.

I'm not a fan with having a setting for the condensing as well, but I could be convinced if others think it is useful.

christer88:
> I'm not a fan with having a setting for the condensing as well, but I could be convinced if others think it is useful.

@YaManicKill Condensing is definitively useful and will make Lounge a lot better. However, I don't think everyone will like this. Might be confusing for quite a few people out there. Not a bad thing to give them the option to just turn it off, if it doesn't require a big chunk of code to maintain.

YaManicKill:
> Might be confusing for quite a few people out there.

I really don't think it would be confusing

> However, I don't think everyone will like this.

Yeah, I guess maybe I'm expecting more people to want the feature than actually will. I'm ok with adding a setting if we get rid of the separate join/part/etc hides. 

Just 1 setting for showing all of them, and 1 for condensing them. Both default to 'on'. I'm ok with that. @thelounge/maintainers ?

src/client.js
```diff
@@ -383,8 +383,14 @@ Client.prototype.more = function(data) {
 		return;
 	}
 	var chan = target.chan;
-	var count = chan.messages.length - (data.count || 0);
-	var messages = chan.messages.slice(Math.max(0, count - 100), count);
+	var index;
+	chan.messages.some((val, i) => {
+		if (`msg-${val.id}` === data.lastId) {
```
xPaw: Remove `msg-` on the client and send an integer. And I think this `more` change can be split into a separate PR, this will make it a bit easier to review.

client/js/render.js
```diff
@@ -19,22 +22,50 @@ module.exports = {
...
+	}
+}
+
 function buildChatMessage(data) {
 	const type = data.msg.type;
 	let target = "#chan-" + data.chan;
 	if (type === "error") {
 		target = "#chan-" + chat.find(".active").data("id");
 	}
+	data.msg.previews = data.msg.previews || [];
```
xPaw: This shouldn't be necessary as `previews` is always defined as `[]` on the server.

client/js/render.js
```diff
@@ -19,22 +22,50 @@ module.exports = {
...
+	if (constants.condensedTypes.indexOf(messageType) !== -1 && chanType !== "lobby") {
+		var condensedTypesClasses = "." + constants.condensedTypes.join(", .");
+		var lastChild = container.children("div.msg").last();
+		var lastDate = (new Date(lastChild.attr("data-time"))).toDateString();
+		var msgDate = (new Date(msg.attr("data-time"))).toDateString();
+		if (lastChild && $(lastChild).hasClass("condensed") && !$(msg).hasClass("message") && lastDate === msgDate) {
+			lastChild.append(msg);
+			condense.updateText(lastChild, [messageType]);
+		} else if (lastChild && $(lastChild).is(condensedTypesClasses) && showAll()) {
+			var condensed = buildChatMessage({msg: {type: "condensed", time: msg.attr("data-time")}, chan: chan});
```
xPaw: This is where you're messing up `previews`.

client/js/lounge.js
```diff
@@ -430,6 +430,14 @@ $(function() {
 		}
 	});
 
+	chat.on("click", ".condensed", function() {
+		$(this).toggleClass("closed");
+	});
+
+	chat.on("click", ".condensed div", function(e) {
```
xPaw: Pretty sure you can just do `e.target !== e.currentTarget` in the event above instead of having this extra event.

client/views/msg_condensed.tpl
```diff
@@ -0,0 +1,5 @@
+<div class="msg {{type}} closed" data-time="{{time}}">
+	<span class="time hide-text">{{tz time}}</span>
```
xPaw: Is there any use of having timestamps on condensed block if it's going to be hidden with CSS anyway?

client/js/render.js
YaManicKill: Oh yeah, I can stick it in here instead, good catch.

src/client.js
YaManicKill: It'll need to go in before this PR if I separate it out.

client/views/msg_condensed.tpl
YaManicKill: The width is messed up if we don't have this. I could figure it out properly, but ... lazy lazy

client/js/constants.js
YaManicKill: Probably.

client/js/lounge.js
YaManicKill: Huh, good point. Will try that out.

client/views/msg_condensed.tpl
xPaw: Won't empty space do the job?

YaManicKill: If we made it non-breaking spaces, possibly, but it would need to be 5 and if the user isn't using a fixed-width font, it won't end up the exact same width as the other lines.

dgw:
> I'm ok with adding a setting if we get rid of the separate join/part/etc hides.

Might I suggest reducing the number of settings instead of condensing them all down to one? Perhaps joins+parts+quits(+kicks)(+modes), and nick-changes. Being able to see when people change names is likely to be important even if one doesn't care to see users entering and leaving the channel. Modes can be lumped in, too, I think, but nick changes seem important enough to separate out.

All that said, I enjoy having control of the individual options, but since I keep them all enabled I don't personally need separate settings for every single item.

metsjeesus:
Just add my 2 cents. This pull is adding the condensing feature. Mainly the HTML and CSS side so it wont break the layout. #1047 will add enchanched grouping mechanism, testing environment for all the weird cases and maybe even translations, if master merges them. Because the big change on client side js, its not worth to fix it while it is not in focus. Once this pull merges to master, ill continue the work.  Soon™

YaManicKill:
> Might I suggest reducing the number of settings instead of condensing them all down to one?

I think I'd rather just have 1 setting, and 1 for condensing.

The number of people that would, say, show joins but not parts, or kicks but not joins, or whatever, are tiny. And the amount of space those settings use is ridiculous.

It would still be really easy for someone to add custom css to hide the type they don't want to show, so I think it makes sense to have all-or-nothing as the setting.

> Modes can be lumped in, too, I think, but nick changes seem important enough to separate out.

I think the problem is if we try and decide how many to make, we'll be here till kingdom come. Everyone will have a different opinion as to which should be separate.

So let's go with just the 2 settings, and anyone who wants more granular can just put in the work on custom css:

* Show joins, parts, kicks, nick and mode changes.
* Condense joins, parts kicks, nick and mode changes

(Happy to change the wording on those, though)

metsjeesus:
My pull will change condensing grouping anyway, so if someone wants to change it, there will be easy ways  to make them on client side js. On/Off button will be enough.

YaManicKill:
Updated the options, and pulled out #1354 as a seperate PR. Once that one is merged, this is good to go, I think.

YaManicKill:
Now that #1354 is merged, this should be good to go. As With #1266, I'm off on holiday for 3 weeks after today, so a nudge would be good.

client/index.html
```diff
@@ -218,38 +218,20 @@ <h1 class="title">Settings</h1>
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
-									<input type="checkbox" name="join">
-									Show joins
+									<input type="checkbox" name="types">
+									Show <abbr title="Joins, parts, kicks, nick changes, and mode changes">status messages</abbr>
```
xPaw: Either this needs to be checked by default, or it has to be `Hide`.

client/index.html
```diff
@@ -218,38 +218,20 @@ <h1 class="title">Settings</h1>
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
-									<input type="checkbox" name="join">
-									Show joins
+									<input type="checkbox" name="types">
```
xPaw: Oh, that name should be `optional`, not `types`.

client/js/lounge.js
xPaw: @YaManicKill did you get to try this?

YaManicKill: I didn't, this is what I get when I have such a long running PR and don't put everything in the to-do list. I'll hopefully have time tonight to try it.

YaManicKill: I don't have time for this before I go. You can either fix it yourself or we can just leave it. Sorry.

client/views/msg_condensed.tpl
```diff
@@ -0,0 +1,7 @@
...
+	<span class="from"><span class="expander"></span></span>
```
astorije: Anyone with strong arguments against having this on the right of the text, to make it consistent with preview toggles? I.e. instead of:

screen shot 2017-07-31 at 00 48 36

We would have:

screen shot 2017-07-31 at 00 48 25

I understand that having no icons on the left can be confusing since this is not a regular message, but it's odd to have it on the left of the column separator, knowing it is clickable. We could (but don't have to) have an icon on the left, if so, not sure which one, but it could be any of:
- http://fontawesome.io/icon/bolt/
- http://fontawesome.io/icon/question/
- http://fontawesome.io/icon/asterisk/
- http://fontawesome.io/icon/info/
- http://fontawesome.io/icon/certificate/
- ...

I honestly don't have a preference myself, I'm not sure we need an icon anyway, so I'm open for suggestions.

astorije: Also, I'll animate the toggle consistently with the preview toggle, not a difficult thing to do anyway.

client/index.html
```diff
@@ -218,38 +218,20 @@ <h1 class="title">Settings</h1>
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
-									<input type="checkbox" name="join">
-									Show joins
+									<input type="checkbox" name="optional">
+									Show <abbr title="Joins, parts, kicks, nick changes, and mode changes">status messages</abbr>
```
astorije: I know there were a lot of discussions around options here, and I think the current situation is really nice: show nothing at all, show everything, or show a condensed/human-friendly version. 👏 👏 on getting there!

That being said, there is one state right now that's possible with the checkboxes and has no effect: whenever "Show status messages" is unchecked, "Condense status messages" has no meaning.

- One easy way to fix this would be to make the latter checkbox disabled when the former one is unchecked.
- Another, more meaningful way (but also a tiny bit more involved) would be to make the latter more visually dependent on the former. Something along the lines of:
  screen shot 2017-07-31 at 01 25 59 screen shot 2017-07-31 at 01 25 38
  Of course, with better wording, presentation, etc.
  Any suggestions welcome :)

astorije: Checking/Unchecking "Condense status messages" checkbox does not apply before refresh.

client/views/msg_condensed.tpl
astorije: No opacity change on hover, and also the whole line is clickable, meaning mostly blank space in most cases.

astorije:
Also, rebased this on latest `master` and fixed conflicts.

client/views/msg_condensed.tpl
bews: >  Anyone with strong arguments against having this on the right of the text, to make it consistent with preview toggles?

I against it, because it will a problem later, in #1047: there are clickable nicknames, so the safest way to expand it is to click on toggle icon. And here is a problem: toggle icon is small enough to miss it. Current way allow to expand condensing by clicking on a huge area, which is a big help:

bews: Same thing here - #1047 partly change this behavior.

client/index.html
xPaw: Sure, do that.

xPaw: I think this could be done in a new PR, as this one has waited too long. But I like the idea.

xPaw: Re-rendering is most likely too much of a hassle to implement.

astorije: Then at the very least, we should refresh the page for this option.

At the moment, when enabling/disabling this option, one needs to realize it did not work, close the app, re-open the app (since there is no refresh button when app is added to homescreen on mobile).

Would you like me to implement that?

astorije: Done and mixed with change mentioned in next comment:

screen shot 2017-08-06 at 00 49 39

astorije: Actually, I did something so that I wouldn't have to repeat that question mark icon, or just put it once but being awkward. Same screenshot as comment above, but putting it here in case GitHub messes up comment order:

screen shot 2017-08-06 at 00 49 39

client/css/style.css
```diff
@@ -1130,12 +1130,12 @@ kbd {
...
-#chat.hide-optional .nick,
-#chat.hide-optional .part,
-#chat.hide-optional .quit,
-#chat.hide-optional .condensed,
+#chat.hide-status-messages .join,
+#chat.hide-status-messages .mode,
+#chat.hide-status-messages .nick,
+#chat.hide-status-messages .part,
+#chat.hide-status-messages .quit,
+#chat.hide-status-messages .condensed,
```
astorije: I merged both options together, and figured this was a more appropriate name than `optional` :man_shrugging:

client/css/style.css
```diff
@@ -1401,12 +1401,11 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
 #settings .opt {
 	display: block;
-	padding: 5px 0 10px 1px;
+	padding: 5px 0 5px 1px;
 }
 
 #settings .opt input {
-	float: left;
-	margin: 4px 10px 0 0;
+	margin-right: 6px;
```
astorije: That float was just unnecessary, just affected the right margin, so this is roughly equivalent.

client/js/options.js
```diff
@@ -9,22 +9,29 @@ const tz = require("./libs/handlebars/tz");
...
-}, JSON.parse(storage.get("settings")));
+};
+const userOptions = JSON.parse(storage.get("settings"));
+const options = {};
+
+for (const key in defaultOptions) {
+	if (userOptions[key]) {
+		options[key] = userOptions[key];
+	}
+}
```
astorije: I noticed that my settings page was broken, and it wasn't code: I had an old options set in local storage, and it was being carried over, then picked up by the UI. This ensures that only the options defined in the defaults can be overridden by what's in local storage.
This will be useful when a given options takes the same name as an old one by accident. Maybe that was my case, not sure, but I figured a proper fix would be nicer than just wipe local storage.

client/views/msg_condensed.tpl
astorije: Relying on blank space to be clickable is a user experience issue, first because there is no discoverability or affordance/feedback (e.g. no hover effect, visual indication, etc.), and also because it's easy to click by accident: users constantly click on blank space as a way to exit the current interaction, such as closing a modal, removing input focus, closing a side navigation menu, etc.

So I tried something, this is how it looks:

screen shot 2017-08-06 at 12 32 14

screen shot 2017-08-06 at 12 32 21

In practice, it actually works really nicely and is very intuitive: you can expand/collapse the entries by clicking either the text or the arrow.
You won't trigger this by clicking on white space by accident (for example, to select something), which has no visible bounding box anyway.

And the day nicks are added to the text will not be an issue: the condensed message will be so much more informative than opening the details view will be something users do much more rarely (i.e. respecting principles than most recurring actions are the easiest to trigger), and when they do need to do that, they can still click on the arrow or any part of the text that is not clickable, i.e. not a colored nickname.

So to me that experiment is very conclusive, it makes much more sense as is and does not affect usability, on the contrary. It also adds a bunch of goodies, like transition on hover, and reuse the styling of the thumbnail toggle for consistency. 🎉

astorije: See comment above for details.

client/css/style.css
```diff
@@ -286,22 +285,6 @@ kbd {
...
-}
-
-#chat .condensed .expander:before {
-	content: "\f0d7"; /* http://fontawesome.io/icon/caret-down/ */
-}
-
-#chat .condensed.closed .expander:before {
-	content: "\f0da"; /* http://fontawesome.io/icon/caret-right/ */
-}
-
```
astorije: I simply merged this with the thumbnail toggle button :)

client/css/style.css
```diff
@@ -1178,19 +1171,14 @@ kbd {
...
+#chat .toggle-button.opened, /* Thumbnail toggle */
+#chat .msg.condensed:not(.closed) .toggle-button { /* Expanded status message toggle */
 	transform: rotate(90deg);
 }
 
-#chat .toggle-button:hover {
-	/* transform and opacity together glitch, so need to use RGBA transition */
-	color: rgba(102, 102, 102, .8); /* #666 x .8 opacity */
-}
-
```
astorije: I removed this because it became impossible to deal with knowing that opacity was applied directly on the button for thumbnail toggles, but on the parent element for condensed toggle. More importantly, the glitch was still present with this hack, and reverting this mostly went back to an un-glitchy state :man_shrugging:.

client/js/renderPreview.js
```diff
@@ -47,7 +47,7 @@ function renderPreview(preview, msg) {
 	container.trigger("keepToBottom");
 }
 
-$("#chat").on("click", ".toggle-button", function() {
+$("#chat").on("click", ".text .toggle-button", function() {
```
astorije: This makes sure that the thumbnail toggle handler does not react to condensed toggle clicks. It is possible because `text` class is now absent from the condensed template.

client/themes/morning.css
```diff
@@ -208,17 +204,9 @@ body {
...
-}
-
-#chat .toggle-button:hover {
-	color: rgba(243, 243, 243, .8);
-}
-
 #chat .toggle-content {
 	background: #242a33;
+	color: #f3f3f3;
 }
```
astorije: All this makes the toggle button the same color than neighboring text. It was silly of me to color it differently, and it looked really bad on the condensed toggle.

client/views/msg_condensed.tpl
```diff
@@ -1,7 +1,7 @@
 <div class="msg {{type}} closed" data-time="{{time}}">
 	<span class="time hide-text">{{tz time}}</span>
-	<span class="from"><span class="expander"></span></span>
+	<span class="from"></span>
 	<span class="content">
-		<span class="text condensed-msg"></span>
+		<span class="condensed-text"></span>
```
astorije: Removed `text` class here to differentiate normal messages from condensed messages. See above for a further explanation.

client/js/lounge.js
astorije: This is not an issue anymore in the last commit I added 😃

client/views/msg_condensed.tpl
bews: > consistency

Another word for breaking good things :disappointed: 

>  transition on hover

Good luck with usernames later.

astorije: > Another word for breaking good things 😞

This is both an untrue and unfair statement. Please refrain yourself from making those, they don't take the discussion forward, and only diminish people's work.

In this case, as outlined in the relevant commit, said consistency was for reusing the CSS for that element instead of duplicating the code (because this PR was started before the other work introducing a caret toggle was done), I honestly don't see how that can be a bad thing.

When it comes to "breaking good things", see my snippet above on usability and user experience. Expecting a user to guess what a blank area that should be inert does when they click on it, as well as the boundaries of the clickable area is not one of those good things.
I am not making a personal attack to @YaManicKill or his tremendous work, obviously, simply answering the comment.

Also, I don't just break things because I like it. I did put quite some UX/usability efforts into this that you might not realize, trying different solutions, at different screen sizes, in different contexts. I suggest you look at the code and run it locally, you'll realize this is a non-issue.
Should changes to this features be made that cause the current usability to be negatively affected, we'll obviously resolve this. In time.

bews: > This is both an untrue and unfair statement. 

No, that's exactly what's happening here.

> said consistency was for reusing the CSS for that element instead of duplicating the code
> how that can be a bad thing

Breaking a good UI for a better code... Are you doing this for users or just for yourself?

> Expecting a user to guess what a blank area that should be inert does when they click on it, as well as the boundaries of the clickable area is not one of those good things.

There's no "guess". It's intuitive: if user miss that symbol for a few times and condensed message still opens as expected - it becomes intuitive and more relaxed way.
Good UI should help user to do things he wants to do, but you break this and go the opposite direction "for consistency and better code". 

> Also, I don't just break things because I like it

C'mon, just admit it :rofl: I have 5 closed PRs because of this.

>  I suggest you look at the code and run it locally, you'll realize this is a non-issue.

I run it since 15.05.2017 and telling you: it's A BIG issue. How long did you run it before changing things?

> that cause the current usability to be negatively affected, we'll obviously resolve this. In time.

Oh, don't worry, that won't be a case. How many users use this PR? When this go to the release most of them won't even know that there was such nice UI, so I don't expect any complains.

client/js/condensed.js
```diff
@@ -33,5 +34,6 @@ function updateText(condensed, addedTypes) {
 			text += obj[messageType] > 1 ? "s" : "";
 		}
 	}
-	condensed.find(".condensed-msg").text(text);
+	condensed.find(".condensed-text")
+		.html(text + templates.msg_condensed_toggle());
```
xPaw: @astorije So the potential issue I see with this, is that unlike previews, condensed text can change and thus moving the button. A case can happen when you want to click the button, and someone leaves the channel at that moment and the button moves as a result of increased text.

astorije: > Unless the entire line is clickable

Yep, entire line, not just the toggle button ;-)

astorije: So bounding box would roughly be:

28763185-2e2aba90-758a-11e7-8933-c7b8a40f9d44

(I quickly edited this with Mac's preview, not set background on actual link, so *picture may differ from actual product* 😅)

client/views/msg_condensed.tpl
astorije: > Are you doing this for users or just for yourself?

You know, I am just about done with you and your insidious comments and personal attacks, ALWAYS (not even talking about the comment where you talk about my self-esteem, who do you think you are?!).

You *clearly* have no idea what UX means, and what are the best practices in that regard. Your opinion is not fact. It is not the undisputed golden rule, it is not what we should blindly take. No, when you open a PR with 3 changes and the team disagrees with 2 of them, we *will not* merge it as is just because you say so.
I am more than happy to discuss any decision I make, and open to genuine and justified arguments. I suggest you start to do something similar.

I will also stop answering to you because I have noticed that you *constantly* quote people out of context to your advantage, and put words in people's mouth (I could find some examples, but I won't).
For the n-th time, I allocate a **tremendous** amount of my free, personal time, almost *every day* that I take off other things like hobbies, family, and sleep. It's a trade-off I am happy to make because mot people are very thankful for the work the entire team puts in, and it's very rewarding to know that this efforts are helping people and making their daily lives easier. But, hear me out, **I do not work for you**. You might have missed the many times I have said this, but I am genuinely thankful for the contributions you have made. It does not mean however that you are entitled to see all your PRs merged with no questions asked. I am not thankful for the tons of hours wasted on pointless arguments, and frankly, being shat upon by you.
I do not have the time, the energy, nor the will to deal with this kind of things.

One last thing: I sincerely hope that you'll realize the damages that you do, apologize for your rudeness, and start acting more respectfully of people and their work. I would be more than happy to restart with a clean slate, help you thrive in the open source world and, overall, make The Lounge an even better IRC client, as we all want. Based on previous exchanges, I do think however that you'll see me as the unthankful bad guy, which is a shame. My door will always be open for respectful and open-minded discussions.

*To the uninformed reader: this is not me taking out on a commenter out of the blue. There have been many of those interactions, to the point where frustration becomes too big to ignore. All team members have spoken up and asked for a better conduct before. I personally believe very strongly in open source and its values, and I don't think things should go that way. There is always room for healthy discussions, even when disagreeing, and I hope we can get back to that soon enough.*

bews: Ha ha and again HA! You really made my day with this comment, thank you.

You're asking for "healthy discussions" but it's you who wrote completely unrelated comment of 2196 characters just to attack me.
I pointed a problem, explained it and expected a healthy discussion you mentioned, but this is what I got. Again.

Also, you ignored a problem pointed out by @xPaw here https://github.com/thelounge/lounge/pull/759#pullrequestreview-54848305 which I overlooked. 
It's the same thing I was talking about, but on a larger scale. With #1047 you absolutely can not click on the text (or even on the empty space left) because
1) message can change while you click
2) it contain clickable usernames

Thus this may result in unexpected behaviour: instead of expanding block, a whois window can be opened.

This is one more proof that your change is really bad UX-wise and I doubt you can say anything against it. 
So feel free to do the same things you always do: ignore it/say that I know nothing about UX/whine about wasted time/ask for a better conduct and healthy discussion.

But it's not that bad, there's something I agree with!
> I do not have the time, the energy, nor the will to deal with this kind of things.

I have the same thought all the time writing previous comments in hope that you stop ignoring problems I pointed. 

ANYWAY, 
I always open for a constructive discussion if you up for it. Also, it's not like my vision of the problem is absolutely right and I really can change it, if you provide arguments instead of forcing your way.
I do want to make Lounge a better client, that's why I'm here. If you really want this too, then listen to other opinions, admit problems and try to find the way to fix them. 

P.S.
> I do not work for you

What is this? I can't understand. Is it supposed to be the answer of the question you quoted? But I'm one of the users. Are you doing this not for every user, but for chosen ones? If so, what the criteria?

xPaw: >  With #1047 you absolutely can not click on the text (or even on the empty space left) because

I have seen you point out multiple times issues that are only related to that specific PR, and not this one. While the idea of that PR is great, it's really some subpar code at the moment. So to get that one merged, this one has to be merged first, and work needs to be done on the other one.

@astorije's changes do make sense, and they simplify things, and they absolutely do not make it harder for #1047 to work. It should be just as simple to move the arrow to beginning of the line, instead of the end, to fix the jumping button problem. I do think that having clickable arrow in nickname row as it was before is not a very good solution, as it may be confusing and not that obvious that it does something unique compared to other icons there.

xPaw:
@astorije the default state for the setting appears to be nothing.

client/views/msg_condensed.tpl
bews: >  It should be just as simple to move the arrow to beginning of the line, instead of the end, to fix the jumping button problem.

Change it now to revert later. Makes no sense to me.

> I do think that having clickable arrow in nickname row as it was before is not a very good solution, as it may be confusing and not that obvious that it does something unique compared to other icons there.

Yeah, but `cursor: pointer` is savior here. With the whole line clickable it's hard not to notice something special about that line :smirk_cat: 

It may be the good idea to completely remove that icon and add something big and obvious here: 
28764196-7e9ae1d4-75cc-11e7-8940-12be4be232f1
because it's the safest spot for click action.

astorije:
Ah, missed that @xPaw, it shouldn't as the default is supposed to be "shown" (i.e. expanded), but I'll look into it.
While I'm here, should the default be expanded or collapsed? I'd actually prefer collapsed, and I think that's what @YaManicKill did, so my commit might have canceled that. If that's okay by you, I'll put back to "collapsed" as default.
Also, if collapsed by default, should I switch the 2 options in the UI? I like the idea of order being "everything"/"less things"/"nothing", but I also think that defaulting to something else than first option is meh. I know I'm grasping at straws though 😅

christer88:
> While I'm here, should the default be expanded or collapsed? I'd actually prefer collapsed, and I think that's what @YaManicKill did, so my commit might have canceled that. If that's okay by you, I'll put back to "collapsed" as default.

I think default should be collapsed. This will highlight the future (not everyone will click on settings and change this because they don't know of it), as well as this is probably what most people would like anyway. 

> Also, if collapsed by default, should I switch the 2 options in the UI? I like the idea of order being "everything"/"less things"/"nothing", but I also think that defaulting to something else than first option is meh. I know I'm grasping at straws though 😅

I agree about having "everything"/"less things"/"nothing". What I don't agree about is the lack of informational text. Can we put "joins/parts/quits/nick changes" somewhere to make it very easy to see what the settings are for? Please mind I have not seen what the tooltip on ? says.

astorije:
Hey @christer88, thanks for your feedback! 😊

> I think default should be collapsed.

Cool, your explanation makes sense!

astorije:
Hey @christer88, thanks for your feedback! 😊

> I think default should be collapsed.

Cool, your explanation makes sense!

> I agree about having "everything"/"less things"/"nothing".

👍

> What I don't agree about is the lack of informational text. Can we put "joins/parts/quits/nick changes" somewhere to make it very easy to see what the settings are for? Please mind I have not seen what the tooltip on ? says.

It turns out, this is exactly what this ❔ tooltip is for :)

screen shot 2017-08-06 at 00 49 39

In practice, this is exactly the text that @YaManicKill put in a `title` attribute on hover, except that this is on an instant tooltip, both "hoverable" and clickable so that mobiles can display it. Same thing, just a bit more reachable by the user!

astorije:
@xPaw, fixed that issue with default option, as well as another one.
Also (/Cc @christer88), I (re-)made `condensed` the default.

Finally, did the last thing I wanted to do, i.e. made the strings friendlier and closer to what we have on expanded view:

screen shot 2017-08-13 at 15 26 13
screen shot 2017-08-13 at 15 26 37

It's not the cleanest code for that, but it will be amended by #1271 and/or #1047 anyway. Also, I did clean up this file a little bit to balance the karma 😅 

Next step for me is to do a full review again, will do that shortly.

xPaw:
Whatever, let's merge this and do work on top of master. This has been stalled for too long.

YaManicKill:
Wait, this is actually finally merged? Holy crap. I'm so happy. No more rebasing.

Issue #760, opened 2016-11-24
Date-markers showing diffrent dates for link expander (on znc playback)


stepie22:
As you can see here, the link expander and the messages have diffrent dates. I believe this is because I am behind znc with playback buffer enabled, which sends the date with the messages, and the server only fetched the links now, hence why they are today's date. (This happened after a restart after updating to master)

A check should probably be made to ignore the prefetch's date

![](http://image.prntscr.com/image/5a9e7ad4d156471ba9d3e319a11fd8fa.png)

YaManicKill:
Good catch, thanks. 

Just to check, the messages (not the link expanders) are in the correct date, yeah?

stepie22:
Yes, thanks to the `znc.in/server-time-iso` capability thelounge supports and the znc playback buffer

xPaw:
I think the bug is that expander button doesn't use server-time as it should.

PR #765, opened 2016-11-25
Date separator fixes

### LINKED ISSUE #763: Date-separator not being removed when it should sometimes
When a 'new messages' marker is at the top of the screen, followed by a date-marker, and the 'show more' button is pressed, the date marker should be removed and inserted at the correct place. Instead it is not, because it only checks the first element after the show-more button if it is a date-marker.

stepie22:
Fixes #763 and #760; I decided to do this in one pr since both are so small fixes. 

For #763 an extra check was added for the new-messages marker and skips it

For #760 I added the reported time, if that was not available, the current time

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -420,9 +420,12 @@ $(function() {
 
 		// Remove the date-change marker we put at the top, because it may
 		// not actually be a date change now
-		var firstChild = $(chan).children().eq(0);
-		if (firstChild.attr("class") === "date-marker") {
-			firstChild.remove();
+		var children = $(chan).children();
+		if (children.eq(0).attr("class") === "date-marker") { // Check top most child
+			children.eq(0).remove();
+		} else if (children.eq(0).attr("class") === "unread-marker" && children.eq(1).attr("class") === "date-marker") {
+			// Otherwise the date-marker would get 'stuck' because of the new-massages marker
+			children.eq(1).remove();
 		}
 
 		// get the scrollable wrapper around messages
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -424,7 +424,7 @@ $(function() {
 		if (children.eq(0).attr("class") === "date-marker") { // Check top most child
 			children.eq(0).remove();
 		} else if (children.eq(0).attr("class") === "unread-marker" && children.eq(1).attr("class") === "date-marker") {
-			// Otherwise the date-marker would get 'stuck' because of the new-massages marker
+			// Otherwise the date-marker would get 'stuck' because of the new-message marker
 			children.eq(1).remove();
 		}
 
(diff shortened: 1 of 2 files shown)


src/plugins/irc-events/link.js
```diff
@@ -32,6 +32,7 @@ module.exports = function(irc, network) {
 		var msg = new Msg({
 			self: data.nick === irc.user.nick,
 			type: Msg.Type.TOGGLE,
+			time: data.time || new Date().getTime(),
```
xPaw: Msg handles the fallback

stepie22: ah, will remove that part then

client/js/lounge.js
```diff
@@ -420,9 +420,12 @@ $(function() {
...
 		// Remove the date-change marker we put at the top, because it may
 		// not actually be a date change now
-		var firstChild = $(chan).children().eq(0);
-		if (firstChild.attr("class") === "date-marker") {
-			firstChild.remove();
+		var children = $(chan).children();
+		if (children.eq(0).attr("class") === "date-marker") { // Check top most child
+			children.eq(0).remove();
+		} else if (children.eq(0).attr("class") === "unread-marker" && children.eq(1).attr("class") === "date-marker") {
+			// Otherwise the date-marker would get 'stuck' because of the new-massages marker
```
MaxLeiter: slight nit-pick: `new-massages` -> `new-message`

MaxLeiter:
Am just going to mention this here: once the client can access npm packages, we can use moments .calender() to create fuzzy time stamps (i.e. "Yesterday")

client/js/lounge.js
```diff
@@ -420,9 +420,12 @@ $(function() {
 
 		// Remove the date-change marker we put at the top, because it may
 		// not actually be a date change now
-		var firstChild = $(chan).children().eq(0);
-		if (firstChild.attr("class") === "date-marker") {
-			firstChild.remove();
+		var children = $(chan).children();
+		if (children.eq(0).attr("class") === "date-marker") { // Check top most child
```
YaManicKill: Ah wait, can you use jquery's hasClass instead? Because this will only give oyu 1 of the classes, which means that it could create bugs in the future if the date marker ends up having more than 1 class on it.

https://api.jquery.com/hasclass/

client/js/lounge.js
```diff
@@ -420,9 +420,12 @@ $(function() {
...
+			children.eq(0).remove();
+		} else if (children.eq(0).attr("class") === "unread-marker" && children.eq(1).attr("class") === "date-marker") {
```
YaManicKill: Same here

stepie22:
Err. @astorije didn't this need to be squashed? I am away for a bit so I can't do it, but pretty sure there is an option for you guys to do so. Referring to the last two commits

astorije:
Argh, I suck, I meant for it then forgot. Too late now, sorry git history!

PR #775, opened 2016-12-03
Auto away when no clients are connected

xPaw:
2 issues:
- [x] Setting away message to empty should disable auto-away (and disable by default?)
- [x] Should keep in mind away message set by using /away command. However this is not blocking issue for this PR.

Auto away is useful for people that use lounge along with Znc as it has checks for away clients.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -65,6 +65,7 @@ function Client(manager, name, config) {
 		config = {};
 	}
 	_.merge(this, {
+		awayMessage: "",
 		lastActiveChannel: -1,
 		attachedClients: {},
 		config: config,
@@ -485,6 +486,16 @@ Client.prototype.clientAttach = function(socketId) {
 
 	client.attachedClients[socketId] = client.lastActiveChannel;
 
+	if (client.awayMessage && _.size(client.attachedClients) === 0) {
+		client.networks.forEach(function(network) {
+			// Only remove away on client attachment if
+			// there is no away message on this network
+			if (!network.awayMessage) {
+				network.irc.raw("AWAY");
+			}
+		});
+	}
+
 	// Update old networks to store ip and hostmask
 	client.networks.forEach(network => {
 		if (!network.ip) {
@@ -508,7 +519,19 @@ Client.prototype.clientAttach = function(socketId) {
 };
 
 Client.prototype.clientDetach = function(socketId) {
+	const client = this;
+
 	delete this.attachedClients[socketId];
+
+	if (client.awayMessage && _.size(client.attachedClients) === 0) {
+		client.networks.forEach(function(network) {
+			// Only set away on client deattachment if
+			// there is no away message on this network
+			if (!network.awayMessage) {
+				network.irc.raw("AWAY", client.awayMessage);
+			}
+		});
+	}
 };
 
 Client.prototype.save = _.debounce(function SaveClient() {
@@ -518,6 +541,7 @@ Client.prototype.save = _.debounce(function SaveClient() {
 
 	const client = this;
 	let json = {};
+	json.awayMessage = client.awayMessage;
 	json.networks = this.networks.map(n => n.export());
 	client.manager.updateUser(client.name, json);
 }, 1000, {maxWait: 10000});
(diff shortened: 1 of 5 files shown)


src/client.js
```diff
@@ -457,11 +458,23 @@ Client.prototype.quit = function() {
...
+
 	this.attachedClients[socketId] = this.lastActiveChannel;
 };
 
 Client.prototype.clientDetach = function(socketId) {
 	delete this.attachedClients[socketId];
+
+	if (_.size(this.attachedClients) === 0) {
+		this.networks.forEach(function(network) {
+			network.irc.raw("AWAY", this.awayMessage);
```
xPaw: `this` should be client from upper scope.

astorije:
> Setting away message to empty should disable auto-away (and disable by default?)

What do you mean by this?

Shouldn't this come with an entry in the client settings? We probably don't want to force everyone to be away when no client is connected :)

src/client.js
```diff
@@ -63,6 +63,7 @@ function Client(manager, name, config) {
 		config = {};
 	}
 	_.merge(this, {
+		awayMessage: "no users in the lounge",
```
astorije: Now that this branch is leaving the status of patch, we should probably have a better default. Actually, empty by default since it will be disabled by default.

src/client.js
```diff
@@ -457,11 +458,23 @@ Client.prototype.quit = function() {
 };
 
 Client.prototype.clientAttach = function(socketId) {
+	if (_.size(this.attachedClients) === 0) {
+		this.networks.forEach(function(network) {
+			network.irc.raw("AWAY");
```
astorije: @xPaw, I'm seeing this and wondering: could this PR be improved once #745 is merged?

src/client.js
astorije: As I understand [MDN's article on arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_binding_of_this), this would come for free if `forEach` was using one of them. I haven't used them much, so I might be wrong.

src/plugins/irc-events/connection.js
```diff
@@ -20,6 +21,10 @@ module.exports = function(irc, network) {
 			}), true);
 		}
 
+		if (_.size(client.attachedClients) === 0) {
```
astorije: Would it make sense to use [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) for `client.attachedClients`?
I noticed we use it like this in a couple places, and we could use `client.attachedClients.size` directly (actually, this is one of the arguments of `Object` vs. `Map` used in the MDN article!).

src/client.js
xPaw: Not really related directly.

xPaw:
@maxpoulin64 See the second todo in main post. I wanted to incorporate your `/away` message into it.

xPaw:
Updated the PR to also account for network `/away` messages, which aren't overriden by default auto away.

astorije:
Just curious, if we merge this, `/away` state will be automatically disabled as soon as user is connected to The Lounge?

xPaw:
@astorije Nope, there's code to prevent that.

MuffinMedic:
Any updates on getting this merged into master?

YaManicKill:
@MuffinMedic We are currently doing final checks on version 2.2.0, and after that we will start looking at this again.

astorije:
@xPaw, can we have a front-end checkbox for this? We already do server-synced stuff with custom highlights so we should be able to do that for this one too.
And I know https://github.com/thelounge/lounge/pull/275 was addressing that sort of things but I don't think this is going to be worked on any time soon, while I would love to see your PR working end-to-end (i.e. not having to edit the config file on the server to enable this!).

astorije:
FYI, rebased this on master and got rid of the merge conflict.

xPaw:
> We already do server-synced stuff with custom highlights so we should be able to do that for this one too.

We'd don't though?

astorije:
> We'd don't though?

Well I'll be damned, I thought so but I see this is client-side only! I guess this is good to go then, but I wish there was a way to do settings better here. Quite the undertaking, but that would be such a great enhancement to solve globally.

xPaw:
Settings branch exists, but it needs major work.

xPaw:
🎉

PR #777, opened 2016-12-06
Add a README badge pointing to vulnerability reports

astorije:
Since this is an end-user project, used in production on actual servers, security is a main concern. While this does not improve anything per se, it will help keeping track of new vulnerabilities and force us taking action when there is one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 [![AppVeyor Build Status](https://img.shields.io/appveyor/ci/astorije/lounge/master.svg?label=windows+build)](https://ci.appveyor.com/project/astorije/lounge/branch/master)
 [![Dependencies Status](https://img.shields.io/david/thelounge/lounge.svg)](https://david-dm.org/thelounge/lounge)
 [![Developer Dependencies Status](https://img.shields.io/david/dev/thelounge/lounge.svg)](https://david-dm.org/thelounge/lounge?type=dev)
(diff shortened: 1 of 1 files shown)


dgw:
Lucky 777! 🎉

MaxLeiter:
I'm personally against any more badges but not a big deal

xPaw:
David-dm is supposed to show vulnerable packages too?

astorije:
> David-dm is supposed to show vulnerable packages too?

Supposed to, yes, but I feel like a tool dedicated to vulnerabilities is more appropriate, like this one has PR integration (not enabled here obviously), etc.

I want something to keep an eye on our security, but this tool doesn't have to be the one. I'm happy to not use it, or use it until we find better, etc.
I do agree with you though that Snyk only seems to track dependency vulnerability, which may duplicate David's stuff, but I don't know how reliable or efficient David DM is for security vulnerabilities.
If we found however a tool that also analyses the codebase (i.e. has a real knowledge of Node's API and not just compares metadata between npm and a DB of known vulnerability) for stuff like filesystem modes, HTTP headers set up, detection of passwords sent in clear, DoS vectors, enforcement of best practices, etc.
Googling around very quickly, we might be happier with (more official) https://nodesecurity.io/ and going through stuff like https://expressjs.com/en/advanced/best-practice-security.html.

Feel free to close it if you don't like it, add thoughts on what to do instead, merge if it's any valuable in the meantime, etc. :)

YaManicKill:
I'm happy either way, more badges isn't exactly the end of the world, and there is a reason for this, I guess. But yeah, fix that hash issue first :-P

astorije:
I'm OK to close this for now as it doesn't seem like Snyk adds more than what David DM can display here.
However, remind me when you notice one of our deps has a vulnerability and I'll compare David and Snyk reports.

And I'll also be on the lookout for better tools, it would be really good to have one that actually looks at our code. If someone knows something, they are more than welcome to suggest :-)

PR #779, opened 2016-12-07
Make user autoload more transparent in the app

astorije:
This is really possible because @xPaw provided a really nice and natural way to watch user config files in #751, so there is now no need to be cheap about it (it used to be run every second, possibly why it could be disabled via settings?).

Similarly, we don't provide a setting for reconnecting to an IRC server, for letting users logged in after a server restart, etc. all of which have been added since the fork.

I also slightly improved the console logs:

Before | After
--- | ---
screen shot 2016-12-07 at 00 59 49 | screen shot 2016-12-07 at 01 02 10

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -26,11 +26,26 @@ ClientManager.prototype.findClient = function(name, token) {
 	return false;
 };
 
-ClientManager.prototype.loadUsers = function() {
-	var users = this.getUsers();
-	for (var i in users) {
-		this.loadUser(users[i]);
-	}
+ClientManager.prototype.autoloadUsers = function() {
+	this.getUsers().forEach(name => this.loadUser(name));
+
+	fs.watch(Helper.USERS_PATH, _.debounce(() => {
+		const loaded = this.clients.map(c => c.name);
+		const updatedUsers = this.getUsers();
+
+		// New users created since last time users were loaded
+		_.difference(updatedUsers, loaded).forEach(name => this.loadUser(name));
+
+		// Existing users removed since last time users were loaded
+		_.difference(loaded, updatedUsers).forEach(name => {
+			const client = _.find(this.clients, {name: name});
+			if (client) {
+				client.quit();
+				this.clients = _.without(this.clients, client);
+				log.info("User '" + name + "' disconnected");
+			}
+		});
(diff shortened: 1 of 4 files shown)


xPaw:
Autoloader bees to handle file changes too, as we have a CLI command to change the password. Also it needs to ignore file changes caused by the server itself (user saves)

astorije:
> Autoloader bees to handle file changes too, as we have a CLI command to change the password. Also it needs to ignore file changes caused by the server itself (user saves)

I totally agree with you, but this is beyond the scope of this PR, IMO. Plus, these comments already apply on `master`, so I don't believe my changes are making this required.
I don't have much time to focus on this right now, so I'd rather finalize what's necessary to get this merged, and leave the opportunity to someone else to open a PR for this. Sounds good?

xPaw:
Yeah this is fine. Maybe we need to add a console warning saying autoload option has been removed and it's automatic now. Might only print this message it autoload is false in the server config.

astorije:
> Yeah this is fine. Maybe we need to add a console warning saying autoload option has been removed and it's automatic now. Might only print this message it autoload is false in the server config.

Good idead! I'll add that, and we can remove it in a couple versions.

astorije:
@xPaw, done!

> Might only print this message it autoload is false in the server config.

I've actually decided to show the warning whatever the value of the `autoload` option, to invite users to clean up their config file anyway, and a wording that makes it clear it's always enabled now. Looks good to you?

src/clientManager.js
```diff
@@ -26,11 +27,24 @@ ClientManager.prototype.findClient = function(name, token) {
...
-	}
+ClientManager.prototype.autoloadUsers = function() {
+	this.getUsers().forEach(name => this.loadUser(name));
+
+	fs.watch(Helper.USERS_PATH, _.debounce(() => {
+		const loaded = this.clients.map(c => c.name);
+		const added = _.difference(this.getUsers(), loaded);
+		added.forEach(name => this.loadUser(name));
+
+		const removed = _.difference(loaded, this.getUsers());
```
xPaw: Two `getUsers` calls, cache it.

astorije: Good call, done.

PR #782, opened 2016-12-09
Use correct channel when pushing link prefetch messages

### LINKED ISSUE #781: Link previews in PMs get displayed in the PM channel with yourself (and only there)
Image of the PM channel with another user: 

Image of the PM channel with myself:

running latest lounge on npm (2.1) and chrome 55 on win10 64bit.

xPaw:
Fixes #781 by using the same code flow as the channel messages themselves, instead of assuming the wrong target.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -24,7 +24,6 @@ var events = [
 	"mode",
 	"motd",
 	"message",
-	"link",
 	"names",
 	"nick",
 	"part",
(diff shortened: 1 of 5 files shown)


MaxLeiter:
Prefetching is still partially broken:

image

Only the first and last link embed/prefetch, the rest do nothing (the toggle button doesn't work)

xPaw:
Sounds like an unrelated bug, but I'll take a look.

xPaw:
> I'm thinking, should we move link.js out of irc-events to avoid confusion in the future?

Most likely when we refactor the link fetcher, yeah.

astorije:
Great job!

PR #783, opened 2016-12-10
Option to log raw IRC traffic

xPaw:


defaults/config.js
```diff
@@ -368,4 +368,11 @@ module.exports = {
 	// @default  false
 	//
 	debug: false,
+
+	// Enables logging raw IRC messages into each server window.
+	//
+	// @type     boolean
+	// @default  false
+	//
+	debugRaw: false,
```
astorije: I don't like having more than one debug option. If this is specifically for debug, we should reuse `debug` (and change its description accordingly). If they are indeed unrelated, then this one should not be called `debugXXX`. Makes sense?

xPaw: For example, I have debug logs enabled to debug connection issues on the server (connection stuff being logged to server console), and in that case I wouldn't want my client to get all the raw traffic.

xPaw: We could make `debug` an object and add options inside of it.

astorije: I hear you and that's a good use case, but current configuration options seem a bit off, and an object might be overkill? Could do an array of strings, that saves a few characters but same stuff in the end...

I don't have anything other than that to suggest, so your idea is probably best:

- `debug: false` means no debug at all
- `debug: true` means all kind of debugs enabled
- `debug: { connection: true, raw: true, ... }` or `debug: ["connection", "raw", ...]` to enable some of the debugging options. To only enable `raw`, `debug: { connection: false, raw: true }` or `debug: { raw: true }` or `debug: ["raw"]`.

That actually seems like a good idea to offer "flags" like that: by default, in prod, most people will have debug options to `false`, but things can get more complex if necessary.
I prefer the array way as it's simpler in the config and reads more like a list of flags.

Then you can have a helper `Helper.config.hasDebug("connection")` that returns `true` if `config.debug === true || (Array.isArray(config.debug) && config.debug.includes("connection"))` or something like that.

`debug: true` doesn't have to be a thing if you want options to be always passed in explicitly, but it's a nice opposite to `debug: false` === all debug flags disabled.

xPaw: I generally don't like array like configs as they're not very explicit.

astorije: You mean you prefer an object instead of an array?
`{ connection: true, raw: true }` doesn't seem any more explicit than `["connection", "raw"]` to me, so I must not be following you correctly.

PolarizedIons:
Doesn't display sent/outgoing messages, only incoming; at least, for me

astorije:
> irc-framework 2.5.1 also breaks this

Do you have a stack trace or error message?

PolarizedIons:
@astorije it only prints out `[object Object]`

xPaw:
Yeah 2.5.1 changed the event. It also emits all raw events in debug event...

defaults/config.js
xPaw: Object is more explicit because you disable options without removing them. As with the list, you will need to find the options elsewhere (e.g. comment above).

DanielOaks:
Nice. I find turning on raw-traffic every now and then really useful while devving/debugging stuff (though I'm probably weird in that regard), so having something like this would be really cool. The ability to enable it on-the-fly would be cool as well, at least for private instances.

astorije:
@xPaw, ping re: debug object.

xPaw:
@astorije waiting for irc-fw update.

YaManicKill:
If this is the last thing for 2.2.0, I say we push it to 2.3.0 instead of waiting for the irc-fw update. But there are a few other things 2.2.0 is waiting for, so we'll see what happens.

xPaw:
@YaManicKill it was updated already, greenkeeper bot picked it up.

YaManicKill:
Alrighty, then ignore me :-P I guess that's waiting in #826 ?

defaults/config.js
```diff
@@ -361,11 +361,24 @@ module.exports = {
...
+		//
+		ircFramework: false,
+
+		// Enables logging raw IRC messages into each server window.
+		//
+		// @type     boolean
+		// @default  false
+		//
+		raw: false,
+	},
```
astorije: See https://github.com/thelounge/lounge/pull/783#discussion_r91851434, I still think that it should be possible to set `debug` to `false` and that it should be the default. `debug` mode is an exceptional situation and most users will set it to a global `false`, which essentially disables it for all sub-modes.
`true` should enable all debug options.

xPaw: I think my comment (the same you linked) gives a good reason why enabling all available debugging options is not always desirable. Logging raw traffic is only needed in very specific situations.

astorije: Yeah, I simply suggest `true` as a mirror value for `false`, but without being 100% convinced. However, I am for `false` :)

src/helper.js
```diff
@@ -67,6 +67,12 @@ function setHome(homePath) {
 		var userConfig = require(this.CONFIG_PATH);
 		this.config = _.extend(this.config, userConfig);
 	}
+
+	// TODO: Remove in future release
+	if (this.config.debug === true) {
+		log.warn("debug option is now an object, see defaults file for more information.");
+		this.config.debug = {ircFramework: true};
+	}
```
astorije: If we continue to do this, maybe we should do it in the `Helper` instead, so all warnings show up together (well, they probably do right now, but at least they'd be together in the code). No big deal here though.

Issue #421, opened 2016-06-20
Using The Lounge on a flaky network leads to a very schizophrenic front-end


dgw:
The client is not particularly good at recovering from network problems, it seems. Even in 2.0.0-pre.3 any timeout while the client is open leads to a full-page reload.

Unfortunately my LAN is very, very flaky tonight and can't keep connections open for more than about 30 seconds. I was hoping The Lounge would be more stable than a desktop client, but neither allows me to reliably send and receive IRC messages; lines both to and from the server are randomly dropped without any warning in both.

I don't suppose The Lounge's architecture would support falling back to some kind of polling if the network is so flaky that persistent connections simply don't work?

YaManicKill:
> I don't suppose The Lounge's architecture would support falling back to some kind of polling if the network is so flaky that persistent connections simply don't work?

Socket.io already does this.

The real problem is that when we lose connection, we tell the page to reload. Ideally, we should not do that, and let socket.io reconnect. The problem with just switching that one (we've tried) is that the server never expects this to happen, so we get a bunch of initial connection information and we send all the channels and their messages again, and it becomes a mess.

dgw:
I see. So the client could do it, in theory, but it would require server-side changes (I'm simplifying a lot).

Could making those changes to the server be a goal for 2.1? 3.0? There's already a lot slated for the 2.0.0 release—and it's been 3 pre-releases already anyway—but improving the reconnection support seems like a good goal now that it's at least implemented for both client<->server and server<->network.

YaManicKill:
> Could making those changes to the server be a goal for 2.1? 3.0?

astorije:
So, if someone ever wants to tackle with this one (which would be really awesome) and wants some history/archives, @unconed gave a try at this in #267.
Feel free to stop by the channel (`#thelounge`) if you need help with this.

kode54:
Absolutely useless on my Android tablet, as it times out by the time it finally finishes Authorizing after a login. It takes about 30 seconds to Authorize, then briefly flashes the last channel viewed buffer, then the names list appears empty, then it immediately flashes back to Authorizing... repeat until I grow bored of watching my tablet cycle itself to death.

xPaw:
Will close this issue as #1471 most likely fixes this, and there's not any specific problem we can tackle besides that.

PR #784, opened 2016-12-10
Stop refreshing the page on every socket.io error

xPaw:
Partial fix for #421 to allow users to read messages and whatnot after the initial page load. From my own experience this is much more desired than instantly reloading the page and hitting a network error page instead. This PR does not introduce reconnections as that would require a lot more changes.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -58,6 +58,7 @@
 				<div id="chat-container" class="window">
 					<div id="chat"></div>
 					<form id="form" method="post" action="">
+						<div id="connection-error"><span>Client connection lost. Click to reconnect.</span></div>
 						<div class="input">
 							<span id="nick">
 								<span id="nick-value" spellcheck="false"></span><!-- Comments here remove spaces between elements
--- a/client/index.html
+++ b/client/index.html
@@ -57,8 +57,8 @@
 				</div>
 				<div id="chat-container" class="window">
 					<div id="chat"></div>
+					<button id="connection-error" class="btn btn-reconnect">Client connection lost &mdash; Click here to reconnect</button>
 					<form id="form" method="post" action="">
-						<div id="connection-error"><span>Client connection lost. Click to reconnect.</span></div>
 						<div class="input">
 							<span id="nick">
 								<span id="nick-value" spellcheck="false"></span><!-- Comments here remove spaces between elements
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1,10 +1,12 @@
 "use strict";
 
 $(function() {
-	$("#loading-page-message").text("Connecting…");
-
 	var path = window.location.pathname + "socket.io/";
-	var socket = io({path: path});
+	var socket = io({
+		path: path,
+		autoConnect: false,
+		reconnection: false
+	});
 	var commands = [
 		"/close",
 		"/connect",
@@ -74,14 +76,29 @@ $(function() {
 		}
 	);
 
-	socket.on("error", function(e) {
-		console.log(e);
+	[
+		"connect_error",
+		"connect_failed",
+		"disconnect",
+		"error",
+	].forEach(function(e) {
+		socket.on(e, function(data) {
+			$("#loading-page-message").text("Connection failed: " + data);
+			$("#connection-error").addClass("shown").one("click", function() {
+				window.onbeforeunload = null;
+				window.location.reload();
+			});
+
+			console.error(data);
+		});
 	});
 
-	$.each(["connect_error", "disconnect"], function(i, e) {
-		socket.on(e, function() {
-			refresh();
-		});
+	socket.on("connecting", function() {
+		$("#loading-page-message").text("Connecting…");
+	});
+
+	socket.on("connect", function() {
+		$("#loading-page-message").text("Finalizing connection…");
 	});
 
 	socket.on("authorized", function() {
@@ -1343,11 +1360,6 @@ $(function() {
 		}
 	}
 
-	function refresh() {
-		window.onbeforeunload = null;
-		location.reload();
-	}
-
 	function sortable() {
 		sidebar.find(".networks").sortable({
 			axis: "y",
@@ -1491,4 +1503,7 @@ $(function() {
 			}
 		}
 	);
+
+	// Only start opening socket.io connection after all events have been registered
+	socket.open();
 });
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -89,6 +89,18 @@ $(function() {
 				window.location.reload();
 			});
 
+			// Disables sending a message by pressing Enter. `off` is necessary to
+			// cancel `inputhistory`, which overrides hitting Enter. `on` is then
+			// necessary to avoid creating new lines when hitting Enter without Shift.
+			// This is fairly hacky but this solution is not permanent.
+			$("#input").off("keydown").on("keydown", function(event) {
+				if (event.which === 13 && !event.shiftKey) {
+					event.preventDefault();
+				}
+			});
+			// Hides the "Send Message" button
+			$("#submit").remove();
+
 			console.error(data);
 		});
 	});
--- a/client/themes/zenburn.css
+++ b/client/themes/zenburn.css
@@ -253,3 +253,11 @@ body {
 #chat .toggle-content .body {
 	color: #d2d39b;
 }
+
+.btn-reconnect {
+	background: #e74c3c;
+	color: #fff;
+	border: 0;
+	border-radius: 0;
+	margin: 0;
+}
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1302,6 +1302,7 @@ button {
 	-webkit-flex: 0 0 auto;
 	flex: 0 0 auto;
 	padding: 5px;
+	position: relative;
 }
 
 #windows #form .input {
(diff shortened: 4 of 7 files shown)


xPaw:
@YaManicKill done

xPaw:
> we vertically center that text

Is there a legit easy way I can accomplish this here?

> maybe just put a link on Refresh for convenience?

Ideally this refresh bollocks would go away in the future with proper reconnections.

client/index.html
```diff
@@ -58,6 +58,7 @@ <h1 class="title">The Lounge is loading…</h1>
 				<div id="chat-container" class="window">
 					<div id="chat"></div>
 					<form id="form" method="post" action="">
+						<div id="connection-error"><span>Connection failed: <span id="connection-error-reason">??</span>. Refresh the page to reconnect.</span></div>
```
astorije: Outer `<span>` is an accident here?

I'd definitely replace second sentence with `Click here to reconnect` and make it a link or a button (maybe even prepend this with http://fontawesome.io/icon/refresh/ ?).
Reason why we have to make it clickable is that, before your PR, app would refresh by itself when connection is back, or something like that. With your PR, app in standalone mode on mobile cannot be refreshed with pull-to-refresh (because of #618). So that means once connection is lost, user now has to fully close the app and open it again. Makes sense?

Also, I don't think there is any need for specific message. `Connection lost. Click here to reconnect` is enough, and most widespread messaging clients will display the same. Plus if for some reason Socket.io spits some weird looking message, that would make it even more confusing.
I just tried disabling wifi on my phone, and I got `Connection failed: transport close`. As a user, that doesn't tell me anything, but `Connection lost` seems more appropriate and generic enough for all cases.
Plus, you are logging to console anyway, if intent is to debug.

client/js/lounge.js
```diff
@@ -72,14 +74,26 @@ $(function() {
...
-	socket.on("error", function(e) {
-		console.log(e);
+	[
+		"connect_error",
+		"connect_failed",
+		"disconnect",
+		"error",
+	].forEach(function(e) {
+		socket.on(e, function(data) {
+			$("#loading-page-message").text("Connection failed: " + data);
```
astorije: Is this to handle cases where connection is lost at page loading time?

client/index.html
astorije: It would be great if this could be displayed above the text input instead of on it. One of the most annoying consequence to the bug you're fixing is that connection is often lost while typing a message (while your PR is 👏 👏 when this happens while reading messages).
Until the full reconnection is fixed, having this outside the form input makes it possible to copy into clipboard what I was typing, refresh, paste. Not ideal, but as you say ideally this whole thing would be fixed someday anyway, and in the meanwhile, it'll help.

What do you think?

You might have to disable the sending of the message though, but that is also probably necessary in the current solution as one can still press <kbd>Enter</kbd> (but since it has no effect...)

xPaw: > Outer <span> is an accident here?

Nope. It's to correctly do the flexbox.

> Also, I don't think there is any need for specific message. 

Specific messages absolutely help when you hit a certain issue.

client/js/lounge.js
xPaw: Yep.

client/index.html
astorije: > Nope. It's to correctly do the flexbox.

Really?? It's so weird to look at 😅.
Anyway, if you move it above the input as mentioned below, I think height should be normal text height + whatever margin, as it won't have to take up the input's height exactly.

> Specific messages absolutely help when you hit a certain issue.

I strongly disagree that `Connection failed: transport close` (which is what you hit every time 3G/Wifi connection is lost, i.e. 95% of the time this feature will be useful) is helpful to the user. What "failed", I didn't attempt any action? What transport? I didn't close anything?

I understand that from an engineering standpoint, it's interesting to know (which is why I'm perfectly OK to have it in the browser console), but from a UX standpoint, it's terrible. In fact, I don't know any general public app that displays anything else than `Connection lost`, `Connecting...`, `Please wait while you are being connected`, etc.
By making it more detailed, we make it confusing in most if not all cases that most users will experience.
This reminds me a discussion with my UX designer roommate who was telling me how his developers were always trying to display server messages, while what users really want to see is if it's working or not and what to do next when it's not :D

So let's please make it `Connection lost. Click here to reconnect` and be done with it. This will also simplify code of this PR, which is great if it's not meant to live forever :-)

xPaw: Displaying on top of the input means the latest message in channel won't be visible, no?

astorije: Not necessarily, can we make it so it's not part of the scrollable area?

xPaw:
@astorije Changed message and made it clickable. I don't know where to move the banner that it displays nicely.

astorije:
> I don't know where to move the banner that it displays nicely.

I'll try to take a stab today.

astorije:
I couldn't get to that today, give me a few and I'll give it a try.

astorije:
Alright, I am done!

Example, Morning, Zenburn | Crypto
--- | ---
screen shot 2016-12-13 at 01 54 10 | screen shot 2016-12-13 at 01 54 24

- Error message now appears on above the message input, instead of right in front of it (see reasons explained above, mainly be able to copy/paste message being written 😄)
- Left the "mechanism" (displaying/hiding the error) in `style.css`, moved design in themes. This is nice as ideally themes would entirely design their form control (buttons, inputs, etc.), they can do it through the class.
- For now, all themes but Crypto use same red color than Example, but I'll let Morning/Zenburn users come up with better reds (so far, only Crypto has a different red than Example).
- Made the whole thing a `button`, so browsers handle themselves cursor, accessibility features, etc.
- Reused `.btn` class to make this one look consistent with other buttons. Hovering will surprise some: it goes green. Well at least it did surprise me at first, but I actually really liked it that way: clicking on it is an intent to bring back the app, so green looks soothing :)

  error_hover

- Block is itself inside a flex container now (`#chat-container`). Thanks to that, no need for absolute/relative positioning and 100% sizing, nor the verbatim `span` tag 🎉 
- Rebased with `master` (that was an accident actually, sorry)

One caveat: I struggled to disable the message input so that pressing <kbd>Enter</kbd> would not clear it. It should be trivial, but maybe I'm tired or maybe the jQuery plugins dealing with input history / auto-complete / submit when hitting <kbd>Enter</kbd> are messing with me.
I stopped playing around after a while. I think that this should be fine if indeed that whole feature is not supposed to stay forever. I think this should be addressed as part of the "lot more changes" @xPaw mentioned in the description: the whole app needs to know when it's connected or not, and should handle gracefully disconnected state. Until that, meh. Is that a big deal?

client/themes/zenburn.css
```diff
@@ -253,3 +253,11 @@ body {
 #chat .toggle-content .body {
 	color: #d2d39b;
 }
+
+.btn-reconnect {
+	background: #e74c3c;
+	color: #fff;
+	border: 0;
+	border-radius: 0;
+	margin: 0;
```
astorije: Repeating these overrides does not bother me: not all themes will have bordered/radiused buttons, and `.btn-reconnect` *is* an extension of `.btn`, so it makes sense. This reuses the Bootstrap's methodology (but not its values/design, I'm not adding any ties to Bootstrap here).

YaManicKill:
> One caveat: I struggled to disable the message input so that pressing Enter would not clear it. It should be trivial, but maybe I'm tired or maybe the jQuery plugins dealing with input history / auto-complete / submit when hitting Enter are messing with me.

Wouldn't it be best just to disable the input entirely (using html disabled) because any text they type in will be lost on refresh, anyway?

astorije:
The whole point for me to do this was to prevent losing message being written, i.e. be able to copy message to clipboard before refreshing, so disabling the input wouldn't help.
Making it read-only, maybe?

YaManicKill:
> The whole point for me to do this was to prevent losing message being written, i.e. be able to copy message to clipboard before refreshing, so disabling the input wouldn't help.

You can copy from a disabled input.

astorije:
Well, TIL :) I'll so that tomorrow, thanks!

astorije:
> Great, so just disable the input and the send button using the html attribute, and then I'm 100% on this one.

Sadly, I couldn't do that: `disabled` on the `textarea` was preventing any selecting/copying to clipboard (at least on Chrome). `readonly` was allowing selecting fine... as well as submitting the message lol.

Issue was indeed coming from that cr... library `inputhistory`. You'd think it **only** deals with remembering what was sent before, not messing up with hitting <kbd>Enter</kbd>, well, no.
I ended up disabling it in disconnected mode and re-enabling the <kbd>Shift</kbd> + <kbd>Enter</kbd> combination (if not adding that `on`, this creates a new line...).
Solution is far from optimal, but hopefully this is not permanent and will be wiped out when improving either the disconnected mode or that input and history management.

Hope I can put that behind me, that was more annoying than expected :D

YaManicKill:
> Sadly, I couldn't do that: disabled on the textarea was preventing any selecting/copying to clipboard (at least on Chrome). readonly was allowing selecting fine... as well as submitting the message lol.

I'm confused about that because it works on chrome here, but alright.

I guess this isn't the worst solution ever, and we'll remove it when we get reconnect working, so...yeah...

I mean there is a point that you can go back to messages sent, so the message isn't lost anyway, but I'm not sure it matters enough. It's all still better than having it refresh on an error. So, I'm alright with whatever we go with.

PR #786, opened 2016-12-10
Add support for displaying emoji

yashsriv:
* First all : enclosed emojis are converted to unicode equivalent
* Then twemoji.js is used to convert it into img tags

This ensures uniformity irrespective of whether user actually inputted `: zap :` or the unicode equivalent...both look the same.
Preview:
preview

### DIFF
--- a/client/css/emoji.css
+++ b/client/css/emoji.css
@@ -1,0 +1,6 @@
+img.emoji {
+  height: 1em;
+  margin: 0 .05em 0 .1em;
+  vertical-align: -0.1em;
+  width: 1em;
+}
--- a/client/index.html
+++ b/client/index.html
@@ -14,6 +14,7 @@
 	<title>The Lounge</title>
 
 	<link rel="stylesheet" href="css/bootstrap.css">
+	<link rel="stylesheet" href="css/emoji.css">
 	<link rel="stylesheet" href="css/style.css">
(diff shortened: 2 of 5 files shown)


stepie22:
Just adding on to this, it would be nice to have a autocomplete feature, like github has:

!()[https://zippy.gfycat.com/UnfortunateDisfiguredAgouti.webm]

so we know what emoji's there are, for example: I would have thought :smilyface: would be a emoji, but it is instead :smile:

MaxLeiter:
My thoughts from IRC:

11:24 MaxLeiter  If slack wanted emoji to be visible they could use Unicode lol

11:24 MaxLeiter  However I'm fine with a emoji picker that converts to actual Unicode emojis (so type :zap: turns into whatever the emoji is before being sent)

maxpoulin64:
This is something I feel like should be done while sending messages, not on every incoming message. This will lead to many people sending out `:smile:` on IRC which isn't standard in any way and will look horrible (and stupid). We should convert these to actual emojis while sending out so it displays properly on everyone's client including non-lounge ones. I'm also not quite sure if we want to parse incoming messages as such either: if the user wanted an emoji there, they would have put an actual emoji.

yashsriv:
I have incorporated the suggestions here into #787

PR #789, opened 2016-12-11
Simplify introduction on README

astorije:
Features listed here are the same as on Shout repo, in same order, participating to the feeling of fork with nothing new. Instead of listing features here, we should refer to the website and improve it to make it as current as possible (there was some recent action there, and more coming, so it is reasonable to point there).

Also, this "Why the fork?" section was useful right when we forked, but now it gives unnecessary and lengthy information (it is now the most verbose section of the README!). The Lounge has enough momentum as that point to be treated as its own project.

Finally, shortening this section moves the screenshot back up on the page, and mobile view now has more context in the description.

Before | After
--- | ---
screen shot 2016-12-11 at 04 16 18 | screen shot 2016-12-11 at 04 16 04

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,36 +7,15 @@
 [![Dependencies Status](https://img.shields.io/david/thelounge/lounge.svg)](https://david-dm.org/thelounge/lounge)
 [![Developer Dependencies Status](https://img.shields.io/david/dev/thelounge/lounge.svg)](https://david-dm.org/thelounge/lounge?type=dev)
 
-__What is it?__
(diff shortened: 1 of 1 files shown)


astorije:
We can improve the intro a bit now or more later, but moving away from the current layout, pretty much a 1:1 copy of Shout's README, is a good thing IMO.

xPaw:
IMO, readme should list a couple of features (doesn't have to be an actual list).

astorije:
Fine by me (that's what I meant in my previous comment): which ones would you see there?

PR #796, opened 2016-12-14
[WIP] Transition to classes

MaxLeiter:
This is a WIP, and I'm not 100% sure I'm doing it correctly - any comments appreciated

What's done:

- clientManager


xPaw:
You're doing it correctly, but you rushed a little. We need to deal with existing PRs (e.g. get 2.2.0 out of the way), then we can start doing this. I rewrote identd/oidentd in #804 and it uses classes. User log can probably just export itself as a function (drop `.write`) no need to make it a class.

astorije:
I think this has more chances of success if it is dealt with just like @xPaw has done in #804: switching to classes when doing big work in some part of the code.

For places that we know we won't be touching/refactoring before a while, I am happy to have PRs that convert them though, especially if these PRs can be small.

> User log can probably just export itself as a function (drop .write) no need to make it a class.

@xPaw, @MaxLeiter, leave the `UserLog` to me and #663 ([here](https://github.com/thelounge/lounge/pull/663/files#diff-e96eecba92dc83c3fe89610e625adc9dx) specifically). It's still a WIP, and will probably stay that way for a little while as I try to catch up with other open PRs, but I would hate to have to rebase over such a drastic code change 😅.

PR #798, opened 2016-12-14
Add colon to first param in unhandled template, closes #797

### LINKED ISSUE #797: Add colons to unhandled messages first param
image

MaxLeiter:
image 

becomes

image

The capitalization difference is due to the server, not the PR. Only effects unknown commands.

### DIFF
--- a/client/views/msg_unhandled.tpl
+++ b/client/views/msg_unhandled.tpl
@@ -5,7 +5,9 @@
 	<span class="from">[{{command}}]</span>
 	<span class="text">
 		{{#each params}}
-			<span>{{this}}</span>
+			<span>
+				{{this}}{{#equal ../command "421"}}{{#if @first}}:{{/if}}{{/equal}}
+			</span>
 		{{/each}}
 	</span>
 </div>


MaxLeiter:
forgot that all commands passed to server are considered unhandled, so working on a fix

astorije:
@MaxLeiter, I saw some discussion around it on the channel, did you manage to fix your issues?

MaxLeiter:
I did, the #equal fixes it

client/views/msg_unhandled.tpl
```diff
@@ -5,7 +5,9 @@
 	<span class="from">[{{command}}]</span>
 	<span class="text">
 		{{#each params}}
-			<span>{{this}}</span>
+			<span>
+				{{this}}{{#equal ../command "421"}}{{#if @first}}:{{/if}}{{/equal}}
```
astorije: Why the `../`? Also, could you add a comment maybe? It's very cryptic looking at this without an explication 😅

xPaw:
Proper fix would be to handle this message on the server and emit a normal pretty error message to the client.

client/views/msg_unhandled.tpl
MaxLeiter: `command` is out of the scope of the `#each`.

astorije:
@MaxLeiter, do you want to rework this PR or close it and get back to it when you have time?

MaxLeiter:
I'll close it and re open it as I'm away this weekend

Issue #517, opened 2016-07-19
Mode flags for "start" command don't work


nirvdrum:
I may just be doing something wrong, but it looks like the `--public` and `--private` flags to the `start` command are completely ignored. The value from the config.js file always prevails. I started poking around at [the source](https://github.com/thelounge/lounge/blob/412f28aaee784bc395da2fcc61b66f962d62711f/src/command-line/start.js#L18-L20) a bit and `program.public` and `program.private` are always undefined.

I'm running 2.0.0-pre.4.

maxpoulin64:
Hmm, [I think the params are assigned the wrong way](https://github.com/thelounge/lounge/blob/master/src/command-line/start.js#L10-L11) and that might be why they are undefined.

Just as a note, there was discussion about deprecating those options in favor of a more generic one that would allow to override any config options through the command-line. So those will likely go in the near future. I would always recommend to use the config.js anyway: it is created and loaded either way, so might as well use it.

nirvdrum:
Fair enough. I was just working through the docs in order and was confused when it didn't work the way I expected. Only then did I start looking at the config file in detail.

PR #799, opened 2016-12-15
Improve CLI a bit (output formatting and bug fix)

astorije:
First commit was started as a finishing touch of [this](https://github.com/thelounge/lounge/pull/779/files#diff-cf27c1d543e886c89cd9ac8b8aeaf05bR93), then I went a bit more crazy.

Second commit of this PR should fix https://github.com/thelounge/lounge/issues/517.
I know it is proposed to deprecate them in favor of a more generic one (`-c port=80`), but in the meantime the existing ones should work properly.

### DIFF
--- a/src/log.js
+++ b/src/log.js
@@ -2,6 +2,7 @@
 
 var colors = require("colors/safe");
 var moment = require("moment");
+const read = require("read");
 var Helper = require("./helper");
 
 function timestamp(type, messageArgs) {
@@ -31,7 +32,7 @@ exports.debug = function() {
 	console.log.apply(console, timestamp(colors.green("[DEBUG]"), arguments));
 };
 
-exports.rawInfo = function() {
-	const newArguments = timestamp(colors.blue("[INFO]"), arguments);
-	return Array.prototype.slice.call(newArguments).join(" ");
+exports.prompt = (options, callback) => {
+	options.prompt = timestamp(colors.cyan("[PROMPT]"), [options.text]).join(" ");
+	read(options, callback);
 };
--- a/src/log.js
+++ b/src/log.js
@@ -30,3 +30,8 @@ exports.info = function() {
 exports.debug = function() {
 	console.log.apply(console, timestamp(colors.green("[DEBUG]"), arguments));
 };
+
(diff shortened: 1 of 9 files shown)


src/log.js
```diff
@@ -30,3 +30,8 @@ exports.info = function() {
 exports.debug = function() {
 	console.log.apply(console, timestamp(colors.green("[DEBUG]"), arguments));
 };
+
+exports.rawInfo = function() {
+	const newArguments = timestamp(colors.blue("[INFO]"), arguments);
```
xPaw: Fine by me, unsure if it needed a new file though.

src/log.js
```diff
@@ -31,7 +31,7 @@ exports.debug = function() {
 	console.log.apply(console, timestamp(colors.green("[DEBUG]"), arguments));
 };
 
-exports.rawInfo = function() {
-	const newArguments = timestamp(colors.blue("[INFO]"), arguments);
+exports.prompt = function() {
+	const newArguments = timestamp(colors.cyan("[PROMPT]"), arguments);
 	return Array.prototype.slice.call(newArguments).join(" ");
 };
```
xPaw: Keep in log for now.

src/command-line/utils/prompt.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const log = require("../../log");
+const read = require("read");
+
+module.exports = (options, callback) => {
+	options.prompt = log.prompt(options.text);
```
astorije: `prompt` != `text` on purpose for 2 reasons:
- We could still use `prompt` if we want to define a tailor-made CLI prompt without the `log.prompt` (we'll need to do `... = options.prompt || ...` or something though)
- `prompt({ prompt: "..." }, ...);` feels weird to me :D

Stupid? 😅

xPaw: I think we can do it simpler. `log.prompt("Password:", (err, password) => {});`

astorije:
Should be all set, @xPaw!

astorije:
This is a small PR, but with a big enough diff: I'll wait for a second review before merging.

PR #801, opened 2016-12-16
Do not automatically focus on touch devices

### LINKED ISSUE #758: Keyboard pops up on channel change on mobile
Because we now put focus to the text input, it means that my keyboard is popping up everytime I change channel on Android Chrome.

This seems less than ideal.

xPaw:
Closes #758.

There's one exception where it does focus: cycle nicks button, and it does that by triggering click AND focus on the input, which keeps the virtual keyboard open.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -823,15 +823,46 @@ $(function() {
 		})
 		.tab(complete, {hint: false});
 
+	var focus = $.noop;
+	if (!("ontouchstart" in window || navigator.maxTouchPoints > 0)) {
+		focus = function() {
+			if (chat.find(".active").hasClass("chan")) {
+				input.focus();
+			}
+		};
+
+		$(window).on("focus", focus);
+
+		chat.on("click", ".chat", function() {
+			setTimeout(function() {
+				var text = "";
+				if (window.getSelection) {
+					text = window.getSelection().toString();
+				} else if (document.selection && document.selection.type !== "Control") {
+					text = document.selection.createRange().text;
+				}
+				if (!text) {
+					focus();
+				}
+			}, 2);
+		});
+	}
+
+	// Triggering click event opens the virtual keyboard on mobile
+	// This can only be called from another interactive event (e.g. button click)
+	var forceFocus = function() {
+		input.trigger("click").focus();
+	};
+
 	// Cycle through nicks for the current word, just like hitting "Tab"
 	$("#cycle-nicks").on("click", function() {
 		input.triggerHandler($.Event("keydown.tabcomplete", {which: 9}));
-		focus();
+		forceFocus();
 	});
 
 	$("#form").on("submit", function(e) {
 		e.preventDefault();
-		focus();
+		forceFocus();
 		var text = input.val();
 
 		if (text.length === 0) {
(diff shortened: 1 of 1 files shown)


xPaw:
@astorije for the purposes of this PR I can't be bothered to add a fully fledged setting.

client/js/lounge.js
```diff
@@ -823,9 +823,21 @@ $(function() {
 	// Cycle through nicks for the current word, just like hitting "Tab"
 	$("#cycle-nicks").on("click", function() {
 		input.triggerHandler($.Event("keydown.tabcomplete", {which: 9}));
-		focus();
+		input.trigger("click").focus(); // Triggering click keeps the keyboard open on mobile
 	});
 
+	// Do not automatically focus on touch devices as this will open the virtual keyboard
+	var focus = $.noop;
```
YaManicKill: Wouldn't it be easier to do

var focus = function () {
  if (!("ontouchstart" in window... etc
}

rather than setting it to a noop?

YaManicKill: Good point. Happy, then.

astorije:
> @astorije for the purposes of this PR I can't be bothered to add a fully fledged setting.

I didn't mean a setting but detection at loading. That's fair, I guess we don't need it for now.

client/js/lounge.js
```diff
@@ -1039,9 +1042,7 @@ $(function() {
 			socket.emit("names", {target: self.data("id")});
 		}
 
-		if (screen.width > 768 && chan.hasClass("chan")) {
-			input.focus();
-		}
+		focus();
```
astorije: Actually, this never works on my phone (Nexus 4, stock Android, stock Chrome). After I send a message, focus is set to the input. I'm guessing the detection failed there.

xPaw: That means it worked as intended. You want to keep the keyboard open after sending message? Or at least focus on the input without doing touch check?

astorije: > That means it worked as intended.

My bad then, for some reason I thought the intent was different.

> You want to keep the keyboard open after sending message? Or at least focus on the input without doing touch check?

Indeed, it seems counter-intuitive at the moment. I realize that *every time* I send a message, I expect the keyboard to stick around. This is probably because all other messaging apps I use on my phone (Slack, Telegram, SMS) do that.
I think we should do that as well after sending a message, setting/canceling nick, and cycling through nicks. But when changing channels, we should definitely not open the keyboard 😅

xPaw: @YaManicKill what you think? Should we keep (re-open in reality) the keyboard after clicking the send button?

astorije: I was concerned for a bit but just clarifying after testing: if this follows the same pattern than the nick-cycler, programmatically, it's re-opening indeed, but in practice, there is no flickering of the keyboard, it stays open.

I have just been playing with The Lounge and other messaging apps, and I definitely see this is as a UX improvement.

xPaw: On current master clicking the cycle button closes the keyboard for me, this PR does a neat trick by triggering click, then focusing on the input to keep the keyboard open. I can do the same for the send button.

YaManicKill: I think keeping it open when clicking send and clicking the cycle button should both keep the keyboard open (or, technically reopen). I think that's most expected.

xPaw: Done.

xPaw:
Also disabled the "clicking on messages but not selecting it should focus" code block on touch.

Issue #803, opened 2016-12-16
Combine oidentd and identd implementations


xPaw:
Implement identd on top of oidentd as the code is newer and more sane (e.g. uses `Map` for connections)

astorije:
This was addressed by #804.

PR #804, opened 2016-12-17
Rewrite identd server, combine with oidentd

xPaw:
- #803 Combines identd and oidentd into a single implementation.
- Fixes a race condition where server could autoload users faster than starting the identd server, which in result would connect to IRC servers without providing ident for these connections.
- Adds a stricter identd resolution by also comparing remote and local addresses (for future multi bind support).
- Correctly binds identd server on same address as the users' bind address.
- Closes connections for invalid identd requests (when both input ports aren't in 1-65535 range).


src/clientManager.js
```diff
@@ -5,17 +5,25 @@ var colors = require("colors/safe");
...
 var Helper = require("./helper");
-var Oidentd = require("./oidentd");
 
 module.exports = ClientManager;
 
 function ClientManager() {
 	this.clients = [];
+}
+
+ClientManager.prototype.init = function(identHandler, sockets) {
```
xPaw: Can't use constructor because CLI commands also initalize client manager.

astorije: I wouldn't want any of this in a constructor anyway :)

src/clientManager.js
```diff
@@ -5,17 +5,25 @@ var colors = require("colors/safe");
...
+		if ("autoload" in Helper.config) {
+			log.warn(`Autoloading users is now always enabled. Please remove the ${colors.yellow("autoload")} option from your configuration file.`);
+		}
 
-	if (typeof Helper.config.oidentd === "string") {
-		this.identHandler = new Oidentd(Helper.config.oidentd);
+		this.autoloadUsers();
 	}
-}
+};
```
astorije: What's the reason behind moving autoloading users from `server.js` to `clientManager.js`?

src/identification.js
```diff
@@ -0,0 +1,109 @@
...
+class Identification {
+	constructor(startedCallback) {
+		this.connectionId = 0;
+		this.connections = new Map();
+
+		if (typeof Helper.config.oidentd === "string") {
+			this.oidentdFile = Helper.expandHome(Helper.config.oidentd);
+			log.info(`Oidentd file: ${colors.green(this.oidentdFile)}`);
+
+			this.Refresh();
```
astorije: What? No no, `thisIsAFunction` vs. `ThisIsAClass`. Same with `AddSocket` vs `addSocket`, etc.

src/identification.js
```diff
@@ -0,0 +1,109 @@
...
+
+				startedCallback();
+			});
+		} else {
+			startedCallback();
+		}
+	}
+
+	ServerConnection(socket) {
+		socket.on("data", (data) => {
```
astorije: Minor nitpick: `(data) =>` can be `data =>` with a single argument. I don't have a preference, especially on multiline callbacks, but I think it's nice on short callbacks like `users.map(u => u.name)`.

src/identification.js
```diff
@@ -0,0 +1,109 @@
...
+			this.RespondToIdent(socket, data);
+			socket.end();
+		});
+	}
+
+	RespondToIdent(socket, data) {
+		data = data.toString();
+		data = data.split(",");
```
astorije: `data = data.toString().split(",");` instead? Or does that hide something else like a missing `if` or whatever?

src/identification.js
```diff
@@ -0,0 +1,109 @@
...
+			&& connection.socket.localAddress === socket.localAddress) {
+				return socket.write(`${lport}, ${fport} : USERID : UNIX : ${connection.user}\r\n`);
+			}
+		}
+
+		socket.write(`${lport}, ${fport} : ERROR : NO-USER\r\n`);
+	}
+
+	AddSocket(socket, user) {
+		const id = this.connectionId++;
```
astorije: Is that a problem outside of exceeding maximum integer (which is [rather high](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER))?

Could do something like http://stackoverflow.com/a/3439649/1935861 where you store the deleted ids in an array and when you add a socket, you pick one of them (first, last, w/e) and if none, increase counter. That seems overkill to me, but I don't know enough about the overall feature to know if it's a DoS vector or not (if not, no need to care IMO).

src/plugins/irc-events/connection.js
```diff
@@ -63,23 +62,15 @@ module.exports = function(irc, network) {
...
-	}
+	let identSocketId;
 
-	if (identHandler) {
-		let identSocketId;
-
-		irc.on("raw socket connected", function(socket) {
-			identSocketId = identHandler.addSocket(socket, client.name || network.username);
-		});
+	irc.on("raw socket connected", function(socket) {
```
astorije: Was there a discussion on the IRC channel with prawnsalad saying we should not use `raw` event? Very likely I am wrong on this, but just double-checking.

src/server.js
```diff
@@ -75,25 +81,13 @@ module.exports = function() {
...
-		}
+	manager = new ClientManager();
 
-		manager.autoloadUsers();
-	}
+	new Promise((resolve) => {
+		identHandler = new Identification(resolve);
+	}).then(() => {
+		manager.init(identHandler, sockets);
+	});
```
astorije: This seems weird to me, as you're manipulating 2 promises without using the resulting one.

It seems equivalent to:

```js
identHandler = new Identification(() => manager.init(identHandler, sockets));
```

A nicer way to do this IMO would be to have the constructor construct his thing only (see [this](http://stackoverflow.com/a/24686979/1935861)) then have another method that returns a promise.

Then you can do:

```js
identHandler = new Identification().start(); // Or whatever name is appropriate here
identHandler.then(() => manager.init(identHandler, sockets));
```

Not a fan of having a constructor do anything else than creating its own instance...

src/clientManager.js
xPaw: Makes more sense, users stuff in the manager that handles users.

src/plugins/irc-events/connection.js
xPaw: The discussion was that we *should* use `raw`.

xPaw:
@astorije Addressed.

astorije:
> Looks fine, with my limited knowledge of that area :-P

Yeah, I felt the same, so I'd rather rely on @maxpoulin64's review and testing here :)

ronilaukkarinen:
Built-in identd confirmed working when no identd servers running on system. Thanks!

astorije:
@maxpoulin64, I know you are fairly distant from the project right now, sorry to bother you, but would you mind giving this a review and if possible a quick test? Thanks a lot! :)

YaManicKill:
@astorije Looks like I guess he's too busy for this just now. Do we want to merge this in now, so that we can get people testing it on master for a while before we release? At least a week would be good.

astorije:
In case of merge-and-see-what-happens, I'd be more comfortable if we do this after v2.2.0 is released. @xPaw is back to a no-laptop mode during the week, so he wouldn't be able to report and/or fix any issues that could potentially arise, and it's likely that I wouldn't be able to fix them myself.
Are you okay with moving this one to v2.3.0? I don't have a strong position either if you really want it in.

maxpoulin64:
I'm still fairly busy so I haven't had time to test it out yet. If we're lucky I might have time to do my review before 2.2.0 but lets move it to 2.3.0 if it's to delay the release.

YaManicKill:
> Are you okay with moving this one to v2.3.0?

Yeah, that's fine. Unless @xPaw has a strong reason to put it in 2.2.0, I think it's better to get 2.2.0 out than wait for this.

> I'm still fairly busy

Don't worry about it, @maxpoulin64, it's a busy time of year and life comes first :-)

xPaw:
It can wait.

astorije:
@maxpoulin64, no pressure but any chance you can test this soon-ish? Thanks!

astorije:
Alright, merging this now. We have until v2.3.0 to deal with the fallout :)
It would be shame to have to fix conflicts on this.

PR #810, opened 2016-12-17
Implement color hotkeys

xPaw:
Same key bindings as http://www.ircbeginner.com/ircinfo/colors.html

Even though the characters are invisible (in Chrome at least), it's still useful as an initial implementation.

I moved all hotkey bindings into their own scope (add `?w=1` to url for saner diff)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1284,45 +1284,81 @@ $(function() {
 		forms.find(".username").val(nick);
 	});
 
-	Mousetrap.bind([
-		"command+up",
-		"command+down",
-		"ctrl+up",
-		"ctrl+down"
-	], function(e, keys) {
-		var channels = sidebar.find(".chan");
-		var index = channels.index(channels.filter(".active"));
-		var direction = keys.split("+").pop();
-		switch (direction) {
-		case "up":
-			// Loop
-			var upTarget = (channels.length + (index - 1 + channels.length)) % channels.length;
-			channels.eq(upTarget).click();
-			break;
+	(function HotkeysScope() {
+		Mousetrap.bind([
+			"command+up",
+			"command+down",
+			"ctrl+up",
+			"ctrl+down"
+		], function(e, keys) {
+			var channels = sidebar.find(".chan");
+			var index = channels.index(channels.filter(".active"));
+			var direction = keys.split("+").pop();
+			switch (direction) {
+			case "up":
+				// Loop
+				var upTarget = (channels.length + (index - 1 + channels.length)) % channels.length;
+				channels.eq(upTarget).click();
+				break;
 
-		case "down":
-			// Loop
-			var downTarget = (channels.length + (index + 1 + channels.length)) % channels.length;
-			channels.eq(downTarget).click();
-			break;
+			case "down":
+				// Loop
+				var downTarget = (channels.length + (index + 1 + channels.length)) % channels.length;
+				channels.eq(downTarget).click();
+				break;
+			}
+		});
+
+		Mousetrap.bind([
+			"command+shift+l",
+			"ctrl+shift+l"
+		], function(e) {
+			if (e.target === input[0]) {
+				clear();
+				e.preventDefault();
+			}
+		});
+
+		Mousetrap.bind([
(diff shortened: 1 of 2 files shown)


client/js/lounge.js
```diff
@@ -1269,45 +1269,65 @@ $(function() {
...
+			contextMenuContainer.hide();
+		});
 
-	Mousetrap.bind([
-		"escape"
-	], function() {
-		contextMenuContainer.hide();
-	});
+		var colorsHotkeys = {
+			k: "\x03",
```
astorije: `command+k` is already used by clearing the screen on macOS (see L1298)... I don't want us to clearing this to `command+l` because it focuses the URL bar, which is immensely useful.
Any other hot key we can use for colors?

dgw: Is there any reason not to simply bind <kbd>Ctrl</kbd>+<kbd>K</kbd> on every platform? That's what a number of existing clients use, and it would reduce The Lounge's learning curve if that keybinding remained the same.

xPaw: @astorije On windows it uses `ctrl+shift+l`. While `ctrl+l` focuses address bar, shift modifier does not. Can we do the same on mac?

astorije: So <kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>L</kbd> to clear the screen and <kbd>cmd</kbd>+<kbd>K</kbd> to set up color? Works for me.

xPaw: Done.

DanielOaks:
Nice, this works well 👍. After this gets merged perhaps making it easier to see that you've pressed this would be good -- right now it's just a little hard to tell that you've actually entered one of these chars until you hit enter.

astorije:
Sadly this is almost impossible with the current `textarea`, but if we ever go for a rich-content equivalent, it would. Not on the agenda for now though.
I reckon @xPaw had a crazy idea with tailor-made font or something, but I can't find it in the issues/PRs.

xPaw:
@astorije it's a branch

astorije:
Ah ah, I see the result, pretty crazy indeed! I don't think the font-family falls back on the next font when a char is missing (i.e. all non-control-code chars in the case of a font with just B, I, U symbols), does it?

client/js/lounge.js
```diff
@@ -1275,45 +1275,65 @@ $(function() {
...
 
-		case "down":
-			// Loop
-			var downTarget = (channels.length + (index + 1 + channels.length)) % channels.length;
-			channels.eq(downTarget).click();
-			break;
-		}
-	});
+		Mousetrap.bind([
+			"command+shift+l",
```
xPaw: Write a changelog entry for this change.

client/js/lounge.js
```diff
@@ -1275,45 +1275,65 @@ $(function() {
...
+			o: "\x0F",
+		};
+
+		for (var hotkey in colorsHotkeys) {
+			Mousetrap.bind([
+				"command+" + hotkey,
+				"ctrl+" + hotkey
+			], function(e) {
+				e.preventDefault();
+				input.val(input.val() + colorsHotkeys[e.key]);
```
xPaw: This does not care about cursor position, need to fix.

xPaw:
I've updated the PR to correctly insert formatting characters at the position where the cursor is. Also supports selections.

astorije:
@xPaw, I added the new shortcuts in the help windows, let me know what you think. I'll review your recent changes as well.

williamboman:
I'm seeing issues like this on macOS;

colors

xPaw:
@astorije I don't quite understand why we have separate sections for windows and macOS bindings. The only difference is the meta key (ctrl vs ⌘)

astorije:
Because that might not be the case for long, due to the different nature of OSes. As a matter of fact, they used to be different before a recent PR.
Also, I think we should quickly hide this distinction. It must not be difficult to detect if we are on Windows/Linux or macOS on the front-end and display the according section only. I am less certain on how to be sure we have a keyboard on the client or not.

PR #812, opened 2016-12-18
No prefetch URLs unless are messages

birkof:


astorije:
What made you open this?
Did you notice a message type that causes a prefetch? It would be unfortunate that parts/quits trigger a prefetch.

I just tried with a `/me` (`ACTION`) message and it did not prefetch, which is something we should do.

birkof:
I've found that URIs in notice are parsed on server window

xPaw:
@astorije Introduced in 62d4cd8fe824026f37c9564cd6b39f34c78a67de, as it previously only worked in `privmsg`, and it goes via a common function to parse messages, notices, wallops...

astorije:
Gotcha @xPaw. What about allowing `Msg.Type.ACTION` as well?

xPaw:
@astorije Dunno, IRC rfc tends to say that actions should not be acted upon anyhow.

astorije:
Hmm, odd. In terms of UX, as a user, I would expect that "I love that picture: URL" and "/me loves this picture: URL" are being expanded the same in The Lounge.

src/plugins/irc-events/message.js
```diff
@@ -91,6 +91,9 @@ module.exports = function(irc, network) {
 		});
 		chan.pushMessage(client, msg, !self);
 
-		LinkPrefetch(client, chan, msg);
+		// No prefetch URLs unless are simple messages.
+		if (data.type === Msg.Type.MESSAGE) {
```
astorije: I would add `Msg.Type.ACTION` here. If strong opinion against this, I'll change my mind, but as a user I'd expect that.

astorije:
@xPaw, @birkof, ping? I really think `Msg.Type.ACTION` should allow for prefetching as well, along with URL parsing, channel parsing, etc.

astorije:
BTW, this is a regression introduced after 2.1.0 so I'm tagging this one as required for 2.2.0 😢

src/plugins/irc-events/message.js
YaManicKill: Action is "/me", yeah? I agree that prefetch should work with those as well.

birkof:
Hey guys. Sorry for delay. In my opinion, I think it is irrelevant to the user to have notices prefetched on Lobby window (just my 2 cents) that's why i've request this merge

src/plugins/irc-events/message.js
astorije: Correct.

astorije:
@birkof, I believe we all agree on this :)
`ACTION` represents messages starting with `/me`:

screen shot 2017-01-04 at 21 03 21

Could you add this change here please?
Thanks!

astorije:
@birkof, mind doing this small change so we can put this behind us and get one step closer to the super-near [v2.2.0](https://github.com/thelounge/lounge/milestone/2)? :))

src/plugins/irc-events/message.js
astorije:
```js
if (data.type === Msg.Type.MESSAGE || data.type === Msg.Type.ACTION) {
```

would probably do the trick.

YaManicKill:
@birkof If you won't have time in the next few weeks, can you let us know, so we can either push it to the next release or update it ourselves. I don't think this is something we want to hold up the release for.

src/plugins/irc-events/message.js
xPaw: Filter out notices instead?

birkof:
I've also added the ACTION type to prefetching  condition.

astorije:
Woops, looks like you missed your rebase seeing the massive commit list and changes. If you make sure that the "Allow maintainers to make changes" checkbox is checked on the top right section of this page and agree with it, I can fix your branch.

src/plugins/irc-events/message.js
astorije: @xPaw, I'd rather not. I'd prefer a whitelist than a blacklist, at least at this stage and for v2.2.
That would mean filtering out topics as well probably, parts, quits, aways, etc.

xPaw:
Look at what that file handles, it only affects messages.

birkof:
@xPaw look at message.js file. it handles messages and actions.
image

birkof:
@astorije I allowed edits from maintainers. My mistake was to not make a new branch with this fix, and put it into master.

birkof:
Peeeeerfect! Thanks

astorije:
Hey @birkof, we have sticker packs for our contributors now!

Issue #814, opened 2016-12-18
Give unhandled messages css class name with numeric


DanielOaks:
So in my theme I'd like to special-case the display of some messages like MAP, which TheLounge currently doesn't parse out specially. Right now I'm not able to do so since they show up with just the `unhandled` class:

![thelounge-msg-unhandled](https://cloud.githubusercontent.com/assets/251281/21291584/246eba56-c531-11e6-8097-91807db73184.PNG)

It'd be really cool if the command/verb could be slugified, prepended with "unhandled-", and added to that as well (final css classes ending up as something like `msg unhandled unhandled-015`.

Any chance something along these lines could be done? Basically, it'd allow themes to react specially to specific messages that are currently unhandled.

xPaw:
Yeah the slug idea is great. I've been thinking about doing slug classnames also for nicknames, channels, servers, to allow for more customization.

PR #816, opened 2016-12-18
Add slug with command to unhandled messages

DanielOaks:
This PR implements #814, essentially adding a class like `msg-002` to all `unhandled` messages. This allows themes to treat those messages specially (when otherwise, all they could say was "that message is an unhandled message".

Looks like this:

lounge-unhandled

### DIFF
--- a/client/views/network.tpl
+++ b/client/views/network.tpl
@@ -1,5 +1,5 @@
 {{#each networks}}
-<section id="network-{{id}}" class="network" data-id="{{id}}" data-nick="{{nick}}" data-options="{{tojson serverOptions}}">
+<section id="network-{{id}}" class="network name-{{slugify name}} network-{{slugify serverOptions.NETWORK}}" data-id="{{id}}" data-nick="{{nick}}" data-options="{{tojson serverOptions}}">
 	{{> chan}}
 </section>
 {{/each}}
--- a/client/js/libs/handlebars/slugify.js
+++ b/client/js/libs/handlebars/slugify.js
@@ -1,0 +1,5 @@
+"use strict";
+
+module.exports = function(orig) {
+	return orig.toLowerCase().replace(/[^a-z0-9]/, "-");
+};
--- a/client/views/chan.tpl
+++ b/client/views/chan.tpl
@@ -1,5 +1,5 @@
 {{#each channels}}
-<div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}}">
+<div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
 	<button class="close" aria-label="Close"></button>
 	<span class="name" title="{{name}}">{{name}}</span>
--- a/client/views/msg_unhandled.tpl
+++ b/client/views/msg_unhandled.tpl
@@ -1,4 +1,4 @@
-<div class="msg {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}}" data-time="{{time}}">
+<div class="msg msg-{{slugify command}} {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}}" data-time="{{time}}">
 	<span class="time tooltipped tooltipped-e" aria-label="{{localetime time}}">
 		{{tz time}}
 	</span>
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -25,6 +25,7 @@ function Network(attr) {
 		irc: null,
 		serverOptions: {
 			PREFIX: [],
+			NETWORK: "",
 		},
 		chanCache: [],
 	});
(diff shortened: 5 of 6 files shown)


src/plugins/irc-events/unhandled.js
```diff
@@ -14,6 +15,7 @@ module.exports = function(irc, network) {
 		network.channels[0].pushMessage(client, new Msg({
 			type: Msg.Type.UNHANDLED,
 			command: command.command,
+			commandslug: slugify(command.command),
```
xPaw: Kinda silly to do this on the server. After `webpack` PR is merged, we will be able to directly require `slugify` on the client, and use it there.

To be fair, we don't even need to use a library for this, as it can be accomplished with a simple replace.

DanielOaks: Makes sense, this just seemed like the easiest place to throw it. You could do it using a replace, I'm just paranoid about any dodgy messages coming through from the server or commands using weird characters, which is why I'd tend towards slugifying it fully I guess.

xPaw: Yes, but this would be solved by removing characters we don't want.

1. Replace space to `-`
2. Remove `/[^a-z0-9\-]/`

xPaw: Ah you are right. Neither commands, usernames, nicknames, channel names or network names can contain spaces, so there is indeed no need to do that. Probably need to deal with a couple of special characters that are allowed in these though (e.g. dots in hostnames). As I mentioned in the issue, it would be good to slugify all these.

DanielOaks:
@xPaw Maybe something like that?

xPaw:
Make the slug a handlebars template, then you don't need to touch `lounge.js`. Also slugify networks and channels in sidebar as I mentioned previously.

client/views/network.tpl
```diff
@@ -1,5 +1,5 @@
 {{#each networks}}
-<section id="network-{{id}}" class="network" data-id="{{id}}" data-nick="{{nick}}" data-options="{{toJSON serverOptions}}">
+<section id="network-{{id}}" class="network name-{{slugify name}} net-{{slugify serverOptions.NETWORK}}" data-id="{{id}}" data-nick="{{nick}}" data-options="{{toJSON serverOptions}}">
```
xPaw: I was thinking users' displayed network name, not the one server gives, but hm..

DanielOaks: This does both yeah, since I could see themes special-handling specific networks as well.

xPaw: `name-xxx` would accomplish that already, no?

DanielOaks: Nah because `name-xxx` is totally freeform. For instance, if I was connecting to IRCHighway I could name the network `irchighway`, `irch`, `rp1`, etc. Where in all those cases, the `name-xxx` would be different it would always have `net-irchighway` (since that's what's given in `RPL_ISUPPORT`).

I could see themes trying to, say, show an icon / use specific colours depending on the network, which is why I included both.

xPaw: > Nah because name-xxx is totally freeform.

Is that a problem? It would be consistent to the displayed name.

xPaw: By the way, I do think it's a fine idea, not saying to remove it. But change `net-` to `network-` to keep it consistent.

DanielOaks: I'm thinking that, say in my theme I want to display a little Rizon-specific logo for Rizon, I can make it show on `net-rizon` and no matter what the user names the network in their client, it'll show up appropriately. I think that'd be the most common use case for themes making use of the network name.

astorije:
@DanielOaks, @xPaw, what about using `data` attributes instead?
We could have `data-key="value"` instead of `class="key-value"`. This seems more appropriate to me, as it's pretty much what `data` attributes are for.

xPaw:
@astorije styling css by targeting data attributes is not the greatest thing to do.

astorije:
Why? It's actually pretty good IMO: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes#CSS_access

Using CSS classes feels messy to me when it comes to network/chan names, etc.
I am less concerned about the unhandled codes as they are closer to message types.

DanielOaks:
I can do `data` attributes for the network/channel stuff, sure – makes sense.

xPaw:
Can you also add slug to each channel window? This would allow styling channels individually (e.g. I'd like to hide nickname column on a channel where only a single bot can post).

astorije:
I'm afraid this is going to start very messy in the markup / custom CSS, but I'm not against as there is no better way to do that currently.

When/if we ever have perfect settings and/or plugins, we can clean these up, but until then, that will do the job.

astorije:
@DanielOaks, ping? :)

astorije:
> I'm afraid this is going to start very messy in the markup / custom CSS, but I'm not against as there is no better way to do that currently.

What do other @thelounge/maintainers think? I'm not strongly against, just make sure we're not setting a precedent we'll have a hard time dealing with in the future.

MaxLeiter:
Personally am a fan, unless someone can think of an alternative

astorije:
@DanielOaks, sorry but could you rebase on master and fix conflicts before we can merge, please? :)

astorije:
Also, merge commits

astorije:
@DanielOaks, going over this when building the changelog entry, do you think you'd have a tiny bit of bandwidth to have a few unit tests for the `slugify` helper?

PR #817, opened 2016-12-18
Webpack

xPaw:
All discussion in #640. This PR is finalized and ready.

Separate PR because I moved the branch to our repo for easier fixing (don't have to deal with the fork repo).


package.json
```diff
@@ -59,14 +58,22 @@
...
   },
   "devDependencies": {
+    "babel-core": "6.21.0",
+    "babel-loader": "6.2.10",
+    "babel-preset-es2015": "6.18.0",
     "chai": "3.5.0",
     "eslint": "3.10.2",
     "font-awesome": "4.7.0",
     "handlebars": "4.0.6",
+    "handlebars-loader": "1.4.0",
```
xPaw: > comes with its opinionated (as in, it differs from the vanilla Handlebars) format

Does it?

I think we benefit a lot more from using a well test covered loader that solves some of the pain points we hit during creation of this PR. What this allows us to do is `import` templates directly, and the returned template is already compiled ready for consumption. Helpers are turned into generic functions which can also be used by other parts of our code without hitting Handlebars at alal.

YaManicKill: I'm with @xPaw on this one. The new format is actually more generic (the old one binds directly to handlebars, this one uses standard module formats and just does the job).

Also makes them infinitely more testable.

package.json
```diff
@@ -59,14 +58,22 @@
     "ldapjs": "1.0.1"
   },
   "devDependencies": {
+    "babel-core": "6.21.0",
+    "babel-loader": "6.2.10",
+    "babel-preset-es2015": "6.18.0",
     "chai": "3.5.0",
     "eslint": "3.12.2",
     "font-awesome": "4.7.0",
     "handlebars": "4.0.6",
```
astorije: So far `handlebars` in `devDependencies` was used just for testing, but now it is used to build the bundle. Version we had in the repo was v4.0.5. Are we ok going with v4.0.6 as part of this PR or play it safe and go one change at a time?

FYI diff is at https://github.com/wycats/handlebars.js/compare/v4.0.5...v4.0.6.

package.json
```diff
@@ -59,14 +58,22 @@
...
+    "handlebars-loader": "1.4.0",
     "istanbul": "0.4.5",
+    "jquery": "2.1.1",
+    "jquery-ui": "1.12.1",
```
astorije: There is a massive [diff between 1.11.1 (version in currently in repo) and 1.12.1](https://github.com/jquery/jquery-ui/compare/1.11.1...1.12.1). I tried and we can't use the tag directly from GitHub so I guess we'll have to go with that.
I don't think we use a lot from jQuery UI anyway, this can probably go at some point...

client/js/lounge.js
```diff
@@ -1,4 +1,18 @@
-"use strict";
+// vendor libraries
+import "jquery-ui/ui/widgets/sortable";
+import $ from "jquery";
+import io from "socket.io-client";
```
astorije: At the moment, this does build (at least on my machine), and I'm very confused why it builds on Travis CI (at least for the version of Node/npm we have in common).

I get this at the end of `npm install` on a clean repo (`git clone https://github.com/thelounge/lounge && cd lounge && git checkout pr/640 && npm i`):

```js
   [0] multi vendor 88 bytes {1} [built] [1 error]
    + 75 hidden modules

ERROR in multi vendor
Module not found: Error: Cannot resolve module 'socket.io-client' in /private/tmp/lounge
 @ multi vendor

ERROR in ./client/js/lounge.js
Module not found: Error: Cannot resolve module 'socket.io-client' in /private/tmp/lounge/client/js
 @ ./client/js/lounge.js 9:14-41
```

App does not run in the browser, I get this in the browser console:

```js
 bundle.vendor.js:105 Uncaught Error: Cannot find module "socket.io-client"
    at bundle.vendor.js:105
    at Object.<anonymous> (bundle.vendor.js:105)
    at t (bootstrap 0a4ff42…:50)
    at bootstrap 0a4ff42…:93
    at bootstrap 0a4ff42…:93
lounge.js:4 Uncaught Error: Cannot find module "socket.io-client"
    at lounge.js:4
    at Object.<anonymous> (lounge.js:4)
    at t (bootstrap 0a4ff42…:50)
    at window.webpackJsonp (bootstrap 0a4ff42…:21)
    at bundle.js:1
```

I replaced this line with `import io from "socket.io/lib/client";` to fix it.
An alternative is to add `socket.io-client` in `package.json` but I'd rather count on the version provided by the server. I'll add this in an extra commit to see if Travis is happy with it, let me know what you think.

package.json
astorije: Alright, I'll go with the flow. I agree that there are some clear advantages to it despite my concerns.

astorije:
Alright, I give up, I am unable to push to this branch. Either Git thinks I created a new branch or I get `fatal: pr/640 cannot be resolved to branch.`. Branch name at the top of this PR is `pr/640` and in my working copy, it comes from upstream as `origin/PR/640`. First time I've ever had that crappy issue...

Anyway, diff I was trying to push:

```diff
───────────────────────────────────────────────────────────────────────────────────────
modified: client/js/lounge.js
───────────────────────────────────────────────────────────────────────────────────────
@ lounge.js:4 @
 // vendor libraries
 import "jquery-ui/ui/widgets/sortable";
 import $ from "jquery";
-import io from "socket.io-client";
+import io from "socket.io/lib/client";
 import Mousetrap from "mousetrap";
 import URI from "urijs";

───────────────────────────────────────────────────────────────────────────────────────
modified: webpack.config.js
───────────────────────────────────────────────────────────────────────────────────────
@ webpack.config.js:14 @ module.exports = {
            "jquery",
            "jquery-ui/ui/widgets/sortable",
            "mousetrap",
-           "socket.io-client",
+           "socket.io/lib/client",
            "urijs",
        ],
    },
```

client/js/lounge.js
xPaw: That's the same issue I hit in the server tests PR.

package.json
xPaw: It's only used for sorting channels.

.gitignore
```diff
@@ -6,6 +6,7 @@ coverage/
 # Built assets created at npm install/prepublish time
 # See https://docs.npmjs.com/misc/scripts
 client/fonts/
-client/js/libs.min.js.map
-client/js/libs.min.js
-client/js/lounge.templates.js
+client/js/bundle.js
+client/js/bundle.js.map
+client/js/bundle.vendor.js
+client/js/bundle.vendor.js.map
```
xPaw: I think we can ignore vendor bundle map. Map for our own code is useful to have.

package.json
```diff
@@ -14,16 +14,15 @@
   "scripts": {
     "coverage": "istanbul cover node_modules/mocha/bin/_mocha",
     "start": "node index",
+    "start-dev": "npm-run-all --parallel watch start",
```
astorije: Minor: What about `start:watch` instead?

astorije: Minor: What about `start:watch` instead?

xPaw:
@astorije So I think just listing `socket.io-client` dependency is a lot more cleaner *(we do depend on it)*, same issue was in #838 where I listed it.

package.json
xPaw: `start-dev` is easier to type, imo

xPaw:
By the way, final JS size decreases:
- master: lounge.js + lounge.templates.js + libs.min.js + socket.io.js = 389.8 kB
- webpack: bundle.js + bundle.vendor.js = 313.8 kB

package.json
xPaw: Handlebars was bumped in https://github.com/thelounge/lounge/commit/6023035838dacb0c487246ab8aef6833a6df44d2 (master), not in this PR.

package.json
astorije: Aka a feature that one person out of 10 discovers months after using the app :D

package.json
astorije: Nah, that was Handlebars used by the helpers tests, but the version in repo, i.e. [the one currently used by the client is v4.0.5](https://github.com/thelounge/lounge/blob/master/client/js/libs/handlebars.js#L3).

astorije:
> So I think just listing socket.io-client dependency is a lot more cleaner (we do depend on it)

It's not less clean in any way. And actually, [`socket.io` even exposes `/lib/` and its content](https://github.com/socketio/socket.io/blob/master/package.json#L14-L17). We don't depend on `socket.io-client`, we depend on `socket.io` and the client it exposes.
`socket.io-client` is useful when you are building a client-only app, where the server is another project/repo and you just want a client lib to connect to it. In our case, we are doing both, we have no interest in manually keeping track of a disconnected `socket.io-client` since the server already does it for us.

Also, this is not acceptable as long as we support npm v2: specifying both `socket.io` and `socket.io-client` will cause the latter to be installed twice, which we certainly don't want.

package.json
xPaw: Right, client has been tested on 4.0.6 already and it doesn't break.

xPaw:
For history: @astorije's method was to use "socket.io/lib/client", but that code is not the client code that can run in browser, it's some client instance code for the server. The serveClient method serves a JS directly from `socket.io-client` package.

astorije:
@xPaw, kind of a good news, I just realized that serving the minified client is only done as of `socket.io` v1.7.0 (https://github.com/socketio/socket.io/pull/2766) released in November 27.
Greenkeeper has updated the lib from 1.5.1 to 1.7.2 [8 days ago](https://github.com/thelounge/lounge/commit/55f1f844788dcb966f989f9f7cee6f2e6c66967e#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L57) which is most likely why none of us could see the minified version served.

I just tested and, indeed, while my normal instance (v2.1.0) is serving a 181k (!!) client, `master` is correctly serving a 70k client as expected.

Not being minified was the only drawback of using the server-provided version of the client, so I definitely think we should leverage it, at least as long as we are using `socket.io` on the server.

PR #818, opened 2016-12-18
Strip control characters from notifications

xPaw:
Removes 0-31 and 127 ascii characters as they appear as unrenderable boxes in notifications (at least in Chrome).

Ref: https://en.wikipedia.org/wiki/ASCII#Control_characters


astorije:
Requested review from either @maxpoulin64 or @YaManicKill.

PR #820, opened 2016-12-19
Fix unread marker being removed from DOM

### LINKED ISSUE #762: Unread marker sometimes doesn't show up
So, it appears the date separator means that sometimes the unread marker doesn't seem to work.

Reproduction (I think):

* Open a channel, ensure that there is a date marker somewhere visible on the channel
* Move to a different channel
* Wait for the channel to show there is a new message (with the unread badge)
* Go to channel
* Note lack of unread marker

In my testing, it doesn't seem to happen every single time, as I have seen the unread marker once or twice, but more often than not it doesn't appear.

Example:

90295a13f6667954dcbf7a7f0a710651

xPaw:
Fixes #762.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1318,26 +1318,14 @@ $(function() {
 	setInterval(function() {
 		chat.find(".chan:not(.active)").each(function() {
 			var chan = $(this);
-			if (chan.find(".messages .msg:not(.unread-marker)").slice(0, -100).remove().length) {
+			if (chan.find(".messages .msg").slice(0, -100).remove().length) {
 				chan.find(".show-more").addClass("show");
 
 				// Remove date-seperators that would otherwise be "stuck" at the top
 				// of the channel
-				var prev;
-				$(chan.find(".messages").children()).each(function() {
-					if (!prev) {
-						// Should always be a date-seperator, because it's always added
-						prev = $(this);
-					} else {
-						var current = $(this);
-
-						if (current.attr("class") === "date-marker") {
-							prev.remove();
-						} else {
-							return false;
-						}
-
-						prev = current;
+				chan.find(".date-marker").each(function() {
+					if ($(this).next().hasClass("date-marker")) {
+						$(this).remove();
 					}
 				});
 			}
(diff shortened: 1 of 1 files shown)


client/js/lounge.js
```diff
@@ -1331,7 +1331,7 @@ $(function() {
 					} else {
 						var current = $(this);
 
-						if (current.attr("class") === "date-marker") {
+						if (current.hasClass("date-marker") && prev.hasClass("date-marker")) {
```
YaManicKill: If we have the unread marker at the top, and we have 2 date markers straight after, this will not delete 1 of the date markers, will it? Because it'll drop into the else and stop there?

client/js/lounge.js
```diff
@@ -1345,9 +1345,10 @@ $(function() {
 	}, 1000 * 10);
 
 	function clear() {
-		chat.find(".active .messages .msg:not(.unread-marker)").remove();
-		chat.find(".active .show-more").addClass("show");
-		chat.find(".active .date-marker").remove();
+		chat
+			.find(".active")
+				.find(".show-more").addClass("show").end()
```
YaManicKill: Dodgy indentation.

client/js/lounge.js
xPaw: The intended behaviour is to keep the lowest date marker, does this code not accomplish this?

YaManicKill: So, if we have:

* unread-marker
* date-marker
* date-marker
* message

We want to end up with 

* unread-marker
* date-marker
* message

Correct?

I'm pretty sure this code will not do that, because it will set the first to be unread-marker, then it will check that both 0 and 1 are date-markers, but they aren't. So it will return false. This will then stop it doing the job, I think.

Am I missing something?

xPaw: That's correct. And you might be right the code is flawed. Any ideas on how to improve it?

YaManicKill: Simplest thing I can think of is changing the first one.

Currently we do:

```
if (!prev) {
 	// Should always be a date-seperator, because it's always added
 	prev = $(this);
} else {
...
```

So, if we want to make sure that prev is always a date-marker, we can just do this:

```
if (!prev) {
	// Should always be a date-seperator, because it's always added
	if ($(this).hasClass("date-marker")) {
	 	prev = $(this);
	}
} else {
...
```

Then you don't need to add this extra check for the date-marker on the previous one here, as we will always have a date marker on this one.

YaManicKill:
Do we need a second review of this one, or is it small enough to go with 1?

astorije:
@xPaw is on a roll!! 👏

PR #824, opened 2016-12-19
Use css tooltips on time elements, ability to view time on mobile

### LINKED ISSUE #761: Be able to tap time on mobile to view the full date

xPaw:
Fixes #761

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1763,12 +1763,8 @@ kbd {
 	  * - https://www.w3.org/TR/mediaqueries-4/
 	  * - https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover
 	  */
-	.tooltipped:hover:before,
-	.tooltipped:hover:after,
-	.tooltipped:active:before,
-	.tooltipped:active:after,
-	.tooltipped:focus:before,
-	.tooltipped:focus:after {
+	.tooltipped-no-touch:hover:before,
+	.tooltipped-no-touch:hover:after {
 		visibility: hidden;
 		opacity: 0;
 	}
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -839,7 +839,6 @@ kbd {
 	word-wrap: break-word;
 	word-break: break-word; /* Webkit-specific */
 	display: flex;
-	overflow: hidden;
 	position: relative;
 }
 
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1519,9 +1519,10 @@ kbd {
 }
 
 /**
-  * Tooltips
+  * Tooltips v0.5.3
   * See http://primercss.io/tooltips/
   */
+
 .tooltipped {
 	position: relative;
 }
@@ -1529,12 +1530,11 @@ kbd {
 .tooltipped:after {
 	position: absolute;
 	z-index: 1000000;
-	display: inline-block;
-	visibility: hidden;
-	opacity: 0;
+	display: none;
 	padding: 5px 8px;
 	font: 12px Lato;
 	line-height: 1.2;
+	-webkit-font-smoothing: subpixel-antialiased;
 	color: #fff;
 	text-align: center;
 	text-decoration: none;
@@ -1547,23 +1547,40 @@ kbd {
 	content: attr(aria-label);
 	background: #222;
 	border-radius: 3px;
-	-webkit-font-smoothing: subpixel-antialiased;
-	transition: .2s;
+	opacity: 0;
 }
 
 .tooltipped:before {
 	position: absolute;
 	z-index: 1000001;
-	display: inline-block;
-	visibility: hidden;
-	opacity: 0;
+	display: none;
 	width: 0;
 	height: 0;
 	color: #fff;
 	pointer-events: none;
 	content: "";
 	border: 5px solid transparent;
-	transition: .2s;
+	opacity: 0;
+}
+
+@-webkit-keyframes tooltip-appear {
+	from {
+		opacity: 0;
+	}
+
+	to {
+		opacity: 1;
+	}
+}
+
+@keyframes tooltip-appear {
+	from {
+		opacity: 0;
+	}
+
+	to {
+		opacity: 1;
+	}
 }
 
 .tooltipped:hover:before,
@@ -1572,9 +1589,18 @@ kbd {
 .tooltipped:active:after,
 .tooltipped:focus:before,
 .tooltipped:focus:after {
-	visibility: visible;
-	opacity: 1;
+	display: inline-block;
 	text-decoration: none;
+	-webkit-animation-name: tooltip-appear;
+	animation-name: tooltip-appear;
+	-webkit-animation-duration: .1s;
+	animation-duration: .1s;
+	-webkit-animation-fill-mode: forwards;
+	animation-fill-mode: forwards;
+	-webkit-animation-timing-function: ease-in;
+	animation-timing-function: ease-in;
+	-webkit-animation-delay: .4s;
+	animation-delay: .4s;
 }
 
 .tooltipped-s:after,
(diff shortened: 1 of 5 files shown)


astorije:
@xPaw, ping?

astorije:
@xPaw, what do you think we should do with this? At the moment it's way too epileptic...

xPaw:
We can experiment with timed tooltips for sure.

astorije:
Hey @xPaw, I have "bumped" our version of Tooltips to the latest one, v0.5.3, which is a 1 stone / 2 birds situation since the timed tooltips are the default ones coming from stock lib.

Only changes I made were the colors + our custom Stylelint config (like `:after` instead of `::after` sadly).

Playing with this on my laptop and mobile, I think it works nicely like that.

Only issue I could not figure out is the following:

screen shot 2017-04-04 at 01 22 56

(Obviously the first "Change nick" is fake, it was to better compare)
For a reason I could not find, border radius and padding are not respected on the timestamp tooltips. Any idea?

client/css/style.css
```diff
@@ -1757,12 +1850,8 @@ kbd {
...
-	.tooltipped:hover:before,
-	.tooltipped:hover:after,
-	.tooltipped:active:before,
-	.tooltipped:active:after,
-	.tooltipped:focus:before,
-	.tooltipped:focus:after {
+	.tooltipped-no-touch:hover:before,
+	.tooltipped-no-touch:hover:after {
 		visibility: hidden;
 		opacity: 0;
```
astorije: Not sure if this needs updating and/or is completely necessary anymore. I just know it works as expected on my mobile at the moment...

xPaw:
@astorije tooltips are cut (no border radius) because `.msg` has `overflow: hidden`.

client/css/style.css
```diff
@@ -1672,6 +1696,17 @@ kbd {
 	border-right-color: #222;
 }
 
+@media only screen and (-webkit-min-device-pixel-ratio: 2),
```
xPaw: @astorije Do we need this?

client/css/style.css
xPaw: Yeah we probably no longer need to disable tooltips on mobile as they are on a delay (?)

astorije:
> reduced the tooltips patch.

Aaaargh, why did you get rid of my commit, @xPaw??? I spent quite some time on it to make sure we wouldn't have to do it again, and made the commit purposedly separate so it made sense to keep it and not squash it and lose history of it.

astorije:
This is the full-size of it, coming directly from the lib. We really *really* should keep it that way:

```diff
commit 7b34216c72d4458e127c41cd2b1055e0958c93f9
Author: Jérémie Astori <jeremie@astori.fr>
Date:   Tue Apr 4 01:27:13 2017 -0400

    Update Primer tooltips to latest v0.5.3

    This:

    - Makes tooltips appear after timer instead of instantly, necessary for timestamp tooltips (see https://github.com/thelounge/lounge/pull/824#pullrequestreview-13676231)
    - Uses Primer default animation (not sure if .2s transition was ours or theirs but here it is)
    - Goes closer to default tooltips which will help to bump future versions and/or to streamline this in build process

─────────────────────────────────────────
modified: client/css/style.css
─────────────────────────────────────────
@ style.css:1516 @ kbd {
 }

 /**
-  * Tooltips
+  * Tooltips v0.5.3
   * See http://primercss.io/tooltips/
   */
+
 .tooltipped {
    position: relative;
 }
@ style.css:1527 @ kbd {
 .tooltipped:after {
    position: absolute;
    z-index: 1000000;
-   display: inline-block;
-   visibility: hidden;
-   opacity: 0;
+   display: none;
    padding: 5px 8px;
    font: 12px Lato;
    line-height: 1.2;
+   -webkit-font-smoothing: subpixel-antialiased;
    color: #fff;
    text-align: center;
    text-decoration: none;
@ style.css:1544 @ kbd {
    content: attr(aria-label);
    background: #222;
    border-radius: 3px;
-   -webkit-font-smoothing: subpixel-antialiased;
-   transition: .2s;
+   opacity: 0;
 }

 .tooltipped:before {
    position: absolute;
    z-index: 1000001;
-   display: inline-block;
-   visibility: hidden;
-   opacity: 0;
+   display: none;
    width: 0;
    height: 0;
    color: #fff;
    pointer-events: none;
    content: "";
    border: 5px solid transparent;
-   transition: .2s;
+   opacity: 0;
+}
+
+@-webkit-keyframes tooltip-appear {
+   from {
+       opacity: 0;
+   }
+
+   to {
+       opacity: 1;
+   }
+}
+
+@keyframes tooltip-appear {
+   from {
+       opacity: 0;
+   }
+
+   to {
+       opacity: 1;
+   }
 }

 .tooltipped:hover:before,
@ style.css:1586 @ kbd {
 .tooltipped:active:after,
 .tooltipped:focus:before,
 .tooltipped:focus:after {
-   visibility: visible;
-   opacity: 1;
+   display: inline-block;
    text-decoration: none;
+   -webkit-animation-name: tooltip-appear;
+   animation-name: tooltip-appear;
+   -webkit-animation-duration: .1s;
+   animation-duration: .1s;
+   -webkit-animation-fill-mode: forwards;
+   animation-fill-mode: forwards;
+   -webkit-animation-timing-function: ease-in;
+   animation-timing-function: ease-in;
+   -webkit-animation-delay: .4s;
+   animation-delay: .4s;
+}
+
+.tooltipped-no-delay:hover:before,
+.tooltipped-no-delay:hover:after,
+.tooltipped-no-delay:active:before,
+.tooltipped-no-delay:active:after,
+.tooltipped-no-delay:focus:before,
+.tooltipped-no-delay:focus:after {
+   opacity: 1;
+   -webkit-animation: none;
+   animation: none;
+}
+
+.tooltipped-multiline:hover:after,
+.tooltipped-multiline:active:after,
+.tooltipped-multiline:focus:after {
+   display: table-cell;
 }

 .tooltipped-s:after,
@ style.css:1711 @ kbd {
    border-right-color: #222;
 }

+.tooltipped-multiline:after {
+   width: -webkit-max-content;
+   width: -moz-max-content;
+   width: max-content;
+   max-width: 250px;
+   word-break: break-word;
+   word-wrap: normal;
+   white-space: pre-line;
+   border-collapse: separate;
+}
+
+.tooltipped-multiline.tooltipped-s:after,
+.tooltipped-multiline.tooltipped-n:after {
+   right: auto;
+   left: 50%;
+   -webkit-transform: translateX(-50%);
+   transform: translateX(-50%);
+}
+
+.tooltipped-multiline.tooltipped-w:after,
+.tooltipped-multiline.tooltipped-e:after {
+   right: 100%;
+}
+
+@media screen and (min-width: 0\0) {
+   .tooltipped-multiline:after {
+       width: 250px;
+   }
+}
+
+.tooltipped-sticky:before,
+.tooltipped-sticky:after {
+   display: inline-block;
+}
+
+.tooltipped-sticky.tooltipped-multiline:after {
+   display: table-cell;
+}
+
+@media only screen and (-webkit-min-device-pixel-ratio: 2),
+   only screen and (min--moz-device-pixel-ratio: 2),
+   only screen and (-moz-min-device-pixel-ratio: 2),
+   only screen and (min-device-pixel-ratio: 2),
+   only screen and (min-resolution: 192dpi),
+   only screen and (min-resolution: 2dppx) {
+   .tooltipped-w:after {
+       margin-right: 4.5px;
+   }
+}
+
 /* End tooltips */
```

xPaw:
We don't gain anything from including the extra styles, they were removed initially too.

astorije:
> We don't gain anything from including the extra styles

Yes we do, as documented in the commit message you got rid of. Benefit is I don't have to spend another hour doing that same 💩 job next time they bump their lib, or investigating why this is not working as expected because we got rid of something we thought was not needed.

Can I re-add my commit, *please*?

xPaw:
> Benefit is I don't have to spend another hour doing that same 💩 job next time they bump their lib, or investigating why this is not working as expected because we got rid of something we thought was not needed.

They haven't updated it since they released it. What I got rid of is `multiline` and `sticky` variants because we don't use either of these (and most likely ever won't). I can add the comment about the delay.

client/css/style.css
astorije: This comes from [here](https://github.com/primer/primer-tooltips/blob/a8b5f34d4f5e0d019005d4c97610e19a0e528152/lib/tooltips.scss#L276-L280). I'd keep it as long as it's in the upstream project.

client/css/style.css
astorije: Let's keep these at the moment, see how things go at next release or something with the new tooltips, and then we could remove this on mobile. At least for a future PR, this is good to go.

astorije:
I was actually afraid removing the `overflow: hidden` would make us subject to zalgo text but testing it out, it seems like no.

PR #829, opened 2016-12-20
Increase IRC colors contrast

### LINKED ISSUE #335: Update some IRC colors to be readable in all themes

xPaw:
Old | New
-- | --
before | after

Closes #335

### DIFF
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -114,6 +114,11 @@ body {
 	opacity: 1;
 }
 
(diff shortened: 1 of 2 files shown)


astorije:
Minor fix, merging.

PR #832, opened 2016-12-21
Remind channel name or nick in input placeholder

### LINKED ISSUE #343: Channel name placeholder in input
This was suggested in the irc:

```
15:42   Parabola requests that my name next to the editbox be replaced with the current channel name
15:42   Parabola    seems like seeing my own name is pointless anyway
15:43   gweedo767   channel name could be more help in the mobile view too
15:44   gweedo767   since the channel list isn't visible then
```

I quite like this idea. Thoughts?

astorije:
Fixes #343.

Result on small device:

screen shot 2016-12-20 at 20 11 38 screen shot 2016-12-20 at 20 42 49

(Content is different only because I messed up my screenshots, it's 2 different windows 😅)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1035,6 +1035,12 @@ $(function() {
 		}
 		document.title = title;
 
+		var placeholder = "";
+		if (chan.data("type") === "channel" || chan.data("type") === "query") {
+			placeholder = "Write to " + chan.data("title") + "...";
+		}
+		input.attr("placeholder", placeholder);
+
 		if (self.hasClass("chan")) {
 			$("#chat-container").addClass("active");
 			setNick(self.closest(".network").data("nick"));
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -46,11 +46,6 @@ body {
 	padding: 15px 16px;
 }
 
-#chat .search::-webkit-input-placeholder {
-	color: #99a2b4;
-	opacity: .5;
-}
-
 /* Borders */
 #chat .from,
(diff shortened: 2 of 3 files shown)


MaxLeiter:
I think this is unnecessary clutter on the desktop - maybe hide it there?

astorije:
Nah, I think it's just as valuable on desktop than mobile. I was kinda not fond of the idea at first, but you get used to it in 5 seconds and it's actually not bad to have it there.

Slack added this fairly recently too, pretty sure some people complained about accidentally messaging the wrong coworkers :D

astorije:
@MaxLeiter, once this is enabled, if you really don't like it, you can easily hide it with `#input::-webkit-input-placeholder { opacity: 0; }` (or `visibility: hidden`), using the prefix you need, in your custom CSS.
It's not possible for someone to display this otherwise though.

MaxLeiter:
That's fine, but sidenote: I really think we should avoid a "you can use custom CSS if you don't like it" mentality, because we shouldn't assume the average user understands CSS (or even knows that box exists).

Can't wait for tabbed-page settings (or some equivalent)...

astorije:
> but sidenote: I really think we should avoid a "you can use custom CSS if you don't like it" mentality, because we shouldn't assume the average user understands CSS

I completely agree with you on this, we should make an app that's optimal and well-thought in the first place, and leave that custom CSS to those of us who have very specific and sometimes weird requirements.
In this case however, I'm with @xPaw and @YaManicKill, I think it makes for a better user experience.

PR #836, opened 2016-12-21
Calculate menu width on touch start

### LINKED ISSUE #831: menuWidth incorrectly calculated when changing sidebar width with themes
https://github.com/thelounge/lounge/blob/6887b66f675cf116e17e8ebe696cf6f7d975a39c/client/js/libs/slideout.js#L8

This is computed before the theme is loaded, so sliding out the menu results in wrong width.

xPaw:
Fixes #831


astorije:
Alright, calling this one a minor fix, merging.

PR #838, opened 2016-12-23
Add web server tests

xPaw:
Who doesn't like more tests?

### DIFF
--- a/test/fixtures/env.js
+++ b/test/fixtures/env.js
@@ -1,4 +1,11 @@
 "use strict";
 
+global.log = {
+	error: () => console.error.apply(console, arguments),
+	warn: () => {},
+	info: () => {},
+	debug: () => {},
+};
+
 var home = require("path").join(__dirname, ".lounge");
 require("../../src/helper").setHome(home);
--- a/test/server.js
+++ b/test/server.js
@@ -1,0 +1,100 @@
+"use strict";
+
+const Helper = require("../src/helper");
+const expect = require("chai").expect;
+const request = require("request");
+const io = require("socket.io-client");
+
+describe("Server", () => {
+	const server = require("../src/server");
+	server();
+
+	const webURL = `http://${Helper.config.host}:${Helper.config.port}/`;
+
+	describe("Express", () => {
+		it("should run a web server on " + webURL, done => {
+			request(webURL, (error, response, body) => {
+				expect(error).to.be.null;
+				expect(body).to.include("<title>The Lounge</title>");
+				expect(body).to.include("https://thelounge.github.io/");
+
+				done();
+			});
+		});
+
+		it("should serve static content correctly", done => {
+			request(webURL + "manifest.json", (error, response, body) => {
+				expect(error).to.be.null;
+
+				body = JSON.parse(body);
+				expect(body.name).to.equal("The Lounge");
+
+				done();
+			});
+		});
+	});
+
+	describe("WebSockets", () => {
+		let client;
+
+		beforeEach(() => {
+			client = io(webURL, {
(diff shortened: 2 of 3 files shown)


lenovouser:
Not sure, but you might want to move the `socket.io-client` dependency to `devDependencies` since it is not necessary for the normal installation.

package.json
```diff
@@ -55,6 +55,7 @@
     "request": "2.79.0",
     "semver": "5.3.0",
     "socket.io": "1.7.2",
+    "socket.io-client": "1.7.2",
```
xPaw: `socket.io` requires `socket.io-client` on it's own as dependency, so there is no extra download or anything.

test/fixtures/env.js
```diff
@@ -1,4 +1,13 @@
 "use strict";
 
+global.log = {
+	error: function() {
+		console.error.apply(console, arguments);
+	},
+	warn: function() {},
+	info: function() {},
+	debug: function() {},
```
xPaw: Hides the unnecessary logging when running test.

test/server.js
```diff
@@ -0,0 +1,75 @@
...
+
+describe("Server", function() {
+	const server = require("../src/server");
+	server();
+
+	const webURL = `http://${Helper.config.host}:${Helper.config.port}/`;
+
+	describe("Express", function() {
+		it("should run a web server on " + webURL, function(done) {
+			request(webURL, function(error, response, body) {
```
xPaw: Not really, to be honest.

xPaw:
> how feasible would it be to start an IRC server as part of the CI build so we could have some end-to-end tests in there?

Very feasible, but I was thinking of making a listening socket and sending raw IRC stuff myself, since we don't need a fully fledged stateful server.

package.json
astorije: Under Node v4 / npm v2 (I don't know when they changed that), packages are not all at the root level, so this will cause to have one directory in `node_modules/` and one in `node_modules/socket.io/node_modules/`.

But more generally, any reason why not putting the dep in `devDependencies` if we (we being `thelounge` package, i.e. excluding what `socket.io` does) don't require it directly?
Do you know if there is a way to not need it at all, like `require("socket.io/client")` or something like that in your test? That would be optimal as it would use the same piece of code we actually serve.

test/fixtures/env.js
astorije: Right, sorry, I meant to suggest the arrow function format.

test/server.js
astorije: Fair enough :) I'll re-evaluate this if/when we need more async tests.

astorije:
> Very feasible, but I was thinking of making a listening socket and sending raw IRC stuff myself, since we don't need a fully fledged stateful server.

Not bad either, good idea. This is good for unit tests and will be much faster anyway.

But at some point it would be nice to have a big end-to-end session that does clicks and inputs and talks to an actual IRC server. Then we can try with maybe the 3-4 most used IRCds out there and make sure the scenario works every time. But we're far from here, and your idea is already good for unit tests, which we lack as well.

test/fixtures/env.js
```diff
@@ -1,4 +1,11 @@
 "use strict";
 
+global.log = {
+	error: () => console.error.apply(console, arguments),
+	warn: () => {},
+	info: () => {},
+	debug: () => {},
+};
+
```
astorije: I've seen this is usually done via a `LOG_LEVEL` env var or the like, but I'm fine either way.

xPaw:
I'm going to merge this with 1 review, as this is just adding tests.

PR #839, opened 2016-12-23
add codecov.io

Bonuspunkt:
code coverage + badge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,3 +33,7 @@ deploy:
     condition: "$BUILD_ENV = production"
     tags: true
     repo: thelounge/lounge
+
+after_success:
+    - npm run coverage
+    - bash <(curl -s https://codecov.io/bash)


YaManicKill:
So, first of all, it doesn't seem to work yet.

```
==> Travis CI detected.
    project root: .
    Yaml not found, that's ok! Learn more at http://docs.codecov.io/docs/codecov-yaml
==> Running gcov in . (disable via -X gcov)
==> Searching for coverage reports in:
    + .
==> Python coveragepy not found
--> No coverage report found.
    Please visit http://docs.codecov.io/docs/supported-languages
```

But let's discuss code coverage.

In general, I'm not a massive fan of it because it gives us a number to try and hit, which ends up making people focus on the quantity and not the quality.

We need to improve test coverage, we all know we do. But I don't think this is the way to do it. It'll also become easier once we have webpack, we can test things much better.

@thelounge/maintainers any opinions?

xPaw:
Need a `npm run coverage`

MaxLeiter:
Personally, I'm against more badges, but I think code coverage is important. I think we should slowly add tests overtime when working on PRs related to the code being tested (astorije explained this well in #796)

Bonuspunkt:
well tbh its just a stupid number.
- it doesn't show if you have validated any results.
- it doesn't show if you have validated all edge cases.
but you can see if your test hit all branches and it supports you long term
imho :)

astorije:
The actual metric is usually useful with libraries, as it gives you a sense of trust if you're considering using it as a dependency.
For a end-user application actively maintained, this has less meaning. For example, if we start adding end-to-end tests that actually interact with the app, there won't be a way to assess coverage. Same with frontend tests, it's likely that a single metric won't include that.

That being said, it's useful information for us developers of the app, and that's why we have `npm run coverage`. It creates a coverage report in HTML that one can open on a browser locally and that gives the same level of data than services like codecov or coveralls.

Lastly, considering our current level of coverage, I don't think there is anything we can brag about :)

So to summarize, I'm not sure we want to include this, or at least not yet. Happy to change my mind if other @thelounge/maintainers strongly feel otherwise. 

Thanks for your help and interest though, @Bonuspunkt!

xPaw:
I think we can report to codecov or coveralls, but disable the auto comments. We can create a README with the badge inside the test folder, so it's not visible in main readme, but visible where it matters.

astorije:
> I think we can report to codecov, but disable the auto comments. We can create a README with the badge inside the test folder, so it's not visible in main readme, but visible where it matters.

We sure can but... for what benefits? Codecov or Coveralls is another service that can fail, another potential source of breaking our Travis builds, etc.
Just to be clear, I have nothing against them, I have used Coveralls a lot in the past and still heavily rely on it in a few projects, but I'm not sure what it brings to us here. If you open the HTML report generated by `npm run coverage`, it has everything we need! For an end-user project like The Lounge, I don't think it is useful at all to have this.

I'd be happy to change my mind if there are compelling arguments / benefits to make this available  to the non-developers, but if there aren't, I don't think we should add this just because we can 😃.

astorije:
@xPaw, what's your final thought on this? :)

xPaw:
Code coverage tool should not be failing our builds. Thats most likely configurable.

Bonuspunkt:
should be at https://codecov.io/gh/thelounge/lounge/settings/yaml

xPaw:
I think we can merge this if the badge from readme is removed.

astorije:
I honestly don't care for it but `¯\_(ツ)_/¯` I'll go with the flow.
Yes, remove badge from README and also make sure this is run only on a single build, like the Node v7 one. It's arbitrary but let's make it match the version we use to release, to keep things simple.

xPaw:
Looks like none of the maintainers are fond of displaying test coverage right now. Should we close this?

astorije:
Yeah, there is very little value about it at the moment. We can definitely re-evaluate this later when we get our 💩 together around testing.

PR #844, opened 2016-12-28
Switched client pop to mp3, closes #512

### LINKED ISSUE #512: The audio for the notification doesn't play on safari
I think this is because Safari doesn't support .ogg files
ref: http://stackoverflow.com/questions/10583827/ogg-audio-in-safari

Converting the file from .ogg to .wav and changing the path to the file should fix it

MaxLeiter:
mp3 is more universally supported - see https://en.wikipedia.org/wiki/HTML5_Audio#Supported_audio_coding_formats


astorije:
MP3 might be more supported, but it is also not an open format, and the restrictions apply to us. I am not 100% sure, but it might not be the case for AAC. While AAC codec developers require a patent, it is free to use (unlike MP3).

If this happens to not be true, it seems like OGG defaulting to WAV is our best option, granted it will not work for IE.

astorije:
2 interesting links: http://www.html5gamedevs.com/topic/7095-best-audio-format-to-use-mp3-wav-or-ogg/ and https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats#Browser_compatibility

MaxLeiter:
Is there really a benefit to shipping ogg and wav? Why not just use wav and save shipping an extra file?

astorije:
Yeah I'm dum-dum. Let me rephrase: OGG with fallback to AAC if there are any licensing issues (it looks like there shouldn't be) for maximum compatibility, or only WAV otherwise and we acknowledge no support for IE.

xPaw:
> OGG with fallback to AAC

That seems to be the way to go from looking at caniuse.com.

PR #845, opened 2016-12-28
Use away-notify to show updates on users away state

MaxLeiter:
Probably need better wording for the settings option. Handles duplicate states (i.e. if someone goes away multiple times in a row it will only be shown once).

image

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -225,8 +225,8 @@
 							<div class="col-sm-12">
 								<h2>
 									Status messages
-									<span class="tooltipped tooltipped-n tooltipped-no-delay" aria-label="Joins, parts, kicks, nick changes, and mode changes">
-										<button class="extra-help" aria-label="Joins, parts, kicks, nick changes, and mode changes"></button>
+									<span class="tooltipped tooltipped-n tooltipped-no-delay" aria-label="Joins, parts, kicks, nick changes, away changes, and mode changes">
+										<button class="extra-help" aria-label="Joins, parts, kicks, nick changes, away changes, and mode changes"></button>
 									</span>
 								</h2>
 							</div>
--- a/client/views/actions/away.tpl
+++ b/client/views/actions/away.tpl
@@ -1,0 +1,3 @@
+{{> ../user_name nick=from}}
+is away
+<i class="away-message">({{{parse text}}})</i>
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -7,6 +7,7 @@ module.exports = User;
 function User(attr, prefixLookup) {
 	_.defaults(this, attr, {
 		modes: [],
+		away: "",
 		mode: "",
 		nick: "",
 		lastMessage: 0,
--- a/src/plugins/irc-events/away.js
+++ b/src/plugins/irc-events/away.js
@@ -1,0 +1,30 @@
+"use strict";
+
+const _ = require("lodash");
+const Msg = require("../../models/msg");
+
+module.exports = function(irc, network) {
+	const client = this;
+	irc.on("away", (data) => {
+		const away = data.message;
+
+		network.channels.forEach((chan) => {
+			const user = _.find(chan.users, {nick: data.nick});
+
+			if (!user || user.away === away) {
+				return;
+			}
+
+			const msg = new Msg({
+				type: away ? Msg.Type.AWAY : Msg.Type.BACK,
+				text: away || "",
+				time: data.time,
+				from: data.nick,
+				mode: user.mode
+			});
+
+			chan.pushMessage(client, msg);
+			user.away = away;
+		});
+	});
+};
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -208,6 +208,8 @@ kbd {
 #settings .extra-help,
 #settings #play::before,
 #form #submit::before,
+#chat .away .from::before,
+#chat .back .from::before,
 #chat .invite .from::before,
 #chat .join .from::before,
 #chat .kick .from::before,
@@ -257,6 +259,12 @@ kbd {
 #footer .sign-out::before { content: "\f011"; /* http://fontawesome.io/icon/power-off/ */ }
 
 #form #submit::before { content: "\f1d8"; /* http://fontawesome.io/icon/paper-plane/ */ }
+
+#chat .away .from::before,
+#chat .back .from::before {
+	content: "\f017"; /* http://fontawesome.io/icon/clock-o/ */
+	color: #7f8c8d;
+}
 
 #chat .invite .from::before {
 	content: "\f003"; /* http://fontawesome.io/icon/envelope-o/ */
@@ -1157,6 +1165,8 @@ kbd {
 }
 
 #chat .condensed .content,
+#chat .away .content,
+#chat .back .content,
 #chat .join .content,
 #chat .kick .content,
 #chat .mode .content,
--- a/client/js/condensed.js
+++ b/client/js/condensed.js
@@ -23,6 +23,12 @@ function updateText(condensed, addedTypes) {
 	constants.condensedTypes.forEach((type) => {
 		if (obj[type]) {
 			switch (type) {
+			case "away":
+				strings.push(obj[type] + (obj[type] > 1 ? " users have gone away" : " user has gone away"));
+				break;
+			case "back":
+				strings.push(obj[type] + (obj[type] > 1 ? " users have come back" : " user has come back"));
+				break;
 			case "join":
 				strings.push(obj[type] + (obj[type] > 1 ? " users have joined the channel" : " user has joined the channel"));
 				break;
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -60,6 +60,8 @@ const commands = [
 ];
 
 const actionTypes = [
+	"away",
+	"back",
 	"ban_list",
 	"invite",
 	"join",
@@ -77,6 +79,8 @@ const actionTypes = [
 ];
 
 const condensedTypes = [
+	"away",
+	"back",
 	"join",
 	"part",
 	"quit",
--- a/client/views/actions/back.tpl
+++ b/client/views/actions/back.tpl
@@ -1,0 +1,2 @@
+{{> ../user_name nick=from}}
+is back
--- a/client/views/index.js
+++ b/client/views/index.js
@@ -3,6 +3,8 @@
 module.exports = {
 	actions: {
 		action: require("./actions/action.tpl"),
+		away: require("./actions/away.tpl"),
+		back: require("./actions/back.tpl"),
 		ban_list: require("./actions/ban_list.tpl"),
 		channel_list: require("./actions/channel_list.tpl"),
 		ctcp: require("./actions/ctcp.tpl"),
--- a/src/client.js
+++ b/src/client.js
@@ -16,6 +16,7 @@ module.exports = Client;
 
 var id = 0;
 var events = [
+	"away",
 	"connection",
 	"unhandled",
 	"banlist",
--- a/src/models/msg.js
+++ b/src/models/msg.js
@@ -29,7 +29,9 @@ class Msg {
 
 Msg.Type = {
 	UNHANDLED: "unhandled",
+	AWAY: "away",
 	ACTION: "action",
+	BACK: "back",
 	ERROR: "error",
 	INVITE: "invite",
 	JOIN: "join",


client/views/actions/away.tpl
```diff
@@ -0,0 +1,7 @@
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{mode}}{{from}}</span>
+{{#if text}}
+is now away
+<i class="away-message">({{{parse text}}})</i>
+{{else}}
+is back
```
MaxLeiter: is now back?

client/index.html
```diff
@@ -243,6 +243,12 @@ <h1 class="title">Settings</h1>
 									Show quits
 								</label>
 							</div>
+							<div class="col-sm-6">
+								<label class="opt">
+									<input type="checkbox" name="away">
+									Show aways/backs
```
MaxLeiter: Must be a better way to word this

src/models/user.js
```diff
@@ -16,4 +17,5 @@ function User(attr, prefixLookup) {
 	// TODO: Remove this
 	this.name = this.nick;
 	this.mode = (this.modes && this.modes[0]) || "";
+	this.away = this.away;
```
MaxLeiter: Not necessary, just added for uniformity... should remove

client/index.html
dgw: Ideas: "Show when users go away", "Show away notifications", "Notify on user presence changes", "Show user presence changes".

client/views/actions/away.tpl
```diff
@@ -0,0 +1,7 @@
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{mode}}{{from}}</span>
+{{#if text}}
+is now away
+<i class="away-message">({{{parse text}}})</i>
+{{else}}
+is now back
```
dgw: I'm partial to wording it as "went away" and "came back", but whatever.

astorije:
No matter what I try, multiple clients and users running this PR on 3 different IRC servers (including Freenode and the demo server), I cannot get to have anything showing up. Using `scripts/run-pr.sh` so the webpacking is done correctly.
Any idea? Any chance this PR could be currently broken? :/

client/index.html
```diff
@@ -243,6 +243,12 @@ <h1 class="title">Settings</h1>
 									Show quits
 								</label>
 							</div>
+							<div class="col-sm-6">
+								<label class="opt">
+									<input type="checkbox" name="away">
+									Show when users go away
```
astorije: As opposed to @dgw's opinion, I think we should go with consistency, therefore something like `Show away/back changes` or the likes. It's not pretty, but it sticks out compared with the other options right now.

client/views/actions/away.tpl
```diff
@@ -0,0 +1,7 @@
+<span role="button" class="user {{colorClass from}}" data-name="{{from}}">{{mode}}{{from}}</span>
+{{#if text}}
+went away
+<i class="away-message">({{{parse text}}})</i>
+{{else}}
+came back
+{{/if}}
```
astorije: `/nick` shows present tense (`x is now known as y`) so we could just go with `x is away` / `x is back` for consistency.

Also, I believe on The Lounge `/away` with no reason sets an away message of ` ` (one whitespace) so I think the message should now show up at all if trimmed `text` is empty.

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+
+let _ = require("lodash");
+let Msg = require("../../models/msg");
```
astorije: `const` instead.

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,36 @@
...
+
+module.exports = function(irc, network) {
+	let client = this;
+	irc.on("away", function(data) {
```
astorije: Arrow function? :)

MaxLeiter:
No idea why not working, will investigate. Regarding duplicate messages, people can't change their nick to the same nick, part twice, join twice, etc (without actions in between)

client/views/actions/away.tpl
MaxLeiter: Is there an easy way to tell what 'text' contains?

astorije: I think you would have to define a helper for that which would do a stupid check like `text.trim().length === 0`. Maybe @xPaw would have a better idea?

xPaw: Do we trim other messages like kicks? If not, don't do it here.

astorije: I'm not saying we should trim `text`. I'm saying the check to know if there is a text should take this into account. This is because when using The Lounge, sending `/away` [will send a `" "` reason](https://github.com/thelounge/lounge/blob/0d8b58425eff9cc2a9f7468fcc678f8b2e1e7156/src/plugins/inputs/away.js#L7) so anyone else using The Lounge would see `nick is now away ( )`.
In such situation, it should not display the parens.

MaxLeiter:
Addressed all changes except the empty parenthesis if no away message is supplied (@astorije)

MaxLeiter:
@astorije even if these messages condensed, I still don't see a reason to show multiple away's from the same user - how about they're hidden if the message is the same, and shown if not?

astorije:
Testing this on testnet.inspircd.org using 2 accounts, I cannot see status change when using `/away` with no argument or when using `/back`.
Using `/away foobar` does show `astorije is away (foobar)` correctly however.

astorije:
@MaxLeiter, did you see my last message?

dgw:
That's a mess of commits…

YaManicKill:
So, what's the current state of this? Is it still not working for @astorije entirely?

astorije:
Right, it wouldn't work at all on Freenode (but I believe this is normal as not supported there) and not working properly on that other network.
@MaxLeiter, do you think you can/want to work on this before v2.3.0 or should we remove it from the milestone? :(

MaxLeiter:
Ill get it done

astorije:
@MaxLeiter, ping on this? Would be really nice to put this behind us!

astorije:
@MaxLeiter, any chance you could fix conflicts / wrap up this PR? :)

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+
+const _ = require("lodash");
+const Msg = require("../../models/msg");
+
+module.exports = function(irc, network) {
+	let client = this;
```
astorije: To fix ESLint, use `const` here.

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,36 @@
...
+	irc.on("away", (data) => {
+		let away = false;
+		if (data.message) {
+			away = true;
+		}
+		network.channels.forEach((chan, index) => {
+			let user = _.find(chan.users, {name: data.nick});
```
astorije: To fix ESLint, use `const` here.

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,36 @@
...
+			if (typeof user === "undefined") {
+				return;
+			}
+			if (user.away === away) {
+				return;
+			}
+
+			let msg = new Msg({
```
astorije: To fix ESLint, use `const` here.

MaxLeiter:
@xPaw if you could take a look at this that'd be great -- I'm lost. Basically, the empty `/away` command should emit `/away <space>`, however -somewhere- that " " is being stripped so it's being received by the IRC server as `/away`, which is really `/back`.

xPaw:
@MaxLeiter are you talking about you sending `/away` not receiving it? If so, why is that a problem for you in this particular PR?

MaxLeiter:
I guess so, @xPaw, because for some reason `/away` no longer works with this PR.

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,42 @@
...
+			const msg = new Msg({
+				type: Msg.Type.AWAY,
+				text: away,
+				time: data.time,
+				from: data.nick,
+				mode: chan.getMode(data.nick)
+			});
+
+			chan.pushMessage(client, msg);
+			if (index === network.channels.length - 1) { // Last iteration, let's update the users status
```
xPaw: User object is a child of channel, you have to update it for every channel.

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,42 @@
...
+
+const _ = require("lodash");
+const Msg = require("../../models/msg");
+
+module.exports = function(irc, network) {
+	const client = this;
+	irc.on("away", (data) => {
+		let away = null;
+
+		if (data.message) {
```
xPaw: If `away.message` is an empty string here and you do want undefined:
```
const away = data.message || undefined;
```
otherwise just doing `const away = data.message;` is enough.

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		network.channels.forEach((chan, index) => {
+			const user = _.find(chan.users, {nick: data.nick});
+
+			if (typeof user === "undefined") {
+				return;
+			}
+
+			if (user.away === away) {
```
xPaw: You can combine the ifs. `if (!user || user.away === away)`

src/plugins/inputs/away.js
```diff
@@ -7,7 +7,6 @@ exports.input = function(network, chan, cmd, args) {
 
 	if (cmd === "away") {
 		reason = args.length > 0 ? args.join(" ") : " ";
-
```
xPaw: unnecessary change

MaxLeiter:
Did some testing, PR is ready to go 👍

astorije:
@xPaw, can you give the first review/approval to this please? Thanks!

xPaw:
@astorije The code is fine, but I really don't know if we want to allow showing away changes like this.

astorije:
Why is that? I think it's at least as valuable as nick changes, so I'm def okay with that personally.

client/index.html
```diff
@@ -258,6 +258,12 @@ <h1 class="title">Settings</h1>
 									Show seconds in timestamp
 								</label>
 							</div>
+							<div class="col-sm-6">
+								<label class="opt">
+									<input type="checkbox" name="away">
+									Show away/back changes
```
YaManicKill: Should this be condensed when we do condensed messages? I reckon it should be just included like the rest of the join/part stuff.

MaxLeiter:
Fine by me -- I'll rebase this once that's merged

astorije:
@MaxLeiter, now that condensed stuff is out, do you think you'd be able to pick this up?
If not, @YaManicKill, do you think you could rebase/fix conflicts for this? As far as I can tell, all conflicts here are consequences of #759.

YaManicKill:
The conflicts don't look that difficult, so if @MaxLeiter doesn't have time to do that, I'd be happy to fix the conflicts.

astorije:
@YaManicKill, @MaxLeiter mentioned he wouldn't be able to take this up, so I assigned you to it, whenever you have a moment :)

client/index.html
MaxLeiter: Condensed -- do we still want this option?

MaxLeiter:
Rebased + updated to condense. Ready for review
image

Recommend testing on testnet.inspircd.org

client/index.html
astorije: I don't think so. This is a status message like any others, I think. So something like "X users went away, Y users came back" and 💥, part of the condensed stuff.

astorije: Just noticed that's what you actually did so don't mind me 👍

MaxLeiter:
Don't think we can control that, @astorije. I could make it show up but it wouldn't be too pretty -- I don't think it's necessary.

astorije:
I think setting yourself as away/back should show, but since it would have nothing to do with away-notify, don't worry about it on this PR.

Glad we're actually going to merge this thing!! 😅

astorije:
@YaManicKill or @xPaw, could one of you review this? :)

xPaw:
@MaxLeiter Can you also mention away messages in help tooltip in the setting for status messages?

astorije:
@MaxLeiter, could you squash?

MaxLeiter:
Done

src/plugins/irc-events/away.js
```diff
@@ -0,0 +1,30 @@
...
+			if (!user || user.away === away) {
+				return;
+			}
+
+			const msg = new Msg({
+				type: away ? Msg.Type.AWAY : Msg.Type.BACK,
+				text: away || "",
+				time: data.time,
+				from: data.nick,
+				mode: chan.getMode(data.nick)
```
xPaw: You can do `user.mode` here, as you've already found the user.

astorije:
Will merge once you address @xPaw's suggestion :)

MaxLeiter:
Done

PR #847, opened 2016-12-28
Move linter and build configs to scripts folder

xPaw:
This tidies up our root folder as much as possible (other files can't be moved as far as I'm aware).


scripts/istanbul.yml
```diff
@@ -1,5 +1,6 @@
 instrumentation:
-  include-all-sources: true
+# TODO: disabled all sources as istanbul crashes on `import` token in client code
+#  include-all-sources: true
```
astorije: Seeing this made me create #850 for a fix, btw. Not perfect, but at least `npm run coverage` runs.

.npmignore
```diff
@@ -3,17 +3,15 @@
 # See https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
 
 client/js/bundle.vendor.js.map
+client/js/lounge.js
+client/js/libs/
```
astorije: This will not break source maps?
Either way I'm not sure if we should ignore them or not, but I don't have a compelling argument.

package.json
```diff
@@ -12,18 +12,18 @@
...
-    "watch": "webpack -w",
+    "build:webpack": "webpack --config scripts/webpack.config.js ",
+    "watch": "webpack --config scripts/webpack.config.js -w",
     "test": "npm-run-all -c test:* lint",
     "test:mocha": "mocha",
     "lint": "npm-run-all -c lint:*",
-    "lint:js": "eslint .",
-    "lint:css": "stylelint \"**/*.css\"",
+    "lint:js": "eslint --config scripts/eslint.yml --ignore-path scripts/eslintignore.txt .",
+    "lint:css": "stylelint --config scripts/stylelint.json \"**/*.css\"",
```
astorije: I'm not too sure about all this, really.
While it's often possible to move these around, relying on default locations/filenames whenever possible is usually a sane idea and not doing so can result in asking for trouble.

At the very least, default locations/filenames helps following conventions, and also helps newcomers get comfortable with what they're getting into.

xPaw: Now that you mention it, linters in IDEs will indeed look in root. What about nyc/webpack though?

astorije: Very good point. I would say that if we move the Webpack config file, it should actually go into the `client/` dir. Then there is only the config file for nyc left, and at that point it would be weird to have just this one in `scripts` :/

astorije:
I think the only thing we want here is moving the Webpack config into the `client/` dir.

xPaw:
We sure can do that, but will be easier to make a new PR later.

PR #853, opened 2016-12-29
you have to require("babel-register") before require("testee")

Bonuspunkt:
you have to `require("babel-register")` before `require("testee")`

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "babel-core": "6.21.0",
     "babel-loader": "6.2.10",
     "babel-preset-es2015": "6.18.0",
+    "babel-register": "6.18.0",
     "chai": "3.5.0",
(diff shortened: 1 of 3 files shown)


Bonuspunkt:
to discuss if you want to keep `import` or switch back to `require` :)

Bonuspunkt:
also polyfills for es2016/7 on node4 / client?

astorije:
@xPaw, is this something we want? I'm still learning my ways with Webpack, but this seems a bit peculiar. I'd like to have your 2c.

Also, @Bonuspunkt, if this ends up going in `master`, you'll want a better commit message, as-is I can't tell what your intent is 😅.

xPaw:
I think we should just use `require()`

PR #854, opened 2016-12-29
Fix stuff that breaks in jQuery 3

xPaw:
For the eventual upgrade.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -300,10 +300,6 @@ $(function() {
 			text.html(templates.actions[type](data.msg));
 		}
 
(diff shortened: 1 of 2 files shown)


client/js/lounge.js
```diff
@@ -300,10 +300,6 @@ $(function() {
 			text.html(templates.actions[type](data.msg));
 		}
 
-		if (text.find("i").size() === 1) {
-			text = text.find("i");
```
xPaw: Shout's parser had some unnecessary `<i>` wrapping which we removed. And this code isn't used anymore.

Issue #851, opened 2016-12-29
User list search only matches from beginning of the line


xPaw:
If you type `test`, it will not find `_test` user.

Issue #66, opened 2016-02-18
A proper structure for the client javascript code


xPaw:
Currently most of the code is in a single file with a lot of anonymous functions. Anon functions make it harder to debug/view stacktraces when analyzing stuff. Having a neat structure where each element is in a separate namespace/scope would be preferred.

astorije:
I reckon @YaManicKill was interested into making this a reality too: https://github.com/erming/shout/issues/612

YaManicKill:
Yeah, I would like to modularise it. @MaxLeiter's work on #62 is going to help that a lot, as the structure will already be there, and we can then just abstract parts out and organise the code in a way that makes it really easy to find specific features (for example, it's really hard to find where ... say the settings code is).

nornagon:
I'm interested in this :) See #610

astorije:
Now that Webpack has landed, can we discuss the structure we want for the client code here before making any drastic change?

Any suggestion?

It seems very appealing to split per component (message input, chat window, user list, user search, etc.) but that's very high level. How do we want to apply this in practice?

YaManicKill:
I think splitting based on functionality is a good way to do things. This would kinda work with what you are doing, but also allows us to split out other things similarly (settings can be it's own, one for the new code I've written for message condensing etc). But it's not as simple as define exactly how we split things out. It's hard to generalise it without actually just doing it. 

However, I don't think we should do a big bang approach of "remove all the things into other files" because that'll get nasty.

Instead, we should move code out as we change it. For example, next person to do the settings code could move that out to a module, (which is a better way to deal with settings as it means it's easier to change out how we are saving the settings without changing other code, etc, etc).

Does that make sense?

astorije:
> Does that make sense?

Definitely, but I would like to discuss the structure (like, folder structure) here before someone actually univocally decides to do it.

YaManicKill:
> I would like to discuss the structure (like, folder structure) here before someone actually univocally decides to do it.

Sure, well yeah, if you want an example of what I was suggesting, I would make module structure for some of it like this:

> settings.js
> input.js
> contextMenu (this makes sense to split into different ones depending on which you are clicking on, maybe with an index.js that would include some common code between them? Not sure yet)
  > userMenu.js
  > chanMenu.js
> channelList (I reckon we'd have a couple for this, because there are a number of things to do with it)
  > display.js
  > sorting.js

Does that make sense as a structure?

MaxLeiter:
[...]

PR #856, opened 2016-12-30
Implement fuzzy-matching for the user list

astorije:
I have been wishing for something like this for a while, then wanted to look at it when #851 was opened.
It is a fuzzy-finder for the user list, where matching parts of the results are being emphasized. It is the same mechanism than fuzzy-finders on GitHub, Atom, etc.

This PR is in a horrible state, partly because of recent addition of Webpack, partly because of how the user list is managed (state entirely in DOM, in a "rendered" way, i.e. mode and nick are concatenated and source data is not kept around).

I added `fuzzy.js` in this PR only because for the love of me I'm having so much trouble with Webpack! It's of course not meant to stay, but until then it's there for who wants to try. Help welcome if anyone wants to save me from my own stupidity!

## TODO

- [x] Use package from npm
- [x] Highly improve the code, ugly for now. Happy to discuss how to improve it (I personally would love to have a global state outside the DOM that gets used to render the list, but maybe I'm dreaming too much)
- [ ] Potentially improve performance. It's probably not efficient at all right now, but I didn't find any slowness on `#freenode` (1.5k users). Maybe more of an issue on mobile.
- [ ] Write tests for it. This relies on #66 so it might come later.
- [x] Have CI builds pass
- [ ] Squash commits

Note that this PR is not meant to solve other issues, such as #544.

## Result

fuzzy_userlist

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1068,7 +1068,7 @@ $(function() {
 	});
 
 	chat.on("input", ".search", function() {
-		const value = $(this).val().toLowerCase();
+		const value = $(this).val();
 		const names = $(this).closest(".users").find(".names");
 
 		names.find(".user").each((i, el) => {
@@ -1078,7 +1078,7 @@ $(function() {
 		const fuzzyOptions = {
 			pre: "<b>",
 			post: "</b>",
-			extract: el => $(el).text().toLowerCase().replace(/[+%@~]/, "")
+			extract: el => $(el).text()
 		};
 
 		fuzzy.filter(
@@ -1086,8 +1086,7 @@ $(function() {
 			names.find(".user").toArray(),
 			fuzzyOptions
 		).forEach(el => {
-			const firstChar = $(el.original).text()[0].replace(/[^+%@~]/, "");
-			$(el.original).html(firstChar + el.string).show();
+			$(el.original).html(el.string).show();
 		});
 	});
 
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1069,11 +1069,15 @@ $(function() {
 
 	chat.on("input", ".search", function() {
 		const value = $(this).val();
-		const names = $(this).closest(".users").find(".names");
+		const parent = $(this).closest(".users");
+		const names = parent.find(".names-original");
+		const container = parent.find(".names-filtered");
 
-		names.find(".user").each((i, el) => {
-			$(el).text($(el).text().replace(/<\/?b>;/, "")).hide();
-		});
+		if (!value.length) {
+			container.hide();
+			names.show();
+			return;
+		}
 
 		const fuzzyOptions = {
 			pre: "<b>",
@@ -1081,13 +1085,14 @@ $(function() {
 			extract: el => $(el).text()
 		};
 
-		fuzzy.filter(
+		const result = fuzzy.filter(
 			value,
 			names.find(".user").toArray(),
 			fuzzyOptions
-		).forEach(el => {
-			$(el.original).html(el.string).show();
-		});
+		);
+
+		names.hide();
+		container.html(templates.user_filtered({matches: result})).show();
 	});
 
 	chat.on("msg", ".messages", function(e, target, msg) {
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -258,7 +258,10 @@ $(function() {
 
 	function renderChannel(data) {
 		renderChannelMessages(data);
-		renderChannelUsers(data);
+
+		if (data.type === "channel") {
+			renderChannelUsers(data);
+		}
 	}
 
 	function renderChannelMessages(data) {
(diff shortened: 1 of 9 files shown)


MaxLeiter:
Could concatinating mode be helpful so you could search by mode?

client/js/lounge.js
```diff
@@ -1107,16 +1108,26 @@ $(function() {
...
+		const names = $(this).closest(".users").find(".names");
+
+		names.find(".user").each((i, el) => {
+			$(el).text($(el).text().replace(/<\/?b>;/, "")).hide();
+		});
+
+		const fuzzyOptions = {
+			pre: "<b>",
+			post: "</b>",
+			extract: el => $(el).text().toLowerCase().replace(/[+%@~]/, "")
```
xPaw: Do you really need the `replace` when you're fuzzy searching?

Also `fuzzy` has `caseSensitive`, which does not appear to be on by default.

client/js/lounge.js
```diff
@@ -1,6 +1,7 @@
 // vendor libraries
 import "jquery-ui/ui/widgets/sortable";
 import $ from "jquery";
+import "fuzzy";
```
xPaw: Should this be `import fuzzy from "fuzzy";`?

client/js/lounge.js
astorije: > Do you really need the `replace` when you're fuzzy searching?

Nope, removed.

> Also `fuzzy` has `caseSensitive`, which does not appear to be on by default.

Ack, removed the lowercasing.

client/js/lounge.js
astorije: I did try that, and at the time I couldn't get it to work. Now, it seems like it works.
Maybe I didn't test correctly, or maybe now is not 100% what it was during my first attempt, I don't know, but now it seems OK.

astorije:
> Could concatinating mode be helpful so you could search by mode?

Yup, thanks, it's now possible (or rather, trimming mode is now disabled).

client/js/lounge.js
```diff
@@ -1107,16 +1108,25 @@ $(function() {
...
+			post: "</b>",
+			extract: el => $(el).text()
+		};
+
+		fuzzy.filter(
+			value,
+			names.find(".user").toArray(),
+			fuzzyOptions
+		).forEach(el => {
+			$(el.original).html(el.string).show();
```
xPaw: That's a lot of HTML shuffling imo, what if we had a separate container for results and rendered it once? This would also render #890 unnecessary.

package.json
```diff
@@ -47,15 +47,16 @@
     "event-stream": "3.3.4",
     "express": "4.14.0",
     "fs-extra": "1.0.0",
+    "fuzzy": "0.1.3",
     "irc-framework": "2.5.3",
+    "ldapjs": "1.0.1",
```
xPaw: I keep seeing a lot of PRs affecting this package due to bad sorting, I'll make a separate PR to fix ite.

xPaw:
I've pushed code to use separate container for search results. I believe it's going to be buggy when user list refreshes while you have a filter applied.

xPaw:
I think this PR is rendered unnecessary by #856.

xPaw:
I've implemented a fix to correctly update user lists while searching (updates search results too!). Also removed some sidebar hackery (only renders sidebar in channel type). This should be good to go now.

astorije:
So this is what it looks like at the moment:

search_results

I'm not entirely sure I like the "merging" of all sections when search is being used. What do other @thelounge/maintainers think?
@xPaw, how much effort do you think it would be to keep search results in their respective sections?

YaManicKill:
> What do other @thelounge/maintainers think?

I can see both sides. I can understand wanting users to be in the section because that's where you see them normally, but I can also see putting them all in 1 list so you don't have to guess which section they are in (less of an issue when you type more letters). So  ¯\_(ツ)_/¯

astorije:
> but I can also see putting them all in 1 list so you don't have to guess which section they are in

Your comment is generally fair but this hit the spot. I think that's a reasonable thought, but it just felt super weird the first time I used it. I'm wondering if it's not being used to the current way that drew my attention.
I'm interested in other opinions as well (@MaxLeiter?) but now that you pointed a decent use case, I can get along with that move. I'll review this a bit more tomorrow.

MaxLeiter:
I think we should keep sections - keeps it uniform (i.e. sections are displayed when not searching). Channels don't generally have a ton of OPs/voiced/etc users, so sifting through them isn't too much of a pain.

astorije:
@MaxLeiter, the plan is not to remove sections altogether, @xPaw's changes just has a single one called "Search results" when typing in the search input. I think it's okay for now (maybe even better if we sort by nick without the mode, but that can go in later).

MaxLeiter:
@astorije i meant keep sections when searching (like search now), but not a big deal

Issue #737, opened 2016-11-16
Npm Prepublish will be changing


YaManicKill:
So, with regards to #722, we do need to do **something** about it.

As mentioned, here is my suggestion:

* Add prepare to do the same as prepublish

This should mean that npm 1-4 will work with prepublish, while npm 5 (when released in Q1 2017) will work with prepare (having prepublish as well will mean that it is run twice on publish, but it's just building, so it'll just overwrite).

As far as I can see, there should be no side effects of this. Am I missing anything?

xPaw:
Here's what I think we should do:
1. Change prepublish` to `prepublishOnly`, only @astorije makes releases, so not a big problem.
2. Explicitly document running `npm run build` after installing, as the new `npm prepare` doesn't run in 100% cases after `npm install` (e.g. when using root). #725.

astorije:
https://github.com/thelounge/lounge/pull/858 resolves that issue.

> Change `prepublish` to `prepublishOnly`, only @astorije makes releases, so not a big problem.

Actually, only Travis does it. #858 now uses `prepublishOnly`, but that means deploying using v7 which is not a LTS, as it's the first version with npm v4. I'm OK with that, just an FYI.

> Possibly add a compilation marker (e.g. in lounge.js) and confirm it when starting the server to tell users to build the stuff (to alert users that they can possibly be running outdated scripts or not have them at all). We've had multiple issues already where users forgot to build and thought why the client was broken.

Agreed, too many people end up thinking their app is broken. #858 does not go as far as a compilation marker, I'm not sure how to do this reliably. It only checks that `client/js/bundle.js` exists (not that it's outdate), but that better than nothing.

YaManicKill:
> but that means deploying using v7 which is not a LTS, as it's the first version with npm v4

Not necessarily, you can update npm separately from node.

astorije:
> Not necessarily, you can update npm separately from node.

I'd rather avoid messing up with versions on CI at deployment time.
I think #858 does things nicely. What do you think, @YaManicKill?

YaManicKill:
> I think #858 does things nicely. What do you think, @YaManicKill?

Yeah, looks fine for this. We will still need to change again once we get to npm 5, and change it back to "prepublish" as that'll replace "prepublishonly". But that's a while away, we probably won't get it till node 9, I expect, and we don't need to support it until node 10, because that'll be the first stable release with npm 5. So it's over a year away. But we'll keep an eye on it.

[...]

Issue #857, opened 2016-12-30
Front-end develop/debug experience is much worse than pre-Webpack era


astorije:
I was trying to work on #856 and kept bashing my head against the wall: many things that were possible and helping code or debug are now gone.

Granted, part of my frustration is caused by my current lack of understanding of Webpack/Babel and the like, but I am going to only list here the technical issues that I encountered.
I'm happy to RTFM, but at the very least it should be as easy to code/debug/test as it used to be.

- [x] `debugger` statements are being dropped when running Webpack. This is due to `UglifyJsPlugin` which drops them by default.
  Current workarounds include adding `drop_debugger: false` to the `compress` object in the Webpack config or commenting out the plugin locally. What do we want to do in repo? Both options are meh, but first one is less meh since ESLint ensures `debugger` statements do not end up in `master`. Is there a better solution?

- [x] When reaching breakpoints, it is now not possible to evaluate code using the browser REPL. To reproduce, add `var foo = "bar"; debugger;` anywhere, run webpack, open the browser devtools and make sure to reach the breakpoint. When evaluating `foo`, browser says `VM72:ReferenceError: foo is not defined`.
  I did not find an alternative yet and this is driving me crazy 😅

- [ ] ~~Likely a symptom of the same issue, but~~ none of the globals are available anymore. This means it is not possible to open a browser console, select an element and evaluate `$($0)` anymore because `$` is undefined<sup>1</sup>.
  Temporary workaround I have found is to add `window.$ = $;` in `lounge.js` to have access to `$`. We obviously will not be able to ship this in production, which means that it will be hell to debug a production instance after next release if not fixed (hence `priority` label).

- [ ] Sometimes `npm run start-dev` loses it and will not rebuild anymore. This forces a whole server restart, which is annoying as we lose history and such. Not sure what to do here, I ended up restarting the server every time so far...

---

<sup>1</sup> `$0` simply references the element currently selected by the inspector

astorije:
[This](https://webpack.github.io/docs/list-of-hints.html#developer-performance) has some recommendations:

> Use webpack devtools for debugging in browser: Better debugging experience, real source code, real module names, at the cost of slower build and difference to production build (Developer, for every app)

But this does not solve the 3rd point (and yay, more complexity...).

astorije:
> I would consider this actually a win for us, because we no longer pollute globals, and thus the entire client code is contained and less likely to be tampered with.

[...]

PR #858, opened 2016-12-30
Do not uglify builds when running start-dev

xPaw:
Tweaks bundle paths (`client` vs `client/js`) to not confuse sourcemaps as much. Disables uglifyjs plugin when running with `--watch`. Also reduces production built file size a little more.


webpack.config.js
```diff
@@ -2,11 +2,12 @@
 
 const webpack = require("webpack");
 const path = require("path");
+const isWatch = process.argv.includes("--watch");
```
astorije: Should we be relying on `NODE_ENV` instead?

package.json
```diff
@@ -17,8 +17,8 @@
     "start-dev": "npm-run-all --parallel watch start",
     "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",
-    "build:webpack": "webpack",
-    "watch": "webpack -w",
+    "build:webpack": "webpack -p",
+    "watch": "webpack -d --watch",
```
astorije: For whoever is wondering, `node_modules/.bin/webpack -h` says:

```
  -d    shortcut for --debug --devtool sourcemap --output-pathinfo
  -p    shortcut for --optimize-minimize
```

I see this as unofficial development/production flags, am I correct?
If so, it's a bit confusing that `build:webpack` implies production and `watch` implies development. I know I sometimes use `npm run build:webpack` when I just want a single build without a FS watch.

IMO (but MO is fairly Webpack-inexperienced 😅), I think we should have the commands be environment-independent, and the configuration files behave differently depending on the environment (or select different files depending on environment, if the configurations are too disjoint).
Is that stupid?

astorije:
Playing a bit with the current state, we definitely cannot have configurations differ solely on `build:webpack` vs. `watch`. When you check out a branch and run `npm install`, this (at least until #737 gets addressed) this assumes prod.

I think our best bet is to rely on `NODE_ENV` to be `production` (which is also set when using `--production`).

xPaw:
I wanted to use env at first, but setting it in script and making it crossplatform is not easy.

astorije:
I don't think we have a choice though. Could use https://www.npmjs.com/package/cross-env for cross-platform in `scripts`?

xPaw:
I think we can do a two liner js to set the environment.

astorije:
You mean to emulate `cross-env`?

Anyway, it doesn't really matter I think, since it's mostly `NODE_ENV=production npm install` which will be useful. The rest should assume default `NODE_ENV` or use the one being set.

webpack.config.js
```diff
@@ -53,14 +54,18 @@ module.exports = {
...
-			compress: {
-				warnings: false
-			}
-		}),
+		new webpack.optimize.DedupePlugin(),
+		new webpack.optimize.CommonsChunkPlugin("js/bundle.vendor.js"),
 	]
 };
+
+if (!isWatch) {
```
ddrmanxbxfr: You could use the node_env directly from here if you wanted to filter it from production and development build.

```
process.env.NODE_ENV
```

ddrmanxbxfr:
Using the process.env.NODE_ENV to determine what to build or not is how most of the webpack-built project does. I left a comment in the code if we want to include the uglify plugin only when in production (NODE_ENV == PRODUCTION)...

.travis.yml
```diff
@@ -6,11 +6,17 @@ node_js:
 
 matrix:
   fast_finish: true
+  include:
+    - node_js: 7 # Version used to deploy to npm registry
+      env: BUILD_ENV=production
```
astorije: This only adds one extra job to the build (which still runs all its jobs concurrently, so build time does not increase) to make sure what we deploy or will is actually correct.

.travis.yml
```diff
@@ -6,11 +6,17 @@ node_js:
...
 
 cache:
   directories:
     - ~/.npm
 
+before_script:
+  - NODE_ENV=$BUILD_ENV npm run build
+
```
astorije: This works fine when `BUILD_ENV` is not defined, which is the case outside of that `include`d combination.

astorije:
Alright, spent a looot of time on this, between reading docs and GitHub threads for Webpack, npm, etc. I believe this PR puts us in an optimal state regarding recent introduction of Webpack.

This PR resolves the first 2 items of https://github.com/thelounge/lounge/issues/857#issue-198109751.

Posting here my commit message details + extra info:

#### 1. Use Webpack configuration based on NODE_ENV instead of watch/no-watch

Also, move the `DedupePlugin` to the prod-specific section. [Webpack doc](https://webpack.github.io/docs/list-of-plugins.html#dedupeplugin) itself recommends to not run this outside of production.

~~Note that this currently breaks cross-OS support of `npm run build`. This will be fixed in a latter commit.~~ This was temporary and on purpose to not mix the unrelated commits of Webpack-for-production and no-`npm run build`-on-`npm install`. It's fixed by the next commit.

This commit addresses my comments made in this PR. Non-cross-OS character of this commit is the reason why the next commits are part of this PR.

#### 2. Make and document `npm run build` now a mandatory step of the install/build setup

This has several reasons, benefits and consequences:

- When running on root (which is not recommended anyway), `npm run build` was already necessary.
- This allows to not use the `prepublish` hook, whose behavior is going to change in npm v5 and again in npm v6.
- This allows to create both production and development builds when running from source.
- It makes `npm run build` compatible with Windows again for development environments (lost in previous commit).
- It uses the `prepublishOnly` hook added in npm v4. Since this hook is not available prior to that, deployment to npm from Travis has to be done on the Node.js v7 environment.

This fixes https://github.com/thelounge/lounge/issues/737.

#### 3. Add an extra Travis CI job to make sure production build succeeds and passes tests

This only adds one extra job to the build (which still runs all its jobs concurrently, so build time does not increase) to make sure what we deploy or will is actually correct.

This is highly desirable since we now produce 2 different files in development/testing and production.

#### 4. Add a basic check for bundled application when starting the server

Note that this will not detect if the client application was built with an old version of the repo.

This partially addresses https://github.com/thelounge/lounge/issues/737#issuecomment-261702678, but a naive check is better than no checks!

package.json
```diff
@@ -18,13 +18,13 @@
...
     "build:webpack": "webpack",
-    "watch": "webpack -w",
+    "watch": "webpack --watch",
     "test": "npm-run-all -c test:* lint",
     "test:mocha": "mocha",
     "lint": "npm-run-all -c lint:*",
     "lint:js": "eslint .",
     "lint:css": "stylelint \"**/*.css\"",
-    "prepublish": "npm run build"
+    "prepublishOnly": "NODE_ENV=production npm run build"
```
xPaw: This won't execute on Windows, fyi.

astorije:
Yeah I know but it's OK as it's only relevant when publishing, which we
only do from Travis builds.
If comes a time when we don't publish through Travis as a one-off, chances
are I'd publish myself (not Windows) and if not, one can tweak to use
Windows' `set` for this one-off.
All other commands, for day to day development, work on all platforms.

src/server.js
```diff
@@ -18,6 +18,11 @@ var authFunction = localAuth;
 module.exports = function() {
 	manager = new ClientManager();
 
+	if (!fs.existsSync("client/js/bundle.js")) {
```
xPaw: Ideas on how we would check hashes to tell the file is out of date?

webpack.config.js
```diff
@@ -17,8 +21,8 @@ module.exports = {
 	},
 	devtool: "source-map",
 	output: {
-		path: path.resolve(__dirname, "client/js"),
-		filename: "bundle.js",
+		path: path.resolve(__dirname, "client"),
+		filename: "[name]",
```
ddrmanxbxfr: Usually with webpack, you'll be using the compilation hash that can easily be appended at build.

You would configure webpack to output like so inside the webpack config...

```
module.exports = {
...
	output: {
		path: path.join(__dirname, 'dist'),
		pathinfo: (env !== 'production'),
		filename: 'app-[hash:6].js'
       }
...
}
```

This will output a app-rand6c.js (app- 6 random characters as i've requested 6). filename in the output folder.

From there you could include it easily in the server.js

src/server.js
astorije: Solutions I can think of are not ideal. For example, I don't know if this is possible without forcing people committing on the client to include such marker in their PR.

Either way, can we do this in a latter PR please? This PR should not go further than what it currently does.

astorije:
@thelounge/maintainers, yes, no, anyone? :)
v2.2.0 is sooo close... !

YaManicKill:
@astorije Sorry, didn't realise you guys had stopped messing around with this PR :-P I sort of phased out the notifications. I'll take a look now.

.travis.yml
YaManicKill: So, if it isn't defined, does it just ignore that, and default to dev?

package.json
YaManicKill: That should be fine, because we should only ever be publishing from travis.

src/server.js
YaManicKill: Yeah, let's look into this in another one. I guess you could do it by the first line of the bundle being a version number or git hash  or something.

xPaw:
Fyi, building from git will most likely not minify (as you have to explicitly set node env to prod). The changes are fine for now.

.travis.yml
```diff
@@ -6,11 +6,17 @@ node_js:
 
 matrix:
   fast_finish: true
+  include:
+    - node_js: 7 # Version used to deploy to npm registry
+      env: BUILD_ENV=production
```
xPaw: Can't set node env directly?

README.md
```diff
@@ -50,14 +50,11 @@ The following commands install the development version of The Lounge:
...
+NODE_ENV=production npm run build
 npm start
 ```
 
-A word of caution:
-
-- While it is the most recent codebase, this is not production-ready!
-- It is not recommended to run this as root. However, if you decide to do so,
-  you will have to run `npm run build`.
+⚠️ Run at your own risk. It is also not recommended to run this as root.
```
xPaw: Keep the mention that all development happens in master and it may not be stable.

src/server.js
xPaw: Also, I would probably move this block into index.js. I'm not sure it belongs in server.

.travis.yml
astorije: No, otherwise `devDependencies` aren't installed. The rationale is: run `npm install` in non-production environment to be able to build, but build for production environments at release time.

.travis.yml
astorije: Correct, if `BUILD_ENV` isn't defined, it runs `NODE_ENV= npm run build` which is similar as setting an empty environment (i.e. a not production).

README.md
astorije: Will do.

package.json
astorije: What @YaManicKill said. You can still run `npm run build` on Windows, and if someone ever needs to deploy from Windows, which should only happen in case something goes wrong on Travis CI, one can always change `prepublishOnly` so a Windows-friendly command temporarily.

src/server.js
astorije: Ok, I'll move it.

astorije: So, I've tried to move it to `index.js` and I'd definitely keep it where it is: `index.js` doesn't have access to `log` at the moment (probably not a big deal but still) but way more importantly, it's blocking `-h`, the `list` subcommand, etc.

Let's keep it that way for now and we can improve that later.

astorije:
Alright, I believe (well, I hope 😅) this one is finally good to go. @xPaw, are you happy with it?

xPaw:
:+1:

YaManicKill:
Well, I'll merge, then.

PR #868, opened 2017-01-04
Make log style when referring user consistent with other places

astorije:
Same as https://github.com/thelounge/lounge/pull/779. Not a fix per se, but super simple (made from GitHub editor). Feel free to merge with only 1 👍.


YaManicKill:
Merging with 1 because it's pretty simple.

Issue #843, opened 2016-12-28
Filling in username and then nick will override username


xPaw:
This piece of code produces very poor UX. https://github.com/thelounge/lounge/blob/6466680a8efc2f18608ae0204b027cbff60048cf/client/js/lounge.js#L1279-L1282

astorije:
There should be an extra check: override the username value only if it is the same as the nick one.

astorije:
Fixed by #873.

PR #873, opened 2017-01-09
Fix filling in the nickname, overriding the username (in add network tab)

PolarizedIons:
Fixes issue #843

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1283,9 +1283,27 @@ $(function() {
 		);
 	});
 
+	forms.on("focusin", ".nick", function() {
+		// Need to set the first "lastvalue", so it can be used in the below function
+		var nick = $(this);
+		nick.data("lastvalue", nick.val());
+	});
+
 	forms.on("input", ".nick", function() {
 		var nick = $(this).val();
-		forms.find(".username").val(nick);
+		var usernameInput = forms.find(".username");
+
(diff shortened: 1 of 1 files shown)


MaxLeiter:
Tests are failing due to some trailing spaces, @PolarizedIons 

https://ci.appveyor.com/project/astorije/lounge/build/1569#L410

PolarizedIons:
Oh whoops, I could have sworn I checked for that. Thank you @MaxLeiter - fixed now

PolarizedIons:
ping?

astorije:
@xPaw, can you think of a less convoluted way to do so? This is a fairly simple feature, we shouldn't need 2 handlers and 20+ lines of code for it IMO :/

xPaw:
@astorije While it does seem silly, the code quite makes sense.

client/js/lounge.js
```diff
@@ -1283,9 +1283,29 @@ $(function() {
 		);
 	});
 
+	forms.on("focusin", ".nick", function() {
+		// Need to set the first "lastvalue", so it can be used in the below function
+		var nick = $(this);
+		if (nick.data("lastvalue") === undefined) {
```
xPaw: What does this check accomplish?

PolarizedIons: I can't seem to find a reason for it either, probably something that got left over after I rewrote it. Thanks for the catch, removed it

PR #884, opened 2017-01-20
Add humans.txt file

Raqbit:
Added a static [humans.txt](http://humanstxt.org/) file that mentions the contributors and maintainers of thelounge.
Also added an author `<link>` in index.html, pointing to this file.

Got maintainer lines from [here](https://github.com/thelounge/lounge/wiki/Maintainers'-corner#maintainers-and-responsibilities).

### DIFF
--- a/client/humans.txt
+++ b/client/humans.txt
@@ -1,0 +1,19 @@
+
+/* Thanks to everyone who contributed to the lounge project! */
+
+https://github.com/thelounge/lounge/graphs/contributors
+
+
+/* Maintainers */
+
+* @astorije: UX, (future) package system, overall overview of the entire project, releases
+* @maxpoulin64 (Max-P on the IRC channel): Performance, bugs
+* @xPaw: IRC core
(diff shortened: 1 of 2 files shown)


MaxLeiter:
I think, if it doesn't include every contributor, it should just have the four primary maintainers + @erming - astorije, YaManicKill, Max-P, and xPaw. That being said, I'm not too thrilled about shipping another root-directory file

Raqbit:
Good point, it just came to my mind after I saw web-starter-kit have a humans.txt included. Of course it's up to debate whom to include in the file, I just wanted to have something in there to show what it's it's for.

xPaw:
:-1: Don't think we need this. Lounge is a collaborative effort. We could define maintainers in package.json though.

YaManicKill:
I agree, this feels unnecessary. We have a link to the github in the url, and we can see a list of everyone who contributes there. We don't want to add something we will just have to keep updated.

PR #889, opened 2017-01-25
Un-ellipsis placeholder and use ES6 template

astorije:
I'm not sure what I was thinking in https://github.com/thelounge/lounge/pull/832 but I am really annoyed by the looks of it in practice.
Plus that way there is a bit of extra room in case screen isn't wide enough. Very unlikely, but if it happens, that would be a few missing chars at best, so why not.

Also hey, first ES6 feature on the client right there.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1040,7 +1040,7 @@ $(function() {
 
 		var placeholder = "";
 		if (chan.data("type") === "channel" || chan.data("type") === "query") {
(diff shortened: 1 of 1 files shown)


astorije:
If reviewer is happy with it, feel free to merge and add to v2.2.0 as it's a minor improvement over a feature introduced in v2.2.0. Heck, this PR doesn't even deserve a change log entry if it makes it to v2.2.0.

PR #896, opened 2017-01-28
Sort depedencies in package.json

xPaw:
Minor fix so that other PRs don't keep touching it.


astorije:
Very straight-forward and side-effect-free, merging!

PR #903, opened 2017-01-29
Prompt admin for user log at user creation

astorije:
I feel like more and more people are asking where to get channel logs on the IRC channel. We usually tell them to go enable the logs for individual users, but at that point it's already too late as logging was not enabled.

~~This ensures new users have logs enabled by default when using `lounge add <username>`. What do people think? I also feel this should be a prompt after asking for password (`Enable user logging? (yes)`). Happy to hear opinions.~~

This prompts admin about it, which defaults to being enabled. The prompt is rather forgiving, as `yes`, `y`, `Y`, etc. will be evaluated to `true` and everything else to `false`. It is fairly standard in CLIs to really care about the first letter when it's a yes/no type of question.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -82,7 +82,7 @@ ClientManager.prototype.getUsers = function() {
 	return users;
 };
 
-ClientManager.prototype.addUser = function(name, password) {
+ClientManager.prototype.addUser = function(name, password, enableLog) {
 	var users = this.getUsers();
 	if (users.indexOf(name) !== -1) {
 		return false;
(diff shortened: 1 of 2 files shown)


dgw:
I'd say enabling anything that will silently consume disk space by default is a bad idea. Prompting during user creation works, though. Maybe `Save user's chat logs to disk? (yes)` to make it clear that disk space will be used.

astorije:
I agree with all that. Thanks for your input, @dgw, I'll do that after releasing v2.2.0 (incomiiiiing!).

astorije:
@dgw, done!

That syntax is not the cleanest thing one can do (a prompt in a prompt) but considering the overall cleanliness of our CLI code, I think that's as good as it gets.
I'm sure when prompts become more complex (n levels of callbacks), we will have a chat about how to refactor it, but in the meantime, that should do.

astorije:
Would one of @thelounge/maintainers be able to review this? It's a pretty simple change so I think it could go with a patch release instead of waiting for a bigger minor release.

PR #913, opened 2017-02-06
Fix body height

### LINKED ISSUE #793: Bar appears on bottom of screen on mobile Chrome Canary and Beta
Accessing the lounge from Android with Chrome Beta or Chrome Canary produces the following visual bug, where a bar appears at the bottom of the screen:

YaManicKill:
#821 fixed #793, but it appears since then, Chrome has changed the behaviour back to what it was befor ethe fix, so 2.2.0 on Chrome 56+ is even more broken.

Only needs 1 reviewer.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -31,7 +31,7 @@
 
 html,
 body {
-	height: 100vh;
+	height: 100%;
 }
 
 body {


xPaw:
Can you attach a screenshot how it's currently broken?

YaManicKill:
These are from PK43 in the irc, but I can confirm this is what it showed like for me.

As add to homescreen: 

In Chrome: 

As mentioned before, it's only on lounge 2.2.0 and Chrome 56+.

Issue #925, opened 2017-02-16
CLI options are ignored when given in a specific order


nhooyr:
at https://thelounge.github.io/docs/getting_started/usage.html under start it says

```
lounge start --port 80 --private
```

that should be

```
lounge start -- --port 80 --private
```

or

```
lounge start --private --port 80
```

(edit: for the last one I meant `lounge --private --port 80`)

astorije:
This is an error with The Lounge and not the docs, unfortunately.

Could you specify how you installed The Lounge (through `npm install -g thelounge`? From source using `git clone && npm install`? Using one of our OS packages?), as well as the actual behavior please?

Thanks!

nhooyr:
I used nix. I created a nix package based off of the latest npm package (like just a few hours ago).

astorije:
FYI, I could reproduce this on v2.2.1. However, none of the following worked:

```sh
lounge start --port 80 --private
lounge start --private --port 80
```

The alternative using `--` worked.
https://github.com/thelounge/lounge/pull/799 should have fixed that issue but it looks like it didn't fix it completely (because this has different root cause).

nhooyr:
sorry for the last one I meant `lounge --private --port 80`

astorije:
Wow, indeed, that works and that's even more messed up as `lounge` without any subcommand shouldn't do anything.

Definitely a lot of work needed on the CLI! If someone passes by and wants to contribute, lots of room for improvement!

xPaw:
> lounge without any subcommand shouldn't do anything

https://github.com/thelounge/lounge/blob/00a3e3f14c01b2e8bd596af8f0bf5d102e091df7/src/command-line/index.js#L45-L47

astorije:
Good point! We should have `lounge` == `lounge --help` instead of `lounge` == `lounge start` when this bug gets fixed though!

msaun008:
I did a little digging into [commander](https://github.com/tj/commander.js) and saw that the lounge is only defining the ```options``` and the ```action``` callback function.  The lounge is not actually parsing ```process.argv```. 

Created PR https://github.com/thelounge/lounge/pull/929.

PR #929, opened 2017-02-18
Update arg parsing and default 'lounge' to 'lounge --help'

msaun008:
This PR does 2 things: 
1. Updates ```lounge``` with no arguments to default to ```lounge --help```.
2. Parse the command line arguments (fix [#925](https://github.com/thelounge/lounge/issues/925)).

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -9,11 +9,9 @@ var fsextra = require("fs-extra");
 var path = require("path");
 var Helper = require("../helper");
 
-program.version(Helper.getVersion(), "-v, --version");
-program.option("");
-program.option("    --home <path>" , "home path");
-
-var argv = program.parseOptions(process.argv);
+program.version(Helper.getVersion(), "-v, --version")
+	.option("-h, --home <path>", "path to configuration folder")
+	.parseOptions(process.argv);
 
 Helper.setHome(program.home || process.env.LOUNGE_HOME);
 
@@ -40,8 +38,8 @@ require("./remove");
 require("./reset");
 require("./edit");
 
-program.parse(argv.args);
+program.parse(process.argv);
 
 if (!program.args.length) {
-	program.parse(process.argv.concat("start"));
+	program.help();
 }
--- a/src/command-line/start.js
+++ b/src/command-line/start.js
@@ -8,11 +8,11 @@ var Helper = require("../helper");
 
 program
 	.command("start")
-	.option("-H, --host <ip>", "host")
-	.option("-P, --port <port>", "port")
-	.option("-B, --bind <ip>", "bind")
-	.option("    --public", "mode")
-	.option("    --private", "mode")
+	.option("-H, --host <ip>", "set the IP address or hostname for the web server to listen on")
+	.option("-P, --port <port>", "set the port to listen on")
+	.option("-B, --bind <ip>", "set the local IP to bind to for outgoing connections")
+	.option("    --public", "start in public mode")
+	.option("    --private", "start in private mode")
 	.description("Start the server")
 	.action(function(options) {
 		var users = new ClientManager().getUsers();
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "homepage": "https://thelounge.github.io/",
   "scripts": {
     "coverage": "nyc mocha",
-    "start": "node index",
+    "start": "node index start",
     "start-dev": "npm-run-all --parallel watch start",
     "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",


src/command-line/start.js
```diff
@@ -13,6 +13,7 @@ program
 	.option("-B, --bind <ip>", "bind")
 	.option("    --public", "mode")
 	.option("    --private", "mode")
+  .parse(process.argv)
```
astorije: @xPaw, do you know why ESLint would not catch this?!

astorije: Actually, nevermind, I found the missing option and added it to #930.

@msaun008, sorry this is a nitpick, but can you replace the 2 spaces with 1 tab please? (Then use `git add src/command-line/start.js && git commit --amend && git push --force` to not create an extra commit)

src/command-line/index.js
```diff
@@ -43,5 +43,5 @@ require("./edit");
 program.parse(argv.args);
 
 if (!program.args.length) {
-	program.parse(process.argv.concat("start"));
+	program.parse(process.argv.concat("--help"));
```
xPaw: Instead of this crappy hack, can `isDefault` be used?

src/command-line/start.js
xPaw: I'm trying to understand why we need this here. I would assume the bug is happening in `index.js` due to incorrect order of parsing. There's `program.parseOptions` and `program.parse` calls in there.

src/command-line/index.js
astorije: I wasn't clear so I asked clarification to @xPaw and:

- `--help` is auto-generated by `commander`, not something we define, so we can't use `isDefault` on it.
- Replace `program.parse(process.argv.concat("--help"));` with `program.help()`.

astorije:
@msaun008, fix to our linter was merged in master, could you rebase with the upstream project please?

src/command-line/index.js
msaun008: 👍 This makes sense.

msaun008:
Tested the fix @xPaw suggested.  This works and I agree is the better fix.

lounge_start

xPaw:
@astorije I suspect that `--home` might be broken.

astorije:
Argh, good catch...
Indeed:

```sh
$ node index start --port 1234 --home /tmp
2017-02-24 23:23:00 [INFO] The Lounge source (66795b0) is now running using node 6.9.4 on darwin (x64)
2017-02-24 23:23:00 [INFO] Configuration file: /Users/jeremie/.lounge/config.js
2017-02-24 23:23:00 [INFO] Available on http://0.0.0.0:1234/ in private mode
2017-02-24 23:23:00 [INFO] Press Ctrl-C to stop
```

And this time it' not fixed with `--`.
Any suggestions for fixing this, @xPaw?

Huuuuh we do need unit tests for the CLI that would have been detected :(

xPaw:
Spend some time testing it, and there's really no easy way of handling global options, but here's the fix I came up with: https://gist.github.com/xPaw/224285fcae39f426159b252b39252097

astorije:
@xPaw, regarding your cleanups, why removing the config path displayed? It saved me a couple times to see that info, considering how weirdly we handle the config path.

xPaw:
@astorije OK, left it as it was.

astorije:
@msaun008, mind trying out the new fix suggested by @xPaw?

msaun008:
@xPaw Great fix.  Just finished testing.

```
$ node index start --port 1234 --home /tmp/delme
2017-03-04 20:54:33 [INFO] Configuration file created at /tmp/delme/config.js.
2017-03-04 20:54:34 [INFO] The Lounge source (66795b0) is now running using node 4.3.1 on darwin (x64)
2017-03-04 20:54:34 [INFO] Configuration file: /tmp/delme/config.js
2017-03-04 20:54:34 [INFO] Available on http://*:1234/ in public mode
2017-03-04 20:54:34 [INFO] Press Ctrl-C to stop
```

astorije:
Okay with this now, @xPaw? I tried to think and cannot think of another edge case we missed.

xPaw:
@astorije OK. In release, document that there's `-h` alias now.

williamboman:
This broke docker images :(. This looks like a change requiring a major bump per semver conventions imo

astorije:
If this was a library I would agree, but for an end-user client application, going from v2 to v3 simply because `lounge start` is the new `lounge` (which was never documented as such, btw) it kinda extreme.
Thanks for fixing this in the Docker images!

williamboman:
> going from v2 to v3 simply because lounge start is the new lounge (which was never documented as such, btw) it kinda extreme.

I agree, it does feel very ridiculous to require a major bump for this - but that's what semver dictates. However, I think the CLI is part of the "API surface" that should be covered by semver in the case of thelounge. I have a feeling more people launching lounge via supervisor/systemd/pm2/etc. will run into this issue when upgrading to `2.2.2`.

In hindsight and imo, I think the best course of action would've been to initially introduce a deprecation notice when running `lounge` and make the switch in the next major release.

YaManicKill:
> but that's what semver dictates 

We never agreed to follow semver.

I agree with the rest of your points (we probably should have deprecated this rather than just removing it), but for semver to mean something, we have to actually follow semver. We agreed at the beginning that we wouldn't follow semver because, as @astorije says, we are an end user application.

williamboman:
> We never agreed to follow semver.

Oh, I thought we adhered to it. That makes my point less valid :smile:

YaManicKill:
Huh, I just noticed that @astorije put that we follow semver in the CHANGELOG, but from this irc conversation last year, I thought we had agreed we wouldn't. But maybe that was my assumption and no one else agreed.

```
[2016-04-10 23:16:48] <astorije> Impaloo, I don't think it will, as it doesn't involve any breaking changes, per the semver document
[2016-04-10 23:17:40] <astorije> In fact, I don't believe we'll go to a v2 before any public API gets involved in The Lounge (like a package system for example)
[2016-04-10 23:17:47] <Impaloo> astorije: I feel like semver + client applications makes for a weird combination
[2016-04-10 23:18:57] <astorije> That's true, it will be more relevant when such API is implemented
[2016-04-10 23:19:10] <Impaloo> Or maybe not, what else is there to base versioning on
[2016-04-11 07:38:57] <xPaw> I would say 2.0 is warranted if we flip on reconnections after merging irc-fw
[2016-04-11 07:39:08] <xPaw> following semver to the stongest point of "breaking changes" is a bit silly
[2016-04-11 07:50:15] <YaManicKill> I think semver for software only used by a person does seem silly. Once we have packages, if we break the package api then we should use semver, but until then, , I think we switch to 2.0 when we have a number of good features, not just 1.
[2016-04-11 07:52:05] <xPaw> I'd say a major internal change (basically switching out entire IRC parser and adding ircv3) can be breaking
[2016-04-11 08:16:21] <YaManicKill> Breaking for who?
[2016-04-11 08:25:48] * hardtheory join
[2016-04-11 08:28:33] <xPaw> clients? there's always a chance there's an obscure server that will do something weird
[2016-04-11 08:44:27] <YaManicKill> But the point of semver isn't "We might have introduced bugs" but "This api has changed and will not work the same as before"
[2016-04-11 08:44:37] <YaManicKill> Ours is very definitely the second.
[2016-04-11 08:46:24] <xPaw> we have no api
[2016-04-11 08:46:47] <xPaw> and following your logic, we will be stuck in 1.x forever
[2016-04-11 08:50:44] <lpoujol> Well, there's the internal API. Anyone who have thinkered in the code will probably have broken stuff after the update to irc-fw
[2016-04-11 08:54:08] * FortuneDays action  tinkered with the code
[2016-04-11 08:55:33] <lpoujol> p
[2016-04-11 09:31:41] <YaManicKill> xPaw: no, my logic is that we don't follow semver, but when we feel we have a few good features to distinguish us from shout, then we can bump.
[2016-04-11 09:32:01] <YaManicKill> lpoujol: well yeah, but by that logic we could be breaking stuff every time we add or remove anything.
[2016-04-11 09:32:14] <YaManicKill> At what point do we say that it's enough to be breaking?
[2016-04-11 09:33:39] <lpoujol> I would say "could break a lot" so.
[2016-04-11 09:33:55] <YaManicKill> Also, I know we have no API. That's exactly the point. Semver is for things with APIs.
[2016-04-11 09:34:13] <YaManicKill> http://semver.org/ "MAJOR version when you make incompatible API changes,"
[2016-04-11 09:34:13] *  toggle
[2016-04-11 09:34:39] <YaManicKill> "MAJOR version when you make incompatible API changes,"
[2016-04-11 09:34:41] <YaManicKill> Ah wrong copy
[2016-04-11 09:35:13] <CornishPasty> I do think it's something worth bumping major for
[2016-04-11 09:35:55] <lpoujol> Oh
[2016-04-11 09:36:07] <CornishPasty> Even without semver
[2016-04-11 09:36:09] <lpoujol> For this system to work, you first need to declare a public API
[2016-04-11 09:36:31] <lpoujol> No public API -> You can do whatever you want
[2016-04-11 09:36:38] <lpoujol> It's the case, right?
[2016-04-11 09:36:41] <CornishPasty> Is lounge getting a plugin framework any time soon?
[2016-04-11 09:36:59] <xPaw> no
[2016-04-11 09:37:14] <YaManicKill> "For this system to work, you first need to declare a public API."
[2016-04-11 09:37:24] <YaManicKill> lpoujol: yes, exactly
[2016-04-11 09:37:30] <YaManicKill> CornishPasty: now, that's a different discussion
[2016-04-11 09:37:47] <YaManicKill> IF we think that irc framework is enough for 2.0.0 (I don't), then we can discuss that.
[2016-04-11 09:38:00] <YaManicKill> But if we are talking about whether we follow semver or not, the answer is we can't.

```

astorije:
I agree with both of you but I think we are overthinking this a little bit :)
AFAIK the only documented way to start a server is through `lounge start`, and this is true since our first release. The fact that `lounge` **also** started the server is not documented anywhere.

If we had moved starting a server *from* `lounge` *to* `lounge start`, deprecation + transition time would be in order indeed. But this release does not break `lounge start`, it makes it less dramatic `lounge`: showing help instead of starting a full server.

Also, considering our currently... ahem, questionable CLI, it's likely that a lot of changes would/will be breaking stuff :)
If we were designing a library, things would be very different as libraries are consumed by other software. In our case, our CLI is *mostly* consumed by humans, who should read the first lines of the release notes before upgrading at the very least 😅.

I am fine to remove `This project adheres to Semantic Versioning` [in the changelog](https://github.com/thelounge/lounge/blob/master/CHANGELOG.md#change-log) if you want. It looks like we all agree on the fact that we are following a sorta-semver for end-user applications and that things are less black and white there. This is really not the kind of things I want to focus on 😄.

PR #931, opened 2017-02-18
Update to jQuery 3

xPaw:


MaxLeiter:
Worked for me, didn't notice any difference.

PR #939, opened 2017-02-24
Disable (temporarily) client ping timeouts

xPaw:
Currently client pings are causing us some issue due to the way it's implemented in irc-fw. As I expect a full fix to take some time to be implemented I propose we disable it.


astorije:
@xPaw, what are the consequences of disabling this outside of the ping timeout issue we're seeing? Or in other words, what were the supposed benefits of having that set?

xPaw:
@astorije Currently it provides more harm than good. It's supposed to find stale connections and drop them (e.g. if server is unresponsive, your network silently died, etc.)

PR #955, opened 2017-03-11
Implement pgup/pgdown keys

### LINKED ISSUE #263: Page up/down keys do not work
Since we almost always have a focus in the input, pressing page up/down keys on your keyboard does nothing.

I tried hooking keydown on input and propagating it to chat window, but that does not work.

xPaw:
Fixes #263.

There's a problem I am unable to currently to solve. Pressing page down while at the bottom and then pressing page up right after does not seem to trigger a scroll.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1302,6 +1302,33 @@ $(function() {
 
 	(function HotkeysScope() {
 		Mousetrap.bind([
+			"pageup",
+			"pagedown"
+		], function(e, key) {
+			let container = windows.find(".window.active");
+
+			// Chat windows scroll message container
+			if (container.attr("id") === "chat-container") {
+				container = container.find(".chan.active .chat");
+			}
+
+			const offset = container.get(0).clientHeight * 0.94;
+			let scrollTop = container.scrollTop();
+
+			if (key === "pageup") {
+				scrollTop -= offset;
+			} else {
+				scrollTop += offset;
+			}
+
+			container.stop().animate({
+				scrollTop: scrollTop
+			}, 200);
+
+			return false;
+		});
+
+		Mousetrap.bind([
 			"command+up",
 			"command+down",
 			"ctrl+up",
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -808,7 +808,7 @@ $(function() {
 
 	var input = $("#input")
 		.history()
-		.on("input keyup", function() {
+		.on("input", function() {
 			var style = window.getComputedStyle(this);
 
 			// Start by resetting height before computing as scrollHeight does not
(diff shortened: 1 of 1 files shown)


williamboman:
How is this supposed to work? Should it paginate when `#input` is focused? If so, this doesn't work on condensed Mac keyboards with `fn` + arrow keys.

xPaw:
Does fn+arrow key on Mac books trigger a page key codes? And yes, it should work with focus.

williamboman:
> Does fn+arrow key on Mac books trigger a page key codes? And yes, it should work with focus.

My bad, I was triggering the `Home` and `End` keys. It registers fine with `fn` + arrow keys.

astorije:
@williamvoman do you confirm it works as expected for you?

astorije:
@williamboman do you confirm it works as expected for you?

astorije:
> There's a problem I am unable to currently to solve. Pressing page down while at the bottom and then pressing page up right after does not seem to trigger a scroll.

That's odd. I'll have to look into it but would appreciate help debugging this.

xPaw:
Could be Chrome/Windows specific oddness, i am unsure it's caused by this code alone.

xPaw:
> Instead of scrolling the whole container height, could you leave some headroom? Like container.clientHeight * .8 or something

I was looking at Discord, they scroll entire height, but head room could be desired.

> Do we want to catch these events everywhere in the app or only when focusing on the input element, to leave the OS freedom when focusing somewhere else (regardless of the input being currently focus greedy)?

Global for consistency. You *have* to be focused in the message container to scroll it with keyboard, which we can't accomplish due to input focus.

> I believe the bug you mention is because of the lack of bounds. If I press pagedown 5 times at the bottom (doesn't happen at the top), I'll have to press pageup 6 times to go up. I have noticed this in the past and scrollTop has a mind of his own. Probably want to make sure scrollTop has a maximum.

https://developer.mozilla.org/en/docs/Web/API/Element/scrollTop

I tested it, and it didn't help, also:
> scrollTop  doesn't respond to negative values; instead, it sets itself back to 0.
If set to a value greater than the maximum available for the element, scrollTop settles itself to the maximum value.

astorije:
Looks nice like that! Love the animation, it looks very natural.

One glitch though if, at the bottom, you press pagedown then pageup right afterwards, it animates up then moves down...

Also, I don't even see one full line on my screen so I'd do `.9` instead of `.94` :man_shrugging:

xPaw:
> One glitch though if, at the bottom, you press pagedown then pageup right afterwards, it animates up then moves down...

Wtf. Is `.stop()` not working as expected?

astorije:
Hey @xPaw, after spending quite some time on this, I found out the source of the glitch: another handler on `input` was listening to `keyup` and adding `.sticky` to it. When making the animation `2000` to test, that would always cancel out the pageup.
I removed `keyup` as `input` event catches all actual changes to the input value. I tested a bit (both the scroll and the autoresize input) and did not see something wrong, but hopefully someone else will test and let us know (@williamboman?).

Also reduced the magic number as on my screen I didn't even have a full line of context, and `floor`ed/`ceil`ed the value: scrolling up then down at the bottom would not scroll down entirely, making the scrollbar not "stickied" to the bottom.

Let me know what you think!

xPaw:
I think that keyup event was to correctly resize multiline inputs.

astorije:
> I think that keyup event was to correctly resize multiline inputs.

This is completely captured by the `input` event. Essentially we want to update the input height on *content change*, not on *key press*. As far as I could test, this works as expected.

xPaw:
Fine by me, merge then.

PR #970, opened 2017-03-17
Fix variable shuffling around ident handler

### LINKED ISSUE #965: Lounge crashes on joining a network (master/ff72ebb)
I tested this with private mode on, but it seems to also happen in public mode. 

Crash: https://hastebin.com/inucebezil.js

entire terminal session: https://hastebin.com/ijuzarusex.txt (tl;dr: cloned repo, installed, and removed .lounge config folder (so default configs) before running)

xPaw:
Fixes #965


MaxLeiter:
Can confirm this fixes #965

PR #972, opened 2017-03-18
New IRC message parser

xPaw:
Supersedes #699. Brings the parser code back into the repository, does not use a polyfill, fixes node4 compat (for tests) @bonuspunkt gave his OK in https://github.com/thelounge/lounge/pull/699#issuecomment-285988183

Fixes #654, #15, #199, #583, #928, #1001.

Let's get this merged and work on issues (if any) along the way, it's been a PR for a while now, we do really need to get this moving.

PR #973, opened 2017-03-18
Use require() instead of import in client code

### LINKED ISSUE #895: Use require() instead of import
Ref: #853

xPaw:
Closes #895.

Got bitten by `Block-scoped declarations (let, const, function, class) not yet supported outside strict mode` while working on #972 and to get strict mode gotta drop `sourceType` in eslint.

### DIFF
--- a/client/.eslintrc.yml
+++ b/client/.eslintrc.yml
@@ -1,3 +1,0 @@
----
-parserOptions:
-  sourceType: module
--- a/client/js/libs/handlebars/colorClass.js
+++ b/client/js/libs/handlebars/colorClass.js
@@ -1,3 +1,5 @@
+"use strict";
+
 // Generates a string from "color-1" to "color-32" based on an input string
 module.exports = function(str) {
 	var hash = 0;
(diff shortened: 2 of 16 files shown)


Bonuspunkt:
just to keep in mind
Tree-shaking should perform better when used with import (ES-modules)
http://2ality.com/2015/12/webpack-tree-shaking.html

xPaw:
@Bonuspunkt Don't think that's a problem in our case.

xPaw:
It's due to node4 support.

astorije:
> It's due to node4 support.

These are all client-related though so that shouldn't matter.
This is probably a problem of configuration, but when it comes to CI, we should have:

- Node 4, 6, 7, etc. run against our server tests
- Browser(s) or Phantom or whatever run against our client tests

I remember @Bonuspunkt brought up this issue some time ago and there is probably a good way (or several) to address it.
Maybe something like [this](http://engineering.widen.com/blog/testing-future-web-stack/) could bring an element of solution?

xPaw:
It matters if we test client code, which we do for the message parser.

astorije:
Right, I agree, but yeah we should run client tests against a browser, not against Node (or at least disable client tests on Node v4, but probably not the path we want to take :S).

astorije:
This also closes #974.

xPaw:
:+1:

PR #977, opened 2017-03-19
Do not build json3 module with Webpack

xPaw:
Both of these are only used by `socket.io` and it contributes an unnecessary amount of size in the final vendor chunk. It removes 23kB from the final vendor chunk in production build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "start-dev": "npm-run-all --parallel watch start",
     "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",
-    "build:webpack": "webpack",
+    "build:webpack": "webpack --progress",
     "watch": "webpack --watch",
     "test": "npm-run-all -c test:* lint",
     "test:mocha": "mocha",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -60,6 +60,9 @@ let config = {
 			},
 		]
 	},
+	externals: {
+		json3: "JSON", // socket.io uses json3.js, but we do not target any browsers that need it
+	},
 	plugins: [
 		new webpack.optimize.CommonsChunkPlugin("js/bundle.vendor.js")
 	]


package.json
```diff
@@ -17,7 +17,7 @@
     "start-dev": "npm-run-all --parallel watch start",
     "build": "npm-run-all build:*",
     "build:font-awesome": "node scripts/build-fontawesome.js",
-    "build:webpack": "webpack",
+    "build:webpack": "webpack --progress",
```
astorije: Is this a side-effect of the rest of this PR, or just something you've added along the way?

package.json
```diff
@@ -75,6 +75,7 @@
     "socket.io-client": "1.7.3",
     "stylelint": "7.9.0",
     "urijs": "1.18.9",
-    "webpack": "2.2.1"
+    "webpack": "2.2.1",
+    "webpack-remove-debug": "0.1.0"
```
astorije: This comes from https://github.com/johngodley/webpack-remove-debug.
1 commit only, 0 stars/forks, pretty much no downloads... Shouldn't we inline the [6 lines of code](https://github.com/johngodley/webpack-remove-debug/blob/master/index.js) of this project directly into our Webpack config, instead?

Also, I'm curious of such little usage of this, what do other projects do, they just leave the debug stuff in there?
Finally, do we want to remove this from both production and development environment, or is there an interest in keeping this in dev mode?

package.json
xPaw: Along the way.

package.json
xPaw: Yes i was thinking about inlining it.

astorije: I tried to inline it (using a function) but it looks like we have to make it an internal module in its own file and set it as `loader: "./file"`. That sucks. Anyone knows a way we could use to simply define this one-line function at the top of the webpack config and call that function from the loader?!

xPaw: Could just exclude it and only keep the json3 trick.

xPaw:
👍

astorije:
I moved to debug stuff to #993 for reference.

Merging this with one review as it's a very simple change and comes from a maintainer.

PR #979, opened 2017-03-21
README: add DockerHub badge

williamboman:
¯\\\_(ツ)_/¯


astorije:
I'd rather not. On the Docker-specific repo however, sure!
Also, new docs (yes, I'm getting back to them) will give a more prime-time visibility to Docker and other official ways to install The Lounge than the current docs.

williamboman:
gotcha

PR #980, opened 2017-03-21
Implement a proper LDAP authentication process

eliemichel:
The Lounge first log as a special user in order to search (as in LDAP's
'"search" verb) for the user's full DN. It then attempts to bind using the
found user DN and the user provided password.

See comments in LDAP config section for more information about the
process.


astorije:
The problem with LDAP is that currently, no one in the team has an extensive knowledge or usage of it with The Lounge. Is there someone who could vet/review these changes?

Also, would someone from the community be willing to step up and help us regarding LDAP integration on a more permanent basis? 😄

Ideally, we really make changes to extract LDAP integration into its own repository. That would make for our first brick of plugins :)

eliemichel:
I implemented this given the guidelines that my LDAP server administrator provided me with. This search-then-bind pattern seems quite standard to me after a bit of research, though I am not personnaly an LDAP expert.

Anyway, since I'm know maintaining an instance of The Lounge using LDAP authentication and that I'm likely to keep it, I'm ready to follow up on this.

How would you imagine the auth plugin basic API? I guess a initial `registration()` function and an `auth()` one, or sth like this.

astorije:
What do other @thelounge/maintainers feel about this?

xPaw:
Would this PR break backwards compatibility with previous versions?

eliemichel:
@xPaw Yes, it will. You're right, I'm going to edit this to enable back the previous behaviour too. It is actually common to have the administrator choose between pattern based (the old one) and search based (this PR) LDAP auth. [Dovecot](https://wiki2.dovecot.org/AuthDatabase/LDAP/AuthBinds) does it, for instance.

xPaw:
cc @thisisdarshan for review.

astorije:
We'll remove this PR from v2.3.0 milestone for now unless LDAP folks think they can help us review this very soon 😕 Sorry!

astorije:
@gramakri, we sure could help here. A few questions:

- Is this still relevant?
- Is this still working?
- Is this something we want and need in the project?

ollipa:
I've been running this on my own build based on 2.3.0-rc1 for a while now without any problems. I am authenticating against Windows Active Directory and could not get the authentication to work without this pull request. It would be great to see this feature in master branch. I'm happy to help if any further testing is needed.

YaManicKill:
> I've been running this on my own build based on 2.3.0-rc1 for a while now without any problems.

Great thanks, @ollipa, that's good to know.

@eliemichel Does this still break backwards compatibility?

@thelounge/maintainers Sounds like this might be good to go soon, depending on the answer to the above question.

eliemichel:
Sorry guys I haven't been able to find the time for this... Still using the lounge though.

@YaManicKill As far as I know it still breaks backward compatibility because I did not change anything. The most simple way I can think of to get backward compatibility "for free" would be to have completely separate config fields for "old ldap" and "new ldap" (this one) and keep both systems at the same time.

What would be cleaner would be to have the old and new ldap auth in some kind of plugins-like things so that it would be easy to start depreciate the old one in the next major update.

MaxLeiter:
I think it's worth breaking backwards compatibility if we write in a catch to detect if the old setup is present

defaults/config.js
```diff
@@ -329,6 +329,23 @@ module.exports = {
 	// @type    object
 	// @default {}
 	//
+	// The authentication process works as follows:
+	//
+	//   1. Lounge connects to the LDAP server with its system credentials
+	//   2. It performs a LDAP search query to find the full DN associated to the
+	//      user requesting to log in.
+	//   3. Lounge tries to connect a second time, but this time using the user's
+	//      DN and password. Auth is validated iff this connection is successful.
```
MaxLeiter: if

MaxLeiter: for those curious, DN == `distinguished name`

dgw: re: "if", [not necessarily](https://en.wiktionary.org/wiki/iff)

quanah:
A few comments:

* The "ldaps" protocol was a bastardization introduced for LDAPV2, where there was no native support for TLS in LDAP.  Starting with LDAPv3, the startTLS extended operation was introduced.  The code should support both startTLS and ldaps
* It should not be required to configure a user to do a search as.  It should be possible to do an anonymous search if desired
* It should be possible to simply define the userDN and account base, allowing binds with no requirement to search
* I would note that memberOf is primarily an Active Directory thing, and it is not part of any approved LDAP standard.  I'm not sure I'd use it as part of a default configuration.

Hope that helps!

eliemichel:
@quanah 

> The "ldaps" protocol was a bastardization

The current PR code only supports ldaps. Adding starttls is a matter of adding an option in the config and us [starttls()](http://ldapjs.org/client.html#starttls).

> It should not be required to configure a user to do a search as.

That's right, yet another option to add to config.

> It should be possible to simply define the userDN and account base

This is the so-called "old behavior". I think it's ok to keep it as an option, since it is easier to setup, but not as default, since it is not that much recommended. As @MaxLeiter said, there must be a way to detect if people were using old behavior to turn it in. Since this PR does not make use of the `baseDN` config entry, one could just assume that as soon as this entry is specified in conf, behavior should switch to the old one.

> I would note that memberOf is primarily an Active Directory thing

Ok, I just wanted the default config to show of that filter can do advanced things, but it can be something else than `memberOf`.

xPaw:
@eliemichel Do you want to implement the changes you've suggested in your last comment?

eliemichel:
@xPaw Back from holidays, I'm on it

src/server.js
```diff
@@ -312,7 +314,7 @@ function initializeClient(socket, client, generateToken, token) {
 				type: "notification",
 				timestamp: Date.now(),
 				title: "The Lounge",
-				body: "🚀 Push notifications have been enabled"
+				body: " Push notifications have been enabled"
```
MaxLeiter: why this change?

src/server.js
```diff
@@ -538,11 +437,17 @@ function performAuthentication(data) {
 	}
 
 	// Perform password checking
+	let auth = function(){};
```
MaxLeiter: I think this should be `function() {}`

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,21 @@
+var Helper = require("../../helper");
```
MaxLeiter: change the var's to `const`

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,21 @@
...
+var ldap = require("ldapjs");
+
+var _ldapAuthCommon = require("./_ldapCommon");
```
MaxLeiter: here too

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,21 @@
...
+
+function ldapAuth(manager, client, user, password, callback) {
+	if (!user) {
+		return callback(false);
+	}
+
+	var config = Helper.config;
```
MaxLeiter: same for here, `userDN`, `bindDN`

src/server.js
```diff
@@ -11,7 +11,9 @@ var path = require("path");
 var io = require("socket.io");
 var dns = require("dns");
 var Helper = require("./helper");
-var ldap = require("ldapjs");
+var ldapAuth = require("./plugins/auth/ldap");
```
MaxLeiter: `const` again

defaults/config.js
```diff
@@ -436,27 +433,55 @@ module.exports = {
...
-		base: "dc=example,dc=com",
-
-		//
-		// LDAP search scope
-		//
-		// @type     string
-		// @default  "sub"
-		//
-		scope: "sub"
+		// LDAP search dn settings. This defines the procedure by which the
```
MaxLeiter: can we replace dn with defined name for the first instance?

src/plugins/auth/advancedLdap.js
```diff
@@ -0,0 +1,70 @@
+var Helper = require("../../helper");
```
MaxLeiter: `const` here too

MaxLeiter: and for the whole file replace var with either `const` or `let`

MaxLeiter:
Failing builds due to some linting issues, but overall seems like a major improvement. 👏 @eliemichel

src/server.js
```diff
@@ -491,11 +437,17 @@ function performAuthentication(data) {
 	}
 
 	// Perform password checking
+	let auth = function() {};
```
xPaw: Just do `let auth;`, there is no need to assign anything.

src/server.js
```diff
@@ -491,11 +437,17 @@ function performAuthentication(data) {
...
 	if (!Helper.config.public && Helper.config.ldap.enable) {
-		ldapAuth(client, data.user, data.password, authCallback);
+		if ("baseDN" in Helper.config.ldap) {
```
xPaw: Can this be done inside of the `ldap` plugin file instead?

xPaw:
Cool changes, however for the sake of simplicity I think `local` auth should be kept where it is, and this PR to only deal with ldap code.

src/server.js
eliemichel: Because of some copy pasting that skipped it, I guess. Not intended.

src/server.js
eliemichel: I'm not so fan of a variable switching its type as the program goes. :/

xPaw: yay for javascript /s

realistically you're doing an unnecessary allocation here that gets thrown away.

eliemichel:
(bug fix pending, do not merge yet)

eliemichel:
All right, should be good now. :) I'll be back later on.

xPaw:
@eliemichel Do you think you will be able to create some tests for the ldap stuff? `ldapjs` also has a server, so it should be doable.

eliemichel:
@xPaw I can give it a try, what's the guidelines for making tests for the lounge?

xPaw:
@eliemichel Basically just require the relevant file, call functions in it and see if the output matches what you expect to happen.

https://github.com/thelounge/lounge/tree/master/test you can take a look around on how it's done.

eliemichel:
@xPaw How can I ask mocha to run one test in particular? I don't see it running my test.

xPaw:
@eliemichel `npm test` should run find all the tests recursively automatically. Looking at Travis, it did run and fail:
```
  1) LDAP authentication "before all" hook:
     TypeError: authorizedUsers.keys is not a function
``

**Comment by eliemichel** [comment]
Yes, I was looking at Travis. `npm test` on my machine does not run the tests.. But calling `mocha test/plugins/auth` did the job. I'm fixing this.

**Comment by eliemichel** [comment]
@xPaw Any idea how my changes could have perturbed the WebSocket tests that fail now? I cannot reproduce this at home.

**Comment by xPaw** [comment]
@eliemichel I can imagine `Helper.config.public = false;` doing that as other tests depend on being public afaik.

**Comment by eliemichel** [comment]
Shouldn't all the tests set their config up as they expect it? Or shouldn't the config be reloaded from defaults between each test session?

**Comment by xPaw** [comment]
Sounds like it yeah.

File: client/js/options.js [hunk 12]
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 const $ = require("jquery");
+require("jquery-textcomplete");
```
xPaw: Very odd that you've had build issues with this. But this is an issue unrelated to this PR so please remove it.

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,132 @@
...
+	// TODO: move this out of here and get rid of `manager` and `client` in
+	// auth plugin API
+	function callbackWrapper(valid) {
+		if (valid && !client) {
+			manager.addUser(user, null);
+		}
+		callback(valid);
+	}
+
+	let auth = function() {};
```
xPaw: Don't do the unnecessary allocation please.

src/plugins/auth/local.js
```diff
@@ -0,0 +1,44 @@
...
+
+			callback(matching);
+		}).catch((error) => {
+			log.error(`Error while checking users password. Error: ${error}`);
+		});
+}
+
+module.exports = {
+	auth: localAuth,
+	isEnabled: function() {
```
xPaw: This can be `() => true`

src/server.js
```diff
@@ -491,11 +436,13 @@ function performAuthentication(data) {
 	}
 
 	// Perform password checking
-	if (!Helper.config.public && Helper.config.ldap.enable) {
-		ldapAuth(client, data.user, data.password, authCallback);
-	} else {
-		localAuth(client, data.user, data.password, authCallback);
+	let auth = function() {};
```
xPaw: No allocation please.

test/plugins/auth/ldap.js
```diff
@@ -0,0 +1,156 @@
...
+	});
+
+	const p2 = new Promise((resolve) => {
+		ldapAuth.auth(manager, client, user, wrongPassword, function(valid) {
+			expect(valid).to.equal(false);
+			resolve();
+		});
+	});
+
+	Promise.all([p1, p2]).then(function() {
```
xPaw: Can you split these promises into separate tests? It will be a lot cleaner in the test report.

Also a test for invalid user might not be a terrible idea.

test/plugins/auth/ldap.js
```diff
@@ -0,0 +1,156 @@
+"use strict";
+
+const ldapAuth = require("../../../src/plugins/auth/ldap");
+const Helper = require("../../../src/helper");
+const ldap = require("ldapjs");
+const expect = require("chai").expect;
+
+const user = "johndoe";
+const correctPassword = "loremipsum";
+const wrongPassword = "dolorsitamet";
```
xPaw: Doesn't need to be a const here.

xPaw:
@eliemichel Thanks for your great work! Are you in `#thelounge` by the way?

astorije:
Thanks a lot for this, keep up the good work @eliemichel!
We are very excited you are eager to bring this to completion, and even more that you are shipping tests with this 😄

client/js/options.js
eliemichel: Removed

src/plugins/auth/ldap.js
eliemichel: Ok, will do

src/plugins/auth/local.js
eliemichel: Done.

src/server.js
eliemichel: Done.

test/plugins/auth/ldap.js
eliemichel: Good idea, I did not though about just using different `it()` instead of promises trickery. Done.

test/plugins/auth/ldap.js
eliemichel: Mmh but why not making it const since it can be const?

eliemichel:
Still this websocket issue that Ido not understand

xPaw:
@eliemichel Try setting `public = true` in `before` callback in `test/server.js`.

eliemichel:
Done. We'll see what it gives, but arn't websocket supposed to work on private instances as well?

xPaw:
Tests don't include auth so they fail.

xPaw:
Looks solid to me now. Could you squash some of the commits for a nicer history?

eliemichel:
Ok, this is done in #1478, which now superseeds this PR.

dgw:
@eliemichel In the future, you might find it useful to simply force-push to the PR branch after rebasing/squashing locally. GitHub handles that gracefully, and it keeps more of the discussion together. :smile_cat:

PR #983, opened 2017-03-22
bcrypt: switch to faster JS implementation

### LINKED ISSUE #982: Raspberry Pi 2 issues with login in private mode
I tried today to install TheLounge on my raspberry pi. Unfortunately, the private mode did not work, the public without user management worked fine. After some digging into the code I think I found the culprit to be the bcrypt function call. My guess is that the pi is not fast enough in calculating the hashes and therefore nodeJS throws a time-out as during the login the CPU is at 100%. 

It does not go beyond: https://github.com/thelounge/lounge/blob/c409328ddf38db3b421c61bf2f688317f8e07f9d/src/server.js#L270

Hardcoding the return value to `return true` lets me use TheLounge without further issues. 

I tried reducing the rounds of bcrypt [here](https://github.com/thelounge/lounge/blob/c409328ddf38db3b421c61bf2f688317f8e07f9d/src/helper.js#L124) to as low as 3 but it did not help.

Node.JS Version: v6.10.0
NPM Version 3.10.10
TheLounge Version: v2.2.2 - 2017-03-13
Raspberry Pi 2 

I wrote this issue on request of @MaxLeiter

williamboman:
Closes #982.

Results from [evaluating](https://hastebin.redwill.se/zosomugoxi) `bcrypt`, `bcryptjs`, `bcrypt-nodejs` (via their async API) on a Raspberry Pi 3 with 3 rounds;

```sh
[bcrypt] Computing 58 password hashes.
[bcrypt] total: 282.228ms

[bcryptjs] Computing 58 password hashes.
[bcryptjs] total: 2483.178ms

[bcrypt-nodejs] Computing 58 password hashes.
[bcrypt-nodejs] total: 394372.514ms

```

`bcrypt-nodejs` performs really bad, and `bcrypt` is quite a heavy dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "node": ">=4.2.0"
   },
   "dependencies": {
-    "bcrypt-nodejs": "0.0.3",
+    "bcryptjs": "2.4.3",
     "cheerio": "0.22.0",
     "colors": "1.1.2",
     "commander": "2.9.0",
--- a/src/helper.js
+++ b/src/helper.js
@@ -6,7 +6,7 @@ var path = require("path");
 var os = require("os");
 var fs = require("fs");
 var net = require("net");
-var bcrypt = require("bcrypt-nodejs");
+var bcrypt = require("bcryptjs");
 
 var Helper = {
 	config: null,


williamboman:
Closing in favor of #985

astorije:
@williamboman, why did you prefer #985 over this? If this PR provides a 150x performance improvement, this seems like a no-brainer to me :)
Also the underlying repo seems much more sane and maintained, while the repo of bcrypt-nodejs is not even linked from the 4-year old Node package anyway.

Also, if even bcryptjs is too slow for #982, then I'm not certain the problem comes from this/us and more an environment issue.

I understand the last alternative is https://github.com/kelektiv/node.bcrypt.js, what made you say it is heavier?

xPaw:
#985 uses asynchronous code, it's generally more preferable for functions that take time to execute.

astorije:
Aaaah for some reason I mixed #984 and #985! So I thought this switch was closed in favor of reducing the number of rounds in the config, my bad.
Don't mind me then!

PR #984, opened 2017-03-22
bcrypt: make amount of bcrypt rounds configurable

williamboman:


defaults/config.js
```diff
@@ -291,6 +291,14 @@ module.exports = {
 	},
 
 	//
+	// Amount of rounds to use for bcrypt passwords.
+	//
+	// @type      number
+	// @default   11
+	//
```
williamboman: Maybe add a notice of the implications of setting a low or high value here?

xPaw: Could have a link to https://security.stackexchange.com/a/3993

MaxLeiter:
Personally against adding -more- config options, esp with #986 and others. However, no real other way to make this configurable.

williamboman:
I thought about perhaps automatically adjusting the amount of rounds depending on CPU capabilities, with a fixed min/max

astorije:
I missed a couple weeks of discussion, what's the purpose of this? Performance? How bad is it? (I'm running on an Rpi 2 and didn't notice anything bad)

This makes the app weak-able by default, probably not something we should do. If people want to mess with the source code, sure, but for a lot of inexperienced sysadmins (which is possible in our case, since The Lounge is not exactly enterprise-grade!), they might see this as low = high perf.

williamboman:
> I missed a couple weeks of discussion, what's the purpose of this? Performance? How bad is it? (I'm running on an Rpi 2 and didn't notice anything bad)

Is it the 700MHz 512MB one? JS implementations of bcrypt with 11 rounds sort of times out on it (hence #984).

astorije:
No it's the Rpi 2 so 900Mhz and 1GB of RAM.

It looks like switching to `bcryptjs` in #985 should be enough to become more than acceptable with 11 rounds, doesn't it?

williamboman:
> It looks like switching to bcryptjs in #985 should be enough to become more than acceptable with 11 rounds, doesn't it?

No, doesn't look like it. It performs really really bad on the 700 MHz version.

```sh
$ node bcrypt.js 11
[bcryptjs] Computing 1 password hashes.
[bcryptjs] total: 140516.808ms
```

[This is the test program I wrote for it](https://hastebin.redwill.se/adevexatem.coffeescript)

astorije:
Wow, that's bizarre... Not sure I want us to switch to `bcryptjs` then 😅.

Just curious, could you test your benchmark with https://www.npmjs.com/package/bcrypt please (regardless of dependency issues)?

astorije:
Also, what about switching to PBKDF2, which seems to me like it's [available directly in Node's native `crypto` module](https://nodejs.org/api/crypto.html#crypto_crypto_pbkdf2_password_salt_iterations_keylen_digest_callback) and the recommendation these days (I haven't looked further than that though).

xPaw:
PBKDF2 is weaker than bcrypt. And switching hashing algorithms would require us to maintain both for backwards compatibility.

astorije:
> And switching hashing algorithms would require us to maintain both for backwards compatibility.

True, but it doesn't have to be definitive, we can convert successful passwords first time they log in, and disable this in a couple versions. I'll admit it's not ideal though, but neither is the current situation.

williamboman:
> Wow, that's bizarre... Not sure I want us to switch to bcryptjs then 😅.

No no, `bcryptjs` is infinitely times better. It's just that it takes `~2m20s` vs. `~1m56s` with 11 rounds on a 700MHz RPI (both performs so absurdly bad that the difference is marginal). It heavily outperforms `bcrypt-nodejs` in most cases.

williamboman:
> Just curious, could you test your benchmark with https://www.npmjs.com/package/bcrypt please (regardless of dependency issues)?

On 700MHz RPI:

```sh
[bcrypt] Computing 1 password hashes.
[bcrypt] total: 1217.932ms

[bcryptjs] Computing 1 password hashes.
[bcryptjs] total: 141651.885ms

[bcrypt-nodejs] Computing 1 password hashes.
[bcrypt-nodejs] total: 120035.263ms

```

astorije:
Hey @williamboman, thanks for these, it's very useful.

> (both performs so absurdly bad that the difference is marginal)

Indeed, I didn't pay attention the numbers were so high it was not even relevant, just noticed the difference quickly.

The 700MHz is the RPi gen 1 (whatever model), and I own the version 2 (900MHz). I am surprised to see so much difference between them as I have no problem on mine.

> but....

Is this:

- The same amount of time on all versions?
- Does it take similar time upon reinstall or is it the first time you install?
- Is there an official package repository for RPi we could recommend that would contain this as a binary?
- On an x86/x64 arch, is it instant?

Last question is because it's not unexpected to have to recompile for the RPi arch considering of its nature, while common architectures would have this available as binary somewhere.

Are we ruling out switching to `bcrypt`? If it's *only* about install time, I'd say it's worth it because in general usage pain is more of a concern than install pain (with a warning). But if it has other implications, like known to not work on a big chunk of architectures out there, then that's more problematic.

williamboman:
> The same amount of time on all versions?
> Does it take similar time upon reinstall or is it the first time you install?
> Is there an official package repository for RPi we could recommend that would contain this as a binary?
> On an x86/x64 arch, is it instant?
> Last question is because it's not unexpected to have to recompile for the RPi arch considering of its nature, while common architectures would have this available as binary somewhere.

There are no prebuilt binaries for any architecture (that currently works with `node-pre-gyp`), see https://github.com/kelektiv/node.bcrypt.js/issues/473. It builds from source on all architectures.

> Are we ruling out switching to bcrypt? If it's only about install time, I'd say it's worth it because in general usage pain is more of a concern than install pain (with a warning). But if it has other implications, like known to not work on a big chunk of architectures out there, then that's more problematic.

I'd say it's 20% about install time and 80% about the additional requirement of `python v2.7`, `make` and `gcc` should we use `bcrypt`. macOS will need Xcode CLI tools installed as well. Not sure what status on Windows is.

astorije:
> It builds from source on all architectures.

Uuuuh that's too bad...

> I'd say it's 20% about install time and 80% about the additional requirement

Yeah I was afraid you'd say that. [This thread](https://groups.google.com/forum/#!topic/node-red/V7_aFjtpEVE) mentions an idea at the end: `bcrypt` in optional dependencies, `bcryptjs` in prod dependencies, and a `try...catch` to pick one or the other. Definitely not ideal because now that's 2 deps to maintain, install, etc.

Also, could you run https://www.npmjs.com/package/bcrypt-speed on your RPis? This is the result on my gen 2:

```
$ time npm i
/
> bcrypt@1.0.2 install .../node_modules/bcrypt
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! Tried to download(404): https://github.com/kelektiv/node.bcrypt.js/releases/download/v1.0.2/bcrypt_lib-v1.0.2-node-v46-linux-arm.tar.gz
node-pre-gyp ERR! Pre-built binaries not found for bcrypt@1.0.2 and node@4.8.0 (node-v46 ABI) (falling back to source compile with node-gyp)
...
bcrypt-nodejs@0.0.3 node_modules/bcrypt-nodejs

async@0.9.2 node_modules/async

twin-bcrypt@2.1.1 node_modules/twin-bcrypt

bcryptjs@2.4.3 node_modules/bcryptjs

bcrypt@1.0.2 node_modules/bcrypt
...

real	1m5.532s
user	1m4.910s
sys	0m4.990s
```

```
$ node bcrypt-speed.js 10000
test duration per number of rounds, up to 10000ms

bcrypt-nodejs (pure js, no deps)
rounds  ms
  4    132
  5    112
  6    196
  7    381
  8    727
  9   1437
 10   2875
 11   5704
 12  11408

twin-bcrypt (pure js, no deps, asm.js)
rounds  ms
  4    122
  5     93
  6    167
  7    333
  8    646
  9   1292
 10   2565
 11   5124
 12  10243

bcryptjs (pure js, no deps)
rounds  ms
  4    148
  5     35
  6     62
  7    122
  8    245
  9    475
 10    954
 11   1889
 12   3792
 13   7484
 14  15002

bcrypt (c++, has deps)
rounds  ms
  4     10
  5     16
  6     32
  7     51
  8     83
  9    165
 10    329
 11    659
 12   1315
 13   2627
 14   5239
 15  10477
```

Looking at these, I'm still a bit surprised by the results of your RPi gen 1. I mean, I know gen 2 is a lot faster than gen 1, but it looks to me like we get 100x ratio.

So it looks like the switch from `bcrypt-nodejs` and `bcryptjs` is a no-brainer: better and faster lib, more up-to-date, etc. But the switch to native package `bcrypt` is different story that comes with caveats.
Let's see how #985 goes but I really hope we'll be able to keep bcrypt settings outside of The Lounge settings, effectively offering the possibility to make it insecure through the settings.

A possibility (maybe over the top, but I'd still take that over a setting) could be to do exactly what that small benchmark is doing: post-install, run 4 to N (11? More?) rounds and take the biggest value that fits in a max time we are comfortable having users wait (I'd say something like 5 seconds max), use that value on this server.

astorije:
Alright, after looking around a bit, making the number of rounds (cost) customizable is definitely not something we should do.
[This](https://security.stackexchange.com/questions/17207/recommended-of-rounds-for-bcrypt) gives a nice overview of why, and what alternatives there are.

I think we should address this differently (also, it's a much more interesting change then :P), a bit like what I mentioned in my last comment: we should decide how much time do we want users to wait when logging in (1s seems ok, that's 10 rounds on a RPi 2 with `bcryptjs`), then compute the cost factor based on this. Someone running The Lounge on a Core i7 should not serve with the same cost factor than on a RPi 1!

The way I'm seeing it, we could have a `postinstall` script that essentially runs [this](https://github.com/timkuijsten/node-bcrypt-speed/blob/7477c55547bc47041f3b7df7546dff584ef13936/bcrypt-speed.js#L76-L96), stops iterating whenever elapsed time is > 1s and take the previous value, saves it in a file that server reads upon startup (probably with a min value or 4 or 6 depending on the result of `node-bcrypt-speed` on @williamboman's RPi 1).
In theory, this could add a few seconds to the install time but not more than that. We're talking less than 5 if my math is correct.
Alternative could be to do that at server startup since this is so fast and that would avoid saving to a file. I don't have a preference.

This seems overkill but I don't think it is. Probably very easy to implement, and that also guarantees optimal security and page load. Just need to discuss the details (postinstall, server startup, etc.).
Thoughts?

williamboman:
Sure!

PR #985, opened 2017-03-23
Switch to bcryptjs and make password comparison async

### LINKED ISSUE #982: Raspberry Pi 2 issues with login in private mode

rockhouse:
Closes #982 

Changed to faster brcypt implementation (brcyptjs) for better performance and made the compare call async so it does not timeout anymore. The check still takes ages on resource constraint devices as on the older models of the raspberry pi but it is usable now, you just need a little patience ;)

Performance comparison of the different bcrypt implementations can be found here: https://github.com/thelounge/lounge/pull/983

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -6,7 +6,7 @@ var path = require("path");
 var os = require("os");
 var fs = require("fs");
 var net = require("net");
-var bcrypt = require("bcrypt-nodejs");
+var bcrypt = require("bcryptjs");
 
 var Helper = {
 	config: null,
@@ -125,5 +125,5 @@ function passwordHash(password) {
 }
 
 function passwordCompare(password, expected) {
-	return bcrypt.compareSync(password, expected);
+	return bcrypt.compare(password, expected);
 }
--- a/src/server.js
+++ b/src/server.js
@@ -192,27 +192,33 @@ function init(socket, client) {
 						});
 						return;
 					}
-					if (!Helper.password.compare(old || "", client.config.password)) {
-						socket.emit("change-password", {
-							error: "The current password field does not match your account password"
+
+					Helper.password
+						.compare(old || "", client.config.password)
+						.then(matching => {
+							if (!matching) {
+								socket.emit("change-password", {
+									error: "The current password field does not match your account password"
+								});
+								return;
+							}
+							const hash = Helper.password.hash(p1);
+
+							client.setPassword(hash, success => {
+								const obj = {};
+
+								if (success) {
+									obj.success = "Successfully updated your password, all your other sessions were logged out";
+									obj.token = client.config.token;
+								} else {
+									obj.error = "Failed to update your password";
+								}
+
+								socket.emit("change-password", obj);
+							});
+						}).catch(error => {
+							log.error(`Error while checking users password. Error: ${error}`);
 						});
-						return;
-					}
-
-					var hash = Helper.password.hash(p1);
-
-					client.setPassword(hash, function(success) {
-						var obj = {};
-
-						if (success) {
-							obj.success = "Successfully updated your password, all your other sessions were logged out";
-							obj.token = client.config.token;
-						} else {
-							obj.error = "Failed to update your password";
-						}
-
-						socket.emit("change-password", obj);
-					});
 				}
 			);
 		}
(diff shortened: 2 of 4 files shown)


src/helper.js
```diff
@@ -21,6 +21,7 @@ var Helper = {
 	password: {
 		hash: passwordHash,
 		compare: passwordCompare,
```
xPaw: Instead of creating separate async method, just update all the code to use async code.

src/server.js
```diff
@@ -354,3 +355,4 @@ function auth(data) {
 		}
 	}
 }
+
```
astorije: Could you remove this, please? :)

src/server.js
```diff
@@ -267,19 +267,20 @@ function localAuth(client, user, password, callback) {
...
-
-	return callback(result);
+			client.setPassword(hash, function(success) {
+				if (success) {
+					log.info(`User ${colors.bold(client.name)} logged in and their hashed password has been updated to match new security requirements`);
+				}
+			});
+		}
+		callback(matching);
+	}, (error) => log.error(`Error while checking users password. Error: ${error}`));
```
astorije: No need for parentheses around `error`, unless we set up ESLint to be strict like that.

src/helper.js
rockhouse: Done

test/tests/passwords.js
```diff
@@ -9,15 +9,26 @@ describe("Client passwords", function() {
...
 		let comparedPassword = Helper.password.compare(inputPassword, "$2a$11$zrPPcfZ091WNfs6QrRHtQeUitlgrJcecfZhxOFiQs0FWw7TN3Q1oS");
+		comparedPassword.then((result) => {
+			expect(result).to.be.true;
+		});
+	});
 
-		expect(comparedPassword).to.be.true;
+	it("hashed password should not match", function() {
+		// Generated with third party tool to test implementation
+		let comparedPassword = Helper.password.compare(inputPassword, "123");
```
xPaw: I would test against a real hash with a letter difference.

astorije:
FYI I'll give this one more try before merging.

test/tests/passwords.js
```diff
@@ -9,15 +9,26 @@ describe("Client passwords", function() {
 	it("hashed password should match", function() {
 		// Generated with third party tool to test implementation
 		let comparedPassword = Helper.password.compare(inputPassword, "$2a$11$zrPPcfZ091WNfs6QrRHtQeUitlgrJcecfZhxOFiQs0FWw7TN3Q1oS");
+		comparedPassword.then((result) => {
+			expect(result).to.be.true;
+		});
```
astorije: Actually, this doesn't test anything. Whether you test against `true` or `false`, it passes.
With promises, you need to do `return comparedPassword.then(...);`.

Alternative will be to use `chai-as-promised` to be able to do `return expect(comparedPassword).to.eventually.be.true;` but we will add that later, not worth doing it for 3 tests only :)

test/tests/passwords.js
astorije: I did this in my extra commit (not a single-letter difference but still).

src/server.js
```diff
@@ -267,19 +273,22 @@ function localAuth(client, user, password, callback) {
...
 	}
 
-	var result = Helper.password.compare(password, client.config.password);
-
-	if (result && Helper.password.requiresUpdate(client.config.password)) {
-		var hash = Helper.password.hash(password);
+	Helper.password
+		.compare(password, client.config.password)
+		.then(matching => {
+			if (Helper.password.requiresUpdate(client.config.password)) {
```
xPaw: Note from the future, this `if` check missed `matching` comparison and introduced a security vuln.

dgw: April Fools security!

PR #986, opened 2017-03-24
Show and remotely log out multiple connections

metsjeesus:
Adds a feature in private mode, to show current user connections. HTML side needs a bit fixing. Server side should have a better way to get hostnames on async. Added bonus, to remotely disconnect connections.  Example if you forgot lounge open in someone computer, now you can remotely disconnect them.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -213,7 +213,7 @@
 							<div class="col-sm-12">
 								<h1 class="title">Settings</h1>
 							</div>
-							<% if (!public) { %>
+							{{#unless public}}
 							<div class="col-sm-12">
 								<h2>Connected clients</h2>
 							</div>
@@ -222,7 +222,7 @@
 							<div class="col-sm-12">
 								<h2>Messages</h2>
 							</div>
-							<% } %>
+							{{/unless}}
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="motd">
--- a/client/index.html
+++ b/client/index.html
@@ -209,20 +209,12 @@
 						<button class="lt" aria-label="Toggle channel list"></button>
 					</div>
 					<div class="container">
+						<h1 class="title">Settings</h1>
+
 						<div class="row">
-							<div class="col-sm-12">
-								<h1 class="title">Settings</h1>
-							</div>
-							{{#unless public}}
-							<div class="col-sm-12">
-								<h2>Connected clients</h2>
-							</div>
-							<div id="connection_list" class="col-sm-12">
-							</div>
 							<div class="col-sm-12">
 								<h2>Messages</h2>
 							</div>
-							{{/unless}}
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="motd">
@@ -393,6 +385,18 @@
 								<textarea class="input" name="userStyles" id="user-specified-css-input" placeholder="You can override any style with CSS here"></textarea>
 							</div>
 						</div>
+
+						{{#unless public}}
+							<h2>Active sessions</h2>
+
+							<h3>Current session</h3>
+
+							<p id="current_connection_info"></p>
+
+							<h3>Other sessions</h3>
+
+							<div id="connection_list"></div>
+						{{/unless}}
 					</div>
 				</div>
 
--- a/client/index.html
+++ b/client/index.html
@@ -213,6 +213,13 @@
 							<div class="col-sm-12">
 								<h1 class="title">Settings</h1>
 							</div>
+							{{#if displayConnection}}
+							<div class="col-sm-12">
+								<h2>Connected clients</h2>
+							</div>
+							<div id="connection_list" class="col-sm-12">
+							</div>
+							{{/if}}
 							<div class="col-sm-12">
 								<h2>Messages</h2>
 							</div>
--- a/client/index.html
+++ b/client/index.html
@@ -213,13 +213,11 @@
 							<div class="col-sm-12">
 								<h1 class="title">Settings</h1>
 							</div>
-							{{#if displayConnection}}
 							<div class="col-sm-12">
 								<h2>Connected clients</h2>
 							</div>
 							<div id="connection_list" class="col-sm-12">
 							</div>
-							{{/if}}
 							<div class="col-sm-12">
 								<h2>Messages</h2>
 							</div>
--- a/client/index.html
+++ b/client/index.html
@@ -213,6 +213,7 @@
 							<div class="col-sm-12">
 								<h1 class="title">Settings</h1>
 							</div>
+							<% if (!public) { %>
 							<div class="col-sm-12">
 								<h2>Connected clients</h2>
 							</div>
@@ -221,6 +222,7 @@
 							<div class="col-sm-12">
 								<h2>Messages</h2>
 							</div>
+							<% } %>
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="motd">
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -44,6 +44,26 @@ $(function() {
 		pop.play();
 	});
 
+	// server signal to update connected client list
+	socket.on("update-clients-list", function(data) {
+		$("#connection_list").empty();
+		for (var i = 0 ; i < data.connection.length ; i++) {
+			data.connection[i].id = i;
+			$("#connection_list").append(
+				templates.connection_list(data.connection[i])
+			);
+			$("#connection_" + i).on("click" , function(input) {
+				socket.emit("remote-sign-out", {socket_id: input.target.value});
+			});
+		}
+	});
+
+	// server signal to remotely sign out
+	socket.on("sign-out", function() {
+		window.localStorage.removeItem("token");
+		location.reload();
+	});
+
 	// Autocompletion Strategies
 
 	const emojiSearchTerms = Object.keys(emojiMap);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -46,16 +46,30 @@ $(function() {
 
 	// server signal to update connected client list
 	socket.on("update-clients-list", function(data) {
+		$("#current_connection_info").empty().append(
+			templates.connection_list({
+				hostname: data.currentSession.hostname,
+				ip: data.currentSession.ip,
+				socketId: data.currentSession.socketId,
+				userAgent: data.currentSession.userAgent,
+			})
+		);
+
 		$("#connection_list").empty();
-		for (var i = 0 ; i < data.connection.length ; i++) {
-			data.connection[i].id = i;
+		data.otherSessions.forEach((connection, index) => {
 			$("#connection_list").append(
-				templates.connection_list(data.connection[i])
+				templates.connection_list({
+					id: index,
+					hostname: connection.hostname,
+					ip: connection.ip,
+					socketId: connection.socketId,
+					userAgent: connection.userAgent,
+				})
 			);
-			$("#connection_" + i).on("click" , function(input) {
-				socket.emit("remote-sign-out", {socket_id: input.target.value});
+			$(`#connection_${index}`).on("click", (input) => {
+				socket.emit("remote-sign-out", {socketId: input.target.value});
 			});
-		}
+		});
 	});
 
 	// server signal to remotely sign out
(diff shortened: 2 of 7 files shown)


src/server.js
```diff
@@ -354,3 +367,74 @@ function auth(data) {
...
+					data.active_host = true;
+				} else {
+					data.active_host = false;
+				}
+				send_data.connection.push(data);
+			}
+			manager.sockets.of("/").connected[connection_list[j].socket_id].emit("update-clients-list", send_data);
+		}
+	});
+}
```
MaxLeiter: Need line break

src/server.js
```diff
@@ -149,10 +149,22 @@ function init(socket, client) {
...
 		});
 		client.clientAttach(socket.id);
 
 		socket.on(
+			"remote-sign-out",
+			function(data) {
+				var connection_list = Object.keys(client.attachedClients);
+				for (var i = 0; i < connection_list.length; i++) {
+					if (data.socket_id === connection_list[i]) {
+						manager.sockets.of("/").connected[connection_list[i]].emit("sign-out", {});
```
xPaw: Merely emitting an event for client to act upon is not enough, you *have* to kill the socket connection on the server.

client/index.html
```diff
@@ -205,6 +205,13 @@ <h1 class="title">
 							<div class="col-sm-12">
 								<h1 class="title">Settings</h1>
 							</div>
+							<% if (displayConnection === true) { %>
```
xPaw: When the option is disabled, all you do is hide this in HTML, but all the socket/dns lookups are still performed.

Also, I'm not entirely sure the option is even needed, perhaps just always have it work in private mode.

src/server.js
```diff
@@ -354,3 +367,76 @@ function auth(data) {
 		}
 	}
 }
+
+
+function IterateOver(list, iterator, callback) {
```
xPaw: Not a fan of this implementation.

src/server.js
metsjeesus: Idea was, emitting event to client, client disconnects itself and then server gets disconnection event. This way you get confirmation that client got disconnected. Just emitting disconnection event wont log out the client and you see recent history there.

client/index.html
metsjeesus: Because i struggled with html and css side, I hide it by default. Some help with design would be nice.

src/server.js
metsjeesus: I would use promises, if its allowed. I remember something about avoiding them in lounge, so i made chunky workaround. If it is allowed, a example would be nice.

astorije: You can use promises. I don't think we are already, but #985 does.

metsjeesus:
Changed callbacks to promise. Better?

src/server.js
astorije: > Merely emitting an event for client to act upon is not enough, you have to kill the socket connection on the server.

Why is that, @xPaw? Not arguing, trying to understand the implications of one vs. the other.

xPaw: If the connection exists on the server, but the client died, you will not be able to remove it.

astorije:
This is a really cool feature, @metsjeesus, thanks!

I am seeing a bug when connecting from my phone to an instance started on my laptop (how I test PRs locally 😄). Server crashes with:

```js
.../lounge/src/server.js:429
			if (err && host.length) {
			               ^

TypeError: Cannot read property 'length' of undefined
    at .../lounge/src/server.js:429:19
    at QueryReqWrap.asyncCallback [as callback] (dns.js:62:16)
    at QueryReqWrap.onresolve [as oncomplete] (dns.js:216:10)
```

This does not appear from laptop to itself.

Outside of that, a couple comments:

- Regarding the config option, I agree with @xPaw: this should be there in private and that's all. You can get rid of `displayConnection` 🎉 
- Regarding the UI, don't worry about that for now. We'll come up with adequate UI/UX once the server side works flawlessly (though I would definitely see this under the CSS block until we have [better settings](https://github.com/thelounge/lounge/issues/85)).

src/server.js
astorije: So, both, right? Emit disconnect event to the client, and after a timeout kill the connection on the server? Does this seem acceptable @xPaw?

metsjeesus: I thought that, if client dies, server itself emits disconnect event after timeout and removes connection. How you can test it?

client/index.html
```diff
@@ -206,6 +206,11 @@ <h1 class="title">
 								<h1 class="title">Settings</h1>
 							</div>
 							<div class="col-sm-12">
+								<h2>Connected clients</h2>
+							</div>
+							<div id="connection_list" class="col-sm-12">
+							</div>
```
astorije: Shouldn't this be displayed in private mode only?

astorije:
Hey @metsjeesus, I let myself in and made some UI changes since you mentioned you were having issues with HTML/CSS. I think it's nice like that, let me know what you think :)

Regardless of that remote log out / event @xPaw mentioned, a few things:

- When I connect from my mobile, if it loses connection, its line does not get removed from the list.
- At the moment, opening the app on a same device, same browser but different tabs creates multiple entries. That means that if I disconnect another tab, I get kicked out (but not instantly, upon refresh, which is also weird. These should not create a new line, but different browser / private connection should. Not sure how feasible this is.
- It would be nice to have more details: last time something happened or connection duration or something time-related, browser, OS, maybe geolocation (but at the moment my link to a geo-ip service will do), etc. That as well I'm not sure how feasible.

metsjeesus:
Seems you dont like underscore on variable names. Now i know.

I try to understand what happens on mobile. If you lose connection, `disconnection` event should be fired on server and it will repopulate all connected clients with new info. On other hand, disconnected client will not see changes, because client is disconnected and wont recieve data. 

`location.reload();` should force refresh on disconnecting every connection. And it removes token too from localStorage. Because localStorage is shared between all the tabs, other tabs wont recieve reload event, but lose the token. If you do refresh, you generate new token and need to log in again. 

It is possible to have more info about client. Anything what socket can give you can use, http headers are bit tricky, but usable too. Currently, client wont save any info about timings, so it needs to be implemented to have them. Moment, lounge lists every connection and every open tab counts as one. Maybe group them as (2 tabs) or so?

astorije:
> Seems you dont like underscore on variable names. Now i know.

Ahah, nothing like that, it's just for consistency sake :)
Also, in JS it's mostly camelCase convention, but it this were Ruby or Python, it would not be.

xPaw:
Welcome to networking where nothing ever works as expected. Reason for disconnect event not being called is silent network failures or incorrectly closed sockets. This is why you always have to just terminate the connection in server without waiting for client to respond (you can wait to ACK that it got a final logout event, but it has to be on a timeout)

metsjeesus:
If i get it right, socket.io plays ping pong. By default, `pingTimeout` is `60000` and `pingInterval` is `25000` milliseconds. What gives 60-85 seconds until disconnect event happens. Question is, do both client and server fire disconnection event? There should be a reliable way to test it?

src/server.js
```diff
@@ -354,3 +367,69 @@ function auth(data) {
...
+					.emit("update-clients-list", sendData);
+			}
+		})
+		.catch(function(err) {
+			log.error("Unable to resolve hostname", err);
+		});
+}
+
+function hostnamePromise(data) {
+	if (Array.isArray(data)) {
```
xPaw: Seems pointless. Split into different functions so that you don't need this check.

metsjeesus: My idea was, that i allow `hostnamePromise("192.168.1.1")` and `hostnamePromise(["192.168.1.1"])`. One way or other, you will need `IsArray` check.

xPaw: I understand that, while nice, this is an internal function, and we define how and where it gets called, so we don't need to over engineer anything. If JS had function overloads based on different types, then this wouldn't be an issue :)

astorije:
@metsjeesus are you still stuck with this? Also, mind rebasing on `master` and fixing conflict?

metsjeesus:
@astorije I know what should be done, use cookies per connections, not one for all. I looked on it and it will interfere with https://github.com/thelounge/lounge/pull/1011 what is made to be a start of rewriting the authentication system. I feel that the change of the cookie managment is too big for that pull.  On server side, it changes alot how auth and client works. I would rather keep those things on separate PR.

astorije:
It looks like you made some progress, @metsjeesus. Are you still stuck with this / do you need some more help? What's the next step? 😃

metsjeesus:
@astorije next step would be rewriting auth system with better cookie managment and that needs async calls from #1011 to be functional. From html/css style, feel free to change as you need.

astorije:
> next step would be rewriting auth system with better cookie managment and that needs async calls from #1011 to be functional.

Is this actually needed for this PR or just a nice-to-have? It seems to me like this PR could do without, but I haven't deeply looked at the server code so I might be wrong. Also interested in @xPaw's opinion.

> From html/css style, feel free to change as you need.

Did you delete my commit from [27 days ago](https://github.com/thelounge/lounge/pull/986#issuecomment-291051904)? 😱 I had made a lot of changes, I really hope not :( I don't have your copy of the repo on my laptop anymore so if you force-pushed over my commit, then my changes are in the trash...

astorije:
Alright, it turns out, I had this partially git-stashed somewhere, so it was not completely lost. Let's not do that again 😅

@metsjeesus, I added a commit to your branch (*Alert! Alert!* `git pull` your branch!) and it now looks like this:

screen shot 2017-04-30 at 03 23 18

It's not perfect of course, but I think that's way slicker :D
Good luck with the rest!

client/js/lounge.js
```diff
@@ -117,6 +118,40 @@ $(function() {
...
+				})
+			);
+			$(`#connection_${index}`).on("click" , input => {
+				socket.emit("remote-sign-out", {socketId: input.target.value});
+			});
+		});
+	});
+
+	// server signal to remotely sign out
+	socket.on("sign-out", function() {
```
MaxLeiter: arrow function?

client/views/connection_list.tpl
```diff
@@ -0,0 +1,11 @@
+<p class="clearfix">
+	{{#if id includeZero=true}}
+		<button class="btn pull-right" id="connection_{{id}}" value="{{socketId}}">Disconnect</button>
+	{{/if}}
+
+	<strong>{{hostname}}</strong>
+	<br>
+	<a class="" href="https://geoiptool.com/?ip={{ip}}">{{ip}}</a>
```
MaxLeiter: Not sure if we should rely on a third party service -- do we need to link to the tool?

src/server.js
```diff
@@ -160,10 +161,31 @@ function init(socket, client) {
 
 		socket.on("disconnect", function() {
 			client.clientDetach(socket.id);
+			sendConnectionInfo(client);
 		});
 		client.clientAttach(socket.id);
 
 		socket.on(
+			"remote-sign-out",
+			function(data) {
```
MaxLeiter: arrow function?

client/views/connection_list.tpl
astorije: There is no way we can ever have this in-repo. Note that it's not fetching anything, just adding a link to it. I'd rather have this than nothing.

astorije:
@metsjeesus, now that #1199 is merged, we'll definitely want this PR! :D
Actually, #1199 came with a feature regression (it's not possible to log out other sessions anymore) and your PR re-introduces it, so we know we cannot release the next version without your work :)

Do you know what else is needed on this PR? Do you need help with the rebase and fixing conflicts?
Make sure you pull locally before committing/force-pushing or you'll lose [my commit](https://github.com/thelounge/lounge/pull/986/commits/2d4a495721315932518d77fe1ddba49edcd57340) again :)

Thanks for your help!!

astorije:
@metsjeesus, do you think you might be able to take this up soon?

astorije:
Rebased on `master` to fix conflicts (that was quite the pain!). It's not working right now and it's missing 30 lines compared to before rebase (kinda expected it with such a rough rebase), but I'll figure it out.

xPaw:
I also did some separate work: https://github.com/thelounge/lounge/compare/xpaw/session-list

But since sessions that don't use 'remember me' will not generate a token, we have to combine both approaches.

astorije:
@xPaw, now that we don't have the "Remember me" option, does your branch need any work?

astorije:
@metsjeesus, @xPaw, pretty ping? 😁

xPaw:
Superceded by #1536

PR #987, opened 2017-03-27
Monkey patch for iso/utf8 miscoded characters

metsjeesus:
irc-framework updated and old way dont work. Old fix is here: https://github.com/thelounge/lounge/pull/656 

That pull has other problems, its hard to use in lounge, because deep dependencies. This monkey patch is easier to use and update on future. Its not a correct way to fix things, but its working and easier to test. Do not merge the pull, but keep it open, so people can see and use it, when needed.

Howto use this

1. Make fork from lounge.
2. (optional) Make a branch.
3. Make a pull request
4. Use your master or optional branch as base
5. Use metsjeesus:unicode_fix as head fork.
6. Merge
7. Clone your version of the lounge as source in your server and run it. ( See [how ](https://github.com/thelounge/lounge#running-from-source))
8. ???
9. Profit!

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,8 @@
     "request": "2.81.0",
     "semver": "5.3.0",
     "socket.io": "1.7.3",
-    "spdy": "3.4.4"
+    "spdy": "3.4.4",
+    "string.prototype.at": "^0.2.0"
   },
   "devDependencies": {
     "babel-core": "6.23.1",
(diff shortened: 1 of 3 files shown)


xPaw:
This is definitely a monkey fix and not something I'd want to have. It's not productive to have this PR open without an intention to merge too.

astorije:
It does seem like we wouldn't want this in `master`, is it a correct assumption, @metsjeesus?

MiniDigger:
the best place for that is probably the wiki. you can link the branch there. keeping this is a open pr is bad since nobody will really search PRs for a solution for their problem (maybe google will bring them here, but google would also bring them to the wiki)

PR #988, opened 2017-03-28
Restrict extra parentheses inside argument list of a function call/definition/etc.

astorije:
Things like `console.log( "Hello world" );` are now prohibited.

While I was playing with #977 and the dependency it refers to, I noticed we allow these for now (but thankfully we have none in-repo).


astorije:
Don't mind me, I'm useless.
#977 allows this indeed, but #975 that was created/merged after added this already :man_facepalming:

PR #996, opened 2017-03-30
Adds http to https redirect option

metsjeesus:
I know, correct way is to use reverse proxy. I reroute ports, http and https ports, to 2 random ports to my rasperry pi, one listens lounge, other redirects, no root access required. Maybe its useful?

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -272,6 +272,16 @@ module.exports = {
 		enable: false,
 
 		//
+		// Enable HTTP to HTTPS redirect port.
+		// port number to enable
+		// null to disable
+		//
+		// @type     int
+		// @default  null
+		//
+		http_redirect_port: null,
+
+		//
 		// Path to the key.
 		//
 		// @type     string
--- a/src/server.js
+++ b/src/server.js
@@ -73,6 +73,15 @@ in ${config.public ? "public" : "private"} mode`);
 		authFunction = ldapAuth;
 	}
 
+	if (config.https.http_redirect_port) {
+		// Redirect from httpt to https
+		var http = require("http");
+		http.createServer(function(req, res) {
+			res.writeHead(301, {Location: "https://" + req.headers.host + req.url});
+			res.end();
+		}).listen(config.https.http_redirect_port);
+	}
+
 	var sockets = io(server, {
 		serveClient: false,
 		transports: config.transports


astorije:
I'm wondering what the use case of having 2 ports is (unless I'm missing network knowledge, which is faaar from impossible 😅).

Ideally, if HTTPS is enabled, assuming port used is `9000` and domain is `example.com`, The Lounge should redirect from `http://example.com:9000/` to `https://example.com:9000/`.
I wonder however if that's possible (after all, usual HTTP and HTTPS ports *are* 2 different ones...).

And if that's possible, there shouldn't be an option, just redirect when HTTPS is on.

metsjeesus:
In my case, I type in `example.com` , what ends up in `http://example.com/` what needs to redirect with that code to `https://example.com/` . Router is set up to listen both 80 and 443 and pass them on some ports on my pi machine, like 9000 for https lounge and 9001 for http server for redirecting. If port 80 is closed,  `example.com` would not show anything.

xPaw:
I'm not sure this is useful to have. I'm not aware of software that has https and allows http redirections (besides web servers and proxies).

astorije:
@xPaw, what do you think of this?
I'm not too excited in the current state.

If we were to do something, I'd say this is 2-fold:
- When HTTPS is enabled, augment non-HTTPS into HTTPS (I googled for a full 5 seconds and it looks like this is not unusual/impossible).
- When HTTPS is enabled *and* port is 80 (or 443?), listen to traffic on 80 and redirect non-HTTPS to 443.

Note that I might be completely missing something as I've never used the webserver directly without proxying it, nor have I ever served web services I host in ports other than usual 80/443 through Nginx.

xPaw:
I think we shouldn't be doing this at all.

dgw:
The config file example should warn against enabling HTTPS with port=80. It might even be worth displaying a warning on startup if HTTPS is enabled and port is set to 80. Otherwise, it's probably best left to the user rather than listening on two ports when the user only specifies one—especially since ports below 1024 usually require root (or special privileges) on Unix-like systems to even start a socket listener.

Automatically upgrading HTTP to HTTPS for the port The Lounge is already listening on would be useful, though.

xPaw:
This could work. https://github.com/mattkrea/express-https/blob/master/index.js

astorije:
Who wants to give it a try? 😅

xPaw:
I came up with this, but it doesn't work.
```js
function allRequests(req, res, next) {
	if (Helper.config.https.enable && req.protocol === "http") {
		if (Helper.config.port === 443) {
			return res.redirect(302, `https://${req.hostname}${req.originalUrl}`);
		}
		
		return res.redirect(302, `https://${req.hostname}:${Helper.config.port}${req.originalUrl}`);
	}
	
	res.setHeader("X-Content-Type-Options", "nosniff");
	return next();
}
```

metsjeesus:
If im not mistaken, browser will show a error if you try to use ssl on 80 and http on 443.  

Another thing is, redirecting on `${Helper.config.port}` might not be good solution, because lounge can open a port what is accessible only on local network, not from outside.

xPaw:
Redirecting without port is problematic too. I think handling redirects on our side is bound to bring more problems and we shouldn't just do it.

astorije:
I think I'm with @xPaw on that one. Unless you find a straightforward way to do that without known issues (because it's not specific to The Lounge, so if there is a solution, it's definitely been solved and documented somewhere), I think we should avoid tempting the devil.
We have enough problems maintaining the current codebase as it is 😅.

Closing for now, but feel free to open a new PR (and link to this one for reference) if you find something else!

PR #1003, opened 2017-04-01
Store channel keys

### LINKED ISSUE #412: Join keys are not stored

xPaw:
Fixes #412 and supercedes #715. This PR gets channel key from channel mode, instead of hooking join command.

cc @spookhurb, @DanielOaks

### DIFF
--- a/src/plugins/irc-events/mode.js
+++ b/src/plugins/irc-events/mode.js
@@ -1,13 +1,40 @@
 "use strict";
 
-var _ = require("lodash");
-var Chan = require("../../models/chan");
-var Msg = require("../../models/msg");
+const _ = require("lodash");
+const Chan = require("../../models/chan");
+const Msg = require("../../models/msg");
 
 module.exports = function(irc, network) {
-	var client = this;
+	const client = this;
+
+	// The following saves the channel key based on channel mode instead of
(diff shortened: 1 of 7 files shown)


astorije:
Just like the parser PR, I think the commit author should be @spookhurb :)

xPaw:
I didn't re-use his code (besides test), and it uses a different approach.

src/plugins/irc-events/mode.js
```diff
@@ -6,6 +6,29 @@ var Msg = require("../../models/msg");
...
+			return;
+		}
+
+		const targetChan = network.getChannel(data.channel);
+		if (typeof targetChan === "undefined") {
+			return;
+		}
+
+		for (var i = 0; i < data.modes.length; i++) {
+			var mode = data.modes[i];
```
astorije: This is a minor nitpick, but any reason why not to use `data.modes.forEach(mode => { ... });` instead?
Also, `const` when possible?

xPaw: Done.

PR #1009, opened 2017-04-04
Add support for banlist messages

### LINKED ISSUE #1008: Add support for banlist messages
We ignore banlist messages which means we can't view them. Ideally a GUI would be nice eventually, but we should at least support seeing the messages.

YaManicKill:
Fixes #1008

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1005,13 +1005,16 @@ kbd {
 	padding-left: 20px;
 }
 
-#chat table.channel-list {
+#chat table.channel-list,
+#chat table.ban-list {
 	margin: 5px 10px;
 	width: calc(100% - 30px);
 }
 
 #chat table.channel-list th,
-#chat table.channel-list td {
+#chat table.ban-list th,
+#chat table.channel-list td,
+#chat table.ban-list td {
 	padding: 5px;
 	vertical-align: top;
 	border-bottom: #eee 1px solid;
@@ -1022,7 +1025,10 @@ kbd {
 }
 
 #chat table.channel-list .channel,
-#chat table.channel-list .topic {
+#chat table.channel-list .topic,
+#chat table.ban-list .hostmask,
+#chat table.ban-list .banned_by,
+#chat table.ban-list .banned_at {
 	text-align: left;
 }
 
--- a/client/index.html
+++ b/client/index.html
@@ -556,6 +556,15 @@
 
 						<div class="help-item">
 							<div class="subject">
+								<code>/banlist</code>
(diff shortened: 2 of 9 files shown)


MaxLeiter:
is it common to push to lobby? not to the channel the command is being called?

xPaw:
Could use the special Chan type, same as /list.

YaManicKill:
> is it common to push to lobby? not to the channel the command is being called?

In my experience, most IRC clients only push to the channel if it is a message directly.

I mean, ideally we would want to show a modal or something, with buttons to remove from the banlist, but this is just a first version so that we support it **at all**.

> Could use the special Chan type, same as /list.

I guess we could, yeah.

I think the thing is we don't want to end up just everything being a channel, because then we will end up not doing it in a nicer way, as we are a web app and we want to do things nicely for the user, not just in the simplest way possible.

But if you think that's the best thing to do for this first version, then I'm happy to do that.

YaManicKill:
Well, I've pushed a commit to have a special chan for banlist entries. I'm not using any special message format (like the chanlist is). It's probably best if I do, and I pass the other information through to the client for showing.

For reference, this is what we get back from the server:

```
[ { channel: '#channelthatdoesntexist',
    banned: 'test!*@*',
    banned_by: 'ymk2!~ymk2@29.93.2.81.in-addr.arpa',
    banned_at: '1491320726' } ]
```

So it's probably worth showing the banned by and the banned at. I'll add that. See what you think about this.

astorije:
@YaManicKill, mind adding an entry in the "commands" section of the help window, between `/back` and `/ctcp`?

YaManicKill:
Alright, I have completely changed this.

Now it is a custom chan with custom messages, basically the same as how `/list` works, but obviously showing useful information for the banlist.

`/banlist` also wasn't a command we supported, so I added that so people don't have to remember the correct mode.

I have also added `/banlist` to the help section, as requested.

It works quite nicely now, I think.

YaManicKill:
>Mind rebasing on master to take into account the recent changes to special channels by #1018?

Yeah, I'll do that soon.

YaManicKill:
Rebased onto master, and using the new stuff from #1018.

astorije:
@YaManicKill just so you know I didn't make any changes but when I wanted to test I got that `undefined` bug back so I rebased with `master` to be able to test then figured I might as well send that up-to-date version.

YaManicKill:
Alright, I have added the command to the autocompletion list, and I have rebased and squashed. I think this is good to go now, then,

PR #1011, opened 2017-04-04
Change fs.sync operations to promises and async methods.

metsjeesus:
Its my first try on promises, so all work is in progress, I hope i did not miss some weird method calls somewhere.


xPaw:
Nice! However we would want to just get rid of sync methods (instead of creating separate ones).

xPaw:
I'm wondering how eslint passed with that spacing and indentation though, it's all over place.

metsjeesus:
I did not have the balls to remove all the sync methods.  They work very differently. Might be some use somewhere?

For eslint, `./node_modules/.bin/eslint . --fix`

metsjeesus:
Next thing would be authentication, it looks very ugly there.

astorije:
I don't know about other @thelounge/maintainers, but I know I'm not going to be comfortable reviewing/merging 600+ lines of changes like this :/

The easiest you could do this IMO, @metsjeesus, would be to split this. For example, looking at your commit list, a good start would be to make the changes to user autoload its own PR.
That would help focusing on a smaller portion of the code, identifying misconfiguration of our linter, testing fewer features at a time, etc. And also if we would identify earlier changes required, leading to less work from you.

What do you think about this?

metsjeesus:
It is possible to do "smaller" portions. I wanted a system, where you can add, update and remove users with promises and remove those sync fs call dependencies and it pretty much ends up like it is. There are some methods what should be promises but currently are not, because its not needed. I am aware, that this will have a long testing period and probably goes to 3.x.x version of the lounge if ever. This is needed for my personal project, so i can test a bit there.

About linting, its my first iteration, i have some ideas how to make things more usable and readable. If you have some examples, how to do it better, let me hear.

metsjeesus:
Figured out some cases when promises wont work so well as intended. So i made some adjustments on those edge cases when `this` is lost on promise hell. Is there a way that code is simply replaced, not just weirdly chunked on github? Its really hard to see in github whats going on there.

xPaw:
Will you be able to split this into smaller PRs? I really would like to land this.

src/clientManager.js
```diff
@@ -36,135 +36,233 @@ ClientManager.prototype.findClient = function(name, token) {
...
-				users.push(file.replace(".json", ""));
-			}
-		});
-	} catch (e) {
-		log.error("Failed to get users", e);
-		return;
-	}
-	return users;
-};
+ClientManager.prototype.loadUsersPromise = function(data) {
```
xPaw: Doesn't feel like a necessary function.

src/clientManager.js
```diff
@@ -36,135 +36,233 @@ ClientManager.prototype.findClient = function(name, token) {
...
-	let user = this.readUserConfig(name);
-	const currentUser = JSON.stringify(user, null, "\t");
-	_.assign(user, opts);
-	const newUser = JSON.stringify(user, null, "\t");
-
-	// Do not touch the disk if object has not changed
-	if (currentUser === newUser) {
-		return callback ? callback() : true;
-	}
+ClientManager.prototype.addUserPromise = function(data) {
```
xPaw: Why did you replace params with a single `data`?

src/clientManager.js
```diff
@@ -36,135 +36,233 @@ ClientManager.prototype.findClient = function(name, token) {
...
-	fs.writeFile(Helper.getUserConfigPath(name), newUser, (err) => {
-		if (err) {
-			log.error("Failed to update user", err);
-		}
+ClientManager.prototype.updateUserPromise = function(data) {
+	let self = data.this || this;
+	let name = data.name || undefined;
+	let opts = data.opts || undefined;
+	return new Promise((resolve, reject) => {
+		self.getUsersPromise()
```
xPaw: Get rid of this `getUsers`, all calls to `updateUserPromise` seem to already check for user existance.

src/clientManager.js
```diff
@@ -36,135 +36,233 @@ ClientManager.prototype.findClient = function(name, token) {
...
-		fs.unlinkSync(Helper.getUserConfigPath(name));
-	} catch (e) {
-		throw e;
-	}
-	return true;
+ClientManager.prototype.readUserConfigPromise = function(data) {
+	let self = data.this || this;
+	let name = data.name || undefined;
+	return new Promise((resolve, reject) => {
+		self.getUsersPromise()
```
xPaw: This shouldn't be necessary too.

src/command-line/start.js
```diff
@@ -15,26 +15,29 @@ program
...
+				if (options.public) {
+					mode = true;
+				} else if (options.private) {
+					mode = false;
+				}
 
-		if (!mode && !users.length && !Helper.config.ldap.enable) {
-			log.warn("No users found.");
-			log.info(`Create a new user with ${colors.bold("lounge add <name>")}.`);
+				if (!mode && !users.length && !Helper.config.ldap.enable) {
```
xPaw: Can you move this check to `autoloadUsers`? Then `start` CLI command won't need to deal with users at all.

src/server.js
```diff
@@ -343,7 +348,7 @@ function auth(data) {
 			if (success) {
 				if (!client) {
 					// LDAP just created a user
-					manager.loadUser(data.user);
+					manager.loadUser({name: data.user});
```
xPaw: `loadUserPromise`?

xPaw:
Couple concerns:
1. Whitespace is all over the place. I've added a couple of extra eslint rules here to help: https://github.com/thelounge/lounge/compare/eslint4
2. Promise functions replace params into a singular `data`, I don't think we need to do this.
3. Renaming functions to `*Promise`, I think we can keep the same names.

metsjeesus:
Made some tests in mocha, but no actual test on server. I hope im on right direction with that.

xPaw:
This is quite a blocking refactor, and it's already conflicting with a lot of files. I think we want to eventually use async operations, we have to do in incremental steps.

I will close this PR as we're unlikely to get anywhere with this.

PR #1012, opened 2017-04-05
Alternative way to use http and https on same port

metsjeesus:
@astorije closed last pull request https://github.com/thelounge/lounge/pull/996 and i found a new way to fix the problem. So there it is. If hostname is set, it will redirect to that hostname. Http works same way, https is changed. On https, server now allows both http and https connections on same port. If connection is not secure, redirects to https. Its important, that same lounge server can response different hostnames, thats the reason why i dont use host as base in config file, because it locks me with one hostname. 

Other then that, now i let router redirect both 80 and 443 ports to 9000 on my pi lounge server. -1 port open, added bit security.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -272,6 +272,16 @@ module.exports = {
 		enable: false,
 
 		//
+		// Enable HTTP to HTTPS redirect port.
+		// port number to enable
+		// null to disable
+		//
+		// @type     int
+		// @default  null
+		//
+		http_redirect_port: null,
+
+		//
 		// Path to the key.
 		//
 		// @type     string
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -22,6 +22,15 @@ module.exports = {
 	host: undefined,
 
 	//
+	// hostname for the web server to redirect.
+	// Setting this to undefined will disable hostname redirecting.
+	//
+	// @type     string
+	// @default  undefined
+	//
+	hostname: undefined,
+
+	//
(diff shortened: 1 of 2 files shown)


metsjeesus:
Seems Travis could not make a build. I dont figure out whats wrong there? any help?

metsjeesus:
Seems it wont work well. It will run for a while, but when connection gets worse, you will get error like this and lounge stops. I did not find a way to replicate it consistently though. And all ways to supress the error seems not to work either. Used node v7.8.0. 

```
Uncaught Error: read ECONNRESET

FROM
Socket.emit (events.js:163:7)
emitErrorNT (net.js:1284:8)
process._tickCallback (internal/process/next_tick.js:80:11)
Illegal instruction
```

astorije:
I would defer the decision to bring this in or not to @xPaw (and/or @maxpoulin64 but he is mostly away/busy). I still can't think of an ideal solution for this and this is unfortunately not a case of "better is the enemy of good" where we can go for something that's just good enough as it could consistently break. Just my 2c but happy to go with the flow.

xPaw:
Just as with previous PR, I think we ahould leave this work to reverse proxies.

astorije:
I am with @xPaw on this. Sorry @metsjeesus, but hope you understand. For simple use cases, provided server should work, but anything more than basic needs should go through a reverse proxy, we can't re-implement Nginx or the like in our codebase.

PR #1015, opened 2017-04-06
Use Referrer-Policy header instead of CSP referrer

astorije:
According to MDN:

> `referrer`
>   Used to specify information in the referer (sic) header for links away from a page.
>   Use the Referrer-Policy header instead.

See:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/referrer
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

This fixes the following warning in JS console:

screen shot 2017-04-06 at 02 26 38

**Note that I have merely switched from the old to new format, I have not tested security implications of this.** My understanding is there should be no implications beyond renaming it, but I could be wrong.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -131,7 +131,8 @@ function index(req, res, next) {
 			return css.slice(0, -4);
 		});
 		var template = _.template(file);
-		res.setHeader("Content-Security-Policy", "default-src *; connect-src 'self' ws: wss:; style-src * 'unsafe-inline'; script-src 'self'; child-src 'self'; object-src 'none'; form-action 'none'; referrer no-referrer;");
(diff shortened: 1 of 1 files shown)


astorije:
@xPaw do you want to test it before merging or should we merge as is?

xPaw:
We can probably drop `rel` from links in favour of the header.

astorije:
¯\_(ツ)_/¯
Feel free to take over this PR if you don't want to merge as is. You clearly know more than me and I can't really find time to explore things as much as I'd like these days :/

xPaw:
That was just a suggestion, looking at browser support of the headers, this PR is fine.

astorije:
Cool

PR #1016, opened 2017-04-06
Unread marker takes hidden messages into account.

### LINKED ISSUE #604: Unread marker shows up incorrectly when joins/parts are hidden

swordbeta:
Stops showing the unread messages marker when there are hidden joins/parts/quits/nick changes which is rather annoying. Would close #604, went with the solution suggested by xPaw.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -444,7 +444,9 @@ $(function() {
 			]);
 
 		if (data.msg.self
-			|| container.find("div:visible").last().hasClass("unread-marker")) {
+			|| container.find("div:visible").last().hasClass("unread-marker")
+			|| (container.find("div:visible").last().hasClass("date-marker")
+				&& container.find("div:visible").last().prev().hasClass("unread-marker"))) {
 			container
 				.find(".unread-marker")
 				.appendTo(container);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -442,11 +442,11 @@ $(function() {
 				target,
 				data
 			]);
-
+		var lastVisible = container.find("div:visible").last();
 		if (data.msg.self
-			|| container.find("div:visible").last().hasClass("unread-marker")
-			|| (container.find("div:visible").last().hasClass("date-marker")
-				&& container.find("div:visible").last().prev().hasClass("unread-marker"))) {
+			|| lastVisible.hasClass("unread-marker")
+			|| (lastVisible.hasClass("date-marker")
+				&& lastVisible.prev().hasClass("unread-marker"))) {
 			container
 				.find(".unread-marker")
 				.appendTo(container);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -443,7 +443,8 @@ $(function() {
 				data
 			]);
 
-		if (data.msg.self) {
+		if (data.msg.self
+			|| container.find("div:visible").last().hasClass("unread-marker")) {
 			container
 				.find(".unread-marker")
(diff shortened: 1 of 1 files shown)


client/js/lounge.js
```diff
@@ -443,7 +443,9 @@ $(function() {
 				data
 			]);
 
-		if (data.msg.self) {
+		var lastVisibleChild = container.find("div:visible:last-child");
+		if (data.msg.self
+			|| (lastVisibleChild.length === 1 && lastVisibleChild.hasClass("unread-marker"))) {
```
xPaw: The `length` check isn't really needed.

client/js/lounge.js
```diff
@@ -443,7 +443,9 @@ $(function() {
 				data
 			]);
 
-		if (data.msg.self) {
+		var lastVisibleChild = container.find("div:visible:last-child");
```
xPaw: I'm wondering what kind of performance such a selector will have.

swordbeta:
@xPaw Removed the length check.

Not sure about performance, according to [this jsperf](https://jsperf.com/jquery-visible-test) `:visible` makes things significantly slower. I wonder if this will actually matter in the lounge though. And according to [this jsperf](https://jsperf.com/last-child-vs-jquery-last) `.last()` is faster than `:last-child`, so I changed that.

client/js/lounge.js
```diff
@@ -443,7 +443,8 @@ $(function() {
 				data
 			]);
 
-		if (data.msg.self) {
+		if (data.msg.self
+			|| container.find("div:visible").last().hasClass("unread-marker")) {
```
xPaw: Also, this might not work if there is a date marker appended in the code above.

astorije: Is it something that an extra check for the date marker could fix? Not sure.

astorije:
@xPaw, mind helping with this? I'm not sure how to go with your comment. @swordbeta, do you need help or just time to amend this?

astorije:
@xPaw, ping?

swordbeta:
@xPaw @astorije Updated, I'm not sure about performance though

xPaw:
Yeah that definitely looks like performance destroyer.

swordbeta:
@xPaw Updated it, any other suggestions that might improve performance?

client/js/lounge.js
```diff
@@ -442,8 +442,11 @@ $(function() {
 				target,
 				data
 			]);
-
-		if (data.msg.self) {
+		var lastVisible = container.find("div:visible").last();
+		if (data.msg.self
+			|| lastVisible.hasClass("unread-marker")
+			|| (lastVisible.hasClass("date-marker")
+				&& lastVisible.prev().hasClass("unread-marker"))) {
```
xPaw: Would this prev() need visible check too? I understand why its necessary to check previous element, but this does not feel like performant code 😞

astorije:
@xPaw, mind reviewing/approving this first before I give it a review?

astorije:
@xPaw, ping?

xPaw:
@astorije I guess we can try merging this, there's really no better way of fixing it I don't think.

astorije:
Thank you for your help on this, @swordbeta!

astorije:
Hey @swordbeta, we have free sticker packs for our contributors now!

PR #1018, opened 2017-04-06
Improve channels list.

swordbeta:
- Set fixed width to channel and users column.
- Sort by number of users in channel.
- Executing /list multiple times wont show multiple tables. (Also clears loading message)
- Scroll to top after /list is finished.
- Limit channels to 500. Scrolling through 1k is very slow on my system.

screenshot from 2017-04-06 18-27-22

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -421,6 +421,10 @@ $(function() {
 		var target = "#chan-" + data.chan;
 		var container = chat.find(target + " .messages");
 
+		if (data.msg.type === "channel_list") {
+			$(container).empty();
+		}
+
         // Check if date changed
 		var prevMsg = $(container.find(".msg")).last();
 		var prevMsgTime = new Date(prevMsg.attr("data-time"));
@@ -1050,7 +1054,7 @@ $(function() {
 		}
 
 		var chanChat = chan.find(".chat");
-		if (chanChat.length > 0) {
+		if (chanChat.length > 0 && chan.data("type") !== "special") {
 			chanChat.sticky();
 		}
 
--- a/src/plugins/irc-events/list.js
+++ b/src/plugins/irc-events/list.js
@@ -5,7 +5,7 @@ var Msg = require("../../models/msg");
 
 module.exports = function(irc, network) {
 	var client = this;
-	var MAX_CHANS = 1000;
+	var MAX_CHANS = 500;
 
 	irc.on("channel list start", function() {
 		network.chanCache = [];
@@ -23,7 +23,9 @@ module.exports = function(irc, network) {
 	irc.on("channel list end", function() {
 		updateListStatus(new Msg({
 			type: "channel_list",
-			channels: network.chanCache.slice(0, MAX_CHANS)
+			channels: network.chanCache.sort(function(a, b) {
+				return b.num_users - a.num_users;
+			}).slice(0, MAX_CHANS)
 		}));
 
 		if (network.chanCache.length > MAX_CHANS) {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1022,6 +1022,10 @@ kbd {
 	border-bottom: #eee 1px solid;
 }
 
+#chat table.channel-list .channel {
+	width: 80px;
+}
+
 #chat table.channel-list .channel,
 #chat table.channel-list .topic {
 	text-align: left;
@@ -1029,6 +1033,7 @@ kbd {
 
 #chat table.channel-list .users {
 	text-align: center;
+	width: 50px;
 }
 
 #chat table.channel-list td.channel .inline-channel {


src/plugins/irc-events/list.js
```diff
@@ -23,7 +23,9 @@ module.exports = function(irc, network) {
 	irc.on("channel list end", function() {
 		updateListStatus(new Msg({
 			type: "channel_list",
-			channels: network.chanCache.slice(0, MAX_CHANS)
+			channels: network.chanCache.sort(function(a, b) {
+				return a.num_users === b.num_users ? 0 : a.num_users > b.num_users ? -1 : 1;
+			}).slice(0, MAX_CHANS)
```
astorije: [Per MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description), the following would work too:

```js
network.chanCache.sort((a, b) => b.num_users - a.num_users).slice(0, MAX_CHANS)
```

Could you give this a shot please?

src/plugins/irc-events/list.js
```diff
@@ -5,7 +5,7 @@ var Msg = require("../../models/msg");
 
 module.exports = function(irc, network) {
 	var client = this;
-	var MAX_CHANS = 1000;
+	var MAX_CHANS = 500;
```
astorije: A couple ideas for improvement of this, but **not necessarily for this PR**:

- *Easy*: Load 501 chans and in the front-end add a note saying "Only the 500 most crowded channels were displayed" if there is more than 500 chans. It's a cheap trick to avoid having an extra property `hasMoreThanFiveHundredChannels` 😄

- *Hard*: Load the first 100 only and add a "Show more channels" at the bottom or make it infinite scrolling that loads N new channels as you scroll. There is a mention of this pattern somewhere (I forget which issue/PR) for something like this: when opening a channel with lots of unread messages, ideally we would display the first unread ones and load new ones as we scroll down. These could use the same mechanism, but it's obviously much harder than the cheap solution above.

client/js/lounge.js
```diff
@@ -448,6 +452,12 @@ $(function() {
 				.find(".unread-marker")
 				.appendTo(container);
 		}
+		if (data.msg.type === "channel_list") {
+			// Delay by 1 second so the browser has time to render all channels
+			$(target + " .chat").stop().delay(1000).animate({
+				scrollTop: 0
+			});
+		}
```
astorije: I'm a bit concerned about this hardcoded 1 second. What does it accomplish?
I tried `/list` on Freenode just now (on v2.2.2, not your PR) and it never loads, tab actually hangs for a long time, and I'm running a fairly powerful laptop (though the server itself is on a RPi).

So yeah, not convinced about this, but maybe I'm missing something.

astorije:
Another nice and easy improvement once `/list` is fast enough to be usable even on Freenode would be to add an entry to the server context menu.

src/plugins/irc-events/list.js
swordbeta: Nice one!

client/js/lounge.js
xPaw: What does this scroll even accomplish?

swordbeta: @astorije I'm running the server on my VPS and loading takes about ~10 seconds (on v2.2.2), so that might be your RPi.

The delay is needed client side because the browsers needs a while (about 150-400ms here) to render the table with 500 rows. The 1 second is arbitrary.

@xPaw The table is sorted descending meaning channels with the most users (and most interesting channels) can be found on top. Having to scroll through them manually isn't very user friendly.

xPaw: What probably needs to be done is to disable "stick to bottom" for special channel types like this one.

swordbeta: @xPaw Would be cleaner, I don't know how to though.

astorije: @swordbeta, I believe this would be as simple as callin `unsticky()` like [here](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L1021), but that needs testing as I haven't played with this myself yet.

astorije:
@xPaw, thoughts?

xPaw:
Unsticky needs to be implemented, then it can be merged.

astorije:
@swordbeta, are you able to continue this or do you need any help from us?

swordbeta:
@astorije Updated, using the sticky plugin now!

astorije:
Unrelated to this PR, but maybe you'd have an idea: on the channel list, the scrollbar goes behind the message input. Do we want to hide the input in special types, is that even possible? If not, it will have to wait until we move the input within the channels.

swordbeta:
@astorije Squashed and fixed quotes

astorije:
Got @xPaw on IRC, merging.

PR #1024, opened 2017-04-10
Add SSL CA bundle option

metsjeesus:
Letsencrypt serts give me errors on firefox. `ERR_CERT_AUTHORITY_INVALID` , to fix it, it needs a CA bundle file. Currently, lounge does not accept this, so i made one.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -44,6 +44,7 @@ module.exports = function() {
 	} else {
 		const keyPath = Helper.expandHome(config.https.key);
 		const certPath = Helper.expandHome(config.https.certificate);
+		const caPath = Helper.expandHome(config.https.ca);
 
 		if (!keyPath.length || !fs.existsSync(keyPath)) {
 			log.error("Path to SSL key is invalid. Stopping server...");
@@ -55,10 +56,16 @@ module.exports = function() {
 			process.exit();
 		}
 
+		if (caPath.length && !fs.existsSync(caPath)) {
+			log.error("Path to SSL ca bundle is invalid. Stopping server...");
+			process.exit();
+		}
+
 		server = require("spdy");
 		server = server.createServer({
 			key: fs.readFileSync(keyPath),
-			cert: fs.readFileSync(certPath)
+			cert: fs.readFileSync(certPath),
+			ca: caPath ? fs.readFileSync(caPath) : undefined
 		}, app);
 	}
 
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -287,7 +287,16 @@ module.exports = {
 		// @example  "sslcert/key-cert.pem"
 		// @default  ""
 		//
-		certificate: ""
+		certificate: "",
+
+		//
+		// Path to the CA bundle.
+		//
+		// @type     string
+		// @example  "sslcert/bundle.pem"
+		// @default  ""
+		//
+		ca: ""
 	},
 
 	//


src/server.js
```diff
@@ -53,10 +54,22 @@ module.exports = function() {
 			log.error("Path to SSL certificate is invalid. Stopping server...");
 			process.exit();
 		}
-		server = server.createServer({
-			key: fs.readFileSync(keyPath),
-			cert: fs.readFileSync(certPath)
-		}, app);
+		if (config.https.ca.length && fs.existsSync(caPath)) {
```
YaManicKill: This logic doesn't seem right.

I'm assuming you mean `(config.https.ca.length && !fs.existsSync(caPath))`

Also, you'd probably want to do "caPath" rather than "config.https.ca", yeah?

src/server.js
```diff
@@ -53,10 +54,22 @@ module.exports = function() {
...
+			log.error("Path to SSL ca bundle is invalid. Stopping server...");
+			process.exit();
+		}
+		if (config.https.ca.length) {
```
YaManicKill: Would it be dreadful to restructure this to do:

```
server = server.createServer({
  key: fs.readFileSync(keyPath),
  cert: fs.readFileSync(certPath),
  ca:  caPath ? fs.readFileSync(caPath) : undefined
}, app);
```

That should? work exactly the same (test it), but it also removes the duplicate code, and looks much nicer, imo.

metsjeesus:
Changed as requested, can test it on nightime.

metsjeesus:
For my case, its working.

src/server.js
```diff
@@ -53,9 +54,14 @@ module.exports = function() {
 			log.error("Path to SSL certificate is invalid. Stopping server...");
 			process.exit();
 		}
+		if (config.https.ca.length && !fs.existsSync(caPath)) {
```
YaManicKill: Change `config.https.ca.length` to `caPath.length`.

metsjeesus: Why? Idea is to check if configuration parameter is wrongly set. That means parameter exists and file is missing.

YaManicKill: caPath is just config.https.ca with normalised "~" is it not? So what's the need to use config.https.ca? That's extra characters. If config.https.ca is "", then caPath will be "".

metsjeesus: Personally, its easier to understand what is `config.https.ca` then `Helper.expandHome(config.https.ca)`. Speciel when same style is applied on `config.https.key` and `config.https.certificate`

YaManicKill: Those would be changed as well (not as part of your pr). You created the variable, so it should be used here. It's confusing not using it because it leads one to wonder why you aren't using it, and the only reason is because that's what we do with the others.

metsjeesus:
Changed, I could change other ssl file paths too, so they use same style.

src/server.js
```diff
@@ -53,9 +54,14 @@ module.exports = function() {
 			log.error("Path to SSL certificate is invalid. Stopping server...");
 			process.exit();
 		}
+		if (caPath.length && !fs.existsSync(caPath)) {
+			log.error("Path to SSL ca bundle is invalid. Stopping server...");
```
xPaw: CA bundle is not required to run TLS, don't kill the server if there isn't one provided. This whole `if` isn't even required.

src/server.js
```diff
@@ -53,9 +54,14 @@ module.exports = function() {
...
+			process.exit();
+		}
 		server = server.createServer({
 			key: fs.readFileSync(keyPath),
-			cert: fs.readFileSync(certPath)
+			cert: fs.readFileSync(certPath),
+			ca: caPath ? fs.readFileSync(caPath) : undefined
```
xPaw: `caPath.length && fs.existsSync(caPath) ? fs.readFileSync(caPath) : undefined`

src/server.js
YaManicKill: I think the point of this is just that if you specify a ca, but it doesn't exist...that's an error.

If you don't specify a ca, then it'll go through fine. But we don't want to pass through nonsense.

xPaw: Oh whoops, yeah, that makes sense.

xPaw:
Squash the commits and then 👍

astorije:
Also @metsjeesus, mind rebasing with `master` to fix the conflicts on `src/server.js`?

astorije:
Thanks!

astorije:
Hey @metsjeesus, we have sticker packs for our contributors now!

PR #1028, opened 2017-04-13
Use local variables to check length

xPaw:
As @YaManicKill pointed out. A minor fix.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -42,17 +42,20 @@ module.exports = function() {
 		server = require("http");
 		server = server.createServer(app);
 	} else {
-		server = require("spdy");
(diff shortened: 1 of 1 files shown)


src/server.js
```diff
@@ -42,17 +42,20 @@ module.exports = function() {
 		server = require("http");
 		server = server.createServer(app);
 	} else {
-		server = require("spdy");
```
YaManicKill: Is this not needed anymore?

xPaw: I moved it down.

YaManicKill: Oh, so you did. I missed that. The green on github is sometimes really easy to miss.

PR #1032, opened 2017-04-14
Stick to bottom when opening user list

### LINKED ISSUE #1031: Opening/closing userlist scrolls channel
at least, on latest safari and chrome on OS X, opening the userlist scrolls the chat up ~3 messages, and closing scrolls back down.

Open
open

closed
closed

xPaw:
Fixes #1031

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -729,6 +729,7 @@ $(function() {
 		var self = $(this);
 		viewport.toggleClass(self.attr("class"));
 		e.stopPropagation();
+		chat.find(".chan.active .chat").trigger("msg.sticky");
 	});
 
 	function positionContextMenu(that, e) {
@@ -822,7 +823,7 @@ $(function() {
 				+ Math.round(parseFloat(style.borderBottomWidth) || 0)
 			) + "px";
 
-			$("#chat .chan.active .chat").trigger("msg.sticky"); // fix growing
+			chat.find(".chan.active .chat").trigger("msg.sticky"); // fix growing
 		})
 		.tab(complete, {hint: false});
 


client/js/lounge.js
```diff
@@ -729,6 +729,7 @@ $(function() {
 		var self = $(this);
 		viewport.toggleClass(self.attr("class"));
 		e.stopPropagation();
+		chat.find(".chan.active .chat").trigger("msg.sticky");
```
astorije: I was expecting that if I was not stick'ed to bottom and clicked on the user list button this would force-stick me but it doesn't.
It works as expected, i.e. sticks you to bottom only if you were before clicking, but I can't figure out why. Do you happen to know what magic is this, @xPaw? If I read this correctly, any click should stick you to bottom regardless of scroll position before clicking.

xPaw: That's the expected behaviour, that event only keeps at the bottom.

astorije: Hmm yes, but that's not what I meant. Probably explained myself poorly.

Let's put it that way: go to a channel, scroll all the way up, then click on the user list button. According to this change, I'm reading that this should force the window to go stickied to bottom, even though you probably don't want that if you weren't at the bottom before clicking. But this behavior doesn't happen, user stays at the top. This is great, you don't want to get moved to bottom if you weren't before, but I don't understand how that's possible.

xPaw: https://github.com/thelounge/lounge/blob/ecb60e3c787511d175f4d2f29b41fbadd7ba08ab/client/js/libs/jquery/stickyscroll.js#L13-L17 and line 40 for the event.

PR #1033, opened 2017-04-14
Preload scripts as soon as possible

xPaw:
Starts downloading them as soon as possible, instead of waiting for the whole HTML file to load.


astorije:
I see more and more pages/snippets that, instead of having the JS declarations just before `</body>` would put their `<script src=...>` back in the `<head>` block and mark the `script` tag with `async`. Would that achieve the same goal?

xPaw:
Having multiple scripts with async may have issues due to execution order.

astorije:
Fair enough.

astorije:
Merging, minor.

PR #1034, opened 2017-04-14
Add rel noopener to URLs in index.html

xPaw:
Ref: https://developers.google.com/web/tools/lighthouse/audits/noopener

Also replaced mirc colors url to @DanielOaks's documentation.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -49,7 +49,7 @@
 								<h1 class="title">The Lounge is loading…</h1>
 							</div>
 							<div class="col-xs-12">
-								<p id="loading-page-message">Loading the app… <a href="http://enable-javascript.com/" target="_blank">Make sure to have JavaScript enabled.</a></p>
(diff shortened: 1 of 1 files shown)


astorije:
Minor, merging.

client/index.html
```diff
@@ -478,7 +478,7 @@ <h1 class="title">Help</h1>
 								</p>
 								<p>
 									A color reference can be found
-									<a href="http://www.mirc.com/colors.html" target="_blank">here</a>.
+									<a href="http://modern.ircdocs.horse/formatting.html#colors" target="_blank" rel="noopener">here</a>.
```
astorije: @DanielOaks, would you be able to serve these with HTTPS? You have free and easy options now (Let's Encrypt), and it would be a good thing for us to point our users to HTTPS content only. Thanks in advance! 😃

DanielOaks: Ah, didn't see the ping here, need to reduce the amount of stuff I'm following :)

Definitely, no problem! I'll yell in that other issue once it's sorted.

Issue #207, opened 2016-03-20
Automatically load more history when scrolling up


tomnx:
When I am reading channel logs backwards it forces me to keep clicking "Show More" (and after a while clicking again and again becomes annoying). Please enable auto-scrolling. Thanks.

YaManicKill:
Just to clarify, when you say auto-scrolling, you mean when you scroll to the top of the existing scrollback, thelounge should automatically load x number more mesages and put them above where you are so you can keep scrolling back rather than having to press the "Show More" button?

tomnx:
@YaManicKill Yes.

tomnx:
https://github.com/thelounge/lounge/blob/eadcca0b5e54bb653072edab5fbe0730698076db/client/js/lounge.js#L825

	chat.on("click", ".show-more-button", function() {
		var self = $(this);
		var count = self.parent().next(".messages").children().length;
		socket.emit("more", {
			target: self.data("id"),
			count: count
		});
	});

astorije:
@davibe, do you want to take a look at this now that you've nailed https://github.com/thelounge/lounge/pull/496? :-)
Maybe @tomnx's suggestion is enough (or maybe not otherwise there would be a PR :D), not sure.

davibe:
I am lazy, but i could try

PR #1038, opened 2017-04-15
Automatically scroll up when reaching the top of the chat window

astorije:
Horrible attempt at fixing #495 and #207.

This is horrible and really buggy.
It feels really slow at the moment, which gets me to think it's attempting to load multiple times.
Also, this should not have to live in the `open` event handler but I couldn't make it work outside of it yet.
Also the top height should be determined in % and not in pixels.
Also when functional the physical button should not be necessary anymore.
Also when removing the button the "clear chat" shortcut will not work anymore (but one thing at a time).

I thought it could be a 10 minute hack but looks like I was too optimistic 😅.
Help welcome.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -477,6 +477,8 @@ $(function() {
 
 		if (data.messages.length !== 100) {
 			scrollable.find(".show-more").removeClass("show");
+		} else {
+			scrollable.find(".show-more").addClass("show");
 		}
 
 		// Date change detect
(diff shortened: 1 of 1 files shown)


client/js/lounge.js
```diff
@@ -1060,6 +1062,19 @@ $(function() {
...
 		focus();
+
+		let oldScrollTop = 0;
+		$("#chat .active .chat").on("scroll", function() {
+			const scrollTop = $(this).scrollTop();
+
+			// Make sure we are scrolling up and nearing the top
+			// FIXME Use a window percentage instead of a height in px
+			if (oldScrollTop > scrollTop && scrollTop < 200) {
+				$(this).find(".show-more.show .show-more-button").click();
```
xPaw: Hiding the button does not prevent `click` callback from triggering again.

astorije:
Closing this as #1045 has a simple fix for the underlying issue.
If someone wants to pick up where I left off (it would still be really cool to address #207), please be my guest but right now I have way too much on my plate to commit to this.

bews:
> If someone wants to pick up where I left off

Not sure what do you want to do with this. Right now it works fine without any major changes.
This is my test code (and the only change needed to achieve autoload):



So:
> It feels really slow at the moment, which gets me to think it's attempting to load multiple times.

Fixed already.

> Also, this should not have to live in the open event handler but I couldn't make it work outside of it yet.

It's not possible to autobind scroll event to ajax-loaded elements (or at least I couldn't find the working way).

> Also the top height should be determined in % and not in pixels.

Um.. Why?

> Also when functional the physical button should not be necessary anymore.

True, but other modern messengers still have it. Don't know why, maybe there are reasons to keep it?

> Also when removing the button the "clear chat" shortcut will not work anymore

+1 reason to keep it :)

PR #1041, opened 2017-04-16
Process received messages in queue

xPaw:
Creating early PR to gather feedback. Want to hear what  @thelounge/maintainers think.

Connecting to ZNC with history playback no longer freezes my browser, and appends messages in small queues instead of as-soon-as-received.

What I want to achieve is to process multiple messages for one channel at once, and append them at once (document fragments). Thus reducing DOM manipulations and style recalculations. This should also prioritise currently active channel.

- [x] Sort messages per channel
- [ ] Prioritize currently active channel
  - [ ] How does this handle for PMs/whois which force a new window to appear?
- [x] Use document fragments to append multiple messages at once
- [x] Trigger events (notifications, sticky scroll) once per processed channel
  - [ ] Look more into notifications
- [x] #96 Remove 10 second interval timer to hide messages in hidden channels
- [x] Limit notification sound to play every 500 milliseconds
- [ ] Fix condensed stuff

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -14,7 +14,6 @@ const emojiMap = require("./libs/simplemap.json");
 require("./libs/jquery/inputhistory");
 require("./libs/jquery/stickyscroll");
 require("./libs/jquery/tabcomplete");
-const helpers_roundBadgeNumber = require("./libs/handlebars/roundBadgeNumber");
 const slideoutMenu = require("./libs/slideout");
 const templates = require("../views");
 const socket = require("./socket");
@@ -28,20 +27,6 @@ $(function() {
 	var chat = $("#chat");
 
 	$(document.body).data("app-name", document.title);
-
-	var pop;
-	try {
-		pop = new Audio();
-		pop.src = "audio/pop.ogg";
-	} catch (e) {
-		pop = {
-			play: $.noop
-		};
-	}
-
-	$("#play").on("click", function() {
-		pop.play();
-	});
 
 	// Autocompletion Strategies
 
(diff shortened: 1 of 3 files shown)


astorije:
@xPaw, adding the `do not merge` label for now, feel free to assign folks and remove the label when you're done with it.

xPaw:
Did another optimization pass, and removed that ugly 10 second interval timer to hide messages in non visible channels (#96).

astorije:
Rebased with `master`. Now back to boarding a plane 😄

astorije:
Alright, so I started looking at the code (not tested yet) and I am so confused 😅.
It's some very convoluted/complex code. I'm not criticizing, I know it's for good reasons, but it makes reviewing rather difficult and pretty sure within 2 months after merging we'll be wondering why this or that change :)
Any way you could comment this code, @xPaw? I know it takes some efforts to do so, but I don't think I can fully understand the changes right now.

xPaw:
The main idea of this PR is to append messages to DOM on browser idle callback (giving it time to breathe), and while it might be waiting for the idle event to callback, multiple messages can be received, and if they are, as such, they are processed in a single go. This gives a very nice boost for multiple messages in a single channel, as they get appended to actual DOM with single operation.

kode54:
Does not work on iOS Safari:

```ReferenceError: Can't find variable: requestIdleCallback```

xPaw:
@kode54 Fixed. However, since Safari does not have `requestIdleCallback`, it will not benefit from this PR.

astorije:
@xPaw, I see some empty checkbox in here, do you still want to work on this? I wouldn't like to start reviewing 300+ lines of changes if you happen to still edit it :)

xPaw:
@astorije these checkbox might not get even done (the code is fine as is).

xPaw:
I have rebased this somewhat, but this certainly breaks condensed messages due to the use of document fragments.

astorije:
@xPaw, what do you think for this PR? Do you need something else done with condensed message so this PR is compatible? I understand you want your other PRs merged before taking up any more work, which is totally fair, do you want to close this PR until condensed stuff makes it possible again?
I really want this and your other work merged too, not sure how you want to proceed.

xPaw:
@astorije No reason to close this.

PR #1044, opened 2017-04-17
Add fix for undefined name being slugified

YaManicKill:


xPaw:
I would rather figure out why its undefined.

YaManicKill:
So, it comes from this slugify:

```
{{#each networks}}
<section id="network-{{id}}" class="network name-{{slugify name}} network-{{slugify serverOptions.NETWORK}}" data-id="{{id}}" data-nick="{{nick}}" data-options="{{tojson serverOptions}}">
	{{> chan}}
</section>
{{/each}}
```

Because the serverOptions.NETWORK is undefined, which is something we get from the server itself, connection.js:134:

```
network.serverOptions.NETWORK = data.options.NETWORK;
```

So, we could make some assumptions if we want, but it sounds like the server is just being stupid. This is snoonet server, btw.

YaManicKill:
Actually, I guess we could use the name rather than serverOptions.NETWORK as that's user-defined, and means that if someone connects to the same server twice, the 2 slugs can have different names if the user wants them to, whereas this wouldn't be possible to do that. Also, means that we know it won't be undefined as it's required.

astorije:
@xPaw, would it be reasonable to merge this until finding a proper fix? I'm in @YaManicKill's case right now: crash on startup, instance unusable without this.

YaManicKill:
Based on what we know, I think this is an acceptable temporary fix (obviously I wrote it) as I feel we could easily bikeshed about the best way to solve it otherwise, but we can open another one to decide the best way to deal with it. But as of just now, master is broken for many people. 

@astorije What server is yours failing on? I bet this isn't a massively common thing as 3/4 of my servers send the string back, it's just the 1 that doesn't, and that's snoonet.

xPaw:
What about adding `|| ""` network name? Or before calling slug in template, check if there is a network name. `name-` is kind of useless as is anyway. I really hate masking bugs like this PR currently does.

YaManicKill:
> What about adding || "" network name?

What do you mean?

> Or before calling slug in template, check if there is a network name.

That could work.

> I really hate masking bugs like this PR currently does.

I don't think we are masking a bug, I believe this **is** a bug. I think we should also do something else, but this caused crashes because we didn't check that we have something that isn't undefined, that's the problem.

There are other problems as well, but this is still an issue.

xPaw:
Just add an `if` around the slugify call then.

YaManicKill:
> Just add an if around the slugify call then.

No, but if we have other places we use this, they will crash if we don't have this fix anyway.

xPaw:
Other places slugify unknown commands and channel names, if these were to be unknown, i would rather get a bug report.

YaManicKill:
> Other places slugify unknown commands and channel names, if these were to be unknown, i would rather get a bug report.

But this is not a good way to deal with this because it completely shuts down whatever we were trying to do. If we want to know about these things we shouldn't just let it all fail with an error on trying to call something from undefined.

xPaw:
These shouldn't fail.

YaManicKill:
Aight, have your "proper" fix. I hope you are happy.

YaManicKill:
Merging as it's a small fix. @astorije will be happy to wake up to this.

PR #1047, opened 2017-04-17
Condense join into nicknames

metsjeesus:


metsjeesus:
I did not like the way the condensing is made. Alot of information is getting lost, so i rewrite it a bit to store all condensed message info, so its usable later on to produce condensed text. @YaManicKill , can you fix the conflict, i dont know the correct way. 

Speaking of formating, i feel it should be done with templates. I just ask before i do something stupid. What is the prefered way to do them? There are lot of combinations how to shorten all the types and messages. Would be nice to get some examples, maybe even a example test. It should be now testable.

YaManicKill:
I think I've fixed the conflicts, but not sure. I think I want to rewrite the whole thing, cause it has gotten into quite a horrible mess. I'll wait until #1049 is merged, though.

> Speaking of formating, i feel it should be done with templates

What do you mean? What  formatting?

metsjeesus:
Condensed text, all those "name1 and name2 joined, name 3 quit" etc. It feels its not right to just hack all the types in to a text string. It should be something similar to a normal message with html tags and css.  Backside of it, testing will be harder.

After my last commit, its no so horrible, easily movable to modules. Still, Its not tested enough to put to `master`, but i hope to update `thelounge:yamanickill/condense-joins`.

YaManicKill:
> It feels its not right to just hack all the types in to a text string. It should be something similar to a normal message with html tags and css. 

Ah yeah, good point actually. Try it and see how it works.

client/views/msg_action.tpl
```diff
@@ -1,6 +1,6 @@
 <div class="msg {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}}"
 	data-type="{{type}}" id="msg-{{id}}" data-time="{{time}}">
-	<span class="time tooltipped tooltipped-e" aria-label="{{localetime time}}">
+	<span class="time" tooltipped tooltipped-e" title="{{localetime time}}">
```
metsjeesus: This seems kind of wrong?

YaManicKill: That's because it is. Groan, I'll take a look later. I will be focusing on the modularising and my other PRs over the condensing one just now. I will pick it up after that.

metsjeesus:
@YaManicKill , ive changed bit methods, modularized it, changed some functionality and made a test cases so we could test the functionality better. Would be nice to make some more tests how the condensing should work. Keep in mind, that it all depends on order of the input messages, Its still work in progress on mode change and maybe even some specific edge cases. This system should hopefully manage to condense correctly anything you can figure out.

bews:
Wah, so much code change. Gonna test this too.

client/js/lounge.js
```diff
@@ -296,70 +297,59 @@ $(function() {
 		return msg;
 	}
 
-	function updateCondensedText(condensed, addedTypes) {
-		var obj = {};
+	function appendMessage(input) {
+		var container = input.container || undefined;
+		var chan = input.chan || undefined;
```
MaxLeiter: Opportunity to use let > var?

bews:
So, here what I got:
- This PR conflicts with #1141 
- Kick message is not condensed (is this intended?)
- Nicknames are not colored

 In #1141 time column doesn't have a default width anymore, and this PR leaves time column empty:



To fix this time should be posted, but hidden (like in embed).

-----------------

I'll try to fix this and nicknames coloring and post changes here later.

bews:
My fixes:

> This PR conflicts with #1141

https://github.com/bews/lounge/commit/32f05122345d8d65793bf8469b3d66c2b8ddbfcb

> Nicknames are not colored

https://github.com/bews/lounge/commit/8cdfa448fc77f755bfaee414d1fb123fbfeb9c43 - requires #1096



> Repeating nicknames bug

https://github.com/bews/lounge/commit/d06a93f2952409ffd48c2660ad10aa0e4db1db44

> Text layout is broken when condense message is too long (flex-wrap: wrap; is the reason)

https://github.com/bews/lounge/commit/8f28d1387adb1610cb77c38cd05b7e76766d8c68

> Date marker is broken:

https://github.com/bews/lounge/commit/cb9e37fde724afe72c8a93594312232988d1c922

------------------------

Only `Show mode` problem left.

bews:
Yep, when we fix `tests` this PR will be ready to go. 

I also made support for #1166 (which is fantastic with this PR):



I'll post it later, when either of these PRs will be merged (to prevent more dependencies).

metsjeesus:
I try on evenings add the modifications to this pull.  I did not implement mode change, because there are so many ways of them and its annyoing to see many of them.  I propose a condensing, that if user joins and gets voiced/halfopped/opped and closed, it will show as a separate voice/halfop/op group.  Example `user1 opped joined `or `user1 opped reconnected`. Would it be useful?

metsjeesus:
Other way would be that the mode mark is in user name, like `@user1 joined` and `@user1 opped`. Would it be better?

bews:
IRCcloud adds mode mark too, also I like it having different color:

xPaw:
This PR wants to merge to `yamanickill/condense-joins` and not `master`, and since that PR has been merged into master now, it can not be merged anymore.

PR #1049, opened 2017-04-17
Create modules for lots of code

YaManicKill:
Because seriously, it was getting ridiculous, and we have a module system now.

### DIFF
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -1,0 +1,36 @@
+"use strict";
+
+const commands = [
+	"/away",
+	"/back",
+	"/close",
+	"/connect",
+	"/deop",
+	"/devoice",
+	"/disconnect",
+	"/invite",
+	"/join",
+	"/kick",
+	"/leave",
+	"/me",
+	"/mode",
+	"/msg",
+	"/nick",
+	"/notice",
+	"/op",
+	"/part",
+	"/query",
+	"/quit",
+	"/raw",
+	"/say",
+	"/send",
+	"/server",
+	"/slap",
+	"/topic",
+	"/voice",
+	"/whois"
+];
+
+module.exports = {
+	commands: commands
+};
--- a/client/js/contextMenu.js
+++ b/client/js/contextMenu.js
@@ -1,0 +1,92 @@
+"use strict";
+
+const templates = require("../views");
+const $ = require("jquery");
+const contextMenuContainer = $("#context-menu-container");
+const contextMenu = $("#context-menu");
+const viewport = $("#viewport");
+
+function positionContextMenu(that, e) {
+	var offset;
+	var menuWidth = contextMenu.outerWidth();
+	var menuHeight = contextMenu.outerHeight();
+
+	if (that.hasClass("menu")) {
+		offset = that.offset();
+		offset.left -= menuWidth - that.outerWidth();
+		offset.top += that.outerHeight();
+		return offset;
+	}
+
+	offset = {left: e.pageX, top: e.pageY};
+
+	if ((window.innerWidth - offset.left) < menuWidth) {
+		offset.left = window.innerWidth - menuWidth;
+	}
+
+	if ((window.innerHeight - offset.top) < menuHeight) {
+		offset.top = window.innerHeight - menuHeight;
+	}
+
+	return offset;
+}
+
+function showContextMenu(that, e) {
+	var target = $(e.currentTarget);
+	var output = "";
+
+	if (target.hasClass("user")) {
+		output = templates.contextmenu_item({
+			class: "user",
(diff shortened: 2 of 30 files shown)


astorije:
@YaManicKill, is there any way you can split this in multiple PRs? 2.2k lines of change at once is a really rough PR to review and test 😅.

YaManicKill:
@astorije Yeah, I get that. The problem with multiple PRs is that it holds up the massive change for longer, and it'll end up being a lot of PRs (how small do I go?). How about I split it into multiple commits but 1 PR, then you can look at them individually but it's still only 1 PR, reducing the amount of time creating, rebasing and merging, and decreasing the amount of time before it's merged in. This is necessary, but it's going to cause merge conflicts for basically **every** PR.

xPaw:
Yeah, it's better to get such a refactor in one go.

YaManicKill:
Alright, I've split this into multiple commits. I hope it's small enough to understand. It should be easy enough to check if it works if many people run it and try all the features. If all the features work, and there are no errors in the console, then it has worked. I understand this is a hard PR to review, but it'll make every other PR in the future better, will decrease issues with merge conflicts, and will hopefully encourage more contributions. The file was getting utterly ridiculous, and we were never going to manage to get it done slowly. Big bang is the only real approach for this.

astorije:
I'm unclear, if we can split this in multiple commits, why would we want to keep this in one go?

The advantage I'm seeing when splitting this (which btw good job on doing that!) is that we can probably merge some of these right now without affecting other open PRs or delaying the release.

If we keep this in one PR, it will definitely have to go post-v2.3.0, and if I feel bad asking contributors to rebase their code on something that completely changed and where they have to struggle.

Don't get me wrong, I think it's an amazing job you did right there, and I do want to ship this, I'm just trying to think of how we can make this as little painful as possible for everyone (and that includes you, I'm not saying all the dirty work has to fall down on you :) ).

astorije:
The really cool thing about you doing everything at once is that this gives a perfect glance at the final structure of everything, instead of what I would have done, identify and work things piece by piece, and then realize that damn, that pattern doesn't work too well, and too late to back out now.

@xPaw, what do you think of the structure @YaManicKill set up? At first glance it seems very reasonable to me.

Looking at the commits and open PRs, it looks like the first one, `socket.js`, is not affected by any open PR. What about making a PR for this one so we can merge it on its own? I think that would be the fastest way to get this in `master` **and** not cause pain and frustration to authors of other PRs.

YaManicKill:
> I'm unclear, if we can split this in multiple commits, why would we want to keep this in one PR?

So, my reasoning for keeping it in 1 PR is as follows:

1. As you mention, it allows us to discuss the structure of how it goes rather than doing 1 thing at a time and everyone assuming it;s going to be done a different way
2. The admin over creating 10 PRs (it could easily be more commits as I bundled a bunch of stuff together) is quite a lot, both for me and other people. It involves creating a PR, getting it reviewed, merging it, rebasing all my other ones, and starting again. And for everyone else, it could mean having conflicts multiple times if they are dealing with code in multiple places that are spread across 2 or more of my PRs. This causes them angst as well.
3. There is never going to be a "right time" to do this. That would imply that for all the code in this PR there will be a time where there isn't a PR covering that code at some point. This just simply isn't possible. For some (as you mention, the socket module is unlikely to cause merge conflicts) then it'll be fine, but for others (the render file) it probably won't ever happen. We don't want it to either because that means there's stagnation in the project.
4. It keeps all the commits together, making it clear that they are all very clearly related and making it much easier to see how things happened in the git history

I absolutely understand what you are saying and I know this is a hard PR to sort out, but I think doing it multiple times causes more problems for everyone. Splitting into multiple commits makes it easier to review, but obviously no easier to merge. Multiple PRs makes it more difficult in the ways I have specified above.

Having said that, if you are strongly against the 1 PR, then I could do one for just the socket part of it and we can see where we go from there.

xPaw:
I agree with @yamanickill, it's better to do it in a single go, otherwise we will keep blocking other contributions.

MaxLeiter:
I'm with @astorije on this; lots of conflicts will be made and lots of PRs will be set back. I think, due to the fact it was relatively easily split into multiple PRs, we could merge one at a time to limit conflicts and regressions. That being said, I think this merging should be a pretty big priority, and maybe this is just how it has to be done

YaManicKill:
Alright, I'll start by doing some small ones, just because fine. But I think a lot of these will have to go together. We shall see.

xPaw:
I'll close this but keep the branch. Feel free to create more PRs for more splitting.

YaManicKill:
Yeah, I'll add them over time. We'll see how quickly we can get these in :-P

PR #1050, opened 2017-04-17
Fix to helper.expandhome to correctly resolve "" and undefined

metsjeesus:
My CA bundle pull broke some things when SSL is set and CA not.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -101,6 +101,9 @@ function ip2hex(address) {
 }
 
 function expandHome(shortenedPath) {
+	if (!shortenedPath) {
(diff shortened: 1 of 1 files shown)


src/helper.js
```diff
@@ -101,6 +101,9 @@ function ip2hex(address) {
 }
 
 function expandHome(shortenedPath) {
+	if (shortenedPath === "" || shortenedPath === undefined) {
```
YaManicKill: I would probably change this to just "if(!shortenedPath)" but I don't massively mind.

PR #1051, opened 2017-04-17
Use HTTPS version to the IRC color guide

astorije:
@DanielOaks very nicely and blazingly quickly set up HTTPS (see https://github.com/ircdocs/modern-irc/issues/38) so we can now direct our users to that version!

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -404,7 +404,7 @@
 								</p>
 								<p>
 									A color reference can be found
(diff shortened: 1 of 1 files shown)


xPaw:
Due to your osx bindings duplication, there are 2 of these URLs.

astorije:
Well, that's embarrassing... :man_facepalming: 
Fixed, @xPaw.

PR #1054, opened 2017-04-19
Use moment on the client to display friendly dates

astorije:
Some time ago, @MaxLeiter [said](https://github.com/thelounge/lounge/pull/765#issuecomment-263306173):

> Am just going to mention this here: once the client can access npm packages, we can use moments .calender() to create fuzzy time stamps (i.e. "Yesterday")

There was some discussion [in here](https://github.com/thelounge/lounge/pull/671) as well.

**Result:**

screen shot 2017-04-19 at 02 04 08

Also, unread and date markers are now half-transparent based on their colors and not parent opacity. This is necessary to display a non-half-transparent tooltip.

Vendor bundle is a bit bigger now, but I think it's for the best and we will most likely benefit from having Moment.js available on the client anyway.

⚠️ **Caveat:** At the moment, this does not refresh at midnight, so any date marker already on the page will stay at "Today" when they should switch to "Yesterday". I don't think this is a big deal but happy to get opinions. Either way, if people feel strongly about this, I'll happily accept help (extra commit on that branch?) as I'll be moving to something else tomorrow anyway 😅.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -841,7 +841,6 @@ kbd {
 #chat .unread-marker {
 	position: relative;
 	text-align: center;
-	opacity: .5;
 	margin: 0 10px;
 	z-index: 0;
 	font-weight: bold;
@@ -855,13 +854,13 @@ kbd {
 	left: 0;
 	right: 0;
 	top: 50%;
-	border-top: 1px solid #e74c3c;
+	border-top: 1px solid rgba(231, 76, 60, .5);
 }
 
 #chat .unread-marker-text:before {
 	content: "New messages";
 	background-color: white;
-	color: #e74c3c;
+	color: rgba(231, 76, 60, .5);
 	padding: 0 10px;
 }
 
@@ -872,7 +871,6 @@ kbd {
 #chat .date-marker {
 	position: relative;
 	text-align: center;
-	opacity: .5;
 	margin: 0 10px;
 	z-index: 0;
 	font-weight: bold;
@@ -886,13 +884,13 @@ kbd {
 	left: 0;
 	right: 0;
 	top: 50%;
-	border-top: 1px solid #006b3b;
+	border-top: 1px solid rgba(0, 107, 59, .5);
 }
 
 #chat .date-marker-text:before {
-	content: attr(data-date);
+	content: attr(data-label);
 	background-color: white;
-	color: #006b3b;
+	color: rgba(0, 107, 59, .5);
 	padding: 0 10px;
 }
 
--- a/client/js/libs/handlebars/friendlydate.js
+++ b/client/js/libs/handlebars/friendlydate.js
@@ -1,0 +1,13 @@
+"use strict";
+
+const moment = require("moment");
+
+module.exports = function(time) {
+	// See http://momentjs.com/docs/#/displaying/calendar-time/
+	return moment(new Date(time)).calendar(null, {
+		sameDay: "[Today]",
+		lastDay: "[Yesterday]",
+		lastWeek: "L", // Locale
+		sameElse: "L"
+	});
+};
--- a/test/client/js/libs/handlebars/friendlydateTest.js
+++ b/test/client/js/libs/handlebars/friendlydateTest.js
@@ -1,0 +1,24 @@
+"use strict";
+
+const expect = require("chai").expect;
+const moment = require("moment");
+const friendlydate = require("../../../../../client/js/libs/handlebars/friendlydate");
+
+describe("friendlydate Handlebars helper", () => {
+	it("should render 'Today' as a human-friendly date", () => {
+		const time = new Date().getTime();
+		expect(friendlydate(time)).to.equal("Today");
+	});
+
+	it("should render 'Yesterday' as a human-friendly date", () => {
+		const time = new Date().getTime() - 24 * 3600 * 1000;
+		expect(friendlydate(time)).to.equal("Yesterday");
+	});
+
+	it("should not render any friendly dates prior to the day before", () => {
+		[2, 7, 30, 365, 1000].forEach(day => {
+			const time = new Date().getTime() - 24 * 3600 * 1000 * day;
+			expect(friendlydate(time)).to.equal(moment(time).format("L"));
+		});
+	});
+});
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -3,6 +3,7 @@
 // vendor libraries
 require("jquery-ui/ui/widgets/sortable");
 const $ = require("jquery");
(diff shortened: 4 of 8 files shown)


client/css/style.css
```diff
@@ -855,13 +854,13 @@ kbd {
 	left: 0;
 	right: 0;
 	top: 50%;
-	border-top: 1px solid #e74c3c;
+	border-top: 1px solid rgba(231, 76, 60, .5);
```
YaManicKill: This is just tidying up, yeah? I think we want to put this in a different one.

client/js/libs/handlebars/friendlydate.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+const moment = require("moment");
+
+module.exports = function(time) {
+	// See http://momentjs.com/docs/#/displaying/calendar-time/
+	return moment(new Date(time)).calendar(null, {
+		sameDay: "[Today]",
+		lastDay: "[Yesterday]",
+		lastWeek: "L", // Locale
```
YaManicKill: This is just a general comment, not a comment on you.

I hate where we have to comment on these stupid magic shortcodes to figure out what on earth we are wanting to show. It's ridiculous. Why can't they have a better way of writing these out so it's self-documenting? Gagh!

YaManicKill: Actually, here's an idea to make this better.

Create a constant for this string and name it "Locale", or something, and then you don't need the comment. But still, my point remains.

xPaw:
What about other places that can benefit from moment? Like message time tooltips?

YaManicKill:
> What about other places that can benefit from moment?

Anywhere we currently use dates. I would like us to switch to using moment entirely because there are many weird edge cases that it deals with.

client/css/style.css
MaxLeiter: Adds opacity, Astorije explains why in the commit message:

`This is necessary to display a non-translucide tooltip.`

astorije: What @MaxLeiter said. If I don't do this *on the date marker*, the tooltip is half-transparent.
The fact that I changed the unread marker as well was 1) to be consistent, 2) because there were some shared CSS so I had to either do that, or do a mixed ugly version of it, not worth it. Better to fix completely (it's also nicer for the themes).

client/js/libs/handlebars/friendlydate.js
astorije: I feel you @YaManicKill.
I don't think a constant is a much better solution than a comment though (and I often find it's not a better one at all), I think we should just move on :)
Also, as we'll be adding more and more Moment.js stuff on the front-end, we'll probably have a common config piece for it so that will be more suitable there. But for later IMO (or else, for someone else 😅).

astorije:
Re: midnight switch, fair enough, I'll see if I can try something quickly. I may have an idea for a quick and nice solution. If not, I guess I'll close this PR then.

Re: anywhere else, again, I'll see what I can do, but if that gets out of hand, I'll vote for another PR, or someone to take over.

astorije:
@YaManicKill, I now recompute the templates for all "Today"/"Yesterday" markers (leaving the others alone since they're on a fixed date) when midnight comes. Hope that's good enough now 😊

client/css/style.css
YaManicKill: Oh, I saw it had changed for the 1 that already had opacity and hadn't noticed it was also added to others. Fine, fine.

client/js/libs/handlebars/friendlydate.js
YaManicKill: > I don't think a constant is a much better solution than a comment though

I disagree, I think it's better because apart from anything, we are duplicating the string twice currently, and if we change how we are doing it, we'll have to updatei t in 2 places. I know it's not a big thing, but I'd really like you to do this.

YaManicKill:
Yeah, I'm with @astorije on just keeping this to this current one. I think it's worth replacing everywhere else that we use dates to use moments, but that's a different PR.

astorije:
> Also, boo adding more Javascript not in a module for me to fix in my PR ;-)

I agree, I did feel bad about it, so let's see if I can fix that quickly.
Looking at your PR for consistency, what if I create a module for date markers, which for now just exports a `refreshEveryday()` (or better wording), run in `lounge.js`?
I would only put my new code however, I will not touch existing date marker code :)
Not sure what I'd need to pass the function, like `templates`, etc. to make it clean but I'll be without an internet connection in a couple days so I can't get too fancy without affecting my work on other PRs.

Ideally, we would have a `DateMarker` component which encapsulates such ability, that would be the cleanest, but our current arch doesn't allow anything like that I believe.

Anyway, let me know what you think and I'll see what I can do assuming it's small enough that I can take it (I've already spent way too much time on this thing lol). Thanks 😃

YaManicKill:
I would just leave it as is and I'll include it in my modularising. This is nice and small nad sorted and I haven't done the modularising yet, so don't worry about it.

Once the modules are done though, everything should be in a module, the main file will just be for tying things together.

astorije:
Rebased and fixed conflict.

client/js/libs/handlebars/friendlydate.js
xPaw: I kind of find using locale aware date formatting a bit shit. What if we formatted like `April 22, 2017`?

test/client/js/libs/handlebars/friendlydateTest.js
```diff
@@ -0,0 +1,24 @@
...
+
+	it("should render 'Yesterday' as a human-friendly date", () => {
+		const time = new Date().getTime() - 24 * 3600 * 1000;
+		expect(friendlydate(time)).to.equal("Yesterday");
+	});
+
+	it("should not render any friendly dates prior to the day before", () => {
+		[2, 7, 30, 365, 1000].forEach(day => {
+			const time = new Date().getTime() - 24 * 3600 * 1000 * day;
+			expect(friendlydate(time)).to.equal(moment(time).format("L"));
```
xPaw: Use hardcoded timestamps and compare to hardcoded strings, otherwise you don't know if you're actually getting expected output.

astorije: I thought about that but that depends on locale, which is already causing issues in https://github.com/thelounge/lounge/blob/master/test/client/js/libs/handlebars/localetimeTest.js. I would need to look into mocking the locale, etc. and I won't have time for this on this PR.
I can either let it as is and someone can look into that post-merge, or remove the test, or close the PR. Obviously I'm advocating for the first option :)

Issue #1055, opened 2017-04-19
Split index.html into multiple files


YaManicKill:
Now that we are on the way to getting the client Javascript split up, I think it would be nice to get the index.html split up as well, because that is a *behemoth*.

We are currently apparently using lodash templating for the themes and the git commit, which seems a little weird...but I guess we could keep using that or switch to handlebars which we use for everything else.

xPaw:
We should indeed use handlebars to pre-render it, as it would introduce consistency.

YaManicKill:
Yeah, we need to use handlebars if we want to use an include,because lodash's one doesn't have an include.

xPaw:
Now that we switched to handlebars this is easily doable. Anyone want to do it?

astorije:
@xPaw, would splitting this make it more difficult to localize (cf. #1271), easier, or have no effect at all, you think?

MaxLeiter:
Wouldn't have any effect

JemBijoux:
Hey! I'd love to help out and take this on as part of Haktoberfest. Just need an evening to orient myself a bit to some of the code, so maybe have something over the next day or so...

astorije:
We'd love that, @JemBijoux!
If this is of any help, I had attempted it very briefly in https://github.com/thelounge/lounge/commit/32c580be5618efcdb3c8b461592943f4bb426c75 but didn't move forward.
Good luck with that!

PR #1057, opened 2017-04-20
Change index.html to be rendered using handlebars

YaManicKill:
index.html is huge (as explained in #1055) and we currently use lodash's templating to make the file. The problem is that lodash doesn't have an "include" function, meaning that we can't split the file up into multiple (think, 1 file for the help page, 1 for the settings page, etc). So if we want to split it up, we need to switch to something else. As we use handlebars on the client, it felt like a good situation to use handlebars for the initial index.html as well.

So to be consistent, and to allow us to split do #1055, we need to switch to handlebars.


client/js/libs/handlebars/firstLetterUppercase.js
```diff
@@ -0,0 +1,5 @@
+"use strict";
+
+module.exports = function(string) {
+	return string.charAt(0).toUpperCase() + string.slice(1);
+};
```
astorije: I know this is a stupid-ass function, but do you mind adding a couple test cases for this?
Our handlebars helpers are actually the most tested area of the client so far lol (especially once #972 lakes it into `master`).

client/index.html
```diff
@@ -25,7 +25,7 @@
 	<meta name="theme-color" content="#455164">
 
 	</head>
-	<body class="signed-out <%= public ? "public" : "" %>">
+	<body class="signed-out {{ternary public 'public' '' }}">
```
astorije: OMG Handlebars...
Actually, I found several SO answers that say you can inline `if` like this:

```js
{{ if myBool "True" "False" }}
```

Mind giving it a shot? Equally awkward but we don't have to have a helper just for that then lol...

package.json
```diff
@@ -46,6 +46,7 @@
     "commander": "2.9.0",
     "event-stream": "3.3.4",
     "express": "4.15.2",
+    "express-handlebars": "^3.0.0",
```
astorije: Pinned version instead.

client/js/libs/handlebars/index.js
```diff
@@ -0,0 +1,18 @@
...
+	localetime: require("./localetime"),
+	modes: require("./modes"),
+	parse: require("./parse"),
+	roundBadgeNumber: require("./roundBadgeNumber"),
+	slugify: require("./slugify"),
+	ternary: require("./ternary"),
+	tojson: require("./tojson"),
+	tz: require("./tz"),
+	users: require("./users"),
+};
```
astorije: Was that unnecessary before due to the Webpack/Handlebars integration?
Does this make [that](https://github.com/thelounge/lounge/blob/8658f15751606da72809dd02545a36b4cf7813bc/webpack.config.js#L52-L54) unnecessary?

src/server.js
```diff
@@ -29,7 +31,10 @@ module.exports = function() {
 	var app = express()
 		.use(allRequests)
 		.use(index)
-		.use(express.static("client"));
+		.use(express.static("client"))
+		.engine("html", expressHandlebars({extname: ".html", helpers: require("../client/js/libs/handlebars")}))
+		.set("view engine", "html")
+		.set("views", path.join(__dirname, "..", "client"));
```
astorije: I don't know of an alternative but this feels... awkward to me. @xPaw, what do you think?

astorije:
This seems sweet @YaManicKill. Can you specify in the PR description why we need to do that? Not that I'm against it or anything, but it adds some complexity and the benefits are unclear at the benefits.
I have probably missed a discussion about that, but it would be good to have for future references.

client/js/libs/handlebars/firstLetterUppercase.js
YaManicKill: Yeah, will do.

client/index.html
YaManicKill: Yeah, I saw them as well, which was good, but it didn't work. I also can't find it anywhere on their official documentation. I imagine it was something that worked a while ago but was never deliberate. If I try it, I get an error now.

package.json
YaManicKill: Ah yes.

client/js/libs/handlebars/index.js
YaManicKill: Yeah, it isn't necessary because with the webpack handlebars loader, we just point it at a folder and it takes all of them and registers them as handlebars helpers.

The same doesn't exist for anything express (that I could find, feel free to find one if it exists). We could write something for it, but meh. The one that exists for express, you have to specify everything specifically. Not ideal, but meh.

src/server.js
YaManicKill: So, this is the way you are actually **meant** to do views in express. The way we were doing it manually caused us lots of weirdnesses, and when I tried to just throw our existing file into handlebars it didn't seem to work. so I decided to simplify it and use res.render. It works well and just a little bit of extra set up.

package.json
YaManicKill: Like, do you mean run webpack on the server, or do you mean just render the index on the client.

If you mean server, then no, let's not do that, that's horrible.

If you mean on the client, we could, but it would increase our initial render time because currently we render it on the server before passing it back, if we passed back the index.html (which will be smaller soon) and all the other templates, plus the data for the client to render, before we can do any connections to the socket etc, that will increase it quite a bit. So, I'd really rather not. It's just 1 extra dependency and doesn't seem that big a deal.

YaManicKill:
> Can you specify in the PR description why we need to do that?

Yeah, sure. I'll leave it in this comment as well, just for the conversation.

Basically, index.html is huge (as explained in #1055) and we currently use lodash's templating to make the file. The problem is that lodash doesn't have an "include" function, meaning that we can't split the file up into multiple (think, 1 file for the help page, 1 for the settings page, etc). So if we want to split it up, we need to switch to something else. As we use handlebars on the client, it felt like a good situation to use handlebars for the initial index.html as well.

So to be consistent, and to allow us to split do #1055, we need to switch to handlebars.

client/index.html
astorije: :man_facepalming:
I mean, it feels so awkward to have to roll-your-own ternary operator. I get the minimalist approach, but still frustrating.

Looking at all uses of `ternary`, I think we should go for native `{{#if public}}public{{/if}}`, especially since all have an empty string in their `else` block.

Kind of a nitpick, sorry, but I think we should embrace the Handlebars way instead of defining our programming language 😕.

client/index.html
```diff
@@ -122,75 +122,81 @@ <h1 class="title">Sign in to The Lounge</h1>
 						<div class="row">
 							<div class="col-sm-12">
 								<h1 class="title">
-									<%= public ? "The Lounge - " : "" %>
+									{{ternary public "The Lounge " " " }}
```
astorije: Looks like a typo here. I think you meant

```handlebars
{{ternary public "The Lounge - " "" }}
```
or
```handlebars
{{#if public}}The Lounge - {{/if}}
```
which I prefer as more native (and also find more readable due to the quotes, but maybe that's personal).

src/server.js
astorije: Sorry, I highlighted the wrong line, but I'm glad you detailed this as well. I agree though I need to read up a bit on it to make sure I'm following.

What I meant to highlight was the fact that we require a client subfolder in a server file... I don't know, feels wonky. I won't hold on that obviously, just sharing thoughts.

package.json
xPaw: I mean add it to our current build step. Then compiled index.html could be put in a public folder. This would require stuff like version to be added via js or some other step, since they would be hardcoded on compile. But I think this approach would be better overall, since we compile other handlebars templates with webpack already.

astorije: Webpack on the server, I'm hoping @xPaw did not mean that 😅.

Apart from that, it's usually a good practice in webapps to pre-render as much as you can so `index.html` comes back from server with all it needs already there, so moving rendering on the client would not necessarily be a bonus.

astorije:
Thank you for the details @YaManicKill, it makes a lot of sense!

xPaw:
You're right, ternary operations can just be ifs.

client/index.html
```diff
@@ -261,14 +267,14 @@ <h1 class="title">Settings</h1>
 							<div class="col-sm-12">
 								<label for="theme-select" class="sr-only">Theme</label>
 								<select id="theme-select" name="theme" class="input">
-									<% themes.forEach(function(themeName) { %>
-										<option value="<%= themeName %>">
-											<%= themeName.charAt(0).toUpperCase() + themeName.slice(1) %>
+									{{#each themes}}
+										<option value="{{this}}">
+											{{firstLetterUppercase this}}
```
xPaw: Can we just not capitalise this. Quite pointless to add such a helper function.

package.json
xPaw: Okay, looking at the theme, it would be quite hard to pre compile.

src/server.js
xPaw: I wonder if we can simplify this since we only render index with handlebars.

YaManicKill: We can change it, but it really isn't simplified. You can speifically just use fs to grab the file as we have done before, and throw it into handlebars, but I couldn't actually get it to work and it definitely isn't simpler. This also allows us more flexibility to handle more files if we want to.

@astorije We could move it to a common folder if you want, or we can move the ones I created into the server and keep them seperate (as we don't use either of the same ones yet) but I don't think it's that big of an issue.

client/index.html
astorije: I think the capitalized version looks way less sloppy. Since we can't have JS in the template (I believe), it kinda makes sense we need a helper.
@YaManicKill, I would call this `capitalize`, if that makes sense. I can see how we would need such a helper in other places as our code grows.

package.json
YaManicKill: Yeah, so a couple of things:

1. We don't compile handlebars in our build step. We just bundle them together. The browser compiles them when we need them. We can't compile those templates because we don't know the data yet.
2. We could do this for this file, but that would require rerunning the build whenever you change a config or add a theme or anything like that, which I would be very strongly against.

client/index.html
YaManicKill: Yeah, the "-" was missed.

client/index.html
YaManicKill: Actually yeah, you are right. I don't know why I didn't think of using the proper if helper, that is better. I shall do that.

client/index.html
YaManicKill: What do you mean?

xPaw: Do it on the server then, same place where .CSS gets stripped. I think that place fits more, as we eventually want to have theme manifests, which would define proper theme names.

YaManicKill: I think that we don't want to remove helpers just for the sake of removing helpers. This is the way you are meant to do handlebars, and we do it for loads of stuff in the client.

I'm happy to call it "capitalize" but I thought that wasn't obvious enough because it isn't capitalising the whole thing, just the first letter. But if you think it's self explanatory enough, I guess.

YaManicKill: We can do it on the server, but that does mean completely chaning the data structure for themes, as we need the lowercase ones as well.

It would go from ["default","crypto"] to [{"name": "Default", fileName: "default"},{"name": "Crypto", fileName: "crypto"}]

Are you both happy with this?

dgw: `capitalize` is too generic, IMO. `ucFirst` or something like that can be used if the current name is too long, but it's ambiguous exactly what a function named `capitalize` would capitalize.

xPaw: I think we need a structure either way. One string to define actual theme name (internal/file), other with display friendly name. This would slowly prepare us for a better theme system.

astorije: I don't care re: naming. I'm not native speaker so I would probably look at stuff like https://atom.io/packages/change-case or the like, see what people do.

Re: structure of theme, I don't think this PR should care about that. My opinion is to just port it in this PR then deal with themes in a further PR. It seems out-of-place and an easy source of bikeshedding while this PR is mostly ready. Just my 2c.

YaManicKill: Alright, I'll do it, then.

Oh look, now we have no handlebars helpers for this file, isn't that handy...

client/js/libs/handlebars/firstLetterUppercase.js
YaManicKill: This is no longer in this file, it's in the server. So it's a bit more complicated to test just now. I could pull it out into a module but that feels overkill fo rsomething so small just now.

client/index.html
YaManicKill: Done, that's much nicer :-)

package.json
YaManicKill: Done

client/js/libs/handlebars/index.js
YaManicKill: This doesn't exist anymore because we don't use any helpers. So we will need to figure this out when we want to use any of the helpers, but for now it is not a problem.

src/server.js
YaManicKill: The require doesn't happen anymore, so that's easier.

But I'm still using the express rendering because I think that's much simpler than the way we were previously doing it, and will make other things much easier to add on in the future.

client/index.html
YaManicKill: Done

client/index.html
YaManicKill: I moved this into the server and just did the simple structure. I accept that we could easily bikeshed, so please ... unless there's a very good reason don't spend time arguing over the structure. This way we don't have to rename the helper function (because, you know, that's a big bikeshed as well) and it has simplified other things as well. We can think about a better structure when we get the theme manifest done sometime down the line.

YaManicKill:
Should be up to date now with all your comments. It's much simpler without the helpers, which is obviously good.

client/index.html
```diff
@@ -122,75 +122,81 @@ <h1 class="title">Sign in to The Lounge</h1>
...
 								<h1 class="title">
-									<%= public ? "The Lounge - " : "" %>
+									{{#if public}}The Lounge - {{/if}}
 									Connect
-									<%= !displayNetwork && lockNetwork ? "to " + defaults.name : "" %>
+									{{#unless displayNetwork}}
+										{{#if lockNetwork}}
+										to {{defaults.name}}
+										{{/if}}
+									{{/unless}}
```
astorije: This is gross, but not your fault...

client/index.html
```diff
@@ -122,75 +122,81 @@ <h1 class="title">Sign in to The Lounge</h1>
...
 								</h1>
 							</div>
-							<div <%= typeof(displayNetwork) !== "undefined" && !displayNetwork ? 'style="display: none;"' : ''%>>
+							{{#unless displayNetwork}}
```
astorije: So we stop assuming that these can be not found in the configuration? I'm fine with it because we default to the config file that's used to create the local config at startup time, but I just want to make sure we're all on the same page with that one.

xPaw: We've changed our config long ago to be bootstrapped from example config, so missing keys in users config are set to default values.

PR #1060, opened 2017-04-21
Create socket module

YaManicKill:
Have a small PR. Let's see how this goes.


xPaw:
I realised that modularising code like this might reveal code smells that we have. Win-win overall.

astorije:
lol pretty sure what triggered your code smell detector is that big comment that just screams my name

PR #1064, opened 2017-04-22
Rewrite server code of channel sorting

xPaw:
Sending malformed `order` would crash the server. Sending incomplete `order` would lose references to channels or networks.

[](https://vimeo.com/180103193)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -412,44 +412,40 @@ Client.prototype.open = function(socketId, target) {
 };
 
 Client.prototype.sort = function(data) {
-	var self = this;
+	const order = data.order;
 
-	var type = data.type;
-	var order = data.order || [];
-	var sorted = [];
+	if (!_.isArray(order)) {
+		return;
+	}
 
-	switch (type) {
+	switch (data.type) {
 	case "networks":
-		order.forEach(i => {
-			var find = _.find(self.networks, {id: i});
-			if (find) {
-				sorted.push(find);
-			}
+		this.networks.sort((a, b) => {
+			return order.indexOf(a.id) > order.indexOf(b.id);
 		});
-		self.networks = sorted;
+
+		// Sync order to connected clients
+		this.emit("sync_sort", {order: this.networks.map(obj => obj.id), type: data.type, target: data.target});
+
 		break;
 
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -412,44 +412,40 @@ Client.prototype.open = function(socketId, target) {
...
+	switch (data.type) {
 	case "networks":
-		order.forEach(i => {
-			var find = _.find(self.networks, {id: i});
-			if (find) {
-				sorted.push(find);
-			}
+		this.networks.sort((a, b) => {
+			return order.indexOf(a.id) > order.indexOf(b.id);
 		});
```
astorije: Assuming about Node's (unstable) sorting algorithm, this is at best `O(n² * log n)`. Some people hang out in easily dozens of channels: 50 channels and that's 14k operations, 100 channels is 64k operations, etc.

I know this is not worse than it was before (I think), but do you think there is any way we can improve that? The user provides the order themselves, so in my understanding we shouldn't have to do a full sort, instead loop through the given order (which would be `O(n²)`, not much better, surely there is a better way...)

Also, when is this needed exactly? When a user reorders their channels/networks on the UI?

xPaw: Can't rely entirely on user given input (will have to filter it to exactly contain every channel ID on the network, missing channels have to be added, invalid ones removed. This was the simplest approach.

This could be improved by modifying order into a hash map so that no indexOf calls are required.

This method is entirely just for the UI sorting.

astorije: Ok, this can be done later though, your fix is a good thing anyway.

src/client.js
```diff
@@ -412,44 +412,40 @@ Client.prototype.open = function(socketId, target) {
...
+
 		break;
 	}
 
-	self.save();
-
-	// Sync order to connected clients
-	const syncOrder = sorted.map(obj => obj.id);
-	self.emit("sync_sort", {order: syncOrder, type: type, target: data.target});
+	this.save();
```
astorije: Specific reason why you separated the event emit and file save?

xPaw: Nothing specific, just didn't want to keep the variable.

PR #1066, opened 2017-04-22
Create options module

YaManicKill:


client/js/localStorage.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+module.exports = {
```
xPaw: Should this implement `get` and `remove` for consistency?

YaManicKill: Yeah, probably should.

YaManicKill: Added them :+1:

client/js/localStorage.js
```diff
@@ -0,0 +1,19 @@
...
+			// Do nothing. If we end up here, web storage quota exceeded, or user is
+			// in Safari's private browsing where localStorage's setItem is not
+			// available. See http://stackoverflow.com/q/14555347/1935861.
+		}
+	},
+	get: function(key) {
+		return window.localStorage.getItem(key);
+	},
+	remove: function(key, value) {
+		window.localStorage.removeItem(key, value);
```
williamboman: `removeItem` doesn't take a value does it?

YaManicKill: You are correct, copy/pasta error.

YaManicKill: Of course it doesn't cause a bug, because it just ignores the extra parameter. But still, I'll remove it :-P

client/js/localStorage.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+module.exports = {
+	set: function(key, value) {
```
astorije: Note that ES6 introduced the method definition shorthand (which you're probably aware of but just reminding here if useful to anyone):

```js
set(key, value) { ...
```

Not a big deal at all but might be worth using in your future modularization work.

astorije:
Merging regardless of my minor comment, as it's mostly style preference.

Issue #285, opened 2016-04-27
Banlist view


YaManicKill:
It would be nice if we had a nice way for ops to view the banlist and add/remove rather than having to figure out the command. It's something I do so infrequently I never remember, but I want it there.

I was thinking if we had an entry in the context menu for channels, and then a modal popped up with the banlist, with "x" next to each entry to remove from the banlist. Banning a user would probably work through the context menu on a user itself.

vectr0n:
+1 the same for +e (ban exception) and +I (invite exception) lists would be nice as well.

astorije:
@YaManicKill, I believe you changed https://github.com/thelounge/lounge/pull/1009 to include that GUI, right? If so, we should close this issue and add the 2.3.0 milestone.

YaManicKill:
@astorije Nah, it was just a special channel to show the images in. This doesn't allow adding/removing from the banlist, just viewing of what is on it.

astorije:
Got it. Your commands module PR links to this issue so I was a bit
confused. All good now

YaManicKill:
Ah woops, I meant to link to #1009, mistake.

PR #1067, opened 2017-04-22
Move commands into constants module

YaManicKill:


YaManicKill:
This will conflict with #285 so whichever gets in first will need to deal with adding that to the module

astorije:
Can someone resolve conflicts on this?

YaManicKill:
Resolved conflicts.

PR #1077, opened 2017-04-24
Add ban/unban command

### LINKED ISSUE #1073: Add ban and unban commands
Currently, users have to do "/mode username +b" and "/mode username -b". Many irc clients have /ban and /unban as links to that.

Also, [irc-fw just added some nice new commands for these (and banlist)](https://github.com/kiwiirc/irc-framework/pull/89/files) so we should be able to simplify the /mode command quite a bit (it's quite ... interesting).

YaManicKill:
Fixes #1073

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -556,6 +556,16 @@
 
 						<div class="help-item">
 							<div class="subject">
+								<code>/ban nick</code>
+							</div>
+							<div class="description">
+								<p>Ban (<code>+b</code>) a user from the current channel.
+								This can be a nickname or a hostmask.</p>
+							</div>
+						</div>
+
+						<div class="help-item">
+							<div class="subject">
 								<code>/banlist</code>
 							</div>
 							<div class="description">
@@ -796,6 +806,16 @@
 							</div>
 							<div class="description">
 								<p>Set the topic in the current channel.</p>
+							</div>
+						</div>
+
+						<div class="help-item">
+							<div class="subject">
+								<code>/unban nick</code>
+							</div>
+							<div class="description">
+								<p>Unban (<code>-b</code>) a user from the current channel.</p>
+									This can be a nickname or a hostmask.</p>
 							</div>
 						</div>
 
--- a/src/plugins/inputs/ban.js
+++ b/src/plugins/inputs/ban.js
@@ -1,0 +1,44 @@
+"use strict";
+
+var Chan = require("../../models/chan");
+var Msg = require("../../models/msg");
+
+exports.commands = [
+	"ban",
+	"unban",
+	"banlist"
+];
+
+exports.input = function(network, chan, cmd, args) {
+	if (chan.type !== Chan.Type.CHANNEL) {
+		chan.pushMessage(this, new Msg({
+			type: Msg.Type.ERROR,
+			text: `${cmd} command can only be used in channels.`
+		}));
+
+		return;
+	}
+
+	if (cmd !== "banlist" && args.length === 0) {
+		if (args.length === 0) {
+			chan.pushMessage(this, new Msg({
+				type: Msg.Type.ERROR,
+				text: `Usage: /${cmd} <nick>`
+			}));
+
+			return;
+		}
+	}
+
+	switch (cmd) {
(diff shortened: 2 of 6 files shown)


src/plugins/irc-events/banlist.js
```diff
@@ -9,7 +9,7 @@ module.exports = function(irc, network) {
 	irc.on("banlist", function(banlist) {
 		const channel = banlist.channel;
 		const bans = banlist.bans;
-		if (!bans) {
+		if (!bans || bans.length === 0) {
```
YaManicKill: It appears that we now get through an empty array rather than undefined when there are no bans. So this needs updated. This was added in [irc-framwork](https://github.com/kiwiirc/irc-framework/commit/1cc8892f6bf5414e677f0d37daa5984b77b58374).

src/plugins/inputs/ban.js
```diff
@@ -0,0 +1,44 @@
...
+				text: `Usage: /${cmd} <nick>`
+			}));
+
+			return;
+		}
+	}
+
+	switch (cmd) {
+	case "ban":
+		network.irc.ban(chan.name, args[0]);
```
xPaw: Will this ban just the nick? I think most channels do hostname bans.

YaManicKill: It'll bad whatever you pass it, so it'll either be nick or hostmask depending on what you pass it.

"/ban test" will ban the user
"/ban *!*@test.com" will ban that hostmask

MiniDigger: typing out hostnames is meh, what about a /banh command to ban the hostname of a user?

YaManicKill: I think that's out of scope of this PR. I would expect that ban doesn't automatically assume I want to ban the hostname or anything like that, and just dumbly send it on. So if we want a new command, that's fine, but that's not part of this PR.

xPaw: Bans only apply to hostmasks though, do all server support simply sending a ban for a nickname?

YaManicKill: I mean, this isn't any new functionality, because if we send a "/mode +b username" then it's exactly the same. All this does is basically symlink "/ban" to "/mode +b" with some checks that you are sending something. 

[The spec is vague on it.](https://tools.ietf.org/html/rfc1459#section-4.2.3)

client/index.html
```diff
@@ -556,6 +556,15 @@ <h1 class="title">Help</h1>
 
 						<div class="help-item">
 							<div class="subject">
+								<code>/ban user</code>
```
astorije: All other helpers are `/foo nick`, is this one different on purpose? If it's to highlight that not only nicks can be passed, then I suggest `/ban target` and then in the description `<code>target</code> can be a nickname, or a...` (with a better wording than what my 5-yo English can come up with lol).

src/plugins/irc-events/banlist.js
astorije: Kind of a breaking change on their end if you ask me, but oh well. Good catch!

YaManicKill: Yeah, probably, but too late now :-P

YaManicKill:
> (You do have a passion for banning people, don't you? :D)

What can I say? I don't like people :-P

client/index.html
YaManicKill: Nah, accidental. I'll add it just as "nick" but point out it can be a hostmask,

YaManicKill: Done.

client/index.html
```diff
@@ -801,6 +811,15 @@ <h1 class="title">Help</h1>
 
 						<div class="help-item">
 							<div class="subject">
+								<code>/unban user</code>
```
astorije: Aaaand the other one? :D

YaManicKill: https://media.giphy.com/media/Kh296M49Zr1Sg/giphy.gif

Goodness, it's clearly too early.

astorije: And I thought I was the one who hasn't slept all night :P

PR #1078, opened 2017-04-24
Add fix for slow scrolling when holding pg-up/pg-dn

### LINKED ISSUE #1022: Holding pageup/pagedown scrolls very slowly
What I assume happens is that previous animation gets stopped, and it scrolls up from that position where the animation stopped.

YaManicKill:
Fixes #1022

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1271,6 +1271,9 @@ $(function() {
 			"pagedown"
 		], function(e, key) {
 			let container = windows.find(".window.active");
+			if (container.is(":animated")) {
+				return;
+			}
 
 			// Chat windows scroll message container
 			if (container.attr("id") === "chat-container") {


xPaw:
Well, this looks easier fix than i was initially trying to do. However this means pressing pgup twice will only scroll once.

YaManicKill:
> However this means pressing pgup twice will only scroll once.

Hmmm, true, but you have to be pretty quick and pressing it. In my testing, the way I would normally press would be slow enough that it would trigger twice.

I'm not sure, what was your suggestion?

astorije:
Testing on my browser, hitting page up twice "accumulates" both events. Not sure if we can do that easily though. I'll defer decision to @xPaw for this PR.

Also, with this change, is `.stop()` L1292 still necessary?

PR #1080, opened 2017-04-24
Replace the state on init rather than adding a new entry

### LINKED ISSUE #1042: Back button needs 2 presses to go back to another page (or quit)
> Testing this after your fix, I'm still seeing something odd: say I'm on page A (not The Lounge), then type the URL of The Lounge (localhost when testing), then I have to hit "Back" twice to go back to A.

YaManicKill:
Fixes #1042

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -132,7 +132,9 @@ $(function() {
 		$("#sign-in").remove();
 
 		var id = data.active;
-		var target = sidebar.find("[data-id='" + id + "']").trigger("click");
+		var target = sidebar.find("[data-id='" + id + "']").trigger("click", {
+			replaceHistory: true
+		});
 		if (target.length === 0) {
 			var first = sidebar.find(".chan")
 				.eq(0)
@@ -820,7 +822,11 @@ $(function() {
 		}
 
 		if (history && history.pushState) {
-			history.pushState(state, null, null);
+			if (data && data.replaceHistory && history.replaceState) {
+				history.replaceState(state, null, null);
+			} else {
+				history.pushState(state, null, null);
+			}
 		}
 	});
 


astorije:
@xPaw, @williamboman, what do you think?

YaManicKill:
Ah, maybe I should explain what the deal is with this.

Basically, what happened before was we loaded the main page, which added the base page to the history, and then we clicked on the channel we want to load, which added another entry to the history. Instead, this replaces the initial page load in the history with the channel we load instead.

williamboman:
Never heard of `replaceState`. If this works I'm 👍

YaManicKill:
@williamboman You can read about it [here](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_replaceState()_method).

Example of it working:

a1a7c969478dabc6db30fb854c3453d1

Without the fix:

e61649dd10b588186c6ddde23411b93c

client/js/lounge.js
```diff
@@ -140,7 +140,9 @@ $(function() {
 		$("#sign-in").remove();
 
 		var id = data.active;
-		var target = sidebar.find("[data-id='" + id + "']").trigger("click");
+		var target = sidebar.find("[data-id='" + id + "']").trigger("click", {
+			replace: true
```
astorije: Nitpick, could you rename this into `replaceState`, or `replaceHistory`, or something like that? `replace` is too generic for my taste (we could need it later, but even now it's hard to figure out what it means without reading your code below since event handling is so "detached" compared to normal function calls).

astorije:
Tested locally and this works great 👏

client/js/lounge.js
YaManicKill: Done.

PR #1082, opened 2017-04-24
Use irc-framework setTopic() for topic command

MaxLeiter:
Possible due to #1070

### DIFF
--- a/src/plugins/inputs/topic.js
+++ b/src/plugins/inputs/topic.js
@@ -14,9 +14,6 @@ exports.input = function(network, chan, cmd, args) {
 
 		return;
 	}
-
-	var irc = network.irc;
-	irc.raw("TOPIC", chan.name, args.join(" "));
-
+	network.irc.setTopic(chan.name, args.join(" "));
 	return true;
 };


src/plugins/inputs/topic.js
```diff
@@ -14,9 +15,6 @@ exports.input = function(network, chan, cmd, args) {
 
 		return;
 	}
-
-	var irc = network.irc;
-	irc.raw("TOPIC", chan.name, args.join(" "));
-
+	client.setTopic(chan.name, args.join(" "));
```
astorije: Why not use `this`?

MaxLeiter: client is more clear imo, also consistent  with other files, but nbd to change

astorije: Need to change on L10/R11 as well then. I'll approve after that and let @xPaw decide if he cares or not.

xPaw: It should be `network.irc` though.

MaxLeiter:
Actually got on the laptop and fixed it. Good now.

xPaw:
Did you test that `/topic` still retrieves the topic?

MaxLeiter:
Still works, @xPaw

Issue #1071, opened 2017-04-23
Text Rendering


bews:
Don't know why (intentionally or it's just a bug) right now text is rendering in subpixel mode (https://en.wikipedia.org/wiki/Subpixel_rendering), which is looking bad on my display (TFT TN LED).

Now:
![lounge_a1](https://cloud.githubusercontent.com/assets/3627488/25311724/c35f2a0a-2810-11e7-80bb-40e50d19c22a.png)

Before (Shout):
![lounge_a2](https://cloud.githubusercontent.com/assets/3627488/25311730/e0a342b8-2810-11e7-8ee4-4d6948422eb1.png)

So, I would like to make a fix, which will change this rendering into old one. What do you think?

YaManicKill:
I'm not sure on the advantages/disadvantages of different text rendering, but from what I can see it depends on what type of display you have? Is there an easy way for us to figure out different display types and use the right rendering for them? Feel free to correct me if I'm wrong, this is a very weak area of my knowledge.

bews:
It's impossible to detect matrix type or display model.
As for advantages, I don't know. I think if some displays have some unique features then they should have special drivers, which change rendering everywhere.

We can make rendering type changeable in settings, but is it really needed? I still believe it was a bug in the first place.

xPaw:
So how would you fix it? We don't set subpixel rendering anywhere besides tooltips.

bews:
> So how would you fix it?
https://github.com/bews/lounge/commit/de666c33f60acb04085c0f8b11df5d6b0d0cc714

astorije:
Does this break sidebars?

bews:
No, I haven't noticed any side effects.

PR #1083, opened 2017-04-24
Rendering fix (subpixel -> old one)

bews:
Was discussed in #1071

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -787,8 +787,6 @@ kbd {
 	right: 0;
 	width: 180px;
 	transition: all .4s;
-	-webkit-transform: translateZ(0);
-	transform: translateZ(0);
 }
 
 #chat .lobby .chat,
@@ -1852,11 +1850,6 @@ kbd {
 }
 
 @media (min-width: 769px) {
-	#viewport {
-		-webkit-transform: none !important;
-		transform: none !important;
-	}
-
 	#viewport.menu-open {
 		transition: none;
 	}


YaManicKill:
So, I have no idea what this code was even meant to do in the first place. @thelounge/maintainers anyone?

xPaw:
It's a trick to force certain elements to be rendered on the GPU.

YaManicKill:
Alright, is there a reason we are currently doing that?

astorije:
I believe it was added for opening/closing channel list and user list, but why that was deferred to the GPU, that I have no idea...
Maybe the animation wasn't super smooth on mobiles?

astorije:
Also, just to be clear, you reported this as a regression from Shout, but this was introduced 3 years ago by erming himself, so I'm not sure how The Lounge regressed from it :)

xPaw:
Opening the sidebar uses translate3d, probably yes, to force GPU before it starts. Forcing GPU layer is also helpful for scroll performance, I was debugging some perf stuff in devtools and doing such a trick improved things. I cant provide more details right now as I am on mobile. 

What i want to know, why does translateZ of 0 does different font rendering?

astorije:
Note that it's not the first time that someone reports that the text in the user list is not as clean as the text of the chat window. Maybe some GPUs do not render things correctly and make simplifications instead?

bews:
> Also, just to be clear, you reported this as a regression from Shout, but this was introduced 3 years ago by erming himself, so I'm not sure how The Lounge regressed from it :)

Just checked myself, you're right. But Shout doesn't use subpixel rendering with this. Guess `transform: translateZ(0);` should be combined with something else make this change, but I couldn't find it. So the easiest way to fix this is to remove those lines.

astorije:
@xPaw, @YaManicKill, what do we want to do with this?

xPaw:
So this PR removes animation when opening or closing the channel list (with hamburger button).

bews:
> So this PR removes animation when opening or closing the channel list (with hamburger button).

No, it doesn't. What browser do you use? I can't reproduce it here.

> Can you test if adding will-change: transform fixes your issue? 

Yes, replacing 
```
-webkit-transform: none !important;
transform: none !important;
```
with 
```
will-change:transform;
```
fixes rending too. Should we use this way?

client/css/style.css
```diff
@@ -787,8 +787,6 @@ kbd {
 	right: 0;
 	width: 180px;
 	transition: all .4s;
-	-webkit-transform: translateZ(0);
-	transform: translateZ(0);
```
bews: It's really strange, but sidebar rendering was fixed by one of the recent commits, so it's not necessary to remove these lines anymore. Magic! Although there is a chance that this problem will appear again in future.

astorije:
> It's really strange, but sidebar rendering was fixed by one of the recent commits, so it's not necessary to remove these lines anymore.

Should we close this PR (for now at least) then?

bews:
> Should we close this PR (for now at least) then?

No, chat block is still broken, so at least other 2 lines should be removed.

client/css/style.css
```diff
@@ -1852,11 +1850,6 @@ kbd {
 }
 
 @media (min-width: 769px) {
-	#viewport {
-		-webkit-transform: none !important;
-		transform: none !important;
-	}
-
```
bews: These ones.

astorije:
According to [this link](https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas) given by @xPaw in https://github.com/thelounge/lounge/pull/1120, `translateZ(0);` was used to improve performance in Safari and Safari Mobile.

Also, I'm thinking https://github.com/thelounge/lounge/pull/1120 will supersede this, won't it, @xPaw? If so we should close for now.

PR #1087, opened 2017-04-25
Fix showing prefetch options

YaManicKill:


YaManicKill:
Minor fix, merging.

astorije:
Good catch. Make sure to add the correct label and milestone next time you merge though :D

YaManicKill:
Pssst, can we not have a bot that does that? ;-)

PR #1091, opened 2017-04-26
Fix broken nick autocomplete caused by #856

PolarizedIons:
Since #856 changed the structure of the userlist, it couldn't find the users to use to autocomplete


xPaw:
Thanks!

astorije:
Woops, good catch and great community work!!

Issue #1092, opened 2017-04-26
Userbar causing spacing/width issues


PolarizedIons:
The userbar automatically hides or shows under circumstances, but the width of the chat/channel doesn't change correctly leading the stuff like this:

https://gfycat.com/ThreadbareHappyHectorsdolphin
https://i.imgur.com/v7C7vvv.png

This is on master (06af65c)

xPaw:
Yeah, this is responsive bullshit. However, has this been recently broken in such a way, or always has been?

MaxLeiter:
cant recreate on [5281411](https://github.com/thelounge/lounge/tree/5281411)

xPaw:
Can you bisect it?

PolarizedIons:
bisecting pointed to c583d6e

xPaw:
I can't reproduce this. Could you figure out what CSS is doing this effect in dev tools?

PR #1093, opened 2017-04-26
Fix network layout displaying the scrollbar incorrectly

astorije:
This was introduced by [this change R773](https://github.com/thelounge/lounge/pull/856/files#diff-97db1f70168fb5f12457b238ff6052b5R773) (and L794-798): a right position got introduced for all channel containers, but default position for other types of containers was absent before this script.

Before | After
--- | ---
screen shot 2017-04-26 at 23 32 08 | screen shot 2017-04-26 at 23 30 33

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -762,6 +762,7 @@ kbd {
 #chat .chat {
 	bottom: 0;
 	left: 0;
+	right: 0;
 	overflow: auto;
 	-webkit-overflow-scrolling: touch;
 	position: absolute;


astorije:
@PolarizedIons, could you let me know if this happens to also fix https://github.com/thelounge/lounge/issues/1092 by any chance?

PolarizedIons:
That did fix another issue I wasn't sure of if it was related, but no. It does not fix the issue described in #1092

PR #1095, opened 2017-04-27
Fix nick autocomplete

### LINKED ISSUE #1119: JS exception when typing for a private message
To reproduce:

- Open your browser JS console
- Open a private message window (`/whois` or click, didn't try other methods)
- Start typing and enjoy the following in your console:
  screen shot 2017-04-30 at 04 23 16

xPaw:
Fixes #1119.

Also simplifies code by getting rid of the `move` function.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -300,8 +300,9 @@ $(function() {
 			var nicks = chan.find(".users").data("nicks");
 			if (nicks) {
 				var find = nicks.indexOf(data.msg.from);
-				if (find !== -1 && typeof move === "function") {
-					move(nicks, find, 0);
+				if (find !== -1) {
+					nicks.splice(find, 1);
+					nicks.unshift(data.msg.from);
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


MaxLeiter:
Can confirm works on top of master 👍

client/js/lounge.js
```diff
@@ -1534,17 +1535,6 @@ $(function() {
...
-		if (new_index >= array.length) {
-			var k = new_index - array.length;
-			while ((k--) + 1) {
-				this.push(undefined);
-			}
-		}
-		array.splice(new_index, 0, array.splice(old_index, 1)[0]);
-		return array;
-	}
-
```
astorije: Shouldn't we get rid of https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L303-L305 as well then? Also, why was that a `typeof` check on a function of ours?!

xPaw: That code was replaced, look at this diff line 303. `typeof` because bad Shout code.

astorije: Next step, I learn to read.

PR #1096, opened 2017-04-27
Better user colors

bews:
This PR improves 3 things:
1) User colors now more distinctive, ~~number increased 32 -> 70~~ - [preview](https://colors.bews.me/test.html)
2) ~~Now it's possible to override user color for any nickname in settings.~~
3) Nicknames in chat is now colored.

-------------------
Things to do:

- [x] Reduce number of colors 
-  ~~Refactor color overriding~~
- [x] Fix colored mentions

-------------------
**User colors**
Default user colors are not so good:

I tried to fix them, but generating even remaining colors turned out to be difficult >_>
So I decided to go deeper. And after some sleepless nights it is finished: [color generator](https://colors.bews.me/)
Click "Generate" button to see colors used in this PR.

-------------------

**~~Color overriding~~**
~~Even with 70 colors users can have the same color. And this become awful when you chat with your friends and everyone is... pink.
So now you can set any color you want for any nickname:~~

### EDIT
This feature was removed for now. Needs a better implementation.

-------------------

**Colored nicknames in mentions**

lounde_u1


maxpoulin64:
For reference, I attempted something similar in #357. Generating readable colors is a real pain! I originally wanted to make a algorithm that could bring back the old behavior of dynamically generating colors based on some hash rather than the fixed colors.

I like the color override for nicks, could greatly help people stop changing nicks!

bews:
BTW this code does nothing:
lounge_c

I guess it was supposed to sort nicknames for autocomplete, but broke somehow. So I decided to keep it for now:


This function is also broken:
lounge_c2

astorije:
Hi @bews,

Waow, what a greatly documented PR!

So, to individually comment the 3 enhancements you made:

> 1. User colors now more distinctive, number increased 32 -> 70.

Sorry but that would be a definitive no from me 😅
As mentioned in the IRC channel (I think, or maybe an issue?), UX-wise human brains can only dissociate about 13 colors given a single page. I do not have source, this was shared to me by a UX expert. So to be taken with a grain of salt, but we went for 32 as twice 16, the closest power of 2 (up) to 13. So going to 70 is way overkill IMO. Whether colors are the same or not, that many colors will always bring undistinguishable colors whatever the number of tries.

Case in point, I just tried your generator with default values and the last colors are:

screen shot 2017-04-28 at 01 07 25

That's 29 colors, where really what's usable is only dark blue, dark/light purple, and pink. That's 4 colors, make it 8 (double) to add a bit of shade headroom, and when we extrapolate to the ration of added / current numbers of colors, that makes only 19. With 32, it means we even have extra values.
So maybe other maintainers strongly disagree with this, but personally I don't think we should do this.

Also, I mentioned power of 2 above. If we were to do this, we would do 64 anyway to keep it a power of 2. Easier to evenly drop the number of colors, for example to make an accessible theme with very few colorblind-friendly colors.

> 2. Now it's possible to override user color for any nickname in settings.

This is actually already possible using the custom CSS, as each nick has a `data-name` attribute.

```css
.user[data-name^='Bews'] { /* Would match both `Bews` and `Bews2` */
  color: #000000;
}
```

I know that anti-settings rep of mine will be mentioned again (which is untrue :D), but I do think a setting for this alone does not bring much more value than custom CSS. I actually think it's a bit nicer because you can use a standard syntax (CSS) *and* you are not limited to color but can also make it bold, change the case, etc.

> 3. Nicknames in chat is now colored.

This can be cool!
I'm wondering if we should not make this part of the message parser though, at least for consistency but also because it's nice and clean. Then we could also make the nicks clickable (and maybe later get the custom highlight list from that same parser?).

What I'm fairly concerned about however is the potential amount of false positives coming with this, because nicks do not have restrictions like channel names have. I'm guessing this is the main reason why we don't parse nicks at the moment.
I'm torn about this, but I'd be happy to give a shot if we all agree on the details and consequences.

---

Thanks for opening this, @bews, I wish I was a bit more on board with the changes, but it's always nice to see contributions anyway 😃

bews:
> human brains can only dissociate about 13 colors given a single page

Well, I can dissociate ~95% of these 70 colors (superpower, yay)

> So going to 70 is way overkill IMO.

There is a problem: string to color number function. The fewer colors available - the more users will have the same colors. IMO even a little different color is better than the same:



This is the main reason why I want increase color list.

> we would do 64 anyway to keep it a power of 2. 

70 -> 64 is very easy change (just pick colors you don't like and add them to ignore list).
Also I think shuffling colors will give better result, because str -> color function tends to pick close numbers for similar names. Will do this tomorrow.

>  setting for this alone does not bring much more value than custom CSS

It's easy and intuitive. A lot of people don't even know that it's possible with CSS. Also there is a problem syncing settings between devices, especially on mobiles. So the easier - the better.

> Then we could also make the nicks clickable (and maybe later get the custom highlight list from that same parser?).

It's already done. It's the same tpl as in the user list.

bews:
> bring back the old behavior of dynamically generating colors based on some hash rather than the fixed colors.

I thought about this too, but I think pre-generated colors is better approach, since they can be adapted for dark/bright themes. So there is basically only 1 difference now between that old behavior and current colors: number of colors. It means if we want to make a perfect user color system number of colors should be increased. This is why I decided to make that color generator.

dgw:
I will leave discussion of the number of colors to people with more UX knowledge than I have, but I want to give a huge :+1: for colored nicknames mentioned in chat. I find having that feature in my Textual theme on desktop makes conversations much easier to follow.

bews:
I decided to refactor color overriding and make it in more stylish/awesome/elegant way. Not sure when, though.

astorije:
> Well, I can dissociate ~95% of these 70 colors (superpower, yay)

Lol there is no way this is possible. Just to be clear, I didn't mean that colors were undistinguishable from each other, but at a cognitive level, there is no way one can associate names to 70 different colors in a discussion, in a way that it is meaningful to have.
Sorry if I'm not super clear, but simply worded it's unreasonable that The Lounge needs such an overkill palette, which includes the example of 29 colors I posted above, with very few distinctions. We really don't need this IMO...

> It's already done. It's the same tpl as in the user list.

My bad, I didn't notice that (haven't gone through the code yet).

---

Anyway, this is my opinion on this, based on the arguments listed above: I'm all for 3 if we can make sure of the details, not in favor of 1, and definitely not in favor of 2.
Curious of what other @thelounge/maintainers think 😃

bews:
> which includes the example of 29 colors I posted above, with very few distinctions

In fact, those colors looks better then others on my display. It depends on many factors, I guess.

> simply worded it's unreasonable that The Lounge needs such an overkill palette

As I said in the other comment, I believe that it's better to have a little different colors than the same - just compare current coloring system and the old one in Shout.

> I'm all for 3 if we can make sure of the details, not in favor of 1, and definitely not in favor of 2.

I'll make changes in 1 and 2, so we can discuss this again later.

bews:
So, I reduced colors to 64 and shuffled them. And while I still not fully satisfied with this result (my original plan was to generate ~95 colors), it is definitely better than current ones.

And thus **color overriding** is the only thing left. Will do it later.

astorije:
Well, I'm still 👎 on doubling the number of colors and the color override thingy, so maybe you should wait for other maintainers to chime in :D

bews:
> What I'm fairly concerned about however is the potential amount of false positives coming with this, 

You're right. Added `RegExp("(^| )" + nick + "([.,: ]{1})", "g")`. Will test more.

Also removed color overriding for now, since it needs a better implementation.

bews:
Added brackets to nicknames on mobile, since it hard to read when mentions are highlighted too.

bews:
So, after some tests number of colors was decreased back to 32. Reason is somewhat unexpected.

When I tried to generate new colors, I was thinking about 2 criteria:
- colors should be distinctive
- visible on dark/bright themes

Thereby 70 colors were generated. But after some testing it turned out that there should be one more criterion:
- colors should be good looking and fit current theme

With this a lot of colors received no-no label (too pale, too dark, etc).

And it's obvious that the more criteria you have the less colors it possible to generate. Of course 32 isn't the absolute maximum, but it doesn't worth to increase number of colors for just a few more.

New generator settings can be found here: https://colors.bews.me/test.html

client/js/lounge.js
```diff
@@ -332,12 +332,26 @@ $(function() {
...
+				data.msg.target = data.msg.whois.nick;
+			}
+
 			text.html(templates.actions[type](data.msg));
 		}
 
 		if ((type === "message" || type === "action") && chan.hasClass("channel")) {
 			var nicks = chan.find(".users").data("nicks");
 			if (nicks) {
+				nicks.forEach(function(nick) {
```
xPaw: This is going to be a massive performance problem.

client/views/user_name.tpl
```diff
@@ -0,0 +1,5 @@
+{{#if from}}
```
xPaw: Since you created a username template, you could pass in the string into it, then you won't need three different span generations.

client/js/lounge.js
bews: I don't see any other way to do this, since you need to search in text for each user.

client/views/user_name.tpl
bews: Fixed this.

xPaw:
Split username template into a separate PR please.

client/css/style.css
```diff
@@ -1915,6 +1916,16 @@ kbd {
 		padding: 0;
 	}
 
+	#chat .message .from .user:before {
```
xPaw: This could be a separate PR as well.

bews:
> Split username template into a separate PR please.

Um... This PR require it for mentions.

> This could be a separate PR as well.

This is needed only when mentions are colored, so I added it to this PR.

xPaw:
The way you did colouring regex, it most likely won't be merged. The username template is a simpler cleanup that can be merged quickly.

client/js/lounge.js
```diff
@@ -321,7 +321,7 @@ $(function() {
 			"whois",
 			"ctcp",
 			"channel_list",
-			"ban_list",
+			"ban_list"
```
astorije: @xPaw, do we want/need an ESLint rule to prevent those? (i.e. to keep the trailing commas)

xPaw: Keeping is better for minimal diffs.

astorije: Right, I agree. My question was: do we want to enforce that?

bews:
> The way you did colouring regex

If you have a better idea how to do this - please share. 

Also: https://jsperf.com/test-testmatching/1
I thought .indexof will be faster, so used regex just for filtering, but it seems using regex alone is faster.

xPaw:
Creating a regex for each user in channel for each received message is certainly not going to be performant. You don't need benchmarks to tell that.

bews:
Fixed bug in "show more" event causing #chan-undefined class. Now coloring works in history too.

astorije:
@bews, are you planning on splitting your code into separate PRs like @xPaw requested?

metsjeesus:
Actions like /me and kicking have a problem, they miss a space between spans.

bews:
> @bews, are you planning on splitting your code into separate PRs like @xPaw requested?

No, I don't see any reason to remove this new awesome coloring. It works well even on my old mobile phone, which is very slow, so there no any performance problem.

Also, I don't see any other way to implement this. Even if we deside to move this regexp to server-side (which is better way to implement this IMO), there still be regexp on client side - you can't replace user name with tpl on server-side :smiley_cat: 

If you want to improve Lounge load time, it's better to change `init` rendering: build only active channel, others can be build in the background later. It will easy to implement this after merging #1185. But it has nothing to do with this PR :cat: 

> Actions like /me and kicking have a problem, they miss a space between spans.

This PR doesn't have such problem.

xPaw:
> No, I don't see any reason to remove this new awesome coloring.

I didn't say remove it, I said to split it into a separate PR. As it stands, this PR currently does changes somewhat unrelated to each other. That username template refactor as a standalone PR has a high chance of being merged as soon as possible.

bews:
> I didn't say remove it, I said to split it into a separate PR.

Since this PR require that new template, things I need to do:
- remove coloring code
- test everything
- wait until this PR is merged
- create new PR
- test everything again

And I still don't see any reason why I should spend time on this unproductive actions.

client/css/style.css
astorije: It actually has nothing to do with colors, and because it's shoved in the middle of other changes/commits, I didn't see it until now.

I don't think we want this. I'm not too much in favor, but I would need to play with it to see how it feels / test usability. In the meantime, @bews please remove it or move it to a dedicated PR.

client/css/style.css
```diff
@@ -938,38 +938,39 @@ kbd {
...
-#chat.colored-nicks .user.color-24 { color: #00367d; }
-#chat.colored-nicks .user.color-25 { color: #009ec4; }
-#chat.colored-nicks .user.color-26 { color: #006119; }
-#chat.colored-nicks .user.color-27 { color: #008bb8; }
-#chat.colored-nicks .user.color-28 { color: #469c00; }
-#chat.colored-nicks .user.color-29 { color: #ff0f95; }
-#chat.colored-nicks .user.color-30 { color: #ff7615; }
-#chat.colored-nicks .user.color-31 { color: #ff4846; }
-#chat.colored-nicks .user.color-32 { color: #ff199b; }
+/* Generate different shades for your theme here: https://colors.bews.me/ */
```
astorije: lol wat. Shameless plug in the code 😂

client/js/lounge.js
astorije: @bews, please revert this change.

client/js/options.js
```diff
@@ -67,7 +67,7 @@ settings.on("change", "input, select, textarea", function() {
 		"nick",
 		"part",
 		"quit",
-		"notifyAllMessages",
+		"notifyAllMessages"
```
astorije: @bews, please revert this change.

astorije:
Unfortunately, it seems like we are facing a disagreement here.

@bews, while your PR has a couple straight-up beneficial changes (s/`chan`/`id` and the template refactor) and could be merged right away, there is another one that needs more discussion, work, etc. that the team is not agreeing with right now.
While @xPaw is saying 👍 to the former and 👎 to the latter, it seems like you would rather keep this PR unchanged. It is entirely your right to refuse the change requests, but that unfortunately means we also have to exercise our right to refuse this PR.

In light of this, I will close this PR for now to try to close the loop, though I am sad about it and would have preferred a solution where we all move forward a bit.
Feel free to request a re-open if you are willing to break down this PR.
If not, if @xPaw or anyone wants to cherry-pick this PR to keep only the 👍'ed parts, this should be an easy win :)

metsjeesus:
seems | character will screw the coloring

bews:
> seems | character will screw the coloring if there is a user with a name as |

What coloring? Can you give an example, screen maybe?

metsjeesus:
something like that

https://www.upload.ee/image/7140166/bug.png

bews:
@metsjeesus fixed this: https://github.com/bews/lounge/commit/5b8fe5673951448821fbaaccc9962b0eb9e33704

bews:
fix: https://github.com/bews/lounge/commit/981df5f17c08b4cc8ab851700997a2150e884b27

PR #1100, opened 2017-04-28
Add support for 0x04 hex colors

xPaw:
Ref: https://modern.ircdocs.horse/formatting.html#hex-color

At least IRCCloud and AdiIRC support it.

cc @Bonuspunkt

### DIFF
--- a/client/js/libs/handlebars/ircmessageparser/parseStyle.js
+++ b/client/js/libs/handlebars/ircmessageparser/parseStyle.js
@@ -3,6 +3,7 @@
 // Styling control codes
 const BOLD = "\x02";
 const COLOR = "\x03";
+const HEX_COLOR = "\x04";
 const RESET = "\x0f";
 const REVERSE = "\x16";
 const ITALIC = "\x1d";
@@ -11,6 +12,9 @@ const UNDERLINE = "\x1f";
 // Color code matcher, with format `XX,YY` where both `XX` and `YY` are
 // integers, `XX` is the text color and `YY` is an optional background color.
 const colorRx = /^(\d{1,2})(?:,(\d{1,2}))?/;
+
+// 6-char Hex color code matcher
+const hexColorRx = /^([0-9a-f]{6})(?:,([0-9a-f]{6}))?/i;
 
 // Represents all other control codes that to be ignored/filtered from the text
 const controlCodesRx = /[\u0000-\u001F]/g;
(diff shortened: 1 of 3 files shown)


client/js/libs/handlebars/ircmessageparser/parseStyle.js
```diff
@@ -12,6 +13,9 @@ const UNDERLINE = "\x1f";
 // integers, `XX` is the text color and `YY` is an optional background color.
 const colorRx = /^(\d{1,2})(?:,(\d{1,2}))?/;
 
+// 6-char Hex color code matcher
+const hexColorRx = /^([a-fA-F0-6]{6})?/;
```
Bonuspunkt: pattern seems wrong, should be `/^([0-9a-f]{6})?/i` ?

test/client/js/libs/handlebars/ircmessageparser/parseStyle.js
```diff
@@ -104,12 +109,70 @@ describe("parseStyle", () => {
 		expect(actual).to.deep.equal(expected);
 	});
 
+	it("should parse hex colors", () => {
+		const input = "test \x04FFFFFFnice \x02\x04RES006 \x0303,04\x04ff00FFcolored";
```
Bonuspunkt: is hex formatting foreground only?
also more color variation would be nice (see issue with pattern)

client/js/libs/handlebars/ircmessageparser/parseStyle.js
xPaw: Yeah that `0-6` is wrong.

test/client/js/libs/handlebars/ircmessageparser/parseStyle.js
xPaw: Yes, only foreground.

DanielOaks: Hmm, AdiIRC seems to allow foreground and background. I should make it more clear on the doc that it allows setting both foreground and background colours with hex in this way

astorije: @xPaw, do you want/need to update your PR to account for foreground/background as per @DanielOaks' comment or not?

xPaw: This currently does not support background colors. Considering @DanielOaks didn't know that too, I wonder how many clients implement it, compared to just the foreground color.

astorije: Right, it doesn't. But don't we want to be the first IRC client leading the way forward? :D (Ok ok it's a minor lead 😅)

xPaw: Considering there is no formal definition, we want to know how it's already implemented in other clients.

astorije: Fair enough?

xPaw:
Added support for background colors.

PR #1104, opened 2017-04-28
Fix away message disappearing

### LINKED ISSUE #1102: awayMessage setting cleared on start
After setting either a network or user-wide `awayMessage` in the user config file and starting lounge, the string is cleared and set back to `""`.

xPaw:
Closes #1102

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -66,8 +66,9 @@ function Client(manager, name, config) {
 	if (typeof config !== "object") {
 		config = {};
 	}
+
 	_.merge(this, {
-		awayMessage: "",
+		awayMessage: config.awayMessage || "",
 		lastActiveChannel: -1,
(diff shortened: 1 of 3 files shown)


src/client.js
```diff
@@ -494,6 +493,8 @@ Client.prototype.clientAttach = function(socketId) {
 		});
 	}
 
+	client.attachedClients[socketId] = client.lastActiveChannel;
+
```
astorije: > 18:13	**astorije**	Just curious, why did you move this?
> 18:13	**xPaw**	look at line 487
> 18:14	**xPaw**	it added to array, then compared to see if size of array is 0, which it would be 1
> 18:14	**xPaw**	I could change the comparison to 1, or move assignment

PR #1106, opened 2017-04-28
Add a default quit message

xPaw:
Quite many IRC clients self-promote this, I think we could benefit from this too.

This could be a config option if we really wanted to.


astorije:
Why did you close this, @xPaw?

xPaw:
I realized `/quit` is a separately implemented command. Will fix another time.

Issue #1105, opened 2017-04-28
Autocomplete dropdown for colors


astorije:
Whenever hitting <kbd>cmd</kbd>+<kbd>K</kbd>, a dropdown should open with available colors.

Issue #1086, opened 2017-04-24
Use fuzzy-matching for nick/channel/emoji/command autocompletion


astorije:
*This will only be relevant once #787 and #856 are merged.*

Essentially apply a similar fuzzy-matching mechanism than user list to the completion search. I don't expect this to be super trivial...

@yashsriv, will you be interested into looking into this?

yashsriv:
Yup. I'll have a look at it in a few days!

PR #1109, opened 2017-04-28
Add autocomplete strategy for foreground and background colors

astorije:
Addresses #1105

~~One limitation of this is it only works for foreground color. Making it support background color as well is not impossible, but it's more efforts and I can't spend them on that right now. I still think the current PR adds some value anyway!~~ Well, not anymore it doesn't.

Color labels were taken from https://modern.ircdocs.horse/formatting.html#colors, which we link to from the client.

autocomplete-colors2

### DIFF
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -1,5 +1,24 @@
 "use strict";
 
+const colorCodeMap = [
+	["00", "White"],
+	["01", "Black"],
+	["02", "Blue"],
+	["03", "Green"],
+	["04", "Red"],
+	["05", "Brown"],
+	["06", "Magenta"],
+	["07", "Orange"],
+	["08", "Yellow"],
+	["09", "Light Green"],
+	["10", "Cyan"],
+	["11", "Light Cyan"],
+	["12", "Light Blue"],
+	["13", "Pink"],
+	["14", "Grey"],
+	["15", "Light Grey"],
+];
+
 const commands = [
 	"/away",
 	"/back",
@@ -35,5 +54,6 @@ const commands = [
 ];
 
 module.exports = {
+	colorCodeMap: colorCodeMap,
 	commands: commands
 };
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -111,6 +111,25 @@ $(function() {
 		index: 1
 	};
 
+	const foregroundColorStrategy = {
+		id: "foreground-colors",
+		match: /\x03(\d{0,2}|[A-Za-z ]{0,10})$/,
+		search(term, callback) {
+			term = term.toLowerCase();
+			const matchingColorCodes = constants.colorCodeMap
+				.filter(i => i[0].startsWith(term) || i[1].toLowerCase().startsWith(term));
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${parseInt(value[0], 10)}">${value[1]}</span>`;
+		},
+		replace(value) {
+			return "\x03" + value[0];
+		},
+		index: 1
+	};
+
 	socket.on("auth", function(data) {
 		var login = $("#sign-in");
 		var token;
@@ -724,7 +743,7 @@ $(function() {
 			chat.find(".chan.active .chat").trigger("msg.sticky"); // fix growing
 		})
 		.tab(completeNicks, {hint: false})
-		.textcomplete([emojiStrategy, nicksStrategy, chanStrategy, commandStrategy], {
+		.textcomplete([emojiStrategy, nicksStrategy, chanStrategy, commandStrategy, colorStrategy], {
 			dropdownClassName: "textcomplete-menu",
 			placement: "top"
 		}).on({
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -130,6 +130,26 @@ $(function() {
 		index: 1
 	};
 
+	const backgroundColorStrategy = {
+		id: "background-colors",
+		match: /\x03(\d{2}),(\d{0,2}|[A-Za-z ]{0,10})$/,
+		search(term, callback, match) {
+			term = term.toLowerCase();
+			const matchingColorCodes = constants.colorCodeMap
+				.filter(i => i[0].startsWith(term) || i[1].toLowerCase().startsWith(term))
+				.map(pair => pair.concat(match[1])); // Needed to pass fg color to `template`...
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${parseInt(value[2], 10)} irc-bg irc-bg${parseInt(value[0], 10)}">${value[1]}</span>`;
+		},
+		replace(value) {
+			return "\x03$1," + value[0];
+		},
+		index: 2
+	};
+
 	socket.on("auth", function(data) {
 		var login = $("#sign-in");
 		var token;
@@ -743,7 +763,10 @@ $(function() {
 			chat.find(".chan.active .chat").trigger("msg.sticky"); // fix growing
 		})
 		.tab(completeNicks, {hint: false})
-		.textcomplete([emojiStrategy, nicksStrategy, chanStrategy, commandStrategy, colorStrategy], {
+		.textcomplete([
+			emojiStrategy, nicksStrategy, chanStrategy, commandStrategy,
+			foregroundColorStrategy, backgroundColorStrategy
+		], {
 			dropdownClassName: "textcomplete-menu",
 			placement: "top"
 		}).on({
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1533,6 +1533,10 @@ kbd {
 	display: inline-block;
 }
 
+.textcomplete-item .irc-bg {
+	display: block;
+}
+
 /**
   * Tooltips v0.5.3
   * See http://primercss.io/tooltips/


client/js/lounge.js
```diff
@@ -111,6 +111,47 @@ $(function() {
...
+		match: /^\x03(\d{0,2}|[A-Za-z ]{0,10})$/,
+		search(term, callback) {
+			term = term.toLowerCase();
+			const matchingColorCodes = colorCodeMap
+				.filter(i => i[0].toString().startsWith(term) || i[1].toLowerCase().startsWith(term));
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			if (value === colorCodeMap[0]) {
```
xPaw: This is kinda theme dependant, probably better to handle in css.

xPaw:
What about autocompleting background colors after you type a comma?

client/js/lounge.js
astorije: Agreed, but I couldn't find a good way to do that simply without having to duplicate the entire `.irc-(f|b)g*` CSS palette, which I really don't want to do.
An alternative is to... not care, and first line is going to be white text on white bg + non-white hover.
Any suggestions?

astorije:
> What about autocompleting background colors after you type a comma?

I really didn't want to spend one more minute on that to go back to my original tasks, but I guess it's my fault for starting this. Let me try this and if it's more than 5 minutes, I'll give up.

client/js/lounge.js
xPaw: `.textautocomplete-item .irc-fg1 { background: #000; }` something like that.

astorije: Duh, of course 😊
But I actually went for white on default background, for several reasons: white on black looks gross in terms of UI, I reuse the text color for the background color dropdown text, the hovering background shows text alright.

Only thing is hovering background is veeeery dim, we really need to change it someday :D

astorije:
Alright alright, I added strategy for background colors. Please don't make me work on this again :D

astorije:
Oh also, updated the screenshot in PR description to show background colors dropdown as well.

client/js/lounge.js
```diff
@@ -111,6 +111,64 @@ $(function() {
...
+		[11, "Light Cyan"],
+		[12, "Light Blue"],
+		[13, "Pink"],
+		[14, "Grey"],
+		[15, "Light Grey"],
+	];
+
+	const foregroundColorStrategy = {
+		id: "foreground-colors",
+		match: /\B\x03(\d{0,2}|[A-Za-z ]{0,10})$/,
```
xPaw: Remove `\B`. You should be able to autocomplete colors no matter where it is.

client/js/lounge.js
```diff
@@ -111,6 +111,64 @@ $(function() {
...
+				.filter(i => i[0].toString().startsWith(term) || i[1].toLowerCase().startsWith(term))
+				.map(pair => pair.concat(match[1])); // Needed to pass fg color to `template`...
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${value[2]} irc-bg${value[0]}">${value[1]}</span>`;
+		},
+		replace(value) {
+			return "\x03$1," + value[0];
```
xPaw: You should add zero padding for <10 colors, so that `1` turns into `01`.

client/js/lounge.js
```diff
@@ -111,6 +111,64 @@ $(function() {
...
+		search(term, callback, match) {
+			term = term.toLowerCase();
+			const matchingColorCodes = colorCodeMap
+				.filter(i => i[0].toString().startsWith(term) || i[1].toLowerCase().startsWith(term))
+				.map(pair => pair.concat(match[1])); // Needed to pass fg color to `template`...
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${value[2]} irc-bg${value[0]}">${value[1]}</span>`;
```
xPaw: I think the color number should be displayed too. Perhaps the same way emojis are displayed.

client/js/lounge.js
```diff
@@ -111,6 +111,64 @@ $(function() {
...
+			const matchingColorCodes = colorCodeMap
+				.filter(i => i[0].toString().startsWith(term) || i[1].toLowerCase().startsWith(term));
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${value[0]}">${value[1]}</span>`;
+		},
+		replace(value) {
+			return "\x03" + value[0];
```
xPaw: Same here, zero padding.

client/js/lounge.js
```diff
@@ -111,6 +111,64 @@ $(function() {
...
+		},
+		index: 1
+	};
+
+	const backgroundColorStrategy = {
+		id: "background-colors",
+		match: /\B\x03(\d{1,2}),(\d{0,2}|[A-Za-z ]{0,10})$/,
```
xPaw: Same here, remove `\B`.

client/js/lounge.js
```diff
@@ -111,6 +111,64 @@ $(function() {
 		index: 1
 	};
 
+	const colorCodeMap = [
```
xPaw: Move to `constants.js`

astorije:
I hesitated with this:

screen shot 2017-04-30 at 12 49 29

It reflects the reality less but it does look nicer. Thoughts?

xPaw:
@astorije It would look nicer if the entire item/line was colored.

client/js/lounge.js
astorije: I didn't work without `\B`. Feel free to play with it, but I'll leave it as is, I've spent enough time on these regexes I hate 😅.

xPaw: @yashsriv?

client/js/lounge.js
astorije: I sure can, but do we need to, knowing that it works fine as is? Asking because I don't know the IRC details, you probably know a reason I don't.

client/js/lounge.js
astorije: I don't think so, it doesn't add value to the autocomplete. If you know the code, you won't rely on the dropdown. It mainly is for user-friendliness.

client/js/lounge.js
xPaw: https://modern.ircdocs.horse/formatting.html#mistaken-eating-of-text

client/js/lounge.js
xPaw: I think it would help to autocomplete colors faster, as you would see the color number and can type it in, without guessing.

client/js/lounge.js
astorije: Ok, it looks like it's working now. I probably messed something with my regexes earlier, nevermind...

astorije: Done

client/js/lounge.js
astorije: Done

client/js/lounge.js
astorije: Done

astorije:
> What if you included foreground color name? Like Green on Black.

It looks like this:

screen shot 2017-04-30 at 13 30 19 screen shot 2017-04-30 at 13 30 30

(It says "Blue" simply because I hardcoded it for the screenshot)

I'm not sure. What do you think?

client/js/lounge.js
astorije: Done

client/js/lounge.js
astorije: Done

astorije:
Alright, really hoping I'm done with this 🙏
I have to move on with life now 😅

client/js/constants.js
```diff
@@ -1,5 +1,24 @@
 "use strict";
 
+const colorCodeMap = [
+	[0, "White"],
```
xPaw: I just realized you don't need numbers here. This could simply be an array of color names.

astorije: I kinda do for [this](https://github.com/thelounge/lounge/blob/astorije/autocomplete-colors/client/js/lounge.js#L120-L125).
And actually, I'm going to change the first value from a regular int to `"01"`, etc.

xPaw: `.filter` gives you index and value in array, so it should work just fine.

astorije: Ok did that, it's much cleaner like this and puts the mapping of color codes/names as a source of truth in the constants.
I could also reconstruct the codes from index + leading zero + `toString` but it feels hacky. Better to keep the codes here, and also makes it possible to support future stuff like `"99"`.

astorije: Please tell me I'm done lol, I so should have never done this in the first place 😅 (anyway, I'm out so I won't be able to work on this more before a day or 2 so that's my cue)

astorije: > .filter gives you index and value in array, so it should work just fine.

Current (new) solution is cleaner IMO.

xPaw:
@astorije Well, now to get a suggestion for <10 colors, you have type like `05` as typing just `5` will not show a suggestion.

astorije:
> @astorije Well, now to get a suggestion for <10 colors, you have type like 05 as typing just 5 will not show a suggestion.

Yes, that will be addressed with fuzzy matching on completion (see #1086).

Please no more bikeshedding. Close this PR if you don't like it but I can't spend another day on this 😅

xPaw:
Don't worry, I'll merge this PR. Also fuzzy matching for colors is pretty lol.

client/js/lounge.js
```diff
@@ -111,6 +111,45 @@ $(function() {
...
+				.filter(i => i[0].startsWith(term) || i[1].toLowerCase().startsWith(term))
+				.map(pair => pair.concat(match[1])); // Needed to pass fg color to `template`...
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${value[2]} irc-bg irc-bg${parseInt(value[0], 10)}">${value[1]}</span>`;
+		},
+		replace(value) {
+			return "\x03$1," + ("00" + value[0]).slice(-2);
```
xPaw: Forgot to remove padding.

client/js/lounge.js
```diff
@@ -111,6 +111,45 @@ $(function() {
...
+		search(term, callback, match) {
+			term = term.toLowerCase();
+			const matchingColorCodes = constants.colorCodeMap
+				.filter(i => i[0].startsWith(term) || i[1].toLowerCase().startsWith(term))
+				.map(pair => pair.concat(match[1])); // Needed to pass fg color to `template`...
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${value[2]} irc-bg irc-bg${parseInt(value[0], 10)}">${value[1]}</span>`;
```
xPaw: Do you need to `parseInt` on `value[2]`?

astorije:
Not so much actually.  Fuzzy matching at the whole completion mechanism
level would let you match "Blue" and "Light Blue" when you type "blue" :)

astorije:
Dammit. Will remove when I have a second on laptop.

astorije:
I do, good catch.

This solution feels like a crappy hack though. Would you be ok if I add the
classes  in the CSS file like `.irc-fg03, .irc-fg3 { color: ... }`? Maybe
later we can remove the versions without leading zeroes (after converting
places that use them) but we could also keep them both permanently for
convenience.

xPaw:
> `.irc-fg03, .irc-fg3 { color: ... }`?

That would only look like code smell to me.

astorije:
Why?!

client/js/lounge.js
astorije: Done.

client/js/lounge.js
astorije: Done. But don't you think we should be using `irc-fg04` and the likes instead of `irc-fg4`, to better reflect color code format in general?
That would remove need for `parseInt` here, but also I think it's nicer to match 1:1 with actual codes.
Not for this PR though, as they are currently used in a couple places I believe (which is why I wanted to use both until transition is over at least).

client/js/lounge.js
```diff
@@ -111,6 +111,45 @@ $(function() {
...
+		search(term, callback, match) {
+			term = term.toLowerCase();
+			const matchingColorCodes = constants.colorCodeMap
+				.filter(i => i[0].startsWith(term) || i[1].toLowerCase().startsWith(term))
+				.map(pair => pair.concat(match[1])); // Needed to pass fg color to `template`...
+
+			callback(matchingColorCodes);
+		},
+		template(value) {
+			return `<span class="irc-fg${parseInt(value[2], 10)} irc-bg irc-bg${parseInt(value[0], 10)}">${value[1]}</span>`;
```
xPaw: Is that an erroneous extra `irc-bg`?

Issue #98, opened 2016-02-24
Timestamp Show Seconds


vectr0n:
It would be really handy to be able to see the seconds in the time stamps.

TheSonAlsoRises:
Agreed. Along with #89, I would say the two would make the messages much more legible.

YaManicKill:
We had 2 different PRs trying to fix this, and neither ended up being agreed, so we should probably agree on how we want this.

There were 2 approaches:

1. Just having a setting to show or hide seconds
2. Having an input to choose a time format

If anyone has any other ideas, share them now. For me, I guess 1 makes most sense. @astorije pointed out that that would be the absolute most common reason to want to change the timestamp, and that having a text input for time format would be confusing.

vectr0n:
I think it would be easier to just have a show seconds or hide seconds option, or maybe have the option for a custom time format in the server config?

Dr-Horv:
1. seems much easier and less confusing for the end user. Should cover most people needs. Could reopen this issue if anyone desires anything else.

hho:
Can you please add the date as well?

dgw:
And this is why 2. was proposed…because everyone wants slightly different information in their timestamps.

hho:
Maybe there's some middle ground, like having a dropdown with different formats? That'd be more user friendly (though not as flexible). Or output complete timestamps and have CSS classes for date, second, etc. so that themes can hide the different parts.

YaManicKill:
Alright I suggest that we have a setting to enable seconds, but we have a hovertext on the time to show the full date time.

That coupled with a date seperator on the page should be enough to satisfy everyone?

dgw:
Setting for seconds and hovertext with the full timestamp…sounds good to me. ISO8601 style for the hovertext?

YaManicKill:
> ISO8601 style for the hovertext?

dgw:
That'd be a hell of a long tooltip. Twitter's hoverstamps (is that a good name for them?) read e.g. "8:21 AM - 27 May 2016", nice and concise. Why include the timezone when the UTC date can be passed to the client and converted to the useragent's local timezone for display?

DanielOaks:
I like to get my instance of lounge to hook up to my ZNC box. Since messages can sometimes come from days or weeks back, I find having dates on the messages really useful as well (at least month/day, kinda like `09/23 12:38:04`).

[...]

PR #1114, opened 2017-04-29
Use moment to render dates everywhere

xPaw:
Also changes dates to be rendered like `29 April 2017`.

### DIFF
--- a/client/js/libs/handlebars/friendlydate.js
+++ b/client/js/libs/handlebars/friendlydate.js
@@ -4,10 +4,10 @@ const moment = require("moment");
 
 module.exports = function(time) {
 	// See http://momentjs.com/docs/#/displaying/calendar-time/
-	return moment(new Date(time)).calendar(null, {
+	return moment(time).calendar(null, {
 		sameDay: "[Today]",
 		lastDay: "[Yesterday]",
-		lastWeek: "L", // Locale
-		sameElse: "L"
+		lastWeek: "D MMMM YYYY",
+		sameElse: "D MMMM YYYY"
 	});
 };
--- a/client/js/libs/handlebars/localetime.js
+++ b/client/js/libs/handlebars/localetime.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const moment = require("moment");
+
 module.exports = function(time) {
-	return new Date(time).toLocaleString();
+	return moment(time).format("D MMMM YYYY, HH:mm:ss");
 };
--- a/client/js/libs/handlebars/tz.js
+++ b/client/js/libs/handlebars/tz.js
@@ -1,17 +1,7 @@
 "use strict";
 
+const moment = require("moment");
+
 module.exports = function(time) {
-	time = new Date(time);
-	var h = time.getHours();
-	var m = time.getMinutes();
-
-	if (h < 10) {
-		h = "0" + h;
-	}
-
-	if (m < 10) {
-		m = "0" + m;
-	}
-
-	return h + ":" + m;
+	return moment(time).format("HH:mm");
 };
--- a/client/js/libs/handlebars/localedate.js
+++ b/client/js/libs/handlebars/localedate.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const moment = require("moment");
+
 module.exports = function(time) {
-	return new Date(time).toLocaleDateString();
+	return moment(time).format("D MMMM YYYY");
 };
--- a/test/client/js/libs/handlebars/friendlydateTest.js
+++ b/test/client/js/libs/handlebars/friendlydateTest.js
@@ -18,7 +18,7 @@ describe("friendlydate Handlebars helper", () => {
 	it("should not render any friendly dates prior to the day before", () => {
 		[2, 7, 30, 365, 1000].forEach(day => {
 			const time = new Date().getTime() - 24 * 3600 * 1000 * day;
-			expect(friendlydate(time)).to.equal(moment(time).format("L"));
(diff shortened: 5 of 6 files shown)


client/js/libs/handlebars/friendlydate.js
```diff
@@ -4,10 +4,10 @@ const moment = require("moment");
...
 module.exports = function(time) {
 	// See http://momentjs.com/docs/#/displaying/calendar-time/
-	return moment(new Date(time)).calendar(null, {
+	return moment(time).calendar(null, {
 		sameDay: "[Today]",
 		lastDay: "[Yesterday]",
-		lastWeek: "L", // Locale
-		sameElse: "L"
+		lastWeek: "MMMM DD, YYYY",
+		sameElse: "MMMM DD, YYYY"
```
astorije: I don't think this is a win. That way of displaying dates is only common in very few countries, so relying on locales is safer (unless I'm missing something).

I think this should be dealt with when internationalizing the client, and left alone until then.

client/js/libs/handlebars/localetime.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
+const moment = require("moment");
+
 module.exports = function(time) {
-	return new Date(time).toLocaleString();
+	return moment(time).format("MMMM DD, YYYY – HH:mm:ss");
```
astorije: Similar comment here and above. While the previous change (`L` to `MMMM DD, YYYY`) forces an American standard, this one will change from `4:30:00 AM` to `16:30:00`, actually parting from the American standard. So even if we forget about internationalization and go for an opinionated change, there is a loss of consistency in the standard used.
(Though currently tooltip show `4:30:00 AM` but message time shows `16:30`... Ah, if only we had proper localization :D)

client/js/libs/handlebars/tz.js
```diff
@@ -1,17 +1,7 @@
...
-	}
+const moment = require("moment");
 
-	if (m < 10) {
-		m = "0" + m;
-	}
-
-	return h + ":" + m;
+module.exports = function(time) {
+	return moment(time).format("HH:mm");
```
astorije: Nice!

client/js/libs/handlebars/localetime.js
xPaw: Exactly, messages already use 24 hour format, and so will this.

client/js/libs/handlebars/friendlydate.js
xPaw: I think locale formatting is not that good, especially when we can't specify that we want the longer representation. The format I chose here is *not* american, but rather a sane approach to formatting dates, where full month name is displayed. This removes any confusion that there can be for "is it month or day?"

client/js/libs/handlebars/tz.js
bews: Need this for #98 implementation to avoid conflicts. Hope it will be merged soon.

astorije: As long as it's a "Show seconds" checkbox and not a time format text input, then it should be fine 😄

client/js/libs/handlebars/friendlydate.js
astorije: Ah ah, calling me uncultured much? :D

I would be arguing if the entire app was consistent, but since we are not even consistent with one format (I thought we were, with the American one, but didn't realize we had `16:30`, I'm just too vested in this format to see it!), then not so much.

I do think we'll need to bring back this discussion when we localize the whole client, but until then, I'm convinced (especially with the conversion from `02-04` into `2 May`. Or is it `4 February`? :D)

However, since we're here, let's go for `DD MMMM YYYY`. It feels more balanced (no need for a comma to visually separate day and year), it represents 10 times more populations than the alternative according to your link, etc. (it's also the standard date format for W3C standards :D)

client/js/libs/handlebars/localetime.js
astorije: Yeah, my comment in parens betrayed me. I never really noticed that 12/24h format difference before today.

Same comment as before though, I'd advocate for `DD MMMM YYYY, HH:mm:ss`.

client/js/libs/handlebars/tz.js
bews: It's a "Show seconds" checkbox, don't worry :laughing:

client/js/libs/handlebars/friendlydate.js
dgw: > (especially with the conversion from `02-04` into `2 May`. Or is it `4 February`? :D)

It'd be `2 April` anyway :D

astorije: Goddammit

xPaw: Can we keep the comma? `DD MMMM, YYYY`

astorije:
"1 May, 2017" feels odd, I don't recall seeing this anywhere. "1 May 2017"
looks more balanced and is nicer when adding the time as we need only one
(or zero) separator.
Nitpick, granted, but I do think it's more standard-looking :)

client/js/libs/handlebars/friendlydate.js
astorije: Looking at https://en.wikipedia.org/wiki/Date_format_by_country and others, it looks like `1 May 2017` is the most common, `May 1, 2017` is essentially US-only, but `1 May, 2017` exists virtually nowhere.

Do you mind if we go for the first of the 3 versions?

PR #1115, opened 2017-04-29
Fix channel sorting messing up the order

### LINKED ISSUE #1088: Channel sorting broken
Running on master (dd5e50f).

As you can see the order is messed up when syncing to other clients, but the strange order also persists when refreshing. (Also note how the network lobby isn't where it should be)

https://gfycat.com/TalkativeKindlyAlbino

PolarizedIons:
Fixes #1088.

How it sorted after PR #1064:

After fix:


PolarizedIons:
Still not sure why only some people could reproduce it though

PR #1117, opened 2017-04-29
More transition effects

bews:
This adds transition effects for almost every element with hover support. 
Also transition + hover effects were added to a lot of buttons:

In this menu background change covers the whole block now:

Tooltipped animation-delay was decreased 0.4 -> 0.1 to match new transition effects.

-----------------------
TODO:

- [x]  Make this PR compatible with #1130

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -581,6 +581,11 @@ kbd {
 	display: inline-block;
 	line-height: 34px;
 	padding: 0 12px;
+	transition: color .2s;
+}
+
+#footer .icon:hover {
+	color: #fff;
 }
 
 .signed-out #footer .sign-in {
@@ -926,6 +931,17 @@ kbd {
 
 #chat a {
 	color: #50a656;
+}
+
+#chat button,
+#form button,
+#chat .user {
+	transition: opacity .2s;
+}
+
+#chat button:hover,
+#form button:hover {
+	opacity: .6;
 }
 
 /* Nicknames */
@@ -1461,10 +1477,6 @@ kbd {
 	flex: 0 0 auto;
 }
 
-#form #submit:hover {
-	opacity: .6;
-}
-
 #context-menu-container {
 	display: none;
 	position: absolute;
@@ -1492,7 +1504,6 @@ kbd {
 
 .context-menu-divider {
 	height: 1px;
-	margin: 6px 0;
 	background-color: rgba(0, 0, 0, .1);
 }
 
@@ -1500,10 +1511,9 @@ kbd {
 .textcomplete-item {
 	cursor: pointer;
 	display: block;
-	padding: 4px 8px;
+	padding: 10px 8px;
 	color: #333;
-	margin-top: 6px;
-	margin-bottom: 6px;
+	transition: background-color .2s;
 }
 
 .context-menu-item:hover,
@@ -1614,8 +1624,8 @@ kbd {
 	animation-fill-mode: forwards;
 	-webkit-animation-timing-function: ease-in;
 	animation-timing-function: ease-in;
-	-webkit-animation-delay: .4s;
-	animation-delay: .4s;
+	-webkit-animation-delay: .1s;
+	animation-delay: .1s;
 }
 
 .tooltipped-s:after,


client/css/style.css
```diff
@@ -1614,8 +1624,8 @@ kbd {
 	animation-fill-mode: forwards;
 	-webkit-animation-timing-function: ease-in;
 	animation-timing-function: ease-in;
-	-webkit-animation-delay: .4s;
-	animation-delay: .4s;
+	-webkit-animation-delay: .1s;
+	animation-delay: .1s;
```
astorije: Could you revert this please? We actually re-added the animation delay some time ago on purpose to avoid fidgeting on timestamps. We'd be better off with default values.

client/css/style.css
```diff
@@ -1492,18 +1504,16 @@ kbd {
...
 
 .context-menu-item,
 .textcomplete-item {
 	cursor: pointer;
 	display: block;
-	padding: 4px 8px;
+	padding: 10px 8px;
 	color: #333;
-	margin-top: 6px;
-	margin-bottom: 6px;
```
astorije: Actually, this has a tiny issue, in which it adds extra margin around elements within the same "group":

Before your PR | After your PR
--- | ---
screen shot 2017-06-12 at 01 36 33 | screen shot 2017-06-12 at 01 37 04

Mind fixing this please?

client/css/style.css
bews: As I mentioned before, it was done to match new transition effects. There are cases when these 2 animations happen simultaneously (settings, help, toggle user list, etc) , and having different animation duration makes them look bad.

client/css/style.css
bews: Um.. I didn't expect any groups here. This can be fixed only by returning to the old design. Are those groups a planned feature or I'm missing smth?

astorije: Correct, the separator (`divider` in our HTML) is not meant to be present between all rows. The nearest and easiest addition to these dropdowns I can think of is having multiple user actions: kick, ban, etc. So rows would be: nick / divider / ban / kick / ...

client/css/style.css
astorije: Yes, I read your PR description 😅

However, your changes are not on an animation _duration_ but on an animation _delay_. There is a reason we have a 0.4s delay on tooltips, it's to avoid the flickering when moving the mouse over by chance.

We need to keep this at `.4s` and I won't be able to 👍 until they're reverted. Would you mind doing so? 😃

bews: We can keep `.4` for timestamps only.

astorije: No, we cannot 😅

We would really appreciate if you fix this so we can 👍 and merge this. Thanks! :)

bews: > No, we cannot

Why?

> if you fix this

You're asking to break this, not fix.

astorije: > Why?

To keep all our tooltips consistent, and this is not negotiable 😅

> You're asking to break this, not fix.

*Sigh*
No, I'm not. How is keeping reasonable tooltip *delays* (again, **not** animation durations) breaking anything?
I may very well be missing something, but please explain me *why* it is needed to reduce the tooltip apparition **delay**? I honestly have no idea how the animation delay could collide with any other animation out there, but if it did, `.1s` would be more subject to do that than `.4s`.

Anyway, I really don't think there is anything to break here, so unless I'm missing something obvious, please change them back.

bews: `.1`: 

`.4`:

astorije: Yes, and again there are good reasons why we added a delay.

Anyway, I feel like we are going in circles here. This is my final review for this PR: **please revert the animation delay change and you should get my 👍.** Thanks!

astorije:
@bews, are you okay reverting the animation delay change as requested in the review?

This PR is simple and 🎉 outside of that minor change needed, it should make it to `master` very quickly.

bews:
> are you okay reverting the animation delay change

No, I'm not, it's a crucial thing here. Also, this PR is incomplete without #1130 (since I removed some parts from here), therefore it's not ready anymore.

Anyway, it might be better to leave these things to themes, so I close this.

astorije:
That's frustrating, it was almost perfect (and some more time wasted...).

bews:
I wasted more time on these PRs than you, so things are not so bad :smile_cat: 
All of this was supposed to be plugins from the beginning, but I was inspired by another project and made PRs instead. From now on I switch back to the plugins plan, so no wasting time anymore :smiley_cat:

astorije:
*The Good Parts* of this has been ported to #1314. It literally took me 2 minutes :)

Issue #1048, opened 2017-04-17
[Request] log file with failed login attempts


notaduck:
It would be nice if we could get a log file where all failed and successful login attempts where shown.  
It would make it easier to locate where the intruder came from if the system gets compromised, and also make it easier to prevent it from happen, and also allowing us users to sue fail2ban or tools like that.

bews:
It will be better to be able to exec any cmd you like when invalid password is entered. For example I prefer to send info to gitter instead of logging.

notaduck:
That would also be nice. But in my opinion the best thing will be to do both, then you can use fail2ban to automaticly block the intrusion attempts, and still get it notifications through gitter.

astorije:
I'm going to close this because we just merged #2247 which should provide this now.

PR #1118, opened 2017-04-30
Run any command when invalid password was entered

bews:
Very useful to detect and stop bruteforce attacks.

Related issue: #1048

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -370,6 +370,16 @@ module.exports = {
 		primaryKey: "uid"
 	},
 
+	// Run command when invalid password was entered.
+	// * {{user}} will be replaced with the login name;
+	// * {{IP}} - client IP;
+	// Use it to track bad login attempts.
+	//
+	// @type     string
+	// @default  ""
+	//
+	authErrorCmd: "",
+
 	// Extra debugging
 	//
 	// @type     object
@@ -388,6 +398,6 @@ module.exports = {
 		// @type     boolean
 		// @default  false
 		//
-		raw: false,
+		raw: false
 	},
 };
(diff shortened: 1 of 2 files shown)


xPaw:
This looks like a security disaster.

bews:
> This looks like a security disaster.

You're right, completely forgot about some things :sob: 
Check it now.

astorije:
Shouldn't we be emitting auth logs like other tools (ssh, etc.) and then monitoring, fail2ban, others, rely on output of this, instead of running a command from our program?

astorije:
I get the intent, but lots of these would be too much of a privacy
intrusion for The Lounge in my opinion.
Logging all abnormal situations like failed passwords, that yes.

bews:
> login, password change

Not sure about this. Definitely not in the current event -> cmd implementation.


> Shouldn't we be emitting auth logs like other tools (ssh, etc.) and then monitoring

Logs will add more steps before any action + daemon is needed:
```
event
  ↓
logs
  ↑
daemon
  ↓
action
```
Current scheme:
```
event
  ↓
action
```
If you really need logs, then use this cmd to generate them in any format you want.

Current way has more flexibility, so I do not think limiting output to logs only is a good idea. For example, this feature can be used to report strange activity to slack/discord/gitter/ifttt and get notifications right away (without any daemons).

xPaw:
I think base code should have logging, but not this. This is better suited as a plugin (whenever we get these!).

astorije:
Fully agreed with @xPaw.
Most software I know use that logging mechanism, including Nginx and sshd. I am not aware of mechanisms where these 2 would run arbitrary commands.

Closing for reasons detailed above. We can reopen this if other @thelounge/maintainers feel strongly about the opposite.

PR #1120, opened 2017-04-30
Include trickery to reduce paints and improve performance

### LINKED ISSUE #1071: Text Rendering

xPaw:
Fixes #1083.
Fixes #1071.

I tested this with #1041 and it showed lower CPU usage.

https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas
https://developers.google.com/web/updates/2016/06/passive-event-listeners

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -140,10 +140,6 @@ kbd {
 .container {
 	margin: 80px auto;
 	max-width: 480px;
-	overflow: hidden;
-	overflow-y: auto;
-	-webkit-overflow-scrolling: touch;
-	padding: 0 30px;
 }
 
 ::-moz-placeholder {
@@ -374,8 +370,7 @@ kbd {
 }
 
 #viewport.rt #chat .sidebar {
-	-webkit-transform: translate3d(180px, 0, 0);
-	transform: translate3d(180px, 0, 0);
+	right: -180px;
 }
 
 #sidebar {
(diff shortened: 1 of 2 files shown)


astorije:
@xPaw, is this ready for review?

xPaw:
@astorije should be.

astorije:
(Simply rebased on `master`, no code change at all)

client/css/style.css
```diff
@@ -374,8 +370,7 @@ kbd {
 }
 
 #viewport.rt #chat .sidebar {
-	-webkit-transform: translate3d(180px, 0, 0);
-	transform: translate3d(180px, 0, 0);
+	right: -180px;
```
astorije: From the links you posted in PR description, I understand that `right` would need a repaint as opposed to `transform`. Is that what `will-change: transform;` below is trying to prevent?
Also, how does `will-change: transform;` work if you removed the `transform` property above?

Sorry, still trying to understand how all of this works...

xPaw: You are correct that it would be doing a repaint, however this was needed to fix the font rendering issue.

astorije:
@YaManicKill, what do you think, merge this before or after #1175? :/

YaManicKill:
Before is fine, afaics it doesn't edit any of the code I have edited.

Issue #1056, opened 2017-04-19
Serve client website from a public folder


xPaw:
We currently do it from a client folder, which contains uncompiled files, and compiled ones get mixed in. Having a separate public/ folder would simplify some things.

PR #1124, opened 2017-05-01
Implement push notifications

xPaw:
No configuration required (no GCM/FCM setup) at all. Uses latest available tech.

Needs better UI for subscribing and unsubscribing. Currently enabling notification checkbox does the job.

- [x] Implement unsubscribing
- [x] Test in Firefox
- [x] Better UI for (un)subscribing
- [ ] Deal with browser and push notifications (desktop can have both enabled)
- [x] Remove dead subscriptions (e.g. when push endpoint returns `not registered`)
- [x] Currently the code limits to 5 notifications per minute (should tags be used, etc?)
- [x] Currently notifications are only emitted when there are no clients attached, what logic should actually be used?
- [ ] Clicking on notification should focus channel where it came from
- [x] Masked `badge` icon
- [x] Register service worker on `init`, this will allow `add to home screen` banner to show up in Chrome, [ref](https://developers.google.com/web/fundamentals/engage-and-retain/app-install-banners/)
- [x] Should subscribe button only be visible if Lounge has been loaded over https? Service workers only work over HTTPS.
- [ ] Tests?
- [ ] Deal with notifications in queries
- [x] Push to hide notifications when any client opens lounge?
- [x] Do not send notifications for old messages (if server-time cap is available)
- [ ] Do not allow enabling notifications if "remember me" was not used
- [ ] Marking a channel as read should probably push `read` for that channel

Ref: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -688,7 +688,8 @@ kbd {
 	width: 100%;
 }
 
-#windows label {
+#windows label,
+#settings .error {
 	font-size: 14px;
 }
 
--- a/client/index.html
+++ b/client/index.html
@@ -293,20 +293,36 @@
 								</label>
 							</div>
 							{{/if}}
+							{{#unless public}}
 							<div class="col-sm-12">
-								<h2>Notifications</h2>
+								<h2>Push Notifications</h2>
+							</div>
+							<div class="col-sm-12">
+								<button type="button" class="btn" id="pushNotifications" disabled data-text-alternate="Unsubscribe from push notifications">Subscribe to push notifications</button>
+								<div class="error" id="pushNotificationsHttps">
+									<strong>Warning</strong>:
+									Push notifications are only supported over HTTPS connections.
+								</div>
+								<div class="error" id="pushNotificationsUnsupported">
+									<strong>Warning</strong>:
+									<span>Push notifications are not supported by your browser.</span>
+								</div>
+							</div>
+							{{/unless}}
+							<div class="col-sm-12">
+								<h2>Browser Notifications</h2>
 							</div>
 							<div class="col-sm-12">
 								<label class="opt">
 								<input id="desktopNotifications" type="checkbox" name="desktopNotifications">
-								Enable desktop notifications<br>
+								Enable browser notifications<br>
 								<div class="error" id="warnUnsupportedDesktopNotifications">
 									<strong>Warning</strong>:
-									Desktop notifications are not supported by your browser.
+									Notifications are not supported by your browser.
 								</div>
 								<div class="error" id="warnBlockedDesktopNotifications">
 									<strong>Warning</strong>:
-									Desktop notifications are blocked by your browser.
+									Notifications are blocked by your browser.
 								</div>
 								</label>
 							</div>
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -22,6 +22,7 @@ require("./socket-events");
 const constants = require("./constants");
 const storage = require("./localStorage");
 const utils = require("./utils");
+require("./webpush");
 
 $(function() {
 	var sidebar = $("#sidebar, #footer");
--- a/client/js/webpush.js
+++ b/client/js/webpush.js
@@ -1,0 +1,127 @@
+"use strict";
+
+const $ = require("jquery");
+const storage = require("./localStorage");
+const socket = require("./socket");
+
+const pushNotificationsButton = $("#pushNotifications");
+let clientSubscribed = null;
+let applicationServerKey;
+
+module.exports.configurePushNotifications = (subscribedOnServer, key) => {
+	applicationServerKey = key;
+
+	// If client has push registration but the server knows nothing about it,
+	// this subscription is broken and client has to register again
+	if (clientSubscribed === true && subscribedOnServer === false) {
+		pushNotificationsButton.attr("disabled", true);
+
+		navigator.serviceWorker.register("service-worker.js")
+			.then((registration) => registration.pushManager.getSubscription())
+			.then((subscription) => subscription && subscription.unsubscribe())
+			.then((successful) => {
+				if (successful) {
+					alternatePushButton().removeAttr("disabled");
+				}
+			});
+	}
+};
+
+if (isAllowedServiceWorkersHost()) {
+	$("#pushNotificationsHttps").hide();
+
+	if ("serviceWorker" in navigator) {
+		navigator.serviceWorker.register("service-worker.js").then((registration) => {
+			if (!registration.pushManager) {
+				return;
+			}
+
+			return registration.pushManager.getSubscription().then((subscription) => {
+				$("#pushNotificationsUnsupported").hide();
+
+				pushNotificationsButton
+					.removeAttr("disabled")
+					.on("click", onPushButton);
+
+				clientSubscribed = !!subscription;
+
+				if (clientSubscribed) {
+					alternatePushButton();
+				}
+			});
+		}).catch((err) => {
+			$("#pushNotificationsUnsupported span").text(err);
+		});
+	}
+}
+
+function onPushButton() {
+	pushNotificationsButton.attr("disabled", true);
+
+	navigator.serviceWorker.register("service-worker.js").then((registration) => {
+		registration.pushManager.getSubscription().then((existingSubscription) => {
+			if (existingSubscription) {
+				socket.emit("push:unregister");
+
+				return existingSubscription.unsubscribe();
+			}
+
+			return registration.pushManager.subscribe({
+				applicationServerKey: urlBase64ToUint8Array(applicationServerKey),
+				userVisibleOnly: true
+			}).then((subscription) => {
+				const rawKey = subscription.getKey ? subscription.getKey("p256dh") : "";
+				const key = rawKey ? window.btoa(String.fromCharCode.apply(null, new Uint8Array(rawKey))) : "";
+				const rawAuthSecret = subscription.getKey ? subscription.getKey("auth") : "";
+				const authSecret = rawAuthSecret ? window.btoa(String.fromCharCode.apply(null, new Uint8Array(rawAuthSecret))) : "";
+
+				socket.emit("push:register", {
+					token: storage.get("token"),
+					endpoint: subscription.endpoint,
+					keys: {
+						p256dh: key,
+						auth: authSecret
+					}
+				});
+
+				return true;
+			});
+		}).then((successful) => {
+			if (successful) {
+				alternatePushButton().removeAttr("disabled");
+			}
+		});
+	}).catch((err) => {
+		$("#pushNotificationsUnsupported span").text(err).show();
+	});
+
+	return false;
+}
+
+function alternatePushButton() {
+	const text = pushNotificationsButton.text();
+
+	return pushNotificationsButton
+		.text(pushNotificationsButton.data("text-alternate"))
+		.data("text-alternate", text);
+}
+
+function urlBase64ToUint8Array(base64String) {
+	const padding = "=".repeat((4 - base64String.length % 4) % 4);
+	const base64 = (base64String + padding)
+		.replace(/-/g, "+")
+		.replace(/_/g, "/");
+
+	const rawData = window.atob(base64);
+	const outputArray = new Uint8Array(rawData.length);
+
+	for (let i = 0; i < rawData.length; ++i) {
+		outputArray[i] = rawData.charCodeAt(i);
+	}
+
+	return outputArray;
+}
+
+function isAllowedServiceWorkersHost() {
+	return location.protocol === "https:" || location.hostname === "localhost" || location.hostname === "127.0.0.1";
+}
--- a/client/service-worker.js
+++ b/client/service-worker.js
@@ -1,0 +1,41 @@
+// The Lounge - https://github.com/thelounge/lounge
+/* global clients */
+"use strict";
+
+self.addEventListener("push", function(event) {
+	if (!event.data) {
+		return;
+	}
+
+	const payload = event.data.json();
+
+	if (payload.type === "notification") {
+		event.waitUntil(
+			self.registration.showNotification(payload.title, {
+				badge: "img/logo-64.png",
+				icon: "img/touch-icon-192x192.png",
+				body: payload.body,
+				timestamp: payload.timestamp,
+			})
+		);
+	}
+});
+
+self.addEventListener("notificationclick", function(event) {
+	event.notification.close();
+
+	event.waitUntil(clients.matchAll({
+		type: "window"
+	}).then(function(clientList) {
+		for (var i = 0; i < clientList.length; i++) {
+			var client = clientList[i];
+			if ("focus" in client) {
+				return client.focus();
+			}
+		}
+
+		if (clients.openWindow) {
+			return clients.openWindow(".");
+		}
+	}));
+});
--- a/src/client.js
+++ b/src/client.js
@@ -94,10 +94,20 @@ function Client(manager, name, config) {
 		client.config.sessions = {};
 	}
 
+	_.forOwn(client.config.sessions, (session) => {
+		if (session.pushSubscription) {
+			this.registerPushSubscription(session, session.pushSubscription, true);
+		}
+	});
+
 	if (client.name) {
 		log.info(`User ${colors.bold(client.name)} loaded`);
 	}
 }
+
+Client.prototype.isRegistered = function() {
+	return this.name.length > 0;
+};
 
 Client.prototype.emit = function(event, data) {
 	if (this.sockets !== null) {
@@ -310,11 +320,11 @@ Client.prototype.updateSession = function(token, ip, request) {
 		friendlyAgent += ` on ${agent.os.name} ${agent.os.version}`;
 	}
 
-	client.config.sessions[token] = {
+	client.config.sessions[token] = _.assign({
 		lastUse: Date.now(),
 		ip: ip,
 		agent: friendlyAgent,
-	};
+	}, client.config.sessions[token]);
 };
 
 Client.prototype.setPassword = function(hash, callback) {
@@ -410,7 +420,7 @@ Client.prototype.more = function(data) {
 Client.prototype.open = function(socketId, target) {
 	// Opening a window like settings
 	if (target === null) {
-		this.attachedClients[socketId] = -1;
+		this.attachedClients[socketId].openChannel = -1;
 		return;
 	}
 
@@ -423,7 +433,7 @@ Client.prototype.open = function(socketId, target) {
 	target.chan.unread = 0;
 	target.chan.highlight = false;
 
-	this.attachedClients[socketId] = target.chan.id;
+	this.attachedClients[socketId].openChannel = target.chan.id;
 	this.lastActiveChannel = target.chan.id;
 
 	this.emit("open", target.chan.id);
@@ -493,7 +503,7 @@ Client.prototype.quit = function() {
 	});
 };
 
-Client.prototype.clientAttach = function(socketId) {
+Client.prototype.clientAttach = function(socketId, token) {
 	var client = this;
 	var save = false;
 
@@ -507,7 +517,10 @@ Client.prototype.clientAttach = function(socketId) {
 		});
 	}
 
-	client.attachedClients[socketId] = client.lastActiveChannel;
+	client.attachedClients[socketId] = {
+		token: token,
+		openChannel: client.lastActiveChannel
+	};
 
 	// Update old networks to store ip and hostmask
 	client.networks.forEach((network) => {
@@ -545,6 +558,40 @@ Client.prototype.clientDetach = function(socketId) {
 			}
 		});
 	}
+};
+
+Client.prototype.registerPushSubscription = function(session, subscription, noSave) {
+	if (!_.isPlainObject(subscription) || !_.isPlainObject(subscription.keys)
+	|| typeof subscription.endpoint !== "string" || !/^https?:\/\//.test(subscription.endpoint)
+	|| typeof subscription.keys.p256dh !== "string" || typeof subscription.keys.auth !== "string") {
+		session.pushSubscription = null;
+		return;
+	}
+
+	const data = {
+		endpoint: subscription.endpoint,
+		keys: {
+			p256dh: subscription.keys.p256dh,
+			auth: subscription.keys.auth
+		}
+	};
+
+	session.pushSubscription = data;
+
+	if (!noSave) {
+		this.manager.updateUser(this.name, {
+			sessions: this.config.sessions
+		});
+	}
+
+	return data;
+};
+
+Client.prototype.unregisterPushSubscription = function(token) {
+	this.config.sessions[token].pushSubscription = null;
+	this.manager.updateUser(this.name, {
+		sessions: this.config.sessions
+	});
 };
 
 Client.prototype.save = _.debounce(function SaveClient() {
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -3,6 +3,7 @@
 const Chan = require("../../models/chan");
 const Msg = require("../../models/msg");
 const LinkPrefetch = require("./link");
+const Helper = require("../../helper");
 
 module.exports = function(irc, network) {
 	var client = this;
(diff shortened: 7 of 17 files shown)


astorije:
This is so slick! I'm adding this to v2.4.0 milestone but I can't wait for this to land! 🎉

christer88:
Cannot wait for this feature! Good work!

MaxLeiter:
I think subscribe button should show when Lounge is over HTTP, but have it disabled with an explanatory tooltip / label

xPaw:
Implemented proper UI button and unsubscription support.

xPaw:
I rebased this PR on top of #1199 which will make it easier for me to update it later. #1199 has to be merged in first.

client/js/options.js
```diff
@@ -143,3 +248,19 @@ if (("Notification" in window)) {
 	desktopNotificationsCheckbox.attr("disabled", true);
 	desktopNotificationsCheckbox.attr("checked", false);
 }
+
+function urlBase64ToUint8Array(base64String) {
```
YaManicKill: I'm not sure about what the logic is with this...

I get why we have it, but why do we do what we do in the function itself?

client/js/options.js
```diff
@@ -111,6 +138,84 @@ $("#desktopNotifications").on("change", function() {
...
+					alternatePushButton();
+				}
+			});
+		}).catch((err) => {
+			$("#pushNotificationsUnsupported p").text(err);
+		});
+	}
+}
+
+function alternatePushButton() {
```
YaManicKill: So this is to change the text on the button? Feels a little unusual. Why not just have 2 buttons, and only show the appropriate one?

client/js/options.js
```diff
@@ -111,6 +138,84 @@ $("#desktopNotifications").on("change", function() {
 	}
 });
 
+function onPushButton() {
```
YaManicKill: Can we abstract this out into another module? It doesn't feel like it's anything actually to do with the options itself.

client/js/options.js
```diff
@@ -111,6 +138,84 @@ $("#desktopNotifications").on("change", function() {
...
+		});
+	})).catch((err) => {
+		console.error(err);
+		window.alert(err);
+	});
+
+	return false;
+}
+
+if (location.protocol === "https:" || location.hostname === "localhost" || location.hostname === "127.0.0.1") {
```
YaManicKill: Would be nice if this was just if (allowedToEnableServiceWorkers()) or something which took this logic into it. But not a massive thing.

client/js/options.js
```diff
@@ -111,6 +138,84 @@ $("#desktopNotifications").on("change", function() {
...
+	$("#pushNotificationsHttps").hide();
+
+	if ("serviceWorker" in navigator) {
+		navigator.serviceWorker.register("service-worker.js").then((registration) => {
+			if (!registration.pushManager) {
+				return;
+			}
+
+			return registration.pushManager.getSubscription().then((subscription) => {
+				$("#pushNotificationsUnsupported").hide();
```
YaManicKill: Is there any reason not to hide this as soon as we know we have service workers? Or is the error from registering actually ever going to happen?

src/client.js
```diff
@@ -76,6 +76,10 @@ function Client(manager, name, config) {
 		id: id++,
 		name: name,
 		networks: [],
+		push: { // TODO: This should not ship
```
YaManicKill: Oh? Should we not have this at all, is this just dev stuff?

src/plugins/irc-events/message.js
```diff
@@ -95,5 +96,27 @@ module.exports = function(irc, network) {
 		if ([Msg.Type.MESSAGE, Msg.Type.ACTION].indexOf(data.type) !== -1) {
 			LinkPrefetch(client, chan, msg);
 		}
+
+		// TODO: For now, do not send notifications if any client is attached
```
YaManicKill: Yeah, good thing to start with. We probably want to eventually add this as an option, maybe? Or do some cleverness with away, etc. But for a first version this makes perfect sense, we'll have less complaints about this feature than the other way around ;-)

src/plugins/irc-events/message.js
```diff
@@ -95,5 +96,27 @@ module.exports = function(irc, network) {
...
+		if (highlight
+		&& (!data.time || data.time > Date.now() - 2.16e+7) // Message not older than 6 hours
```
YaManicKill: Why? Arbitrary, or is there a reason for this? What if we have an issue with timezones?

src/server.js
```diff
@@ -289,10 +289,36 @@ function init(socket, client, generateToken) {
...
+				return;
+			}
+
+			const registration = client.registerPushSubscription(client.config.sessions[subscription.token], subscription);
+
+			if (registration) {
+				client.manager.webPush.pushSingle(client, registration, {
+					timestamp: Date.now(),
+					title: "The Lounge",
+					body: "🚀 Push notifications have been enabled"
```
YaManicKill: Classy

src/server.js
```diff
@@ -289,10 +289,36 @@ function init(socket, client, generateToken) {
...
+				});
+			}
+		});
+
+		socket.on("push:unregister", (token) => {
+			if (!client.isRegistered()) {
+				return;
```
YaManicKill: I feel like this would be a "sigh" moment if this happens much :-P

src/webpush.js
```diff
@@ -0,0 +1,69 @@
...
+
+			if (typeof parsedData.publicKey === "string" && typeof parsedData.privateKey === "string") {
+				this.vapidKeys = {
+					publicKey: parsedData.publicKey,
+					privateKey: parsedData.privateKey,
+				};
+			}
+		}
+
+		if (!this.vapidKeys) {
```
YaManicKill: I have no idea what these are.

src/webpush.js
```diff
@@ -0,0 +1,69 @@
...
+			this.vapidKeys = WebPushAPI.generateVAPIDKeys();
+
+			fs.writeFileSync(vapidPath, JSON.stringify(this.vapidKeys, null, "\t"));
+
+			log.info("New VAPID key pair has been generated for use with push subscription.");
+		}
+
+		WebPushAPI.setVapidDetails(
+			"https://github.com/thelounge/lounge",
+			this.vapidKeys.publicKey,
```
YaManicKill: \>\>\> INSERT MAGIC <<<

src/webpush.js
```diff
@@ -0,0 +1,69 @@
...
+			this.vapidKeys.privateKey
+		);
+	}
+
+	push(client, payload) {
+		const now = Date.now();
+
+		// TODO: Currently, limit to 5 notifications in a minute
```
YaManicKill: What is the "TODO"? To make this configurable or change the limit?

client/js/options.js
xPaw: PushManager is a newer api, so service worker might get registered, but not the pushmanager.

src/client.js
xPaw: It's here until we figure out push limits (e.g. how many messages per minute and stuff like that.)

src/plugins/irc-events/message.js
xPaw: Arbitrary mostly. It's all UTC timestamps.

src/server.js
xPaw: I chose to implement a `isRegistered` function over simply doing `public` check to allow for mixed mode in the future (if we ever do it).

src/webpush.js
xPaw: It's the keys that are used to encrypt notifications, they're required as per the web push spec.

https://datatracker.ietf.org/doc/draft-ietf-webpush-vapid/

src/webpush.js
xPaw: Either, really, this was just a quick hack fix until we know what limits we want to impose.

client/js/options.js
xPaw: Not sure I understand the question.

client/js/options.js
xPaw: Felt a little simpler to implement at the time.

client/js/options.js
xPaw: Will do.

client/js/options.js
xPaw: Will do.

src/webpush.js
YaManicKill: Yeah, that much I guessed, but it's a bit of a weird name and seems odd. I'm not saying you should do anything about it. Just braindumping, really.

src/plugins/irc-events/message.js
YaManicKill: Aight

client/js/options.js
YaManicKill: I'm not sure I do either.

src/webpush.js
YaManicKill: Maybe stick in an issue and put the number here? Otherwise we will probably forget.

client/index.html
```diff
@@ -301,20 +301,36 @@ <h1 class="title">Settings</h1>
 								</label>
 							</div>
 							{{/if}}
+							{{#unless public}}
+							<div class="col-sm-12">
+								<h2>Push Notifications</h2>
```
astorije: We should consolidate "Push Notifications" and "Browser Notifications" into one group called "Notifications". From the user perspective, it just means being notified when not looking at the app, whatever strategy used underneath.

You mention both can be enabled on desktop, what is the use case for this?
I would expect that when app is opened (on browser), desktop notifications apply, but when it's shut down, push notifications are in use. The "Enable notifications" checkbox should enable them all.

If you have details as of what works in what context (push vs. normal notifications), that would help me figure out what would be the most usable thing to do :)

astorije: Okay, playing with this on the laptop actually shows me the usage of both, which is pretty much how I've described it.

So definitely one checkbox "Enable notifications" that enables both, based on capabilities. The only issue I see is going to be on the side of the code, as we currently disable the checkbox when `Notification` is not available. Now it's going to be a bit more intricate, but we should still be able to do all that with a single checkbox, which would be the most natural in the UI.

client/index.html
```diff
@@ -301,20 +301,36 @@ <h1 class="title">Settings</h1>
...
+							{{/unless}}
 							<div class="col-sm-12">
-								<h2>Notifications</h2>
+								<h2>Browser Notifications</h2>
 							</div>
 							<div class="col-sm-12">
 								<label class="opt">
 								<input id="desktopNotifications" type="checkbox" name="desktopNotifications">
-								Enable desktop notifications<br>
+								Enable browser notifications<br>
```
astorije: This would just be `Enable notifications` now.

client/js/lounge.js
```diff
@@ -22,6 +22,7 @@ require("./socket-events");
 const constants = require("./constants");
 const storage = require("./localStorage");
 const utils = require("./utils");
+require("./webpush");
```
astorije: This is very much side-effecty, I don't like that too much. But rather a nitpick, I guess.

src/client.js
astorije: How do other systems handle this? I feel like if someone sends me 5 PMs quickly, we should rate limit them. At least, if Telegram was doing that to me, I'd be pissed lol.
I have noticed though that it only shows up to a certain amount, then just rings but doesn't show (or show "N new messages") but that might be handled by the OS itself.

astorije:
I haven't gone through the whole code, but skimmed it + ran it locally, it's pretty cool! My biggest issue right now is with the UI/UX, but nothing impossible I think. Once #1199 gets merged, I'll rebase this (if @xPaw is not around) and the PR should be much less meaty :)

Regarding what's in the PR description:

> - [x] Better UI for (un)subscribing

Yup, left a few comments inline, I might play with this on my end once #1199 is merged and I have a deeper knowledge of the code, if you are not able to.

> - [ ] Deal with browser and push notifications (desktop can have both enabled)

That's related to previous point. In a nutshell, Though they can have both, we shouldn't push notify if app is open (i.e. browser notification are available) if that's possible. Again, I'll have a clearer vision when I know the code more.

> - [ ] Remove dead subscriptions (e.g. when push endpoint returns `not registered`)

I'm not sure I'm following this one.

> - [ ] Currently the code limits to 5 notifications per minute (should tags be used, etc?)

If we have a per-minute limit, this should be really high, like an order of magnitude higher I think. Think PMs that can come at a fast rate from a single person.

> - [ ] Currently notifications are only emitted when there are no clients attached, what logic should actually be used?

In my opinion, notifications should be emitted to all registered clients that are not currently active. If I have The Lounge running in the background on my laptop, I expect a browser notification on the laptop, and a push notification on my phone. I don't know about others but that's how Telegram does it, I believe, with the addition of removing notifications from all devices if it is read on another. In the previous example, this means that if I go to The Lounge on my laptop after receiving a notification, it disappears from both the laptop and the phone.

> - [ ] Clicking on notification should focus channel where it came from

I believe #1283 would help with that, right?

> - [x] Masked `badge` icon

What do you mean by "masked"?

> - [x] Should subscribe button only be visible if Lounge has been loaded over https? Service workers only work over HTTPS.

Once we consolidate the "Enable notifications" checkboxes, we would not need to think about this anymore. However, we could add a warning to mention that notifications are not available if not on HTTPS, but I wouldn't worry about it.

> - [ ] Tests?

👍 :D

> - [ ] Deal with notifications in queries

What's the behavior at the moment? You don't get any notifications?

Also, does this feature apply to custom highlights as well?

> - [ ] Push to hide notifications when any client opens lounge?

Yes! 👍 So annoying when you go on a device and see dozens of notifications you have already acted upon (I'm looking at you, Gmail).

> - [x] Do not send notifications for old messages (if server-time cap is available)

Is this  a concern outside of ZNC?

src/client.js
```diff
@@ -529,6 +548,44 @@ Client.prototype.clientDetach = function(socketId) {
...
+
+	session.pushSubscription = data;
+
+	if (!noSave) {
+		this.manager.updateUser(this.name, {
+			sessions: this.config.sessions
+		});
+	}
+
+	log.debug(`Subscribed ${this.name} to ${subscription.endpoint}`);
```
astorije: Do you want to add a `notifications` property to [`debug` in the config file](https://github.com/thelounge/lounge/blob/master/defaults/config.js#L395) or was this just for dev purpose, to remove before merge?

xPaw: Thats to be removed.

astorije: Okay, I have removed them in an extra commit so we don't forget.

xPaw:
@astorije, I'm not sure what work is involved in combining the notifications into a single button, but i feel like it might get very tricky if we also want to correctly report state to user.

astorije:
Rebased with `master` now that #1199 is merged.

> I'm not sure what work is involved in combining the notifications into a single button, but i feel like it might get very tricky if we also want to correctly report state to user.

I'll give it a shot (in extra commits), see where we can get :)

xPaw:
I have fixed some bugs with new changes to master. I have also implemented hiding of all notifications whenever you open the client on any device.

xPaw:
I have implemented sending notifications to detached clients, even if there's some other client attached.

xPaw:
Looks like hiding notifications when opening lounge is not a feasible feature, as Chrome will show "This site has been updated in the background" instead.

PolarizedIons:
Great work on the PR so far :smile: 

However, I have some concerns: @astorije regarding this paragraph:
> In my opinion, notifications should be emitted to all registered clients that are not currently active. If I have The Lounge running in the background on my laptop, I expect a browser notification on the laptop, and a push notification on my phone. I don't know about others but that's how Telegram does it, I believe, with the addition of removing notifications from all devices if it is read on another. In the previous example, this means that if I go to The Lounge on my laptop after receiving a notification, it disappears from both the laptop and the phone.

I do not agree with this at all. For example, let's say I'm connected on my pc, and start having a conversation with someone. Each reply is sent as a push notification to my phone, which plays its notification sound. This is annoying & unnecessary - I know they sent a reply, I've already seen it on the pc. This makes it so I either have to mute/ignore my phone, or not have push notifications enabled on it :( 

(This is the current behaviour of the PR)

Telegram does this, but only when it isn't minimised to the taskbar (at least for me)

--- 

The not being able to hiding notifications is also :(

christer88:
> I do not agree with this at all. For example, let's say I'm connected on my pc, and start having a conversation with someone. Each reply is sent as a push notification to my phone, which plays its notification sound. This is annoying & unnecessary - I know they sent a reply, I've already seen it on the pc. This makes it so I either have to mute/ignore my phone, or not have push notifications enabled on it :(

This is how Facebook Messenger works. If someone send you a message, you get a notification sound in your browser, phone, tablet etc. Reluctant to not having them sent out to all non-active clients.

astorije:
> I do not agree with this at all.

@PolarizedIons, are you sure though? :)
If I read what you posted correctly, this seems to be exactly what I meant. Now, I might have been unclear myself, which is totally possible 😅
Essentially, if you are actively having a conversation, you shouldn't be getting any notifications, since you are currently looking at the app and what the notifications would notify you about anyway.

xPaw:
@astorije Basically this PR is just the push implementation, we should handle "logic on when to send a push" in a separate PR, otherwise we will never get this done.

Initially I wanted to send a push to hide notifications whenever you interact with a channel, but Chrome wants any push to show a notification, so that doesn't work and defeats the purpose.

client/js/webpush.js
```diff
@@ -0,0 +1,129 @@
...
+					p256dh: key,
+					auth: authSecret
+				}
+			});
+
+			alternatePushButton().removeAttr("disabled");
+		});
+	})).catch((err) => {
+		console.error(err);
+		window.alert(err);
```
astorije: Mmmmh that's a leftover from development, right?

client/service-worker.js
```diff
@@ -0,0 +1,41 @@
+// The Lounge - https://github.com/thelounge/lounge
+/* global clients */
```
astorije: Does this have a special meaning? It looks like a machine-readable comment, like ESLint does.

client/service-worker.js
```diff
@@ -0,0 +1,41 @@
...
+"use strict";
+
+self.addEventListener("push", function(event) {
```
astorije: I take it this file gets distributed as is? Meaning, we don't webpack it, correct?

I noticed the syntax is a bit less ES6-looking (`var` instead of `const`, no arrow functions, etc.), and I'm assuming that's because of the above. If a client supports service workers, isn't it fair to assume it will support `const` and `=>`?

Also, is ESLint going through this file, or ignoring it?

client/js/webpush.js
xPaw: It's not as there is no proper UI for this.

client/service-worker.js
xPaw: It is for eslint.

client/service-worker.js
xPaw: Eslint does not ignore this file. I'm actually not sure about arrow support.

client/js/webpush.js
astorije: Can we add an extra block next to `pushNotificationWarning` stuff? We really can't be reporting in `window.alert`...

client/service-worker.js
astorije: You think arrow functions are more recent than push notifications / service workers stuff? I don't care either way, I'm just curious now :)

YaManicKill: Simple solution, another webpack script for this file.

xPaw: Well, then we need to gitignore another file since we don't have a public/built folder...

astorije: Webpacking a single file seems overkill to me. Also, Webpack generates fairly obfuscated code, and I don't know if we can/should ship sourcemaps for it, and we probably don't want to obfuscate what runs in service workers so people can see what they're running if they want (they can look at the repo, but that's different, IMO).

Anyway, I think we should save ourselves this trouble and just live it as is.

xPaw: Agreed, the file is small enough that we won't benefit from webpack.

client/js/webpush.js
```diff
@@ -0,0 +1,127 @@
...
+					.on("click", onPushButton);
+
+				clientSubscribed = !!subscription;
+
+				if (clientSubscribed) {
+					alternatePushButton();
+				}
+			});
+		}).catch((err) => {
+			$("#pushNotificationsUnsupported span").text(err);
```
astorije: Does this need a `.show()` like [here](https://github.com/thelounge/lounge/pull/1124/files#diff-ab9ed8f3a9f1719ea73cdb90cf1599a3R95)?

src/plugins/irc-events/message.js
```diff
@@ -102,5 +103,24 @@ module.exports = function(irc, network) {
 		}
 
 		chan.pushMessage(client, msg, !self);
+
+		// Do not send notifications for messages older than six hours (znc buffer for example)
```
astorije: Why 6h and not, like, 15 minutes or less?

src/server.js
```diff
@@ -296,6 +296,31 @@ function initializeClient(socket, client, generateToken, token) {
...
+
+		const registration = client.registerPushSubscription(client.config.sessions[token], subscription);
+
+		if (registration) {
+			client.manager.webPush.pushSingle(client, registration, {
+				type: "notification",
+				timestamp: Date.now(),
+				title: "The Lounge",
+				body: "🚀 Push notifications have been enabled"
+			});
```
astorije: We don't do this for non-push notifications, I'm not sure we need this.

client/css/style.css
```diff
@@ -688,7 +688,8 @@ kbd {
 	width: 100%;
 }
 
-#windows label {
+#windows label,
+#settings .error {
 	font-size: 14px;
 }
```
astorije: Before | After
--- | ---
screen shot 2017-08-22 at 01 18 09| screen shot 2017-08-22 at 01 21 11

client/index.html
astorije: We decided to keep merging notification types and other UX/UI major improvements for later.

client/index.html
```diff
@@ -293,20 +293,36 @@ <h1 class="title">Settings</h1>
 								</label>
 							</div>
 							{{/if}}
+							{{#unless public}}
+							<div class="col-sm-12">
+								<h2>Push Notifications</h2>
+							</div>
+							<div class="col-sm-12">
+								<button type="button" class="btn" id="pushNotifications" disabled data-text-alternate="Unsubscribe from push notifications">Subscribe to push notifications</button>
```
astorije: So yeah, as discussed, let's keep UX improvements of this for a latter PR. But it's definitely going to take some time, discussions and attempts before we hit a correct UI/UX.
In the meantime, could you make this an "Enable push notifications" checkbox instead? Thanks! That's my last request before approving this 😊

client/index.html
astorije: We decided to keep merging notification types and other UX/UI major improvements for later.

client/js/webpush.js
xPaw: Should not need it, as the `hide` is inside this same chain.

src/server.js
xPaw: Web push is all new stuff, so I'd rather have a reassuring confirmation that it actually worked.

src/plugins/irc-events/message.js
xPaw: Changed to 15m.

client/service-worker.js
YaManicKill: So far, sure. But it'll be very useful when we need to add other things to the service worker (offline support) and we'll get the benefits of transpiling just now.

src/server.js
YaManicKill: I agree, as a user it is nice to see it is all working fine.

client/service-worker.js
xPaw: What do we need to transpile? If the brower supports service workers, it's most likely than not supports ES6.

YaManicKill: Well, for 1 you just said earlier that you don't know if all browsers support arrow functions.

The thing is, that ES6 isn't a binary thing. Plenty of browsers support some features and not others. And transpiling is for future things as well. 

I don't really care enough, I was just giving you a solution for the problem you were discussing.

xPaw: It's easy to look up that info though:

https://caniuse.com/#feat=arrow-functions
https://caniuse.com/#feat=serviceworkers

I know it's not binary, but it's fairly safe to assume that if SW is supported in a browser, most of the ES6 features we would want to use are supported. We can webpack it in #1056 (if anyone ever wants to do it...)

astorije:
Just for the record, I'm seeing a build breaking in Travis, but it's a one-time thing for now so putting it here in case it comes back and we want more examples: https://travis-ci.org/thelounge/lounge/jobs/267458577

PR #1125, opened 2017-05-01
Remove autocompletion section from the help page

astorije:
Rationale for this is that the whole point of autocompletion is to be intuitive and show up naturally when starting to type something else. Helper for this is just extra noise user does not need.

For example, all commands start with `/`, so obviously starting to type a command will trigger autocomplete. This is true for channels as well.
Emoji are a bit particular because all systems that support emoji open their completion with `:`.
The only not-so-intuitive completion strategy is for nicks because it is not so common to start them with `@` on IRC, but as long as we keep tab completion after any set of characters, this is fine. It will be even nicer once regular tab completion uses the same autocompletion dropdown.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -529,52 +529,7 @@
 							</div>
 						</div>
 
-						<h2>Autocompletion</h2>
-
-						<p>
-							Start typing the following characters followed by any letter to
-							trigger the autocompletion dropdown:
-						</p>
-
-						<div class="help-item">
-							<div class="subject">
-								<code>@</code>
-							</div>
-							<div class="description">
-								<p>Nickname</p>
-							</div>
-						</div>
-
-						<div class="help-item">
-							<div class="subject">
-								<code>#</code>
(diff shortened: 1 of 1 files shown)


client/index.html
```diff
@@ -529,53 +529,8 @@ <h1 class="title">Help</h1>
...
-							</div>
-							<div class="description">
-								<p>Emoji</p>
-							</div>
-						</div>
-
 						<h2>Commands</h2>
 
-						<p>All commands can be autocompleted with <kbd>tab</kbd>.</p>
-
```
astorije: This used to be necessary because nothing on the UI would indicate that commands could be completed. Now, typing `/` opens the autocompletion dropdown, so visual help is making this redundant.

YaManicKill:
I feel like this was merged too quickly. Only 1 approval feels odd for this, especially as I was assigned by @astorije, it feels odd for it to have been merged so suddenly.

PR #1132, opened 2017-05-03
Fix date-marker not being removed on loading new messages

### LINKED ISSUE #763: Date-separator not being removed when it should sometimes

PolarizedIons:
Fixes #763 again. Affects mastrer (5eb5b99). This is caused by 648cfd12, where @astorije put it inside a container, but didn't update the class we searched for.

What it currently looks like:

Expected behavior:

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -469,20 +469,20 @@ $(function() {
 			.find("#chan-" + data.chan)
 			.find(".messages");
 
-		// Remove the date-change marker we put at the top, because it may
-		// not actually be a date change now
-		var children = $(chan).children();
-		if (children.eq(0).hasClass("date-marker")) { // Check top most child
-			children.eq(0).remove();
-		} else if (children.eq(0).hasClass("unread-marker") && children.eq(1).hasClass("date-marker")) {
-			// Otherwise the date-marker would get 'stuck' because of the new-message marker
-			children.eq(1).remove();
-		}
-
 		// get the scrollable wrapper around messages
 		var scrollable = chan.closest(".chat");
 		var heightOld = chan.height();
 		chan.prepend(documentFragment).end();
+
+		// Remove the date-change marker we put at the top, because it may
+		// not actually be a date change now
+		var children = $(chan).children();
+		if (children.eq(0).hasClass("date-marker-container")) { // Check top most child
+			children.eq(0).remove();
+		} else if (children.eq(1).hasClass("date-marker-container")) {
+			// The unread-marker could be at index 0, which will cause the date-marker to become "stuck"
+			children.eq(1).remove();
+		}
 
 		// restore scroll position
 		var position = chan.height() - heightOld;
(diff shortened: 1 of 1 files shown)


PolarizedIons:
err. Now yes :)

xPaw:
Looking at your gifts, looks like the viewport jumps a little when content loads now.

PolarizedIons:
This is the behavior for bc5b03d, before this broke. Not really sure how to fix it either, tbh.

bews:
Current master.  Does this PR fix this too?

PolarizedIons:
Pushed the changes @xPaw, not as hard as I though :smile: 


@bews if those are "stuck" at the top (instead of being caused by join/parts being hidden), then yes, this should fix that.

client/js/lounge.js
```diff
@@ -472,16 +472,19 @@ $(function() {
...
+		} else if (children.eq(0).hasClass("unread-marker") && children.eq(1).hasClass("date-marker-container")) {
+			// Check for unread-marker too - oherwise the date-marker would get 'stuck' because of the new-message marker
+			dateMarkerHeight = children.eq(1).height();
 			children.eq(1).remove();
 		}
 
 		// get the scrollable wrapper around messages
 		var scrollable = chan.closest(".chat");
-		var heightOld = chan.height();
+		var heightOld = chan.height() + dateMarkerHeight; // Add the date marker's height because it was already removed
```
xPaw: Why don't you just get the height before removing the date marker?

PolarizedIons: I guess I was trying to not assume anything, but it is a guarantee that there will be one at the top, so thanks. Moved calculating the height to before I remove it.

client/js/lounge.js
```diff
@@ -469,21 +469,21 @@ $(function() {
...
+
 		// Remove the date-change marker we put at the top, because it may
 		// not actually be a date change now
 		var children = $(chan).children();
-		if (children.eq(0).hasClass("date-marker")) { // Check top most child
+		if (children.eq(0).hasClass("date-marker-container")) { // Check top most child
 			children.eq(0).remove();
-		} else if (children.eq(0).hasClass("unread-marker") && children.eq(1).hasClass("date-marker")) {
-			// Otherwise the date-marker would get 'stuck' because of the new-message marker
+		} else if (children.eq(0).hasClass("unread-marker") && children.eq(1).hasClass("date-marker-container")) {
```
xPaw: Should that first one check for container too?

PolarizedIons: Hmm. I guess it isn't needed, because there will never be normal messages at index 0

astorije:
Unfortunately, I noticed this is still a bug in `v2.3.0-rc.1` :(

date-marker-bug

@PolarizedIons, do you think you could take a look?

PolarizedIons:
Ah whoops, simple mistake, though I wish I would stop making them in PR's :\

astorije:
It's ok, we all do, and we caught this before releasing :)

Issue #404, opened 2016-06-13
electron lounge shell


pguth:
This really should get an electron stand-alone release :D

dgw:
So…it would run the server and the client on the local machine, and quit the server when the user closes the window?

pguth:
To give Lounge (and other webapps) a native feeling Chrome offers the `--webapp` flag but that solution has two downsides:

- When switching Wifis the client goes down and does not come up again when the connectivity is restored.
- `Assign to: [Desktop 1...n]` is not available on MacOS.

That's why I think an electron container that just cares about reloading on restored connectivity would be quite an improvement on MacOS.

<img width="249" alt="screen shot 2016-06-13 at 23 20 23" src="https://cloud.githubusercontent.com/assets/85259/16023542/7a461f12-31bd-11e6-8225-7d86679c785c.png">

williamboman:
[nativefier](https://github.com/jiahaog/nativefier) might be of interest.

pguth:
Oh, yes. Thanks @williamboman!

Maybe give that hint in the readme right away like so https://github.com/thelounge/lounge/pull/405?

PR #1137, opened 2017-05-05
optimize `.png` icons, normalize filenames, add precomposed touch icon

ajb:
Hey there,

Big fan of the project, and much ❤️ for all your hard work.

I used [nativefier](https://github.com/jiahaog/nativefier) to wrap The Lounge into a "native" app (as per https://github.com/thelounge/lounge/issues/404#issuecomment-225716492), and the application icon that it chose was low-res and crappy looking:

screen shot 2017-05-05 at 10 29 13 am

I decided to dig into the code that nativefier uses to choose an application icon, and saw that it selects the icon with the largest filesize, since it assumes that it's the highest resolution. 

Upon investigation of the icons in `client/img`, I noticed that the `.png` icons weren't optimized, and that the 64x64 icon had the largest filesize. This PR does the following:

- Renames `touch-icon-192x192.png` to `apple-touch-icon-192x192.png`, to standardize the filenames
- Optimizes all of the `.png` icons with [Imageoptim](https://imageoptim.com/mac)
- Adds a "precomposed" touch icon, which results in a nativefier'd application icon that looks like this:

screen shot 2017-05-05 at 10 41 33 am

(It's not perfect... it could stand to be a lot higher resoluton, but it's way better!)

Thanks for your consideration of this PR 😄


xPaw:
I will close this. We should be getting a new logo design soon, and we will handle all the relevant icon stuff in that, and will surely keep this PR in mind.

PR #1141, opened 2017-05-06
Show seconds in timestamp

bews:
Just a simple option in settings (disabled by default):

Here how it looks:

Related issue: #98

### DIFF
--- a/client/js/libs/handlebars/tz.js
+++ b/client/js/libs/handlebars/tz.js
@@ -1,7 +1,10 @@
 "use strict";
 
 const moment = require("moment");
+const constants = require("../../constants");
 
 module.exports = function(time) {
-	return moment(time).format("HH:mm");
+	const options = require("../../options");
+	const format = options.showSeconds ? constants.timeFormats.msgWithSeconds : constants.timeFormats.msgDefault;
+	return moment(time).format(format);
 };
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -3,6 +3,7 @@ const $ = require("jquery");
 const settings = $("#settings");
 const userStyles = $("#user-specified-css");
 const storage = require("./localStorage");
+const tz = require("./libs/handlebars/tz");
 
 const windows = $("#windows");
 const chat = $("#chat");
@@ -19,6 +20,7 @@ const options = $.extend({
 	notifyAllMessages: false,
 	part: true,
 	quit: true,
+	showSeconds: false,
 	theme: $("#theme").attr("href").replace(/^themes\/(.*).css$/, "$1"), // Extracts default theme name, set on the server configuration
 	thumbnails: true,
 	userStyles: userStyles.text(),
@@ -79,6 +81,10 @@ settings.on("change", "input, select, textarea", function() {
 			// Ensure we don't have empty string in the list of highlights
 			// otherwise, users get notifications for everything
 			return h !== "";
+		});
+	} else if (name === "showSeconds") {
+		chat.find(".msg > .time").each(function() {
+			$(this).text(tz($(this).parent().data("time")));
 		});
 	}
 }).find("input")
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -901,8 +901,7 @@ kbd {
 
 #chat .time {
 	color: #ddd;
-	text-align: right;
-	width: 46px;
+	padding-left: 10px;
 }
 
 #chat .from {
--- a/client/index.html
+++ b/client/index.html
@@ -249,6 +249,12 @@
 									Show quits
 								</label>
 							</div>
+							<div class="col-sm-6">
+								<label class="opt">
+									<input type="checkbox" name="showSeconds">
+									Show seconds in timestamp
+								</label>
+							</div>
 							<div class="col-sm-12">
 								<h2>Visual Aids</h2>
 							</div>
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -54,7 +54,13 @@ const commands = [
 	"/whois"
 ];
 
+const timeFormats = {
+	msgDefault: "HH:mm",
+	msgWithSeconds: "HH:mm:ss"
+};
+
 module.exports = {
 	colorCodeMap: colorCodeMap,
+	timeFormats: timeFormats,
 	commands: commands
 };


MaxLeiter:
not looking at code, am fan of feature :+1:

christer88:
It's a good start, but I think it would be even better if we could set the date format yourself, e.g. https://www.npmjs.com/package/dateformat
Default setting can include seconds. :)

astorije:
> I think it would be even better if we could set the date format ourself

We have discussed this in other tickets, this is not something we will do as part of the main codebase (what a nice way to say "packages!" until we get to that 😅). I invite you to look for them for rationale, I'm on a connection that's too spotty to browser around, sorry. Essentially, 12/24h + show seconds checkboxes will cover 95-99% of use cases while not having to learn a syntax for that (in fact, most if not all OS settings offer the same) so checkboxes = ease of use.

@bews, we should be able to get that out with the next release (v2.4.0) once reviewed, thanks! 😃

client/views/msg.tpl
```diff
@@ -1,6 +1,6 @@
 <div class="msg {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}}" id="msg-{{id}}" data-time="{{time}}" data-from="{{from}}">
-	<span class="time tooltipped tooltipped-e" aria-label="{{localetime time}}">
-		{{tz time}}
+	<span class="time tooltipped tooltipped-e {{#if seconds}}seconds{{/if}}" aria-label="{{localetime time}}">
```
xPaw: I think it's better to set option classes on `body`, like `option-seconds`. How would this mix with 12h+seconds display anyway? Perhaps it's better to make it a flexible column, if possible.

bews: >  How would this mix with 12h+seconds (01:11:11 PM) display anyway?

Is it a planned feature or I'm missing smth?

YaManicKill: I think he's just meaning as a future plan, trying to think about how this would interoperate with it.

client/js/libs/handlebars/tz.js
```diff
@@ -2,6 +2,6 @@
 
 const moment = require("moment");
 
-module.exports = function(time) {
-	return moment(time).format("HH:mm");
+module.exports = function(time, seconds) {
+	return seconds ? moment(time).format("HH:mm:ss") : moment(time).format("HH:mm");
```
YaManicKill: Can I be a code pedant and request something like this:
```
const format = seconds ? "HH:mm:ss" : "HH:mm";
return moment(time).format(format);
```

I would also prefer if the formats were in constants somewhere, but I won't push that one too hard.

client/js/libs/handlebars/tz.js
```diff
@@ -2,6 +2,6 @@
 
 const moment = require("moment");
 
-module.exports = function(time) {
-	return moment(time).format("HH:mm");
+module.exports = function(time, seconds) {
```
YaManicKill: There's something a little bit weird about passing through seconds here, but I don't know what the better option would be, tbh. If we had control over requring these handlebars templates, then we could use the modules that we are creating, but webpack pulls them in itself, so we'll end up with differing settings, which isn't good. So, I guess this is just me moaning about the system rather than asking you to do anything.

client/views/msg.tpl
YaManicKill: If we cared about it updating correctly when we change the settings (currently if you change the settings only new messages will have the correct format) then we could set it on the body (or container) and then have both time formats rendered, and show/hide in css based on the parameter we set on the container. That doesn't reaaaaally work well if we end up having multiple settings, and maybe feels a bit overkill (it doubles the number of time spans we have, with half of them being hidden). Let's leave that for now, this is good enough for just now I think. This is a setting we should have had a long time ago, personally.

client/js/options.js
```diff
@@ -19,6 +19,7 @@ const options = $.extend({
 	notifyAllMessages: false,
 	part: true,
 	quit: true,
+	seconds: false,
```
YaManicKill: Maybe change this to "showSeconds" because it makes sense here, but everywhere else (data.msg.seconds) it isn't clear exactly what it does.

bews:
So, I refactored this thing and now there even less code :smile: 

Decided to keep the previous commit for now to compare.

client/js/libs/handlebars/tz.js
```diff
@@ -1,7 +1,9 @@
 "use strict";
 
 const moment = require("moment");
+const options = require("../../options");
```
xPaw: Does this happen to register options stuff twice? That code isn't so modular to be required multiple times afaik.

bews: > Does this happen to register options stuff twice?

Just checked - it doesn't.

YaManicKill: So, I know for certain that node caches require, so that it only runs the stuff once, and you get the same object back if you require it multiple times. So that's exactly how the options module was designed, so that we can require it anywhere within the client code.

I'm not 100% sure on that with webpack, but if @bews says it works, then I guess it does.

I'll do some testing of this before. I thought it might not work just because the handlebars templates are included into webpack in a separate way, so I just assumed it wouldn't. If this works, it's definitely better than the old way.

astorije:
@xPaw, any idea how we could make it work with existing messages? Sorry @YaManicKill but your suggestion is... ahem, less than ideal 😅
Each `.msg` line has a `data-time`, so we could recompute the timestamp from that and update the DOM for everything. Not ideal either.

The more we move forward with client, the more I think our HTML should be the rendered view of a state we keep in JS. Right now the state *is* the DOM, and this has been biting us quite a bit.

YaManicKill:
Well I guess here's a better (and still css related) options.

Wrap the seconds in a span that is shown/hidden appropriately. So we always generate the format as "HH:mm:ss" but the ":ss" is inside another span. 

The only other way I can think of is going through them all to change the format when we change the setting. Which doesn't feel idea.

xPaw:
Hiding seconds in Dom approach won't work if we want to also support 12H format. Re-rendering should be a couple lines of code.

astorije:
Agreed with @xPaw. Re-rendering is not going to be that expensive, and it's an option you change once and keep forever (or like, until your personal preferences change), not something you'd change every day.

YaManicKill:
Yeah, fine, I guess you are right. Re-rendering all the times when the user changes the settings sounds like the best idea and won't be too difficult. @bews are you happy to add this to the PR? I'll be happy with the PR if that is done.

bews:
Yep, will do that later.

bews:
Done.

xPaw:
So you want to update this PR to support 12h (am/pm) as well?

client/js/options.js
```diff
@@ -80,6 +83,11 @@ settings.on("change", "input, select, textarea", function() {
 			// otherwise, users get notifications for everything
 			return h !== "";
 		});
+	} else if (name === "showSeconds") {
+		var format = options.showSeconds ? constants.timeFormats.msgWithSeconds : constants.timeFormats.msgDefault;
+		chat.find(".msg > .time").each(function() {
+			$(this).text(moment($(this).parent().data("time")).format(format));
```
xPaw: Call the TS helper here, then you don't need duplicate code.

YaManicKill:
> Do you want to update this PR to support 12h (am/pm) as well?

Let's leave that to another PR.

client/js/options.js
bews: > Call the TS helper here, then you don't need duplicate code.

TZ you mean? I can't, since it imports `options` now.

YaManicKill: There's a fix for that. Move the import into the function on the export (so it only gets callled when the function gets called) and then add the import for the helper here. It's a little less efficient, but there should be no cyclical dependency.

bews: done

client/js/options.js
```diff
@@ -62,7 +64,7 @@ settings.on("change", "input, select, textarea", function() {
 		"nick",
 		"part",
 		"quit",
-		"notifyAllMessages",
+		"notifyAllMessages"
```
xPaw: Revert this line.

client/js/libs/handlebars/tz.js
```diff
@@ -1,7 +1,10 @@
 "use strict";
 
 const moment = require("moment");
+const constants = require("../../constants");
 
 module.exports = function(time) {
-	return moment(time).format("HH:mm");
+	const options = require("../../options");
```
xPaw: Did this require outside of the function not work?

YaManicKill: No, because otherwise we have a cyclical dependency, with options requiring this file.

client/js/options.js
bews: done

xPaw:
Could you squash the commits? I will merge after that.

bews:
done

astorije:
Hey @bews, we have free sticker packs for our contributors now!

Issue #1131, opened 2017-05-03
Autoload backlog instead of "show more" button


bews:
History should automatically load when scrolling up. I think it may be easy to implement.

MiniDigger:
duplicate of #207

astorije:
> History should automatically load when scrolling up.

Ahah, famous last words 😄
I did say the same thing then attempted it in https://github.com/thelounge/lounge/pull/1038 then gave up.

PR #1148, opened 2017-05-09
Autoload more history on scrolling

bews:
Works the same way as on Discord.

Related: #1131, #207, #1038

- [x] Find the way to fix scroll bugs

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -795,6 +795,7 @@ kbd {
 #chat .show-more {
 	display: none;
 	padding: 10px;
+	height: 44px;
 	padding-bottom: 0;
 	width: 100%;
 }
@@ -807,7 +808,7 @@ kbd {
 	border-radius: 2px;
 	color: #555;
 	font-size: 12px;
-	height: 34px;
+	height: 100%;
 	line-height: 0;
 	width: 100%;
 }
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -553,7 +553,7 @@ $(function() {
 			lastDate = msgDate;
 		});
 
-		scrollable.find(".show-more-button").prop("disabled", false);
+		scrollable.find(".show-more-button").prop("disabled", false).data("scroll", position).removeClass("hide");
 	});
 
 	socket.on("network", function(data) {
@@ -1027,7 +1027,7 @@ $(function() {
 			setNick(self.closest(".network").data("nick"));
 		}
 
-		var chanChat = chan.find(".chat");
+		var chanChat = chan.find(".chat").off("scroll");
 		if (chanChat.length > 0 && chan.data("type") !== "special") {
 			chanChat.sticky();
 		}
@@ -1038,6 +1038,19 @@ $(function() {
 		}
 
 		focus();
+
+		chanChat.on("scroll", function() {
+			var button = $(this).find(".show-more.show .show-more-button");
+			if ($(this).scrollTop() < 50 && $(this).find(".msg").length >= 100) {
+				if (!button.data("loading")) {
+					button.data("scroll", false).data("loading", true).click();
+				} else if (button.data("scroll")) {
+					$(this).scrollTop(button.data("scroll"));
+				}
+			} else if (button.data("loading")) {
+				button.data("loading", false);
+			}
+		});
 	});
 
 	sidebar.on("click", "#sign-out", function() {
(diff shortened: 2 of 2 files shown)


YaManicKill:
How well does this work? Is it nice and smooth? The code looks really simple, but I'm just a little suspicious about it being so small and simple because @astorije tried this and had some bugs. I've not looked into how he did it, though.

bews:
> How well does this work? Is it nice and smooth?

Yep, I didn't notice any problems.

> I'm just a little suspicious about it being so small and simple because @astorije tried this and had some bugs.

This PR doesn't have any of that bugs:
- #1045 fixed multiple load
- "Show more" button is still here + message count check was added, so /clear works
- `.off("scroll")` was added to fix multiple event listeners

bews:
Wait, found some bugs on mobile.

bews:
Finally fixed scroll bugs :weary:

xPaw:
We've ran into same problem with restoring scroll position when more history is loaded. Your solution is going to be a problem if `more` does not arrive instantly (e.g. slow mobile network), and will produce an unwanted jump (e.g. scroll up and down quickly).

You can see my code here https://github.com/thelounge/lounge/compare/xpaw/infinite-scroll
It uses IntersectionObserver to not deal with scroll position and counting how many messages there are.

bews:
In my instance `more` never arrive instantly (because of CF and PI) and I haven't noticed this before. Is this really an issue?

Anyway, I believe this can be easy fixed by updating scroll position in `more` event.

xPaw:
It should be noticeable on a slow network (think second delay)

bews:
> It should be noticeable on a slow network (think second delay)

I have 2-3 seconds delay here and noticed this only after you mentioned it. Maybe it's because I usually don't scroll much while `more` is loading.

bews:
Wait... I already have this in `more` event, maybe this is the reason I don't see any bugs.

xPaw:
Try triggering the load event and scroll down quickly (e.g. page up then page down), I'd be surprised if that didn't jump you back after more messages loaded.

bews:
> I'd be surprised if that didn't jump you back after more messages loaded.

Be surprised :rofl: 

Also, why are you trying to reimplement the same feature instead of just merging this?

xPaw:
Using proper API for this is better than it checking scroll and counting messages.

bews:
Well, you can't fix this unless you start checking scroll :laughing: So you'll have a proper API with a scroll check :smiley:

xPaw:
I tested this PR, it definitely has it's own issues. Doing a pageup triggers it multiple times and it even gets visibly glitchy. There's a bit unrelated issue where position does not reset correctly when more messages get appended (this bug existed before).

I have fixed the scroll reset in my branch by using `requestAnimationFrame` by the way.

bews:
> Doing a pageup triggers it multiple times and it even gets visibly glitchy.

Wat, it absolutely can't get triggered multiple times, look at the code. I also tested this to be sure. It may be some other bug with pageup (which is already fixed on master IIRC).

xPaw:
> Wat, it absolutely can't get triggered multiple times, look at the code.

I tested whatever you currently had in the current state, and it absolutely did produce glitchy UI.

bews:
I definitely don't have such problem on my instance, so it isn't related to this PR. I'll test current branch again tomorrow.

bews:
Turned out you were right: it triggers multiple times (but only on good connection, so I don't have this on my instance). The reason for this was pageup animation, fixed it.

MaxLeiter:
Tests not passing

bews:
It's because of conflicts. I'll fix them if we agree to merge this instead of #1318.

xPaw:
> Turned out you were right: it triggers multiple times (but only on good connection, so I don't have this on my instance).

And this is precisely why you need to be calm instead of telling me to "look at the code".

bews:
> And this is precisely why you need to be calm instead of telling me to "look at the code".

Yeah, but I still agreed to test it again. It's just hard to believe that tested for months code can have such obvious bugs.

MaxLeiter:
I haven't tested this, but I'd rather this over #1318 due to more widespread browser support. No reason to use a brand new API

xPaw:
> No reason to use a brand new API

The reason would be that new API is completely asynchronous, so it's a performance win.

xPaw:
Didn't mean to close this, damn github ui.

astorije:
I also very much prefer #1318 over this PR for the fact that it leverages native browser interaction, and is properly async. This only applies if we ship the polyfill the spec authors came up with though.
And when it comes to API change... well, let's hope there won't be too many breaking changes 😅

Thanks for your help, @bews!

astorije:
#1318 being in good shape to be merged (upon the official polyfill being fixed), it all seems like we will go for the browser-native solution instead. Thanks for triggering work on this, @bews!

PR #1150, opened 2017-05-09
Add "reload page" button when Lounge fails to start.

bews:
It's hard to reload when in fullscreen mode:

screenshot_20170509-161420

### DIFF
--- a/client/js/loading-slow-alert.js
+++ b/client/js/loading-slow-alert.js
@@ -15,3 +15,7 @@ setTimeout(function() {
 		element.style.display = "block";
 	}
 }, 5000);
+
+document.getElementById("loading-slow-reload").addEventListener("click", function() {
+	location.reload();
+});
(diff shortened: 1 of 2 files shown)


client/js/loading-slow-alert.js
```diff
@@ -15,3 +15,7 @@ setTimeout(function() {
 		element.style.display = "block";
 	}
 }, 5000);
+
+document.getElementById("loading-slow-reload").onclick = function() {
```
xPaw: Use addEventListener.

astorije:
It would be really sweet if it would look like the reload button on the chat page for consistency, but I won't push too much if other @thelounge/maintainers don't like that idea :)

I can't get it to show up right now, but 100%-wide red ribbon that turns green when clicking on it.

YaManicKill:
I'm not sure I'm happy with the idea of it looking the same as that, it would feel very out of place on this screen, and we plan to remove the other one anyway. I think what we have here is nicer. I understand the consistency though.

bews:
> It would be really sweet if it would look like the reload button on the chat page for consistency

I've thought about it too, but it doesn't look good. Also, #1185 PR removes that wide button completely, so I think it's better to keep current one.

client/js/loading-slow-alert.js
astorije: @bews, could you fix that please? :)

bews: done

Issue #841, opened 2016-12-26
Small default font size


xPaw:
Is it just me or is the font sizes we have defined are incredibly small? Default theme seems to define 12px, and morning 13px.

MiniDigger:
I think its just fine.  all web chat clients I use have 12-13 (whatsapp web, skype, telegram), the forums I use have 12px too. 
have you just got a high res monitor? :D

xPaw:
Morning uses Lato font in chat, which is why it increases to 13px. Default theme uses monospace font at 12px, which I do find more legible than the Morning font setup.

astorije:
I find the message a bit small as well, especially on mobile. On laptop, I'm hardly ever bothered but it looks a tiny bit smaller than Slack/Telegram. On mobile however, it's way too small IMO.

astorije:
I definitely would like to see this, especially on mobile where the text is tiny. Marking as up-for-grabs, it should be an easy fix if anyone wants to take a stab at it!

bews:
So, how much font size should be increased? I think 14px is the best option. It looks great on my desktop and mobile, also GitHub uses 14px in comments too :)

astorije:
Consolas looks really big at 14px. We currently have the same size as Telegram on desktop (we are at 12px right now), and increasing to 13px would put us similar to Slack.

For mobile, I'm not sure, I'd need to test.

@xPaw, whaddya think?

bews:
> Consolas looks really big at 14px

This depends on screen resolution, for my display it's perfect.
Maybe we should discuss UI scaling again >_>

> increasing to 13px would put us similar to Slack

Actually Slack has 15px font size:

![2017-04-29_23-00-54](https://cloud.githubusercontent.com/assets/3627488/25558486/5331cbf0-2d30-11e7-940d-bd63c0427fa7.png)


which is a bit **bigger** than Consolas at 14px:

![2017-04-29_23-02-24-2](https://cloud.githubusercontent.com/assets/3627488/25558492/818c0164-2d30-11e7-8b8d-30ca32ec9c09.png)

YaManicKill:
We have custom CSS, it's not difficult. I think the current one is too big.

astorije:
> I think the current one is too big.

You mean you think the current font size is too big already?! Surely not on mobile, where it's super tiny, correct?

[...]

PR #1153, opened 2017-05-09
Bigger font size

bews:
This PR increases font in chat `12px -> 13px` and channel title `14px -> 15px`.

I'm still not sure about this. I believe it should be changed in settings, because it so much depends on user screen size. For example, 13px is small for my desktop, but perfect for mobile.

It was also discussed in #841.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -757,7 +757,7 @@ kbd {
 }
 
 #windows .header .title {
-	font-size: 14px;
+	font-size: 15px;
 }
 
 #windows .header .topic {
@@ -804,7 +804,7 @@ kbd {
 #windows .header .topic,
 .messages .msg,
 .sidebar {
-	font: 12px Consolas, Menlo, Monaco, "Lucida Console", "DejaVu Sans Mono", "Courier New", monospace;
+	font: 13px Consolas, Menlo, Monaco, "Lucida Console", "DejaVu Sans Mono", "Courier New", monospace;
 	line-height: 1.4;
 }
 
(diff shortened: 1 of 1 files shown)


MiniDigger:
> I'm still not sure about this. I believe it should be changed in settings, because it so much depends on user screen size. 

that's why the css mod textfield is the perfect location for it ^^

bews:
> that's why the css mod textfield is the perfect location for it 

It's inconvenient to type in those rules on every device (especially on mobile). Simple select in setting will be nice.

astorije:
Shouldn't this PR change font size in *way* more places in `style.css`?

bews:
> Shouldn't this PR change font size in way more places in style.css?

Everything else is fine, even on my big display. We can also increase font in input area, but that's not very noticeable.

bews:
What do you think about increasing font to 14px only on big displays? `@media (min-width: 1610px)` for example.

MaxLeiter:
Am a fan of that, @bews

bews:
> We can also increase font in input area

> increase font to 14px only on big displays

Done.

astorije:
@bews, are you planning to fix conflicts and rebase this PR on master? :)

bews:
@astorije done

client/css/style.css
```diff
@@ -1961,6 +1961,14 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
 }
 
+@media (min-width: 1610px) {
+	#windows .header .topic,
+	.messages .msg,
+	.sidebar {
+		font-size: 14px;
+	}
+}
+
```
astorije: @xPaw, @YaManicKill, what do you think about this? Aside of that I'm 👍 but I just wanted to double check if you were okay with this or not.

YaManicKill: I'm not sure what you are referring to, the fact that these have different font sizes on a different sized screen? Or what specifically is set to a bigger size? Or what the bigger size is?

astorije: I meant adding a new responsive breakpoint to our CSS. I guess we're fine with that.

PR #1052, opened 2017-04-19
Issue 1019: Show MOTD by default

KlipperKyle:
Resolves https://github.com/thelounge/lounge/issues/1019.

Testing done:  I have built in a development environment on Slackware 14.2 with

```sh
npm install
NODE_ENV=production npm run build
npm start
```

I then opened an Incognito Window of Google Chrome to http://localhost:9000/ and logged into Freenode.  The MOTD was enabled and shown by default.

Google Chrome was packaged from Slackware's extra directory (on DVD), and nodejs came from slackbuilds.org

I still need to figure out how to tweak npm2tgz so that I can build a custom package.

PR #1157, opened 2017-05-09
Display MOTD by default on client

### LINKED ISSUE #1019: [Feature request] Option to show MOTD by default
Currently, the server's MOTD is hidden unless the user checks the box in Options to show the MOTD.

Some servers put useful information for new users in the MOTD such as links to rules and howto pages.  I am currently maintaining a private server for some coworkers, and I do just that.  irc.twit.tv is an example of a public server that posts important information in the server's MOTD (i.e. rules, and if you don't follow their rules, the operators are not happy).

There are some cases where you might really want people to see the MOTD.  It would be nice to have a setting in config.js to make the "show MOTD" option default to on.  (Users can always uncheck it if they want.)

astorije:
Fixes #1019 (again).

This was changed in https://github.com/thelounge/lounge/pull/1052 by @KlipperKyle but this change was lost during the switch to `options` module in #1066.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -13,7 +13,7 @@ const options = $.extend({
 	join: true,
 	links: true,
 	mode: true,
-	motd: false,
+	motd: true,
 	nick: true,
 	notification: true,
 	notifyAllMessages: false,


YaManicKill:
Minor fix for something which was approved previously. Merging.

Issue #577, opened 2016-08-22
Longer nicks overflow on two lines


TheSonAlsoRises:
![tln5exm](https://cloud.githubusercontent.com/assets/8591366/17857067/9ef38dc8-6880-11e6-912d-9c381a30f9f6.png)

Long nicknames are going overflow a second line, which makes reading the conversation harder than it should be.

xPaw:
Should be fixed by #523

astorije:
@TheSonAlsoRises, could you check if this issue still exists on `master`? #523 was merged and hopefully this is not an issue anymore.

YaManicKill:
It does still happen, just tested after #523 merge.

Fwiw, there are very few servers that support names long enough for this. Synirc is the one we both use for that.

I have a small css tweak that fixes it, but it's not perfect. Once [text-overflow: fade](https://developer.mozilla.org/en/docs/Web/CSS/text-overflow) is supported, I can sort this out in a really nice way, but currently it either involves a horrible fade with ::after, or the name just cutting off. Or using JS. Don't want to use any of these solutions.

```css
/* Truncate username */

#chat .message .from {
  overflow: hidden;
  white-space: nowrap;
}

#chat .message .from:after {
    left: calc(10% - 33px);
    content: '';
    position: absolute;
    top: 0;
    width: 30px;
    height: 100%;
    background-image: linear-gradient(to left, rgba(0, 43, 54, 1), rgba(0, 43, 54, 0));
}
```

MiniDigger:
whats even the expected behavior here? should the nick make the column expand? should it be cut/faded? 

personally, I just yell at ppl till the change the size of their nick :D, we even have ban masks that ban nicks longer than 18 chars (I think thats the limit for lounge too) like '"??????????????????*". what I am trying to say with this is that this is really not something you should worry too much about, as its generally considered to be a nogo since other clients have such problems too. 
I personally would be fine with just cutting them off (we need some way to display that they got cut tho), the full name is always visible via the context menu.

YaManicKill:
>whats even the expected behavior here? 

I expect it to fade out, but have a hover that shows the full name of the user. That's exactly the same behaviour as for channel names.

> what I am trying to say with this is that this is really not something you should worry too much about

I'm not worrying about it, I'm saying that it'll be an easy fix once we have the feature that is coming to browsers (currently in working draft) and I have a not-great fix that people can use just now.

> as its generally considered to be a nogo since other clients have such problems too

[...]

PR #1158, opened 2017-05-11
Fade out for long usernames.

bews:
Issue: #577

I get that [text-overflow: fade](https://developer.mozilla.org/en/docs/Web/CSS/text-overflow) is better solution, but who knows when we can use it. So this a temporary fix, until `fade` option will be available.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -911,6 +911,9 @@ kbd {
 	padding-right: 10px;
 	text-align: right;
 	width: 134px;
+	overflow: hidden;
+	white-space: nowrap;
+	position: relative;
 }
 
 #chat .text {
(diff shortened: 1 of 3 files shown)


MaxLeiter:
when you hover does the whole nick show in a tooltip? What about mobile?

bews:
> when you hover does the whole nick show in a tooltip? 

No, is there a need for that? Full nick is visible at right-click menu and whois tab. 

> What about mobile?

Mobile is unaffected by this PR.

YaManicKill:
I'll test this later, but I'm happy with the idea of this in general. My comment about waiting for text-overflow: fade was about me not having the time to implement something. Happy with someone else implementing this instead.

YaManicKill:
Hmm, am I missing something, @astorije? Why did you ask for my review, I've already reviewed this one.

astorije:
That's a good question, probably a click miss 😅
I meant to ask review from @maxpoulin64 or @xPaw on top of both ours, sorry!

Issue #1099, opened 2017-04-28
Bug: after closing private chat lobby becomes an active window


bews:
Expected: switch back to the previous window (Shout behavior).

astorije:
Yeah, we had numerous discussions about this.

The history behind this is this behavior was lost when working on a better client layout. It was introduced as a bug in https://github.com/thelounge/lounge/pull/465 and then fixed (to the current solution) in https://github.com/thelounge/lounge/pull/489.

We attempted to bring back Shout's behavior you mentioned in https://github.com/thelounge/lounge/pull/503, but first we did not agree on this, but also there was a bug when browsing on channels across networks.
The conclusion of this was to defer this to the future "Home tab" described in https://github.com/thelounge/lounge/issues/791 and therefore the current behavior was and will be kept as such until then.

Anyway, a classic and infamous "it's not a bug, it's a feature" 😅

PR #1159, opened 2017-05-11
Restore Shout behavior in private windows

bews:
@astorije you may say "it's a feature" as much as you like, but for me it's a bug, which needs to be fixed :blush: 

So this PR:
- restores Shout behavior and sends user back to the previous channel when closing current one (window.history is used)
- restores Shout's `close` button in private window **only**

Conflicts with #1148, #1117

Related: #1099, #489, #503

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -799,21 +799,38 @@ kbd {
 	width: 100%;
 }
 
+#windows .header .button,
 #chat .show-more-button {
 	background: #f4f4f4;
-	background-image: linear-gradient(#f4f4f4, #ececec);
+	transition: all .3s;
+	background-size: auto 200%;
+	background-position: 0 100%;
+	background-image: linear-gradient(#fefefe, #ececec);
 	border: 1px solid #d7d7d7;
 	border-bottom-color: #b7b7b7;
 	border-radius: 2px;
 	color: #555;
 	font-size: 12px;
+}
+
+#windows .header .button {
+	height: 25px;
+	line-height: 0;
+	padding: 0 10px;
+	float: right;
+	margin: 11px -8px 0 12px;
+}
+
+#chat .show-more-button {
 	height: 34px;
 	line-height: 0;
 	width: 100%;
 }
 
+#windows .header .button:hover,
 #chat .show-more-button:hover {
 	opacity: 1;
+	background-position: 0 0;
 }
 
 #chat .messages {
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -28,6 +28,8 @@ $(function() {
 
 	var ignoreSortSync = false;
 
+	var initHistoryLength = history.length;
+
 	var pop;
 	try {
 		pop = new Audio();
@@ -585,8 +587,10 @@ $(function() {
 	socket.on("part", function(data) {
 		var chanMenuItem = sidebar.find(".chan[data-id='" + data.chan + "']");
 
-		// When parting from the active channel/query, jump to the network's lobby
-		if (chanMenuItem.hasClass("active")) {
+		// When parting from the active channel/query, jump to the previous channel or to the network's lobby
+		if (history.length - initHistoryLength > 0) {
+			history.back();
+		} else if (chanMenuItem.hasClass("active")) {
 			chanMenuItem.parent(".network").find(".lobby").click();
 		}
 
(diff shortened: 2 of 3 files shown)


MaxLeiter:
indifferent on close button, :+1: on previous window

YaManicKill:
So, these are 2 different things.

In terms of the close button on private windows, I don't really care about that. I will just hide it in css. I definitely don't want the button, but if others want it, that's fine.

But in terms of moving back to the previous channel, I am very strongly against this. I hate this functionality in other places. When I close, and it takes me back to the previous channel, it clears the unread in that channel, and that's not always what I want to do. If I want to go to a specific channel, I want to go to a specific channel, and the idea that we will assume that the user wants to go back to the previous one? I really don't like that. Massive 👎 to that.

bews:
> the idea that we will assume that the user wants to go back to the previous one?

This PR focuses on private chats (which are combined with whois atm). When you open a whois window to check user, it's obvious to returnt to previous chat on close. If you don't like it as a general idea - it's fine, but for private chats it's a must have feature.

Also this history clear on every time `init` is called, so channels will be unaffected by this PR in most cases.

YaManicKill:
> This PR focuses on private chats (which are combined with whois atm).

Right, well I feel like you are combining issues.

To have different behaviour for PMs and rooms is confusing. 

Now, I think the issue you are coming across is that to see whois, you have to be moved to the user's pm, which is a bit confusing. I accept this issue. But the best way to solve it is (imho) to show the whois in a modal, rather than a chat window. I think that would solve this issue.

> but for private chats it's a must have feature.

Just a quick point, is that when people tell us something is a "must have feature" it really puts us off interacting with them. Clearly it's not a must have feature, because we have hundreds of people using the lounge on a daily basis, and they haven't left because of it. So please don't be so hyperbolic.

bews:
>  because we have hundreds of people using the lounge on a daily basis, and they haven't left because of it. So please don't be so hyperbolic.

It's really pointless to argue, when someone mentions thousands/millions/billions of users. Just look at M$ with Win10.

YaManicKill:
My point is just that hyperbolic "must have feature" is a term that really annoys people, and is unnecessary. I'm trying to help you not come across as abrasive, and trying to improve the communication, that's all.

dgw:
> the best way to solve it is (imho) to show the whois in a modal, rather than a chat window

I'd say putting `whois` output in a modal is just changing the problem, from unnecessarily creating and focusing a new chat window that must be closed to unnecessarily creating and focusing a modal that must be dismissed. Putting `whois` responses inline in the current chat window would be my ideal solution, but I won't get too far into that here. It's been discussed.

astorije:
I have a few issues with this PR.

First off, being called out as the first statement of a PR makes me want to run away from it and never review it. Please please avoid making things personal, it's not the place, and even if it was, it's not desired and will always play against you.
I have noticed a few of these, and this is my way of asking you once and for all to stay respectful, before things go out of proportion.

The other thing that does not work well here is calling something a bug because it had a different behavior on Shout. This project is The Lounge. It is not Shout. There have been 1263 commits (+244%), 30 releases, 506 PR merged since the fork from Shout. Things are different, and in almost all cases, there was a good reason they are.
We may disagree on the changes, but you have to realize that even if in the end we agreed on a change, starting with "Restore Shout behavior" or calling something a bugfix because it brings back a direction change from Shout, it's the best way of getting 👎 :)

Anyway, I fully agree with @YaManicKill on all points that he exposed.
UX-wise, this is bad. As an example, Slack does that and is subject to the issue @YaManicKill mentioned, as well as moving the scrollbar all the way each time. Terrible user experience.

An interesting idea (because I understand the underlying statement, don't get me wrong) is what Chrome does: if you open a tab and close it without switching to another tab, your focus will go back to the active tab before opening a new tab. If you open a tab, switch tab, go back, then close, your focus will go to the closest tab.
For us, that couldn't work as is because of the unread messages @YaManicKill mentioned, but we could have an hybrid: when opening / closing a private DM, bring back to the previous window if there was no switch between windows and if that was done within some time (like 30 seconds or something), or to the server window otherwise.
This probably sounds like a stretch saying it like that, but it actually feels so natural that most people don't even realize Chrome does that :)

YaManicKill:
Closing this as well.

PR #1160, opened 2017-05-11
Fix annoying unread marker

bews:
This PR stops the unread marker from appearing again and again in discussions.

Before:
ayk6osxbuy

After:
jal2jd46y1

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -469,6 +469,7 @@ $(function() {
 		var msg = buildChatMessage(data);
 		var target = "#chan-" + data.chan;
 		var container = chat.find(target + " .messages");
+		var stopUnreadMarker = false;
 
 		if (data.msg.type === "channel_list" || data.msg.type === "ban_list") {
 			$(container).empty();
@@ -488,7 +489,11 @@ $(function() {
 			prevMsg.after(templates.date_marker({msgDate: msgTime}));
 		}
 
-        // Add message to the container
+		if (chat.find(target).is(".active") && container.children().last().is(".unread-marker")) {
+			stopUnreadMarker = true;
+		}
+
+		// Add message to the container
 		container
 			.append(msg)
 			.trigger("msg", [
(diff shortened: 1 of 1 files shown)


xPaw:
This is not a bug. It has been kept this way because merely having a channel active does not mean you are reading it, and unread marker moves on user actions.

YaManicKill:
Yeah, I'm with @xPaw . We have discussed this lots, and I don't think we should change this.

But that's the beauty of open source software, @bews, you can use this on your fork and keep merging in the newer features on thelounge.

bews:
> This is not a bug. It has been kept this way because merely having a channel active does not mean you are reading it, and unread marker moves on user actions.

You just don't know how active discussions look like.

>  you can use this on your fork and keep merging in the newer features on thelounge.

Already doing this, since most of my PRs get :-1:. Moreover, I'm thinking of creating my own IRC client from scratch. If I continue making changes here, it eventually will lead to more and more conflicts and merging will be impossible.

xPaw:
I very well do know how active channels look like, and in these channels the current behaviour is especially helpful.

And feel free to hack up your own client, nobody stops you.

bews:
Well, I strongly disagree with you. Having this marker popping up **_every 5-10 seconds_** is nothing but annoying.

There may be over ways to fix it (timers maybe), but this fix is the easiest one.

YaManicKill:
> There may be over ways to fix it (timers maybe), but this fix is the easiest one.

YaManicKill:
> Having this marker popping up every 5-10 seconds is nothing but annoying.

As someone who doesn't constantly watch the chat window when I'm doing other things, I just go back when there are unread messages. Removing this would mean that the unread marker isn't in the right place when I go back. Sure, we could base it on whether the chat window has focus, but there are times that we walk away from our computers with the focus still on the chat, and that is the point of this.

This is entirely intended, and we have discussed this plenty.

YaManicKill:
Closing as maintainers agreed not to pull this in.

Issue #1162, opened 2017-05-11
Add option to prefetch images with undefined size


bews:
`prefetchMaxImageSize`  doesn't help in those cases.

astorije:
This was handled differently than an option by #1246.

PR #1165, opened 2017-05-12
Display images with unknown size in prefetch (#1162)

bews:
Images served via scripts usually don't have size header.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -78,6 +78,14 @@ module.exports = {
 	prefetchMaxImageSize: 512,
 
 	//
+	// Display images with unknown size in prefetch.
+	//
+	// @type     boolean
+	// @default  false
+	//
+	prefetchUndefinedImageSize: false,
+
+	//
 	// Display network
 	//
 	// If set to false network settings will not be shown in the login form.
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -64,7 +64,8 @@ function parse(msg, url, res, client) {
 	case "image/gif":
 	case "image/jpg":
 	case "image/jpeg":
-		if (res.size < (Helper.config.prefetchMaxImageSize * 1024)) {
+		if ((res.size === undefined && Helper.config.prefetchUndefinedImageSize) ||
+			(res.size < (Helper.config.prefetchMaxImageSize * 1024))) {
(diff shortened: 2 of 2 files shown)


xPaw:
It looks like we could just read `body.length` if there is no `content-length` header. This will remove the need for a new option.

astorije:
@bews, mind trying @xPaw's solution? I don't think this scenario grants a dedicated option either, it's rather cryptic, even with proper documentation.

astorije:
Closing this to keep an history of the proposed changes somewhere, feel free to try the `body.length` solution in another PR and add a link to this one.

bews:
`body.length` is always `0` for images, so it doesn't work.

xPaw:
Huh? Why would fetched content length be 0?

bews:
It's `0` even for websites. I guess this is the reason:
```javascript
try {
	body = JSON.parse(data);
} catch (e) {
	body = {};
}
```
So `body` is `{}` in most cases.

xPaw:
Try checking before that then.

bews:
It just doesnt' work, `.length` is still `0`.

Also, `JSON.parse(data);` fails every time:
```javascript
<!DOCTYPE html>
^

SyntaxError: Unexpected token < in JSON at position 0
```

xPaw:
Right, didn't notice there's a separate variable called data, that should have the length then. That json parsing is for rare cases when data on the page is json, but I don't know what we achieve by parsing it...

bews:
So, if answer is html, then original `data` has length, but if it's an image - it's still 0. Maybe there should be another way to get size in that cases?

xPaw:
That sounds odd, surely there must be a way.

bews:
It looks like `request.get` doesn't download images at all: it just checks headers. Therefore length is 0.

xPaw:
Looked at the code, it aborts the request if content type doesn't match text/html or application/json.

bews:
lol, looking but not seeing, here's the reason of this:
```javascript
if (!(/(text\/html|application\/json)/.test(res.headers["content-type"]))) {
	res.req.abort();
}
```
soo... should we remove this?

xPaw:
You could certainly try.

bews:
Yep, this works now. Should we completely ignore `content-length` header then?

Issue #1097, opened 2017-04-28
Make nickname grayed out in backlog when user is disconnected


bews:
It will be easy to notice, that user can't read your answer now.

The main problem I see here is grey color. Shades of grey maybe? (fifty, lol)

clandmeter:
This would be an interesting feature for me. I always have all "Show X messages" disabled (i like a clean backlog). It would be nice if we could directly see from the backlog if a user is still available. Maybe add a setting (#85) to set a custom opacity level.

MiniDigger:
started work on this here, now that vue is a thing
 https://github.com/MiniDigger/thelounge/commit/e2fb3670646f6f26229d7f429b07d7d06423b7e7

once I refactored the userlist into a hashset to improve lookup times ill open a PR

PR #1166, opened 2017-05-12
Make nickname grayed out in backlog when user is disconnected

bews:
This PR makes easy to find out if user is disconnected and can't read your answer.

Issue: #1097

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -972,6 +972,11 @@ kbd {
 #chat.colored-nicks .user.color-31 { color: #ff4846; }
 #chat.colored-nicks .user.color-32 { color: #ff199b; }
 
+#chat .user.disconnected,
+#chat.colored-nicks .user.disconnected {
+	color: #999;
+}
+
 #chat .text {
 	padding-left: 10px;
 	padding-right: 6px;
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -558,7 +558,7 @@ $(function() {
 		// Date change detect
 		// Have to use data instaid of the documentFragment because it's being weird
 		var lastDate;
-		var nicks = chat.find("#chan-" + data.chan + " .users").data("nicks");
+		var nicks = chat.find("#chan-" + data.id + " .users").data("nicks");
 		$(data.messages).each(function() {
 			var msgData = this;
 			var msgDate = new Date(msgData.time);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -478,6 +478,7 @@ $(function() {
 			channel.find(".msg .user[data-name='" + message.from + "']").removeClass("disconnected");
 			break;
 		case "part":
+			// Deliberate fall through
 		case "quit":
 			channel.find(".msg .user[data-name='" + message.from + "']").addClass("disconnected");
 			break;
@@ -676,10 +677,12 @@ $(function() {
 		renderChannelUsers(data);
 
 		// Condensed actions support
-		var messages = chat.find("#chan-" + data.id).find(".messages");
-		data.users.forEach(function(user) {
-			messages.find(".user[data-name='" + user.name + "']").removeClass("disconnected");
-		});
+		setTimeout(function() {
+			let messages = chat.find("#chan-" + data.id).find(".messages");
+			data.users.forEach(function(user) {
+				messages.find(".user[data-name='" + user.name + "']").removeClass("disconnected");
+			});
+		}, 500);
 	});
 
 	var options = require("./options");
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -386,9 +386,10 @@ $(function() {
 
 		if (data.type !== "lobby") {
 			var lastDate;
-			$(chat.find("#chan-" + data.id + " .messages .msg[data-time]")).each(function() {
-				var msg = $(this);
-				var msgDate = new Date(msg.attr("data-time"));
+			$(data.messages).each(function() {
+				var msgData = this;
+				var msgDate = new Date(msgData.time);
+				var msg = $(chat.find("#chan-" + data.id + " .messages #msg-" + msgData.id));
 
 				// Top-most message in a channel
 				if (!lastDate) {
@@ -401,6 +402,8 @@ $(function() {
 				}
 
 				lastDate = msgDate;
+
+				updateDisconnectedUsers(channel, msgData);
 			});
 		}
 	}
@@ -465,6 +468,22 @@ $(function() {
 		}
 	}
 
+	function updateDisconnectedUsers(channel, message) {
+		switch (message.type) {
+		case "nick":
+			channel.find(".msg .user[data-name='" + message.from + "']").addClass("disconnected");
+			channel.find(".msg .user[data-name='" + message.new_nick + "']").removeClass("disconnected");
+			break;
+		case "join":
+			channel.find(".msg .user[data-name='" + message.from + "']").removeClass("disconnected");
+			break;
+		case "part":
+		case "quit":
+			channel.find(".msg .user[data-name='" + message.from + "']").addClass("disconnected");
+			break;
+		}
+	}
+
 	socket.on("msg", function(data) {
 		var msg = buildChatMessage(data);
 		var target = "#chan-" + data.chan;
@@ -474,7 +493,7 @@ $(function() {
 			$(container).empty();
 		}
 
-        // Check if date changed
+		// Check if date changed
 		var prevMsg = $(container.find(".msg")).last();
 		var prevMsgTime = new Date(prevMsg.attr("data-time"));
 		var msgTime = new Date(msg.attr("data-time"));
@@ -488,7 +507,7 @@ $(function() {
 			prevMsg.after(templates.date_marker({msgDate: msgTime}));
 		}
 
-        // Add message to the container
+		// Add message to the container
 		container
 			.append(msg)
 			.trigger("msg", [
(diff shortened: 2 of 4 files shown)


bews:
Doesn't work after more history was loaded :scream: I'll try to fix it later

client/css/style.css
```diff
@@ -972,6 +972,11 @@ kbd {
 #chat.colored-nicks .user.color-31 { color: #ff4846; }
 #chat.colored-nicks .user.color-32 { color: #ff199b; }
 
+#chat .user.disconnected,
+#chat.colored-nicks .user.disconnected {
+	color: #999;
```
YaManicKill: This will need to be added into the other themes as well.

client/js/lounge.js
```diff
@@ -488,7 +507,7 @@ $(function() {
 			prevMsg.after(templates.date_marker({msgDate: msgTime}));
 		}
 
-        // Add message to the container
+		// Add message to the container
```
YaManicKill: Nice catch

client/js/lounge.js
```diff
@@ -465,6 +468,22 @@ $(function() {
...
+	function updateDisconnectedUsers(channel, message) {
+		switch (message.type) {
+		case "nick":
+			channel.find(".msg .user[data-name='" + message.from + "']").addClass("disconnected");
+			channel.find(".msg .user[data-name='" + message.new_nick + "']").removeClass("disconnected");
+			break;
+		case "join":
+			channel.find(".msg .user[data-name='" + message.from + "']").removeClass("disconnected");
+			break;
+		case "part":
```
YaManicKill: If you could just add a comment `// Deliberate fall through` just so it's clear to people in the future.

client/js/lounge.js
```diff
@@ -646,7 +672,15 @@ $(function() {
...
 	});
 
-	socket.on("names", renderChannelUsers);
+	socket.on("names", function(data) {
+		renderChannelUsers(data);
+
+		// Condensed actions support
+		var messages = chat.find("#chan-" + data.id).find(".messages");
+		data.users.forEach(function(user) {
+			messages.find(".user[data-name='" + user.name + "']").removeClass("disconnected");
```
YaManicKill: Why do we need this as well as the updating on each message?

bews: As the comment say - it's for future condensed actions. Condensing happens after messages rendering, so colors in that message are not updated. 

To be honest, I don't like this implementation, it might be better to just call `updateDisconnectedUsers(channel, message)` in `setTimeout`

bews: Ah, wait. Since condensed message is regenerating each time some action happens, we need to update users after that. Also, we can't take data from userlist, because it's updated later by this event.

But I think `setTimeout` is still needed here.

client/css/style.css
bews: done

client/js/lounge.js
bews: done

client/js/lounge.js
bews: Added `setTimeout`.

bews: We can remove this code for now and move this code to #759 or whatever PR it will be.

MaxLeiter:
Maybe it's just a personal issue, but I don't know if I can really tell if it's a faded nick or that's just the nick color.

client/js/lounge.js
astorije: @xPaw, do you see any performance concerns here or somewhere else in this PR? Something tells me there will be some overhead, similar to the PR that was looking for all nicks in messages.

astorije:
Yeah, carrying information solely in a color is usually an accessibility issue. Rather than (or on top of) using a color, we could append/prepend http://fontawesome.io/icon/sign-out/ or maybe http://fontawesome.io/icon/circle-thin/ after the nick, but that could look bloated (especially after #759 / #1047).

We could also do a `not-allowed` mouse cursor and/or disable click when disconnected maybe?

How do nick changes look behave? Do they look like the old nick is a disconnected nick?

bews:
> Maybe it's just a personal issue, but I don't know if I can really tell if it's a faded nick or that's just the nick color (in the example image).

Sure, because this was supposed to work with #1096, and default colors sucks.

> we could append/prepend http://fontawesome.io/icon/sign-out/

It's a bad idea. I thought about this too: the original plan was to add • before every normal user (like ~, @, %, +) and remove those marks on disconnect. 

> Do they look like the old nick is a disconnected nick?

Yes.

astorije:
Why is it a bad idea? What are you proposing so we don't use color as the single way to express information?

Re: nick changes appearing as disconnected users, that would be an issue. Not sure how to fix that unless we keep track of nick changes over time, which is far from ideal.

bews:
> Why is it a bad idea? What are you proposing so we don't use color as the single way to express information?

It's the IRC client, so there shouldn't be any additional symbols. And I'm not proposing anything, I'm saying that changing color is enough. Distinctive or not - it is a problem of coloring, rather than the current system. Again, it's works perfect with my colors from #1096.

> nick changes appearing as disconnected users, that would be an issue

I don't see any issue here. Also, tracking will be difficult because of possible reconnections.

> We could also do a not-allowed mouse cursor and/or disable click when disconnected maybe?

Bad idea, leaving channel doesn't always mean disconnect, so `whois` might still work.

astorije:
> It's the IRC client, so there shouldn't be any additional symbols.

That... doesn't make any sense 😅


> I'm saying that changing color is enough. Distinctive or not - it is a problem of coloring, rather than the current system.

It is not enough, it is an accessibility problem. From [WAI](https://www.w3.org/WAI/gettingstarted/tips/designing.html#dont-use-color-alone-to-convey-information): "While color can be useful to convey information, color should not be the only way information is conveyed."

> I don't see any issue here.

disconnected != nick changed.

> Also, tracking will be difficult because of possible reconnections.

Doesn't this "un-gray out" reconnected/rejoined users?

> Bad idea, leaving channel doesn't always mean disconnect, so whois might still work.

I missed that this applied for both users who either left the channels or got disconnected.

---

All in all, accessibility issue aside, I feel like there are a lot of edge cases and room for false positives, and that this design might be too simplistic to handle them.

bews:
> disconnected != nick changed.

Still, you can't convo user by old nickname and `whois` will fail. It's better to keep current way.

> Doesn't this "un-gray out" reconnected/rejoined users?

Yes, but user can reconnect using the old nickname, or using another nickname and than change it to the old one. If will be hard to track all of this. 

> It is not enough, it is an accessibility problem.

Whatever it may be, adding symbols is a bad idea. 

Also, you may not merged my other PRs, but I use them and have nicknames in condensed messages, mentions, actions etc. And I definitely don't want to have additional symbols everywhere.

It may be possible `underline` connected users, but it won't look good IMO.

client/js/lounge.js
```diff
@@ -646,7 +674,17 @@ $(function() {
 		}
 	});
 
-	socket.on("names", renderChannelUsers);
+	socket.on("names", function(data) {
+		renderChannelUsers(data);
+
+		// Condensed actions support
+		setTimeout(function() {
```
xPaw: Condensed actions is not in master, nor does using setTimeout in event based code is a good idea.

xPaw:
I agree with @astorije, using color alone is not good enough, especially with colored nicknames.

Perhaps italic would be better suited here?

client/js/lounge.js
bews: yeah, I can remove that

bews:
Underline:



Italic:

MaxLeiter:
I like italics a lot, but not too sure how I feel about it in nick changes -- the first nick will -always- be italic. Nice overall

bews:
> the first nick will -always- be italic

1) Not always. At first yes, but then user can change nick back to the first one.
2) In backlog all nicknames are updated anyway, so it may save some time by not updating current message only, but meh... It won't save much time IMO. 
3) I made that changes in condensed tpl for the different reason (mostly to avoid bugs ), but here... Dunno, I think it's unnecessary.

astorije:
Can we try with a discreet marker like http://fontawesome.io/icon/circle-thin/ before or after the nick, just to play with it?
Re: italic version, I believe @xPaw meant italics instead of grayed out. Can you show what it'd look like? :)

bews:
> Re: italic version, I believe @xPaw meant italics instead of grayed out. 



> Can we try with a discreet marker

clandmeter:
An alternative to color could be opacity of which the level could be set in settings?

image

bews:
> An alternative to color could be opacity of which the level could be set in settings?

It may look good with dark themes, but definitely not on white background. 
The default design can be easily changed in theme.css like other things.

YaManicKill:
So, regarding the whole colour vs symbol thing.

I don't like the symbol, it's too busy, let's not do that.

But as for the grey colour, I don't have an issue with this.

> using color alone is not good enough, especially with colored nicknames.

This is the thing, the information that this conveys is not something that we **need** to get across. If someone doesn't realise what the grey means, the whole thing still works, it's all good. It's just they don't get the small advantage of knowing the user is not there.

So I wouldn't worry about this.

christer88:
Can we use 'striketrough' on nicknames and text? Will be visible atleast.

astorije:
Because of the state of our code, we believe there is currently no way to do this efficiently. @xPaw pointed out that we need sooner or later to properly send a user list diff instead of the entire list at every update, which would then make this feature much more digestible 😅

We're unfortunately closing this PR in the meantime, but will be happy to re-open it once this has happened, or if we overlooked a better way to do so.

Oh, and when work on this feature gets worked on again, definitely italics 😁

Issue #1151, opened 2017-05-09
When searching a user, results are not clickable


astorije:
This was most likely broken in #856.

When clicking on a user search result, nothing happens. Click does work however when nothing is being searched.

Example on `v2.3.0-rc.1`:

![user-search-bug](https://cloud.githubusercontent.com/assets/113730/25854713/fc3c8012-349e-11e7-8583-ce2dc6b5105a.gif)

PR #1167, opened 2017-05-13
Fix user list search not responding to clicks on search results

astorije:
Fixes https://github.com/thelounge/lounge/issues/1151.

`original` comes from [here](https://github.com/thelounge/lounge/blob/3e54569624f5b276163b76da6afdfa10a20d6616/client/js/lounge.js).


astorije:
Merging, small and straightforward bugfix.

PR #1175, opened 2017-05-18
Add modules for socket events

YaManicKill:
This one is probably the biggest modularising one. It is quite big, but it's really awkward to split this one down anymore, and it works quite nicely as a chunk together.

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -1,0 +1,193 @@
+"use strict";
+
+const $ = require("jquery");
+const templates = require("../views");
+const options = require("./options");
+const utils = require("./utils");
+const sorting = require("./sorting");
+
+const chat = $("#chat");
+const sidebar = $("#sidebar");
+
+module.exports = {
+	buildChannelMessages,
+	buildChatMessage,
+	renderChannel,
+	renderChannelMessages,
+	renderChannelUsers,
+	renderNetworks
+};
+
+function buildChannelMessages(channel, messages) {
+	return messages.reduce(function(docFragment, message) {
+		docFragment.append(buildChatMessage({
+			chan: channel,
+			msg: message
+		}));
+		return docFragment;
+	}, $(document.createDocumentFragment()));
+}
+
+function buildChatMessage(data) {
+	const type = data.msg.type;
+	let target = "#chan-" + data.chan;
+	if (type === "error") {
+		target = "#chan-" + chat.find(".active").data("id");
+	}
+
+	const chan = chat.find(target);
+	let template = "msg";
+
+	if (!data.msg.highlight && !data.msg.self && (type === "message" || type === "notice") && options.highlights.some(function(h) {
+		return data.msg.text.toLocaleLowerCase().indexOf(h.toLocaleLowerCase()) > -1;
+	})) {
+		data.msg.highlight = true;
+	}
+
+	if ([
+		"invite",
+		"join",
+		"mode",
+		"kick",
+		"nick",
+		"part",
+		"quit",
+		"topic",
+		"topic_set_by",
+		"action",
+		"whois",
+		"ctcp",
+		"channel_list",
+		"ban_list",
+	].indexOf(type) !== -1) {
+		template = "msg_action";
+	} else if (type === "unhandled") {
+		template = "msg_unhandled";
+	}
+
+	const msg = $(templates[template](data.msg));
+	const text = msg.find(".text");
+
+	if (template === "msg_action") {
+		text.html(templates.actions[type](data.msg));
+	}
+
+	if ((type === "message" || type === "action") && chan.hasClass("channel")) {
+		const nicks = chan.find(".users").data("nicks");
+		if (nicks) {
+			const find = nicks.indexOf(data.msg.from);
+			if (find !== -1) {
+				nicks.splice(find, 1);
+				nicks.unshift(data.msg.from);
+			}
+		}
+	}
+
+	return msg;
+}
+
+function renderChannel(data) {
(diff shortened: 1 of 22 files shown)


YaManicKill:
I rebased this, so it should be ready to go. There were a couple of small things that I could see that were added to the socket events since I did this PR, so I copied them over to the modules. A second pair of eyes to check I didn't miss something would be good.

client/js/utils.js
```diff
@@ -0,0 +1,140 @@
...
+		let k = new_index - array.length;
+		while ((k--) + 1) {
+			this.push(undefined);
+		}
+	}
+	array.splice(new_index, 0, array.splice(old_index, 1)[0]);
+	return array;
+}
+
+function sortable() {
```
xPaw: Make its own module?

YaManicKill: We probably could do. I could do that in another PR? It'll be easier to move things out of util.js than lounge.js in the future.

client/js/socket-events/open.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+const $ = require("jquery");
+const sidebar = $("#sidebar, #footer");
```
YaManicKill: I forgot to remove this. Will do so soon.

client/js/socket-events/auth.js
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	const input = login.find("input[name='user']");
+	if (input.val() === "") {
+		input.val(storage.get("user") || "");
+	}
+	if (token) {
+		return;
+	}
+	sidebar.find(".sign-in")
```
YaManicKill: Oh yeah, don't need it, will do so.

client/js/utils.js
YaManicKill: Alright, I just did it.

YaManicKill:
@xPaw Updated

YaManicKill:
Re-rebased again. Could someone else review this so we can get this in before we get more conflicts? ;-)

astorije:
Uuuuh, sorry...

screen shot 2017-06-20 at 01 35 38

client/js/render.js
```diff
@@ -0,0 +1,192 @@
+"use strict";
```
astorije: Any chance you could add a new line between `"use strict";` and what comes after in your files?

If this is just too annoying for you, nevermind then, not a big deal :)

client/js/render.js
```diff
@@ -0,0 +1,192 @@
...
+const $ = require("jquery");
+const templates = require("../views");
+const options = require("./options");
+const utils = require("./utils");
+const sorting = require("./sorting");
+
+const chat = $("#chat");
+const sidebar = $("#sidebar, #footer");
```
astorije: This is probably extracted from `lounge.js`, but do we actually want to keep both selectors in there?

client/js/render.js
```diff
@@ -0,0 +1,192 @@
...
+
+module.exports = {
+	buildChannelMessages,
+	buildChatMessage,
+	renderChannel,
+	renderChannelMessages,
+	renderChannelUsers,
+	renderNetworks
+};
```
astorije: Another no-big-deal question: I was under the impression that it was fairly common/standard to add the `exports` statement at the end of the JS files, am I incorrect?

astorije:
I haven't done a full review because of the conflict (sorry!) but I did test this locally and so far no issues.
I'll do a more comprehensive review when you fix the conflict and I promise to look at merging this next so that you don't have to do it again!

client/js/render.js
YaManicKill: So, there are 2 standards, 1 is the add it at the end and 1 is the add it at the beginning.

I tend to add it at the beginning because I like have it front and centre, one of the first things you see when you go in, because it gives you a clear indication what is exported straight away.

client/js/render.js
YaManicKill: Probably not? I'm not sure why it was in it in `lounge.js` so I'll need to figure that out in the first place before I delete the footer.

client/js/render.js
YaManicKill: Sure, but if this is something we want to enforce we should probably see if there's an eslint rule (or write our own) so we don't have to point it out.

YaManicKill:
> Uuuuh, sorry...

Are you? Are you really?

YaManicKill:
Alright, @astorije I have done everything except removing the "#footer". I'll take a look into that now, but you can review the rest whenever you want.

client/js/render.js
astorije: I guess http://eslint.org/docs/rules/padding-line-between-statements would do?

YaManicKill: It doesn't look like "use strict" is one of the options there, but possibly. Let's not do this in this PR anyway, we can do that in another.

client/js/render.js
YaManicKill: Alrighty, so looking and this is for the buttons in the bottom of the sidebar (add server, help, sign out, etc) which are all still handled in lounge.js, so I have removed it from all the plugins (I checked that it didn't need them in each of them, they all seem to just be doing chan stuff).

YaManicKill:
> It's actually pretty much impossible to fully review this. 

Yeah, I know. But I wasn't making a PR per socket event. That would have been dreadful.

Anyway, 🎉 less than 1000 lines in lounge.js now. I'll give it a break before modularising anything else.

But I would like to ask that new code doesn't go in lounge.js unless it has to now.

Issue #1174, opened 2017-05-18
Add option to autoreload on timeout or enable socket reconnect


bews:
Notifications don't get through because of timeouts.

astorije:
I believe this is a duplicate of https://github.com/thelounge/lounge/issues/421. Make sure to include enough details to your issue so we are all on the same page.

Regarding the issue itself, that definitely won't be an option and should just work for everyone. However, as mentioned some days ago in the IRC channel, it is sadly not just about enabling socket.io's reconnect mode. I forget why but feel free to take a stab at it if you feel adventurous 😅

PR #1185, opened 2017-05-25
Enable client socket reconnection

bews:
Changes:
- socket reconnect option -> `true`
- "click to reconnect" button is replaced with "reconnecting" banner
- socket timeout is increased a bit `20000` -> `40000` 

Here how it looks:
timeout_banner

On reconnect "init" event is parsed and channels are re-rendered. 
The input field is still accessible while the banner is displayed (message sending is disabled).

Still testing it, looks good so far.

Related: #1174, #421, #267

### TODO
- [x] disable "load more" button on reconnection
- [x] fix channels sync
- [x] ~~check embed reload on re-rendering (not sure if they are really reloaded)~~ cached
- [x] fix network/channel leave on server restart
- [x] fix "load more" button when connection is lost on "more" event / on 100 loaded messages
-  do something about performance drop
- [x] fix "new messages" marker

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -220,6 +220,32 @@ $(function() {
 	});
 
 	socket.on("init", function(data) {
+		if (!$("#loading").length) {
+			var channels = $.map(data.networks, function(n) {
+				return n.channels;
+			});
+
+			channels.forEach(function(channel) {
+				renderChannel(channel);
+				if (channel.unread > 0 && channel.type === "channel") {
+					var badge = sidebar.find(".chan[data-title='" + channel.name + "'] .badge");
+					badge.text(channel.unread);
+					if (channel.highlight) {
+						badge.addClass("highlight");
+					}
+				}
+			});
+
+			if (sidebar.find(".highlight").length) {
+				toggleNotificationMarkers(true);
+			}
+
+			$("#connection-error").removeClass("display");
+			$("#input").removeAttr("disabled");
+
+			return;
+		}
+
 		$("#loading-page-message").text("Rendering…");
 
 		if (data.networks.length === 0) {
@@ -369,7 +395,7 @@ $(function() {
 
 	function renderChannelMessages(data) {
 		var documentFragment = buildChannelMessages(data.id, data.messages);
-		var channel = chat.find("#chan-" + data.id + " .messages").append(documentFragment);
+		var channel = chat.find("#chan-" + data.id + " .messages").html(documentFragment);
 
 		if (data.firstUnread > 0) {
 			var first = channel.find("#msg-" + data.firstUnread);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -240,8 +240,9 @@ $(function() {
 				toggleNotificationMarkers(true);
 			}
 
-			$("#connection-error").removeClass("display");
-			$("#input").removeAttr("disabled");
+			$("#connection-error").removeClass("shown");
+			$("#submit").removeAttr("disabled");
+			$("#input").data("disabled", false);
 
 			return;
 		}
@@ -804,6 +805,11 @@ $(function() {
 			},
 			"textComplete:hide": function() {
 				$(this).data("autocompleting", false);
+			},
+			keydown: function() {
+				if (event.which === 13 && !event.shiftKey) {
+					event.preventDefault();
+				}
 			}
 		});
 
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -227,12 +227,10 @@ $(function() {
 
 			channels.forEach(function(channel) {
 				renderChannel(channel);
-				if (channel.unread > 0 && channel.type === "channel") {
-					var badge = sidebar.find(".chan[data-title='" + channel.name + "'] .badge");
-					badge.text(channel.unread);
-					if (channel.highlight) {
-						badge.addClass("highlight");
-					}
+				var badge = sidebar.find(".chan[data-title='" + channel.name + "'] .badge");
+				badge.text(channel.unread ? channel.unread : "");
+				if (channel.highlight) {
+					badge.addClass("highlight");
 				}
 			});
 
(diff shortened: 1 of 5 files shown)


client/js/lounge.js
```diff
@@ -369,7 +396,7 @@ $(function() {
 
 	function renderChannelMessages(data) {
 		var documentFragment = buildChannelMessages(data.id, data.messages);
-		var channel = chat.find("#chan-" + data.id + " .messages").append(documentFragment);
+		var channel = chat.find("#chan-" + data.id + " .messages").html(documentFragment);
```
xPaw: This doesn't seem right.

client/js/lounge.js
```diff
@@ -220,6 +220,33 @@ $(function() {
...
+				if (channel.unread > 0 && channel.type === "channel") {
+					var badge = sidebar.find(".chan[data-title='" + channel.name + "'] .badge");
+					badge.text(channel.unread);
+					if (channel.highlight) {
+						badge.addClass("highlight");
+					}
+				}
+			});
+
+			if (sidebar.find(".highlight").length) {
```
xPaw: This doesn't appear to update sidebar. If the channels or networks change while you are offline, this would be out of sync.

client/js/lounge.js
```diff
@@ -220,6 +220,33 @@ $(function() {
...
 
 	socket.on("init", function(data) {
+		if (!$("#loading").length) {
+			var channels = $.map(data.networks, function(n) {
+				return n.channels;
+			});
+
+			channels.forEach(function(channel) {
+				renderChannel(channel);
+				if (channel.unread > 0 && channel.type === "channel") {
```
xPaw: You always need to update badge, if it gets reset to 0 while offline.

xPaw:
This easily could be improved to only send new messages, keep track of latest message is, send it on auth, and on the server only send messages with higher id. This also needs to handle server restarts though.

client/js/lounge.js
bews: On normal load `#chan-" + data.id + " .messages` is always empty, so it's fine. On reconnection this block should be re-rendered, thus `append` was replaced.

bews:
This is strange... After reconnection there is a significant performance drop. And it's even stays that way until page reload. 
Any idea why this can happen?

bews:
So I really don't know what to do with this. When you return to the Lounge tab after ~12 hours, on mobile it becames almost unusable with **very** low FPS. 

Considering that a "refresh on timeout" option would be a better solution until that bug will be fixed. No point to continue working on this PR :disappointed:

client/js/lounge.js
bews: fixed

client/js/lounge.js
bews: fixed

client/js/lounge.js
```diff
@@ -1183,6 +1235,10 @@ $(function() {
 
 	chat.on("click", ".show-more-button", function() {
 		var self = $(this);
+		if ($("#connection-error").is(".shown")) {
```
xPaw: Disable all load more buttons on connection loss.

bews: You mean `.prop("disabled", true)`?

xPaw: Yep

bews: done

MaxLeiter:
@bews is this good to go? want to run on my instance for a while if it is

bews:
> @bews is this good to go? want to run on my instance for a while if it is

Yeah, I use this on my instance too. There are some performance issues (mostly on mobile), but I'm not sure what to do with them.

MaxLeiter:
ah, i'll try on my SE, if it's noticeable i'll be against merging this PR :(

bews:
> if it's noticeable i'll be against merging this PR :(

It is noticeable. But as I said before, problem is not in this PR. If you keep tab open for a few days (hours on mobile), the same performance issues appear. It's just hard to do so because of timeouts.

MaxLeiter:
@bews could this be solved by unloading messages on the client side? (i.e. removing from buffer, requiring the user to click "Show more messages")

bews:
> could this be solved by unloading messages on the client side? (i.e. removing from buffer, requiring the user to click "Show more messages")

It's already works this way.

xPaw:
I took this PR as an inspiration and implemented reconnection here: https://github.com/thelounge/lounge/compare/xpaw/client-reconnection?w=1

Instead of simply redrawing all channel windows, it only draws new ones, and for existing ones it updates the topic (I still need to implement drawing new messages).

I also think I know what the memory leak/slowness was in this PR, as it did not redraw existing channel windows correctly, but simply appended them again. This also breaks `#chan-` ids, if I read the code correctly.

My approach also asks the server to only send new messages since last known message on the client to save traffic, and to ease drawing of new messages. And as a final touch it will force reload the client page if server restarted between connections (server generates a random number on start), this is to help with buffer problems (e.g. last message id going backwards) and Lounge updates.

xPaw:
Closing in favour of #1471, thanks @bews I looked at your PR as a base.

bews:
> I looked at your PR as a base.

Yeah, it might be the best way to deal with this.

> Instead of simply redrawing all channel windows, it only draws new ones, and for existing ones it updates the topic (I still need to implement drawing new messages).

Big :+1: for this, it will definitely fix recursive reconnect bug.

Regarding this PR:

> I also think I know what the memory leak/slowness was in this PR, as it did not redraw existing channel windows correctly, but simply appended them again. 

This is not the case. This is the first obvious reason to think in the situation like this, so I double checked the code, DOM structure and made a lot of tests. Even after ~50 reconnections in a row there no any performance problems.

> This also breaks #chan- ids,

Not sure about this - there was a problem that I fixed in some of my PRs, so maybe this one is expecting fix from another PR, I don't remember. 

> And as a final touch it will force reload the client page if server restarted between connections

This PR handles restarts without page reloading.

xPaw:
> This PR handles restarts without page reloading.

Mine can too, but the the IDs can be inconsistent (network, user, channel, message ids are incremental). And this is also done to handle upgrades.

Issue #1155, opened 2017-05-09
Focus a channel by joining it


realies:
Using the client with lots of servers and channels might be confusing at times, especially when there's no sorting capabilities. Sometimes I try to join a channel that I'm already part of and nothing happens, then I usually do CTRL+F and find it in the list, select it, etc. It would make sense if the channel goes into focus when joining no matter your presence in it. Looking into the code this seems possible by adding the following code after [this line](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L829).

```
if (text.substring(0, 5) === "/join") {
	var cmd = text.trim();

	if (cmd !== "/join") {
		var name = cmd.split(" ").pop();
		var chan = findCurrentNetworkChan(name);

		if (chan.length) {
			chan.click();
			return;
		}
	}
}
```

astorije:
The rationale seems reasonable. If `/join #newchannel` joins + opens, `/join #joinedchannel` should indeed open, for consistency.
I haven't thought through your example though. I'm thinking @xPaw would know the edge cases for this. Would you be OK opening a PR for this afterwards, @realies?

realies:
Thank should be fine @astorije. I would be happy if someone can review the suggested change, there is a trim and if preventing `findCurrentNetworkChan` from being calling if a user types `/join` without a channel.

astorije:
In terms of implementation, I think it would be better for consistency sake to emit this to the server, and the server itself would take care of ignoring the actual `/join` and simply emit an `open` event.

Not sure though, I'm all ears for a better approach.

realies:
From efficiency prespective, the server shouldn't care that the user is trying to effectively change the active channel.

astorije:
Actually, it should, as this impacts the unread notifications. That's why my approach seems to be the easiest to implement and maintain.

realies:
Good point, it will end up signaling the server anyways. I will look into other ways of implementing it in the next few days.

astorije:
Note that `chan.click();` will go to the server anyway, so I'd rather have something consistent with `/join`ing a new channel, unless I'm missing something (@xPaw?)

YaManicKill:
Bear in mind that chan.click() will do everything we need for dealing with unread notifications. And although it goes to the server anyway, it will do 1 less. If we emit join to the server, and then we change channel, surely that's 1 more call to the server than we need.

PR #1189, opened 2017-05-29
Focus a channel by joining it

realies:
#1155

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -829,7 +829,7 @@ $(function() {
 			return;
 		}
 
-		if (text.substring(0, 5) === "/join") {
+		if (text.substring(0, 6) === "/join ") {
 			var cmd = text.trim();
 
 			if (cmd !== "/join") {


MaxLeiter:
Commit message could be a little more descriptive, maybe do a `git commit --amend`?

dgw:
@MaxLeiter The source branch is `patch-1`, so based on that plus the default title "Update lounge.js", it was probably committed from Github's website. No `--amend`ing there, AFAIK…

client/js/lounge.js
```diff
@@ -829,6 +829,20 @@ $(function() {
 			return;
 		}
 
+		if (text.substring(0, 5) === "/join") {
```
xPaw: I think this will fail if you send `/joinabc`

realies: Your point being?

realies: Updated. A cleaner and a more efficient solution might require refactoring the event handler function.

YaManicKill:
I would quite like this to be pulled out of this funciton a bit, having all this code inline doesn't feel ideal. I'm thinking a better way would be to have modules for commands that we want to deal with client side, rather than server side.

realies:
Perhaps the event handler for form submission should be checking if the first character is `/` after which checking if the command has anything to do with the client in a more structured manner.

xPaw:
This should be fine, however we do need to have some proper logic for handling commands on the client, otherwise we will end up adding even more indexofs/substrings to check for commands there.

realies:
@xPaw, agreed. Perhaps if the first character of the input is `/`, switch between a list of valid client-side handled functionalities. Thoughts?

xPaw:
@realies That probably will work, have the function return true if it has been "eaten", and if the function does not return true, send it to the server as usual.

client/js/lounge.js
```diff
@@ -829,6 +829,20 @@ $(function() {
 			return;
 		}
 
+		if (text.substring(0, 6) === "/join ") {
+			var cmd = text.trim();
```
MaxLeiter: `const`?

client/js/lounge.js
```diff
@@ -829,6 +829,20 @@ $(function() {
...
+
+			if (cmd !== "/join") {
+				var name = cmd.split(" ").pop();
```
MaxLeiter: `const`?

client/js/lounge.js
```diff
@@ -829,6 +829,20 @@ $(function() {
...
+				var chan = findCurrentNetworkChan(name);
```
MaxLeiter: `const`?

astorije:
@realies, let us know when you can apply the few changes @xPaw and @MaxLeiter mentioned, it would be very nice if we could get this in :)

realies:
@astorije, will do as soon as possible.

astorije:
Hey @realies, friendly ping? :)

xPaw:
Superseded by #1491.

PR #1190, opened 2017-05-31
Option to force https on prefetch

### LINKED ISSUE #1180: Link prefetch 404 will still get pushed to the client

bews:
This PR:
1. Adds option to rewrite `http://` to `https://` on prefetch to keep `Secure` status
2. Replaces redirects with the final URI
3. Adds thumbnails handling (they haven't even checked for size before)

Also fixes #1180
Conflicts with #1165

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -31,35 +31,46 @@ module.exports = function(client, chan, originalMsg) {
 
 	const link = escapeHeader(links[0]);
 	fetch(link, function(res) {
-		parse(msg, link, res, client);
+		parse(msg, link, res, null, client);
 	});
 };
 
-function parse(msg, url, res, client) {
-	var toggle = msg.toggle = {
+function parse(msg, url, res, res2, client) {
+	let uri = res.uri;
+	let toggle = msg.toggle = {
 		id: msg.id,
 		type: "",
 		head: "",
 		body: "",
 		thumb: "",
-		link: url,
+		link: uri,
 	};
 
 	switch (res.type) {
 	case "text/html":
 		var $ = cheerio.load(res.text);
+		if (res2 === null) {
+			var image =
+				$("meta[property=\"og:image\"]").attr("content")
+				|| $("meta[name=\"twitter:image:src\"]").attr("content")
+				|| "";
+			if (image) {
+				fetch(image, function(response) {
+					parse(msg, url, res, response, client);
+				});
+				return;
+			}
+		}
 		toggle.type = "link";
 		toggle.head = $("title").text();
 		toggle.body =
 			$("meta[name=description]").attr("content")
 			|| $("meta[property=\"og:description\"]").attr("content")
 			|| "No description found.";
-		toggle.thumb =
-			$("meta[property=\"og:image\"]").attr("content")
-			|| $("meta[name=\"twitter:image:src\"]").attr("content")
-			|| "";
-		if (Helper.config.prefetchForceSSL) {
-			toggle.thumb = toggle.thumb.replace(/^http:\/\//, "https://");
+
+		var thumbnail = checkImage(res2);
+		if (thumbnail) {
+			toggle.thumb = thumbnail;
 		}
 		break;
 
@@ -67,11 +78,10 @@ function parse(msg, url, res, client) {
 	case "image/gif":
 	case "image/jpg":
 	case "image/jpeg":
-		if (res.size < (Helper.config.prefetchMaxImageSize * 1024)) {
+		uri = checkImage(res);
+		if (uri) {
 			toggle.type = "image";
-			if (Helper.config.prefetchForceSSL) {
-				toggle.link = toggle.link.replace(/^http:\/\//, "https://");
-			}
+			toggle.link = uri;
 		} else {
 			return;
 		}
@@ -82,6 +92,29 @@ function parse(msg, url, res, client) {
 	}
 
 	client.emit("toggle", toggle);
+}
+
+function checkImage(res) {
+	let uri = false;
+
+	if (res) {
+		switch (res.type) {
+		case "image/png":
+		case "image/gif":
+		case "image/jpg":
+		case "image/jpeg":
+			if (res.size < (Helper.config.prefetchMaxImageSize * 1024)) {
+				if (Helper.config.prefetchForceSSL) {
+					uri = res.uri.replace(/^http:\/\//, "https://");
+				} else {
+					uri = res.uri;
+				}
+			}
+			break;
+		}
+	}
+
+	return uri;
 }
 
 function fetch(url, cb) {
@@ -96,6 +129,7 @@ function fetch(url, cb) {
 			}
 		});
 	} catch (e) {
+		cb(false);
 		return;
 	}
 	var length = 0;
@@ -122,6 +156,7 @@ function fetch(url, cb) {
 			var body;
 			var type;
 			var size = req.response.headers["content-length"];
+			var uri = req.response.request.uri;
 			try {
 				body = JSON.parse(data);
 			} catch (e) {
@@ -133,6 +168,7 @@ function fetch(url, cb) {
 				type = {};
 			}
 			data = {
+				uri: uri.href,
 				text: data,
 				body: body,
 				type: type,
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -58,6 +58,9 @@ function parse(msg, url, res, client) {
 			$("meta[property=\"og:image\"]").attr("content")
 			|| $("meta[name=\"twitter:image:src\"]").attr("content")
 			|| "";
+		if (Helper.config.prefetchForceSSL) {
+			toggle.thumb = toggle.thumb.replace(/^http:\/\//, "https://");
+		}
 		break;
 
 	case "image/png":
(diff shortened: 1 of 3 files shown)


xPaw:
This would need a ruleset like HTTPS everywhere, simply rewriting wont work. What we need to do is proxy images through lounge like github does (camo).

bews:
> simply rewriting wont work

Embed content is the only place which has http stuff. I'll test this of course, but this 5-minute fix should remove all possible unsecure content.

> What we need to do is proxy images through lounge

Yeah, it's a perfect way to do this, but I don't have time for that.

xPaw:
Of course it will fix mixed content, by breaking a lot of websites.

bews:
> by breaking a lot of websites.

How? In worst case scenario thumbnail won't load, that's all.

src/plugins/irc-events/link.js
```diff
@@ -29,7 +29,10 @@ module.exports = function(client, chan, originalMsg) {
 	});
 	chan.pushMessage(client, msg);
 
-	const link = escapeHeader(links[0]);
+	let link = escapeHeader(links[0]);
+	if (Helper.config.prefetchForceSSL) {
+		link = link.replace(/^http:\/\//, "https://");
```
bews: Ah, I see, you mean this line?

bews:
This should fix that, I'll test it tomorrow.

bews:
Rewrite doesn't work if link has 302 redirect. Need to fix this thing too :weary:

bews:
1. Fixed redirects
2. Added thumbnails handling

Should be good now...

bews:
Some websites redirect to http (while having a valid SSL certificate), and... there goes insecure request.
Maybe we should just use CSP to deal with this? Remove that code from server and add `Content-Security-Policy: upgrade-insecure-requests` as meta tag(?).

xPaw:
👎 on this. This will never work as expected.

astorije:
👎 on this PR as well, for reasons @xPaw described above.

If anything, I would head towards not prefetching HTTP content at all, which is becoming less and less of an issue as HTTPS grows every day ([near 60% of pages loaded by Firefox](https://letsencrypt.org/stats/)).
There is also the option of proxying everything through the server, which is not mutually exclusive with not prefetching HTTP content.
But all these are an entirely different story, very different from this PR, so I'm going to close it since we won't be moving forward with this.

@bews, what do you mean by:

> 3. Adds thumbnails handling (they haven't even checked for size before)

bews:
I've fixed all possible bugs and been testing this for a few weeks. And then get :-1: this just because you _**think**_ it won't work as expected, not even looking at the code. Nice work, guys!

Makes me wonder what I'm doing here.

xPaw:
It's not our personal opinion exclusively, it's been already proven that you can't simply rewrite http to https and expect it to work.

There's a reason why HTTPS Everywhere exists, and has a *massive* amount of rule sets. There's also a reason why GitHub (and other services) proxy images through their own servers to get SSL.

bews:
Again: site previews are still loaded from http, no changes here. Even if site does not support https at all, preview still work.

astorije:
> And then get :-1: this just because you _**think**_ it won't work as expected, not even looking at the code. Nice work, guys!

This is not true. I did look at the code, and we have given you the reasons of our 👎 before doing so. It's not some despotic move against you personally, obviously :)
While I am very thankful and appreciative of you trying to help out on this project, you cannot expect us to blindly merge your PRs if we do not agree with them. This project has a vision, that you may have missed, and sometimes features proposed do collide with it, very much like this one does.
In this case, we do believe there is no need for such option considering the reality of HTTP/HTTPS link handling out there.

> Again: site previews are still loaded from http, no changes here. Even if site does not support https at all, preview still work.

I'm confused, isn't the goal of this PR to (optionally) force HTTPS on all prefetch, and refuse prefetching if the HTTPS version is not present?

bews:
> I did look at the code

It doesn't look like true, otherwise you would not ask such questions :smirk_cat: 

> what do you mean by:
> Adds thumbnails handling (they haven't even checked for size before)

Exactly this, look at the code. I'm talking about `prefetchMaxImageSize` check.

> isn't the goal of this PR to (optionally) force HTTPS on all prefetch, and refuse prefetching if the HTTPS version is not present?

Yes, but site previews are fetched by server, only parsed info is sent to the client. So there no need to rewrite anything and this behaviour is not affected by this PR.
Site thumbnails are another story, they are still loaded by client and here goes http rewrite and blocks. Thus, like I said in previous comment: in worst case scenario thumbnail won't load, that's all.

> This project has a vision, that you may have missed, 

I don't know what is your vision, but in most of my PRs I try to implement very basic features, and other IRC clients/messengers already have them. While my implementation may not be perfect, it's not the reason to throw out my PRs completely. With this PR ideal solution is to proxy all images through Lounge, but I don't have time to implement this, so current way is a **temporary** solution until the correct one will be done.

astorije:
> It doesn't look like true, otherwise you would not ask such questions 😼

You know, without the snarky comments, people would be more willing to spend time on your PRs. I know that, personally, you've called me out and been unpleasant a few times, which is enough to not make these PRs a priority in my life. It may seem petty, but that's one way of prioritizing the gazillions of things I need to do in my free time.
That being said...

> I'm talking about `prefetchMaxImageSize` check.

Oh so you mean before your PR, we would check the size of images against this option when loading a picture preview, but not the thumbnails in website previews?
If so, then that's a great addition, and I don't see any reason why we would not merge this into `master` for the next release. Feel free to move it to its own PR since that's something entirely different and we'll most likely merge it very soon.

> Yes, but site previews are fetched by server, only parsed info is sent to the client. So there no need to rewrite anything and this behaviour is not affected by this PR.
> Site thumbnails are another story, they are still loaded by client and here goes http rewrite and blocks. Thus, like I said in previous comment: in worst case scenario thumbnail won't load, that's all.

That makes sense, but this small impact makes it even more unreasonable to have an option for it + some complex HTTPS handling logic behind that, like @xPaw said, is most likely to fail because of the reality of the Internet (and why HTTPS Everywhere is a complex project).

If anything, I would rather block HTTP images rather than try to augment them to HTTPS which is bound to fail on the short/long term.
Until we proxy all images through the server, @xPaw, would you be okay *not* fetching HTTP images on the client? That would mean no thumbnails and no picture previews at all for not HTTPS content. Yes, it's a breaking change, but it's for the better for users' privacy, and with 60% (according to Firefox) / 70% (according to Google) of all current Internet traffic + most of the top websites being HTTPS, it's likely that few images would be affected.
I wouldn't make this a setting though, see how it goes, and if people are pissed off, make it a client setting :D

> in most of my PRs I try to implement very basic features

The best thing you can do is to open an issue on GitHub about what you are about to implement. When you open a PR with no earlier discussion, work hard on it, and refuse change requests project owners make, you have to accept that this work will not be merged. This is sad on both ends, but this leaves us no other option.
So, open a ticket early, see if owners agree to it and/or have existing ideas about it, and then it's waaay more likely to be merged.

> it's not the reason to throw out my PRs completely.

But here is the thing though: though we have closed a couple of your PRs, we have not thrown them out completely. In fact, there is some acceptable and some unacceptable, and maintainers have asked you if you can split your work into multiple PRs so we can at least quickly merge the good parts, and you have refused that, again leaving us no other options ¯\\\_(ツ)\_/¯

Most of the time, it's very tricky for reviewers and maintainers when a single PR does multiple unrelated things, and in practice, this gets times less chances to be merged. I don't like to play that card usually, but [I have been involved in enough open source projects](https://github.com/astorije?utf8=%E2%9C%93&tab=repositories&q=&type=fork) to assure you that this is rather true in the FOSS world :)

bews:
> So, open a ticket early, see if owners agree to it and/or have existing ideas about it,

I do this changes primarily for myself (and just share here) to make Lounge barely usable :pensive:. So I usually don't have time to open a ticket and wait for the reaction - I need that done right now. 
And if it's not some critical and totally must have feature :rofl:, then yes, I open an issue and so on.

> split your work into multiple PRs so we can at least quickly merge the good parts

Those "good parts" are just side effects of my PRs, I never intended to do them, therefore I have absolutely no motivation to make additional PR just with them. `prefetchMaxImageSize` check fix is the same thing - I wanted to keep rewrite code in one place, so there was no choice :pensive:

> I would rather block HTTP images rather than try to augment them to HTTPS

This was the first version of this PR, but the result was unsatisfactory: ~50% of all content in my daily log was blocked. So I've added a rewrite, and now ~95% is displayed - proof of the efficiency of simple rewriting.

Also, `cheerio` sometimes return site thumbnails with `http`, while site is loaded with `https` (https://www.pixiv.net/ is a good example). Rewriting from this PR fixes this too :smiley:

astorije:
> I do this changes primarily for myself (and just share here) to make Lounge barely usable :pensive:. So I usually don't have time to open a ticket and wait for the reaction - I need that done right now. 

Then I'm afraid you are missing the point of open source software. It's very nice of you to share PRs of what you did to fulfill your *personal* preferences, but you simply cannot be upset that the project *owners* do not accept them all blindly 😉

> And if it's not some critical and totally must have feature :rofl:, then yes, I open an issue and so on.

Well, if it's a totally must-have feature (another term that's usually not too welcome in open source, by the way...), it's actually probably something that *does* need discussion in an issue, so again I think you are looking at it the wrong way.

> Those "good parts" are just side effects of my PRs, I never intended to do them, therefore I have absolutely no motivation to make additional PR just with them.

This is rather sad... But oh well, I guess that's where community effort will help.

PR #1191, opened 2017-05-31
Update node versions for travis

YaManicKill:


YaManicKill:
Currently assuming we will keep node 4 for just now. If we want to remove it, we can do that later.

.travis.yml
```diff
@@ -29,7 +29,7 @@ deploy:
   api_key:
     secure: I9iN31GWI+Mz0xPw81N7qh1M6uidB+3BmiPUXt8QigX45zwp9EhvfZ0U/AIdUyQwzK2RK1zLRQSt+2/1jyeVi+U+AAsRRmaAUx8iqKaQPAkPnQtElolgRP04WSgo7fvNejfM7zS939bQNKG3RlSm04yPgu+ke2igf799p2bpFe2LtyoEeIiUfrUkBiMSpMguN9XF8a7jqCyIouTKjXHR24RmzJ9r7ZoMV27yQauS7XlD81bontzNRZxTytDKdJpZ+sxGIT9mbbtM4LUFX8MeNe3p/bjWavEhrO0ZIpkbOfS/L/w1375YDoNPXxCs288lnGUH+NbGNAEfn+BTz8cmUp7jI7QWR/kNACPeopdAX4OdZxT8wfQcfQZrfCuSpKciOMC7vGgPpQqjQ61t1RKcKs9VUnwC0SwWjyo8LlzkFKnP1ks0eDGYsSoPLdpC9+76UmePkQdxMhscO8TOgkOCcsTMLiyt6ABGOGKu2iE5SsjUYtPiSiRzSBAQENoO560+xBSVTKwqvvhzUAIt4AuAQSgsFjAylDdyzKoObHX12hBdALrqSOOSVwwIQ5/jTgNAsilURHo7KPD407PhRnLOsvumL0qg4sr9S1hjuUKnNla5dg9GY8FVjJ+b2t0A2vgfG1pR1e3vrJRXrpkfRorhmjvKAk2o5you5pQ1Itty7rM=
   on:
-    node: 7
+    node: 8
```
astorije: @YaManicKill, did you happen to check if there were breaking changes around `npm publish`? FYI, last time we bumped from v4 to v7 because we *had to* so I played a bit with it before making the switch. This was the rationale, [from this commit](https://github.com/thelounge/lounge/commit/552fa3fae84dd7c75315dffe6e0bc28afabc14ae):

> - It uses the `prepublishOnly` hook added in npm v4. Since this hook is not available prior to that, deployment to npm from Travis has to be done on the Node.js v7 environment.

Sorry if I look conservative/paranoid, it's just that a breaking change here would be pretty bad as we would most likely not notice it at all until the moment of the next release or, worse, after the release is done (with a broken release). Something like that happened to me once, it was not fun :/

YaManicKill: Well, it's important to note that the breaking changes for npm publish aren't in node, but in npm. However, 8 does bump npm to npm 5. Changes with [npm publish](https://github.com/npm/npm/releases/tag/v5.0.0) are:

* New publishes will now include both sha512 and sha1 checksums.
* prepublishOnly now runs before the tarball to publish is created, after prepare has run.

The first won't really affect us as it's just adding something.

Changing the order of prepare -> prepublishOnly to prepublishOnly -> prepare shouldn't affect us because we don't have a prepare action. So we should be good.

YaManicKill:
I'll let you merge, @astorije, once you are happy with my response.

Issue #1192, opened 2017-05-31
Channel restrictions


ThoreKr:
We currently have the need to be able to restrict the client to join only channels matching a certain pattern.
To make to configuration as individual, and usable for others, as possible support for regular expressions would be cool.
I already asked for it on freenode, but maybe this is the better place for such a discussion.

xPaw:
Closing for the same reason the PR was closed:

> I'll close this as we all agree that doing this in the client is not the right place. Channel restrictions should be done on the IRC server.

PR #1193, opened 2017-05-31
Add support for channel restrictions

ThoreKr:
Directly related to #1192

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -370,6 +370,35 @@ module.exports = {
 		primaryKey: "uid"
 	},
 
+	// Restrict to only certain channels
+	//
+  // @type     object
+	// @default  {}
+	//
+	restrict: {
+		//
+		// Enable channel restrictions.
+		//
+		// @type     boolean
+		// @default  false
+		//
+		enable: false,
+
+		// Pattern for channels to match
+		//
+		// @type 	string
+		// @default ""
+		//
+		pattern: "",
+
+		// Message to return to the user when trying to join a channel not matching the pattern, supports regular expressions
+		//
+		// @type  string
+		// @default "This client ist restricted to certain channels."
+		//
+		restrictMessage: "This client ist restricted to certain channels."
+	},
+	
 	// Extra debugging
 	//
 	// @type     object
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -372,7 +372,7 @@ module.exports = {
 
 	// Restrict to only certain channels
 	//
-  // @type     object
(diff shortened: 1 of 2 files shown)


defaults/config.js
```diff
@@ -370,6 +370,35 @@ module.exports = {
...
+		// @default ""
+		//
+		pattern: "",
+
+		// Message to return to the user when trying to join a channel not matching the pattern, supports regular expressions
+		//
+		// @type  string
+		// @default "This client ist restricted to certain channels."
+		//
+		restrictMessage: "This client ist restricted to certain channels."
```
MaxLeiter: is* restricted

defaults/config.js
```diff
@@ -370,6 +370,35 @@ module.exports = {
 		primaryKey: "uid"
 	},
 
+	// Restrict to only certain channels
+	//
+  // @type     object
```
MaxLeiter: indentation

ThoreKr:
No idea what atom did with the indentation again, but it should be fixed now

xPaw:
Basically what you did is simply not create a UI window for channel, but on the server, you're in the channel and it will send you messages and other stuff from that channel.

ThoreKr:
Well, that's an issue indeed.
It's not possible to control the server and changing it isn't possible due to legacy reasons.
The only way to prevent users from joining certain channels would be in the client.

astorije:
> I also have a hard time grasping why should it be client's job to filter channels.

Client? Isn't this PR trying to do it on the server (though I totally get the first part of your comment)? Do you mean it should be the job of `irc-framework`?

> It's not possible to control the server and changing it isn't possible due to legacy reasons.

What do you mean by that?

ThoreKr:
Oh, you were reffering to the lounge server-side?
My bad, messed that up with considering lounge just as a client and the server as the actual IRC-Server.
Ideally it would limit the channel access for users in that part of the code, which would not be executed in the browser, but in the lounge-server-side.

xPaw:
> Client? Isn't this PR trying to do it on the server 

Client as in an IRC *client*, and server as in an IRC server.

YaManicKill:
I'm not sure how I feel about this...

I definitely get the issue you are trying to get around, but I'm not sure the lounge is the best place for this restriction, to be honest.

I guess we do already have a way to lock the network that the client can connect to, but ... I'm not sure.

xPaw:
I'll close this as we all agree that doing this in the client is not the right place. Channel restrictions should be done on the IRC server.

PR #1194, opened 2017-06-01
Lazily load user list in channels on init, keep autocompletion sort on server

xPaw:
Less bandwidth and processing on initial page load.

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -144,7 +144,7 @@ function renderChannelMessages(data) {
 function renderChannelUsers(data) {
 	const users = chat.find("#chan-" + data.id).find(".users");
 	const nicks = data.users
-		.concat()
+		.concat() // Make a copy of the user list, sort is applied in-place
 		.sort((a, b) => b.lastMessage - a.lastMessage)
 		.map((a) => a.nick);
 
@@ -179,7 +179,13 @@ function renderNetworks(data) {
 			channels: channels
 		})
 	);
-	channels.forEach(renderChannel);
+	channels.forEach((channel) => {
+		renderChannel(channel);
+
+		if (channel.type === "channel") {
+			chat.find("#chan-" + channel.id).data("needsNamesRefresh", true);
+		}
+	});
 
 	utils.confirmExit();
 	sorting();
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -81,7 +81,7 @@ function buildChatMessage(data) {
 		renderPreview(preview, msg);
 	});
 
-	if ((type === "message" || type === "action") && chan.hasClass("channel")) {
+	if ((type === "message" || type === "action" || type === "notice") && chan.hasClass("channel")) {
 		const nicks = chan.find(".users").data("nicks");
 		if (nicks) {
 			const find = nicks.indexOf(data.msg.from);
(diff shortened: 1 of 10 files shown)


src/models/chan.js
```diff
@@ -95,6 +95,7 @@ Chan.prototype.getMode = function(name) {
 
 Chan.prototype.toJSON = function() {
 	var clone = _.clone(this);
+	clone.users = [];
```
YaManicKill: Maybe a quick comment here just mentioning why we do this. I can see it would be confusing for people coming to it.

xPaw:
Don't merge it, it breaks auto completion (sorting) as users will be empty.

xPaw:
I have fixed auto completion!

client/js/render.js
```diff
@@ -145,22 +145,10 @@ function renderChannelMessages(data) {
...
-
-	for (const i in data.users) {
-		nicks.push(data.users[i].nick);
-	}
-
-	nicks = nicks.sort(function(a, b) {
-		return (oldSortOrder[a] || Number.MAX_VALUE) - (oldSortOrder[b] || Number.MAX_VALUE);
-	});
+	const nicks = data.users
+		.concat()
```
YaManicKill: This does nothing, unless I'm missing something.

client/js/render.js
```diff
@@ -145,22 +145,10 @@ function renderChannelMessages(data) {
...
+		.sort((a, b) => b.lastMessage - a.lastMessage)
```
YaManicKill: I think I must have missed the point of renderChannelUsers. Is it not for the users for the sidebar? Because it looks like we are sorting based on when they last sent a message, which wouldn't make sense for the sidebar. So is this for the autocompletion?

src/models/chan.js
YaManicKill: Yeah, still want this comment.

client/js/render.js
xPaw: Make a copy of the object, because it sorts in-place.

client/js/render.js
xPaw: This is sorting auto completion, yes.

client/js/render.js
YaManicKill: Oh. Well that's a very confusing way of doing it. I thin kthe standard way to do it is with slice(0), isn't it?

xPaw: There's no difference. i'll add a comment here

xPaw:
Fixed a case where last message time would be reset when NAMES event is received. This change should also improve GC performance as we will keep `User` objects instead of creating new ones for every user.

client/js/render.js
```diff
@@ -83,7 +83,7 @@ function buildChatMessage(data) {
 		text.html(templates.actions[type](data.msg));
 	}
 
-	if ((type === "message" || type === "action") && chan.hasClass("channel")) {
+	if ((type === "message" || type === "action" || type === "notice") && chan.hasClass("channel")) {
```
astorije: Why did you need this addition? Sorry, not questioning, I'm just not seeing it 😅.

xPaw: Notices are processed in same callback on the server as normal messages and actions. So this change was needed to match ordering correctly.

xPaw:
@astorije Commits are better left without squashing.

PR #1195, opened 2017-06-01
Do not store unnecessary information in user objects

xPaw:
This PR would will reduce memory and bandwidth use. I also found inconsistent `name`/`nick` references that could cause insignificant bugs.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -417,7 +417,7 @@ $(function() {
 		nicks = [];
 
 		for (i in data.users) {
-			nicks.push(data.users[i].name);
+			nicks.push(data.users[i].nick);
 		}
 
 		nicks = nicks.sort(function(a, b) {
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -7,13 +7,21 @@ module.exports = User;
 function User(attr, prefixLookup) {
 	_.defaults(this, attr, {
 		modes: [],
+		mode: "",
 		nick: ""
 	});
 
 	// irc-framework sets character mode, but lounge works with symbols
 	this.modes = this.modes.map(mode => prefixLookup[mode]);
 
-	// TODO: Remove this
-	this.name = this.nick;
-	this.mode = (this.modes && this.modes[0]) || "";
+	if (this.modes[0]) {
+		this.mode = this.modes[0];
+	}
 }
+
+User.prototype.toJSON = function() {
+	return {
+		nick: this.nick,
+		mode: this.mode,
+	};
+};
--- a/src/plugins/irc-events/names.js
+++ b/src/plugins/irc-events/names.js
@@ -10,7 +10,12 @@ module.exports = function(irc, network) {
 			return;
 		}
 
-		chan.users = data.users.map(u => new User(u, network.prefixLookup));
+		chan.users = data.users.map(user => {
+			return new User({
+				nick: user.nick,
+				modes: user.modes,
+			}, network.prefixLookup);
+		});
 
 		chan.sortUsers(irc);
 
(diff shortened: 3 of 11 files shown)


src/models/user.js
```diff
@@ -7,13 +7,21 @@ module.exports = User;
...
 
-	// TODO: Remove this
-	this.name = this.nick;
-	this.mode = (this.modes && this.modes[0]) || "";
+	if (this.modes[0]) {
+		this.mode = this.modes[0];
+	}
 }
+
+User.prototype.toJSON = function() {
```
xPaw: Only send needed variables to client.

src/plugins/irc-events/names.js
```diff
@@ -10,7 +10,12 @@ module.exports = function(irc, network) {
 			return;
 		}
 
-		chan.users = data.users.map(u => new User(u, network.prefixLookup));
+		chan.users = data.users.map(user => {
```
xPaw: `names` event contains hostmasks, indents, which we have no use for.

src/models/user.js
```diff
@@ -7,13 +7,21 @@ module.exports = User;
...
 	_.defaults(this, attr, {
 		modes: [],
+		mode: "",
 		nick: ""
 	});
 
 	// irc-framework sets character mode, but lounge works with symbols
 	this.modes = this.modes.map(mode => prefixLookup[mode]);
 
-	// TODO: Remove this
```
xPaw: 😄

YaManicKill: https://media.giphy.com/media/Kh296M49Zr1Sg/giphy.gif

client/js/lounge.js
```diff
@@ -417,7 +417,7 @@ $(function() {
 		nicks = [];
 
 		for (i in data.users) {
-			nicks.push(data.users[i].name);
+			nicks.push(data.users[i].nick);
```
YaManicKill: So, was name just unnecessary? We were mapping it to nick and using it for ... what reason?

src/plugins/irc-events/names.js
YaManicKill: Hostmasks might be useful for automated banning, etc. I guess no point in keeping it in if we don't use it just now, though.

test/models/chan.js
```diff
@@ -32,9 +32,7 @@ describe("Chan", function() {
 		};
 
 		var getUserNames = function(chan) {
-			return chan.users.map(function(u) {
-				return u.name;
-			});
+			return chan.users.map(u => u.nick);
```
YaManicKill: Can we stick () around "u" because it's one of those things like doing {} round a 1 line if. It's not "wrong" but it can stop us having problemsi n the future if we add stuff to it.

xPaw: `users` is an array and `map` can only call with a single element.

client/js/lounge.js
xPaw: I don't remember to be honest. But I added a TODO when I did the irc-fw conversion.

test/models/chan.js
YaManicKill: Not true: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/map?v=control

Also, even if it can physically only happen with 1, there's no reason not to have them, as it just makes things safer. Just like why we do ';' and {} around things that don't technically need them. It's to help us, because we make mistakes.

xPaw: My bad about the map. But I would leave this discussion for another time (e.g. when we update eslint as I have a branch with extra rules), as existing code in repo already uses the same style.

YaManicKill: Fine, but I'm not letting this go :-P

PR #1196, opened 2017-06-01
Fix invalid prefetch thumbnail url getting pushed to client

### LINKED ISSUE #1180: Link prefetch 404 will still get pushed to the client

Raqbit:
Fixes #1180

When prefetching a page, thelounge will first fetch the actual image before pushing the (possibly invalid) url to the client. When it comes across any type of error it'll simply omit the thumbnail.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -51,13 +51,17 @@ function parse(msg, url, res, client) {
 		toggle.type = "link";
 		toggle.head = $("title").text();
 		toggle.body =
-			$("meta[name=description]").attr("content")
-			|| $("meta[property=\"og:description\"]").attr("content")
-			|| "No description found.";
+				$("meta[name=description]").attr("content")
+				|| $("meta[property=\"og:description\"]").attr("content")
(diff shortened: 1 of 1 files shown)


src/plugins/irc-events/link.js
```diff
@@ -51,13 +51,17 @@ function parse(msg, url, res, client) {
...
+				|| "No description found.";
 		toggle.thumb =
-			$("meta[property=\"og:image\"]").attr("content")
-			|| $("meta[name=\"twitter:image:src\"]").attr("content")
-			|| "";
+				$("meta[property=\"og:image\"]").attr("content")
+				|| $("meta[name=\"twitter:image:src\"]").attr("content");
+		fetch(toggle.thumb, function(prefetchRes) {
+			if (prefetchRes.error) {
+				toggle.thumb = "";
```
bews: This won't work due asynchronous nature of `fetch`.

bews:
Anyway, I've already fixed this in #1190

Raqbit:
Right, should've thought about that🤔

PR #1197, opened 2017-06-01
Correctly configure client socket transports

### LINKED ISSUE #848: Unable to use the websocket transport
1. Edit ~/lounge/config.js and set

```
transports: ["websocket"],
```

2. After restarting the server, the client just keeps reloading.

I had to edit `client/js/lounge.js` and add

```
transports: ['websocket'],
```

in the `io()` call.

Browser: Google Chrome (Win) Version 55.0.2883.87 m (64-bit)

xPaw:
Fixes #848. This will match server/client transport configuration (e.g. websocket only, or polling only, or websocket priorioty over polling).

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -25,7 +25,7 @@
 	<meta name="theme-color" content="#455164">
 
 	</head>
-	<body class="signed-out {{#if public}}public{{/if}}">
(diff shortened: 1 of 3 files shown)


xPaw:
It's relatively standard to pass through settings through data attributes. We can't put in a script tag as we allow no inline JS.

Issue #1139, opened 2017-05-05
Overhaul login code to generate unique session for each login


xPaw:
This is required for #1124 and #986.

It probably needs to be a list of objects, which will contain token, and in the future, push subscription for that token. This way will allow us to send push notifications for browsers that are currently not logged in. We would need to keep track of which browsers/tokens are currently active.

PR #1199, opened 2017-06-02
Implement multiple sessions for users

### LINKED ISSUE #712: Client token timing attacks
The way user tokens are implemented are subject to timing attacks, we search all users and compare each token using `===`.

xPaw:
Initial work for #1139. Use stored username on client to match token instead of iterating every user on the server

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -99,7 +99,7 @@
 							</div>
 							<div class="col-xs-12">
 								<label class="remember">
-									<input type="checkbox" name="remember" id="sign-in-remember" checked>
+									<input type="checkbox" name="remember" value="on" id="sign-in-remember" checked>
 									Stay signed in
 								</label>
 							</div>
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -562,8 +562,12 @@ $(function() {
 	});
 
 	sidebar.on("click", "#sign-out", function() {
+		socket.emit("sign-out", storage.get("token"));
 		storage.remove("token");
-		location.reload();
+
+		if (!socket.connected) {
+			location.reload();
+		}
 	});
 
 	sidebar.on("click", ".close", function() {
--- a/client/js/socket-events/sign_out.js
+++ b/client/js/socket-events/sign_out.js
@@ -1,0 +1,9 @@
+"use strict";
+
+const socket = require("../socket");
+const storage = require("../localStorage");
+
+socket.on("sign-out", function() {
+	storage.remove("token");
+	location.reload();
+});
--- a/src/client.js
+++ b/src/client.js
@@ -10,6 +10,7 @@ var Msg = require("./models/msg");
 var Network = require("./models/network");
 var ircFramework = require("irc-framework");
 var Helper = require("./helper");
+const UAParser = require("ua-parser-js");
 
 module.exports = Client;
 
@@ -81,12 +82,6 @@ function Client(manager, name, config) {
 
 	var client = this;
 
-	if (client.name && !client.config.token) {
-		client.updateToken(function(token) {
-			client.manager.updateUser(client.name, {token: token});
-		});
-	}
-
 	var delay = 0;
 	(client.config.networks || []).forEach((n) => {
 		setTimeout(function() {
@@ -94,6 +89,10 @@ function Client(manager, name, config) {
 		}, delay);
 		delay += 1000;
 	});
+
+	if (typeof client.config.sessions !== "object") {
+		client.config.sessions = {};
+	}
 
 	if (client.name) {
 		log.info(`User ${colors.bold(client.name)} loaded`);
@@ -286,34 +285,51 @@ Client.prototype.connect = function(args) {
 	client.save();
 };
 
-Client.prototype.updateToken = function(callback) {
-	var client = this;
-
-	crypto.randomBytes(48, function(err, buf) {
+Client.prototype.generateToken = function(callback) {
+	crypto.randomBytes(48, (err, buf) => {
 		if (err) {
 			throw err;
 		}
 
-		callback(client.config.token = buf.toString("hex"));
+		callback(buf.toString("hex"));
 	});
+};
+
+Client.prototype.updateSession = function(token, ip, request) {
+	const client = this;
+	const agent = UAParser(request.headers["user-agent"] || "");
+	let friendlyAgent = "";
+
+	if (agent.browser.name.length) {
+		friendlyAgent = `${agent.browser.name} ${agent.browser.major}`;
+	} else {
+		friendlyAgent = "Unknown browser";
+	}
+
+	if (agent.os.name.length) {
+		friendlyAgent = ` on ${agent.os.name} ${agent.os.version}`;
+	}
+
+	client.config.sessions[token] = {
+		lastUse: Date.now(),
+		ip: ip,
+		agent: friendlyAgent,
+	};
 };
 
 Client.prototype.setPassword = function(hash, callback) {
 	var client = this;
 
-	client.updateToken(function(token) {
-		client.manager.updateUser(client.name, {
-			token: token,
-			password: hash
-		}, function(err) {
-			if (err) {
-				log.error("Failed to update password of", client.name, err);
-				return callback(false);
-			}
+	client.manager.updateUser(client.name, {
+		password: hash
+	}, function(err) {
+		if (err) {
+			log.error("Failed to update password of", client.name, err);
+			return callback(false);
+		}
 
-			client.config.password = hash;
-			return callback(true);
-		});
+		client.config.password = hash;
+		return callback(true);
 	});
 };
 
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -25,14 +25,8 @@ ClientManager.prototype.init = function(identHandler, sockets) {
 	}
 };
 
-ClientManager.prototype.findClient = function(name, token) {
-	for (var i in this.clients) {
-		var client = this.clients[i];
-		if (client.name === name || (token && token === client.config.token)) {
-			return client;
-		}
-	}
-	return false;
+ClientManager.prototype.findClient = function(name) {
+	return this.clients.find((u) => u.name === name);
 };
 
 ClientManager.prototype.autoloadUsers = function() {
@@ -105,7 +99,8 @@ ClientManager.prototype.addUser = function(name, password, enableLog) {
 			password: password || "",
 			log: enableLog,
 			awayMessage: "",
-			networks: []
+			networks: [],
+			sessions: {},
 		};
 		fs.writeFileSync(
 			Helper.getUserConfigPath(name),
--- a/src/server.js
+++ b/src/server.js
@@ -17,7 +17,6 @@ const net = require("net");
 const Identification = require("./identification");
 
 var manager = null;
-var authFunction = localAuth;
 
 module.exports = function() {
 	log.info(`The Lounge ${colors.green(Helper.getVersion())} \
@@ -94,10 +93,6 @@ module.exports = function() {
 in ${config.public ? "public" : "private"} mode`);
 	});
 
-	if (!config.public && (config.ldap || {}).enable) {
-		authFunction = ldapAuth;
-	}
-
 	var sockets = io(server, {
 		serveClient: false,
 		transports: config.transports
@@ -105,7 +100,7 @@ in ${config.public ? "public" : "private"} mode`);
 
 	sockets.on("connect", function(socket) {
 		if (config.public) {
-			auth.call(socket);
+			auth.call(socket, {});
 		} else {
 			init(socket);
 		}
@@ -178,7 +173,7 @@ function index(req, res, next) {
 	res.render("index", data);
 }
 
-function init(socket, client) {
+function init(socket, client, generateToken) {
 	if (!client) {
 		socket.emit("auth", {success: true});
 		socket.on("auth", auth);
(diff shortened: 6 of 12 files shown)


metsjeesus:
Im just thinking loud.

If you first time enter to the lounge, browser dont have any tokens stored. After getting right credentials, server sends a token and username.
If you now open new tab on browser, it will send both token and username to server. Server compares sended username tokens and if pair is found, logs in.
If you log out, server stops socket connections and browser forgets user-token pair.

Now some test cases. If you have 2 tabs open with login screen, you could log in as different users on both tabs. Last login will set the username-token pair, so if you do refresh or open new tabs on browser, last user will be used. It might be a feature, i personally dont like it and i feel it will be problem later. Proposed fix, if you log in, send allways current user-token pair and send warning that you already have logged in if those match.

Same problem is with logging out. If you are logged in on 2 tabs and log out on one tab, you still are logged in on other. Proposed fix, generate new user-token pair for every new browser. Logging out will close server side all connections with that user-token pair, even if there are many. #986 will help on that part. Maybe make token unique to IP, example if there are connections from same IP address, dont generate new token, instead use that IP address token from established connection, so there is no need to store IP address with token. Benefit, logging out logs out all connection on that IP. If you do so, there will be problem if you are on wifi and log in on 2 different machines. They share IP addresses and logging out on either one logs both out. Proposed fix, after logging in, show a mild warning that you are already logged in on this IP address and give a option to tell with 1 click that this is different machine what generates new token pair.

xPaw:
> Proposed fix, generate new user-token pair for every new browser.

That's the eventual plan that I want to implement. Every login would generate a new token pair (#1139).

As for different tabs in same browser, you would need to have different browsers or profiles to login twice (e.g. incognito).

src/server.js
```diff
@@ -251,12 +245,48 @@ function init(socket, client) {
...
-			token: client.config.token || null
+
+		socket.on("sign-out", token => {
+			delete client.config.sessions[token];
+
+			client.manager.updateUser(client.name, {
+				sessions: client.config.sessions
+			}, err => {
+				if (err) {
+					log.error("Failed to add session token for", client.name, err);
```
PolarizedIons: This should probably be remove instead of add

client/js/lounge.js
```diff
@@ -232,10 +230,8 @@ $(function() {
 			renderNetworks(data);
 		}
 
-		if (data.token && $("#sign-in-remember").is(":checked")) {
+		if (data.token) {
```
YaManicKill: Why are we not checking if the user wants to be remembered anymore?

client/js/lounge.js
```diff
@@ -1045,8 +1041,12 @@ $(function() {
 	});
 
 	sidebar.on("click", "#sign-out", function() {
+		socket.emit("sign-out", storage.get("token"));
 		storage.remove("token");
-		location.reload();
+
+		if (!socket.connected) {
```
YaManicKill: Hmmm, is this because we get a notification from the server that we have logged out and then we refresh if we already are connected? Also, the emit for sign-out won't work if we aren't connected.

client/js/socket.js
```diff
@@ -52,4 +53,9 @@ socket.on("authorized", function() {
 	$("#loading-page-message").text("Authorized, loading messages…");
 });
 
+socket.on("sign-out", function() {
+	storage.remove("token");
```
YaManicKill: Ah yeah, I see it here.

src/client.js
```diff
@@ -286,34 +285,40 @@ Client.prototype.connect = function(args) {
...
 };
 
+Client.prototype.updateSession = function(token, ip, request) {
+	const client = this;
+	const agent = UAParser(request.headers["user-agent"] || "");
+
+	client.config.sessions[token] = {
+		lastUse: Date.now(),
+		ip: ip,
+		agent: `${agent.browser.name} ${agent.browser.major} on ${agent.os.name} ${agent.os.version}`,
```
YaManicKill: Are we using this anywhere?

client/index.html
```diff
@@ -96,7 +96,7 @@ <h1 class="title">Sign in to The Lounge</h1>
 							</div>
 							<div class="col-xs-12">
 								<label class="remember">
-									<input type="checkbox" name="remember" id="sign-in-remember" checked>
+									<input type="checkbox" name="remember" value="on" id="sign-in-remember" checked>
```
xPaw: It already is.

client/js/lounge.js
xPaw: Server won't generate or send a token if you don't want to be remembered.

client/js/lounge.js
xPaw: If connected, it waits for server ACK. If not, just reload.

src/client.js
xPaw: Not yet, but it's useful metadata when looking at user config. Will be used in #986.

xPaw:
This PR has a consequence of requiring to login again after upgrade (because single `token` is no longer used), but I think this work out great, as couple releases ago we increased our bcrypt rounds, so everyone who upgrades Lounge will also refresh their passwords.

metsjeesus:
I would like to promisify manager and client side code too, but i think its not this pull job to do it.
#1235 and #1236 is mostly needed if you need to add a auth method, plus it easier to understand whats happening.

src/client.js
```diff
@@ -286,34 +285,40 @@ Client.prototype.connect = function(args) {
 	client.save();
 };
 
-Client.prototype.updateToken = function(callback) {
-	var client = this;
-
-	crypto.randomBytes(48, function(err, buf) {
+Client.prototype.generateToken = function(callback) {
```
astorije: Mind adding a test for this function?

src/client.js
astorije: Let's not add it in here then, and definitely not as a pre-formatted string.
I did [something here](https://github.com/thelounge/lounge/pull/986/commits/2d4a495721315932518d77fe1ddba49edcd57340) that I think looks really nice in the UI, when both this PR and #986 are ready.

src/clientManager.js
```diff
@@ -25,14 +25,8 @@ ClientManager.prototype.init = function(identHandler, sockets) {
...
-ClientManager.prototype.findClient = function(name, token) {
-	for (var i in this.clients) {
-		var client = this.clients[i];
-		if (client.name === name || (token && token === client.config.token)) {
-			return client;
-		}
-	}
-	return false;
+ClientManager.prototype.findClient = function(name) {
+	return this.clients.find(u => u.name === name);
```
astorije: Sorry if this a stupid question but I'm having a bit of a hard time following this: why is `token` not needed here anymore?

src/server.js
```diff
@@ -256,12 +250,48 @@ function init(socket, client) {
...
 			}
 		);
-		socket.join(client.id);
-		socket.emit("init", {
-			active: client.lastActiveChannel,
-			networks: client.networks,
-			token: client.config.token || null
+
+		socket.on("sign-out", token => {
+			delete client.config.sessions[token];
```
astorije: Oh, JavaScript...

Any benefits we could get from using a `Map` here?

src/client.js
xPaw: It's either storing full raw user agent, or the pre-formatted one. I opted for the formatted one.

I do want to store user agent info in this PR because I don't know how long it will take for the UI to ship, and I don't want to miss this info.

src/client.js
xPaw: That would be a pretty useless test.

src/clientManager.js
xPaw: Client now sends name+token combo instead of just token, so we can find the user and then compare the token. Previous version iterated all users and compared tokens.

xPaw:
Rebased on master.

src/client.js
astorije: I still don't want this to interfere with what I've done on the UI. I've spent quite some time making it look nice, I'd rather not throw that away :)

src/client.js
astorije: Well, if you say so... :(

src/server.js
```diff
@@ -256,12 +250,48 @@ function init(socket, client) {
...
 			}
 		);
-		socket.join(client.id);
-		socket.emit("init", {
-			active: client.lastActiveChannel,
-			networks: client.networks,
-			token: client.config.token || null
+
+		socket.on("sign-out", (token) => {
+			delete client.config.sessions[token];
```
astorije: Oh, JavaScript...

Any benefits we could get from using a `Map` here?

src/client.js
xPaw: How would this change throw your changes in a future PR?

src/server.js
xPaw: Maybe just a tiny bit better code, but I'm not willing to rewrite and test it again.

client/index.html
YaManicKill: Right, so why this addition then?

client/js/lounge.js
YaManicKill: Ah right, so if we are connected, we'll reload on another message, gotcha. I was wondering why we wouldn't reload if connected.

client/js/socket-events/sign_out.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const socket = require("../socket");
+const storage = require("../localStorage");
+
+socket.on("sign-out", function() {
+	storage.remove("token");
+	location.reload();
```
YaManicKill: Aha, here's that handling.

src/client.js
YaManicKill: Yeah, I'm kinda with @xPaw on this one, it would essentially just be testing crypto library.

client/index.html
xPaw: Just explicitly stating the value to send (for reader's sanity), and I compare against it on the server.

src/client.js
astorije: Okay with this, but adding something that we will remove shortly after is a bit odd (and if the answer is "let's keep both", then 👎, I'm only 👍 if this is a temporary solution until #986 makes it).

xPaw: Why would we remove it? Is your suggestion to store full user agent string?

src/server.js
```diff
@@ -248,8 +243,7 @@ function init(socket, client) {
 								const obj = {};
 
 								if (success) {
-									obj.success = "Successfully updated your password, all your other sessions were logged out";
-									obj.token = client.config.token;
+									obj.success = "Successfully updated your password";
```
astorije: Are we not logging out other sessions with this PR on password change anymore?? That would be a regression security-wise.

astorije:
Why is "Fix #712 to prevent timing attacks on tokens" crossed out? It doesn't fix this anymore?

src/server.js
xPaw: It's pretty silly to log everything out on password change. Push PR will also store push endpoint config in the session object, so resetting all sessions would break push notifications.

xPaw:
Nope it does not fix it.

src/client.js
astorije: #986 already [sends that information](https://github.com/thelounge/lounge/pull/986/files#diff-e6a5b42b2f7a26c840607370aed5301aR437), just in a different format.

xPaw: It sends it sure, but we need to store it on disk, which is what this does.

src/server.js
astorije: You added that behavior in https://github.com/thelounge/lounge/pull/370, and that was a good thing for security purposes.

#986 currently logs out active connections, it will need to be updated to log out inactive connections too (since this PR helps keeping track of connections even after you close The Lounge).

Until #986 happens, we really really should not remove this, otherwise there is no way to log out from other sessions anymore, which was possible before.

src/client.js
astorije: Ah, that makes sense I guess. Can you then use the format I put [here](https://github.com/thelounge/lounge/pull/986/files#diff-e6a5b42b2f7a26c840607370aed5301aR438)? In several cases, some of these attributes were not defined, so this was the best I could do to make it look nice.

astorije:
> This PR has a consequence of requiring to login again after upgrade

That's fine to me, but I'll need to remember to add a mention in the changelog entry, just in case some admins need to broadcast this to their users.

src/server.js
xPaw: I tried to find what's the internet's opinion on this, and can't find anything conclusive. Google seems to be keeping sessions on password change?

If I were to add session destroying back, there's the real consequence of killing push notifications (that PR depends on this PR) when changing the password. Do we want that?

astorije: Honestly, because we have that already, yes. I am happy to not do that anymore when #986 arrives, i.e. when there is an alternate way to do this, but I very strongly welcomed you allowing this in the first place in #370.

So, sure, it's not perfect, but it's times better than not having any option for this. Also, I'm not too worried about push notifications, because if the behavior is meant to kill sessions everywhere, there is no reason to keep receiving notifications.

Again, more than happy to change my mind when #986 makes it (and it's great stuff, I really want that in!) but "not perfect" vs. "back to nothing" is an easy choice to me :)

src/client.js
xPaw: How would that [].join help with undefineds?

astorije:
@xPaw, ping to update this PR? I believe your push notification depends on it, right? Let's get this out of the way either way :)

src/client.js
xPaw: I have done this in a more elaborate way.

astorije: If `agent.browser.major` or `agent.os.version` is `undefined`, you're going to have some issues again, no?

xPaw: I don't think UAParser makes `browser` available without version. Even if it does, this is a minor problem that I would rather keep.

astorije: That's fair. I like that my version is more concise, but it is way more cryptic so I'm okay with your way.

xPaw:
@astorije merge then. @metsjesus can work on that sessions UI then.

astorije:
> @astorije merge then. @metsjesus can work on that sessions UI then.

/Cc @metsjeesus :)

PR #1202, opened 2017-06-03
Parse X-Forwarded-For header correctly

### LINKED ISSUE #1169: Lounge crashes when there more than one IP in X-Forwarded-For header
Lounge crashes when there more than one IP in X-Forwarded-For header if `reverseProxy = true`. 

Fix: `ip = req.headers["x-forwarded-for"].split(",")[0] || req.connection.remoteAddress;`

Also there should be more checks for this header, regex maybe?

xPaw:
Fix #1169. cc @bews

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -13,6 +13,7 @@ var dns = require("dns");
 var Helper = require("./helper");
 var ldap = require("ldapjs");
 var colors = require("colors/safe");
+const net = require("net");
 const Identification = require("./identification");
 
 var manager = null;
@@ -113,13 +114,15 @@ in ${config.public ? "public" : "private"} mode`);
 	});
 };
 
-function getClientIp(req) {
-	var ip;
+function getClientIp(request) {
+	let ip = request.connection.remoteAddress;
 
-	if (!Helper.config.reverseProxy) {
-		ip = req.connection.remoteAddress;
-	} else {
-		ip = req.headers["x-forwarded-for"] || req.connection.remoteAddress;
+	if (Helper.config.reverseProxy) {
+		const forwarded = (request.headers["x-forwarded-for"] || "").split(/\s*,\s*/).filter(Boolean);
+
+		if (forwarded.length && net.isIP(forwarded[0])) {
+			ip = forwarded[0];
+		}
 	}
 
 	return ip.replace(/^::ffff:/, "");


src/server.js
```diff
@@ -36,6 +36,11 @@ module.exports = function() {
 		.set("view engine", "html")
 		.set("views", path.join(__dirname, "..", "client"));
 
+	// Express will handle X-Forwarded-For header for us
+	if (Helper.config.reverseProxy) {
+		app.enable("trust proxy");
```
YaManicKill: As far as I can see from the [docs](https://expressjs.com/en/guide/behind-proxies.html), doing this will take the default IP and not the forward one. I think you'll want to do app.enable("trust proxy", true) which will take the first from x-forwarded-for.

xPaw: ` app.enable("trust proxy", true)` is not a thing.

> Calling app.set('foo', true) for a Boolean property is the same as calling app.enable('foo'). 

https://expressjs.com/en/4x/api.html#app.set

YaManicKill: Ah, I misread it and thought it said "set" not "enable". Alrighty, looks good.

astorije:
@bews, do you think you could review/test this?

bews:
This fix doesn't work: it completely ignores `x-forwarded-for` header. Invalid call maybe?

xPaw:
Looks like we should be using `req.ip` instead of `req.connection.remoteAddress`. Can you test that?

bews:
> Looks like we should be using req.ip instead of req.connection.remoteAddress

Already tried, it's `undefined`.

xPaw:
Must be socket.io screwing with it. I also noticed in socket.io v2, in node v4 and v6, `connection.remoteAddress` is undefined and makes our tests fail.

astorije:
@xPaw, do we want to merge #1143 first then?

xPaw:
So yeah, looks like socket.io is doing something odd, and the `remoteAddress` not being available in 2.0 does not help. Will need to dig in further.

xPaw:
@bews try this one then.

src/server.js
```diff
@@ -113,13 +113,14 @@ in ${config.public ? "public" : "private"} mode`);
...
-	var ip;
+function getClientIp(request) {
+	let ip;
 
 	if (!Helper.config.reverseProxy) {
-		ip = req.connection.remoteAddress;
+		ip = request.connection.remoteAddress;
 	} else {
-		ip = req.headers["x-forwarded-for"] || req.connection.remoteAddress;
+		ip = (request.headers["x-forwarded-for"] || "").split(/\s*,\s*/).filter(Boolean).reverse();
```
bews: remove `.reverse()`

bews:
> @bews try this one then.

It doesn't validate header, so this isn't a proper fix.

xPaw:
> It doesn't validate header, so this isn't a proper fix.

I think that's the job of reverse proxy. This PR only fixes parsing the header if it includes multiple IPs.

bews:
> I think that's the job of reverse proxy.

You mean nginx should do this? How?

xPaw:
nginx would *set* the header if it's publicly facing, so in that case there is nothing to validate.

bews:
> nginx would set the header if it's publicly facing

And if it's not? What about behind NAT config?

Also, if there won't be any validation, when this PR doesn't fix #1169 - Lounge still can be crashed in cases when raw header is passed.

xPaw:
> And if it's not? What about behind NAT config?

What about it? Whatever is at the top must be setting correct IP in the header. You simply wouldn't use the header that client provides to the proxy.

bews:
IIRC every proxy by default just adds client IP to the `x-forwarded-for` header and doesn't validate the previous content at all. For example: `0,0,0,0` -> `0,0,0,0, 192.168.1.1`. But I don't have time to check this.

YaManicKill:
Yeah, I'm actually with @bews on this one. Even if all reverse proxies work fine, if a user isn't on one, a malicious actor could put a malformed IP on the header and cause the lounge to crash. We shouldn't let this happen.

xPaw:
Added `net.isIP`.

src/server.js
```diff
@@ -113,13 +114,15 @@ in ${config.public ? "public" : "private"} mode`);
...
+	let ip = request.connection.remoteAddress;
 
-	if (!Helper.config.reverseProxy) {
-		ip = req.connection.remoteAddress;
-	} else {
-		ip = req.headers["x-forwarded-for"] || req.connection.remoteAddress;
+	if (Helper.config.reverseProxy) {
+		ip = (request.headers["x-forwarded-for"] || "").split(/\s*,\s*/).filter(Boolean);
+
+		if (ip[0] && net.isIP(ip[0])) {
```
bews: This doesn't look right, if IP is invalid `ip` var will be array.

xPaw: You are right, will fix.

astorije:
@YaManicKill, I understand you don't approve this PR anymore? @xPaw, @bews, what do you think is left to do here?

xPaw:
@astorije it's done

astorije:
@bews, are you okay with this now?

bews:
It looks fine, but I haven't tested this.

PR #1205, opened 2017-06-05
Keep original <title> name when changing the title

xPaw:
Changing `title` in `index.html` will no longer require modifying and recompiling the javascript.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -26,6 +26,8 @@ $(function() {
 	var sidebar = $("#sidebar, #footer");
 	var chat = $("#chat");
 
+	$(document.body).data("app-name", document.title);
+
 	var ignoreSortSync = false;
 
 	var pop;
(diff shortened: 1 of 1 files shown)


astorije:
What's the purpose of this change? I'm not sure I'm getting it.

xPaw:
This will allow changing the app title via `<title>` tag in html as it won't reset back to `The Lounge` as soon as a channel name appears in the title.

astorije:
So, if I understand correctly, it's for those who want to brand their instances with their own name, now they just have to change the `<title>` in `index.html` instead of that + in the JS file?

YaManicKill:
> So, if I understand correctly, it's for those who want to brand their instances with their own name, now they just have to change the <title> in index.html instead of that + in the JS file?

Exactly, we had someone in the IRC confused when changing the title in index.html didn't actually change the title.

PR #1216, opened 2017-06-10
Option to display timestamps with 12 hour clock

xPaw:


xPaw:
@astorije the test fails horribly because the module requires `options.js` which requires jQuery, and that straight up just fails on node.

astorije:
I'm not sure we should add untestable code, especially on something that was already tested. What alternatives do we have?

xPaw:
I don't know, I didn't want to spend a lot of time figuring out how we would test jquery/dom code.

YaManicKill:
@xPaw Just use https://www.npmjs.com/package/mock-require and mock out options.js when testing the file. Should be nice and easy to do.

xPaw:
@YaManicKill not so easy, the handlebars module requires options so that it can format dates. I tried mocking out jquery but it crashes because it calls to jquery and needs stuff.

YaManicKill:
@xPaw Yeah, but you can just mock the options themselves, that keeps the test very self contained. Or am I missing something? Mocking jquery would be more difficult, but mocking the options.js should be very easy. I'll take a look sometime if you can't get it to work.

client/js/libs/handlebars/localetime.js
```diff
@@ -1,7 +1,17 @@
 "use strict";
 
 const moment = require("moment");
+const constants = require("../../constants");
 
 module.exports = function(time) {
-	return moment(time).format("D MMMM YYYY, HH:mm:ss");
+	const options = require("../../options");
```
astorije: Why is this part of the function itself? To make sure it uses the most up-to-date values set by user?

client/js/options.js
```diff
@@ -86,9 +88,15 @@ settings.on("change", "input, select, textarea", function() {
...
-	} else if (name === "showSeconds") {
+	} else if (name === "showSeconds" || name === "use12hClock") {
 		chat.find(".msg > .time").each(function() {
-			$(this).text(tz($(this).parent().data("time")));
+			const time = $(this).parent().data("time");
+
+			$(this).text(tz(time));
+
+			if (name === "use12hClock") {
+				$(this).attr("aria-label", localetime(time));
```
astorije: `localetime` requires `options` and `options` requires `localetime` 😰 Not sure we want to go there :/

astorije:
Alright, a few ideas:

- It's kinda heavy to require the whole thing each time we want to simply access the user preferences: why not having an `options.js` module, which just contains the user pref databag (whatever its storage), and another `settings.js` module that exists to take care of setting/changing checkboxes, dealing with notifications warning, etc.
- Instead of reading from localStorage each time, set a global variable in `window` with user pref that mirrors what's in the local storage, without having to parse it every time. I know "global variable" is taboo, but that's exactly what `options.use12hClock` is at the moment, a global variable in disguise.
- At the moment, `localetime` introduces a side effect (and very quickly the word "mocking" appears 😅), we should avoid this if we can. What about adding an argument to the function for the options? Alternatively, can we pass the options at a higher level (like at `require` time, or else)?

Note that these are not mutually exclusive. Either way, this PR feels wrong in multiple ways in the current state: deleting a test, introducing a side-effect, considering mocking, etc.
Sorry!

client/js/libs/handlebars/localetime.js
xPaw: It has to be due to a cyclical dependency.

xPaw:
@astorije `require` is cached, so first time you call it's gonna initialise, next time it's gonna return the same object. Most modules in master already do this (require options/socket multiple times), so I am not doing anything new with this.

astorije:
Fair point re: cached, but deleting a test file just because jQuery gets in the way is a hard no from me.
What do you think of (either or both of) the first 2 points I have mentioned above, specifically:

> - having an `options.js` module, which just contains the user pref databag (whatever its storage), and another `settings.js` module that exists to take care of setting/changing checkboxes, dealing with notifications warning, etc.
> - Instead of reading from localStorage each time, set a global variable in `window` with user pref that mirrors what's in the local storage, without having to parse it every time. I know "global variable" is taboo, but that's exactly what `options.use12hClock` is at the moment, a global variable in disguise.

Separating options storage/data/model (whatever we call it) from the UI to update them would not only make for a nice and clear modularization, it would also mean we can keep the test this PR removes, and manually sets up option values in tests.
I can help with this if you need.

xPaw:
Refactoring settings is not something I am willing to do.

YaManicKill:
Did you ever try mocking out the options? It really shouldn't be difficult. If I have time I'll take that over if you don't want to do it.

xPaw:
@YaManicKill feel free to.

PR #1217, opened 2017-06-10
Fix MOTD underline in Safari

MaxLeiter:
before: before

after: after

Basically, safari's user agent CSS applies an underline to <abbr> tags, which bootstrap already does as well (but bootstrap doesn't use text-decoration so it's not overriden).

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -56,6 +56,18 @@ a:hover {
 	opacity: .8;
 }
 
+/**
+ * From Normalize. See https://github.com/thelounge/lounge/pull/1217
+ * 1. Remove the bottom border in Chrome 57- and Firefox 39-.
+ * 2. Add the correct text decoration in Chrome, Edge, IE, Opera, and Safari.
+ */
+
+abbr[title] {
+	border-bottom: none; /* 1 */
+	text-decoration: underline; /* 2 */
+	text-decoration: underline dotted; /* 2 */
+}
+
 h1,
 h2,
 h3 {


client/css/style.css
```diff
@@ -1280,6 +1280,10 @@ kbd {
 	margin-top: 30px;
 }
 
+#settings abbr {
+	text-decoration: none;
+}
+
```
astorije: Any reason why not `abbr` entirely?

xPaw: Yeah, this could go into `bootstrap.css`, imo.

astorije: Wait no, I'd rather not patch Bootstrap any further...

astorije:
FYI, this has been fixed in Normalize v4+ with the following:

```css
/**
 * 1. Remove the bottom border in Chrome 57- and Firefox 39-.
 * 2. Add the correct text decoration in Chrome, Edge, IE, Opera, and Safari.
 */

abbr[title] {
  border-bottom: none; /* 1 */
  text-decoration: underline; /* 2 */
  text-decoration: underline dotted; /* 2 */
}
```

Unfortunately, our version of Bootstrap (v3.2.0) uses Normalize v3.0.1 and even latest Bootstrap (v3.3.7) uses Normalize v3.0.3.
Code above is integrated in Bootstrap v4 alpha (using Normalize v5.0.0, latest being v7.0.0), but I don't think it's anywhere near completion nor a trivial upgrade for us.

I think we should use the code pasted above. It could go into `style.css` or, like @xPaw says, in `bootstrap.css`. In the end, if we're using the "official" snippet, I don't care where that much.

xPaw:
@astorije What if we included normalize.css and refactored bootstrap.css out? I believe we only require the grid layout from bootstrap, all the other styles do not seem to go used.

astorije:
That's a decent possibility, but way beyond the point of this PR though.
@MaxLeiter needs very few changes so this PR can make it to v2.3.2, and what you are describing could a a bigger refactor in a future minor release IMO.

MaxLeiter:
Changed to just `abbr`, as requested by @xPaw on IRC

astorije:
Hey @MaxLeiter, I stand by my comment above, I think we should definitely apply what Normalize.css does. @xPaw, you okay with that?

MaxLeiter:
@astorije well normalize's looks slightly different (although an improvement over the actual bug for sure). Which do you prefer? 

`underline: none`: first

normalizes solution: second

astorije:
I get where you're going because in terms of styling, I do prefer `underline: none`, but if our end-goal is to get rid of Bootstrap and rely on Normalize.css entirely, we should probably follow what they do.

That being said, when we do have Normalize only (i.e. no Bootstrap anymore), nothing prevents us to do this ourselves (though I'm sure we'll have forgotten).
Ultimately I don't have a strong opinion anymore, I'm sure @xPaw will have a wiser opinion than me so I'll let him pick between the two you just showed.

Thanks for your efforts anyway!!

xPaw:
Use the normalize solution then.

astorije:
@MaxLeiter, do you think you'll manage to do that today? If so I can draft the 2.3.2 release this weekend 🎉

MaxLeiter:
Done

xPaw:
@MaxLeiter use tabs to fix stylelint.

PR #1227, opened 2017-06-14
Fix the copying of multiple lines having too many linebreaks

### LINKED ISSUE #1146: Copying text has new lines between time, username, and message
I think this is due to the new layout in chat.

When you copy something like:

```
09:54 YaManicKill test
```

What you get copied into your clipboard is:

```
09:54
YaManicKill
test
```

This is probably due to the new flexbox layout. Not sure of the best way to do it, we could intercept the copy and remove some of the new lines, but that feels...not great. But I can't think of a better way to do it.

I feel like this should be a blocker for 2.3.0, but if we don't have an idea for a fix this week, maybe not.

YaManicKill:
Fixes #1146 

I apologise profusely for this, and if someone else can think of a better way (no, removing flexbox is not better) then feel free to deny this. But this works.

Also, this will make it incredibly easy to add in "<" and ">" around usernames once this is in (but that's for a different PR).

### DIFF
--- a/client/js/clipboard.js
+++ b/client/js/clipboard.js
@@ -1,0 +1,28 @@
+"use strict";
+
+function addLink() {
+	// Get the selected text
+	const selection = window.getSelection(),
+		newdiv = document.createElement("div");
+
+	let copytext = selection;
+
+	copytext = copytext.toString();
+	copytext = copytext.replace(/\n(\d{2}:\d{2}(:\d{2})?)/g, "<br>$1"); // Replace first newlines on each message with <br>
+	copytext = copytext.replace(/\n/g, " "); // Remove all other newlines
+
+	// hide the newly created container
+	newdiv.style.position = "absolute";
+	newdiv.style.left = "-99999px";
+
+	// insert the container, fill it with the text
+	document.body.appendChild(newdiv);
+	newdiv.innerHTML = copytext;
+	selection.selectAllChildren(newdiv);
+
+	window.setTimeout(function() {
+		document.body.removeChild(newdiv);
+	}, 100);
+}
+
+document.addEventListener("copy", addLink);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -14,6 +14,7 @@ const emojiMap = require("./libs/simplemap.json");
 require("./libs/jquery/inputhistory");
 require("./libs/jquery/stickyscroll");
 require("./libs/jquery/tabcomplete");
+require("./clipboard");
 const helpers_parse = require("./libs/handlebars/parse");
 const helpers_roundBadgeNumber = require("./libs/handlebars/roundBadgeNumber");
 const slideoutMenu = require("./libs/slideout");


YaManicKill:
I should probably add some tests for this as the logic is a little convoluted. Well, at least I should bring out the regex code and test that. The rest should be fine.

YaManicKill:
Also, yes, I'm aware the build is failing. I will fix it tomorrow.

YaManicKill:
IRC chat regarding this:

```
15:24:06 +xPaw YaManicKill, is it not possible to get selected elements? i don't like the fact this uses regexes
15:25:24 lounge-user37 (cbdbf04c@gateway/web/thelounge/ip.203.219.240.76) has quit (Client Quit)
15:25:25 +YaManicKill What are you thinking? Taking the actual elements that are copied, moving them to the invisible div, and changing it so that it isn't flexbox anymore, and the copy-paste will just work?
15:25:42 +YaManicKill I guess that could work, yeah.
15:26:07 +YaManicKill I'm not sure if we get access to that, though.
15:27:18 +YaManicKill This is what we get from window.getSelection: https://developer.mozilla.org/en-US/docs/Web/API/Selection
15:27:24 +xPaw you probably would lose flexbox as is
15:27:45 +YaManicKill What do you mean?
15:27:53 +xPaw css
15:28:11 +xPaw copying to some invisible div means no css aplies to it
15:28:23 +YaManicKill Well no, because the div still has css on it.
15:28:41 +YaManicKill Having said that, the div wouldn't have flexbox enabled on it, so we shouldn't need to do anything.
15:28:46 +xPaw it won't be inside #chat, our selectors wont apply to it
15:28:55 +YaManicKill Yeah
15:29:19 +xPaw does the element need to be in body for it to be selected?
15:29:32 +YaManicKill Yeah
15:30:04 +YaManicKill You certainly could do it, but the selection doesn't have a simple way to get all the elements that are selected (and obviously we'll need to deal with if an element isn't fully selected) so it'll be horrible logic.
15:31:12 +YaManicKill You can get the first element that is in the selection, and the end one. So you'll need to get the parent of the first selected element, and the parent of the last selected one, get all in-between, and get all their children, copy them over, and then remove the text within that that wasn't selected.
15:31:16 +xPaw I see where you copied the code, haha
15:31:20 +YaManicKill But it could be done.
15:31:40 +YaManicKill How dare you, I typed every character myself ;-)
15:32:04 +xPaw are you sure this regex doesn't fuck with copying other text? e.g from help
15:32:12 +xPaw this is really a hack :(
15:32:27 +YaManicKill I'm not sure, no.
15:32:33 +YaManicKill Of course it's a hack. No matter what we do, this will be a hack.
15:32:44 +YaManicKill We are working around a browser bug.
15:32:46 +xPaw could remove flexbox
15:32:49 +YaManicKill Hell no
15:32:56 +YaManicKill That's not the solution.
15:36:41 +YaManicKill I could add a check to see if we are copying from inside a chat, xPaw
15:37:01 +YaManicKill Then it doesn't interfere with copy/paste from other parts of the interface, including help page, etc.
15:37:23 +YaManicKill Because yeah, it does interfere with help page copy
15:37:38 +xPaw and if you copy like this? https://s.xpaw.me/2017-06/14-143711527.png
15:38:00 ScruffyRules That's magical
15:38:18 +YaManicKill Oh goodness, why would you do that?
15:38:28 +YaManicKill But yeah, it will mess up there as well.
15:38:28 +xPaw i don't know, but you technically can
15:38:50 +YaManicKill It'll basically remove all newlines that aren't immediatelly before a "dd:dd" or "dd:dd:dd"
15:38:58 +YaManicKill I accept it's not perfect.
15:39:02 +YaManicKill It needs improvement.
15:39:24 lounge-user11 (2e636633@gateway/web/thelounge/ip.46.99.102.51) has joined the channel
15:39:26 +xPaw that'll break on am/pm
15:39:42 +YaManicKill Sure, but that isn't implemnted yet, and that won't be too difficult to add.
15:40:00 +YaManicKill btu I accept it's less than ideal having to change everytime we change the possible formats
15:40:05 +YaManicKill I want to try your suggestion.
15:40:17 +xPaw I really don't think regexing this is the solution
15:40:20 +YaManicKill It would be a better solution if we can manage to do it.
15:40:31 +xPaw you can copy only nick and message
15:40:37 +xPaw you can copy nick partially, etc
15:40:54 +YaManicKill Yes, yes, I get that.
15:41:09 +YaManicKill I have already said your solution would be better. But only if we actually can do it.
15:41:40 +YaManicKill But actually, it won't fail on those 2 examples.
15:42:08 +YaManicKill If you copy 1 line with just message and nick (or part of nick) it will just remove all of the newlines and you'll have what you expect.
15:42:18 +xPaw discord has flex elements, but then like 3 divs more inside
15:42:42 +YaManicKill Alright? What's your point?
15:42:51 +xPaw was checking how they deal with it
15:43:45 +YaManicKill Ok
15:43:55 +YaManicKill Anyway, I got work to do. I'll take a look at your suggestion tomorrow.
```

Also, I totally just copied all that from the lounge without having to fix the newlines. Beautiful.

YaManicKill:
Fixed the build. It still does the thing that @xPaw doesn't want it to do, but that'll be a lot more work to figure out.

YaManicKill:
Going to close this. It looks like @xPaw has a working version of moving back to table layout, and fixing the problem highlighted in #523.

@thelounge/maintainers Was there any other reason for removing table layout? AFAICS, it was just the awkward unnecessary line wrapping?

starquake:
Am I too late with this? https://gist.github.com/starquake/01c55624796a68ed58d9c5c9d75b77ce

YaManicKill:
Well, if we don't end up managing to move back to table layout, that might be the better way to deal with it than what I did. But obviously it's not ideal...

PR #1231, opened 2017-06-16
Update to eslint 4 and enforce extra rules

xPaw:


.eslintrc.yml
```diff
@@ -9,22 +9,28 @@ env:
   node: true
 
 rules:
+  arrow-body-style: 2
+  arrow-parens: [2, as-needed]
```
xPaw: cc @thelounge/maintainers 

@YaManicKill wants this to be `always`

YaManicKill: I would like this to be always, and here's my reasoning:

We require curly braces around all conditional and loop statements. The reason we do this (and why it's a standard across JS development just now) is because you can try to add a second line to an if, and if you forge tto add in the brackets afterwards, it won't do what you expect (it'll run the second line always).

I feel like we have the same issue with arrow functions, where you can technically do a 1 argument arrow function which doesn't have parenthesis, but if you add in a second argument, and forget to add in the brackets, you will end up with issues. The extra characters don't add much work, and they stop us having issues in the future. Also, it keeps all arrow functions looking consistent.

astorije: I personally would prefer `as-needed` over `always`, but I won't fight over it :)

I completely get what you're saying, but there is a small subtlety to it though:

```js
if (something)
  doThis();
  doThat();
```

The above is perfectly valid JS, but probably not what one wants. However:

```js
arr.map(x, y => x * 2); // 💥
```

is simply broken JS, and a typo that one of our checks (tests, linters, compilers) should detect, regardless of the style preference.

Reason why I'd rather keep `as-needed` is because of the beauty/simplicty of the new syntax: 

```js
arr.map(x => x * 2);
```

This is much nicer than:

```js
arr.map((x) => x * 2);
```

And the reason why I won't fight over it is because parentheses are needed in other scenarios:

```js
myFunction(() => doSomething());
myOtherFunction((x, y) => doSomethingElse(x, y));
```

And also because there is more important in life 😅
Anyway, my recommendation is to keep `as-needed` and if we start seeing some issues with it, we can become more strict about it.

.eslintrc.yml
```diff
@@ -33,19 +39,23 @@ rules:
...
+  semi-style: [2, last]
   semi: [2, always]
   space-before-blocks: 2
   space-before-function-paren: [2, never]
   space-in-parens: [2, never]
   space-infix-ops: 2
   spaced-comment: [2, always]
   strict: 2
+  template-curly-spacing: 2
+  yoda: 2
```
astorije: Was this removed from `recommended` or something?

astorije: My bad, it was never in there, good addition.

.eslintrc.yml
```diff
@@ -9,22 +9,28 @@ env:
   node: true
 
 rules:
+  arrow-body-style: 2
+  arrow-parens: [2, always]
```
astorije: Oh well :(

YaManicKill: Yay!

astorije: :trollface:

PR #1232, opened 2017-06-16
Format messages being copied to the clipboard

starquake:
Attempt at fixing or at least improving #1146

Extension of https://github.com/thelounge/lounge/pull/1227.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1960,3 +1960,7 @@ kbd {
 ::-webkit-scrollbar-thumb:vertical:active {
 	background: rgba(0, 0, 0, .6);
 }
+
+.clipboard-only {
+	display: none;
+}
--- a/client/js/clipboard.js
+++ b/client/js/clipboard.js
@@ -1,0 +1,45 @@
+"use strict";
+
+const $ = require("jquery");
+
+function copyMessages(event) {
+	// Get the selected text
+	const selection = window.getSelection();
+	var copyLines = [];
+	var copyText = "";
+
+	if (selection.anchorNode === selection.focusNode) {
+		// Selection does not span messages
+		copyText = selection.toString();
+	} else {
+		// Get the first range
+		let range = selection.getRangeAt(0);
+		let documentFragment = range.cloneContents();
+		let messages = documentFragment.querySelectorAll(".msg");
+
+		messages.forEach(function(message) {
+			let el = $("#" + message.id);
+			if (el) {
+				let time = el.find(".time").get(0).innerText;
+				let from = "";
+				let fromEl = el.find(".from .user").get(0);
+				if (fromEl) {
+					from = fromEl.innerText;
+				}
+				let text = el.find(".text").get(0).innerText;
+				if (from !== "") {
+					copyLines.push(time + " <" + from + "> " + text);
+				} else {
+					copyLines.push(time + " " + text);
+				}
+			}
+		});
+
+		copyText = copyLines.join("\n");
+	}
+
+	(event.originalEvent.clipboardData || window.clipboardData).setData("Text", copyText);
+	event.preventDefault();
+}
+
+$("#chat").on("copy", ".messages", copyMessages);
--- a/client/js/clipboard.js
+++ b/client/js/clipboard.js
@@ -13,20 +13,20 @@ function copyMessages(event) {
 		copyText = selection.toString();
 	} else {
 		// Get the first range
-		let range = selection.getRangeAt(0);
-		let documentFragment = range.cloneContents();
-		let messages = documentFragment.querySelectorAll(".msg");
+		const range = selection.getRangeAt(0);
+		const documentFragment = range.cloneContents();
+		const messages = documentFragment.querySelectorAll(".msg");
 
 		messages.forEach(function(message) {
-			let el = $("#" + message.id);
+			const el = $("#" + message.id);
 			if (el) {
-				let time = el.find(".time").get(0).innerText;
+				const time = el.find(".time").get(0).innerText;
 				let from = "";
-				let fromEl = el.find(".from .user").get(0);
+				const fromEl = el.find(".from .user").get(0);
 				if (fromEl) {
 					from = fromEl.innerText;
 				}
-				let text = el.find(".text").get(0).innerText;
+				const text = el.find(".text").get(0).innerText;
 				if (from !== "") {
 					copyLines.push(time + " <" + from + "> " + text);
 				} else {
--- a/client/js/clipboard.js
+++ b/client/js/clipboard.js
@@ -5,37 +5,26 @@ const $ = require("jquery");
 function copyMessages(event) {
 	// Get the selected text
 	const selection = window.getSelection();
-	var copyLines = [];
+	var copyParts = [];
 	var copyText = "";
 
 	if (selection.anchorNode === selection.focusNode) {
-		// Selection does not span messages
-		copyText = selection.toString();
+		// Selection does not span nodes
+		if (selection.focusNode.parentElement.classList.contains("clipboard")) {
+			// Selection is inside an element that has the clipboard class
+			copyText = selection.toString();
+		}
 	} else {
-		// Get the first range
+		// Selection does span nodes
 		const range = selection.getRangeAt(0);
 		const documentFragment = range.cloneContents();
-		const messages = documentFragment.querySelectorAll(".msg");
+		const parts = documentFragment.querySelectorAll(".clipboard");
 
-		messages.forEach(function(message) {
-			const el = $("#" + message.id);
-			if (el) {
-				const time = el.find(".time").get(0).innerText;
-				let from = "";
-				const fromEl = el.find(".from .user").get(0);
-				if (fromEl) {
-					from = fromEl.innerText;
-				}
-				const text = el.find(".text").get(0).innerText;
-				if (from !== "") {
-					copyLines.push(time + " <" + from + "> " + text);
-				} else {
-					copyLines.push(time + " " + text);
-				}
-			}
+		parts.forEach(function(part) {
+			copyParts.push(part.innerText);
 		});
 
-		copyText = copyLines.join("\n");
+		copyText = copyParts.join("");
 	}
 
 	(event.originalEvent.clipboardData || window.clipboardData).setData("Text", copyText);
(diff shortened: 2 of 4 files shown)


starquake:
Right now partly selected lines are fully copied. I think this makes sense if the UI also shows this is happening. Example: 
jun-16-2017 17-57-31

YaManicKill:
Regarding this, using `user-select: all;` might be something to do. It means that if you want to select any of something, you must select all of it. It does mean that you can't select part of a message, but it is first class support from a browser and doesn't require any complications with JS, etc. Worth trying out.

starquake:
That sounds interesting indeed. But I don't see how that would allow adding < > around the username though.

YaManicKill:
> But I don't see how that would allow adding < > around the username though.

It wouldn't, but I was talking to @xPaw and we were wondering why we don't just have < > on the actual UI (not in css content) and then we don't need to mess around with the content.

Also, it doesn't help with the newlines, but it just helps with your requiring a whole line to be copied, which was part of what you were doing.

astorije:
@YaManicKill, would you be okay if we move this to the v2.4.0 milestone? The code is a bit bigger than what I was hoping and therefore requires more review/test time, while v2.3.2 milestone is probably around the corner.

@starquake, could you fix the conflicts please? Sorry!

astorije:
> I was talking to @xPaw and we were wondering why we don't just have < > on the actual UI (not in css content) and then we don't need to mess around with the content.

I really _really_ don't want us to add `<`/`>` around the nicks in our UI 😅.
This is something that always puzzled me in IRC graphical clients, why oh why do they still need, today in 2011, to add them?! It seems to me that it's a relic of the terminal clients and was added as a habit, but in terms of UI / UX, it doesn't make sense: there are tons of graphical ways to separate a nick from the message.
Case in point, no other messaging system that I know of does that: Telegram, Slack, Gitter, Messenger, WhatsApp, etc.
To me it would be a step backwards to add that in our UI. Unless, of course, I misunderstood your statement 😄

> It does mean that you can't select part of a message

So I've tried this PR locally, and this is really disturbing. If I select something and copies it to the clipboard, I really don't want to have any other content copied. It could be horrific for example with sensitive content.
Adding the `<`/`>` to the clipboard is different because it's not content, it's formatting, just like extra newlines are, pretty much. But then the question is: what do we do if user selects half the nick on top of the message? Should we just add `>`, the entire nick (:-1: because similarly, it's content not formatting), do nothing?
In the current state, selecting the end of a nick + message gets you an empty clipboard 😕

starquake:
> @starquake, could you fix the conflicts please? Sorry!

No problem. I stopped working on this for a while because I was hoping #1242 would fix it (or at least improve it), but it doesn't. I would prefer a solution without JS but I haven't figured out a way to do it yet.

> I really really don't want us to add </> around the nicks in our UI 😅.

Yeah you have a point there. Put that means the JS probably is needed.

> I really don't want to have any other content copied.

To me there are 2 cases:
1.  Selecting a part of a single message (in that case `user-select:all` is a no-go)
2. Selecting multiple messages

(1) should work just by copying the part that is selected as long as it doesn't span lines.

(2) 
I currently can think of 3 way to handle this. Take this selection:
image

The result could be one of these: (assuming I fixed issues with newlines)

A. (like Telegram) - note: needs to show more is being selected as illustrated by the GIF

```
11:46 <xPaw> create commit and force push
11:46 <starquake> No the fact that there comments on commits that do not exist anymore
11:46 <+YaManicKill> It just hides them.
11:47 <starquake> Sorry, won't do it again
11:48 <starquake> I'm new at all this :-P
```

B. (like WhatsApp)

```
11:46 <xPaw> force push
11:46 <starquake> No the fact that there comments on commits that do not exist anymore
11:46 <+YaManicKill> It just hides them.
11:47 <starquake> Sorry, won't do it again
11:48 <starquake> I'm new
```

C. (like Slack, and apps that do no formatting of copy to clipboard)

```
force push
11:46 <starquake> No the fact that there comments on commits that do not exist anymore
11:46 <+YaManicKill> It just hides them.
11:47 <starquake> Sorry, won't do it again
11:48 <starquake> I'm new
```

I like A most but it's the hardest, B is okay to (and doable). I think C is wrong.

I prefer the way Telegram (A) does this, for a couple of reasons
1. I think the times the user wants to select **complete** multiple messages is more frequent than a user wanting to select half of the first sentence and half of the last sentence. 
2. If the UI shows the user it's selecting the entire message I think that's okay
3. If you want to select complete sentences it's quite hard to do if it doesn't switch to selecting full sentences
4. If you paste the result it's a bit weird because you miss context 

> In the current state, selecting the end of a nick + message gets you an empty clipboard 😕

That's probably just a bug that needs fixing.

Would love to hear your opinions on this and I will see if I can make it happen

YaManicKill:
> I really really don't want us to add </> around the nicks in our UI .

It's not like we have made this feature up in our heads, plenty of people have actually put this into their custom css. So it is clearly wanted  ¯\_(ツ)_/¯ and as people have them in their UIs lots already, and we want it on copy, then it makes sense (to me) to put them in the UI normally so they are just copied.

> So I've tried this PR locally, and this is really disturbing. If I select something and copies it to the clipboard, I really don't want to have any other content copied.

Yeah, I agree and the only way I would be happy with this is if the UI shows the whole line as being selected. If you see less selected, then no I'm not ok with that.

dgw:
As far as `<`/`>` goes, the style I use for Textual has invisible brackets around the nicknames that are copied, but not displayed. (Well, my fork of it does… The original used to, but removed that feature.)

The brackets are placed in e.g. `<span class="swrapper">&lt;</span>`, then this CSS makes them visually disappear:

````CSS
body .senderContainer .swrapper {
  font-size: 0px !important;
  opacity: 0;
}
````

YaManicKill:
> @YaManicKill, would you be okay if we move this to the v2.4.0 milestone?

Sure, it is already in 2.3.1 as a bug so I have no big issues with it missing 2.3.2. I have the hack installed locally anyway ;-) (yay greasemonkey)

astorije:
> I like A most but it's the hardest, B is okay to (and doable). I think C is wrong.

I actually use Telegram on a daily basis, and I hate how it does select+copy. Copying something a user has not selected is just a horrible experience. This goes against what the user is actively trying to do (meaning it's not a passive action like a notification or something, user is doing one thing, app does something different).
I actually think C is the way to go: this is *exactly* the content that you selected on the app, there is no content added or removed.

That said, kudos on the analysis, that's great work, ideal way to drive the discussion!

> It's not like we have made this feature up in our heads, plenty of people have actually put this into their custom css. So it is clearly wanted ¯_(ツ)_/¯ and as people have them in their UIs lots already, and we want it on copy, then it makes sense (to me) to put them in the UI normally so they are just copied.

Nop, no and nope! 😅
Sorry to be so reluctant about this but this is a complete 👎 for me. Only UX-considerate reason to have this would be if there was no way to differentiate the nick and the message, which we do not have here. That's precisely why it's interesting to have it in the clipboard, since styling information gets lost.
But in the UI, this would be a complete step backward from a modern-looking (or modern-trying :D) client back to a terminal-looking or terminal-influenced client.

But I get that some people have styling preference and the fact that they can do it via CSS is a good thing. But if we add it in the content, then there is actually no way to remove it via CSS (and still, adding/removing the brackets shouldn't affect the copy-to-clipboard) so that's an ever harsher solution.

That being said, you might be onto an interesting solution: I couldn't find but is there a way to copy hidden text we could leverage here? So `<>` would be in the DOM, but made invisible with any of the following: `visibility: hidden`, `font-size: 0`, `width/height: 0`, `overflow: hidden`, `opacity: .001`, etc.
I don't know, just throwing ideas here.
I would really like to have `<>` in the clipboard, but as much as I'd like that, I know I would sacrifice them any day over adding them visually in the UI 😅.

...
...

Doh, I read @dgw's answer after typing my previous paragraph...

How does @dgw's solution sound, @starquake and @YaManicKill? @dgw, do you know why you needed both statements? It seems to me that only the `font-size` would suffice?

Either way, keep up the good work @starquake, it's great to have contributions like that. It's not an easy one but the fact that you keep on trying until we get it right makes it really enjoyable to spend time reviewing and commenting if that helps.

dgw:
> @dgw, do you know why you needed both statements? It seems to me that only the `font-size` would suffice?

I didn't write those styles, @april did. For all I know, it was just ass-covering. But I just ran a quick test using only `font-size` and saw no visible difference between this and the original (before I added the brackets & `span`s): https://jsfiddle.net/dgw/e27r5jvu/ — the angle brackets are copied if I select & copy the text, but not visible (at least in Chrome).

april:
Technically the CSS is only like 20% mine. I've avoided mucking with it as much as possible.

dgw:
I guess that's why that CSS wasn't removed when the hidden angle bracket functionality went AWOL… </thread_hijacking>

starquake:
> That said, kudos on the analysis, that's great work, ideal way to drive the discussion!

Thanks for not getting annoyed :-)

I dislike the solution C a lot. :-) Some other users dislike the other solutions. But it looks like @astorije and @YaManicKill agree on solution C being the best. How about I implement solution C in this PR and improve the copy pasting later by adding a user setting in another PR? Or is this not important enough to create a setting for it?

I will also investigate ways to prevent </> from being copied. Not too happy about `font-size:0` but it might be the best way. If that's the case I will happily implement @dgw's solution!

YaManicKill:
> How about I implement solution C in this PR

Yes please

> and improve the copy pasting later by adding a user setting in another PR?

I'm not sure this is something we want to add a setting for, it feels *very* niche. But we can see.

> I will also investigate ways to prevent </> from being copied. 

It's not about them not being copied, it's about them not being visible on the UI, but being copied. Then themes could just change the css to make them visible if they want.

starquake:
> It's not about them not being copied, it's about them not being visible on the UI, but being copied. Then themes could just change the css to make them visible if they want.
DERP, sorry didn't have much sleep last night. It's so hot here.

starquake:
Another experiment. 

- Chrome will probably not fix the way they copy flexbox stuff, and we want to add </>
- We might want to skip copying things like toggles

I propose decorating elements we want to copy with a 'clipboard' and/or 'clipboard-only' class. Naming is up for debate.

This commit implements it for msg.tpl.

Would love some feedback.

Komic:
This would be a great change, right now copy-pasting stuff is kind of a frustrating experience, thanks!

xPaw:
@starquake Do you want to continue with this PR?

xPaw:
Superseded by #1464.

PR #1233, opened 2017-06-16
Make everything unselectable by default

xPaw:
... then specifies which elements should be selectable (help page, channel and topic in header, message container in channels).

This should feel a little more like a native app (Discord does this too). Less noise when doing ctrl+a ctrl+c.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -40,6 +40,12 @@ body {
 	font: 16px Lato, sans-serif;
 	margin: 0;
 
+	-webkit-user-select: none;
+	-moz-user-select: none;
+	-ms-user-select: none;
+	user-select: none;
+	cursor: default;
+
 	/**
 	  * Disable pull-to-refresh on mobile that conflicts with scrolling the message list.
 	  * See http://stackoverflow.com/a/29313685/1935861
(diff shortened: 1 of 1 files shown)


williamboman:
> This should feel a little more like a native app (Discord does this too).

Might be worth trying out removing most, if not all, `cursor: pointer` too.

xPaw:
@williamboman Any example? I can't think of anything where it does that now.

williamboman:
> @williamboman Any example? I can't think of anything where it does that now.

I mean links, buttons and other elements in general that defaults to `cursor: pointer`. Native desktop applications rarely use the pointer cursor in my experience. I think it's pretty clean with pointer cursors removed.

```css
a, button, .chan {
  cursor: default !important;
}
```

xPaw:
On that I would disagree, pointer cursor is clearly an action, and should be kept that way. I will however add `cursor: default` on body, so that unselectable text doesn't get a text cursor.

client/css/style.css
```diff
@@ -40,6 +40,12 @@ body {
 	font: 16px Lato, sans-serif;
 	margin: 0;
 
+	-webkit-user-select: none;
+	-moz-user-select: none;
+	-ms-user-select: none;
+	user-select: none;
```
YaManicKill: As a side-note, we should add in auto-prexifer...

astorije:
I'll test this to double-check some usability concerns, but I'm thinking this should be fine. I'll do that in a day or 2.

PR #1235, opened 2017-06-18
Promisify reverseDnsLookup

metsjeesus:
So you can use the method somewhere else too, maybe on webirc on some extreme cases.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -297,15 +297,20 @@ function init(socket, client, generateToken) {
 
 function reverseDnsLookup(socket, client) {
 	client.ip = getClientIp(socket.request);
-
-	dns.reverse(client.ip, function(err, host) {
-		if (!err && host.length) {
-			client.hostname = host[0];
-		} else {
-			client.hostname = client.ip;
-		}
-
-		init(socket, client);
+	return new Promise((resolve, reject) => {
+		dns.reverse(client.ip, (err, host) => {
+			if (!err) {
+				if (host.length) {
+					client.hostname = host[0];
+					resolve();
+				} else {
+					client.hostname = client.ip;
+					resolve();
+				}
+			} else {
+				reject(err);
+			}
+		});
 	});
 }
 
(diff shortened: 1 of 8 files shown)


src/server.js
```diff
@@ -373,7 +378,14 @@ function auth(data) {
 		// If webirc is enabled and we do not know this users IP address,
 		// perform reverse dns lookup
 		if (Helper.config.webirc !== null && !client.config.ip) {
-			reverseDnsLookup(socket, client);
+			reverseDnsLookup(socket, client)
+				.then(()=>{
+					init(socket, client);
+				})
+				.catch(e=>{
```
astorije: I believe `=>` should be surrounded with whitespaces. Once #1231 is merged, rebase on `master`, which I think should detect this :)

astorije:
@metsjeesus, why do you still have xPaw's commit here and in your other PR? Also, are you up-to-date with master?

metsjeesus:
Its only 1 method, reverseDnsLookup. Old code has some auth logic in it, what made it useless to use it anywhere else, example on #986 .  Main point is, its cleaner to look and undestand what auth does, its not split up on many places.

xPaw:
I'm gonna agree with @astorije for now. This doesn't provide any direct value. I will take another look at this after session stuff is redone. Login/session code is somewhat critical and subtle issue can easily slip in, so I will close this for now.

Thanks for your help, I'll poke you on IRC on the next release.

astorije:
@metsjeesus, just so you don't take it the wrong way: thanks for your help, we do appreciate it a lot!

Essentially, I believe this PR is doing 2 things: converting async stuff to using promises, and something else. In my understanding, this "something else" is the answer to the underlying problem, not the promisifying, but I'm not sure to understand what is the underlying problem.
Regarding promisifying things, I am for one a big advocate to promises as opposed to callbacks, but in this current case, I don't think they are correctly used here (they are useful when used as values, not functions with no return value + side-effects).

Just like @xPaw said, let's look at this again when the session stuff is ready, there is a lot of crossover right now and it gets confusing and tricky (because of critical component like @xPaw mentioned).

metsjeesus:
@astorije and @xPaw , its all fine. Busy times, many things on the works, understandable. Focus is currently on clinet side anyway.

PR #1240, opened 2017-06-20
Improve the PR tester script a bit

astorije:
I know it's probably not the cleanest / most error-proof solution, but it does the job in most cases. What I wanted to add most obviously is the `git rebase master` part, to make sure what we are testing with these PRs is as up-to-date with `master` as possible.

Of course, if your `master` is not up-to-date itself, there will be troubles, but devs' `master`s should always be up-to-date... right? 😉

Anyway, I'm using this myself a lot to test PRs when reviewing, so I'd surely appreciate those!

### DIFF
--- a/scripts/run-pr.sh
+++ b/scripts/run-pr.sh
@@ -10,6 +10,6 @@ fi
 git fetch https://github.com/thelounge/lounge.git refs/pull/${1}/head
 git checkout FETCH_HEAD
 npm install
-npm run build
+NODE_ENV=production npm run build
 npm test || true
 npm start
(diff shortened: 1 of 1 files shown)


astorije:
> but I doubt many people other than you use this script @astorije, so go for it.

Yeah, and I'd be totally willing to improve that script if other reviewers find this useful (how can you review without it, I literally use it multiple times a week 😅), I'd be totally happy to make it more resilient.

> Just a think to remember that if the rebase fails, the script will continue to run the other commands.

Actually, it would stop thanks to the `set -e` at the top, which is exactly why I'm not too concerned about having it fail.

YaManicKill:
> how can you review without it

I mean, I have an equivalent, but it's not exactly the same. It's just a 1-liner.

> Actually, it would stop thanks to the set -e at the top

Huh, TIL

PR #1237, opened 2017-06-19
Correctly show spaces in messages

starquake:
This tries to fix #1229.

Issue #1229, opened 2017-06-15
Multiple spaces in messages not shown as such


PolarizedIons:
When there are multiple spaces in a message (like in motd ASCII art), lounge only shows it as one, messing up the alignment.

Example: lounge on the left and hexchat on the right, showing EsperNet's motd
![](https://i.imgur.com/jOupDen.png)

DanielOaks:
This also affects the MAP command particularly – since multiple spaces don't get displayed as such, it pretty much becomes unreadable :P

YaManicKill:
Well, it's html, so no matter how many spaces you have next to it, the browser will show it as 1 space.

We'll want to replace them with non-breaking spaces.

starquake:
I think that breaks word-wrapping. May I suggest another fix? I think the correct way would be to use:
```
#chat .text {
  white-space: pre-wrap;
}
```
From MDN
[Sequences of whitespace are preserved. Lines are broken at newline characters, at <br>, and as necessary to fill line boxes.](https://developer.mozilla.org/en/docs/Web/CSS/white-space?v=control)

But that means all the texts going in that SPAN can not have newlines that aren't supposed to be there.

mestaritonttu:
I tested with [KiwiIRC's next client](https://kiwiirc.com/nextclient/) and it has:

```
.kiwi-messagelist-body {
    display: block;
    white-space: pre-wrap;
    word-wrap: break-word;
        overflow-wrap: break-word;
    margin-left: 130px;
}
```

astorije:
I could swear we've had this issue before, and fixed it! Is there someone courageous to look through The Lounge / Shout commit / PR histories? :D

> I tested with KiwiIRC's next client

(Interesting how in some ways it _really_ looks like The Lounge visually :D)

YaManicKill:
> I could swear we've had this issue before, and fixed it! 

Yes we did, and it just came full circle. It was fixed in #239 

And I tried adding it back on locally and found ... that the bug we had with the table layout was caused by this (where it would wrap unnecessarily). It was then removed in #523 (specifically [this commit](https://github.com/thelounge/lounge/pull/523/commits/fb672ab57ff2a6d5ae8a54b2acef64d4e4354513)) so that makes sense.

starquake:
@YaManicKill That happens because with pre-wrap, newlines get meaning and there are newlines in that part. And the newlines are there because there are newlines in the .tpl

[...]

PR #1242, opened 2017-06-20
Correctly show whitespace and newlines in messages

starquake:
Screwed up last PR so here's a clean one. Fix for https://github.com/thelounge/lounge/issues/1229

Extension of https://github.com/thelounge/lounge/pull/1237

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -2085,3 +2085,19 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 #image-viewer .open-btn {
 	margin: 0 auto 10px;
 }
+
+/* Correctly handle multiple successive whitespace characters.
+   For example: user has quit ( ===> L   O   L <=== )  */
+
+#windows .header .topic,
+#chat .message .text,
+#chat .motd .text,
+#chat .notice .text,
+#chat .ctcp-message,
+#chat .part-reason,
+#chat .quit-reason,
+#chat .new-topic,
+#chat .action-text,
+#chat table.channel-list .topic {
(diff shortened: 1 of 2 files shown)


starquake:
Fixing the lint issues...

YaManicKill:
I'll review this tomorrow.

xPaw:
So I don't think it's worth the trouble to apply `white-space` on every message type. Just `notice`, `motd` and `message` would be enough I think. Then modifying templates for different actions wouldn't be required.

starquake:
But... how about some one doing:`/kick You got kicked =>  L  O  L  <=` or `/quit some    fancy   quit 
  message   with   spaces`

xPaw:
@starquake Quit message has `quit-reason` class for example, could apply on it. I just think applying on `.text` as a generic will always bring us problems with whitespace in templates.

starquake:
Mmmm yeah maybe we should. I see ctcpMessage has no class though. Is it okay to add one there? Any suggestions for a name?

xPaw:
Yes, we can safely add classnames on all the stuff that has parsed messages. `ctcp` would be it's own action and can be matched like '.msg.ctcp .text`.

starquake:
There's a newline in the .ctcp .text so this will become visible. Here's the ctcp template:
```
{{> ../user_name nick=from}}
<b>{{ctcpType}}</b> {{{parse ctcpMessage}}}
```
Do you want me to:

1.  change the template
2. add a class in the template

xPaw:
Wrapping ctcpMessage in a span should be good.

PR #1244, opened 2017-06-20
Correctly finish scroll animation when using page keys

### LINKED ISSUE #1022: Holding pageup/pagedown scrolls very slowly

xPaw:
Fixes #1022.

> When .finish() is called on an element, the currently-running animation and all queued animations (if any) immediately stop and their CSS properties set to their target values. All queued animations are removed.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -1306,14 +1306,13 @@ $(function() {
 			"pagedown"
 		], function(e, key) {
 			let container = windows.find(".window.active");
-			if (container.is(":animated")) {
-				return;
-			}
 
 			// Chat windows scroll message container
 			if (container.attr("id") === "chat-container") {
 				container = container.find(".chan.active .chat");
 			}
+
+			container.finish();
 
 			const offset = container.get(0).clientHeight * 0.9;
 			let scrollTop = container.scrollTop();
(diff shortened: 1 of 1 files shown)


client/js/lounge.js
```diff
@@ -1306,15 +1306,14 @@ $(function() {
 			"pagedown"
 		], function(e, key) {
 			let container = windows.find(".window.active");
-			if (container.is(":animated")) {
```
xPaw: This `if` should have been lower by the way, so it doesn't really work anyway.

YaManicKill:
@astorije I'm not bothered, I just want one of us to test it. If you get to it first, feel free. I'll let you know if I manage to get some time to do it this evening or not.

astorije:
Anything you think can be improved, @xPaw?

xPaw:
Initially I wanted to get jQuery's animation state, so I could get final scrollTop target, stop the animation and start a new one with adjusted scrollTop, but there jQuery provides no way of doing that. I could be storing scroll data myself to re-use, but I don't know if it's worth it?

astorije:
> I could be storing scroll data myself to re-use, but I don't know if it's worth it?

It's probably worth a try, but if this PR already improves things, that's a win IMO. I'm going to merge this as-is, and feel free to attempt something better (but probably more experimental) for a future improvement.

Either way, thanks for that!

PR #1246, opened 2017-06-21
Handle images with unknown size in prefetch

bews:
Was discussed in #1165

Related: #1162

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -93,11 +93,12 @@ function fetch(url, cb) {
 		return;
 	}
 	var length = 0;
-	var limit = 1024 * 10;
+	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/(text\/html|application\/json)/.test(res.headers["content-type"]))) {
-				res.req.abort();
+			if (!(/(image\/.+)/.test(res.headers["content-type"]))) {
+				// if not image, limit download to 10kb, since we need only meta tags
+				limit = 1024 * 10;
 			}
 		})
 		.on("error", function() {})
@@ -113,14 +114,13 @@ function fetch(url, cb) {
 				return;
 			}
 
-			var body;
-			var type;
-			var size = req.response.headers["content-length"];
-			try {
-				body = JSON.parse(data);
-			} catch (e) {
-				body = {};
+			let type;
+			let size = parseInt(req.response.headers["content-length"], 10) || length;
+
+			if (size < length) {
+				size = length;
 			}
+
 			try {
 				type = req.response.headers["content-type"].split(/ *; */).shift();
 			} catch (e) {
@@ -128,7 +128,6 @@ function fetch(url, cb) {
 			}
 			data = {
 				text: data,
-				body: body,
 				type: type,
 				size: size
 			};


src/plugins/irc-events/link.js
```diff
@@ -64,7 +64,7 @@ function parse(msg, url, res, client) {
 	case "image/gif":
 	case "image/jpg":
 	case "image/jpeg":
-		if (res.size < (Helper.config.prefetchMaxImageSize * 1024)) {
+		if (res.text.length < (Helper.config.prefetchMaxImageSize * 1024)) {
```
xPaw: This means `content-length` is not used at all now. Should it be `res.size || res.text.length` maybe?

bews: I think it's better to check actual size (since we have it now) rather than trusting the header, which may have invalid value.

YaManicKill: It that guaranteed to be there?

bews: Yeah, should be.

src/plugins/irc-events/link.js
```diff
@@ -93,13 +93,8 @@ function fetch(url, cb) {
 		return;
 	}
 	var length = 0;
-	var limit = 1024 * 10;
+	var limit = Helper.config.prefetchMaxImageSize * 1024;
```
xPaw: In default config, this change will fetch up to 512kb of HTML content instead of the 10kb. Can we keep the `"response"` callback, but instead of aborting the request, reduce the `limit`, so that we still only fetch 10kb of HTML, and `prefetchMaxImageSize` of unknown content?

src/plugins/irc-events/link.js
xPaw: `text.length` would be whatever fetched content length, so it will be there, yes. What we can do is check if `content-length` is under `prefetchMaxImageSize` and compare it to `text.length` to see if it's true. If it's not, then do not send the image to client.

src/plugins/irc-events/link.js
bews: I don't get why HTML should be treated differently. Is there a reason to do so?

xPaw: Because we only need to fetch some meta tags from the page, which are at the top, so 10kb works fine for that purpose.

src/plugins/irc-events/link.js
YaManicKill: That feels overly complicated. If the text will always be there, and it'll always be the actual length, then why bother with the header?

src/plugins/irc-events/link.js
bews: >we only need to fetch some meta tags from the page, which are at the top, so 10kb works fine for that purpose.

I believe preview won't load at all if content > 10kb:

```javascript
if (length > limit) {
	req.response.req.abort();
}
```

xPaw: Why not?

bews: Right, I've missed `response`. It's done.

src/plugins/irc-events/link.js
```diff
@@ -93,11 +93,11 @@ function fetch(url, cb) {
...
 	req
 		.on("response", function(res) {
-			if (!(/(text\/html|application\/json)/.test(res.headers["content-type"]))) {
-				res.req.abort();
+			if ((/(text\/html|application\/json)/.test(res.headers["content-type"]))) {
+				limit = 1024 * 10;
```
xPaw: Can you please add an explanatory comment here

src/plugins/irc-events/link.js
xPaw: Because the connection might timeout or drop, so our `text` can be partial (e.g. received 15kb but content-length is 300kb)

bews:
Fixed everything, should be good now.

src/plugins/irc-events/link.js
```diff
@@ -115,7 +116,7 @@ function fetch(url, cb) {
 
 			var body;
 			var type;
-			var size = req.response.headers["content-length"];
+			var size = parseInt(req.response.headers["content-length"]);
```
xPaw: Add `10` as second param here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description

src/plugins/irc-events/link.js
```diff
@@ -93,11 +93,12 @@ function fetch(url, cb) {
...
 	}
 	var length = 0;
-	var limit = 1024 * 10;
+	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/(text\/html|application\/json)/.test(res.headers["content-type"]))) {
-				res.req.abort();
+			if (!(/(image\/.+)/.test(res.headers["content-type"]))) {
+				// if not image, limit download to 10kb, since we need only meta tags
```
YaManicKill: This might have to change if we end up proxying videos, etc, which we might end up doing. But for now this looks good.

src/plugins/irc-events/link.js
astorije: I'm unclear here: @bews, @xPaw, did that line need further changing?

xPaw: @astorije this PR is ready.

PR #1249, opened 2017-06-21
Use stylelint-config-standard

xPaw:
We're surprisingly quite standard except for three rules.


client/css/style.css
```diff
@@ -1722,10 +1721,7 @@ kbd {
 	border-right-color: #222;
 }
 
-@media only screen and (-webkit-min-device-pixel-ratio: 2),
```
YaManicKill: Are these no longer needed?

.editorconfig
```diff
@@ -6,6 +6,7 @@ root = true
 
 [*]
 indent_style = tab
+tab_width = 4
```
YaManicKill: Such big tabs. Weird people.

xPaw: This was mainly for github (it uses editor config), and it defaults to 8 tabs.

client/css/style.css
xPaw: `resolution` is the way to do it, min-device-pixel-ratio is deprecated.

.editorconfig
YaManicKill: Alright, 4 is definitely better than 8.

astorije: Why not 2 to match L20 and former L24?

xPaw:
Removed `tab-width` from this PR.

PR #1253, opened 2017-06-22
Fix nick fade positioning

xPaw:
Noticed this in my theme.


astorije:
Merging since super trivial fix. Thanks!

PR #1254, opened 2017-06-22
Make sure thumbnail is a valid image

### LINKED ISSUE #1180: Link prefetch 404 will still get pushed to the client

### LINKED ISSUE #1239: Enforce prefetchMaxImageSize on preview thumbnails
@bews noticed that thumbnails are not checked against this maximum size, as opposed to the picture previews.

He included a fix as a side-effect of https://github.com/thelounge/lounge/pull/1190 but the overall feature will not be part of the codebase, effectively losing this fix.

If anyone is willing to help, this should be a fairly easy fix.

xPaw:
Fixes #1239.
Fixes #1180.


PolarizedIons:
Can confirm it fixes #1180

PR #1256, opened 2017-06-22
Check og:title before title tag

xPaw:
Should play a lot more nicer with `og:description`.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -49,7 +49,10 @@ function parse(msg, url, res, client) {
 	case "text/html":
 		var $ = cheerio.load(res.text);
 		toggle.type = "link";
-		toggle.head = $("title").text();
+		toggle.head =
+			$("meta[property=\"og:title\"]").attr("content")
+			|| $("title").text()
+			|| "No title found.";
 		toggle.body =
 			$("meta[name=description]").attr("content")
 			|| $("meta[property=\"og:description\"]").attr("content")


src/plugins/irc-events/link.js
```diff
@@ -49,7 +49,10 @@ function parse(msg, url, res, client) {
 	case "text/html":
 		var $ = cheerio.load(res.text);
 		toggle.type = "link";
-		toggle.head = $("title").text();
+		toggle.head =
+			$("meta[property=\"og:title\"]").attr("content")
+			|| $("title").text()
+			|| "No title found.";
```
astorije: I would actually reword this in `Untitled page` rather than `No title found.` The latter displays like an error message while the former presents a fact about the page. Nitpick, I know.

And because I know this is meant to be consistent with `No description found.`, I would simply remove it. It looks sloppy. If there is no description, let's not show any, it's leaner and doesn't make things obfuscated either.

xPaw: Let's instead refuse to show info if there is no preview image/title/description? Or only show "Untitled page" if there is an image. This will need to done in a separate PR though.

astorije: Very reasonable.

astorije:
Unfortunately, there are cases where this is not a clear win. For example, when entering https://github.com/rails/rails/search?utf8=%E2%9C%93&q=def+sanitize&type=, one gets:

Before | After
--- | ---
screen shot 2017-07-03 at 01 45 14 | screen shot 2017-07-03 at 01 45 27

The first one comes with more details.
... Also, what's up with that cut width?!

astorije:
Oh my, this width is painful:

On a recent release:

screen shot 2017-07-03 at 01 56 25

On `master`:

screen shot 2017-07-03 at 01 56 41

Completely off-topic of this PR obviously, but have you looked into this by any chance, @xPaw?

Issue #1176, opened 2017-05-20
Insert a comma, colon, or other user-configurable string when tab-completing nicks


cmr:
This would only fire at the start of a line though. Lacking this makes it rather hard for me to get used to lounge.

astorije:
Hi @cmr, welcome on The Lounge!

We have discussed in the past and have decided to keep it that way on purpose. Not adding or adding something after the nick would always offend someone, preferring nothing (I am a strong supporter of the current way) or a different character. Keeping it that way lets people add whatever they want instead to have to backspace.
In the end, it's just a matter of habit, and people get used to it very quickly.

Making this configurable, I'm not hugely in favor of adding an option just for that, but I could be convinced if people really want this. This however requires an overhaul of the settings, so this is not possible at the moment.

clandmeter:
Adding my comment here as I would like to have support for this with some config option.

xPaw:
@astorije, I think we should give in and make a setting for this which would default to empty (e.g. current behaviour). This should be relatively simple to implement, and there's a demand for it.

YaManicKill:
I agree with @xPaw 

Whether this should be in an advanced page, or whether it should just be in the normal settings is something we can bikeshed on this, but I think this should definitely be a feature. Happy to default it to empty, though.

Fwiw, the functionality should only insert the seperator if it is used on the first word in the input, as it is generally accepted behaviour, and people don't want it to do it in the middle of a sentence.

astorije:
As long as it defaults to the current state, I think I am ready to give up. Always amazed by this request though, literally no other messaging system I know let you configure that (Telegram, Slack, Facebook, Hangout, you name it) and yet us lacking it is described as an essential issue 😂 

> Fwiw, the functionality should only insert the seperator if it is used on the first word in the input, as it is generally accepted behaviour, and people don't want it to do it in the middle of a sentence.

So, this is why I was not willing to care about this feature: I have seen all sorts of requests for this. Very quickly, people will ask us 2 fields, one for the first nick-as-word in a message, one for the rest, or to make it obscure like a special format in a single field to support both, or a checkbox "Only for first word", etc. This is very much overkill IMO, especially since it's something you get used to in like 2 days top.

We have so much to do and to maintain already, it pains me that an extra character is such a big deal. But whatever, I won't block this obviously.

YaManicKill:
> literally no other messaging system I know let you configure that

[...]

PR #1262, opened 2017-06-23
Add : to nick tab/text complete if it's the first word in the message box

MaxLeiter:
I know the discussion in #1176 is mostly towards a setting being added, but I'm against that. Made this for personal use, but figured I'd open a PR if it's what people want.

### DIFF
--- a/client/js/libs/jquery/tabcomplete.js
+++ b/client/js/libs/jquery/tabcomplete.js
@@ -152,6 +152,11 @@ import jQuery from "jquery";
 
 					// Update element with the completed text.
 					var text = value.substr(0, self[0].selectionStart - last.length) + word;
+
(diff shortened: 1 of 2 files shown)


astorije:
> I know the discussion in #1176 is mostly towards a setting being added, but I'm against that.

I'm going to close this PR for this reason. This is not where we are headed, and I would very strongly 👎 work that would change the current default behavior.
Thanks for sharing your stuff though, very useful for anyone who wants to do just like you! :)

astorije:
FYI, closing this PR doesn't mean it cannot be used by others so nothing got lost when I closed it.
Anyone can still retrieve it using [tricks explained here](https://help.github.com/articles/checking-out-pull-requests-locally/).

PR #1263, opened 2017-06-23
Add common emoji shortcuts/ascii faces to emoji map

MaxLeiter:
Added some emoji shortcuts (:) -> 😃, :( ->🙁, etc).

One downside, and I'm not sure there's a fix which is probably a dealbreaker, is they're ugly when they show up in the autocomplete dropdown:

image


astorije:
I don't know if this what this PR does exactly, but I personally hate when a client replaces ":)", ":(" with emoji. It used to be very fashionable some time ago, but much less so since the `:emoji:` syntax became a de facto standard.
And even clients who do that (well, those I know) either have an option to disable it, or it's disable by default.

MaxLeiter:
astorije: it doesn't replace automatically, it acts the same as regular replacing. If the user hits enter on the emoji, it will change to emoji. Otherwise it will remain `:)` or whatever they type.

YaManicKill:
Why did you close this, @MaxLeiter? I like the idea of it. My main question would be whether it should be :): or ::):

MaxLeiter:
@YaManicKill - couldn't think of a nice UX

PR #1266, opened 2017-06-24
Allow themes from npm

YaManicKill:
This has a bit of work left to do, but it works. Woop.

- [x] Clean up code
- [x] Documentation on how to create a theme for the lounge
- [x] Figure out how I can get rid of the multiple names...
- [x] Get packages from `$LOUNGE_HOME/packages` instead of globally/locally
- [x] rename "plugins"/"themes" to "packages" (except where being a theme is useful info)

screenshot from 2017-06-24 16-20-02

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -264,8 +264,8 @@
 								<label for="theme-select" class="sr-only">Theme</label>
 								<select id="theme-select" name="theme" class="input">
 									{{#each themes}}
-										<option value="{{filename}}">
-											{{name}}
+										<option value="{{name}}">
+											{{displayName}}
 										</option>
 									{{/each}}
 								</select>
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -51,11 +51,12 @@ module.exports = {
 
 	//
 	// Set the default theme.
+	// Find out how to add new themes at https://thelounge.github.io/docs/packages/themes
 	//
 	// @type     string
-	// @default  "themes/example.css"
+	// @default  "example"
 	//
-	theme: "themes/example.css",
+	theme: "example",
 
 	//
 	// Prefetch URLs
--- a/src/helper.js
+++ b/src/helper.js
@@ -12,6 +12,8 @@ const colors = require("colors/safe");
 var Helper = {
 	config: null,
 	expandHome: expandHome,
+	getPackagesPath: getPackagesPath,
+	getPackageModulePath: getPackageModulePath,
 	getStoragePath: getStoragePath,
 	getUserConfigPath: getUserConfigPath,
 	getUserLogsPath: getUserLogsPath,
@@ -94,6 +96,14 @@ function getUserLogsPath(name, network) {
 
 function getStoragePath() {
 	return path.join(this.HOME, "storage");
+}
+
+function getPackagesPath() {
+	return path.join(this.HOME, "packages", "node_modules");
+}
+
+function getPackageModulePath(packageName) {
+	return path.join(Helper.getPackagesPath(), packageName);
 }
 
 function ip2hex(address) {
--- a/src/plugins/themes.js
+++ b/src/plugins/themes.js
@@ -1,0 +1,85 @@
+"use strict";
+
+const fs = require("fs");
+const Helper = require("../helper");
+const colors = require("colors/safe");
+const path = require("path");
+const _ = require("lodash");
+const themes = new Map();
+
+module.exports = {
+	getAll: getAll,
+	getFilename: getFilename
+};
+
+fs.readdir("client/themes/", (err, builtInThemes) => {
+	if (err) {
+		return;
+	}
+	builtInThemes
+		.filter((theme) => theme.endsWith(".css"))
+		.map(makeLocalThemeObject)
+		.forEach((theme) => themes.set(theme.name, theme));
+});
+
+fs.readdir(Helper.getPackagesPath(), (err, packages) => {
+	if (err) {
+		return;
+	}
+	packages
+		.map(makePackageThemeObject)
+		.forEach((theme) => {
+			if (theme) {
+				themes.set(theme.name, theme);
+			}
+		});
+});
+
+function getAll() {
+	return _.sortBy(Array.from(themes.values()), "displayName");
+}
+
+function getFilename(module) {
+	if (themes.has(module)) {
+		return themes.get(module).filename;
+	}
+}
+
+function makeLocalThemeObject(css) {
+	const themeName = css.slice(0, -4);
+	return {
+		displayName: themeName.charAt(0).toUpperCase() + themeName.slice(1),
+		filename: path.join(__dirname, "..", "client", "themes", `${themeName}.css`),
+		name: themeName
+	};
+}
+
+function getModuleInfo(packageName) {
+	let module;
+	try {
+		module = require(Helper.getPackageModulePath(packageName));
+	} catch (e) {
+		log.warn(`Specified theme ${colors.yellow(packageName)} is not installed in packages directory`);
+		return;
+	}
+	if (!module.lounge) {
+		log.warn(`Specified theme ${colors.yellow(packageName)} doesn't have required information.`);
+		return;
+	}
+	return module.lounge;
+}
+
+function makePackageThemeObject(moduleName) {
+	const module = getModuleInfo(moduleName);
+	if (!module || module.type !== "theme") {
+		return;
+	}
+	const modulePath = Helper.getPackageModulePath(moduleName);
+	const displayName = module.name || moduleName;
+	const filename = path.join(modulePath, module.css);
+	return {
+		displayName: displayName,
+		filename: filename,
+		name: moduleName
+	};
+}
--- a/src/server.js
+++ b/src/server.js
@@ -14,6 +14,7 @@ var Helper = require("./helper");
 var colors = require("colors/safe");
 const net = require("net");
 const Identification = require("./identification");
+const themes = require("./plugins/themes");
 
 // The order defined the priority: the first available plugin is used
 // ALways keep local auth in the end, which should always be enabled.
@@ -50,6 +51,15 @@ module.exports = function() {
 		}))
 		.set("view engine", "html")
 		.set("views", path.join(__dirname, "..", "client"));
+
+	app.get("/themes/:theme.css", (req, res) => {
+		const themeName = req.params.theme;
+		const theme = themes.getFilename(themeName);
+		if (theme === undefined) {
+			return res.status(404).send("Not found");
+		}
+		return res.sendFile(theme);
+	});
 
 	var config = Helper.config;
 	var server = null;
@@ -191,16 +201,13 @@ function index(req, res, next) {
 		pkg,
 		Helper.config
 	);
+	if (!data.theme.includes(".css")) { // Backwards compatibility for old way of specifying themes in settings
+		data.theme = `themes/${data.theme}.css`;
+	} else {
+		log.warn(`Referring to CSS files in the ${colors.green("theme")} setting of ${colors.green(Helper.CONFIG_PATH)} is ${colors.bold("deprecated")} and will be removed in a future version.`);
+	}
 	data.gitCommit = Helper.getGitCommit();
-	data.themes = fs.readdirSync("client/themes/").filter(function(themeFile) {
-		return themeFile.endsWith(".css");
-	}).map(function(css) {
-		const filename = css.slice(0, -4);
-		return {
-			name: filename.charAt(0).toUpperCase() + filename.slice(1),
-			filename: filename
-		};
-	});
+	data.themes = themes.getAll();
 
 	const policies = [
 		"default-src *",


src/plugins/themes.js
```diff
@@ -0,0 +1,65 @@
...
+Helper.config.themes.map(getModuleInfo)
+	.filter((name) => name !== undefined)
+	.forEach((theme) => themes[theme.module] = theme);
+
+function get() {
+	return Object.keys(themes).map((key) => themes[key]);
+}
+
+function fileName(module) {
+	return themes[module].filename;
```
xPaw: This is gonna crash if `module` isn't in `themes`.

src/plugins/themes.js
```diff
@@ -0,0 +1,65 @@
...
+		.map(makeThemeObject)
+		.forEach((theme) => themes[theme.name] = theme);
+});
+
+Helper.config.themes.map(getModuleInfo)
+	.filter((name) => name !== undefined)
+	.forEach((theme) => themes[theme.module] = theme);
+
+function get() {
+	return Object.keys(themes).map((key) => themes[key]);
```
xPaw: Maybe make `themes` an actual `Map`?

src/plugins/themes.js
```diff
@@ -0,0 +1,65 @@
...
+	} catch (e) {
+		log.warn(`Specified theme ${colors.yellow(moduleName)} is not installed globally`);
+		return;
+	}
+	const modulePath = requireg.resolve(moduleName);
+	modulePath.substring(0, modulePath.lastIndexOf("/"));
+	return {
+		module: moduleName,
+		name: module.name,
+		filename: `${modulePath.substring(0, modulePath.lastIndexOf("/"))}/${module.css}`,
```
xPaw: What's `module.css` in this case? Is this exports from the plugin?

src/plugins/themes.js
```diff
@@ -0,0 +1,65 @@
...
+		module = requireg(moduleName);
+	} catch (e) {
+		log.warn(`Specified theme ${colors.yellow(moduleName)} is not installed globally`);
+		return;
+	}
+	const modulePath = requireg.resolve(moduleName);
+	modulePath.substring(0, modulePath.lastIndexOf("/"));
+	return {
+		module: moduleName,
+		name: module.name,
```
xPaw: Add `|| moduleName`.

xPaw:
We should totally make the existing themes their own packages 💅

src/plugins/themes.js
xPaw: Also before returning, make sure this file actually exists. Use `path.join`.

src/plugins/themes.js
YaManicKill: Good point, will make it more safe.

src/plugins/themes.js
YaManicKill: Huh yeah, that would work, I see node 4 supports map.values, so that's good.

src/plugins/themes.js
YaManicKill: module is the exports from the plugin, and `css` is the filename for the css. Would you prefer a different name?

YaManicKill:
> We should totally make the existing themes their own packages :nail_care:

If we do that, we'll need to either:

1. Install them as dependencies in lounge, and allow themes to be local packages (currently it only works with global packages)
2. Require people to install the default themes

Thoughts?  ¯\\\_(ツ)_/¯

src/plugins/themes.js
YaManicKill: > Also before returning, make sure this file actually exists. Use path.join and path.normalize.

What should it do if the file doesn't exist? Return undefined? Return an object with no filename? Return something that will show an error in the frontend?

xPaw: Print an error to console and don't add it to `themes`.

xPaw: Also about exports, this technically should work? https://github.com/thelounge/lounge-theme-solarized/blob/d385c0811fa105cb8fd55a1d0e57492b47545dee/package.json#L5-L6

YaManicKill: > Print an error to console and don't add it to themes.

Alright, will do

> Also about exports, this technically should work? 

Well, I'm currently reading from the main file, it's harder to read the package.json (although possible).

I was thinking the main js file just was something like this:

```
module.exports = {
        type: "theme",
        name: "Solarized",
        css: "theme.css"
}
```

Allows you to have a different visible name than the module name if you wanted, and means we don't have to worry about other name clashes in the package.json.

xPaw: Note how I made `main` the `package.json`, technically `require` should load and export it.

YaManicKill: Ah, I hadn't noticed that. Yeah, it works, guess it'll be nicer. Probably best to do something like:

lounge: {
type: "theme",
css: "theme.css"
}

rather than having things as top level items.

YaManicKill:
Added themes docs thelounge/thelounge.github.io#69

YaManicKill:
Alright, I've done everything except making sure the css file exists.

The reason I haven't done that yet is because it'll involve quite a bit of re-architecting how I've done it due to needing to change it to async. It's currently synchronous. So I'll do that sometime, but apart from that it works really nicely.

package.json
```diff
@@ -54,6 +54,7 @@
     "moment": "2.18.1",
     "read": "1.0.7",
     "request": "2.81.0",
+    "requireg": "^0.1.6",
```
xPaw: Remove `^`

src/plugins/themes.js
```diff
@@ -0,0 +1,79 @@
...
+	}
+	dirThemes
+		.filter((theme) => theme.endsWith(".css"))
+		.map(makeLocalThemeObject)
+		.forEach((theme) => themes.set(theme.name, theme));
+});
+
+Helper.config.plugins
+	.map(getModuleInfo)
+	.filter((module) => module !== undefined)
```
xPaw: You're iterating same array 6 times here, I understand ES6 is nice, but this is kinda silly.

src/server.js
```diff
@@ -42,6 +43,11 @@ module.exports = function() {
 		.set("view engine", "html")
 		.set("views", path.join(__dirname, "..", "client"));
 
+	app.get("/plugins/themes/:theme.css", (req, res) => {
+		const theme = req.params.theme;
+		return res.sendFile(themes.fileName(theme));
```
xPaw: You never check if `fileName` returns undefined here.

xPaw:
Just wondering, does default theme work if you specify a plugin theme?

src/plugins/themes.js
```diff
@@ -0,0 +1,79 @@
...
+	}
+	module.lounge.name = moduleName;
+	return module.lounge;
+}
+
+function makeModuleThemeObject(module) {
+	const modulePath = requireg.resolve(module.name);
+	modulePath.substring(0, modulePath.lastIndexOf("/"));
+	const displayName = capitalizeFirstLetter(module.name.replace(/lounge-theme-/, ""));
+	const filename = path.join(modulePath.substring(0, modulePath.lastIndexOf("/")), module.css);
```
xPaw: You're not using `lounge` object as you've documented.

client/index.html
```diff
@@ -274,7 +274,7 @@ <h1 class="title">Settings</h1>
 								<label for="theme-select" class="sr-only">Theme</label>
 								<select id="theme-select" name="theme" class="input">
 									{{#each themes}}
-										<option value="{{filename}}">
+										<option value="{{url}}">
```
xPaw: This breaks backwards compatibility, currently the value is just the theme name like `example`, but this PR changes it to a URL. I assume you changed to url because local themes do not have `plugins/` in it, but maybe a better fix would be to load local themes just the same way as plugins?

src/plugins/themes.js
YaManicKill: I am, check the function called before this, it returns the lounge object.

xPaw: Ah, I see.

src/plugins/themes.js
YaManicKill: Well, it's 6n, as opposed to n, and the number of of themes is going to be low, so it really isn't that much, but I can combine a couple of them.

src/plugins/themes.js
```diff
@@ -0,0 +1,79 @@
...
+
+function getModuleInfo(moduleName) {
+	let module;
+	try {
+		module = requireg(moduleName);
+	} catch (e) {
+		log.warn(`Specified theme ${colors.yellow(moduleName)} is not installed globally`);
+		return;
+	}
+	module.lounge.name = moduleName;
```
xPaw: Probably gotta also check if `module.lounge` even exists, and if not, warn about it.

package.json
YaManicKill: I was so confused with this comment, thought you meant to remove the whole line 😛 I wonder if there is a setting to change npm to save it differently.

src/server.js
YaManicKill: Good point.

client/index.html
YaManicKill: I load the local ones the same way as plugins now, they have a URL as well. Is that what you mean?

src/plugins/themes.js
YaManicKill: Good point. Will want to return early if we don't have it, and it'll filter it out then.

client/index.html
xPaw: Local themes don't have `plugins/` in their url. So basically after upgrading to this PR, everyone will have their theme reset, as `example` will change to `themes/example.css`.

YaManicKill:
> Just wondering, does default theme work if you specify a plugin theme?

What do you mean? It uses the default theme if we get a 404 for another theme. So it'll never be unstyled.

client/index.html
YaManicKill: Ah, gotcha. I guess we can just load the local themes with plugins/ in the URL, that's the nicest way to deal with this, I think. Keep the value as the name and they all have the same URL then.

YaManicKill:
Oh, good question, will check later.

src/server.js
MaxLeiter: Can't set it project-wide (as in, effect other users), but locally you can do `npm config set save-exact true`

YaManicKill: Wrong comment thread? Assuming you mean the version number one.

MaxLeiter: Yeahhhhhhh

src/plugins/themes.js
YaManicKill: I've pulled it down to 2n.

client/index.html
YaManicKill: Fixed this now. All themes use `/theme/themeName.css` and we still use the name of the theme for the value, so this should work now.

src/plugins/themes.js
YaManicKill: Added in a quick check for this.

YaManicKill:
> Just wondering, does default theme work if you specify a plugin theme?

It does now, and I allowed people to specify just the name of the theme rather than the url to the css file, but made sure that we keep backwards compatibility with that.

xPaw:
Good stuff. `lounge-theme-solarized` works with this :)

YaManicKill:
Alrighty, I have updated it to check for themes installed locally as well so people don't need to do install them globally. It uses it globally installed if it exists over a locally installed one.

dgw:
Would it not be better to prefer local over global?

MaxLeiter:
Local over global makes sense to me -- for example, when testing themes.

YaManicKill:
> Would it not be better to prefer local over global?

For what reason? I think that people are more likely to accidentally insatll something locally rather than something globally. And if you do that, you'll not realise it and when you update it globally you'll not get the new version. I think you are much less likely to accidentally install something globally.

But if everyone else wants local, then I'll change it. But I don't think it's right.

YaManicKill:
> for example, when testing themes.

Well, either don't have your theme installed globally, or you can `npm link` it.

defaults/config.js
```diff
@@ -51,9 +51,18 @@ module.exports = {
...
-	// @default  "themes/example.css"
+	// @default  "example"
 	//
-	theme: "themes/example.css",
+	theme: "example",
+
+	//
+	// A list of plugins installed as global npm modules
+	//
+	// Find available themes at https://www.npmjs.com/search?q=keywords%3Alounge-theme
```
astorije: Stupid question but since the npm package is called `thelounge`, do we want to refer to `thelounge-theme` to match the package name, or `lounge-theme` to match the command name?

defaults/config.js
```diff
@@ -51,9 +51,18 @@ module.exports = {
...
 	//
 	// @type     string
-	// @default  "themes/example.css"
+	// @default  "example"
 	//
-	theme: "themes/example.css",
+	theme: "example",
+
+	//
+	// A list of plugins installed as global npm modules
```
astorije: You mentioned `npm link` in the PR comments, do you confirm that having a local theme and `npm link`ing it loads it correctly? It should, but just checking (you'll notice I haven't tested your PR yet...).

What would be really really could would be to have an `lpm` CLI tool to install our packages and themes, and it would simply be a wrapper around `npm` that installs stuff in the `$LOUNGE_HOME/packages` and makes checks around keywords upon installations to make sure you install the correct stuff.
Another benefit of this: in the future, if someone is running multiple instances on their server (for example, one public and one private), we don't want to share packages.

But I'm probably thinking one step too far 😅 Your work is already an amazing step and none of what I said above is made incompatible with your PR. Just something to keep in mind when extending this to generic packages. :)

src/server.js
```diff
@@ -146,16 +156,11 @@ function index(req, res, next) {
 		pkg,
 		Helper.config
 	);
+	if (!data.theme.includes(".css")) { // Backwards compatibility for old way of specifying themes in settings
```
astorije: If memory serves, I think `#includes` is not available before Node.js v... 6? I think that on v5 it was behind a flag, and did not exist on v4 or something like that.

astorije: Mind adding a deprecation notice like the following in this block?

```js
log.warn(`Referring to CSS files in the ${colors.green("theme")} setting of ${colors.green(Helper.CONFIG_PATH)} is ${colors.bold("deprecated")} and will be removed in a future version.`);
```

Unless of course we are not planning to remove support for CSS files, but I think we will eventually, no?

src/plugins/themes.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+fs.readdir("client/themes/", (err, dirThemes) => {
+	if (err) {
+		return;
+	}
+	dirThemes
+		.filter((theme) => theme.endsWith(".css"))
+		.map(makeLocalThemeObject)
+		.forEach((theme) => themes.set(theme.name, theme));
```
astorije: Is this for backwards compatibility only? Mind adding a comment above this block?

defaults/config.js
```diff
@@ -51,9 +51,18 @@ module.exports = {
...
+	//
+	// Find available themes at https://www.npmjs.com/search?q=keywords%3Alounge-theme
+	//
+	// @type	array
+	// @default	[]
+	plugins: [],
```
astorije: In the future, we'll probably want this to be discovered.

astorije:
> 1. Install them as dependencies in lounge, and allow themes to be local packages (currently it only works with global packages)

npm has [`bundledDependencies` in `package.json`](https://docs.npmjs.com/files/package.json#bundleddependencies), could this help us?

> 2. Require people to install the default themes

That's not great because at the bare minimum we'd want to ship at least a default package, I think. Also, that won't scale either as we start to modularize our code in packages that we want to ship by default.

I see Atom has a [`packageDependencies` thingy](https://github.com/atom/atom/blob/master/package.json#L79). I'm guessing it's different from `dependencies` because they can pick which location they want to install this to (cf. my `lpm` comment 😁). But without going that far, since you allow for local packages, I think listing the default themes in `dependencies` should work, no?

src/server.js
xPaw: That's contradicted by http://node.green/#ES2015-built-in-extensions-String-prototype-methods-String-prototype-includes

astorije: That's right, I guess I was looking at the wrong section (it says `.includes` with symbols is not in before Node v6), my bad.
Also, last time we probably hit that when we were still supporting Nove v0.12.
So all good here, thanks!

xPaw:
@YaManicKill If a theme wants to use its own local font files, how can we allow accessing them via Expess?

YaManicKill:
@xPaw We could either allow them to call a function with the static files, and have a large array of static files and statically serve them the same way we are doing it with the css files just now. Or we could allow them to have a "public" folder in their package, which we point to statically using `express.static`. 

Either way, I think, is outside of the scope of this first version.

defaults/config.js
YaManicKill: How do you mean?

src/server.js
YaManicKill: Yeah, sure.

xPaw:
@YaManicKill Sure, just something to keep in mind.

defaults/config.js
YaManicKill: Wow yeah, no, let's ... leave that discussion to a whole other time shall we?

defaults/config.js
YaManicKill: Yeah, can do.

src/plugins/themes.js
YaManicKill: It has been a while since I've looked through the code... but yeah, I think it's just for the existing built in themes.

defaults/config.js
xPaw: This inconsistency is quite annoying. Repo is called `lounge`, that search link is looking for a keyword, not package name.

YaManicKill:
> npm has bundledDependencies in package.json, could this help us?

Doesn't make any difference, and we really shouldn't be bundling things in like that. We end up with the exact same thing afterwards, it's just our bundle on npm is bigger.

> That's not great because at the bare minimum we'd want to ship at least a default package, I think. Also, that won't scale either as we start to modularize our code in packages that we want to ship by default.

Yeah, agreed.

> But without going that far, since you allow for local packages, I think listing the default themes in dependencies should work, no?

Yeah, since my original comment, this has changed so we do allow local packages, so that would make sense. But, again, outside of the scope of this PR, IMO.

astorije:
So, I really like where this PR is going, but there is one thing that bothers me a lot: that we are looking at global npm packages.
This is a problem for several reasons:

- It forces us (well, admins) to keep a list of `plugins` in the config file, which shouldn't be necessary.
- It means that all instances of The Lounge running on a given server will have to offer the same plugins and themes.
- It forces admins to install all third-party plugins in the global namespace, which is not the most secure condition we can offer.

I think we should really be reading packages and themes from `$LOUNGE_HOME/packages`. That solves the issues above and a few others.
The only downside I can think of is that admins would need to install things slightly differently than currently, with `npm install --prefix $LOUNGE_HOME/packages <package>` (which isn't that much worse than `sudo npm install --global <package>`). There are very elegant ways to address this however and make it smoother, but totally not the scope for this PR.

What do you think of this proposal? I can obviously help getting somewhere if you don't want to put yet another effort after the fact :) If you want me to try that, I'd do it in extra commits to not overwrite your stuff.

The reason why I'm bringing the package location now is because it's going to be the base for all our themes/packages. I think it is important to solve it now, because if we do it later, transition for users will be painful.

YaManicKill:
> It forces us (well, admins) to keep a list of plugins in the config file, which shouldn't be necessary.

Afaics, this is a pretty standard way of doing things. Atom does it behind the scenes, but still does it. Everything else I use that has plugins does the same thing.

I see what you are saying, but I'm not sure how I feel about it. Let me try it out and see how it feels to me.

astorije:
Reminder to change terminology from "plugins" to "packages" as discussed on the channel :)
This is so exciting, a first, actual, real, tangible step towards packagifying The Lounge 👏 👏

YaManicKill:
I didn't get any time last night to do it, but I will make time tonight to get this done.

Added tasks to the PR.

defaults/config.js
YaManicKill: Yeah...I agree. We do have issues in having the 2 names.

src/plugins/themes.js
YaManicKill: I've renamed some stuff here, hopefully better?

src/server.js
YaManicKill: It's actually not inside this block as we go into this block if we are using the new way of doing it, so it's in an else block.

YaManicKill:
Think I've done everything now. I'll fix up thelounge.github.io#69 as well

src/helper.js
```diff
@@ -96,6 +98,14 @@ function getStoragePath() {
 	return path.join(this.HOME, "storage");
 }
 
+function getPackagesPath() {
+	return path.join(this.HOME, "packages", "node_modules");
```
astorije: Meeeh, kinda sucks that we can't install in `packages/` itself and need a `packages/node_modules/` but oh well, looks like there is no way to do otherwise...

src/plugins/themes.js
```diff
@@ -0,0 +1,84 @@
...
+	if (themes.has(module)) {
+		return themes.get(module).filename;
+	}
+}
+
+function makeLocalThemeObject(css) {
+	const themeName = css.slice(0, -4);
+	return {
+		displayName: themeName.charAt(0).toUpperCase() + themeName.slice(1),
+		filename: path.join(__dirname, "..", "client", "themes", `${themeName}.css}`),
```
astorije: Is that trailing `}` a typo?

src/plugins/themes.js
```diff
@@ -0,0 +1,84 @@
...
+	packages
+		.map(makePackageThemeObject)
+		.forEach((theme) => {
+			if (theme) {
+				themes.set(theme.name, theme);
+			}
+		});
+});
+
+function get() {
```
astorije: Nitpick but maybe better with `list` or `listAllThemes` or something like that? Don't worry, I'm not bikeshedding, I need to test this PR anyway when I have a minute :)

astorije:
This looks pretty damn good!! I believe I might have noted a typo (see comment), and I need to test this PR, but overall looks really good and it's very exciting!
👏 🎉 

How easy / difficult would it be to make this _not_ require to restart the server when adding/upgrading/removing a theme? (Note that whatever the answer, I'm not saying it should be in this PR 😉)
Restarting the server when you don't use ZNC is rather painful, I would need to find a really good theme to justify the cost of restarting the server :D

src/plugins/themes.js
YaManicKill: Why? Is it because it's a map internally? 

1. The outer code doesn't care how it is stored internally, and shouldn't. 
2. If you want us to say it's an Array (it's not a list...) then let's do that in JS doc, not in the name of the function.
3. This is a getter function, get is a pretty standard name
4. If you want "all", then that's fine. I'd be happy with "getAll"

tl;dr, I'd be happy with "get" or "getAll", or I'd be happy to add jsdoc to make it clear it's an array it returns. Other than that, I'm going to push back.

src/plugins/themes.js
YaManicKill: Hmmmm, yeah, it is. I didn't test local themes at the end, will sort this out.

YaManicKill:
> How easy / difficult would it be to make this not require to restart the server when adding/upgrading/removing a theme?

Not difficult at all, I don't think. We already do the same for user config. But yeah, just because this PR has been going (and is useful in itself) let's leave that to another PR. It's a good addition, though.

src/helper.js
YaManicKill: Yeah, it shouldn't realistically matter because users won't be going directly into this folder, so it doesn't matter how it looks. I get your frustration, but it's really not a massive issue.

package.json
```diff
@@ -52,6 +52,7 @@
     "ldapjs": "1.0.1",
     "lodash": "4.17.4",
     "moment": "2.18.1",
+    "package-json": "4.0.1",
```
astorije: @xPaw, could you remove your commit out of this PR? This PR is essentially ready, and I'm expecting it to be merged during the weekend.

src/plugins/themes.js
astorije: > Why? Is it because it's a map internally?

No no sorry, not at all. It's just that only `get()` seems not very descriptive enough (even for a getter), but I don't mind much anyway, that won't prevent me to approve once I can test. It's just, if there is a more descriptive name to give it, good, otherwise oh well.

src/plugins/themes.js
```diff
@@ -0,0 +1,84 @@
...
+	return Array.from(themes.values());
+}
+
+function fileName(module) {
```
astorije: I see a mix of `filename` and `fileName` around here, could we stick with `filename` only?

src/plugins/themes.js
YaManicKill: I mean, the module is "themes" and the function is "get". Seems pretty descriptive to me. As I said, happy to change to getAll if you would prefer that.

src/plugins/themes.js
YaManicKill: Will. I am never sure whether filename is 1 or 2 words therefore what it should be in camel case.

src/plugins/themes.js
YaManicKill: Weirdly, these still work fine. I assume `path.join` or `express.sendFile` filter it out. But I'll remove it, obviously.

src/plugins/themes.js
YaManicKill: I'm renaming this to getAll to make it clear that it gets ... all the themes.

src/plugins/themes.js
YaManicKill: Replacing all to filename.

YaManicKill:
Should be up to date with requested changes, @astorije

astorije:
Not sure if @xPaw's `lounge install` command does that already, but first time your try to install a package with the current way of doing it, `npm` complains the directory doesn't exist:

```sh
$ npm install --prefix $LOUNGE_HOME/packages lounge-theme-polarized
Error: EACCES: permission denied, mkdir '/packages'
    at Error (native)
```

This is just a reminder for `lounge install`, not for this very PR :)

src/plugins/themes.js
```diff
@@ -0,0 +1,84 @@
...
+		.map(makePackageThemeObject)
+		.forEach((theme) => {
+			if (theme) {
+				themes.set(theme.name, theme);
+			}
+		});
+});
+
+function getAll() {
+	return Array.from(themes.values());
```
astorije: Could you do something like this:

```diff
───────────────────────────────
modified: src/plugins/themes.js
───────────────────────────────
@ themes.js:7 @ const fs = require("fs");
 const Helper = require("../helper");
 const colors = require("colors/safe");
 const path = require("path");
+const _ = require("lodash");
 const themes = new Map();

 module.exports = {
@ themes.js:39 @ fs.readdir(Helper.getPackagesPath(), (err, packages) => {
 });

 function getAll() {
-   return Array.from(themes.values());
+   return _.sortBy(Array.from(themes.values()), "displayName");
 }
```

This sorts the themes in the list, for example:

screen shot 2017-09-17 at 01 27 27

YaManicKill:
Ok yeah, Google Fonts works fine. I'm using a custom css import for roboto with solarised. That's all you need. Why not just tell them to use Google Fonts? That's the only decent cross platform way to deal with it. The standard exists for a reason.

xPaw:
@YaManicKill He meant shipping font files with the theme, which we discussed before (e.g. theme exports files it wants to server via express).

YaManicKill:
I know what he means. I'm just not sure why... Those don't always work cross platform and web Fonts work much nicer. Anyway, discussion outside of the pr.

xPaw:
@YaManicKill You do know you can serve web fonts locally, right? That's what we are suggesting. Like we do with Lato: https://github.com/thelounge/lounge/tree/master/client/css/fonts

YaManicKill:
What's the purpose of that rather than hosted ones?

xPaw:
Not giving traffic to Google. It's silly to argue this in FOSS software, it should be self contained.

astorije:
Exactly what @xPaw said. There is no way we should recommend people to use Google Fonts :)

That being said it's totally okay to discuss this in a dedicated ticket as mentioned, this PR is good on its own!

YaManicKill:
I forget about what some people's reasons for Foss are. Fine, let's deal with the crazy. Anyway, will fix up the lint error this evening.

PR #1271, opened 2017-06-25
Localization support

MaxLeiter:
Major WIP, maybe more of a POC.

Added a few translations to act as examples, not going to add the rest until I'm / we're sure this is how to go about it.

Need to add server-side option for default language and language selector on the client. Also need to setup translations server-side for use in `index.html` and all the plugins/inputs (like what commands return). 

Todo (missing stuff):
- [x] Client settings option to change language
- [x] Support switching language with page reload
- [x] Complete en-US translations
- [x] Server-side default language option
- [ ] Tests
- [ ] Finalize JSON layout


src/server.js
```diff
@@ -35,7 +35,8 @@ module.exports = function() {
 		.engine("html", expressHandlebars({
 			extname: ".html",
 			helpers: {
-				tojson: (c) => JSON.stringify(c)
+				tojson: (c) => JSON.stringify(c),
+				translate: (c) => i18n.t(c)
```
MaxLeiter: doesn't currently do anything

xPaw: Client JS needs to perform most of the translations. Only initial Loading/enable js strings need to be translated.

MaxLeiter: Agreed, but need to somehow translate index.html strings

xPaw: I'm guessing leave most of them untranslated tokens, add a class like `translate`, and iterate them in client JS on load?

client/js/render.js
```diff
@@ -72,6 +72,10 @@ function buildChatMessage(data) {
 		text.html(templates.actions[type](data.msg));
 	}
 
+	if (template === "msg" && data.msg.type !== "motd" && (data.msg.from === "undefined" || data.msg.from === "" || data.msg.type === "error")) {
```
xPaw: What messages have `from` undefined?

client/translations/en-US.json
```diff
@@ -0,0 +1,85 @@
...
+  "network_created": "Network created, connecting to {{host}}:{{port}}...",
+  "enabled_capabilities": "Enabled capabilities: {{caps}}",
+  "connected_network": "Connected to the network",
+  "disconnected_network_will_not_reconnect": "Disconnected from the network, and will not reconnect. Use /connect to reconnect again.",
+  "disconnected_retry": "Disconnected from the network. Reconnecting in {{seconds}} seconds… (Attempt {{attempt}} of {{max_retries}})",
+  "ping_timeout": "Ping timeout, disconnecting...",
+  "now_known_as": "You're now known as {{new_nick}}"
+}
+
+}
```
xPaw: identation has gone wrong somewhere

client/translations/en-US.json
```diff
@@ -0,0 +1,85 @@
...
+  "has_kicked": "has kicked",
+  "sets_mode": "sets mode",
+  "now_known_as": "is now known as",
+  "has_left": "has left the channel",
+  "has_quit": "has quit",
+  "topic": {
+    "has_changed": "has changed the topic to",
+    "topic_is": "The topic is:"
+  },
+  "topic_set_by": {
```
xPaw: This probably needs to be a single string...

client/js/render.js
MaxLeiter: None 👍

MaxLeiter:
This basically works now. Just need to:

1) translate index.html before it renders
2) translate server `log()` calls
3) default language server setting (need to somehow access defaults in `lounge.js`)
4) write tests 😢

williamboman:
> translate server log() calls

I don't think server logs should be localized, I don't think I've ever seen it done elsewhere. It also opens up for misinterpretations due to server logs being more low-level and detailed by nature. Server logs should be English only imo.

MiniDigger:
translated messages are bad. when you google an error, you would find the most results in english. 
also, imagine someone opening a issue here, only able to provide a log in chinese or smth.

MaxLeiter:
Makes my life easier 👍

src/server.js
MaxLeiter: Is now used

client/locales/en-US.json
```diff
@@ -0,0 +1,215 @@
...
+		"network_created": "Network created, connecting to {{host}}:{{port}}...",
+		"enabled_capabilities": "Enabled capabilities: {{caps}}",
+		"connected_network": "Connected to the network",
+		"disconnected_network_will_not_reconnect": "Disconnected from the network, and will not reconnect. Use /connect to reconnect again.",
+		"disconnected_retry": "Disconnected from the network. Reconnecting in {{seconds}} seconds… (Attempt {{attempt}} of {{max_retries}})",
+		"ping_timeout": "Ping timeout, disconnecting...",
+		"now_known_as": "You're now known as {{new_nick}}"
+	}
+
+}
```
astorije: Could you add an EOF line please?

client/locales/fr.json
```diff
@@ -0,0 +1,10 @@
+{
+  "close_channel": "Fermer le canal",
+  "toggle_chan_list": "Baguette",
+  "toggle_user_list": "Du fromage",
+  "open_context_menu": "Oh la la",
+  "show_older_messages": "Je m'appelle Max",
+  "search_among_user_list": "J'adore The Lounge",
+  "disconnect_from": "hopefully this works : {{server}}?"
+
+}
```
astorije: WAT :joy:

client/views/actions/invite.tpl
```diff
@@ -1,9 +1,9 @@
...
+{{translate 'client.invited.invited'}}
 {{#if invitedYou}}
-	you
+	{{translate 'client.invited.you'}}
 {{else}}
 	{{> ../user_name nick=invited}}
 {{/if}}
-to
+{{translate 'client.invited.to'}}
 {{{parse channel}}}
```
astorije: I don't think this is sustainable. Assuming sentence order being `[nick] [invited] [nick] [to] [channel]` is likely not going to be true in a bunch of languages. For example, in French, it would work when reporting about someone else: `[bob] [a invité] [alice] [sur] [#thelounge]`, but when user gets invited, that would be: `[bob] [vous] [a invité sur] [#thelounge]`.

I barely remember anything from my German years, but I believe the structure would be something like: `[bob] [has] [you] [to] [#thelounge] [invited]` or something like that.

client/views/actions/nick.tpl
```diff
@@ -1,3 +1,3 @@
 {{> ../user_name nick=from}}
-is now known as
+{{translate 'client.now_known_as'}}
 {{> ../user_name nick=new_nick}}
```
astorije: Previous comment applies here as well, and most templates I believe. Could this be made so that the strings in the JSON files contain the other elements, like `{{from}} is now known as {{new_nick}}`?

astorije:
I haven't reviewed but just took a random look through the PR.
This is very exciting, I can't wait! Pretty sure language files are going to flourish after that!!

client/views/actions/nick.tpl
MaxLeiter: Yeah, the Handlebars template supports passing in options, just need to figure out how :smile:

MaxLeiter: Actually, not sure how to solve this exactly, because it requires passing a template into a template...

client/locales/fr.json
MaxLeiter: Needed to test and don't know French

client/views/actions/invite.tpl
MaxLeiter: Yeah, not too sure how to solve this -- need to do some reading up on Handlebars. http://i18next.github.io/i18next/pages/doc_templates.html has an "extended handlebars helper" but for the life of me I can't figure out how it works

client/locales/en-US.json
MaxLeiter: Done

MaxLeiter:
Ref https://github.com/wycats/handlebars.js/issues/1360 and https://github.com/wycats/handlebars.js/issues/1361 and

astorije:
Because of https://github.com/wycats/handlebars.js/issues/1360#issuecomment-313910281 (there won't be new features to Handlebars at least for now), how do we want to handle this? :(

MaxLeiter:
There may be a solution we haven't thought of? @xPaw any ideas?

MaxLeiter:
Should i rebase this or not worth the effort until handlebars updates? (which may be never...)

Comes down to if we want half-baked or not

astorije:
If you can rebase, that would be awesome. I am still very much looking forward to get involved with this and would absolutely love for us to support multiple languages soon :)

MaxLeiter:
Rebased to master and updated to translate new strings.

client/index.html
```diff
@@ -83,185 +84,186 @@ <h1 class="title">The Lounge is loading…</h1>
...
 								</label>
 							</div>
 							<div class="col-xs-12">
 								<label>
-								Password
+								<span data-translate="true">index.signin.password</span>
 								<input class="input" type="password" name="password">
 								</label>
 							</div>
 							<div class="col-xs-12 error" style="display: none;">
```
PolarizedIons: Are you missing a `data-translate="true"` here?

MaxLeiter: Nice catch

client/index.html
```diff
@@ -466,23 +479,32 @@ <h1 class="title">Help</h1>
...
 							</div>
 							<div class="description">
-								<p>Switch to the previous/next window in the channel list</p>
+								<p data-translate="true">index.help.switch_next_prev_window</p>
 							</div>
 						</div>
 
 						<div class="help-item">
 							<div class="subject">
+<<<<<<< HEAD
```
PolarizedIons: Looks like you forgot to deal with a merge conflict here

client/index.html
```diff
@@ -561,12 +579,24 @@ <h1 class="title">Help</h1>
...
 							</div>
 							<div class="description">
-								<p>Load the banlist for the current channel.</p>
+								<p data-translate="true">index.commands.banlist</p>
 							</div>
 						</div>
 
 						<div class="help-item">
 							<div class="subject">
+<<<<<<< HEAD
```
PolarizedIons: and here

client/index.html
```diff
@@ -763,10 +778,14 @@ <h1 class="title">Help</h1>
 								<code>/part [channel]</code>
 							</div>
 							<div class="description">
+<<<<<<< HEAD
```
PolarizedIons: and here ;)

client/index.html
```diff
@@ -582,10 +612,8 @@ <h1 class="title">Help</h1>
...
 							</div>
 							<div class="description">
-								<p>
-									Connect to a new IRC network. If <code>port</code> starts with
-									a <code>+</code> sign, the connection will be made secure
-									using TLS.
+								<p data-translate="true">
+									index.commands.connect
 								</p>
 								<p>Alias: <code>/server</code></p>
```
PolarizedIons: These `Alias:`s aren't being translated

client/views/chat.tpl
```diff
@@ -1,32 +1,32 @@
...
+			<button class="show-more-button" data-id="{{id}}">{{translate 'client.show_older_messages'}}</button>
 		</div>
 		<div class="messages"></div>
 	</div>
 	{{#equal type "channel"}}
 	<aside class="sidebar">
 		<div class="users">
 			<div class="count">
-				<input type="search" class="search" aria-label="Search among the user list" tabindex="-1">
+				<input type="search" class="search" aria-label="{{translate 'client.search_among_user_list'}}" tabindex="-1">
```
PolarizedIons: I'm not sure where yet, but this is getting a `placeholder="{{amount}} users"`, the 'users' isn't translated

client/index.html
```diff
@@ -63,18 +64,18 @@ <h1 class="title">The Lounge is loading…</h1>
...
-								--><span id="set-nick-tooltip" class="tooltipped tooltipped-e" aria-label="Change nick"><button id="set-nick" type="button" aria-label="Change nick"></button></span><!--
-								--><span id="cancel-nick-tooltip" class="tooltipped tooltipped-e" aria-label="Cancel"><button id="cancel-nick" type="button" aria-label="Cancel"></button></span><!--
-								--><span id="save-nick-tooltip" class="tooltipped tooltipped-e" aria-label="Save"><button id="submit-nick" type="button" aria-label="Save"></button></span>
+								--><span id="set-nick-tooltip" class="tooltipped tooltipped-e" aria-label="index.change_nick"><button id="set-nick" type="button" aria-label="index.change_nick"></button></span><!--
+								--><span id="cancel-nick-tooltip" class="tooltipped tooltipped-e" aria-label="index.cancel"><button id="cancel-nick" type="button" aria-label="index.cancel"></button></span><!--
+								--><span id="save-nick-tooltip" class="tooltipped tooltipped-e" aria-label="index.save"><button id="submit-nick" type="button" aria-label="index.save"></button></span>
 							</span>
 							<textarea id="input" class="mousetrap"></textarea>
```
PolarizedIons: This is given a `placeholder="Write to {{channel}}"` which isn't translated

client/index.html
```diff
@@ -31,30 +31,31 @@
...
-			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Help"><button class="icon help" data-target="#help" aria-label="Help"></button></span>
-			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Sign out"><button class="icon sign-out" id="sign-out" aria-label="Sign out"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="index.sign_in"><button class="icon sign-in" data-target="#sign-in"
+			aria-label="index.sign_in"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="index.connect_to_network"><button class="icon connect" data-target="#connect" aria-label="index.connect_to_network"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="index.client_settings"><button class="icon settings" data-target="#settings" aria-label="index.client_settings"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="index.html"><button class="icon help" data-target="#help" aria-label="index.help_label"></button></span>
```
PolarizedIons: This `aria-label="index.html"` should be `aria-label="index.help_label"`

client/index.html
```diff
@@ -83,185 +84,186 @@ <h1 class="title">The Lounge is loading…</h1>
...
 							</div>
 							<div class="col-sm-12">
 								<label class="opt">
 									<input type="checkbox" name="coloredNicks">
-									Enable colored nicknames
+									<span data-translate="true">index.settings.enabled_color_nicks</span>
 								</label>
 								<label class="opt">
 									<input type="checkbox" name="autocomplete">
 									Enable autocomplete
```
PolarizedIons: not translated

client/index.html
```diff
@@ -83,185 +84,186 @@ <h1 class="title">The Lounge is loading…</h1>
...
 								</label>
 							</div>
 							<div class="col-sm-12">
-								<h2>Theme</h2>
+								<h2 data-translate="true">Theme</h2>
```
PolarizedIons: `Theme` -> `index.settings.theme`

client/views/chat.tpl
```diff
@@ -1,32 +1,32 @@
...
 			</div>
 			<div class="names names-filtered"></div>
 			<div class="names names-original"></div>
 		</div>
 	</aside>
 	{{/equal}}
 </div>
-{{/each}}
+{{/each}}
```
PolarizedIons: EOF

PolarizedIons: newline

client/views/msg.tpl
```diff
@@ -14,4 +14,4 @@
 			<div class="preview" data-url="{{link}}"></div>
 		{{/each}}
 	</span>
-</div>
+</div>
```
PolarizedIons: newline

client/views/msg_preview.tpl
```diff
@@ -16,4 +16,4 @@
 		</a>
 	{{/equal}}
 </div>
-{{/preview}}
+{{/preview}}
```
PolarizedIons: newline

client/index.html
```diff
@@ -83,185 +84,186 @@ <h1 class="title">The Lounge is loading…</h1>
...
 							</div>
 							<div class="col-sm-12">
-								<h2>Messages</h2>
+								<h2 data-translate="true">index.settings.messages</h2>
 							</div>
 							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="motd">
-									Show <abbr title="Message Of The Day">MOTD</abbr>
+									<span data-translate="true">index.settings.show_motd</span> <abbr title="Message Of The Day">MOTD</abbr>
```
PolarizedIons: Currently `index.settings.show_motd` is just "Show", and then the `abbr` is put after that. This works fine with English, but it would be better if the `abbr` is inserted into the translation string, rather than tacked on at the end

client/views/chat.tpl
```diff
@@ -1,32 +1,32 @@
...
 	<div class="chat">
 		<div class="show-more {{#equal messages.length 100}}show{{/equal}}">
-			<button class="show-more-button" data-id="{{id}}">Show older messages</button>
+			<button class="show-more-button" data-id="{{id}}">{{translate 'client.show_older_messages'}}</button>
 		</div>
 		<div class="messages"></div>
 	</div>
 	{{#equal type "channel"}}
 	<aside class="sidebar">
 		<div class="users">
```
PolarizedIons: I'm not sure about if the 'mode headings' that are put here should be translated, but the `Search results` text when searching probably should

client/views/actions/invite.tpl
PolarizedIons: I still very much don't like this, but I have not found a nice solution either.

The only thing I could come up with is having a `client.invite.userFirst = true` in the translation file, and using that here to decide if the user/'you' or the channel goes first, then using your approach and having `client.invited.prefix`, `.middle`, and `.suffix`. I don't fully like this either though, since it puts logic inside the translation file

client/js/lounge.js
```diff
@@ -417,7 +419,7 @@ $(function() {
 		if (chan.hasClass("lobby")) {
 			cmd = "/quit";
 			var server = chan.find(".name").html();
-			if (!confirm("Disconnect from " + server + "?")) {
+			if (!confirm(t.translate("disconnect_from", {server: server}))) {
```
PolarizedIons: should this be `client.disconnect_from` ?

client/views/actions/part.tpl
```diff
@@ -1,6 +1,6 @@
 {{> ../user_name nick=from}}
 <i class="hostmask">({{hostmask}})</i>
-has left the channel
+{{translate 'client.part'}}
```
PolarizedIons: This has no translation key

client/views/actions/topic.tpl
```diff
@@ -1,8 +1,8 @@
 {{#if from}}
 	{{> ../user_name nick=from}}
-	has changed the topic to:
+	{{translate 'client.topic.has_changed'}}
 {{else}}
-	The topic is:
+	{{translate 'client.topic.topic_is'}}
 {{/if}}
 
 <span class="new-topic">{{{parse text}}}</span>
```
PolarizedIons: The same issue with word ordering here as above: `{{user}} has changed the topic to {{topic}}` may not work with all languages

client/views/actions/topic_set_by.tpl
```diff
@@ -1,3 +1,4 @@
-Topic set by
+{{translate 'client.topic_set_by.before'}}
 {{> ../user_name}}
-on {{localetime when}}
+{{translate 'client.topic_set_by.between'}} {{localetime when}}
+{{translate 'client.topic_set_by.after'}}
```
PolarizedIons: None of these have a translation key & missing newline

client/views/chat.tpl
```diff
@@ -1,32 +1,32 @@
...
+		<button class="lt" aria-label="{{translate 'client.toggle_chan_list'}}"></button>
 		{{#equal type "channel"}}
-			<span class="rt-tooltip tooltipped tooltipped-w" aria-label="Toggle user list">
-				<button class="rt" aria-label="Toggle user list"></button>
+			<span class="rt-tooltip tooltipped tooltipped-w" aria-label="{{translate 'client.toggle_user_list'}}">
+				<button class="rt" aria-label="{{translate 'client.toggle_user_list'}}"></button>
 			</span>
 		{{/equal}}
-		<button class="menu" aria-label="Open the context menu"></button>
+		<button class="menu" aria-label="{{translate 'open_context_menu'}}"></button>
```
PolarizedIons: `open_context_menu` -> `client.open_context_menu`

src/plugins/irc-events/welcome.js
```diff
@@ -9,7 +9,8 @@ module.exports = function(irc, network) {
 
 		var lobby = network.channels[0];
 		var msg = new Msg({
-			text: "You're now known as " + data.nick
+			text: "now_known_as",
```
PolarizedIons: `now_known_as` -> `server.now_known_as`

client/locales/en-US.json
```diff
@@ -0,0 +1,226 @@
...
+      "exit_warning":"Are you sure you want to navigate away from this page?"
+   },
+   "server":{
+      "page_closed":"Page closed",
+      "password":{
+         "change_enter_new":"Please enter a new password",
+         "change_fail":"Failed to update your password",
+         "change_fail_entered_not_match":"Both new password fields must match",
+         "change_fail_entered_not_match_account":"The current password field does not match your account password",
+         "change_success":"Successfully updated your password, all your other sessions were logged out"
```
PolarizedIons: I believe since #1199, sessions are no longer logged out, so this should probably be updated

client/locales/en-US.json
```diff
@@ -0,0 +1,226 @@
...
+         "away":"Mark yourself as away with an optional message",
+         "back":"Remove your away status (set with {{away}}).",
+         "ban":"Ban ({{banMode}}) a user from the current channel. This can be a nickname or a hostmask.",
+         "banlist":"Load the banlist for the current channel.",
+         "connect":"Connect to a new IRC network. If {{port}} starts with a {{plus}} sign, the connection will be made secure using TLS.",
+         "ctcp":"Send a {{ctcp}} request.",
+         "deop":"Remove op ({{deopMode}}) from one or several user in the current channel.",
+         "devoice":"Remove voice ({{devoiceMode}}) from one or several users in the current channel.",
+         "disconnect":"Disconnect from the current network with an optionally-provided message.",
+         "invite":"Invite a user to the specified channel. If {{channel}} is ommitted, user will be invited to the current channel.",
```
PolarizedIons: `ommitted` -> `omitted`

PolarizedIons:
In the console I see
```
i18next::backendConnector: loaded namespace translation for language en-US ...
i18next::backendConnector: loaded namespace translation for language fr ...
```
Does it load ALL languages? Shouldn't it just load the current or default language?

---

It also tries to load `/locales/en.json` and 404s. I remember you saying it will load the `language.json` then override that with `language-COUNTRY.json`, to allow for regional translations. I'm indifferent about this 404, but it would be nice to not have

---

The condense messages aren't translated at all

---

So much for a "quick check" on the pr :stuck_out_tongue: So happy to see this PR mostly working, and looking forward to it eventually being merged!

client/views/chat.tpl
MaxLeiter: That's in CSS; @thelounge/maintainers: should I move it out?

(`content: "Search Results`")

client/js/lounge.js
MaxLeiter: Yep

client/views/actions/part.tpl
MaxLeiter: Should be `client.has_left`

client/views/actions/topic.tpl
MaxLeiter: here's where {{before}} {{end}} could work.
`{{before}} {{translate 'client.topic.topic_is'}} {{end}}`

In english

`before` = `The topic is:`
`end` = ` `

client/views/actions/topic_set_by.tpl
MaxLeiter: This was me deciding if we want to do `{{before}}`, `{{between}}`, `{{after}}`

client/locales/en-US.json
MaxLeiter: I don't see any strings being changed in that PR, so I think this is fine

client/views/chat.tpl
MaxLeiter: This and the comment below are done in javascript. nice catch

client/locales/en-US.json
PolarizedIons: fair enough

client/js/render.js
```diff
@@ -101,6 +102,10 @@ function buildChatMessage(msg) {
 		renderPreview(preview, renderedMessage);
 	});
 
+	if (template === "msg" && type !== "motd" && (msg.from === "" || type === "error")) {
```
xPaw: Can you just send `translate: true` from the server where it sends tokens instead?

MaxLeiter: Done

astorije:
@MaxLeiter, I know this is going to be a huge pain, but would you mind rebasing this? I'd like to start reviewing it.

MaxLeiter:
@astorije sure, but still need to decide how to handle different word ordering/insertion.

Should I go with 
```
{{before}} {{middle}} {{end}}
```
?

astorije:
> @astorije sure, but still need to decide how to handle different word ordering/insertion.

Yes, that's precisely why I want to start reviewing it :)

I can't answer that question before I take a serious look at it. Just rebase and ignore that issue for now.

astorije:
@MaxLeiter I really think we should avoid that as much as we can, which is why I want to spend some time on this PR and play with it, see what can be done. There are internationalization solutions for handlebars out there, I just need to experiment with them.

MaxLeiter:
Closing until Vue is more fleshed out and implemented. Will probably need to be totally rewritten at that point

PR #1273, opened 2017-06-26
Do not display preview if there is nothing to preview

xPaw:


src/plugins/irc-events/link.js
```diff
@@ -74,11 +74,13 @@ function parse(msg, url, res, client) {
 		// Verify that thumbnail pic exists and is under allowed size
 		if (toggle.thumb.length) {
 			fetch(escapeHeader(toggle.thumb), (resThumb) => {
-				if (!(/^image\/.+/.test(resThumb.type)) || resThumb.size > (Helper.config.prefetchMaxImageSize * 1024)) {
+				if (resThumb === null
```
xPaw: Fixes crash on master, @astorije

PR #1275, opened 2017-06-26
Move link prefetcher button to toggle template and warn on the client when image is too big

astorije:
Before that PR, toggle button would appear instantly even if the link cannot be prefetched. We need to wait that a response comes back to know if it actually can be displayed or not.

Probably worth reviewing with `?w=1` as most of this PR changes are just indentation.


xPaw:
Closed in favour of #1276

PR #1280, opened 2017-06-27
Do not hardcode vendor bundles in webpack

xPaw:


webpack.config.js
```diff
@@ -71,7 +60,13 @@ const config = {
 		json3: "JSON", // socket.io uses json3.js, but we do not target any browsers that need it
 	},
 	plugins: [
-		new webpack.optimize.CommonsChunkPlugin("js/bundle.vendor.js")
+		// socket.io uses debug, we don't need it
+		new webpack.NormalModuleReplacementPlugin(/debug/, path.resolve(__dirname, "scripts/noop.js")),
+		// automatically split all vendor dependancies into a separate bundle
+		new webpack.optimize.CommonsChunkPlugin({
+			name: "js/bundle.vendor.js",
+			minChunks: (module) => module.context && module.context.indexOf("node_modules") !== -1
```
astorije: How does this know what packages to build into the vendor bundle specifically?

xPaw: `node_modules` being in the path.

astorije: Right, that tells Webpack where are our packages, but is it going to load everything that's under `node_modules` (like lodash, that we don't use in the client)? Or is it the other way around, it discovers everything that's `require()`d, then puts them in the right file based on this line?

xPaw: It's based on requires. Basically it generates identical bundles to what master has now.

PR #1281, opened 2017-06-27
Add connect URL query to skip connect form

MaxLeiter:
If you add ?connect=true to the URL on a public instance the user will be automatically connected.


xPaw:
I don't think we should add this, seems open to abuse (hidden iframe that auto connects?)

Issue #791, opened 2016-12-12
Add "home" tab


YaManicKill:
As discussed in #503, we could make a "home" tab which will initially be blank, so that when we close a channel that we are currently on, we would go back to the "home" tab so we don't mark any other channel as read. This is how slack does it, fwiw.

This also has a future advantage of having some other possibilities for viewing things (highlights, stats, etc) further down the line, but initially it should probably just be blank.

MaxLeiter:
Similar to #650

YaManicKill:
Yeah, it's not exactly the same, but that could definitely tie in. I assume
that they are influenced by slack having one where its just basically a pm
to yourself, but the idea is that it wouldn't go over irc, I guess?

teamcoltra:
This could be a server tab, which most IRC clients already have.

YaManicKill:
@teamcoltra We already have a server tab, which is slightly different from this. The server tab is for server notices, motd, etc.

PR #1283, opened 2017-06-28
Add anchor tag to URL to signify open page for reloading

MaxLeiter:
example: `https://mylounge.com/#help` opens the help page
`https://mylounge.com/#chan-1` opens the network (because the first connected channel, not the network, is assigned id `0`)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -338,9 +338,9 @@ $(function() {
 
 		if (history && history.pushState) {
 			if (data && data.replaceHistory && history.replaceState) {
-				history.replaceState(state, null, null);
+				history.replaceState(state, null, target);
 			} else {
-				history.pushState(state, null, null);
+				history.pushState(state, null, target);
 			}
 		}
 	});
@@ -594,7 +594,8 @@ $(function() {
 			}
 		});
 	});
-	if ($("body").hasClass("public")) {
+
+	if ($("body").hasClass("public") && window.location.hash === "#connect") {
 		$("#connect").one("show", function() {
 			var params = URI(document.location.search);
 			params = params.search(true);
--- a/client/js/socket-events/init.js
+++ b/client/js/socket-events/init.js
@@ -32,7 +32,10 @@ socket.on("init", function(data) {
 	const target = sidebar.find("[data-id='" + id + "']").trigger("click", {
 		replaceHistory: true
 	});
-	if (target.length === 0) {
+	const dataTarget = document.querySelector("[data-target='" + window.location.hash + "']");
+	if (window.location.hash && dataTarget) {
+		dataTarget.click();
+	} else if (target.length === 0) {
 		const first = sidebar.find(".chan")
 			.eq(0)
 			.trigger("click");


xPaw:
What's the use case? To get `#chan-1` to work, you'd need to have the channel already there on page load (e.g. logged in). We already re-open active channel on page reloads for logged in users.

MaxLeiter:
Help, settings, and connect windows

xPaw:
But does that solve any real use case?

MaxLeiter:
If I'm on the help page and reload, I expect it to stay on the help page?

astorije:
It will be very useful if we have settings that force a page refresh when changed (like changing the language locale), i.e. when it's much cheaper to just reload than to recompute/repaint the whole app.
It will also be very useful, if not mandatory, when we get to #791. When that happens, loading the app should go to the home page, unless it was a refresh, and it won't make sense to keep track of the last opened tab (this behavior really annoys me at the moment).

In the meantime... maybe it's too soon? :/ I can definitely understand @MaxLeiter's points though.

xPaw:
This will probably actually be needed for push notifications (click on notif -> open channel where it came from).

astorije:
@xPaw, mind taking over this PR as @MaxLeiter will not be able to help for some time?

christer88:
Would this let you specify a channel someone join (for the "login" page in public mode)? e.g. demo.thelounge.org/#testchannel takes the user to #testchannel on the specified network in config file.

MaxLeiter:
@christer88 different feature, but on public mode you can do that like this:

https://demo.thelounge.chat/?join=%23testchannel&nick=example

MaxLeiter:
Bug: if you reload the page on #help, #connect, or #settings the sidebar disappears.

astorije:
@MaxLeiter, @xPaw, what do we want to do with this?

MaxLeiter:
Rebased

MaxLeiter:
there's an issue with history on this -- states seem to be duplicated (so need to hit forward/backward twice)

@thelounge/maintainers - any clues? tried to debug, no success.

xPaw:
@MaxLeiter try doing
```
		.trigger("click", {
			pushState: false,
		});
```

MaxLeiter:
I did - changed nothing.

astorije:
Doesn't `window.location.hash = ...` add an entry in the history, explaining why you have duplicates?

MaxLeiter:
Fixed - PR is ready to go.

astorije:
@xPaw, mind giving a first review to this?

client/js/lounge.js
```diff
@@ -594,7 +594,8 @@ $(function() {
 			}
 		});
 	});
-	if ($("body").hasClass("public")) {
+
+	if ($("body").hasClass("public") && window.location.hash === "#connect") {
```
xPaw: Can we support `#channels` in public mode instead?

MaxLeiter:
Can you elaborate?

client/js/lounge.js
MaxLeiter: Can you elaborate?

MaxLeiter: oh, if public mode is trying to connect to channels, populate the channels textbox? Then you're assuming that the user also knows the network, etc. Also, we should probably encourage them to use https://github.com/thelounge/lounge/pull/674

YaManicKill: Yeah, that feels like 2 ways to solve the 1 issue.

MaxLeiter:
@xPaw can you 👍 and we can discuss your request in an issue/PR?

MaxLeiter:
...good to merge?

astorije:
Thanks!

PR #1288, opened 2017-06-30
Don't run anything after help() and don't create config until we have checked that

### LINKED ISSUE #520: The --help option shouldn't create config directory
I've been working on a startup script for lounge and found it convenient to use the `--home` option with the `start` command to store the files in a good location. Confusingly, using the `--help` option auto-creates the config directory and it's almost certainly not the location I want, since I didn't pass the `--home` option along with the `--help` option. I think it would make more sense if `--help` didn't touch the filesystem.

YaManicKill:
Fixes #520

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -15,6 +15,19 @@ program.version(Helper.getVersion(), "-v, --version")
 
 Helper.setHome(program.home || process.env.LOUNGE_HOME);
 
+require("./start");
+require("./config");
+require("./list");
+require("./add");
+require("./remove");
+require("./reset");
+require("./edit");
+
+if (!process.argv.slice(2).length || process.argv.includes("--help")) {
+	program.help();
+	process.exit();
+}
+
 if (!fs.existsSync(Helper.CONFIG_PATH)) {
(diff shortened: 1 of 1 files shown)


src/command-line/index.js
```diff
@@ -15,6 +15,19 @@ program.version(Helper.getVersion(), "-v, --version")
...
 
+require("./start");
+require("./config");
+require("./list");
+require("./add");
+require("./remove");
+require("./reset");
+require("./edit");
+
+if (!process.argv.slice(2).length || process.argv.includes("--help")) {
```
YaManicKill: I can't find a way to use commander to do this. You have to parse the program to get access to program.args, but as soon as you parse, you run the commands. So, sadly we'll have to stick with this. But it works.

YaManicKill:
Just realised this won't fix --version.

I'm not sure the best way to fix it.

Options:

1. Add a custom --version as well as custom --help (easiest solution, but duplicates some of the code)
2. Add a check into each of the command line options that creates the config only when invoked (lots of duplicated code there, but means we are using commander all the way)

Thoughts?

YaManicKill:
Closing in favour of #1350

PR #1293, opened 2017-07-03
Do not truncate link previews if viewport can fit more text

### LINKED ISSUE #1084: Link preview truncates before the right edge of the viewport
With user list **open** (notice the second block not taking full width):

screen shot 2017-04-25 at 00 13 57

With user list **closed** (both are wrong size):

screen shot 2017-04-25 at 00 14 09

This is anecdotic but I have noticed this several times.

@maxpoulin64, is there any chance you could help with that?

xPaw:
Fixes #1084. `float: left` on the thumbnail image pushed text, but did not increase container width.

Before:

After:

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1129,15 +1129,20 @@ kbd {
 }
 
 #chat .toggle-content .thumb {
-	float: left;
 	margin-right: 6px;
 	max-width: 48px;
 	max-height: 32px;
 }
 
+#chat .toggle-text {
+	min-width: 0;
+	display: flex;
+	flex-direction: column;
+	white-space: nowrap;
+}
+
 #chat .toggle-content .head,
 #chat .toggle-content .body {
-	white-space: nowrap;
 	text-overflow: ellipsis;
 	overflow: hidden;
 }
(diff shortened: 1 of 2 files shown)


astorije:
You had me super excited, but unfortunately it doesn't work correctly when reaching the end of the screen :(

screen shot 2017-07-03 at 11 00 39

^ No ellipsis + text on second line instead of next to thumbnail.

xPaw:
Damn it.

xPaw:
@astorije fixed with flexbox.

client/css/style.css
```diff
@@ -1129,15 +1129,21 @@ kbd {
 }
 
 #chat .toggle-content .thumb {
-	float: left;
+	display: flex;
```
astorije: Is this necessary at all?

client/views/msg_preview.tpl
```diff
@@ -8,11 +8,15 @@
 	{{else}}
 		{{#if thumb}}
 			<img src="{{thumb}}" class="thumb">
+			<div class="toggle-text">
```
astorije: Probably worth rendering this div whether or not the thumb appears for consistency sake.

xPaw: Will do

client/css/style.css
xPaw: How else would you do it?

astorije: I meant, this would only apply to its children, which .thumb doesn't have any, no?

PR #1294, opened 2017-07-03
Allow opting out of autocomplete

awalgarg:


astorije:
I have to admit I'm not too fond of this. It's very easy to want to add an opt-out option for features that seem annoying, but most of the time, it's because said features can be improved.

For example, someone already mentioned adding an opt-out option for @YaManicKill's condensing (even though it's not even ready!) just because the feature might not be perfect. I think our time (and UX) is better spent with improving the non-perfect features :)

That auto-complete was a lot of great work by multiple people, I'm sure we'll continue to make it even better, but I would rather have people list their pain points and/or help fixing them rather than just disable because they can.

That being said, if there is a strong push for this opt-out option, I won't push back, just trying to express my opinion on the matter :)

dgw:
It doesn't bother me sufficiently to argue in favor, but I can see where having autocomplete menus popping up all the time could get annoying if one never uses them. As more of a developer-type, I'd be happier with autocomplete triggering on a specific keypress (<kbd>Tab</kbd> or whatever), but eh… it's useful anyway.

awalgarg:
> but I would rather have people list their pain points and/or help
> fixing them rather than just disable because they can.

It is not "because they can". It is because certain people would want to
not have certain features if they are ever so slightly intrusive. An
autocompletion *popup* is one such thing (tab-completion on the other
hand is not so much IMO).

> That auto-complete was a lot of great work by multiple people, I'm
> sure we'll continue to make it even better

Hence why allowing to opt-out instead of disabling it by default. It is
indeed impressive how far a web irc client could come.

-- awal

astorije:
> It is not "because they can".

Perhaps I didn't explain myself correctly. I'm not saying that's what you're trying to do personally. I'm saying that a concern I have if we start adding all sorts of opting-out options for non-perfect features is that the temptation to just disable them instead of coming together to try and fix will be very strong.
As an example, many people (including maintainers) simply disabled previews because they sucked, and therefore it took a year before there was any progress made to them...

Now, if there are usability concerns/tradeoffs with features, outside of their level of completion, then it is definitely a reasonable argument to allow for opting-out, _assuming these concerns cannot be fixed otherwise_. Many times we simply improved the underlying feature, and the need to remove it went away (but again, I understand that it might not be the case here).

awalgarg:
> the temptation to just disable them instead of coming
> together to try and fix will be very strong.

That's a fair perspective. But...

> *assuming these concerns cannot be fixed otherwise*.
> Many times we simply improved the underlying feature,
> and the need to remove it went away (but again, I
> understand that it might not be the case here).

I am arguing that's the case here. A "popup" is by very
definition intrusive. Any amount of sugar coating or design
fixes would not help with that. My problem with this is not
one of design, I do not think that it looks ugly (not that I'd
have worthy opinions on design anyways). I am talking
about the inherent intrusiveness of any popup here. The
fact that it is meant to show up while typing makes it very
irksome.

-- awal

astorije:
I completely understand your argument ;)
I do not argue against your opinion that you personally feel the intrusiveness of a dropup, but because of how standard it is to have said popup these days in certain webapps (heck, the current GitHub textarea where I'm typing right now even has it! :P), to me this is a change that belongs to a personal fork rather than a project-wide setting.
But again, if @thelounge/maintainers do not share my point of view, I'll adjust to their perspective.

YaManicKill:
> I do not argue against your opinion that you personally feel the intrusiveness of a dropup, but because of how standard it is to have said popup these days in certain webapps (heck, the current GitHub textarea where I'm typing right now even has it! :P), to me this is a change that belongs to a personal fork rather than a project-wide setting.

Most of these things need to be decided on a case-by-case basis as to whether we allow opting out. I'm ok with this one because:

1. It is very in-your-face if you type ":" quite a lot
2. It is very little code, so not much to maintain.

It might be nice to change the autocomplete to plugins to remove this once we get to that point.

astorije:
Hey @awalgarg, we have sticker packs for our contributors
If you're interested, you can fill the form at https://goo.gl/forms/f5usqAEp5DWoeXC92 🙂

PR #1297, opened 2017-07-04
Fix background color contrast on Zenburn previews

astorije:
Again, since we're fixing previews for 2.3.3, here is another one, that fixes an accessibility issue.

Before | After
--- | ---
screen shot 2017-07-04 at 13 23 35 | screen shot 2017-07-04 at 13 22 59

### DIFF
--- a/client/themes/zenburn.css
+++ b/client/themes/zenburn.css
@@ -59,7 +59,7 @@ body {
 }
 
 #footer {
-	background: #33332f;
+	background: #333;
 	border-top: 1px solid #000;
 }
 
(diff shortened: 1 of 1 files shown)


client/themes/zenburn.css
```diff
@@ -59,7 +59,7 @@ body {
 }
 
 #footer {
-	background: #33332f;
+	background: #333;
```
astorije: Pretty sure this was a typo, because this is 1 digit off, so probably a <kbd>Ctrl</kbd><kbd>Down</kbd> in a text editor or something.

Junction between footer and input margin, before:

screen shot 2017-07-04 at 13 37 25

PR #1299, opened 2017-07-04
Add a transition when toggling the user list

astorije:
I remember looking at this some time ago, and because of the `transform` mechanism we were using, it wasn't possible. Now, it's doable with a single line of change 🎉 

It removes jump on markers (date, new message) and text/preview overflow.

Before | After
--- | ---
transition_before | transition_after

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -790,6 +790,7 @@ kbd {
 
 #chat .channel .chat {
 	right: 180px;
+	transition: right .4s;
 }
 
 #chat .special {


astorije:
@xPaw, I understand you weren't too happy with this. Do you want to close this PR or is there anything we can do to make it digestible? :)

xPaw:
@astorije I'm not sure we can do anything here.

astorije:
So, should we close? @thelounge/maintainers?
@xPaw's (legitimate concern) is the extra flickering that happens on the window content when toggling the user list. I haven't tried with channel filled up with discussions, so I'm not sure myself, but I understand the concern. Happy to leave this in or close it.

xPaw:
Let's close for now.

PR #1303, opened 2017-07-06
Support multiple previews per message

astorije:
- Load up to 5 previews per message (to avoid abuse)
- Do not load multiple times the same URL
- Prepare preview containers per message instead of appending (to maintain correct order)
- Store an array of previews instead of a single preview in `Msg` objects
- Consolidate preview rendering for new messages and upon refresh/load history (when rendering entire channels)
- Update `parse` tests to reflect previous point
- Add test for multiple URLs
- Switch preview tests from `assert` API to `expect` API

It's probably not the nicest code, but it works pretty nicely, and I'm sure reviews will highlight the required improvements.
Note that this does not improve the UI of displaying toggles/previews, this will be better left off to another PR and/or #1301.

### Example of multiple URLs (screenshots show an outdated toggle button)

multiple_previews
no_duplicate_previews

### UI of toggle button

screen shot 2017-07-06 at 03 34 33

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -205,6 +205,7 @@ kbd {
 #chat .whois .from:before,
 #chat .nick .from:before,
 #chat .action .from:before,
+#chat .toggle-button:after,
 .context-menu-item:before,
 #nick button:before {
 	font: normal normal normal 14px/1 FontAwesome;
@@ -291,6 +292,16 @@ kbd {
 
 #chat .action .from:before {
 	content: "\f005"; /* http://fontawesome.io/icon/star/ */
+}
+
+#chat .toggle-button {
+	/* These 2 directives are loosely taken from .fa-fw */
+	width: 1.35em;
+	text-align: center;
+}
+
+#chat .toggle-button:after {
+	content: "\f0da"; /* http://fontawesome.io/icon/caret-right/ */
 }
 
 #chat .count:before {
@@ -1097,17 +1108,18 @@ kbd {
 }
 
 #chat .toggle-button {
-	background: #f5f5f5;
-	border-radius: 2px;
 	display: inline-block;
 	color: #666;
-	height: 1em;
-	line-height: 0;
-	padding: 0 6px;
+	transition: color .2s, transform .2s;
 }
 
-#chat .toggle-button:after {
-	content: "···";
+#chat .toggle-button.opened {
+	transform: rotate(90deg);
+}
+
+#chat .toggle-button:hover {
+	/* transform and opacity together glitch, so need to use RGBA transition */
+	color: rgba(102, 102, 102, .8); /* #666 x .8 opacity */
 }
 
 #chat .toggle-content {
@@ -1118,7 +1130,7 @@ kbd {
 	font-size: 12px;
 	max-width: 100%;
 	padding: 6px;
-	margin-top: 2px;
+	margin: 2px 0;
 	overflow: hidden;
 }
 
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -81,5 +81,8 @@ module.exports = function parse(text) {
 		}
 
 		return fragments;
+	}).join("") + linkParts.map((part) => {
+		const escapedLink = Handlebars.Utils.escapeExpression(part.link);
+		return `<div data-url="${escapedLink}"></div>`;
 	}).join("");
 };
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -83,6 +83,6 @@ module.exports = function parse(text) {
 		return fragments;
 	}).join("") + linkParts.map((part) => {
 		const escapedLink = Handlebars.Utils.escapeExpression(part.link);
-		return `<div data-url="${escapedLink}"></div>`;
+		return `<div class="preview" data-url="${escapedLink}"></div>`;
 	}).join("");
 };
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -3,6 +3,7 @@
 const $ = require("jquery");
 const templates = require("../views");
 const options = require("./options");
+const renderPreview = require("./renderPreview");
 const utils = require("./utils");
 const sorting = require("./sorting");
 
@@ -15,7 +16,7 @@ module.exports = {
 	renderChannel,
 	renderChannelMessages,
 	renderChannelUsers,
-	renderNetworks
+	renderNetworks,
 };
 
 function buildChannelMessages(channel, messages) {
@@ -35,9 +36,9 @@ function buildChatMessage(data) {
 		target = "#chan-" + chat.find(".active").data("id");
 	}
 
-	if (data.msg.preview) {
-		data.msg.preview.shown = options.shouldOpenMessagePreview(data.msg.preview.type);
-	}
+	data.msg.previews.forEach((preview) => {
+		preview.shown = options.shouldOpenMessagePreview(preview.type);
+	});
 
 	const chan = chat.find(target);
 	let template = "msg";
@@ -71,6 +72,12 @@ function buildChatMessage(data) {
 
 	const msg = $(templates[template](data.msg));
 	const text = msg.find(".text");
+
+	if (data.msg.previews.length) {
+		data.msg.previews.forEach((preview) => {
+			renderPreview(preview, msg);
+		});
+	}
 
 	if (template === "msg_action") {
 		text.html(templates.actions[type](data.msg));
--- a/client/js/socket-events/msg_preview.js
+++ b/client/js/socket-events/msg_preview.js
@@ -1,51 +1,13 @@
 "use strict";
 
 const $ = require("jquery");
+const Handlebars = require("handlebars/runtime");
+const renderPreview = require("../renderPreview");
 const socket = require("../socket");
-const templates = require("../../views");
-const options = require("../options");
 
 socket.on("msg:preview", function(data) {
-	data.preview.shown = options.shouldOpenMessagePreview(data.preview.type);
-
 	const msg = $("#msg-" + data.id);
-	const container = msg.closest(".chat");
-	const bottom = container.isScrollBottom();
 
-	msg.find(".text").append(templates.msg_preview({preview: data.preview}));
-
-	if (data.preview.shown && bottom) {
-		handleImageInPreview(msg.find(".toggle-content"), container);
-	}
-
-	container.trigger("keepToBottom");
+	data.link = Handlebars.Utils.escapeExpression(data.link);
+	renderPreview(data.preview, msg);
 });
-
-$("#chat").on("click", ".toggle-button", function() {
-	const self = $(this);
-	const container = self.closest(".chat");
-	const content = self.parent().next(".toggle-content");
-	const bottom = container.isScrollBottom();
-
-	if (bottom && !content.hasClass("show")) {
-		handleImageInPreview(content, container);
-	}
-
-	content.toggleClass("show");
-
-	// If scrollbar was at the bottom before toggling the preview, keep it at the bottom
-	if (bottom) {
-		container.scrollBottom();
-	}
-});
-
-function handleImageInPreview(content, container) {
-	const img = content.find("img");
-
-	// Trigger scroll logic after the image loads
-	if (img.length && !img.width()) {
-		img.on("load", function() {
-			container.trigger("keepToBottom");
-		});
-	}
-}
--- a/client/views/msg.tpl
+++ b/client/views/msg.tpl
@@ -7,10 +7,5 @@
 			{{> user_name nick=from}}
 		{{/if}}
 	</span>
-	<span class="text">
-		{{~{parse text}~}}
-		{{#if preview}}
-			{{> msg_preview}}
-		{{/if}}
-	</span>
+	<span class="text">{{{parse text}}}</span>
 </div>
--- a/client/views/msg_preview_toggle.tpl
+++ b/client/views/msg_preview_toggle.tpl
@@ -1,0 +1,10 @@
+{{#preview}}
+<button class="toggle-button {{#if shown}} opened{{/if}}"
+  data-url="{{link}}"
+  {{#equal type "image"}}
+    aria-label="Toggle image preview"
+  {{else}}
+    aria-label="Toggle website preview"
+  {{/equal}}
+></button>
+{{/preview}}
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -23,14 +23,19 @@ module.exports = function(client, chan, msg) {
 		return;
 	}
 
-	const link = escapeHeader(links[0].link);
-	fetch(link, function(res) {
-		if (res === null) {
-			return;
-		}
+	Array.from(new Set( // Remove duplicate links
+		links.map((link) => escapeHeader(link.link))
+	))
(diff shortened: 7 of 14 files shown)


client/js/render.js
```diff
@@ -90,6 +99,54 @@ function buildChatMessage(data) {
 	return msg;
 }
 
+function renderPreview(preview, msg) {
```
xPaw: Can we keep all this preview code in `msg_preview`, otherwise we will end up having this render file huge again.

client/js/render.js
```diff
@@ -35,8 +36,10 @@ function buildChatMessage(data) {
 		target = "#chan-" + chat.find(".active").data("id");
 	}
 
-	if (data.msg.preview) {
-		data.msg.preview.shown = options.shouldOpenMessagePreview(data.msg.preview.type);
+	if (data.msg.previews.length) {
+		data.msg.previews.forEach((preview) => {
+			preview.shown = options.shouldOpenMessagePreview(preview.type);
```
xPaw: wtf, you want to open *all* previews at once?

src/plugins/irc-events/link.js
```diff
@@ -23,14 +23,19 @@ module.exports = function(client, chan, msg) {
 		return;
 	}
 
-	const link = escapeHeader(links[0].link);
-	fetch(link, function(res) {
-		if (res === null) {
-			return;
-		}
+	Array.from(new Set( // Remove duplicate links
```
xPaw: Could you add proper caching instead of just deduping? As long as lounge process is running, it would keep X recent urls in memory (and block while one request is in flight, like await an existing promise).

xPaw:
Is this going to append expand button per new line after the text? You added divs with urls, but I don't see any code changes to move buttons there.

client/views/msg.tpl
```diff
@@ -9,8 +9,8 @@
 	</span>
 	<span class="text">
 		{{~{parse text}~}}
-		{{#if preview}}
+		{{#each previews}}
 			{{> msg_preview}}
```
xPaw: This doesn't seem to match what you're doing in appends.

client/js/libs/handlebars/parse.js
```diff
@@ -81,5 +81,6 @@ module.exports = function parse(text) {
 		}
 
 		return fragments;
-	}).join("");
+	}).join("") +
+	linkParts.map((part) => `<div data-url="${part.link}"></div>`).join("");
```
xPaw: `part.link` is not escaped.

xPaw:
Add a test with multiple URLs, both for parse and link previews.

client/js/render.js
astorije: If we do that, then we need to do a `require("client/js/socket-events/msg_preview")` in this file, which I think we shouldn't do. The code in this function is very much related to the renderer, so I think we should leave it there, and modularize the renderer in a later PR if necessary.
For the `click` event, however, I'm less happy to move it here, but we have the same need for `require` otherwise :/

client/views/msg.tpl
astorije: Correct, this should just all go, I just forgot.

client/js/render.js
xPaw: Why exactly do you need it? You already append the preview in template, and you seem to be duplicating it here.

src/plugins/irc-events/link.js
astorije: I agree with the intent, but this goes way beyond the goal of this PR and the time I can allocate on this. I would rather keep the semi-naive solution I have right now (which is not worse than what we have in `master` anyway), and we can work on proper caching/deduping in a later PR, IMO.

client/views/msg.tpl
astorije: Done.

client/js/libs/handlebars/parse.js
astorije: Fixed. Is there something better I could do though?

client/js/render.js
astorije: Previews aren't appended in the template anymore.

I'm not scared of the `socket-events` folder lol. When refreshing the page, we're not getting the preview as a socket event, it doesn't make sense to load it from there. Everything gets loaded from the chain in `render` (build channels -> build messages -> etc.), and the rendering of previews is a part of that chain. So definitely makes sense to have it there.

xPaw: Having it as part of render doesn't stop us from splitting the code.

client/js/render.js
astorije: What do you mean? I haven't improved/changed the UI one bit, so before that PR, if you were posting a URL and that would show as open, with this PR that same behavior applies for multiple URLs.
One reason that I didn't touch UI is because you are improving it in #1301 and I think we should keep those improvements there.

I'm still debating whether we want the caret right next to the URLs (à la Slack) or not, but to be honest the current UI works fairly well even with multiple URLs. Sure, it's taking a lot of space for nothing, but I'm not saying it's definitive :)

astorije:
> Is this going to append expand button per new line after the text? You added divs with urls, but I don't see any code changes to move buttons there.

Correct. Not ideal, but not a definitive solution, and I'm leaving UI improvements to your #1301 that we can build upon after this PR lands. I'm also not planning to release this PR without your work in #1301 :)

client/js/render.js
xPaw: You're changing how many previews are rendered though. So with this PR I assume sending 5 links will render all 5 previews at once? Seems like quite the overkill.

xPaw:
> I'm leaving UI improvements to your #1301 that we can build upon after this PR lands

That PR is unnecessary if buttons are going to appear after links.

astorije:
> Add a test with multiple URLs, both for parse and link previews.

What do you mean? I already did add/update tests, no?

xPaw:
>  I already did add/update tests, no?

There's no test that has multiple `<div data-url>`

client/js/libs/handlebars/parse.js
```diff
@@ -81,5 +81,8 @@ module.exports = function parse(text) {
 		}
 
 		return fragments;
+	}).join("") + linkParts.map((part) => {
+		const escapedLink = Handlebars.Utils.escapeExpression(part.link);
+		return `<div data-url="${escapedLink}"></div>`;
```
astorije: Actually, escaping this is going to make finding stuff with jQuery difficult (it's matching based on `preview.link`, which is unescaped). Do we actually need to escape this, here? It's not for show, it's mostly a key.

xPaw: Do you want a XSS? That's how you're going to get an XSS.

astorije: Fair enough 😅 What do you recommend? Because right now rendering of something that needs escaping will fail.

astorije:
> That PR is unnecessary if buttons are going to appear after links.

Are we set on this idea though? I think it's reasonable to settle on the actual capability of having multiple previews, and then play with different UIs, as yours is already an improvement.

> There's no test that has multiple `<div data-url>`

Will add a test for `parse`, I believe that's the only missing one, no?

astorije:
Test added.

client/js/libs/handlebars/parse.js
astorije: Actually, this is fixed now, yay

client/js/render.js
```diff
@@ -72,6 +75,12 @@ function buildChatMessage(data) {
 	const msg = $(templates[template](data.msg));
 	const text = msg.find(".text");
 
+	if (data.msg.previews.length) {
+		data.msg.previews.forEach((preview) => {
```
xPaw: This can be oneliner

client/js/socket-events/msg_preview.js
```diff
@@ -1,51 +1,13 @@
...
-		handleImageInPreview(content, container);
-	}
-
-	content.toggleClass("show");
-
-	// If scrollbar was at the bottom before toggling the preview, keep it at the bottom
-	if (bottom) {
-		container.scrollBottom();
-	}
+	data.link = Handlebars.Utils.escapeExpression(data.link);
```
xPaw: Are you sure this isn't going to have side effects like double escapes?

client/views/msg_preview_toggle.tpl
```diff
@@ -0,0 +1 @@
+<button class="toggle-button opened" data-url={{url}}	aria-label="Toggle prefetched media"></button>
```
xPaw: Was no quotes intentional?

astorije: image

astorije: Fixed.

client/js/socket-events/msg_preview.js
astorije: In my understanding, it's coming back unescaped from the `preview` socket events. That's why we also escape when getting the `msg` event.

client/js/render.js
astorije: We can't because of this rule: http://eslint.org/docs/rules/brace-style.
Alternative is to do `data.msg.previews.forEach((preview) => renderPreview(preview, msg));`, which has an implicit return value that we shouldn't convey, so this is actually the best option IMO.

astorije: Essentially, the following:

```js
data.msg.previews.forEach((preview) => renderPreview(preview, msg));
```

translates to:

```js
data.msg.previews.forEach((preview) => {
  return renderPreview(preview, msg);
});
```

and not:

```js
data.msg.previews.forEach((preview) => {
  renderPreview(preview, msg);
});
```

That extra `return` is unwanted.

client/css/style.css
```diff
@@ -293,6 +294,16 @@ kbd {
 	content: "\f005"; /* http://fontawesome.io/icon/star/ */
 }
 
+#chat .toggle-button {
+	/* These 2 directives are directly from .fa-fw */
+	width: 1.28571429em;
```
xPaw: I don't think we need such a precise width, the button can be wider to hit it easily.

client/js/socket-events/msg_preview.js
xPaw: But then the `msg_preview` template is going to escape it again, is it not? Try testing it with `&` in url or something.

client/views/msg_preview_toggle.tpl
```diff
@@ -0,0 +1 @@
+<button class="toggle-button opened" data-url="{{url}}" aria-label="Toggle prefetched media"></button>
```
xPaw: Pretty sure we are able to change text depending on type, so `Toggle image` and `Toggle link preview`?

xPaw:
Looks like `fetch` is blocking and it fetches urls one by one instead of asynchronously.

client/views/msg_preview_toggle.tpl
astorije: I can do that most likely yeah.

astorije: Done.

client/css/style.css
astorije: Done. Button is now a bit bigger, and roughly square (takes all height).

client/js/socket-events/msg_preview.js
astorije: I just tried with the msdn URL given in test for `parse`, and yeah it does work fine :)

astorije:
Re: concern of being blocking, I did some tests on the server, using:

```diff
───────────────────────────────────
modified: src/plugins/irc-events/link.js
───────────────────────────────────
@ link.js:31 @ module.exports = function(client, chan, msg) {
    ))
        .slice(0, 5) // Only preview the first 5 URLs in message to avoid abuse
        .forEach((link) => {
+           console.log(`Starting ${link}`);
            fetch(link, function(res) {
                if (res === null) {
                    return;
                }

+               console.log(`Finished ${link}`);
+
                parse(msg, link, res, client);
            });
        });
```

Posting this message (`https://cloudhealthtech.com https://thelounge.github.io https://www.facebook.com https://www.google.com https://imgur.com/gallery/51haX`) multiple times, I get different orders:

```
Starting https://cloudhealthtech.com
Starting https://thelounge.github.io
Starting https://www.facebook.com
Starting https://www.google.com
Starting https://imgur.com/gallery/51haX
Finished https://www.google.com
Finished https://www.facebook.com
Finished https://imgur.com/gallery/51haX
Finished https://thelounge.github.io
Finished https://cloudhealthtech.com

Starting https://cloudhealthtech.com
Starting https://thelounge.github.io
Starting https://www.facebook.com
Starting https://www.google.com
Starting https://imgur.com/gallery/51haX
Finished https://thelounge.github.io
Finished https://imgur.com/gallery/51haX
Finished https://www.google.com
Finished https://cloudhealthtech.com
Finished https://www.facebook.com
```

So unless I'm missing something, it looks fine.

astorije:
@xPaw, I believe I have addressed all your comments, including splitting the rendering code into its own module :))

client/css/style.css
```diff
@@ -293,6 +294,16 @@ kbd {
 	content: "\f005"; /* http://fontawesome.io/icon/star/ */
 }
 
+#chat .toggle-button {
+	/* These 2 directives are loosely taken from .fa-fw */
```
xPaw: No longer valid comment

client/views/msg_preview_toggle.tpl
```diff
@@ -0,0 +1,10 @@
+{{#preview}}
+<button class="toggle-button {{#if shown}} opened{{/if}}"
+  data-url="{{link}}"
+  {{#equal type "image"}}
+    aria-label="Toggle image preview"
```
xPaw: Is this not going to show up on hover? I don't see tooltipped classes.

client/css/style.css
astorije: Nah it is, I updated it from `directly from` to `loosely taken from` :D Just a reminder to keep an eye on `.fw-fa` in the future.

client/views/msg_preview_toggle.tpl
astorije: No, it was always `aria-label`, not tooltips 😅
I'm up for making these tooltips, but mind doing that later after merging? I'm going to be fairly busy this weekend, hence me addressing comments at 5am :)

astorije: See https://github.com/thelounge/lounge/pull/1303/files#diff-22e42391cd970f0b0b44672abb685a0aL3, and no tooltips in UI right now.

client/js/render.js
YaManicKill: > That extra return is unwanted.

Yeah, but it doesn't cause any problems. I wouldn't have an issue with this, tbh.

client/js/libs/handlebars/parse.js
```diff
@@ -81,5 +81,8 @@ module.exports = function parse(text) {
 		}
 
 		return fragments;
+	}).join("") + linkParts.map((part) => {
+		const escapedLink = Handlebars.Utils.escapeExpression(part.link);
+		return `<div class="preview" data-url="${escapedLink}"></div>`;
```
YaManicKill: I'm not 100% sure about putting these right next to the link, but I can't think of a better option. It's probably the best we can do when there are multiple previews.

astorije: Yeah, alternatively it was messing with jQuery to insert at the right position depending on the order, etc. so this feels more "natural" but I completely feel you. Nothing we can't update if a better solution presents itself anyway

xPaw:
For history: This PR broke storage de-referencing because it didn't update the relevant code. Fixed in #2824

PR #1307, opened 2017-07-06
Store preview images on disk for privacy, security and caching

xPaw:
- [x] Add `block-all-mixed-content; img-src 'self'` to our CSP.
- [x] Filenames are SHA256 of the file, split into 2 subfolders.
- [x] Fix tests
- [x] Allow disabling this in the config
- [x] Empty storage on startup
- [x] Delete images when they are no longer referenced in message history

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -67,6 +67,23 @@ module.exports = {
 	prefetch: false,
 
 	//
+	// Store and proxy prefetched images and thumbnails.
+	// This improves security and privacy by not exposing client IP address,
+	// and always loading images from The Lounge instance and making all assets secure,
+	// which in result fixes mixed content warnings.
+	//
+	// If storage is enabled, The Lounge will fetch and store images and thumbnails
+	// in ~/.lounge/storage folder, or %HOME%/storage if --home is used.
+	//
+	// Images are deleted when they are no longer referenced by any message (controlled by maxHistory),
+	// and the folder is cleaned up on every The Lounge restart.
+	//
+	// @type     boolean
+	// @default  false
+	//
+	prefetchStorage: false,
+
+	//
 	// Prefetch URLs Image Preview size limit
 	//
 	// If prefetch is enabled, The Lounge will only display content under the maximum size.
--- a/src/helper.js
+++ b/src/helper.js
@@ -12,6 +12,7 @@ const colors = require("colors/safe");
 var Helper = {
 	config: null,
 	expandHome: expandHome,
+	getStoragePath: getStoragePath,
 	getUserConfigPath: getUserConfigPath,
 	getUserLogsPath: getUserLogsPath,
 	setHome: setHome,
@@ -88,6 +89,10 @@ function getUserConfigPath(name) {
 
 function getUserLogsPath(name, network) {
 	return path.join(this.HOME, "logs", name, network);
+}
+
+function getStoragePath() {
+	return path.join(this.HOME, "storage");
 }
 
 function ip2hex(address) {
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -2,6 +2,7 @@
 
 var _ = require("lodash");
 var Helper = require("../helper");
+const storage = require("../plugins/storage");
 
 module.exports = Chan;
 
@@ -53,7 +54,15 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 	this.messages.push(msg);
 
 	if (Helper.config.maxHistory >= 0 && this.messages.length > Helper.config.maxHistory) {
-		this.messages.splice(0, this.messages.length - Helper.config.maxHistory);
+		const deleted = this.messages.splice(0, this.messages.length - Helper.config.maxHistory);
+
+		if (Helper.config.prefetch && Helper.config.prefetchStorage) {
+			deleted.forEach((deletedMessage) => {
+				if (deletedMessage.preview && deletedMessage.preview.thumb) {
+					storage.dereference(deletedMessage.preview.thumb);
+				}
+			});
+		}
 	}
 
 	if (!msg.self && !isOpen) {
--- a/src/plugins/storage.js
+++ b/src/plugins/storage.js
@@ -1,0 +1,81 @@
+"use strict";
+
+const fs = require("fs");
+const fsextra = require("fs-extra");
+const path = require("path");
+const crypto = require("crypto");
+const helper = require("../helper");
+
+class Storage {
+	constructor() {
+		this.references = new Map();
+
+		// Ensures that a directory is empty.
+		// Deletes directory contents if the directory is not empty.
+		// If the directory does not exist, it is created.
+		fsextra.emptyDirSync(helper.getStoragePath());
+	}
+
+	dereference(url) {
+		// If maxHistory is 0, image would be dereferenced before client had a chance to retrieve it,
+		// so for now, just don't implement dereferencing for this edge case.
+		if (helper.maxHistory === 0) {
+			return;
+		}
+
+		const references = (this.references.get(url) || 0) - 1;
+
+		if (references < 0) {
+			return log.warn("Tried to dereference a file that has no references", url);
+		}
+
+		if (references > 0) {
+			return this.references.set(url, references);
+		}
+
+		this.references.delete(url);
+
+		// Drop "storage/" from url and join it with full storage path
+		const filePath = path.join(helper.getStoragePath(), url.substring(8));
+
+		fs.unlink(filePath, (err) => {
+			if (err) {
+				log.error("Failed to delete stored file", err);
+			}
+		});
+	}
+
+	store(data, extension, callback) {
+		const hash = crypto.createHash("sha256").update(data).digest("hex");
+		const a = hash.substring(0, 2);
+		const b = hash.substring(2, 4);
+		const folder = path.join(helper.getStoragePath(), a, b);
+		const filePath = path.join(folder, `${hash.substring(4)}.${extension}`);
+		const url = `storage/${a}/${b}/${hash.substring(4)}.${extension}`;
+
+		this.references.set(url, 1 + (this.references.get(url) || 0));
+
+		// If file with this name already exists, we don't need to write it again
+		if (fs.existsSync(filePath)) {
+			return callback(url);
+		}
+
+		fsextra.ensureDir(folder).then(() => {
+			fs.writeFile(filePath, data, (err) => {
+				if (err) {
+					log.error("Failed to store a file", err);
+
+					return callback("");
+				}
+
+				callback(url);
+			});
+		}).catch((err) => {
+			log.error("Failed to create storage folder", err);
+
+			return callback("");
+		});
+	}
+}
+
+module.exports = new Storage();
--- a/src/server.js
+++ b/src/server.js
@@ -33,6 +33,10 @@ module.exports = function() {
 		.use(allRequests)
 		.use(index)
 		.use(express.static("client"))
+		.use("/storage/", express.static(Helper.getStoragePath(), {
+			redirect: false,
+			maxAge: 86400 * 1000,
+		}))
 		.engine("html", expressHandlebars({
 			extname: ".html",
 			helpers: {
@@ -152,7 +156,24 @@ function index(req, res, next) {
 			filename: filename
 		};
 	});
-	res.setHeader("Content-Security-Policy", "default-src *; connect-src 'self' ws: wss:; style-src * 'unsafe-inline'; script-src 'self'; child-src 'self'; object-src 'none'; form-action 'none';");
+
+	const policies = [
+		"default-src *",
+		"connect-src 'self' ws: wss:",
+		"style-src * 'unsafe-inline'",
+		"script-src 'self'",
+		"child-src 'self'",
+		"object-src 'none'",
+		"form-action 'none'",
+	];
+
+	// If prefetch is enabled, but storage is not, we have to allow mixed content
+	if (Helper.config.prefetchStorage || !Helper.config.prefetch) {
+		policies.push("img-src 'self'");
+		policies.unshift("block-all-mixed-content");
+	}
+
+	res.setHeader("Content-Security-Policy", policies.join("; "));
 	res.setHeader("Referrer-Policy", "no-referrer");
 	res.render("index", data);
 }
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ package-lock.json
 
 .nyc_output/
 coverage/
+test/fixtures/.lounge/storage/
 
 # Built assets created at npm install/prepublish time
 # See https://docs.npmjs.com/misc/scripts
--- a/client/views/msg_preview.tpl
+++ b/client/views/msg_preview.tpl
@@ -2,7 +2,7 @@
 <div class="toggle-content toggle-type-{{type}}{{#if shown}} show{{/if}}">
 	{{#equal type "image"}}
 		<a class="toggle-thumbnail" href="{{link}}" target="_blank" rel="noopener">
-			<img src="{{link}}">
+			<img src="{{thumb}}">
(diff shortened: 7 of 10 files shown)


astorije:
> split into 2 subfolders

Just curious, why that?

xPaw:
> Just curious, why that?

Some reading: https://softwareengineering.stackexchange.com/questions/301400/why-is-the-git-git-objects-folder-subdivided-in-many-sha-prefix-folders

astorije:
> Allow disabling this in the config

What are the benefits of *not* enabling that? If there are none, might as well make this stock feature, no?

xPaw:
> What are the benefits of not enabling that?

Perhaps not everyone wants to proxy the images and store them on disk. This is the kind of feature where toggle should be implemented.

astorije:
> Some reading: https://softwareengineering.stackexchange.com/questions/301400/why-is-the-git-git-objects-folder-subdivided-in-many-sha-prefix-folders

Good stuff! I do think we are safe with a single sub-dir, maybe even none, but whatever, it's not a big deal (as long as none of these can be listed from a web browser or something).

> Perhaps not everyone wants to proxy the images and store them on disk. This is the kind of feature where toggle should be implemented.

Fair enough. What do you think, enabled or disabled by default?

YaManicKill:
> What do you think, enabled or disabled by default?

Enable. People who care will look into it, but people who don't care will not look into how to enable it, and that'll reduce their privacy by default.

astorije:
That's reasonable. Maybe a warning at startup if filesystem has less than N free megabytes then? Like, default max picture size is 512kB, if less than 512mB then warn at startup (1000 images)?
Or have a default quota or number of pictures before deleting the oldest ones?

I don't know, I don't want to overthink this, but I don't want either to bloat a filesystem or make this super easy to kill a server by feeding it thousands of random pictures, if that makes sense. It seems like an easy attack vector to me if there is no failsafe.

dgw:
`storage` has no delete method? No cleanup, manually initiated or age-based? I'd disable this feature immediately upon upgrading to a version that includes it if there's no pruning.

xPaw:
@dgw need to figure out a prune strategy, as deleted images would show up as errors in the client (depending on the limit, and how we decided to delete files).

astorije:
> We could clear the folder on lounge startup, as we don't load log from disk (and when we will, it will re-fetch the preview information again anyway).

I'd rather not do that. Just like when I enable logs, I don't want them to be remove at startup. I don't know, it's not a strong argument against removing images at startup, but I'd rather avoid that if we can.
For example, things like `pm2` that restart on crash could make this experience pretty bad, no? On the other hand, some people rarely restart their servers, etc.

dgw:
I'm afraid to suggest it, but pruning strategy might be a good setting (not in the UI, obviously). Or maybe it just has to depend on the logging settings (prune on startup if logs are turned off, for example).

Does The Lounge have a mechanism to detect whether it exited cleanly or not? That could be worth the extra complexity, pruning on startup only if the last exit was clean. Or maybe the startup pruning function could only remove images older than _n_ days. Another option would be removing the oldest cached images on a continuing basis when disk usage by the preview function surpasses a given threshold. (@astorije already mentioned this type of strategy.)

The main thing is to take this convenience/privacy feature and balance it against the reality of limited server resources. My objection is not to The Lounge using disk space per se; it's allowing unlimited growth. There are a number of ways of solving the problem, several of which have been mentioned already. The hard part is deciding which one is The Best™. 😆

astorije:
> I'm afraid to suggest it, but pruning strategy might be a good setting

Damn, I am so scary? 😅
I know it's been said, said and said again, but that would be a good candidate for plugins. We come up with the strategy we think is best, give the possibility to disable it, and when plugins are available (maybe before the heat death of the universe), new strategies can come up.

Not trying to be anti-setting (all discussed above will reasonably be in at least 2 settings), but providing options for multiple strategies means that we need to come up with them, code them, test them, maintain them and make sure we are breaking none of them over time. That's a lot of efforts for a team that small :)

astorije:
@xPaw, I am probably waay wrong asking that, and I know that even if this is technically possible, there are lot of tradeoffs and drawbacks that the current solution hasn't (but also advantages), but just wondering: would there be a way that the proxied URLs would just be a redirect on the server?

Sure, no caching, and several other issues (I'm not saying this is _the right way_, just wondering if this is technically possible), but that would make for a solution that doesn't store anything on the server.

Whenever we need file uploads, that won't be enough anymore, but I'm still curious, and also we probably won't want to purge file uploads automatically!

dgw:
> Damn, I am so scary? 😅

Let's just say you've built a reputation over the last several months. 😆 

Your points about maintaining pruning strategies are well put. A future hook for plugins to enable/alter/suppress pruning would be adequate, I think. (Yes, hopefully before the heat death of the universe…)

When you say "a redirect", I suspect you're actually meaning that the server would act as a reverse proxy: fetching the referenced image from the origin server, and returning it to the client while pretending that the image came from The Lounge so the origin server sees only The Lounge's IP address, and not the end user's browser's IP. The main issue with that would be increased bandwidth usage—2*n* bytes instead of (*n* + 1) bytes, for *n* times a client loads a given image.

astorije:
> Let's just say you've built a reputation over the last several months. 😆

Well, I hope the reputation is not *that bad* :)
It's true that I do question the need for new options/settings before we introduce them. It doesn't mean I'm against, but asking why is reasonable IMO. In many cases, we discovered that there was a deeper issue and that we were better off fixing it.
I believe I have approved adding settings more than denied 😅 But in general, adding settings/options is an easy way out (sometimes the only/best one, but still easy one) and you need to assess them before doing so, at the very least to make sure usability and maintainability are not negatively impacted.
Does that make sense? I don't want to come across as the opposite of what I'm trying to achieve :)

> When you say "a redirect", I suspect you're actually meaning that the server would act as a reverse proxy [...] The main issue

Correct, that's what I meant.
And yeah, it does have a bunch of issues (or tradeoffs, rather), but I was merely wondering if that was technically possible. If it turned out that permanently adding files on the filesystem ended up being a problem, that could have been an alternative. I am not going to push for that, I do prefer @xPaw's solution anyway, just considering options.

YaManicKill:
> (maybe before the heat death of the universe)

We are moving in the right direction, I promise. #1266 is technically plugins, it's just themes. So this is happening, be patient with me :-P I'll get that one finished soon, finish off #759 and then get onto some more plugin stuff. But we are moving.

dgw:
> Well, I hope the reputation is not that bad :) […] Does that make sense? I don't want to come across as the opposite of what I'm trying to achieve :)

I'm kinda just pulling your chain, haha. It's all about the bigger picture with you, and I like that. Sometimes people (like me) get bogged down in details, and you force us to step back and realize that maybe there's a better solution none of us has seen because we're focused on the color of the button that triggers the current one. :)

xPaw:
> would there be a way that the proxied URLs would just be a redirect on the server?

This was the initial idea until I realized that we already fetch the images as part of the link preview generation (to check that images are small and exist), then the initial implementation was to store that downloaded data in memory, and after I added disk storage.

src/helper.js
```diff
@@ -90,6 +91,10 @@ function getUserLogsPath(name, network) {
 	return path.join(this.HOME, "logs", name, network);
 }
 
+function getStoragePath() {
+	return path.join(this.HOME, "storage");
```
astorije: I think we will want to improve that in the near future. .lounge mostly has config stuff right now so I personally have it in /etc/ for now. It's not ideal with user logs, but file storage makes this even more inappropriate.
Any suggestions?

src/server.js
```diff
@@ -152,7 +156,24 @@ function index(req, res, next) {
...
+	];
+
+	// If prefetch is enabled, but storage is not, we have to allow mixed content
+	if (Helper.config.prefetchStorage || !Helper.config.prefetch) {
+		policies.push("img-src 'self'");
+		policies.unshift("block-all-mixed-content");
+	}
+
+	res.setHeader("Content-Security-Policy", policies.join("; "));
 	res.setHeader("Referrer-Policy", "no-referrer");
```
astorije: Noice!!

src/storage.js
```diff
@@ -0,0 +1,81 @@
...
+const helper = require("./helper");
+
+class Storage {
+	constructor() {
+		this.references = new Map();
+
+		// Ensures that a directory is empty.
+		// Deletes directory contents if the directory is not empty.
+		// If the directory does not exist, it is created.
+		fsextra.emptyDirSync(helper.getStoragePath());
```
astorije: I'm still unsure about clearing at startup. More importantly, a FS could still be filled until server gets restarted. I know you have that garbage collector when no references exist anymore, but that still leaves room for fs problems.

src/storage.js
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	dereference(url) {
+		// If maxHistory is 0, image would be dereferenced before clietn had a chance to retrieve it,
```
astorije: clietn

test/fixtures/.lounge/config.js
```diff
@@ -3,6 +3,7 @@
 var config = require("../../../defaults/config.js");
 
 config.prefetch = true;
+config.prefetchStorage = false;
```
astorije: Are you going to add tests for this feature or is there something blocking you from that?

src/helper.js
dgw: Why not /var/lib/ instead? I have znc and sopel in there, with their service accounts' home directories set up in subfolders therein. Could be /var/lib/lounge

xPaw: I was thinking of using system temp folder.

astorije:
> We are moving in the right direction, I promise. [...] be patient with me :-P [...] But we are moving.

Sorry, my bad my bad, I didn't mean to attack you personally or anyone else. It was just a (poorly phrased) joke about the fact that we have been saying "when plugins are supported" for ages now. I
 appreciate immensely your time and efforts spent on the project, I would never make you a reproac
 that you are not doing something or not fast enough :) 🙏

src/helper.js
astorije: It could be even worse with distro using tmpfs by default, no?

YaManicKill:
> Sorry, my bad my bad, I didn't mean to attack you personally or anyone else. It was just a (poorly phrased) joke about the fact that we have been saying "when plugins are supported" for ages now.

I know, I was joking with you 😛

defaults/config.js
```diff
@@ -67,6 +67,20 @@ module.exports = {
 	prefetch: false,
 
 	//
+	// Store and proxy prefetched images and thumbnails.
+	// This improves security and privacy by not exposing client IP address,
+	// and always loading images from The Lounge instance and making all assets secure,
+	// which in result fixes mixed content warnings.
```
YaManicKill: :tada:

src/helper.js
YaManicKill: Or allow admins to specify a folder.

*cough*

src/models/chan.js
```diff
@@ -53,7 +54,15 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 	this.messages.push(msg);
 
 	if (Helper.config.maxHistory >= 0 && this.messages.length > Helper.config.maxHistory) {
-		this.messages.splice(0, this.messages.length - Helper.config.maxHistory);
+		const deleted = this.messages.splice(0, this.messages.length - Helper.config.maxHistory);
+
+		if (Helper.config.prefetch && Helper.config.prefetchStorage) {
```
YaManicKill: I think this is good for first step. We will probably want to think up some more ways of dealing with this (the max storage space, etc) but I think this fine to start with, especially as we allow people to turn it off.

src/storage.js
YaManicKill: > I'm still unsure about clearing at startup. 

Hmm? Why? Currently, history is empty at startup, so it makes sense for nothing to be there. When we add history loading, then maybe it'll be different.

> More importantly, a FS could still be filled until server gets restarted.

Yeah, but I do think this is a good first step, personally. What's your suggestion for this first PR?

astorije: I mean, it's very likely that it's going to be enough in 90% cases, but I'm wondering how much time will happen before someone opens a ticket with a blown up FS :/ Maybe a month from now, or maybe never, who knows.

Do you think this is good enough, or do we need to add a setting for allowed storage size?

When I think about it, the logs will not be filled with proxified URLs, so that's not so annoying. But the problem of filling up space is still there, and whenever `Storage` is used for file uploads, we have to nuke this strategy: it would be pretty bad to destroy all uploaded files at startup!
I know this is not done now, just thinking forward since in his PR description, @xPaw mentioned:

> ` Storage` was designed to be used for file uploads in the future ;)

astorije: Look at me, suggesting new settings. Who knew this day would come?! 😅

src/helper.js
astorije: > Or allow admins to specify a folder.

Agreed, but this needs to be solved for all dirs we assume, I would say, don't you think?

Anyway, if it is the long-term goal, then this is probably good enough.

> I was thinking of using system temp folder.

`/var/tmp`? It's usually not `tmpfs`.

Otherwise, I am good with what's in this PR, at least for now.

test/fixtures/.lounge/config.js
astorije: Pretty pleeeease 🙏 :)
Something like [this line](https://github.com/thelounge/lounge/blob/master/test/plugins/link.js#L161) could check the format of the link, no? And with our current tests, we'd still have coverage for when the setting is disabled.
Now you'd have to be able to run tests with different config files, I don't know if any of our test suite would need refactor for that.

src/storage.js
xPaw: So basically I can't implement this in the current state. Images will keep existing if there's any message in memory (buffer) that references it. If we are going to delete images before they are removed from buffer, we will need to find all references and remove them then.

Right now it is not possible to re-download if it's not on the disk.

test/fixtures/.lounge/config.js
xPaw: Added tests.

astorije:
Good job re: tests!

I'm a bit confused, what's the status of the dereferencing stuff? Is there anything you want to do on this PR?

xPaw:
The PR should be done.

src/helper.js
YaManicKill: > Agreed, but this needs to be solved for all dirs we assume, I would say, don't you think?

Yes, probably, but let's not do the "All or nothing" thing with it.

MaxLeiter:
Been running for 24 hours now and no problems 👍

xPaw:
I think we should play on the safe side and have this disabled by default. So people aware of the option can enable it.

src/helper.js
astorije: Totally agree, but we need to have a global strategy/direction before going heads first. Still beyond the reach of this PR, I think. Until then, I'm good with the current.

src/plugins/storage.js
```diff
@@ -0,0 +1,81 @@
+"use strict";
```
astorije: That's an interesting location. Why not just `src/` instead of `src/plugins/`? It's not like our current "plugins" folder contains actual plugins. I think we should leave it with the others in `src/`

defaults/config.js
```diff
@@ -67,6 +67,20 @@ module.exports = {
...
+	//
+	// If storage is enabled, The Lounge will fetch and store images and thumbnails
+	// in ~/.lounge/storage folder, or %HOME%/storage if --home is used.
+	//
+	// @type     boolean
+	// @default  true
```
astorije: `false`

src/plugins/storage.js
xPaw: I think it's a nice little folder for unique components. @YaManicKill has also been doing this (e.g. themes.js) and I did this with push.js in the other PR.

Would keep the main folder clean with a couple of 'global' files.

PR #1309, opened 2017-07-07
Add collapse/expand commands to toggle all previews

astorije:
This is a complete rip-off from Slack. I cannot count how many times I am using this every week at work.

This will only be functional after #1303 gets merged, and it will be even more beneficial to have these when supporting multiple previews.

collapse


astorije:
Note that this needs #1303 to be actually functional

PR #1310, opened 2017-07-07
Reword link preview settings to better match reality

astorije:
- **s/Links and URLs/Link previews/**: From a user perspective, these are the same thing... These options are for the prefetcher, not the URLs themselves
- **s/thumbnails/images/**: What we call thumbnails are for "site" previews, not actual image links
- **s/links/websites/**: Technically, both image and non-image links are links, "websites" carries a tiny bit better the meaning

These are not perfect or ideal for sure, but I feel like these changes reflect better what the settings actually do. Open to suggestions if you feel these are way wrong.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -282,18 +282,18 @@
 							</div>
 							{{#if prefetch}}
 							<div class="col-sm-12">
-								<h2>Links and URLs</h2>
+								<h2>Link previews</h2>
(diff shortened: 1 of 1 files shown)


MaxLeiter:
👍 

I foresee a lot of rebasing #1271 in the future....

PR #1317, opened 2017-07-10
Add service worker for caching

MaxLeiter:
I benchmarked on my machine and saw some improvements, but I'm not 100% sure how to correctly benchmark this so it could just be random. Due to webpack/handlebars, I have it not caching `.tpl` and `.js` files. Some heavy profiling should be done so we're sure it's an improvement.


bews:
Isn't this can be easily achieved by using any proxy, nginx for example?

MaxLeiter:
`Cache-Control: immutable ` probably ends in same result, but this way you don't need to set up a proxy

PR #1318, opened 2017-07-10
Automatically load history when scrolling upwards

### LINKED ISSUE #207: Automatically load more history when scrolling up

xPaw:
This is a different implementation compared to #1148. This uses `IntersectionObserver` to do the heavy lifting without requiring us to deal with scroll positions and counting the messages in DOM.

Technically we can use this approach to lazily load messages on initial page load, as the callback would be callback for empty channel when you switch to it.

Chrome, Opera, Edge, and Firefox support it. Other browsers supported with polyfill.

Fixes #207.

### DIFF
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -38,7 +38,13 @@ socket.on("more", function(data) {
 
 	// restore scroll position
 	const position = chan.height() - heightOld;
-	scrollable.scrollTop(position);
+	scrollable.finish().scrollTop(position);
+
+	// We have to do this hack due to smooth scrolling in browsers,
+	// as scrollTop does not apply correctly
+	if (window.requestAnimationFrame) {
+		window.requestAnimationFrame(() => scrollable.scrollTop(position));
+	}
 
 	if (data.messages.length !== 100) {
 		scrollable.find(".show-more").removeClass("show");
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -12,7 +12,7 @@ socket.on("more", function(data) {
 
 	// get the scrollable wrapper around messages
 	const scrollable = chan.closest(".chat");
-	const heightOld = chan.height();
+	const heightOld = chan.height() - scrollable.scrollTop();
 
 	// If there are no more messages to show, just hide the button and do nothing else
 	if (!data.messages.length) {
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -12,6 +12,8 @@ const condensed = require("./condensed");
 const chat = $("#chat");
 const sidebar = $("#sidebar");
 
+require("intersection-observer");
+
 const historyObserver = window.IntersectionObserver ?
 	new window.IntersectionObserver(loadMoreHistory, {
 		root: chat.get(0)
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -12,6 +12,11 @@ const condensed = require("./condensed");
 const chat = $("#chat");
 const sidebar = $("#sidebar");
 
+const historyObserver = window.IntersectionObserver ?
+	new window.IntersectionObserver(loadMoreHistory, {
+		root: chat.get(0)
+	}) : null;
+
 module.exports = {
 	appendMessage,
 	buildChannelMessages,
@@ -145,6 +150,10 @@ function renderChannel(data) {
 	if (data.type === "channel") {
 		renderChannelUsers(data);
 	}
+
+	if (historyObserver) {
+		historyObserver.observe(chat.find("#chan-" + data.id + " .show-more").get(0));
+	}
 }
 
 function renderChannelMessages(data) {
@@ -219,4 +228,20 @@ function renderNetworks(data) {
 	if (sidebar.find(".highlight").length) {
 		utils.toggleNotificationMarkers(true);
 	}
+}
+
+function loadMoreHistory(entries) {
+	entries.forEach((entry) => {
+		if (!entry.isIntersecting) {
+			return;
+		}
+
+		var target = $(entry.target).find(".show-more-button");
+
+		if (target.attr("disabled")) {
+			return;
+		}
+
+		target.click();
+	});
 }
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "fuzzy": "0.1.3",
     "handlebars": "4.0.10",
     "handlebars-loader": "1.6.0",
+    "intersection-observer": "0.4.2",
     "jquery": "3.2.1",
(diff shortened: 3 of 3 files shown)


client/js/socket-events/more.js
```diff
@@ -14,7 +14,7 @@ socket.on("more", function(data) {
 
 	// get the scrollable wrapper around messages
 	const scrollable = chan.closest(".chat");
-	const heightOld = chan.height();
+	const heightOld = chan.height() - scrollable.scrollTop();
```
xPaw: This fixes scroll position not resetting correctly if we are not scrolled all the way up. This bug can be reproduced manually on master by scrolling a little so that load more button is clickable but not fully visible.

client/js/socket-events/more.js
```diff
@@ -31,7 +31,13 @@ socket.on("more", function(data) {
 
 	// restore scroll position
 	const position = chan.height() - heightOld;
-	scrollable.scrollTop(position);
+	scrollable.finish().scrollTop(position);
```
xPaw: `finish` is here for the same reason we have it when using page keys. As we have no clean way of updating final scroll position while jQuery is animating, we have to stop the animation.

MaxLeiter:
Mentioned in the description, but here's a direct link to browser support: 
https://caniuse.com/#feat=intersectionobserver

Safari (webkit) is currently working on it (whatever that means): https://webkit.org/status/?#specification-intersection-observer

bews:
> Mentioned in the description, but here's a direct link to browser support:
https://caniuse.com/#feat=intersectionobserver

Not supported in Firefox? Really? And you are still going to merge this?

xPaw:
> Not supported in Firefox? Really?

Firefox 55 ships in August which should have this enabled by default (currently behind a feature flag).

> And you are still going to merge this?

No one has merged anything yet.

xPaw:
`do not merge` until Firefox 55 is released.

MiniDigger:
Been running this for a few days now and it works great. I have some feedback tho. On phone loading new messages can take some time, it's not as smooth as on a desktop. Sadly, the ui doesn't display when it's loading messages. 
I think it would make sense to disable the load more button or just completely replace it with a spinner icon or something to show that it's loading messages. That could improve the UX since you can tell that it's doing something.

xPaw:
Firefox 55 shipped, and `IntersectionObserver` is now enabled by default there. [ref](https://developer.mozilla.org/en-US/Firefox/Releases/55#New_APIs)

astorije:
This is pretty cool and works really well.

I'm concerned about the stability of this feature/API though, it's not even a working draft yet, meaning it hasn't started going through all the hoops of W3C's process. Until it reaches Candidate Recommendation, lots of things can change! After that, we're safe since browsers implement it.

And to be clear, though I understand the argument of new API vs. alternatives, I am more concerned about changes to this API than non-support (because there is a polyfill), and I think it's a good idea to rely on browsers' capabilities for that. Also, this is a pretty non-critical feature, so it's fine to rely on this API, I think.

astorije:
> Technically we can use this approach to lazily load messages on initial page load, as the callback would be callback for empty channel when you switch to it.

It's kind of an abuse of the underlying motive for this API, no?

astorije:
Also, as a note, with this smoothly loading old messages, it now makes the "New messages" marker look even more out-of-place now. :man_shrugging:

xPaw:
> I definitely see a jump when I reach the top, probably due to more-button and/or markers.

Perhaps the animation frame fix I did for smooth scrolling is not enough on Mac devices, hmm.

> Agree with @MiniDigger, can we have a loading spin or something (or even nothing!)

On master we've made it change the button text to say it's loading, are you not seeing that?

> I understand there are ways to increase the bounding box?

It's the root margin setting and what not I believe, but I could not get it to work correctly last time I tried it.

astorije:
> Perhaps the animation frame fix I did for smooth scrolling is not enough on Mac devices, hmm.

Sorry, I should have specified: this was on mobile (Android).

> On master we've made it change the button text to say it's loading, are you not seeing that?

Actually, no :/
When reaching the top, nothing happens for a couple seconds, then new messages are loaded (and that's when the jump occurs).

astorije:
> It's the root margin setting and what not I believe, but I could not get it to work correctly last time I tried it.

Are you talking about that `boundingClientRect` thing? Maybe browsers don't entirely support that margin stuff? And maybe it would with the polyfill?
Well, that's unfortunate, but not a huge deal though.

xPaw:
> Are you talking about that boundingClientRect thing?

I'm talking about `rootMargin`.

astorije:
Happy to delay that to a further PR. Are you okay with adding the polyfill?

astorije:
I can confirm this PR now works on both Chrome and Safari 🎉 

However, I am noticing something odd. I generated a lot of `/motd`, refreshed the page (to only display the last 100 messages, as we don't seem to clean up messages on active channel when there are more than 100, but that's a `master` thing), then scrolled up and up so it loads previous messages: consistently, at the 6th load, it loads a 💩-ton of older messages (in fact, all of them).

So in practice: load page with 100 messages, scroll up, load another 100, scroll up, another 100, etc. and during the 6th load, the loading time is longer, and it loads several hundreds messages.
I can reproduce this consistently (as in 6th load every single time lol) on both Chrome and Safari.
The weirdest thing.

xPaw:
@astorije e01e21cdbe237c2ed58b9175587c8f5b6efd2252 this breaks perhaps?

astorije:
I honestly have no idea. Are you able to reproduce on your end?

astorije:
Well, rebased on `master` and it appears that I cannot reproduce after that, so I guess it's fixed... We'll find out soon enough :)

xPaw:
@astorije I would rather wait for a fix.

astorije:
Okay then, hope they'll fix it quickly because I'd love love love to merge this :)

astorije:
@xPaw:

- https://github.com/w3c/IntersectionObserver/issues/234#issuecomment-325499944
- https://github.com/w3c/IntersectionObserver/pull/239

🚢 :shipit: 🚀 ❗️❕❗️

astorije:
Rebased on `master` to fix conflict and updated the polyfill. Diff at https://github.com/w3c/IntersectionObserver/compare/510fa96d67b110e7017029c7392a84f7181f8b15...b7d62353e9a1128dd66d0eeeb532278e2ae38488.

astorije:
:tada:

PR #1319, opened 2017-07-10
Update README for when to run `npm run build`

MaxLeiter:


README.md
```diff
@@ -61,10 +61,11 @@ your own risk. It is also not recommended to run this as root.
 ## Development setup
 
 Simply follow the instructions to run The Lounge from source above, on your own
-fork.
+fork. Be sure to check out `package.json` for available npm scripts that may help or be necessary for your changes.
```
astorije: I don't think this is really necessary. Any specific use case why you added it?

README.md
```diff
@@ -61,10 +61,11 @@ your own risk. It is also not recommended to run this as root.
...
 
 Before submitting any change, make sure to:
 
 - Read the [Contributing instructions](https://github.com/thelounge/lounge/blob/master/CONTRIBUTING.md#contributing)
 - Run `npm test` to execute linters and test suite
-- Run `npm run build` if you change or add anything in `client/js/libs` or `client/views`
+- Run `npm run build` if you change or add anything in `client/js` or `client/views`
```
astorije: Good catch!

README.md
```diff
@@ -61,10 +61,11 @@ your own risk. It is also not recommended to run this as root.
...
+- Run `npm run start-dev` to start an instance that will watch for changes and rebuild automatically
```
astorije: Well, to be fair, this is in the list for "Before submitting any change, make sure to:", so I either don't think we need it, or that this is the right section to add this.

Maybe this section should only contain reading the CONTRIBUTING file + running tests, and then another section (before or after) to help development where you mention `npm run build`, `npm run watch`, `npm run start-dev` and `npm run start` would be more appropriate.

MaxLeiter:
Simplified a lot, now only fixes the mistake.

PR #1130, opened 2017-05-03
Better previews

bews:
# New design

- Pictures now have max-height in vh, so they are bigger/smaller depending on user screen size
- Site description is multiline again
- ~~Site previews now take all width (see #1084)~~
- Site thumbnail is ~2 times smaller than picture preview

# New features
1. When preview is shown, toggle button is hidden now (only if auto-expand is `on`):

It's still possible to hide it again using right click:

-------------------------------

2. Now it's possible to expand images from previews:

This works both for pictures and site thumbnails. It's really awesome, believe me!

# Bugfixes
1. Finally fixed these options:

^^ content in backlog is now affected too.

--------------------------------
2. Extra space is removed (mobile only):

--------------------------------
# TODO

- [x] Revert site previews width
- [x] Bind "back" button to expanded previews (mobile)
- [x] Keep toggle button if auto-expand options are disabled

PR #1325, opened 2017-07-12
An image viewer popup for thumbnails and image previews

astorije:
Inspired from #1130 for the styling, with some improved logic like support for <kbd>Escape</kbd> key.

screen shot 2017-07-12 at 03 00 33

### TODOs

- [x] Split thumbnail and content into 2 adjacent links to make sure clicking on thumbnail margin opens the viewer (will also make code simpler)
  NB: Not sure why, but when I switched to this it fixed the image preview shrinking on `master` :tada:
- [x] Try to fix link going beyond image borders when resizing
- [x] Check / Make GPU taking care of animation

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -337,6 +337,9 @@ kbd {
 #wrap {
 	height: 100%;
 	overflow: hidden;
+
+	/* Needed to transition blurred background when image viewer is open */
+	transition: filter .2s;
 }
 
 #viewport {
@@ -1147,7 +1150,6 @@ kbd {
 	color: #222;
 	font-size: 12px;
 	max-width: 100%;
-	padding: 6px;
 	margin: 2px 0;
 	overflow: hidden;
 }
@@ -1159,22 +1161,32 @@ kbd {
 }
 
 #chat .toggle-content .thumb {
-	margin-right: 6px;
 	max-width: 48px;
 	max-height: 32px;
 }
 
+#chat .toggle-thumbnail {
+	padding: 6px;
+}
+
 #chat .toggle-text {
+	padding: 6px;
 	min-width: 0;
 	display: flex;
 	flex-direction: column;
 	white-space: nowrap;
+	color: inherit;
+}
+
+#chat .toggle-text:not(:first-child) {
+	padding-left: 0;
 }
 
 #chat .toggle-content .head,
 #chat .toggle-content .body {
 	text-overflow: ellipsis;
 	overflow: hidden;
+	color: inherit;
 }
 
 #chat .toggle-content .head {
@@ -1997,4 +2009,73 @@ kbd {
 
 ::-webkit-scrollbar-thumb:vertical:active {
 	background: rgba(0, 0, 0, .6);
+}
+
+/* Image viewer */
+
+#image-viewer,
+#image-viewer .close-btn {
+	/* Vertically and horizontally center stuff */
+	display: flex;
+	flex-direction: column;
+	align-items: center;
+	justify-content: center;
+}
+
+#image-viewer {
+	position: fixed;
+	top: 0;
+	bottom: 0;
+	left: 0;
+	right: 0;
+	background: rgba(0, 0, 0, .8);
+	visibility: hidden;
+	opacity: 0;
+	transition: opacity .2s, visibility .2s;
+	z-index: 999;
+}
+
+.image-viewer-opened #wrap {
+	filter: blur(4px);
+}
+
+.image-viewer-opened #image-viewer {
+	visibility: visible;
+	opacity: 1;
+}
+
+#image-viewer .close-btn {
+	position: fixed;
+	top: 0;
+	right: 0;
+	width: 2em;
+	height: 2em;
+	font-size: 36px;
+	color: white;
+	z-index: 1001;
+	opacity: .6;
+	transition: .2s opacity;
+}
+
+#image-viewer .close-btn:hover {
+	opacity: 1;
+}
+
+#image-viewer .image-link {
+	margin: 10px;
+}
+
+#image-viewer .image-link:hover {
+	opacity: 1;
+}
+
+#image-viewer .image-link img {
+	max-width: 100%;
+
+	/* Top/Bottom margins + button height + image/button margin */
+	max-height: calc(100vh - 2 * 10px - 37px - 10px);
+}
+
+#image-viewer .open-btn {
+	margin: 0 auto 10px;
 }
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -137,12 +137,14 @@ kbd {
 	text-transform: uppercase;
 	transition: background .2s, border-color .2s, color .2s;
 	word-spacing: 3px;
+	cursor: pointer; /* This is useful for `<button>` elements */
 }
 
 .btn:disabled,
 .btn:hover {
 	background: #84ce88;
 	color: #fff;
+	opacity: 1;
 }
 
 .btn:active {
--- a/client/js/renderPreview.js
+++ b/client/js/renderPreview.js
@@ -61,3 +61,47 @@ function handleImageInPreview(content, container) {
 		});
 	}
 }
+
+/* Image viewer */
+
+// FIXME Remove #input focus when this is open
+// See https://github.com/thelounge/lounge/issues/1342
+$("#viewport").on("click", ".toggle-thumbnail", function() {
+	const link = $(this);
+
+	openImageViewer(link);
+
+	// Prevent the link to open a new page since we're opening the image viewer,
+	// but keep it a link to allow for Ctrl/Cmd+click
+	return false;
+});
+
+$("#image-viewer").on("click", function() {
+	closeImageViewer();
+});
+
+$(document).keydown(function(e) {
+	switch (e.keyCode ? e.keyCode : e.which) {
+	case 27: // Escape
+		closeImageViewer();
+		break;
+	}
+});
+
+function openImageViewer(link) {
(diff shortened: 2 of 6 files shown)


astorije:
The flexbox thing behind link/image on the viewer is kinda 💩. Play with resize, or wide images, and the button won't stick to the image anymore, and the link will be on the image + on big top/bottom margins.
Probably not the right way to go, but I have made an unreasonable amount of attempts, so I'd love some help :D (or maybe it's good enough for this PR and can be improved later, I'd vote for that!).

Reason why 2 links btw is because I don't want the black areas around the button to be clickable too. Only image *and* button are/should be clickable.

MaxLeiter:
Personally not sure how I feel about the open image button -- I think most people can figure out save as or drag to desktop. Overall great feature.

MiniDigger:
👍 on this one, I am currently using the one from bews PR and it works great.  

what limits the size of the image? can't we make it bigger? I also don't like how the button looks, but I can just change that via css. I definitely am in favor of having such a button since you might want to get to the actual site where that image is coming from (eg. viewing youtube thumbnail and then watching the video)

client/js/renderPreview.js
```diff
@@ -61,3 +61,56 @@ function handleImageInPreview(content, container) {
...
+	openImageViewer(self, link);
+
+	// Prevent the link to open a new page since we're opening the image viewer,
+	// but keep it a link to allow for Ctrl/Cmd+click
+	event.preventDefault();
+});
+
+// Reason to make this a specific handler is, for images, to open the image
+// viewer even when clicking in the margins of the thumbnail
+$("#viewport").on("click", ".toggle-content.toggle-type-image", function() {
```
xPaw: Have you tried `stopPropagation` in this event? It might be possible to get rid of the event above.

client/js/renderPreview.js
```diff
@@ -61,3 +61,56 @@ function handleImageInPreview(content, container) {
...
+	const self = $(this);
+	const image = self.find("img");
+
+	openImageViewer(image, self);
+
+	return false;
+});
+
+$(document).on("click", "#image-viewer", function() {
```
xPaw: You can cache `#viewport` into a variable and bind all these events on it instead of document.

client/js/renderPreview.js
```diff
@@ -61,3 +61,56 @@ function handleImageInPreview(content, container) {
...
+	closeImageViewer();
+});
+
+$(document).keydown(function(e) {
+	switch (e.keyCode ? e.keyCode : e.which) {
+	case 27: // Escape
```
xPaw: We also do this dropdown menus, might be a good idea to combine or something.

client/js/renderPreview.js
xPaw: Okay that won't work as you want the thumbnail too.

YaManicKill:
> Personally not sure how I feel about the open image button -- I think most people can figure out save as or drag to desktop. Overall great feature.

I disagree, let's not make assumptions on that sort of level, not everyone knows these things. I would however suggest that maybe instead of this button, we could have a button that has [this icon](http://fontawesome.io/icon/external-link/) which opens it in a new tab. But that's just a small thing.

I'll do a proper review of this soon, looks good in general, @astorije

xPaw:
I don't think using blur (especially with transition) is worth the performance hit considering how dark the container already is.

client/js/renderPreview.js
astorije: I'll rewrite that part to make it simpler and remove an edge case (accidentally clicking on margin of image in link previews, not image previews, will open the link instead of the viewer).

client/js/renderPreview.js
astorije: > You can cache #viewport into a variable

Will do!

> bind all these events on it instead of document.

Any benefits? By not binding on `body`, it means your focus can be off and not catch the event, no?

client/js/renderPreview.js
astorije: I don't really want to get into keyboard handling as part of this PR. Can we do it later?
(Also, I'm not 100% sure we should, hesitating a bit, let's postpone this I'd say)

MaxLeiter:
A lot of the time instead of using `blur` people create a div with a slight alpha like here:

https://codepen.io/imprakash/pen/GgNMXO

Could that be a better solution, @xPaw?

astorije:
> Personally not sure how I feel about the open image button -- I think most people can figure out save as or drag to desktop.

Additionally to what @YaManicKill said, it's less trivial on mobile to open context menu than to click on a button, and while it's convenient it's not obvious that image is clickable.
For the record, Google Images shows this kind of button too.
Also, when #1307 is merged, button/link will open the original image instead of the proxied picture, which is usually what you want to do when saving or zooming (at this point we won't be compressing/resizing the image, but user does not know that).

> what limits the size of the image? can't we make it bigger?

Image is: fluid until 100% of the image width max. So if image is 1000px width and your screen is 2000px wide, image will take half of it. If your screen is 500px wide, image will be reduced to fit 100% of the screen.
Does that make sense? I feel like it doesn't :D
To see it in full-size, you can click the image or the button.

> For some reason `object-fit: contain` makes the link hitbox square, so hitting empty space still opens the link (try https://i.redd.it/bk48m38ioz8z.jpg)

Yeah, that's the issue I was mentioning above. I think I found a way to fix this, I'll try this and see what happens.

> I don't think using blur (especially with transition) is worth the performance hit considering how dark the container already is.

You don't like nice things :D
I honestly think there is no issue with this at all. I will try with the slowest devices of our generation and our parent's (i.e. current phones in my household) but I doubt this will have any noticeable impact. Especially since there is no scroll or else on the blurred content.
If it's an issue, I will use the `transform: translate3d(0,0,0);` or `transform: translateZ(0);` trick, but I don't even think we'll need this.
And yeah, of course it's worth it, it's pretty! :)

astorije:
> A lot of the time instead of using blur people create a div with a slight alpha like here:

If I understand correctly, this is doing both. It looks pretty slick as is!

astorije:
I fixed both issues mentioned in the PR description so this should be good for review. The `max-height` for image is pretty disgusting but for the life of me I couldn't find anything else that works (`object-fit: contain` is out of the picture due to how it works).
Any suggestions? Or you're cool with the current solution?

@xPaw, I couldn't find where to check that GPU is being leverage, could you advise?

client/js/renderPreview.js
```diff
@@ -61,3 +61,44 @@ function handleImageInPreview(content, container) {
 		});
 	}
 }
+
+/* Image viewer */
+
+// FIXME Remove #input focus when this is open
```
YaManicKill: If you aren't going to fix this in this PR (which I'm ok with) then stick up an issue on github for it and put in the number here. Otherwise this will get missed. I hate random "FIXME" and "TODO" s that don't have any number attached to them

client/css/style.css
```diff
@@ -1996,3 +2010,73 @@ kbd {
...
+
+#image-viewer .image-link {
+	margin: 10px;
+}
+
+#image-viewer .image-link:hover {
+	opacity: 1;
+}
+
+#image-viewer .image-link img {
```
xPaw: Wouldn't actual padding on here work instead of using calc?

client/js/renderPreview.js
```diff
@@ -61,3 +61,44 @@ function handleImageInPreview(content, container) {
...
+	const link = $(this);
+
+	openImageViewer(link);
+
+	// Prevent the link to open a new page since we're opening the image viewer,
+	// but keep it a link to allow for Ctrl/Cmd+click
+	return false;
+});
+
+$(document).on("click", "#image-viewer", function() {
```
xPaw: Pretty sure the way you're binding the event here does not help and `$("#image-viewer").on("click")` should work exactly the same.

client/js/renderPreview.js
```diff
@@ -61,3 +61,44 @@ function handleImageInPreview(content, container) {
...
+function openImageViewer(link) {
+	const container = $("#image-viewer");
+
+	container.html(templates.image_viewer({
+		image: link.find("img").attr("src"),
+		link: link.attr("href"),
+		type: link.parent().hasClass("toggle-type-image") ? "image" : "link"
+	}));
+
+	$("body").addClass("image-viewer-opened");
```
xPaw: Use `$(document.body)`

client/js/renderPreview.js
```diff
@@ -61,3 +61,44 @@ function handleImageInPreview(content, container) {
...
+}
+
+function closeImageViewer() {
```
xPaw: Clearing out `image-viewer` in here might be a good idea to unload it from memory.

client/js/renderPreview.js
astorije: Good idea, here it is: https://github.com/thelounge/lounge/issues/1342

client/js/renderPreview.js
astorije: Thanks, done.

client/js/renderPreview.js
astorije: Good call, done.

client/js/renderPreview.js
astorije: Agreed, done (after transition has ended).

client/css/style.css
astorije: Actually, left/right padding calc was unnecessary as the link itself has a `margin`, but for the max height, padding doesn't help. What this means is that the image's max height must be all screen minus its margins and the button height, and I can't get to anything else that would work and be less "hardcoded"... :/

astorije:
I believe I have addressed all comments.

PR #1326, opened 2017-07-12
Update screenshot in README

MaxLeiter:
lounge-3

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ The Lounge is a modern web IRC client designed for self-hosting.
 To learn more about configuration, usage and features of The Lounge, take a look at [the website](https://thelounge.github.io).
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


MiniDigger:
any special reason? 
while its cool that you show the embed you are also missing some stuff the other screenshot shows, like the alter in the secret server, the new messages bar, the red favicon etc.

MaxLeiter:
I don't think the red favicon and altered message bubble color are important to have here because they aren't very explicit in their meaning without using the client. Should have new message bar though.

YaManicKill:
My only question: should we make it a 16:9 screenshot, rather than this (which is 16:6) as that's more likely to be what people will be viewing it at. This one seems a little squished on the y axis.

MaxLeiter:
Thoughts on this as a new thumbnail for 2.5?

image

astorije:
@MaxLeiter, :shipit:

PR #1327, opened 2017-07-13
Changelog viewer and updater checker in the client

### LINKED ISSUE #619: Present change logs to users in the client
Since our change logs are very informative (:clap: @astorije) I think users would appreciate reading a curated list of changes should their instance be upgraded. This could be presented through a modal once a user authenticates after a server upgrade. This'll also help make new features like #400 less hidden.

This could be tracked by keeping something like this in the user configs;

``` json
{
    ...
    "changelogs": {
        "enabled": true,
        "lastRead": "2.0.0-rc.1"
    }
}
```

xPaw:
Grabs the version and changelog info from GitHub API. Fixes #619

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -85,6 +85,7 @@
 				<div id="connect" class="window"></div>
 				<div id="settings" class="window" data-type="settings"></div>
 				<div id="help" class="window"></div>
+				<div id="changelog" class="window"></div>
 			</div>
 		</div>
 	</div>
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -332,64 +332,42 @@ $(function() {
 		$(this).closest(".msg.condensed").toggleClass("closed");
 	});
 
-	sidebar.on("click", ".chan, button", function(e, data) {
-		// Pushes states to history web API when clicking elements with a data-target attribute.
-		// States are very trivial and only contain a single `clickTarget` property which
-		// contains a CSS selector that targets elements which takes the user to a different view
-		// when clicked. The `popstate` event listener will trigger synthetic click events using that
-		// selector and thus take the user to a different view/state.
-		if (data && data.pushState === false) {
-			return;
-		}
-		const self = $(this);
-		const target = self.data("target");
-		if (!target) {
-			return;
-		}
-		const state = {};
+	let changelogRequestedAt = 0;
 
-		if (self.hasClass("chan")) {
-			state.clickTarget = `#sidebar .chan[data-id="${self.data("id")}"]`;
-		} else {
-			state.clickTarget = `#footer button[data-target="${target}"]`;
-		}
-
-		if (history && history.pushState) {
-			if (data && data.replaceHistory && history.replaceState) {
-				history.replaceState(state, null, target);
-			} else {
-				history.pushState(state, null, target);
-			}
-		}
-	});
-
-	sidebar.on("click", ".chan, button", function() {
+	const openWindow = function openWindow(e, data) {
 		var self = $(this);
 		var target = self.data("target");
 		if (!target) {
 			return;
 		}
 
-		chat.data(
-			"id",
-			self.data("id")
-		);
-		socket.emit(
-			"open",
-			self.data("id")
-		);
+		// This is a rather gross hack to account for sources that are in the
+		// sidebar specifically. Needs to be done better when window management gets
+		// refactored.
+		const inSidebar = self.parents("#sidebar").length > 0;
 
-		sidebar.find(".active").removeClass("active");
-		self.addClass("active")
-			.find(".badge")
-			.removeClass("highlight")
-			.empty();
+		if (inSidebar) {
+			chat.data(
+				"id",
+				self.data("id")
+			);
+			socket.emit(
+				"open",
+				self.data("id")
+			);
 
-		if (sidebar.find(".highlight").length === 0) {
-			utils.toggleNotificationMarkers(false);
+			sidebar.find(".active").removeClass("active");
+			self.addClass("active")
+				.find(".badge")
+				.removeClass("highlight")
+				.empty();
+
+			if (sidebar.find(".highlight").length === 0) {
+				utils.toggleNotificationMarkers(false);
+			}
+
+			sidebarSlide.toggle(false);
 		}
-
-		sidebarSlide.toggle(false);
 
 		var lastActive = $("#windows > .active");
 
@@ -447,8 +425,49 @@ $(function() {
 			socket.emit("sessions:get");
 		}
 
+		if (target === "#help" || target === "#changelog") {
+			const now = Date.now();
+			// Don't check more than once a day
+			if (now - changelogRequestedAt > 86400 * 1000) {
+				changelogRequestedAt = now;
+				socket.emit("changelog");
+			}
+		}
+
 		focus();
-	});
+
+		// Pushes states to history web API when clicking elements with a data-target attribute.
+		// States are very trivial and only contain a single `clickTarget` property which
+		// contains a CSS selector that targets elements which takes the user to a different view
+		// when clicked. The `popstate` event listener will trigger synthetic click events using that
+		// selector and thus take the user to a different view/state.
+		if (data && data.pushState === false) {
+			return;
+		}
+		const state = {};
+
+		if (self.attr("id")) {
+			state.clickTarget = `#${self.attr("id")}`;
+		} else if (self.hasClass("chan")) {
+			state.clickTarget = `#sidebar .chan[data-id="${self.data("id")}"]`;
+		} else {
+			state.clickTarget = `#footer button[data-target="${target}"]`;
+		}
+
+		if (history && history.pushState) {
+			if (data && data.replaceHistory && history.replaceState) {
+				history.replaceState(state, null, target);
+			} else {
+				history.pushState(state, null, target);
+			}
+		}
+
+		return false;
+	};
+
+	sidebar.on("click", ".chan, button", openWindow);
+	$("#help").on("click", "#view-changelog, #back-to-help", openWindow);
+	$("#changelog").on("click", "#back-to-help", openWindow);
 
 	sidebar.on("click", "#sign-out", function() {
 		socket.emit("sign-out");
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -427,8 +427,8 @@ $(function() {
 
 		if (target === "#help" || target === "#changelog") {
 			const now = Date.now();
-			// Don't check more than once a day
-			if (now - changelogRequestedAt > 86400 * 1000) {
+			// Don't check more than once an hour
+			if (now - changelogRequestedAt > 3600 * 1000) {
 				changelogRequestedAt = now;
 				socket.emit("changelog");
 			}
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -14,6 +14,7 @@ socket.on("configuration", function(data) {
 	$("#settings").html(templates.windows.settings(data));
 	$("#connect").html(templates.windows.connect(data));
 	$("#help").html(templates.windows.help(data));
+	$("#changelog").html(templates.windows.changelog());
 
 	$("#play").on("click", () => {
 		const pop = new Audio();
@@ -65,4 +66,22 @@ socket.on("configuration", function(data) {
 			// Store the "previous" value, for next time
 			$(this).data("lastvalue", nick);
 		});
+
+	$("#view-changelog").on("click", function() {
+		$("#windows > .active")
+			.removeClass("active")
+			.find(".chat")
+			.unsticky();
+
+		$("#changelog")
+			.addClass("active");
+
+		history.pushState({
+			clickTarget: "#view-changelog",
+		}, null, "#changelog");
+
+		return false;
+	}).one("click", function() {
+		socket.emit("changelog");
+	});
 });
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -66,22 +66,4 @@ socket.on("configuration", function(data) {
 			// Store the "previous" value, for next time
 			$(this).data("lastvalue", nick);
 		});
-
-	$("#view-changelog").on("click", function() {
-		$("#windows > .active")
-			.removeClass("active")
-			.find(".chat")
-			.unsticky();
-
-		$("#changelog")
-			.addClass("active");
-
-		history.pushState({
-			clickTarget: "#view-changelog",
-		}, null, "#changelog");
-
-		return false;
-	}).one("click", function() {
-		socket.emit("changelog");
-	});
 });
--- a/client/views/windows/help.tpl
+++ b/client/views/windows/help.tpl
@@ -4,6 +4,28 @@
 <div class="container">
 	<h1 class="title">Help</h1>
 
+	<h2>About The Lounge</h2>
+
+	<div class="about">
+		{{#if gitCommit}}
+			<p>The Lounge is running from source (commit <a href="https://github.com/thelounge/lounge/tree/{{gitCommit}}" target="_blank" rel="noopener"><code>{{gitCommit}}</code></a>) based on <strong>{{version}}</strong></p>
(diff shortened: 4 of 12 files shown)


YaManicKill:
As discussed on IRC, I think we should grab the changelog from the local CHANGELOG.md for multiple reasons:

1. If github is down, we can still get the changelog (although not the latest version, but that's fine)
2. I feel like there's a security issue with grabbing rendered html from a 3rd party site and just dumping it onto the page. You say that any JS on that won't get run, but I would like to test this

I understand the reasoning for getting it from github, and I know you really don't regex, but it's not the spawn of the Devil, and we do control the CHANGELOG.md file, so...

If other @thelounge/maintainers support this over my suggestion, I won't argue it too much. But that's my thought process.

However, in general I love this feature. Will do a proper review of it later.

client/views/changelog.tpl
```diff
@@ -0,0 +1,27 @@
+{{#if latest}}
+	<h3>New version available!</h3>
+	<p>
+		There is a new {{#if latest.prerelease}}pre-{{/if}}release <b>{{latest.version}}</b> of The Lounge available.
+		<a href="{{latest.url}}" target="_blank" rel="noopener">View details on GitHub</a>
+	</p>
+	<p>We strongly suggest updating The Lounge to the latest version.</p>
```
MaxLeiter: I would remove "The Lounge".

`<p>We strongly suggest updating to the latest version.</p>`

client/views/changelog.tpl
```diff
@@ -0,0 +1,27 @@
+{{#if latest}}
+	<h3>New version available!</h3>
+	<p>
+		There is a new {{#if latest.prerelease}}pre-{{/if}}release <b>{{latest.version}}</b> of The Lounge available.
```
MaxLeiter: "You is a newer version of The Lounge available. You can <a>view the details on GitHub<a/>"

client/views/changelog.tpl
```diff
@@ -0,0 +1,27 @@
...
+		<a href="{{latest.url}}" target="_blank" rel="noopener">View details on GitHub</a>
+	</p>
+	<p>We strongly suggest updating The Lounge to the latest version.</p>
+	<hr>
+{{/if}}
+
+{{#if gitCommit}}
+	<p>The Lounge is running from source (commit <a href="https://github.com/thelounge/lounge/tree/{{gitCommit}}" target="_blank" rel="noopener"><code>{{gitCommit}}</code></a>) based on <strong>{{current.version}}</strong></p>
+	<p><a href="https://github.com/thelounge/lounge/compare/{{gitCommit}}...master" target="_blank" rel="noopener">View changes from <code>{{gitCommit}}</code> to <code>master</code> on GitHub</a> to see the changes you are missing</p>
```
MaxLeiter: "Compare changes between <gitCommit>, <version>, and <master> to see what you're missing"

client/views/changelog.tpl
```diff
@@ -0,0 +1,27 @@
...
+	<p><a href="https://github.com/thelounge/lounge/compare/{{current.version}}...{{gitCommit}}" target="_blank" rel="noopener">View changes from <code>{{current.version}}</code> to <code>{{gitCommit}}</code> to master on GitHub</a> to see the changes made</p>
```
MaxLeiter: Same as above minus commit

YaManicKill:
What's the current state of this, @xPaw ?

xPaw:
@YaManicKill waiting for your reviews.

astorije:
Didn't you mean to put this under "Help"? It's rather weird right now:

screen shot 2017-11-22 at 17 40 47

astorije:
So, current thing is like this:

screen shot 2017-11-22 at 17 43 43

I would much prefer to keep the version number directly.
Also, the version check would be much nicer in the help page. In general, the extra page (or this would be very appropriate for a modal actually) should just be the release notes IMO.

astorije:
These appear to be the same heading level:

screen shot 2017-11-22 at 17 46 33

This is because the "Added"/etc. stuff are `h2`s. If this page only contains release notes, then that won't be an issue anymore, because the top title will be about the release, it won't have to be an `h2`.

xPaw:
@astorije Both version check and changelog do the same request to GitHub's API, would be rather weird to split them.

astorije:
Right, I get why technically it's more annoying, but the UI would be much nicer. And it's odd to have to check for new versions to know the current version :s
Want me to do some mockups/changes (in extra commits) to see what I mean?

xPaw:
@astorije Tweaked it.

client/js/socket-events/configuration.js
```diff
@@ -65,4 +65,18 @@ socket.on("configuration", function(data) {
 			// Store the "previous" value, for next time
 			$(this).data("lastvalue", nick);
 		});
+
+	$("#view-changelog").on("click", function() {
```
YaManicKill: Not sure this is the right place for this code, it's not really related to the configuration.

client/index.html
```diff
@@ -83,6 +83,12 @@ <h1 class="title">The Lounge is loading…</h1>
 				<div id="connect" class="window"></div>
 				<div id="settings" class="window" data-type="settings"></div>
 				<div id="help" class="window"></div>
+				<div id="changelog" class="window">
+					<div class="header">
+						<button class="lt" aria-label="Toggle channel list"></button>
```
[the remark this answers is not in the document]
xPaw: Shout way of doing things 🤷‍♂️

All the windows have that button for mobile.

client/js/socket-events/configuration.js
xPaw: This event is used to render windows like help, settings, etc. I just need to make changelog a view too.

YaManicKill: Hmmmm, I'm just not sure why this needs to happen here. The other ones exist here because they are waiting on the configuration call. This one doesn't need it.

client/index.html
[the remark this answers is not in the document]
YaManicKill: Yay, legacy.

client/js/socket-events/configuration.js
xPaw: Actually `#view-changelog` is in the help window, hence why it needs to happen here.

YaManicKill: Oh yeah, you are right.

client/views/windows/help.tpl
```diff
@@ -4,6 +4,26 @@
...
+			<p>The Lounge is running <a href="https://github.com/thelounge/lounge/releases/tag/{{version}}" target="_blank" rel="noopener"><strong>{{version}}</strong></a></p>
+		{{/if}}
+
+	{{#unless public}}
+		<button class="btn" id="view-changelog">View changelog &amp; check for updates</button><br>
+	{{/unless}}
+
+		<a href="https://thelounge.github.io/" target="_blank" rel="noopener">Website</a><br>
+		<a href="https://thelounge.github.io/docs/" target="_blank" rel="noopener">Documentation</a><br>
+		<a href="https://github.com/thelounge/lounge/issues/new" target="_blank" rel="noopener">Report a bug</a>
```
astorije: It looks like those 3 items do not pick the right font-size.

client/views/changelog.tpl
```diff
@@ -0,0 +1,20 @@
...
+	The Lounge <b>{{latest.version}}</b>{{#if latest.prerelease}} (pre-release){{/if}} is now available.
+	Click to view details on GitHub.
+</a>
+{{else if current.changelog}}
+<div class="changelog-version">
+	The Lounge is up to date!
+</div>
+{{/if}}
+
+<h1 class="title">Changelog for {{current.version}}</h1>
```
astorije: I don't think there is a way to go back to the Help window yet:

- The back/forward button does not work
- What about adding a "Back to the Help page" button here?

client/views/windows/help.tpl
```diff
@@ -4,6 +4,26 @@
...
 
+	<h2>About The Lounge</h2>
+
+	<p class="about">
+		{{#if gitCommit}}
+			<p>The Lounge is running from source (commit <a href="https://github.com/thelounge/lounge/tree/{{gitCommit}}" target="_blank" rel="noopener"><code>{{gitCommit}}</code></a>) based on <strong>{{version}}</strong></p>
+			<p><a href="https://github.com/thelounge/lounge/compare/{{gitCommit}}...master" target="_blank" rel="noopener">Compare changes between <code>{{gitCommit}}</code> and <code>master</code> to see what you are missing</a></p>
+			<p><a href="https://github.com/thelounge/lounge/compare/{{version}}...{{gitCommit}}" target="_blank" rel="noopener">Compare changes between <code>{{version}}</code> and <code>{{gitCommit}}</code> to see the changes made</a></p>
+		{{else}}
+			<p>The Lounge is running <a href="https://github.com/thelounge/lounge/releases/tag/{{version}}" target="_blank" rel="noopener"><strong>{{version}}</strong></a></p>
```
astorije: I think I prefer the current:

screen shot 2017-12-14 at 02 23 24

over:

screen shot 2017-12-14 at 02 23 33

The call-to-action link seems more explicit than the link on the version.

dgw: Agreed re the placement of the link to release notes, but not re the wording of the version string. "The Lounge is in version" sounds really, really strange, and I think it would be better phrased "The Lounge is at version" or "The Lounge v2.x.x".

astorije: My English is really, really strange, can't argue with that 😁 

"Version 2.x.x" works too. Could also do "Version 2.x.x (stable)" or "Version 2.x.x-pre.n (pre-release)" but that adds a step.

There are alternatives, like putting the version number in the title line (with a `<small>` or aligned to the right or something) and then the release notes link can be with the list of other links.
I don't have a strong opinion and don't want to bikeshed, so :man_shrugging:

client/views/changelog.tpl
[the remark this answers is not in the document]
xPaw: Changed.

client/views/windows/help.tpl
xPaw: I tweaked it a bit:

client/views/changelog.tpl
xPaw: I made the back/forward buttons sort of work. You can go back, forward, but next back won't work, don't know why.

client/views/windows/help.tpl
astorije: lol you left as-is what we commented about and changed what we didn't comment about :D 

I would definitely leave the links as is instead of button styles.
Common UX (I know I'm gonna get accused to pull UX rules out of my 🍑 again lol, but there are a bunch of good reads on the subject) recommends to use buttons for actions, and links for routing/navigating. All 3 are navigations here (all outside the app), even the last one, which doesn't do the action of reporting a bug, but send the user to a place where to report a bug.
Of course, it's UX / design so it's not black & white, we'll always find counter-examples, but I think in this case it's particularly relevant.

The first button is fine as a button because of the "Check for updates" action.

client/views/changelog.tpl
astorije: Uh, that's odd.
Side-node: There is no... back button on iOS... 😩
I'm glad we have a &times; button on the image preview.

client/views/windows/help.tpl
xPaw: Earlier you were showing to make "See release notes" a link after the version.

So I don't know what you want anymore.

astorije: > Earlier you were showing to make "See release notes" a link after the version.
> So I don't know what you want anymore.

I'm not sure where the contradiction is. "See release notes" makes sense as a link, but since you're bundling it with checking for updates as "View release notes & check for updates", it's fine as a button.

I see you have made more tweaks, let me take a look.

astorije:
@xPaw, do you want me to take a stab at UI? I can do it in an extra commit so it doesn't touch what you did in case you don't like.

xPaw:
@astorije Feel free to.

astorije:
I simply rebased on master to fix conflicts, haven't modified the commit one bit. I'll add stuff onto this PR and see how it goes.

astorije:
Alriiight, I have spent soooooo much time on this, that was unreasonable. But I think it looks really dope right now. Impressive how our window management is terrible and cries for a refactor lol. No wonder why we often have history handling issues!
@xPaw, I haven't touched your commit, added all my stuff on top of it:

- Support `pre` blocks in our style. We use this in our changelogs. Style for this is straight from Bootstrap 3 minus the border.
- Keep consistent width between the Help page and Changelog (which is already different from other windows <U+1F620>)
- Add icons to the About links
- Make sure `li` elements (i.e. all the lists in changelogs) are consistent in size with rest of the client
- Display version and release notes link on the "About The Lounge" header line, smaller, pushed to the right
- Check new releases when opening the Help window in order to display it without having to open the release notes. Release notes are being fed to the Changelog page at that moment to avoid fetching twice.
- Re-check version/fetch release notes after 24h. Since The Lounge can now run 24/7, reconnect when losing the network, we have to assume an "always-on" usage.
- Change icon, animate background color when getting response from GitHub to avoid flashing.
- Combine click handlers with our wonderful window management. These were the same handler, even with similar checks (`target` exists, etc.), just in 2 different places. This is necessary for the next item.
- Combine "Open release notes" and "Go back to Help" button behaviors with window management handlers. The window management code is gross as ever, and is in desperate need of a refactor, but at least there is no duplicated code for the same behavior + history management. This fixes the "Next" history behavior (however reloading the app while viewing the notes does not load on the notes, but this is a bug for a different PR!).
- Added a rule in the history management thingy: if a link we want to add history handling to has an `id`, store that in the state
- Added a button to go back to the Help window
- Fixed links to releases
- Send user to the GitHub issues *list* instead of *new issue form* because if they do not have a GitHub account, they will be redirected to the login page, which is a rather unpleasant experience when you are already confused...
- Fixed a bug that would return data about a new release in `latest` even though it is already the `current`. It was showing the current version as "The Lounge v... is now available".
- Added https://user-images.githubusercontent.com to the CSP rule when prefetch storage is enabled, because that is where we have stored screenshots in the changelog so far. Meh (we can improve that later if we decide to have a dedicated place for screenshots).
- Fetch changelog info even in public mode because users in public mode can access the release notes. They do not see the result of the version checker however.

And for the results:

screen shot 2017-12-22 at 21 56 13 1
screen shot 2017-12-22 at 21 56 31
screen shot 2017-12-22 at 21 57 01
screen shot 2017-12-22 at 21 57 44
screen shot 2017-12-22 at 21 58 46
screen shot 2017-12-22 at 21 57 15
screen shot 2017-12-22 at 21 57 52

client/js/lounge.js
```diff
@@ -447,8 +425,49 @@ $(function() {
 			socket.emit("sessions:get");
 		}
 
+		if (target === "#help" || target === "#changelog") {
```
astorije: ` || target === "#changelog"` because if we're loading the page while on the changelog, we need to retrieve its content. Now, well, this isn't working because of something entirely different (handlers are not set up so we're sent to a lobby), but it's ready to support it.

client/js/lounge.js
```diff
@@ -447,8 +425,49 @@ $(function() {
...
+			const now = Date.now();
+			// Don't check more than once a day
+			if (now - changelogRequestedAt > 86400 * 1000) {
```
astorije: One day was arbitrary. We don't release often so that's reasonable IMO, and there are probably not many people running the app 24/7 at this point, yet.

xPaw:
@astorije I primarily had changelogs only for private instances because it hits GitHub and we don't have caching implemented for that.

astorije:
@xPaw, didn't you have caching though, with the `one` handler you had? Now it is cached for 24h or until next page refresh, but refreshing the page on public mode is way more costly than on private because you need to reconnect to an IRC server every time.

I can disable it, but I think it should be fine, especially for that last reason.

xPaw:
@astorije I mean caching on the server, not client. It's easier to abuse in public mode.

astorije:
Okay, I'll disable it and have the release notes link send user to the GitHub release in public mode. I think that's a bit unnecessary, but out of precaution, I guess it doesn't really hurt. I just don't like ripping off functionalities from the public mode :)

xPaw:
@astorije In a future PR we can re-enable public changelogs if we add caching on the server.

astorije:
@xPaw, rather than re-disabling for public, I directly added simple caching on the server.

Now I'm wondering if we shouldn't display version check on public as well. I don't see why not.

PR #1330, opened 2017-07-14
Add SOURCE CTCP Command

MaxLeiter:
Note: won't work until new irc-framework release with https://github.com/kiwiirc/irc-framework/pull/104

image

ref: https://defs.ircdocs.horse/defs/ctcp.html

Two comments: 
1) Linting is very strange with switch statements, but that should be fixed in a different PR (or are we okay with how it is?)
2) ~iirc, we decided we wanted to slowly convert codebase over to ES6 -- right now I just converted part of `ctcp.js`, should I convert the rest? Or convert none?~

### DIFF
--- a/src/plugins/irc-events/ctcp.js
+++ b/src/plugins/irc-events/ctcp.js
@@ -1,17 +1,17 @@
 "use strict";
 
-var Msg = require("../../models/msg");
+const Msg = require("../../models/msg");
 
 module.exports = function(irc, network) {
-	var client = this;
+	const client = this;
 
 	irc.on("ctcp response", function(data) {
-		var chan = network.getChannel(data.nick);
+		let chan = network.getChannel(data.nick);
 		if (typeof chan === "undefined") {
 			chan = network.channels[0];
 		}
 
-		var msg = new Msg({
+		const msg = new Msg({
 			type: Msg.Type.CTCP,
 			time: data.time,
 			from: data.nick,
@@ -21,14 +21,20 @@ module.exports = function(irc, network) {
 		chan.pushMessage(client, msg);
 	});
 
-	irc.on("ctcp request", function(data) {
+	irc.on("ctcp request", (data) => {
 		switch (data.type) {
-		case "PING":
-			var split = data.message.split(" ");
+		case "PING": {
+			const split = data.message.split(" ");
 			if (split.length === 2) {
 				irc.ctcpResponse(data.nick, "PING", split[1]);
 			}
 			break;
+		}
+		case "SOURCE": {
+			const packageJson = require("../../../package.json");
+			irc.ctcpResponse(data.nick, "SOURCE", packageJson.repository.url);
+			break;
+		}
 		}
 	});
 };


src/plugins/irc-events/ctcp.js
```diff
@@ -21,14 +22,25 @@ module.exports = function(irc, network) {
...
+		case "TIME": {
+			const time = moment().format("dddd, MMMM Do YYYY, h:mm:ss a"); // "Sunday, February 14th 2010, 3:25:50 pm"
+			irc.ctcpResponse(data.nick, "TIME", time);
+			break;
+		}
+		case "SOURCE": {
+			const source = "https://www.github.com/thelounge/lounge";
+			irc.ctcpResponse(data.nick, "SOURCE", source);
+			break;
+		}
```
MaxLeiter: I really think we should have case statements indented in the switch

src/plugins/irc-events/ctcp.js
```diff
@@ -21,14 +21,20 @@ module.exports = function(irc, network) {
...
-			var split = data.message.split(" ");
+		case "PING": {
+			const split = data.message.split(" ");
 			if (split.length === 2) {
 				irc.ctcpResponse(data.nick, "PING", split[1]);
 			}
 			break;
 		}
+		case "SOURCE": {
+			const source = "https://www.github.com/thelounge/lounge";
```
xPaw: We could remove `.git` from `repository.url` in `package.json` and then we could just put it in here :)

astorije: Yeah, if we could avoid hardcoding this in the source code, that'd be preferable. If we don't want to change the `package.json` file, worst case, we could strip it at reading time here.

src/plugins/irc-events/ctcp.js
```diff
@@ -1,17 +1,17 @@
 "use strict";
 
-var Msg = require("../../models/msg");
+const Msg = require("../../models/msg");
 
-module.exports = function(irc, network) {
-	var client = this;
+module.exports = (irc, network) => {
```
astorije: Meh, `function` is fine here, but doesn't matter much.

src/plugins/irc-events/ctcp.js
xPaw: There's no need to strip it, as github handles that url fine.

src/plugins/irc-events/ctcp.js
xPaw: Yeah probably better to keep the function in exports. I'm not sure what scope it would get doing this.

src/plugins/irc-events/ctcp.js
astorije: Not what I meant, sorry. I meant that if we're not okay having `https://www.github.com/thelounge/lounge` in `package.json` and want to keep `https://www.github.com/thelounge/lounge.git` there but `https://www.github.com/thelounge/lounge` in `SOURCE`, then we can still read it from `package.json` and strip it here.

Either way, I really think it shouldn't be hardcoded here.

MaxLeiter: I think xPaw meant Github handles `https://www.github.com/thelounge/lounge.git` just fine, so no need to strip.

MaxLeiter:
should be good to go

src/plugins/irc-events/ctcp.js
```diff
@@ -21,14 +21,20 @@ module.exports = function(irc, network) {
...
+		case "PING": {
+			const split = data.message.split(" ");
 			if (split.length === 2) {
 				irc.ctcpResponse(data.nick, "PING", split[1]);
 			}
 			break;
 		}
+		case "SOURCE": {
+			const packageJson = require("../../../package.json");
+			irc.ctcpResponse(data.nick, "SOURCE", packageJson.repository.url);
```
astorije: So is it expected that this will respond with `https://github.com/thelounge/lounge.git` or should it respond with `https://github.com/thelounge/lounge`?

xPaw: It is fine to responding with `lounge.git`, yes.

PR #1334, opened 2017-07-14
Emoji fuzzysearching

MaxLeiter:
image

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -45,17 +45,26 @@ $(function() {
 
 	// Autocompletion Strategies
 
+	const emojiSearchTerms = Object.keys(emojiMap);
 	const emojiStrategy = {
 		id: "emoji",
 		match: /\B:([-+\w]*):?$/,
 		search(term, callback) {
-			callback(Object.keys(emojiMap).filter((name) => name.indexOf(term) === 0));
+			const results = fuzzy.filter(
+				term,
+				emojiSearchTerms,
+				{
+					pre: "<b>",
+					post: "</b>"
+				}
+			);
+			callback(results.map((el) => [el.string, el.original]));
 		},
-		template(value) {
-			return `<span class="emoji">${emojiMap[value]}</span> ${value}`;
+		template([string, original]) {
+			return `<span class="emoji">${emojiMap[original]}</span> ${string}`;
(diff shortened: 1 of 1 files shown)


MiniDigger:
I so need this, I never can find the emoji I want ^^
Great idea

MaxLeiter:
Should be simple

astorije:
Just to make sure you see this, if you happen to try, probably do it in a different commit so you can easily abort if it doesn't work without having to figure out what was your previous commit. Totally cool if we prefer to do that in a further PR to at least improve the current situation.

MaxLeiter:
screen shot 2017-07-15 at 12 22 36 pm

@astorije there we go

client/js/lounge.js
```diff
@@ -45,14 +45,24 @@ $(function() {
...
+					pre: "<b>",
+					post: "</b>"
+				}
+			);
+			callback(results.map((el) => el.string));
 		},
 		template(value) {
-			return `<span class="emoji">${emojiMap[value]}</span> ${value}`;
+			const cleansed = value.replace(/<b>|<\/b>/g, "");
+			return `<span class="emoji">${emojiMap[cleansed]}</span> ${value}`;
```
astorije: So, I don't like that we add the boldness in one method, then remove it in the next one, especially since `fuzzy` has an `original` key. I played a bit locally, and made this step go away with the following (and some nice ES6 destructured assignments):

```diff
─────────────────────────
modified: client/js/lounge.js
─────────────────────────
@ lounge.js:61 @ $(function() {
                    post: "</b>"
                }
            );
-           callback(results.map((el) => el.string));
+           callback(results.map((el) => [el.string, el.original]));
        },
-       template(value) {
-           const cleansed = value.replace(/<b>|<\/b>/g, "");
-           return `<span class="emoji">${emojiMap[cleansed]}</span> ${value}`;
+       template([string, original]) {
+           return `<span class="emoji">${emojiMap[original]}</span> ${string}`;
        },
-       replace(value) {
-           return emojiMap[value];
+       replace([, original]) {
+           return emojiMap[original];
        },
        index: 1
    };
```

xPaw: Ah didn't think of this, this is good.

PR #1336, opened 2017-07-15
Remove "in" transition on autocomplete items to fix a visual glitch on value refresh

astorije:
@xPaw noticed the refresh is gross when typing further characters.

I disabled only the transition when it shows up, not when it disappears (btw I'm not sure I understand why "in" transitions are disabled when setting them to `none` on `:hover` in CSS...).
@xPaw, can you try this and tell me what you think? Alternative is to simply disable them, but I figured I should give it a try like this first.


xPaw:
Interesting fix, did not expect this to work.

PR #1347, opened 2017-07-18
Hide empty preview divs when there are no previews loaded

### LINKED ISSUE #1343: URL parser adds preview placeholder divs within the part/quit reason
Since #1303, "placeholder" `div`s get added in the middle of the message, because of [`parse` being called inline](https://github.com/thelounge/lounge/blob/9735f2b7a112635c4e75ba8d958e3569cb3c3b24/client/views/actions/part.tpl#L5). This causes line breaks like the following:

screen shot 2017-07-17 at 00 44 47
screen shot 2017-07-17 at 00 46 07

This would also be true for [`kick`](https://github.com/thelounge/lounge/blob/9735f2b7a112635c4e75ba8d958e3569cb3c3b24/client/views/actions/kick.tpl#L5), [`list`](https://github.com/thelounge/lounge/blob/9735f2b7a112635c4e75ba8d958e3569cb3c3b24/client/views/actions/channel_list.tpl#L12), [`quit`](https://github.com/thelounge/lounge/blob/9735f2b7a112635c4e75ba8d958e3569cb3c3b24/client/views/actions/quit.tpl), [`topic`](https://github.com/thelounge/lounge/blob/9735f2b7a112635c4e75ba8d958e3569cb3c3b24/client/views/actions/topic.tpl), etc. though the consequences are not dramatic in most.

Not sure yet what are the best option(s) to fix that.
It's a regression from 2.3.2, do we need to make this a blocker for 2.4.0?

astorije:
Fixes #1343.

Not the cleanest thing in the world, but getting this as clean as can be is actually a bit of effort (sending the list of links to be previewed as part of the `emit("msg")` instead of at parse time) so until then it will have to do.

Before | After
--- | ---
screen shot 2017-07-18 at 00 38 05 | screen shot 2017-07-18 at 00 37 52

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1201,6 +1201,12 @@ kbd {
 	display: inline-flex !important;
 }
 
+/* Do not display an empty div when there are no previews. Useful for example in
+part/quit messages where we don't load previews (adds a blank line otherwise) */
+#chat .preview:empty {
+	display: none;
+}
+
 #chat .count {
 	background: #fafafa;
 	height: 48px;


astorije:
> getting this as clean as can be is actually a bit of effort (sending the list of links to be previewed as part of the `emit("msg")` instead of at parse time)

Working on something along those lines but it'll take a few days before it's ready fo until then this PR will do.

PR #1348, opened 2017-07-18
Fix previews in actions ("/me") after page reload

### LINKED ISSUE #1346: URLs passed in /me do not prefetch after refresh anymore
This was most likely caused by #1303. Sorry :man_shrugging:

astorije:
Fixes #1346 

Took me quite a while to figure this one out!

What happens essentially is that `renderPreview` adds preview templates (toggle button and content) with ``msg.find(`.text a[href="${preview.link}"]`)`` and ``msg.find(`.preview[data-url="${preview.link}"]`)``. At that point, `msg` contained `$(templates["msg_action"](data.msg))` but the content of the potential "actions" templates.

Switching those makes sure that when the previews get added, the content of the generated message is fully loaded before adding the previews.

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -73,15 +73,13 @@ function buildChatMessage(data) {
 	const msg = $(templates[template](data.msg));
 	const text = msg.find(".text");
 
-	if (data.msg.previews.length) {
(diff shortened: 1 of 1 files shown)


xPaw:
Nice catch.

Issue #773, opened 2016-12-02
Lounge setup command


MaxLeiter:
We should probabaly write a cli for setting up the config for new users to make deployment easier.

Maybe the first option can ask if you're a "simple" or "advanced" user; advanced could ask about useHexIp, WebIRC, etc and simple could just ask private/public, first account, etc.

xPaw:
Advanced configuration would just be editing the config.json yourself.

As for the simple setup, we could ask a few simple options when starting the server for the first time. But we need to decide which options to ask for (can't ask for all of them obviously).

1. Public?
2. Logs?
3. Prefetch?
4. Port?
5. identd?

PR #1350, opened 2017-07-18
Only create config folder in 'start' command

### LINKED ISSUE #520: The --help option shouldn't create config directory

xPaw:
Fixes #520.
Fixes #1288.

### DIFF
--- a/src/command-line/add.js
+++ b/src/command-line/add.js
@@ -1,9 +1,9 @@
 "use strict";
 
-var ClientManager = new require("../clientManager");
-var colors = require("colors/safe");
-var program = require("commander");
-var Helper = require("../helper");
+const colors = require("colors/safe");
+const program = require("commander");
+const fs = require("fs");
+const Helper = require("../helper");
 const Utils = require("./utils");
 
 program
@@ -11,6 +11,13 @@ program
 	.description("Add a new user")
 	.on("--help", Utils.extraHelp)
 	.action(function(name) {
+		if (!fs.existsSync(Helper.USERS_PATH)) {
+			log.error(`${Helper.USERS_PATH} does not exist.`);
+			return;
+		}
+
+		const ClientManager = require("../clientManager");
+
 		if (Helper.config.public) {
 			log.warn(`Users have no effect in ${colors.bold("public")} mode.`);
 		}
--- a/src/command-line/list.js
+++ b/src/command-line/list.js
@@ -1,8 +1,8 @@
 "use strict";
 
-var ClientManager = new require("../clientManager");
-var program = require("commander");
-var colors = require("colors/safe");
+const colors = require("colors/safe");
+const program = require("commander");
+const fs = require("fs");
 const Helper = require("../helper");
 const Utils = require("./utils");
 
@@ -11,6 +11,13 @@ program
 	.description("List all users")
 	.on("--help", Utils.extraHelp)
 	.action(function() {
+		if (!fs.existsSync(Helper.USERS_PATH)) {
+			log.error(`${Helper.USERS_PATH} does not exist.`);
+			return;
+		}
+
+		const ClientManager = require("../clientManager");
+
 		if (Helper.config.public) {
 			log.warn(`Users have no effect in ${colors.bold("public")} mode.`);
 		}
--- a/src/command-line/list.js
+++ b/src/command-line/list.js
@@ -28,12 +28,12 @@ program
 			return;
 		}
 
-		if (!users.length) {
-			log.info(`There are currently no users. Create one with ${colors.bold("lounge add <name>")}.`);
-		} else {
+		if (users.length > 0) {
 			log.info("Users:");
-			for (var i = 0; i < users.length; i++) {
-				log.info(`${i + 1}. ${colors.bold(users[i])}`);
-			}
+			users.forEach((user, i) => {
+				log.info(`${i + 1}. ${colors.bold(user)}`);
+			});
+		} else {
+			log.info(`There are currently no users. Create one with ${colors.bold("lounge add <name>")}.`);
 		}
 	});
--- a/src/command-line/start.js
+++ b/src/command-line/start.js
@@ -1,8 +1,11 @@
 "use strict";
 
-var program = require("commander");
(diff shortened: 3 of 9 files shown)


astorije:
This is pretty awful to review lol, and `?w=1` doesn't help much. @xPaw, do you have any advice to help reviewing this?

xPaw:
@astorije review the first commit, the second one is proof of concept.

src/command-line/add.js
```diff
@@ -1,45 +1,67 @@
...
-							enableLog.charAt(0).toLowerCase() === "y"
-						);
-					}
-				});
+
+			add(
+				manager,
+				name,
+				password,
+				enableLog.charAt(0).toLowerCase() === "y"
```
MaxLeiter: I know it's not documented, but maybe || "t" for `true`? Can't hurt?

astorije: Not sure why we would do that. It was fine and `y` only before this PR, and most tools (like Git) will do this too.

astorije:
> @astorije review the first commit, the second one is proof of concept.

Any more advice now that you have squashed the commits? :D

xPaw:
I removed the second commit

src/command-line/start.js
```diff
@@ -15,6 +16,11 @@ program
 	.option("    --private", "start in private mode")
 	.description("Start the server")
 	.action(function(options) {
+		initalizeConfig();
+
+		const ClientManager = require("../clientManager");
+		const server = require("../server");
```
astorije: Just trying to understand, why did you move this here?

src/command-line/add.js
```diff
@@ -1,14 +1,21 @@
...
+const Helper = require("../helper");
 
 program
 	.command("add <name>")
 	.description("Add a new user")
 	.action(function(name) {
+		if (!fs.existsSync(Helper.USERS_PATH)) {
+			log.error(`${Helper.USERS_PATH} does not exist.`);
+			return;
+		}
```
astorije: It's a bit of an awkward workflow now, because to start in private mode, you need at least one user, and to add a user, you need a configuration file / tree, which gets created with `lounge start`.
It's awkward currently too, because you still need to set to private mode _after_ creating a user (creating a user should fail if mode is public, IMO, but that's a different story) but a bit less.

So the workflow after this PR will be:

1. `lounge start` to create config _and_ start a public instance
2. Kill it
3. Update configuration file to set to private mode
4. `lounge add <user>`
5. `lounge start` to start private mode with that user

This is what I think it makes a lot of sense to have a `lounge init` that would replace step 1/2 (and 3 if making it interactive, see #773).

What do you think? How can we easily change this to make it a bit less awkward?

astorije: We mentioned making the default mode private, which I agree, however this PR would make the setup workflow even more convoluted:

0. `lounge start` fails because there is no user (we should still start the instance even without user set up, btw, since we always user files #779)
1. Update configuration file to temporarily setup The Lounge in public
2. `lounge start` to create config _and_ start a public instance
3. Kill it
4. Update configuration file to set to private mode
5. `lounge add <user>`
6. `lounge start` to start private mode with that user

xPaw: So to fix the problem we should just not stop the server if there are no users found?

astorije: I'd say that's a first step, yeah. I've done it in #1443, with a bit of other good stuff.

astorije:
You have a bunch of conflicts and you'll have more after #1443 is merged. Don't bother fixing them now, I'm happy to fix those conflicts after #1443 if you're okay with that.

src/command-line/add.js
```diff
@@ -1,14 +1,21 @@
...
+
+		const ClientManager = require("../clientManager");
```
astorije: I see you have moved that `require` statement in all files, why is that?

astorije:
Ok @xPaw, I should be done with rebasing. I also tested and saw nothing critical. Mind taking a look at what I pushed to see if you see something blatantly wrong + see the questions I posted? Thanks!

src/command-line/add.js
xPaw: Because running any CLI command requires all the subcommands, which in turn would require `ClientManager`, and it would cascade all the way down into loading all the server code before we would want to do it (e.g. before starting the server). This was one of the fixes for not creating config file incorrectly.

src/command-line/list.js
```diff
@@ -1,8 +1,8 @@
 "use strict";
 
-var ClientManager = new require("../clientManager");
-var program = require("commander");
-var colors = require("colors/safe");
+const colors = require("colors/safe");
+const program = require("commander");
+const fs = require("fs");
 const Helper = require("../helper");
```
xPaw: @astorije You may have forgot to move this into callback?

astorije: I can move it into the callback, but I believe it's safe in here, no?
In fact, all `const Helper = require("../helper");` but one were at the top of the file, except one which was in a function, I think. Any reason to not have them at the top (aside your explanation on `ClientManager`, which makes sense, but it shouldn't apply for `Helper`).

xPaw: This PR before rebase had all helpers inside callbacks, after the rebase this one specifically was left out, while others are still in the callback, for consistency sake, requires that run our should live in callbacks (to prevent future problems).

astorije: Sorry, I'm confused. I'd like to double-check with you before committing any further :)

All `src/command-line/*` files (after this PR, in the current state) but 2 (`remove` and `reset`) have the `require("../helper")` statement at the top and not in the callback.
I don't get why we should put `require("../helper")` in the callback, imo they should really be at the top of each file.

Let me know what you want me to do.

xPaw: Okay, keep them at the top for now.

PR #1352, opened 2017-07-18
Fix incorrect mode on kick target

xPaw:


client/views/actions/kick.tpl
```diff
@@ -1,6 +1,6 @@
 {{> ../user_name nick=from}}
 has kicked
-{{> ../user_name nick=target}}
+{{> ../user_name nick=target mode=""}}
```
astorije: Can kick target be voiced or op or whatever? In other word, why would it always be empty mode?

xPaw: This is restoring previous behaviour. Ideally we would show the real mode, yes. But I don't want to touch that code as I have another PR open that touches same code already.

astorije: Makes sense, still better than what's on master anyway.

PR #1353, opened 2017-07-19
Enforce correct order for previews on server-side prefectch rather than at client parsing

astorije:
This is essentially a better way / improvement of https://github.com/thelounge/lounge/pull/1303.

This has the benefit of not adding `.preview` divs everywhere, anytime we use `parse()`, and also to un-tie the position of the preview blocks from the result of the helper. This means that templates that call `parse` and have some extra markup after that are not constrained anymore.

This is effectively an alternative, better way to fix https://github.com/thelounge/lounge/issues/1343, but the initial fix that was put in place (https://github.com/thelounge/lounge/pull/1347) is still relevant, for example to make sure a preview stays hidden (and does not add extra margin/padding/etc.) if the link does not prefetch.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -943,7 +943,7 @@ kbd {
 
 #chat .time,
 #chat .from,
-#chat .text {
+#chat .content {
 	display: block;
 	padding: 2px 0;
 	flex: 0 0 auto;
@@ -965,7 +965,7 @@ kbd {
 	position: relative;
 }
 
-#chat .text {
+#chat .content {
 	flex: 1 1 auto;
 	overflow: hidden;
 }
@@ -1024,7 +1024,7 @@ kbd {
 #chat.colored-nicks .user.color-31 { color: #ff4846; }
 #chat.colored-nicks .user.color-32 { color: #ff199b; }
 
-#chat .text {
+#chat .content {
 	padding-left: 10px;
 	padding-right: 6px;
 }
@@ -1093,14 +1093,14 @@ kbd {
 	display: none !important;
 }
 
-#chat .join .text,
-#chat .kick .text,
-#chat .mode .text,
-#chat .nick .text,
-#chat .part .text,
-#chat .quit .text,
-#chat .topic .text,
-#chat .topic_set_by .text {
+#chat .join .content,
+#chat .kick .content,
+#chat .mode .content,
+#chat .nick .content,
+#chat .part .content,
+#chat .quit .content,
+#chat .topic .content,
+#chat .topic_set_by .content {
 	color: #999;
 }
 
@@ -1978,7 +1978,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 	#chat .time,
 	#chat .from,
-	#chat .text {
+	#chat .content {
 		border: 0;
 		display: inline;
 		padding: 0;
@@ -2097,7 +2097,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 #chat .part-reason,
 #chat .quit-reason,
 #chat .new-topic,
-#chat .action-text,
+#chat .action .text,
 #chat table.channel-list .topic {
 	white-space: pre-wrap;
 }
--- a/client/views/actions/action.tpl
+++ b/client/views/actions/action.tpl
@@ -1,6 +1,6 @@
 {{> ../user_name nick=from}}
 <span class="text">{{{parse text}}}</span>
 
-{{#each links}}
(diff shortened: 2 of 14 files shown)


src/models/msg.js
```diff
@@ -31,6 +31,7 @@ function Msg(attr) {
 	_.defaults(this, attr, {
 		from: "",
 		id: id++,
+		links: [],
 		previews: [],
```
astorije: I originally started this with `previews: new Map()`, where each link in the text was added as a key with an `null` value (because `Map` keep key ordering, and they can be easily looped over without some JS heaviness), but it turns out Socket.io transports stuff to the client as JSON so my `Map` was converted to an `Object`.

It doesn't seem like there is a solution for that on the Socket.io side, people deconstruct/reconstruct using arrays of arrays and such, brr. I couldn't be bothered to do some custom/crazy serialization and this worked well enough.

client/views/msg.tpl
```diff
@@ -7,5 +7,11 @@
 			{{> user_name nick=from}}
 		{{/if}}
 	</span>
-	<span class="text">{{{parse text}}}</span>
+	<span class="text">
+		{{{parse text}}}
```
xPaw: This breaks #1229 FYI (but it was probably already broken by link previews.

src/models/msg.js
xPaw: You can still take advantage of `previews` array`. By default you can set it to have `type` of either empty string or `loading`, and compare against that on the client. I think this will be slightly better than having a separate links array.

client/views/actions/action.tpl
```diff
@@ -1,2 +1,6 @@
 {{> ../user_name nick=from}}
 <span class="action-text">{{{parse text}}}</span>
+
+{{#each links}}
+  <div class="preview" data-url="{{this}}"></div>
```
xPaw: tab?

client/views/msg.tpl
astorije: Good catch, I'll look into it. Any recommendaions?

client/views/actions/action.tpl
astorije: Ah, good find, will fix.

src/models/msg.js
astorije: I have thought about that, but honestly it will look clunky at different levels, if anything because fetching callback will have to find which preview needs to be updated, and do it in place. I agree that having a separate array for ordering is not perfect, but I don't think that would be a better/simpler solution :/

xPaw: That should be a simple `findIndex`. I just don't want to have duplicate links stored in memory.

astorije: 😩
It doesn't seem very logical to me. But I'll try, keep an open mind about it, and see how it looks.

client/css/style.css
```diff
@@ -2090,7 +2090,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
    For example: user has quit ( ===> L   O   L <=== )  */
 
 #windows .header .topic,
-#chat .message .text,
+#chat .message .text span,
```
astorije: I don't like that, but I honestly couldn't find a decent class name. Suggestions?

xPaw: Can we avoid the span by just removing whitespace in the template?

astorije: I did that at first, yeah, but that means that content of previews are also subject to that, plus do a `~` dance (or manually removing new lines, until next person breaks it) in the template... Unfortunately not ideal.

astorije: BTW I believe that this change (as opposed to removing whitespaces in the template) would fix bugs 2 and 3 of https://github.com/thelounge/lounge/issues/1357.

astorije: Fix required was actually bigger than that so I'll extract it to its own PR.

astorije: Actually, it turns out that I can't extract it to its own PR, because the fix relies on the fact that adding the `preview` block is taken out of the `parse` method. So these are inter-dependent, but at least I'll keep the commits separate.

src/models/msg.js
astorije: I could make it so `findIndex` is not even necessary. Still a bit of a brain-twister to me, but I'm good either way.

@xPaw, I have kept the logic change from adding `links` to re-using `previews` in its own commit to see the transition. If you prefer this, I'll merge both commits (first and third of this PR) together.

astorije:
This PR is in the end much bigger than I was hoping, but I had to solve a few issues along the way, all of them connected.

It should be ready now, and should not break #1242 anymore.

PR #1354, opened 2017-07-19
Change more to use id rather than count.

YaManicKill:
This needs to be merged before #759 

I haven't managed to 100% test it because I restarted the server just now and don't have any "more" to do. I'll test it once it has backed up a bit.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -693,11 +693,12 @@ $(function() {
 
 	chat.on("click", ".show-more-button", function() {
 		var self = $(this);
-		var count = self.parent().next(".messages").children(".msg").length;
+		var lastMessage = self.parent().next(".messages").children(".msg").first();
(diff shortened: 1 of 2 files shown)


astorije:
Just curious, what happens in channels that have a ton of join/parts but very few messages, like `#freenode`? Does it load potentially thousands of messages to try and grab 100 actual messages sent by people?

I get that your condensing PR would be solving that anyway, I'm just going over what happens/would happen in the interim.

YaManicKill:
> Just curious, what happens in channels that have a ton of join/parts but very few messages, like #freenode? Does it load potentially thousands of messages to try and grab 100 actual messages sent by people?

This PR doesn't change this at all. It still gets 100 messages.

All it does is send the last message id, rather than the number of messages we currently have in chat window.

YaManicKill:
Thanks, @xPaw, can confirm after I left it on all day that it works. Good stuff.

astorije:
Thanks for the clarification, @YaManicKill!

PR #1356, opened 2017-07-19
Only match emoji autocomplete after two characters

### LINKED ISSUE #1355: Open emoji autocomplete menu after two characters have been entered (instead of just the leading colon)
Example: Typing `:` currently opens the menu and selects 💯.
Instead: Do not open the menu until the leading character (`:`) and two additional have been entered.

When attempting to send a message that ends in an emoticon, the emoji menu attempts to autocomplete and replaces the emoticon with an emoji.

The user is required to either hit `ESC` after typing their emoticon or, as often happens to me, delete the emoji, type the traditional emoticon again, and then hit `ESC` before hitting `Enter`.

The emoticon...
...`:x` becomes ❌ 
...`:o` becomes ⭕️ 
...`:3` becomes 🥉 
...`:p` becomes 😔

This interaction is very frustrating.

The solution I believe is best (and is also employed by Discord) is to require two characters after the leading character (`:`) before opening the menu. Thus, typing `:` doesn't suggest anything (which makes sense); typing `:p` doesn't attempt to autocomplete (which allows unfettered use of traditional emoticon); and typing `:pe` will still bring up the menu and attempt to autocomplete with 😔.

MaxLeiter:
https://regex101.com/r/QNAtHZ/5
Closes #1355

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -48,10 +48,12 @@ $(function() {
 	const emojiSearchTerms = Object.keys(emojiMap);
 	const emojiStrategy = {
 		id: "emoji",
-		match: /\B:([-+\w]*):?$/,
+		match: /\B:([-+\w:?]{2,}):?$/,
 		search(term, callback) {
 			const results = fuzzy.filter(
-				term,
+				// Trim colon from the matched term,
+				// as we are unable to get a clean string from match regex
+				term.replace(/:$/, ""),
 				emojiSearchTerms,
 				{
 					pre: "<b>",


astorije:
Why is this a `do not merge`?

MaxLeiter:
The regex isn't matching correctly and I can't think of a proper fix -- it'll match multiple `:` after the emoji. i.e. `:smile::::`

dgw:
It's not, according to the regex101 you put in the OP. Putting test lines with multiple trailing `:`s in the input there, they don't match.

astorije:
@dgw, I could make `:smile:::` match indeed.

@MaxLeiter, what about `\B:([-+\w]+[-+\w:])$`?

screen shot 2017-07-20 at 01 28 49

If it looks gross, that's normal, that's 'cause I suck at regexes. It works for me, but I don't know if there is a better way.

dgw:
@astorije I could not. But I see you're on regex v5; when I checked, it was still showing v4, so :man_shrugging:

MaxLeiter:
OP is regex 5; maybe weird caching thing going one

dgw:
@MaxLeiter I clicked the link from the GitHub email, and I see you edited the OP, so that's probably it.

astorije:
> @MaxLeiter, what about `\B:([-+\w]+[-+\w:])$`?

Isn't this working in the end? I have tried a few things with it and it seems fine to me.

MaxLeiter:
@maxpoulin64 any thoughts/comments on using `\B:([-+\w]+[-+\w:])$`?

astorije:
Okay so @MaxLeiter and I identified an issue with the new Regex, something that we cannot overcome with a simple Regex. From the channel:

```
21:59 <MaxLeiter> the second : (end of the word) doesn't match correctly
21:59 <MaxLeiter> so if you type the full complete (i.e. :smile: or :+1: as opposed to :+1 it doesn't autocomplete0
22:00 <MaxLeiter> do we want that?
00:03 <astorije> Ah, I see the problem, MaxLeiter, one sec
00:03 <astorije> Essentially, the Regex is fine, but then the matched keyword includes the ":", which is not part of the emoji map
00:07 <astorije> >e can't have ":" part of the ()
00:21 <astorije> Uuuh, I don't think it can be done without altering the lib :/
00:21 <MaxLeiter> yeah
00:22 <MaxLeiter> which makes the regex a PITA
00:22 <MaxLeiter> so i say we go with "lesser of the evils"
00:22 <MaxLeiter> question is which one
00:22 <astorije> So we have 2 cases, which is fine with "|", except that it triggers 2 capture groups, and one will always be undefined
00:22 <astorije> And the lib always takes the first one, even if it's undefined
00:28 <astorije> Honestly, taking my solution and stripping the last char if ":" would work
00:28 <astorije> It feels gross though
00:33 <astorije> Actually, I have an idea
```

After a bunch of attempts, this is the "best" that I can come up with:

```diff
──────────────────────────────
modified: client/js/lounge.js
──────────────────────────────
@ lounge.js:51 @ $(function() {
 	const emojiSearchTerms = Object.keys(emojiMap);
 	const emojiStrategy = {
 		id: "emoji",
-		match: /\B:([-+\w]+[-+\w:])$/,
+		match(textFromHeadToCaret) {
+			// Matching text is either any 2 chars, or 1 char IF followed by a `:`.
+			// Since we cannot have a single capture group to express those two
+			// conditions, we have to check is any matches, and use it as a match
+			// term if so.
+			const oneCharCheck = /\B:([-+\w]):$/;
+			if (oneCharCheck.test(textFromHeadToCaret)) {
+				return oneCharCheck;
+			}
+			return /\B:([-+\w]{2,}):?$/;
+		},
 		search(term, callback) {
 			const results = fuzzy.filter(
 				term,
```

MaxLeiter:
Done

xPaw:
I really don't like the fact this introduces more regex testing (which will run every time you type a character in input), at this point it's most likely cheaper to just trim the `:` in callback function.

MaxLeiter:
I really have no preference. @astorije?

astorije:
I'm not sure it's cheaper in both code or performance (hitting an extra regex per char is not going to change anything, but I see where you're coming from). When I tried, I couldn't get something 👌 when trimming after the fact, it always came back to having to do something like this.

@MaxLeiter, if you think you can pull it off, do you mind adding an extra commit with the attempt? Happy to change my mind if you can make this work.
But overall, I just want this change in like ASAP 😅

astorije:
@MaxLeiter, mind trying this out in an extra commit (for now)?

astorije:
@MaxLeiter, come back, we need youuuuu 🙏 :)

MaxLeiter:
Couldn't make anything work well.

astorije:
@xPaw, do you want to give it a quick try to see if you can help out @MaxLeiter?

xPaw:
As I've commented before:

> at this point it's most likely cheaper to just trim the : in callback function.

astorije:
@xPaw, the ping is because @MaxLeiter couldn't make that work.

xPaw:
Now that was a pain to update fork branch, but I've changed it to trim as I suggested.

PR #1358, opened 2017-07-21
Take into account wordboundaries for custom highlighting

### LINKED ISSUE #719: Custom notifications should only trigger when the message contains the keyword as word
Version 2.0.1
I'm using 2 different nickname frequently and i configured my lounge account to highlight the keywords 'lb, lb1a, @all'. But the highlighting keeps getting triggered when the string 'lb' is part of a complete different word: e.g: too**lb**ar .
I tried different keyword lists with and without empty strings
- lb, lb1a, @all
- lb,lb1a,@all
- lb , lb1a , @all

A fix would be to check if the keyword is not only part of the message string but also has empty strings around it (aka, is a whole word).
This was verified by PolarizedPumpkin in the lounge channel with current source.

starquake:
Fixes #719

When you have `apple` as a keyword is *does* highlight: 
```
Take a look at this magnificent apple!
I wouldn't call that an "apple"
Apple>Microsoft
```
But it *doesn't* highlight:
```
apples applepie grappler
```
Or if it's `sq` in my case it does highlight:
```
Hey sq!
```
But it doesn't highlight:
```
That guy is rather grotesque
```

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const $ = require("jquery");
+const escapeRegExp = require("lodash/escapeRegExp");
 const settings = $("#settings");
 const userStyles = $("#user-specified-css");
 const storage = require("./localStorage");
@@ -98,6 +99,15 @@ settings.on("change", "input, select, textarea", function() {
 			// otherwise, users get notifications for everything
 			return h !== "";
 		});
+		// Construct regex with wordboundary for every highlight item
+		const highlightsTokens = options.highlights.map(function(h) {
+			return escapeRegExp(h);
+		});
+		if (highlightsTokens && highlightsTokens.length) {
+			module.exports.highlightsRE = new RegExp("\\b(?:" + highlightsTokens.join("|") + ")\\b", "i");
+		} else {
+			module.exports.highlightsRE = null;
+		}
 	} else if (name === "showSeconds") {
 		chat.find(".msg > .time").each(function() {
 			$(this).text(tz($(this).parent().data("time")));
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -65,9 +65,11 @@ function buildChatMessage(data) {
 	const chan = chat.find(target);
 	let template = "msg";
 
-	if (!data.msg.highlight && !data.msg.self && (type === "message" || type === "notice") && options.highlights.some(function(h) {
-		return data.msg.text.toLocaleLowerCase().indexOf(h.toLocaleLowerCase()) > -1;
-	})) {
+	// See if any of the custom highlight regexes match
+	if (!data.msg.highlight && !data.msg.self
+		&& options.highlightsRE
+		&& (type === "message" || type === "notice")
+		&& options.highlightsRE.exec(data.msg.text)) {
 		data.msg.highlight = true;
 	}
 


client/js/options.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 const $ = require("jquery");
+const _ = require("lodash");
```
xPaw: You just included a very huge library (we don't use lodash on client)

client/js/options.js
```diff
@@ -91,6 +93,9 @@ settings.on("change", "input, select, textarea", function() {
 			// otherwise, users get notifications for everything
 			return h !== "";
 		});
+		options.highlightsRE = options.highlights.map(function (h) {
```
xPaw: This can be compiled into a single regex like `\b(one|two|three)\b`.

starquake: Will do

client/js/options.js
starquake: How do you suggest to fix that? Duplicate the functionality? Or can I require("lodash.escaperegexp") ?

xPaw: If requiring `lodash.escaperegexp` does not bring half of lodash with it, then that can work, otherwise just copy the function (it's a oneliner anyway).

starquake: Well if I want to require lodash.escaperegexp I need to install it and add it to packages.json as a dependency. Still okay? Or do you prefer copy pasting?

xPaw: Have you tried `lodash/escaperegexp`?

starquake: Ah I suspected something like that should exist, was looking for it. Thanks will do that.

client/js/options.js
```diff
@@ -26,6 +27,7 @@ const options = $.extend({
 	thumbnails: true,
 	userStyles: userStyles.text(),
 	highlights: [],
+	highlightsRE: [],
```
xPaw: This should be an empty regex or something because `options.highlightsRE.exec` may fail.

client/js/options.js
```diff
@@ -91,6 +93,11 @@ settings.on("change", "input, select, textarea", function() {
 			// otherwise, users get notifications for everything
 			return h !== "";
 		});
+		// Construct regex with wordboundary for every highlight item
+		var highlightsTokens = options.highlights.map(function(h) {
+			return escapeRegExp(h);
+		});
+		options.highlightsRE = new RegExp("\\b(?:" + highlightsTokens.join("|") + ")\\b", "i");
```
xPaw: What will this do if `highlightsTokens` is empty (e.g. when clearing out highlights in settings)

starquake:
Thanks for the feedback. Shouldn't commit stuff on fridays... Will fix.

astorije:
What's the status of this PR? @starquake, is it ready for review? @xPaw, what do you think of the feature change overall?

starquake:
@astorije No, I need to fix the things you mentioned. But suddenly I've been swamped with some other stuff. Will pick this up as soon when it all dies down a bit.

starquake:
I have incorporated your comments.

client/js/options.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 const $ = require("jquery");
+const escapeRegExp = require("lodash/escapeRegExp");
```
xPaw: Before we merge, gotta test if this actually only brings the relevant function, or half of lodash with it.

client/js/options.js
```diff
@@ -91,6 +93,13 @@ settings.on("change", "input, select, textarea", function() {
 			// otherwise, users get notifications for everything
 			return h !== "";
 		});
+		// Construct regex with wordboundary for every highlight item
+		var highlightsTokens = options.highlights.map(function(h) {
```
xPaw: There's already a `map` with `trim` in it, you could probably throw `escapeRegExp` into it instead of iterating twice.

client/js/options.js
```diff
@@ -91,6 +93,13 @@ settings.on("change", "input, select, textarea", function() {
...
+			return escapeRegExp(h);
+		});
+		if (highlightsTokens && highlightsTokens.length) {
+			options.highlightsRE = new RegExp("\\b(?:" + highlightsTokens.join("|") + ")\\b", "i");
+		}
```
xPaw: This needs an `else` statement to reset `highlightsRE` to `undefined` or `null`.

client/js/render.js
```diff
@@ -39,9 +39,11 @@ function buildChatMessage(data) {
 	const chan = chat.find(target);
 	let template = "msg";
 
-	if (!data.msg.highlight && !data.msg.self && (type === "message" || type === "notice") && options.highlights.some(function(h) {
-		return data.msg.text.toLocaleLowerCase().indexOf(h.toLocaleLowerCase()) > -1;
-	})) {
+	// See if any of the custom highlight regexes match
+	if (!data.msg.highlight && !data.msg.self
+		&& (type === "message" || type === "notice")
+		&& options.highlightsRE
```
xPaw: I would move this higher to avoid type string checks.

client/js/options.js
starquake: You mean something like this?
```
options.highlights = highlightString.split(",").map(function(h) {
  return escapeRegExp(h.trim());
})
```
Cause that would mean the contents of options.highlight will be escaped. I was afraid that might break other stuff that expected it to not be escaped.

xPaw: Oh you are right, nevermind!

client/js/render.js
starquake: Do you mean the order of the checks, or the entire if statement?

xPaw: Order of checks.

starquake: Like this?
```
	if (!data.msg.highlight && !data.msg.self
		&& options.highlightsRE
		&& (type === "message" || type === "notice")
		&& options.highlightsRE.exec(data.msg.text)) {
		data.msg.highlight = true;
	}
```

xPaw: Sure.

starquake: Are string checks more expensive then checking if something is true? Or do you have some other reason?

xPaw: It sure is cheaper to check a boolean. In this particular case it's an easy change, so no reason not to do it.

starquake: But options.highlightsRE is not a boolean. It's a regex. Or is that irrelevant?

xPaw: It's only checked against not being undefined, so that's not a problem.

client/js/options.js
starquake: Done

client/js/render.js
starquake: Done

client/js/options.js
starquake: Will test

starquake:
```
no require:
➜  lounge git:(highlight-wordboundary) ✗ npm run-script build
➜  lounge git:(highlight-wordboundary) ✗ du client/js/bundle*.js
360     client/js/bundle.js
2496    client/js/bundle.vendor.js

const escapeRegExp = require("lodash/escapeRegExp");
➜  lounge git:(highlight-wordboundary) ✗ npm run-script build
➜  lounge git:(highlight-wordboundary) ✗ du client/js/bundle*.js
360     client/js/bundle.js
2496    client/js/bundle.vendor.js

const escapeRegExp = require("lodash").escapeRegExp;
➜  lounge git:(highlight-wordboundary) ✗ npm run-script build
➜  lounge git:(highlight-wordboundary) ✗ du client/js/bundle*.js
360     client/js/bundle.js
3552    client/js/bundle.vendor.js
```

starquake: It adds nothing to the size because client/js/libs/handlebars/ircmessageparser/findChannels.js already has a reference to it.

starquake:
Incorporated your comments!

client/js/options.js
astorije: Sorry to be a pain for nitpicks like this, but could you make this a bit more ES6-like, please?

```js
const highlightsTokens = options.highlights.map((h) => escapeRegExp(h));
```

client/js/options.js
```diff
@@ -15,6 +16,7 @@ const options = {
 	coloredNicks: true,
 	desktopNotifications: false,
 	highlights: [],
+	highlightsRE: null,
```
astorije: Do we actually want to store this as part of localStorage?

astorije:
@starquake, could you add a few examples to your PR description to mention what this detect / doesn't detect, compared to before your PR? Thanks!

starquake:
@astorije Added some examples.

client/js/options.js
astorije: ^ @xPaw

xPaw: Yeah, I guess it has to be assigned directly to `module.exports`.

astorije: @starquake, would you mind doing that? :) Thanks!

starquake: @astorije Will do!

client/js/options.js
starquake: @astorije No problem, it's nice to be on the other side for once

client/js/options.js
```diff
@@ -98,6 +99,15 @@ settings.on("change", "input, select, textarea", function() {
 			// otherwise, users get notifications for everything
 			return h !== "";
 		});
+		// Construct regex with wordboundary for every highlight item
+		const highlightsTokens = options.highlights.map(function(h) {
+			return escapeRegExp(h);
+		});
+		if (highlightsTokens && highlightsTokens.length) {
+			module.exports.highlightsRE = new RegExp("\\b(?:" + highlightsTokens.join("|") + ")\\b", "i");
```
starquake: Not sure if referencing module.exports here is okay. If not, let me know!

xPaw: I wonder if doing `module.exports` puts it into localStorage regardless, as it may be referencing the same object..

astorije: That's a great point, it probably does. I was wondering why `shouldOpenMessagePreview` would not be stored somehow, but it turns out functions are not serialized by `JSON.stringify`:

```js
> JSON.stringify({ foo: 'bar', test() { return 'foo'; }, regex: new RegExp("test") })
'{"foo":"bar","regex":{}}'
```

Kinda sucks that we do `module.exports = options;` but then also use this file to do other things than just dealing with `options` directly.

@xPaw, what do you think?
I can think of a few ideas: wrapping `options` in a property `module.exports.whatever = options;` (requires editing of every place that has `require("[...]/options")`, creating a `optionsHelper.js` file, making `highlightsRE` a function instead as they're not serialized (that would only delay the issue instead of fixing it though), but none of those things are ideal. Open to ideas.

xPaw: Exporting under a separate object is probably the cleanest way of doing it. Or making a `getOption` function.

I have also found other unintentional variable that would be stored: https://github.com/thelounge/lounge/blob/0b85582744be6671c96c185866171798af4eead9/client/js/sorting.js#L33

starquake:
This pull request fixes a bug. Shouldn't the other problem be handled in a separate pull request. This problem existed before this PR and could maybe be allowed to still exist after it?

astorije:
@starquake, sure, but that means this PR is blocked until this other bug gets fixed. We don't want to pollute people's localStorage :/
@xPaw, do you want to take a look at it? Or someone else?

starquake:
@astorije Well in that case I think this PR should be blocked until the issue is fixed. I mean, I would be fine with that.

xPaw:
@astorije @starquake I don't think we need to block this PR for the fix, as the regex won't end up being saved as stringifying a regex into json does nothing.

starquake:
Oooh so I got lucky!

astorije:
Okay, but let's not pollute localStorage any further then. As in since this PR is ready, that's fine, but let's fix the underlying issue before potentially adding another one :)

astorije:
Hey @keegan, we have sticker packs for our contributors now!

astorije:
Hey @starquake, we have sticker packs for our contributors now!

PR #1359, opened 2017-07-21
Fix remaining issues with link previews

xPaw:


client/js/renderPreview.js
```diff
@@ -13,6 +13,18 @@ function renderPreview(preview, msg) {
...
+	const previewContainer = msg.find(`.preview[data-url="${preview.link}"]`);
+
+	// This is to fix a very rare case of rendering a preview twice
+	// This happens when a very large amount of messages is being sent to the client
+	// and they get queued, so the `preview` object on the server has time to load before
+	// it actually gets sent to the server, which makes the loaded preview sent twice,
+	// once in `msg` and another in `msg:preview`
+	if (previewContainer.data("loaded")) {
+		return;
+	}
```
astorije: I think it would be nicer to check that `previewContainer.hasClass("toggle-type-loading")`, which is the meaning `loading` type was for in #1353, instead of adding a new attribute for this.

client/js/renderPreview.js
```diff
@@ -12,6 +12,19 @@ function renderPreview(preview, msg) {
 		return;
 	}
 
+	preview.link = Handlebars.Utils.escapeExpression(preview.link);
```
astorije: Prefetching https://msdn.microsoft.com/en-us/library/windows/desktop/ms644989(v=vs.85).aspx (which requires escaping according to our `parse` tests) is broken on this PR. Pretty this is the cause.

xPaw: Huh?

client/js/renderPreview.js
xPaw: Good idea, didn't think of that.

xPaw: Wait what, there's no such class.

client/js/renderPreview.js
astorije: On `master`:

screen shot 2017-07-21 at 12 01 47

In this PR:

screen shot 2017-07-21 at 12 02 45

:man_shrugging:

client/js/renderPreview.js
astorije: Yes, my bad, I meant `previewContainer.is(':empty')` 😁

client/js/renderPreview.js
xPaw: `https://google.com/?v="test"wtf` produces a syntax error on master.

PR #1364, opened 2017-07-23
Added a folder for scss-files

brainmaniac:
made a folder where one can put scss files that will be read in to the page.

EDIT: added more modules so need to run npm install


astorije:
Hi @brainmaniac, thanks a lot for chiming in!

However, as I believe @xPaw let you know on the IRC channel, we have no direct need to involve SASS/SCSS in our build system yet (and as you've noted, there would be some trickery needed for themes).
Should we need to do that someday, we'll definitely get back to your PR to take example, at the very least.

Closing for now.

brainmaniac:
Hi @astorije,

Thanks for your kind words. I totally understand not implementing this as it is not integrated for the whole structure of the application.. but maybe as you say some parts can be useful the day (if it comes) you guys decide to implement sass. 

One nice way to handle the themes could be that each theme-file is just a file with scss-variables and that in the run/build proccess it builds a css file for each theme-variable-file... I'll see if get time to make an example of this but it will probably not happen in the near feature.

PR #1365, opened 2017-07-23
Add Previous/Next cycling buttons to the image viewer

astorije:
This adds 2 previous/next buttons on the image viewer to toggle through the images prefetched in the active channel.

- Only images that are expanded are cycled through
- <kbd>←</kbd> and <kbd>→</kbd> also trigger these buttons, just like <kbd>Esc</kbd> closes the image viewer.

Also:

- Move the &times; button of image viewer from HTML to CSS (like our other icons)
- Add an alternative text to the image in viewer

There are no animations/transitions at the moment. I wanted to add some, but it's a fairly different implementation (right now the entire block gets re-templated) so I couldn't in a timely fashion. Also, I'm not super good with CSS beatifulnesses like that, so if anyone wants to take this up, I would definitely be supportive.

I would love for this to make it to milestone **2.4.0** since the image viewer is a new feature in there (and because I'm changing the &times; icon thing). @xPaw, what do you think?

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -209,7 +209,9 @@ kbd {
 #chat .action .from:before,
 #chat .toggle-button:after,
 .context-menu-item:before,
-#nick button:before {
+#nick button:before,
+#image-viewer .previous-image-btn:before,
+#image-viewer .next-image-btn:before {
 	font: normal normal normal 14px/1 FontAwesome;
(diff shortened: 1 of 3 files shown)


MaxLeiter:
How does this work with multiple previews in one message?

astorije:
It's handled, look at the comments :)
Essentially, all visible images are cycled through "linearly", if that makes sense.

PR #1366, opened 2017-07-24
Keep track of preview visibility on the server so it persists at page reload

astorije:
Alright, this will be my last preview-related enhancement for this release.

This is going to be especially useful because of the support for multiple previews per message + the image viewer cycling. Before this PR, collapsing NSFW previews then refreshing then cycling on image previews would display them back, oops!

One issue to keep in mind with this implementation is that running `/collapse`/`/expand` will emit as many times as there are previews to collapse/expand. I don't think this is an issue that is going to have tremendous repercussions performance-wise, so I think any improvements over this can be done in a further PR (help welcome!).

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -280,6 +280,26 @@ function init(socket, client) {
 				client.names(data);
 			}
 		);
+
+		socket.on("msg:preview:toggle", function(data) {
+			const networkAndChan = client.find(data.target);
+			if (!networkAndChan) {
+				return;
+			}
+
+			const message = networkAndChan.chan.findMessage(data.msgId);
+
+			if (!message) {
+				return;
+			}
+
+			const preview = message.findPreview(data.link);
+
+			if (preview) {
+				preview.shown = data.shown;
+			}
+		});
+
 		socket.join(client.id);
 		socket.emit("init", {
 			active: client.lastActiveChannel,
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -36,10 +36,6 @@ function buildChatMessage(data) {
 		target = "#chan-" + chat.find(".active").data("id");
 	}
 
-	data.msg.previews.forEach((preview) => {
-		preview.shown = options.shouldOpenMessagePreview(preview.type);
-	});
-
 	const chan = chat.find(target);
 	let template = "msg";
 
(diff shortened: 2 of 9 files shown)


src/server.js
```diff
@@ -280,6 +280,14 @@ function init(socket, client) {
 				client.names(data);
 			}
 		);
+
+		socket.on("msg:preview:toggle", function(data) {
+			const chan = client.find(data.target).chan;
```
xPaw: Add undefined checks for all these `find`

xPaw:
So doing `/collapse` is going to spam the server with many events, instead of sending one in a list. Something to keep in mind.

src/server.js
astorije: @xPaw, done, are you okay with how I did that?

astorije:
> So doing /collapse is going to spam the server with many events, instead of sending one in a list. Something to keep in mind.

Right, mentioned that in the PR description when I opened it.
I don't think it's a blocker for this PR, but regardless, do you think it's going to be a performance issue since it's going through a websocket, as opposed to plain HTTP request? After all, the same information is being passed to the server, just that there is less socket overhead with sending a list. Note that I'm ignorant of the implementation details of websockets and that might make this question silly :)

src/server.js
```diff
@@ -280,6 +280,20 @@ function init(socket, client) {
 				client.names(data);
 			}
 		);
+
+		socket.on("msg:preview:toggle", function(data) {
+			const networkAndChan = client.find(data.target);
+			if (networkAndChan) {
```
xPaw: You could just `return` in each if instead of going deeper.

xPaw:
Maybe for collapse command we can do an optimization trick where it sends `msg:preview:collapse` just for the channel, and `msg:preview:toggle` is not sent for each click (not sure how you're going to filter that one out), and then the server just loops messages and sets shown to false on all of them (and backwards for expand)

src/server.js
astorije: Done. I honestly prefer the nested version, shorter and makes more sense to me, but whatever.

Adding `do not merge` label so I remember to squash before merging.

astorije:
Agreed we can probably benefit from optimization later, but I won't have the cycles to do that right now. I'll let someone else chime in in a further PR, though I don't believe this will be an issue anyway.

astorije:
Otherwise, I believe this PR is up for review :)

PR #1373, opened 2017-07-25
Add Stack Overflow link in CONTRIBUTING file

astorije:
Eventually, we probably want to move this to a new SUPPORT.md file, but in the meantime, this gives additional guidance and will hopefully push people to ask on SO first.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,6 +5,9 @@ your contributions.
 
 ### I have a question
 
+- Check out [existing questions on Stack Overflow](https://stackoverflow.com/questions/tagged/thelounge)
+  to see if yours has been answered before. If not, feel free to [ask for a new question](https://stackoverflow.com/questions/ask?tags=thelounge)
+  (using `thelounge` tag so that other people can easily find it).
 - Find us on the Freenode channel `#thelounge`. You might not get an answer
   right away, but this channel is full of nice people who will be happy to
   help you.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,15 +3,6 @@
 Welcome to The Lounge, it's great to have you here! We thank you in advance for
 your contributions.
 
-### I have a question
-
-- Check out [existing questions on Stack Overflow](https://stackoverflow.com/questions/tagged/thelounge)
-  to see if yours has been answered before. If not, feel free to [ask for a new question](https://stackoverflow.com/questions/ask?tags=thelounge)
-  (using `thelounge` tag so that other people can easily find it).
-- Find us on the Freenode channel `#thelounge`. You might not get an answer
-  right away, but this channel is full of nice people who will be happy to
-  help you.
-
 ### I want to report a bug
 
 - Look at the [open and closed
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -1,0 +1,11 @@
+## Support
+
+Welcome to The Lounge, it's great to have you here! If you have a question, or
+need help, you have a few options:
+
+- Check out [existing questions on Stack Overflow](https://stackoverflow.com/questions/tagged/thelounge)
+  to see if yours has been answered before. If not, feel free to [ask for a new question](https://stackoverflow.com/questions/ask?tags=thelounge)
+  (using `thelounge` tag so that other people can easily find it).
+- Find us on the Freenode channel `#thelounge`. You might not get an answer
+  right away, but this channel is full of nice people who will be happy to
+  help you.


MaxLeiter:
I personally think we should encourage Github issues and try to stop fragmentation

astorije:
I strongly disagree: GitHub issues for issues, Stack Overflow questions for questions. GitHub issues should not be used for support.

dgw:
I would argue that Superuser is a better place for users and administrators to ask questions, because that site is targeted at server and service administration.

No doubt Stack Overflow is top of your mind because you just asked a question there yourself, @astorije. :D

astorije:
> I would argue that Superuser is a better place for users and administrators to ask questions, because that site is targeted at server and service administration.

I tend to agree (though the line between Stack Overflow and Superuser is very often thin) but at the moment there are a few questions about The Lounge on IRC, and none on Superuser, so we can't have a tag there. And also, I don't have enough rep points to create a tag there.
So until someone asks a question on Superuser and a new tag gets created, Stack Overflow would do I think.

> No doubt Stack Overflow is top of your mind because you just asked a question there yourself, @astorije. :D

I'm not sure what you mean. My last question there was in March, and last answer was in April.

dgw:
> My last question there was in March, and last answer was in April.

Ugh, SO's wonky time display strikes again. In the tag list it says "modified 11 hours ago" by you, with no mention of the question author. Sigh. I should have clicked through.

> So until someone asks a question on Superuser and a new tag gets created, Stack Overflow would do I think.

The problem with that approach, I think, is that it becomes self-fulfilling: The documentation doesn't recommend Super User because there's no tag, and points to Stack Overflow because there is a tag already. Therefore, users go to ask questions on SO, because it's what the documentation says to do, and nobody creates a tag on the other site, perpetuating the problem.

It's a pretty typical Catch-22 problem, and the only solution is if someone else involved in this project has enough rep to create the tag. Sadly it seems I haven't enough points either, but surely someone in the community of maintainers and active users has.

> Serverfault as well... Seriously though, that ecosystem is a bit confusing sometimes :weary:

Yeah, it's not the easiest. I've been involved with Stack Exchange since 2008 (according to my network profile), and participated in the early days of Area 51, and I still get lost sometimes when deciding which site is in-scope for the knowledge I seek. For this case, I picked SU over Server Fault primarily because of SF's focus on the business environment, because many Lounge users aren't going to be setting it up in one.

astorije:
> the only solution is if someone else involved in this project has enough rep to create the tag

And if there are questions on Superuser about The Lounge, which there aren't.

At the end of the day, I just don't want to bikeshed and party-poop this. Both are good options (one is good, one is better, sure), but either is definitely better than nothing. It's one more step towards community support and better/easier adoption/daily use of The Lounge. And Stack Overflow is *de facto* a valid choice in general, and in our case, if anything, because there are a few existing questions already.
For comparison sake, I just looked at the `irc` tag [on Stack Overflow](https://stackoverflow.com/tags/irc/info) and [on Super User](https://superuser.com/tags/irc/info). Both were created at the same time, yet the difference in number of questions and tag documentation is staggering. Also for comparison sake, there is a tag for Kiwi on SO, not on SU.

Anyway, to summarize: I totally agree with you on the fundamentals, but being pragmatic here: we have a door that's opening to us, I really want to use this one instead of the closed one that we _may_ open someday :)
I really want to push this forward, but I don't want to offend anyone either, are you okay with this?

dgw:
> I really want to push this forward, but I don't want to offend anyone either, are you okay with this? :smiley:

Oh, you won't offend me. My only concern was making sure the alternatives were mentioned before a decision got made, which they have been. If I've come out swinging so hard on enough tickets that you're afraid of offending me, I apologize. I'm always doing it for the sake of what seems best for the project, according to my reasoning—sometimes more ideally than pragmatically, but that's where it comes from. :smiley:

Personally, I don't really have an issue with support questions getting raised in GitHub issues. I've seen it work on projects with much smaller communities than this one. Actually, what makes the absolute most sense to me is directing people to the IRC channel. For help with an IRC client, it makes a lot of sense to use IRC itself, but building a more long-lasting Q&A reference over time makes sense too. (And I doubt public logs for `#thelounge` are a good option for that, regardless of whether the channel's users would agree to public logging.) Using SO is good for that, and your point about Kiwi having a tag there is quite relevant, due to the similarities between the two projects.

One thought I had was maybe linking to SU *and* SO, but the potential for fragmentation there is pretty high. (Could turn out to be tags on both sites, with neither getting enough volume to appear "active" and worth using.)

The other thought I had was more explicitly presenting `#thelounge` @ freenode as a support venue, beyond the badge that currently appears in the readme (which is pretty easy to miss, if I'm a user looking for support). That's in addition to pointing at SO/SU/wherever.

tl;dr: I'm totally fine with putting SO in the readme.

astorije:
> If I've come out swinging so hard on enough tickets that you're afraid of offending me, I apologize.

Not at all 😉 It's just that these discussions can easily become religious, and I was just saying that I'd rather push for something simple and useful than the hypothetical perfect we would not be able to have. So just in case someone came afterwards being all 👎 on this :D

> I'm always doing it for the sake of what seems best for the project

Yep, we're all there :D
And I do appreciate your dedication, btw :)

> tl;dr: I'm totally fine with putting SO in the readme.

:+1:
(Agreed with the rest, and note that nothing is mutually exclusive, I believe we are going towards this direction)

xPaw:
So why not put this in SUPPORT.md straight away?

astorije:
> So why not put this in SUPPORT.md straight away?

Because I wanted this out of the way quickly, and didn't want to start debates (I didn't think it would be controversial, sorry!) but I guess it's a little too late for that 😅.

Anyway, I added a `SUPPORT.md` file. I didn't touch any of the content, just moved to this new file from the CONTRIBUTING one what was relevant. Happy to improve the wording/content later, but for now this will do for what this PR achieves.

PR #1380, opened 2017-07-28
Show notices and errors inline

### LINKED ISSUE #114: Show notices inline
I'm not sure if this is part of the IRC spec, but most clients generally show notices in-line (ie like a message). At the moment, they're displayed in the network tab which isn't very intuitive.

Alternatively, notices could be displayed in a modal/popup similar to IRCcloud

xPaw:
Fixes #114.

Display received notices and errors in currently active channel, if the actual target is the network lobby. Reloading the page will put them back into the lobby window.

### DIFF
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -27,11 +27,27 @@ socket.on("msg", function(data) {
 });
 
 function processReceivedMessage(data) {
-	const targetId = data.chan;
-	const target = "#chan-" + targetId;
-	const channel = chat.find(target);
-	const container = channel.find(".messages");
+	let targetId = data.chan;
+	let target = "#chan-" + targetId;
+	let channel = chat.find(target);
+	let sidebarTarget = sidebar.find("[data-target='" + target + "']");
 
+	// Display received notices and errors in currently active channel.
+	// Reloading the page will put them back into the lobby window.
+	if (data.msg.showInActive) {
+		const activeOnNetwork = sidebarTarget.parent().find(".active");
+
+		// We only want to put errors/notices in active channel if they arrive on the same network
+		if (activeOnNetwork.length > 0) {
+			targetId = data.chan = activeOnNetwork.data("id");
+
+			target = "#chan-" + targetId;
+			channel = chat.find(target);
+			sidebarTarget = sidebar.find("[data-target='" + target + "']");
+		}
+	}
+
+	const container = channel.find(".messages");
 	const activeChannelId = chat.find(".chan.active").data("id");
 
 	if (data.msg.type === "channel_list" || data.msg.type === "ban_list") {
@@ -65,7 +81,7 @@ function processReceivedMessage(data) {
 
 	// Clear unread/highlight counter if self-message
 	if (data.msg.self) {
-		sidebar.find("[data-target='" + target + "'] .badge").removeClass("highlight").empty();
+		sidebarTarget.find(".badge").removeClass("highlight").empty();
 	}
 
 	let messageLimit = 0;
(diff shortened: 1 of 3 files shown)


client/js/socket-events/msg.js
```diff
@@ -47,4 +58,15 @@ socket.on("msg", function(data) {
 			.find(".unread-marker")
 			.appendTo(container);
 	}
+
+	if ((data.msg.type === "message" || data.msg.type === "action") && chan.hasClass("channel")) {
```
xPaw: I moved this autocompletion ordering here as there is no need to apply this logic when rendering bulk messages (e.g. when opening the page), as the server now keeps track of messages. I also removed `notice` from here as notices always target query windows and lobbies, and neither of these have user list.

dgw:
:tada:

I assume that query windows will return `false` for `chan.hasClass('lobby')`? One of the bots I get notices from frequently is also useful to keep a PM window open to, so I'm guessing the current behavior of its notices going to that query window will not change with this PR.

Still :tada: overall though, since I can just close the query as a workaround.

xPaw:
@dgw Yes, it only applies the active window logic only if the intended target is lobby.

astorije:
> Reloading the page will put them back into the lobby window.

Any way to fix this? That would be pretty bad :/

xPaw:
@astorije, that's not exactly a bug. Otherwise this gets super tricky with multi client support.

christer88:
Is it please possible to differentiate which notices goes in active channel and which goes into server window? All the notices from IRC server when joining network (see screenshot below) is quite a lot of information which for most users are not important nor nice to have in active channel.

e.g. If notices from server when joining network -> send to server window. Else, send notices to active channel.

server_notices

astorije:
@xPaw, instead of showing errors inline, maybe we should have them shown at the top of the screen like error messages often do. Kinda how we currently show connection errors, except at the top instead of bottom.
What do you think?

astorije:
> @astorije, that's not exactly a bug. Otherwise this gets super tricky with multi client support.

Honestly, having things disappear on reload (well, moved somewhere else but the effect is the same) is just something we should try to avoid :/
What alternatives do we have?

client/js/socket-events/msg.js
```diff
@@ -27,11 +27,20 @@ socket.on("msg", function(data) {
...
-	const container = channel.find(".messages");
+	let targetId = data.chan;
+	let target = "#chan-" + targetId;
+	let channel = chat.find(target);
+
+	// Display received notices and errors in currently active channel,
+	// if the actual target is the network lobby.
+	// Reloading the page will put them back into the lobby window.
+	if (data.msg.showInActive) {
+		targetId = data.chan = chat.find(".active").data("id");
```
xPaw: I need to test what this code does if notice arrives on another network.

xPaw: Okay I fixed this, will push in a mo.

xPaw:
@astorije As @YaManicKill said, it's an unfortunate design of notices on IRC. We have to reset them to server window (well technically they're always there on the server) due to our support for multiple clients at once. It gets hairy either way.

As for your error design, how are you going to contain these errors within the network?

client/js/socket-events/msg.js
```diff
@@ -27,11 +27,27 @@ socket.on("msg", function(data) {
...
-	const container = channel.find(".messages");
+	let targetId = data.chan;
+	let target = "#chan-" + targetId;
+	let channel = chat.find(target);
+	let sidebarTarget = sidebar.find("[data-target='" + target + "']");
+
+	// Display received notices and errors in currently active channel.
+	// Reloading the page will put them back into the lobby window.
+	if (data.msg.showInActive) {
+		const activeOnNetwork = sidebarTarget.parent().find(".active");
```
YaManicKill: Oh goodness, this doesn't feel ideal, but I guess that's what we get for only storing the state in the dom  ¯\\\_(ツ)_/¯

xPaw: lounge frontend LUL

astorije:
> As for your error design, how are you going to contain these errors within the network?

I don't think we have to. If an error shows up on Freenode while you're browsing a channel on another network, the banner should say something like "Blabla happened on Freenode" or whatever.

The idea is that error banners would stay until closing them and persist when switching channels. There are many ways we could do that, and I can see a few paradigms that exist in other app that would apply to us. This needs experimenting though.

Issue #888, opened 2017-01-24
Channel assign with URL


daniel-hyuuga:
It is a little trick where I can give out TheLounge URLs with intended channels defined.

lounge.js (line 881)
-------------
```
        document.title = title;

        var t=window.location.href.split("?");

        if (t.length > 1) {
            if (t[1].indexOf(",") > -1) {
                t = t[1].split(",");
            }
            else {
                t = new Array(t[1]);
            }
            $("#connect\\:channels").val(t.join(" "));
        }
```

With this, you can give out urls like `http://chat.exampleirc.net/?#lobby,#help`
**EDIT**: Fixed issue where the javascript would add URL instead of channel if no channel specified in URL

YaManicKill:
Huh, I'm surprised this doesn't exist as an issue already.

First of all, we'd need a way to define the server as well, not just the chan (most people have multiple servers). This was mentioned in #574 as a good idea.

MaxLeiter:
Once 2.2.0 is released, you can do this in public mode as of #674

astorije:
Closing this as this will part of v2.2.0, which is just around the corner.

YaManicKill:
But this won't be the case in private mode, which is still a legitimate
feature.

Issue #666, opened 2016-10-02
Lounge crashes when user logs in from different system (LDAP enabled)


thisisdarshan:
Lounge crashes after the user logs in from a different system.  For the context, I have enabled LDAP for authentication.  It could be related the case (notice the first time user logged in with "Xyz" and the second time the user logged in with "xyz").  Here is the trace log -

```
2016-09-20 23:55:40 [INFO] User 'abc' loaded
2016-09-20 23:55:40 [INFO] User 'bcd' loaded
2016-09-20 23:59:59 [INFO] User 'Xyz' loaded
2016-09-21 16:31:56 [INFO] User 'xyz' loaded
Missing error handler on `socket`.
TypeError: Cannot read property 'replace' of undefined
    at ldapAuth (/bnc/codebase/src/server.js:264:19)
    at Socket.auth (/bnc/codebase/src/server.js:321:4)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at Socket.onevent (/bnc/codebase/node_modules/socket.io/lib/socket.js:335:8)
    at Socket.onpacket (/bnc/codebase/node_modules/socket.io/lib/socket.js:295:12)
    at Client.ondecoded (/bnc/codebase/node_modules/socket.io/lib/client.js:193:14)
    at Decoder.Emitter.emit (/bnc/codebase/node_modules/socket.io/node_modules/socket.io-parser/node_modules/component-emitter/index.js:134:20)
    at Decoder.add (/bnc/codebase/node_modules/socket.io/node_modules/socket.io-parser/index.js:247:12)
    at Client.ondata (/bnc/codebase/node_modules/socket.io/lib/client.js:175:18)
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at exports._errnoException (util.js:907:11)
    at WriteWrap.afterWrite (net.js:785:14)

npm ERR! Linux 3.10.0-327.28.3.el7.x86_64
npm ERR! argv "/usr/bin/node" "/usr/bin/npm" "start"
npm ERR! node v4.5.0
npm ERR! npm  v2.15.9
npm ERR! code ELIFECYCLE
npm ERR! thelounge@2.0.0-rc.1 start: `node index`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the thelounge@2.0.0-rc.1 start script 'node index'.
npm ERR! This is most likely a problem with the thelounge package,
npm ERR! not with npm itself.
[...]

PR #1383, opened 2017-07-30
More contrast for text colors

mestaritonttu:
Didn't check with W3C guidelines, but just darkened some relevant ones a bit.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -958,7 +958,7 @@ kbd {
 }
 
 #chat .time {
-	color: #ddd;
+	color: #777;
 	padding-left: 10px;
 }
 
@@ -1038,11 +1038,11 @@ kbd {
 }
 
 #chat .self .text {
-	color: #999;
+	color: #444;
 }
 
 #chat .msg.channel_list_loading .text {
-	color: #999;
+	color: #444;
 	font-style: italic;
 	padding-left: 20px;
 }
@@ -1109,7 +1109,7 @@ kbd {
 #chat .quit .content,
 #chat .topic .content,
 #chat .topic_set_by .content {
-	color: #999;
+	color: #444;
 }
 
 #chat .action .from,
@@ -1202,7 +1202,7 @@ kbd {
 }
 
 #chat .toggle-content .body {
-	color: #999;
+	color: #444;
 }
 
 #chat .toggle-content.show {
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -207,6 +207,17 @@ body {
 	color: #ccc;
 }
 
+#chat .join .content,
+#chat .kick .content,
+#chat .mode .content,
+#chat .nick .content,
+#chat .part .content,
+#chat .quit .content,
+#chat .topic .content,
+#chat .topic_set_by .content {
+	color: #999;
+}
+
 /* Embeds */
 #chat .toggle-content,
 #chat .toggle-button {
--- a/client/themes/zenburn.css
+++ b/client/themes/zenburn.css
@@ -233,6 +233,17 @@ body {
 	color: #ccc;
 }
 
+#chat .join .content,
+#chat .kick .content,
+#chat .mode .content,
+#chat .nick .content,
+#chat .part .content,
+#chat .quit .content,
+#chat .topic .content,
+#chat .topic_set_by .content {
+	color: #999;
+}
+
 /* Previews */
 
 #chat .toggle-content,


MaxLeiter:
Have you tested this on all default themes?

mestaritonttu:
> Have you tested this on all default themes?

Good point. I now tested with the themes on the demo instance and noticed Morning and Zenburn need to be tweaked so they override the color in one place like so:

```
#chat .join .content,
#chat .kick .content,
#chat .mode .content,
#chat .nick .content,
#chat .part .content,
#chat .quit .content,
#chat .topic .content,
#chat .topic_set_by .content {
	color: #999;
}
```

client/css/style.css
```diff
@@ -958,7 +958,7 @@ kbd {
 }
 
 #chat .time {
-	color: #ddd;
+	color: #777;
```
astorije: Before | After
--- | ---
screen shot 2017-08-12 at 21 17 34 | screen shot 2017-08-12 at 21 18 43

I totally understand that this is not contrasted enough, but with `#777` there is pretty much no difference with normal text. Since time appears with every. single. message, that *After* view is very crowded and harsh...

I would leave them as is, until someday maybe if we add a "high contrast mode" option.

client/css/style.css
```diff
@@ -1038,11 +1038,11 @@ kbd {
 }
 
 #chat .self .text {
-	color: #999;
+	color: #444;
```
astorije: Before | After
--- | ---
screen shot 2017-08-12 at 21 21 42 | screen shot 2017-08-12 at 21 21 51

I honestly can't tell the difference between colors in the second screenshot :/
I understand `#999` isn't contrasted enough, and this is a pretty important message to read, so maybe `#888` or `#777` instead? (second one passes AA levels)

client/themes/morning.css
```diff
@@ -207,6 +207,17 @@ body {
...
+#chat .join .content,
+#chat .kick .content,
+#chat .mode .content,
+#chat .nick .content,
+#chat .part .content,
+#chat .quit .content,
+#chat .topic .content,
+#chat .topic_set_by .content {
+	color: #999;
+}
```
astorije: For future readers, this "resets" colors. There is effectively no change on this theme between master and this PR, but not adding this would make them much darker.

client/themes/zenburn.css
```diff
@@ -233,6 +233,17 @@ body {
...
+#chat .join .content,
+#chat .kick .content,
+#chat .mode .content,
+#chat .nick .content,
+#chat .part .content,
+#chat .quit .content,
+#chat .topic .content,
+#chat .topic_set_by .content {
+	color: #999;
+}
```
astorije: Same as above.

client/css/style.css
```diff
@@ -1202,7 +1202,7 @@ kbd {
 }
 
 #chat .toggle-content .body {
-	color: #999;
+	color: #444;
```
astorije: Same comment than on `self`, it's a pretty important content, but it's too dark as is. I'd go with `#666`, which passes AA levels.

client/css/style.css
```diff
@@ -1109,7 +1109,7 @@ kbd {
 #chat .quit .content,
 #chat .topic .content,
 #chat .topic_set_by .content {
-	color: #999;
+	color: #444;
```
astorije: This is like `time`: it doesn't add a lot of value, and too dark it would clutter the place pretty hard. Even #777 is too dark, IMO, but I would be okay with `#777` since they will be much less cluttered after #759.

client/css/style.css
```diff
@@ -1038,11 +1038,11 @@ kbd {
 }
 
 #chat .self .text {
-	color: #999;
+	color: #444;
 }
 
 #chat .msg.channel_list_loading .text {
-	color: #999;
+	color: #444;
```
astorije: That's fine, I didn't even see that one was bright, good catch.

Turns out, `#chat table.channel-list td` (the content of the table when loaded) is `#555` at the moment. You could make it `#444` to match your change here, or even better IMO, remove the color definition here and in `#chat table.channel-list td` so they both match "normal" color, i.e. `#222`.
What do you think?

mestaritonttu:
Well, change them as you please. I would never use differing contrast to communicate a difference in a chat context, but that is a personal preference.

xPaw:
Closing this as author has no intent of changing it.

PR #1387, opened 2017-07-31
Extend fuzzy search in autocomplete to all strategies

### LINKED ISSUE #1086: Use fuzzy-matching for nick/channel/emoji/command autocompletion

yashsriv:
Fixes #1086
Just building upon the work already done in #1334

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -120,8 +120,18 @@ $(function() {
 		match: /\x03(\d{0,2}|[A-Za-z ]{0,10})$/,
 		search(term, callback) {
 			term = term.toLowerCase();
+
 			const matchingColorCodes = constants.colorCodeMap
-				.filter((i) => i[0].startsWith(term) || i[1].toLowerCase().startsWith(term));
+				.filter((i) => fuzzy.test(term, i[0]) || fuzzy.test(term, i[1]))
+				.map((i) => {
+					if (fuzzy.test(term, i[1])) {
+						return [i[0], fuzzy.match(term, i[1], {
+							pre: "<b>",
+							post: "</b>"
+						}).rendered];
+					}
+					return i;
+				});
 
 			callback(matchingColorCodes);
 		},
@@ -140,7 +150,16 @@ $(function() {
 		search(term, callback, match) {
 			term = term.toLowerCase();
 			const matchingColorCodes = constants.colorCodeMap
-				.filter((i) => i[0].startsWith(term) || i[1].toLowerCase().startsWith(term))
+				.filter((i) => fuzzy.test(term, i[0]) || fuzzy.test(term, i[1]))
+				.map((pair) => {
+					if (fuzzy.test(term, pair[1])) {
+						return [pair[0], fuzzy.match(term, pair[1], {
+							pre: "<b>",
+							post: "</b>"
+						}).rendered];
+					}
+					return pair;
+				})
 				.map((pair) => pair.concat(match[1])); // Needed to pass fg color to `template`...
 
 			callback(matchingColorCodes);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -71,9 +71,10 @@ $(function() {
 		search(term, callback) {
 			term = term.slice(1);
 			if (term[0] === "@") {
-				callback(completeNicks(term.slice(1)).map((val) => "@" + val));
+				callback(completeNicks(term.slice(1), true)
+					.map((val) => ["@" + val[0], "@" + val[1]]));
 			} else {
-				callback(completeNicks(term));
+				callback(completeNicks(term, true));
 			}
 		},
 		template([string, ]) {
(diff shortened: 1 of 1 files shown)


yashsriv:
I'll add fuzzy matching for colors too if this structure is ok.

astorije:
So, I am 👍 💯 🎉 on this feature overall, but I am wondering if there is a way we could define a fuzzy-ready generic strategy, then our strategies could inherit from it. Won't be a blocker for this, but just something to think about :)

astorije:
@yashsriv, ping? :)

yashsriv:
Sadly I couldn't even do the color autocomplete generically because of its requirements :/

But yeah now fuzzy complete works everywhere

astorije:
@yashsriv, mind rebasing and fixing conflicts on this? I can help if you're confused.

client/js/lounge.js
```diff
@@ -932,13 +955,23 @@ $(function() {
...
+		// Lobbies and private chats do not have an user list
+		if (!users.length) {
+			return [];
+		}
+
+		const words = users.data("nicks");
+
+		return fuzzyGrep(word, words);
+	}
+
```
astorije: Actually, instead of duplicating this function, make it work for both with either of these solutions:

- Add a `isFuzzy` boolean to the signature, and if `true` return `fuzzyGrep(...)`, otherwise return `$.grep(...)`.
- Make this functional-like (think of `.map()` or `.filter`): add a `callback` argument and return `callback(words)` so that functions calling this can do `completeNicks(word, (words) => fuzzyGrep(word, words));` (or `$.grep` equivalent).

Or those are overkill and there is a much simpler solution, I don't know, it's pretty late here :)
@xPaw, any preference?

client/js/lounge.js
```diff
@@ -916,29 +927,42 @@ $(function() {
...
+		);
+		return results.map((el) => [el.string, el.original]);
+	}
 
-		// Lobbies and private chats do not have an user list
-		if (!users.length) {
-			return [];
-		}
+	function completeNicks(isFuzzy) {
+		return function(word) {
```
xPaw: Why are you returning a function?

client/js/lounge.js
```diff
@@ -275,7 +286,7 @@ $(function() {
 
 			chat.find(".chan.active .chat").trigger("msg.sticky"); // fix growing
 		})
-		.tab(completeNicks, {hint: false})
+		.tab(completeNicks(false), {hint: false})
```
xPaw: .. probably because of this, but it feels like poor code.

astorije: What poor code about returning a function? That's like functional programming 101.

xPaw: It's unnecessary complexity and the way it's used does not warrant a "dynamic" function.

astorije: Come on, there is no unnecessary complexity in this whatsoever.

- @yashsriv's solution: a function that returns a function, called three times with `true`/`false` as an argument.
- Your solution: that inner function, called three times with `(word) => completeNicks(word, false/true)`

i.e. the exact same thing, just with the argument is shifted. At this point, that's bikeshedding.

> the way it's used does not warrant a "dynamic" function.

What does that even mean?

xPaw: > Your solution: that inner function, called three times with (word) => completeNicks(word, false/true)

It's just not about function calls, it's about being clear what the function does. It is a lot cleaner code when the function just takes 2 arguments instead of 1 argument and return a function that takes a secondary argument (there's literally no reason to do this, we don't do this anywhere else in the entire codebase).

astorije: Whateveeeeer. Let's just make this happen, please.
@xPaw, feel free to commit your change on this PR.

astorije:
@yashsriv, sorry to do that to you, but because of #1356, there is now a conflict on this PR. Could you resolve it please? Also, if you could apply @xPaw's suggestion, we should be all set.
Thanks!

client/js/lounge.js
```diff
@@ -916,29 +927,42 @@ $(function() {
...
-			words,
-			(w) => !w.toLowerCase().indexOf(word.toLowerCase())
-		);
+			const words = users.data("nicks");
+			if (isFuzzy) {
+				return fuzzyGrep(word, words);
+			}
+			return $.grep(
+				words,
+				(w) => !w.toLowerCase().indexOf(word.toLowerCase())
```
xPaw: I just noticed that `word` is lowercased every time in the callback, move `word = word.toLowerCase()` before the `$.grep` call for a performance boost.

xPaw:
Thanks @yashsriv!

Next refactoring step would be moving all the autocompletion logic into it's own file outside of `lounge.js`, do you want to take this task?

astorije:
Hey @yashsriv, we have free sticker packs for our contributors now!

PR #1403, opened 2017-08-09
Change history button text while loading

xPaw:
https://github.com/thelounge/lounge/pull/1318#issuecomment-319295408


bews:
Umm, I think the expected solution to https://github.com/thelounge/lounge/pull/1318#issuecomment-319295408 is to add loading animation in CSS, not this hack.

CSS animation can be also used in picture previews and thumbnails for the reasons I pointed in my closed PR, iirc (in short: browser cache).

PR #1406, opened 2017-08-11
Correctly dereference stored images when leaving channels

### LINKED ISSUE #1404: Stored images are not dereferenced when leaving a channel
... or closing network, or closing the demo page.

xPaw:
Fixes #1404. cc @MaxLeiter

### DIFF
--- a/src/plugins/storage.js
+++ b/src/plugins/storage.js
@@ -17,12 +17,6 @@ class Storage {
 	}
 
 	dereference(url) {
-		// If maxHistory is 0, image would be dereferenced before client had a chance to retrieve it,
-		// so for now, just don't implement dereferencing for this edge case.
-		if (helper.maxHistory === 0) {
-			return;
-		}
-
 		const references = (this.references.get(url) || 0) - 1;
 
 		if (references < 0) {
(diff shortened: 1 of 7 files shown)


src/plugins/storage.js
```diff
@@ -17,12 +17,6 @@ class Storage {
 	}
 
 	dereference(url) {
-		// If maxHistory is 0, image would be dereferenced before client had a chance to retrieve it,
-		// so for now, just don't implement dereferencing for this edge case.
-		if (helper.maxHistory === 0) {
```
xPaw: There was a bug here too, should have been `helper.config.maxHistory`.

astorije: Just curious, why did you move it out of the function to somewhere else?

xPaw: I didn't feel like adding a `force` option to avoid this, so I just moved it instead.

astorije: Sorry, what I meant was: why is this needed in one case and not every time you call `dereference`?

xPaw: Other cases are leaving channel/network/closing public page, which all destroy the channel and messages completely.

PR #1409, opened 2017-08-13
Remove 10 second interval to trim buffer

### LINKED ISSUE #96: "show more" is constantly checked on an interval

xPaw:
Fixes #96. Split from #1041.

### DIFF
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -8,8 +8,11 @@ const templates = require("../../views");
 
 socket.on("msg", function(data) {
 	const msg = render.buildChatMessage(data);
-	const target = "#chan-" + data.chan;
-	const container = chat.find(target + " .messages");
+	const target = data.chan;
+	const channel = chat.find("#chan-" + target);
+	const container = channel.find(".messages");
+
+	const activeChannelId = chat.find(".chan.active").data("id");
 
 	if (data.msg.type === "channel_list" || data.msg.type === "ban_list") {
 		$(container).empty();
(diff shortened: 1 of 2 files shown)


client/js/socket-events/msg.js
```diff
@@ -47,4 +50,17 @@ socket.on("msg", function(data) {
...
+		channel.find(".show-more").addClass("show");
+
+		// Remove date-seperators that would otherwise
+		// be "stuck" at the top of the channel
+		channel.find(".date-marker-container").each(function() {
+			if ($(this).next().hasClass("date-marker-container")) {
+				$(this).remove();
+			}
+		});
+	}
```
astorije: Is it worth putting this in a debounce function so it doesn't fire more than once every N seconds (10?)? I'm concerned this could be called a lot in very active channels.

xPaw: Perhaps.

astorije: Do you wanna look into it?

astorije: From the channel:

```
<xPaw>     astorije, I remember testing the message trimmer in my queue PR on twitch channels
<astorije> So you're saying no debounce needed, right?
<xPaw>     it produced like multiple new message per seconds, and while debugging it, I saw nothing that pointed towards this code
<xPaw>     so yes, shouldnt be needed
```

PR #1411, opened 2017-08-13
Refactor authentication flow

xPaw:
This PR is whitespace heavy so review with `?w=1`.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -300,13 +300,13 @@ Client.prototype.updateSession = function(token, ip, request) {
 	const agent = UAParser(request.headers["user-agent"] || "");
 	let friendlyAgent = "";
 
-	if (agent.browser.name.length) {
+	if (agent.browser.name) {
 		friendlyAgent = `${agent.browser.name} ${agent.browser.major}`;
 	} else {
 		friendlyAgent = "Unknown browser";
 	}
 
-	if (agent.os.name.length) {
+	if (agent.os.name) {
 		friendlyAgent += ` on ${agent.os.name} ${agent.os.version}`;
 	}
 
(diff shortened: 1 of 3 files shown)


src/server.js
```diff
@@ -470,3 +466,13 @@ function auth(data) {
 		localAuth(client, data.user, data.password, authCallback);
 	}
 }
+
+function reverseDnsLookup(ip, callback) {
```
xPaw: Rewrote this function to be an independent helper. This improves authentication flow (e.g. remember me in webirc most likely did not work before).

src/server.js
```diff
@@ -408,18 +397,25 @@ function ldapAuth(client, user, password, callback) {
 	});
 }
 
-function auth(data) {
+function performAuthentication(data) {
 	const socket = this;
 	let client;
 
+	const finalInit = () => initializeClient(socket, client, !!data.remember, data.token || null);
```
xPaw: I am now passing `token` so that we have a session-like reference to it. This will be required for push notifs and other stuff later.

src/server.js
```diff
@@ -100,9 +100,10 @@ in ${config.public ? "public" : "private"} mode`);
 
 	sockets.on("connect", function(socket) {
 		if (config.public) {
-			auth.call(socket, {});
+			performAuthentication.call(socket, {});
 		} else {
-			init(socket);
+			socket.emit("auth", {success: true});
```
xPaw: Moved these 2 lines here so that we don't have a big if/else in `init`.

src/client.js
```diff
@@ -300,13 +300,13 @@ Client.prototype.updateSession = function(token, ip, request) {
 	const agent = UAParser(request.headers["user-agent"] || "");
 	let friendlyAgent = "";
 
-	if (agent.browser.name.length) {
```
xPaw: Fixes crash if UA is missing.

PR #1416, opened 2017-08-15
Use .lounge_config to help distribution packages handle config paths right

xPaw:
This will make `lounge` command execute with correct config path consistently.


astorije:
Maybe it was not optimal to call it `.lounge_config` vs. `LOUNGE_HOME`, (should have been `.lounge_home` maybe?) but oh well, I noticed that afterwards :)
What do you think @xPaw?

astorije:
Actually, I'm totally changing my mind on this: definitely makes more sense to use `.lounge_home`. Will send a PR any minute now.

PR #1419, opened 2017-08-16
Remove os.homedir() polyfill

xPaw:


src/helper.js
```diff
@@ -116,17 +116,10 @@ function expandHome(shortenedPath) {
...
-	if (os.homedir) {
-		home = os.homedir();
-	}
-
-	if (!home) {
-		home = process.env.HOME || "";
-	}
 
+	let home = os.homedir();
 	home = home.replace("$", "$$$$");
```
astorije: `const home = os.homedir().replace("$", "$$$$");` instead? :man_shrugging:

xPaw: You can do it if you want to.

astorije: Done.

PR #1421, opened 2017-08-16
Fix issue with falsey user options getting overridden by truthy defaults

### LINKED ISSUE #1420: "true" default options can override user settings
How to reproduce:

* Go into options
* See that "Auto-expand websites" is checked
* Uncheck
* Refresh page
* Go into options

What should happen: auto-expand websites should still be unchecked
What actually happens: auto-expand websites is checked again

It's due to [this change](https://github.com/thelounge/lounge/commit/b508783101dfc979e92221a884f87812b27633b7#diff-03ed49e301a4e588ec406bc5aace2a02R31) where it will only set the user option if it is truthy, and so if the default is truthy and the user's option is falsy, it will use the default.

We should have an undefined check instead.

YaManicKill:
Fixes #1420

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -25,13 +25,15 @@ const options = {
 	thumbnails: true,
 	userStyles: userStyles.text(),
 };
-const userOptions = JSON.parse(storage.get("settings")) || {};
+let userOptions = JSON.parse(storage.get("settings")) || {};
 
 for (const key in options) {
-	if (userOptions[key]) {
+	if (userOptions[key] !== undefined) {
 		options[key] = userOptions[key];
 	}
 }
+
+userOptions = null;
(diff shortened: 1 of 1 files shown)


client/js/options.js
```diff
@@ -25,14 +25,16 @@ const options = {
...
+let userOptions = JSON.parse(storage.get("settings")) || {};
 
 for (const key in options) {
-	if (userOptions[key]) {
+	if (userOptions[key] !== undefined) {
 		options[key] = userOptions[key];
 	}
 }
 
+userOptions = null;
```
astorije: I get the change above, but why this one?

xPaw: Free memory.

Issue #1432, opened 2017-08-18
Switching between condensed/expanded options for status message does nothing before reloading the app


astorije:
This means that on mobile, users will need to close the application and open it again. On desktop, they will need to figure out that the option does work but only after refreshing.

Ideal solution: re-render the channels when switching the option.

Non-ideal solution until the above can be done: refreshing the page when changing this option. Problem is it's disruptive for the user as the settings don't show up right after reload. This would be addressed by #1283.
This solution is still disruptive even after #1283 makes it, but I'd argue times better than nothing, and until the ideal solution can be done, it's an acceptable temporary solution IMO 😃

PR #1437, opened 2017-08-19
Cleanup condensed appendMessage

xPaw:
I have not yet tested this, but it should be logically the same. The PR is open for @YaManicKill to review the code.

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -7,7 +7,7 @@ const renderPreview = require("./renderPreview");
 const utils = require("./utils");
 const sorting = require("./sorting");
 const constants = require("./constants");
-const condense = require("./condensed");
+const condensed = require("./condensed");
 
 const chat = $("#chat");
 const sidebar = $("#sidebar");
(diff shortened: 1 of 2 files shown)


client/js/render.js
```diff
@@ -33,23 +33,31 @@ function buildChannelMessages(data) {
...
+		container.append(msg);
+		return;
+	}
+
+	const lastChild = container.children("div.msg").last();
+
+	if (lastChild && $(lastChild).hasClass("condensed")) {
+		lastChild.append(msg);
+		condensed.updateText(lastChild, [messageType]);
+	} else if (lastChild && $(lastChild).is(constants.condensedTypesQuery)) {
```
astorije: Since this is used only once, remove the export and use `"." + constants.condensedTypes.join(", .")` directly here. I'm not a fan of `constants` exporting this.

client/js/render.js
```diff
@@ -33,23 +33,31 @@ function buildChannelMessages(data) {
...
-		} else if (lastChild && $(lastChild).is(condensedTypesClasses) && options.statusMessages === "condensed") {
-			var condensed = buildChatMessage({msg: {type: "condensed", time: msg.attr("data-time"), previews: []}, chan: chan});
-			condensed.append(lastChild);
-			condensed.append(msg);
-			container.append(condensed);
-			condense.updateText(condensed, [messageType, lastChild.attr("data-type")]);
-		} else {
-			container.append(msg);
-		}
+	// TODO: To fix #1432, statusMessage option should entirely be implemented in CSS
```
astorije: 👍

client/js/render.js
xPaw: There's no reason constantly imploding an array into string every time a message is processed.

client/js/render.js
```diff
@@ -33,23 +33,31 @@ function buildChannelMessages(data) {
...
+				type: "condensed",
+				time: msg.attr("data-time"),
+				previews: []
+			}
+		});
+
+		condensed.updateText(newCondensed, [messageType, lastChild.attr("data-type")]);
+		container.append(newCondensed);
+		newCondensed.append(lastChild);
+		newCondensed.append(msg);
```
astorije: Any reason why you are changing the order of this?

xPaw: 1. Condensed text will update in memory before it's inserted into DOM.
2. Last child will be moved into condensed container without being removed from DOM first.

client/js/render.js
```diff
@@ -7,7 +7,7 @@ const renderPreview = require("./renderPreview");
 const utils = require("./utils");
 const sorting = require("./sorting");
 const constants = require("./constants");
-const condense = require("./condensed");
+const condensed = require("./condensed");
```
YaManicKill: I think this was coming from when I couldn't decide whether to say "condense" or "condensed".

client/js/render.js
YaManicKill: Yeah, I guess that could simplify the logic here and fix #1432 as well.

PR #1438, opened 2017-08-21
Add documentation for LOUNGE_HOME environment variable in the CLI helper

astorije:
Couldn't do this in a nicer way, see https://github.com/tj/commander.js/issues/682.

screen shot 2017-08-21 at 01 48 29

(`~` in default not expanded on purpose)

### DIFF
--- a/src/command-line/utils.js
+++ b/src/command-line/utils.js
@@ -1,0 +1,18 @@
+"use strict";
+
+const colors = require("colors/safe");
+
+class Utils {
+	static extraHelp() {
+		[
+			"",
+			"",
+			"  Environment variable:",
+			"",
+			`    LOUNGE_HOME   Path for all configuration files and folders. Defaults to ${colors.green("~/.lounge")}.`,
+			"",
+		].forEach((e) => console.log(e));
+	}
+}
+
+module.exports = Utils;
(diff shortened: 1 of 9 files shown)


astorije:
Actually, the default value is wrong, it should be what's in `.lounge_home`. Will fix.

src/command-line/utils.js
```diff
@@ -0,0 +1,38 @@
...
+class Utils {
+	static extraHelp() {
+		[
+			"",
+			"",
+			"  Environment variable:",
+			"",
+			`    LOUNGE_HOME   Path for all configuration files and folders. Defaults to ${colors.green(Utils.defaultLoungeHome())}.`,
+			"",
+		].forEach((e) => console.log(e));
```
YaManicKill: Do we really have to console log manually? Ew.

astorije: Yes, unfortunately, see their recommendation in https://github.com/tj/commander.js/issues/682 :(

astorije:
> I feel like we should have a way to not have to add the .on to each file

I tried several ways, and it looks like `.on("--help", ...)` is a per command thing. Which makes sense, in a way, not all commands will have the same helper, but if there was a way to define a global one, that'd be good, except there isn't... :/

PR #1442, opened 2017-08-23
Fix time placeholder showing on mobile format for condensed messages

astorije:
/Cc @MaxLeiter

As an extra, make sure time placeholder cannot be selected anymore, ~and do not do an extra call to `tz` helper when time is not relevant/displayed~.

I tried to entirely remove `time`/`from` blocks, but because of flexbox layout, that broke message alignment...

Before | After
--- | ---
screen shot 2017-08-22 at 20 12 38 | screen shot 2017-08-22 at 20 13 58
screen shot 2017-08-22 at 20 13 04 | screen shot 2017-08-22 at 20 13 45
screen shot 2017-08-22 at 20 13 15 | screen shot 2017-08-22 at 20 13 34

### DIFF
--- a/client/views/msg_condensed.tpl
+++ b/client/views/msg_condensed.tpl
@@ -1,5 +1,5 @@
-<div class="msg {{type}} closed" data-time="{{time}}">
-	<span class="time hide-text">{{tz time}}</span>
+<div class="msg condensed closed" data-time="{{time}}">
+	<span class="time">{{tz time}}</span>
 	<span class="from"></span>
 	<span class="content">
 		<span class="condensed-text"></span>
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -837,6 +837,10 @@ kbd {
 	display: none;
 }
 
+#chat .condensed .time {
+	visibility: hidden;
+}
+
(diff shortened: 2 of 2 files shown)


MaxLeiter:
👍

client/views/msg_condensed.tpl
```diff
@@ -1,5 +1,5 @@
-<div class="msg {{type}} closed" data-time="{{time}}">
-	<span class="time hide-text">{{tz time}}</span>
+<div class="msg condensed closed">
+	<span class="time">00:00</span>
```
xPaw: This will break layout with seconds enabled

astorije: Yes, I'm stupid. Goddammit, I was so hoping this was the ~right~ _better_ thing... So should I just leave actual time as is, for now? :/

xPaw: Pretty much

astorije: :rage4:

astorije: Alright, fixed. Well, at least you can't select it by accident anymore...

PR #1443, opened 2017-08-23
Improve CLI user management

astorije:
- **Private server can now run when there are no users**
- Add a couple warning to help users understand what's going on
- Make sure server is running before loading users to avoid some potentially weird race conditions, and also to produce a nice output (otherwise it's all mangled)
- **Make The Lounge private by default**
- Fix a couple bugs

This is meant to help #1350. I'll be happy to fix conflicts in #1350 after this PR lands if that helps.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -8,9 +8,9 @@ module.exports = {
 	// Set to 'false' to enable users.
 	//
 	// @type     boolean
-	// @default  true
+	// @default  false
 	//
-	public: true,
+	public: false,
 
 	//
 	// IP address or hostname for the web server to listen on.
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -18,7 +18,7 @@ ClientManager.prototype.init = function(identHandler, sockets) {
 	this.identHandler = identHandler;
 	this.webPush = new WebPush();
 
-	if (!Helper.config.public) {
+	if (!Helper.config.public && !Helper.config.ldap.enable) {
 		if ("autoload" in Helper.config) {
 			log.warn(`Autoloading users is now always enabled. Please remove the ${colors.yellow("autoload")} option from your configuration file.`);
 		}
@@ -32,11 +32,22 @@ ClientManager.prototype.findClient = function(name) {
 };
 
 ClientManager.prototype.autoloadUsers = function() {
-	this.getUsers().forEach((name) => this.loadUser(name));
+	const users = this.getUsers();
+	const noUsersWarning = `There are currently no users. Create one with ${colors.bold("lounge add <name>")}.`;
+
+	if (!users.length) {
+		log.info(noUsersWarning);
+	}
+
+	users.forEach((name) => this.loadUser(name));
 
 	fs.watch(Helper.USERS_PATH, _.debounce(() => {
 		const loaded = this.clients.map((c) => c.name);
 		const updatedUsers = this.getUsers();
+
+		if (!updatedUsers.length) {
+			log.info(noUsersWarning);
+		}
 
 		// New users created since last time users were loaded
 		_.difference(updatedUsers, loaded).forEach((name) => this.loadUser(name));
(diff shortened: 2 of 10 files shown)


src/clientManager.js
```diff
@@ -18,7 +18,7 @@ ClientManager.prototype.init = function(identHandler, sockets) {
 	this.identHandler = identHandler;
 	this.webPush = new WebPush();
 
-	if (!Helper.config.public) {
+	if (!Helper.config.public && !Helper.config.ldap.enable) {
```
xPaw: Any reason to add this?

astorije: Ah right, forgot to add comment for this. Reason is because of removing this https://github.com/thelounge/lounge/pull/1443/files#diff-e23e6a2fa48b03eef4ac89b3a4656f5aL29

So I figured LDAP just takes care of user management. Am I incorrect? Happy to remove, LDAP stuff is a black box to me anyway.

astorije:
Forgot to mention, but I recommend reviewing with [`?w=1`](https://github.com/thelounge/lounge/pull/1443/files?w=1).

src/clientManager.js
```diff
@@ -32,12 +32,29 @@ ClientManager.prototype.findClient = function(name) {
...
+	const users = this.getUsers();
+	const noUsersWarning = `There are currently no users. Create one with ${colors.bold("lounge add <name>")}.`;
+
+	// There was an error, already logged, but we have to crash the server as
+	// user directory could not be accessed
+	if (users === undefined) {
+		process.exit(1);
+	}
+
+	if (!users.length) {
```
YaManicKill: Can we specify `=== 0` instead of relying on falseyness. We've had a couple of bugs based on falseyness and I think it makes it more clear.

src/clientManager.js
```diff
@@ -32,12 +32,29 @@ ClientManager.prototype.findClient = function(name) {
...
+		log.warn(noUsersWarning);
+	}
+
+	users.forEach((name) => this.loadUser(name));
 
 	fs.watch(Helper.USERS_PATH, _.debounce(() => {
 		const loaded = this.clients.map((c) => c.name);
 		const updatedUsers = this.getUsers();
 
+		if (!updatedUsers.length) {
```
YaManicKill: What's the point in logging whenever we get an updated file? Surely the 1 message on startup is enough, no?

defaults/config.js
```diff
@@ -8,9 +8,9 @@ module.exports = {
 	// Set to 'false' to enable users.
 	//
 	// @type     boolean
-	// @default  true
+	// @default  false
```
YaManicKill: :+1:

src/clientManager.js
astorije: Will do.

src/clientManager.js
astorije: I figured it couldn't hurt. This would warn if something happened that led The Lounge to not have any users anymore, which is pretty useful when dealing with FS snafus. In particular, it logs the time this happened, which could be very useful.

I totally get what you're saying, but it's free for us to warn there (including code maintenance) and it adds some value in case something goes wrong, so I think we can afford leaving it here. What do you think? :)

YaManicKill: I guess so. Would it be worth moving it to a debug (or at least info) rather than warn in that case? I can see the value it would add.

xPaw: I have no strong opinion. Info might fit though.

astorije: Alright, info is fine. Made it info consistently to help admins search through their logs or something.

src/clientManager.js
astorije: Argh, I'm a 🍑, completely forgot to do that before merging, sorry! 😩

YaManicKill: :hankey:

src/clientManager.js
xPaw: This change broke loading existing ldap users on server restart.

It should disable fs.watch but still load existing json files from disk.

PR #1444, opened 2017-08-23
Do not unintentionally send incorrect messages from history

xPaw:
Fixes the bug @astorije found.

### DIFF
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -7,7 +7,6 @@ const chat = $("#chat");
 const templates = require("../../views");
 
 socket.on("more", function(data) {
-	const documentFragment = render.buildChannelMessages(data);
 	const chan = chat
 		.find("#chan-" + data.chan)
 		.find(".messages");
@@ -15,6 +14,12 @@ socket.on("more", function(data) {
 	// get the scrollable wrapper around messages
 	const scrollable = chan.closest(".chat");
 	const heightOld = chan.height();
+
+	// If there are no more messages to show, just hide the button and do nothing else
+	if (!data.messages.length) {
+		scrollable.find(".show-more").removeClass("show");
+		return;
+	}
 
 	// Remove the date-change marker we put at the top, because it may
 	// not actually be a date change now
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -403,14 +403,29 @@ Client.prototype.inputLine = function(data) {
...
-	var chan = target.chan;
-	var index = chan.messages.findIndex((val) => val.id === data.lastId);
-	var messages = chan.messages.slice(Math.max(0, index - 100), index);
+
+	const chan = target.chan;
+	const index = chan.messages.findIndex((val) => val.id === data.lastId);
+
+	// If we don't find the requested id, or it's first message we have
+	// Send an empty message array
+	if (index < 1) {
```
YaManicKill: Oh duh, I didn't even think about that possibility. Good catch.

client/js/socket-events/more.js
```diff
@@ -16,6 +15,12 @@ socket.on("more", function(data) {
 	const scrollable = chan.closest(".chat");
 	const heightOld = chan.height();
 
+	// If there are no more messages to show, just hide the button and do nothing else
+	if (!data.messages.length) {
```
YaManicKill: Do we need to take into consideration when we get more messages and previous ones will be pushed off the client?

src/client.js
```diff
@@ -403,14 +403,29 @@ Client.prototype.inputLine = function(data) {
...
+		client.emit("more", {
+			chan: chan.id,
+			messages: []
+		});
+
+		return;
+	}
+
+	const messages = chan.messages.slice(Math.max(0, index - 100), index);
```
YaManicKill: Could we simplify this code by just doing

`const messages = index < 1 ? [] : chan.messages.slice(Math.max(0, index - 100), index);`

?

client/js/socket-events/more.js
xPaw: What do you mean?

src/client.js
xPaw: Sure.

client/js/socket-events/more.js
YaManicKill: Well, if we have 100 messages in a channel, and we receive a new message, the top messages gets removed from the client, doesn't it?

xPaw: Messages in active channel do not get trimmed.

YaManicKill: Ah right, then no :-) cool

PR #1396, opened 2017-08-06
Make emoji bigger

easymac:
Increases the size of emoji by 40% without displacing the surrounding text. Discord, Slack, and Telegram provide a similar functionality to make emoji a bit more legible.

Before: Before preview

After: After preview

(Note: black text is not a bug, just happened because I took the screenshot from different simultaneously running clients.)

This is my first time contributing to The Lounge so I expect I'll have some fixes to make.

PR #1446, opened 2017-08-23
Parse emoji to make them bigger

xPaw:
Takes over #1396 since the author disappeared.

I opted into using a separate package to get the emoji regex. I also made emojis same bigger size in autocompletion dropdown.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1619,6 +1619,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	color: #333;
 	margin-top: 6px;
 	margin-bottom: 6px;
+	line-height: 1.4;
 	transition: background-color 0.2s;
 }
 
@@ -1643,11 +1644,16 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	opacity: 1;
 }
 
-.textcomplete-item .emoji {
-	margin-right: 8px;
-	width: 16px;
-	text-align: center;
+.emoji {
 	display: inline-block;
+	font-size: 1.4em;
+	vertical-align: bottom;
+	line-height: 1;
+}
+
+.textcomplete-item .emoji {
+	width: 32px;
+	text-align: center;
 }
 
 .textcomplete-item .irc-bg {
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -4,6 +4,7 @@ const Handlebars = require("handlebars/runtime");
 const parseStyle = require("./ircmessageparser/parseStyle");
 const findChannels = require("./ircmessageparser/findChannels");
 const findLinks = require("./ircmessageparser/findLinks");
+const findEmoji = require("./ircmessageparser/findEmoji");
 const merge = require("./ircmessageparser/merge");
 
 // Create an HTML `span` with styling information for a given fragment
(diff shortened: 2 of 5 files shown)


client/css/style.css
```diff
@@ -1643,11 +1643,16 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
-	width: 16px;
-	text-align: center;
+.inline-emoji {
 	display: inline-block;
+	font-size: 1.4em;
+	vertical-align: bottom;
+	line-height: 1;
+}
+
+.textcomplete-item .inline-emoji {
```
astorije: Kind of a nitpick but any reason to rename the class from `emoji` to `inline-emoji`? I'd rather keep the former if that's possible. :man_shrugging:

client/css/style.css
```diff
@@ -1643,11 +1643,16 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	opacity: 1;
 }
 
-.textcomplete-item .emoji {
-	margin-right: 8px;
-	width: 16px;
-	text-align: center;
+.inline-emoji {
 	display: inline-block;
+	font-size: 1.4em;
```
astorije: Could you actually go a little higher than that? I like the size of the autocomplete emoji, but the ones in the text are still a little too small IMO. Between `1.6` and `1.8em` would be sweet (I'm talking about default theme with no custom styling, btw), but I haven't looked at alignment stuff.

client/css/style.css
astorije: As far as I can tell, before your PR, emoji and text were both vertically misaligned in dropdown, and after your PR, emoji is nicely centered, but not text. Because they're not aligned with each other anymore, it jumped on me, but never realized both weren't correctly aligned before.
Any chance you could attempt to vertically center the text too?

Before | After
--- | ---
screen shot 2017-08-23 at 23 40 28 | screen shot 2017-08-23 at 23 40 47

client/js/libs/handlebars/parse.js
```diff
@@ -78,6 +81,8 @@ module.exports = function parse(text) {
 		} else if (textPart.channel) {
 			const escapedChannel = Handlebars.Utils.escapeExpression(textPart.channel);
 			return `<span class="inline-channel" role="button" tabindex="0" data-chan="${escapedChannel}">${fragments}</span>`;
+		} else if (textPart.emoji) {
+			return `<span class="inline-emoji">${fragments}</span>`;
```
astorije: I hope that someday we'll be able to display the reverse-lookup'ed text for an emoji in a tooltip. Great when an emoji doesn't show up (for compatibility or whatever reason), also good for accessibility, and things as simple as not recognizing what a damn emoji is meant to look like 😂

client/css/style.css
xPaw: Will do.

client/css/style.css
xPaw: I think 1.4 is the limit here before it starts pushing the lines around, but I will try again.

client/js/libs/handlebars/parse.js
YaManicKill: Yes, this. Currently I copy it and search on emojipedia.

client/css/style.css
astorije: I'd be okay if that means incrementing the line height a little bit. I think it's worth it.

client/css/style.css
astorije: @xPaw, did you manage to fix this?

client/js/libs/handlebars/parse.js
astorije: I just paste in Google but yeah, that's one of the reasons lol.

client/css/style.css
xPaw: I think so.

client/css/style.css
xPaw: I don't know, our line-height of 1.4 is already kinda pushing it.

astorije: Oh well, 1.4 is better than nothing I guess.

PR #1448, opened 2017-08-23
Set default quit default message

xPaw:


src/client.js
```diff
@@ -501,7 +501,7 @@ Client.prototype.quit = function() {
 	}
 	this.networks.forEach((network) => {
 		if (network.irc) {
-			network.irc.quit("Page closed");
+			network.irc.quit(`Page closed - ${pkg.homepage}`);
```
astorije: Can't use "Page closed" + `leaveMessage`?

xPaw: Too long for my liking

Issue #1436, opened 2017-08-19
Allow to "/cycle" current channel


dgw:
I occasionally find it necessary to part and rejoin channels, be it to fix some permissions issue with ChanServ or to prod ZNC into rejoining a stuck channel, etc.

Not to keep stealing ideas from Textual :grin: but it does have a `/cycle` command that does precisely this: It parts the current channel, and then rejoins it.

Maybe this would even be a good first "real" contribution (code, not just ideas) for someone like me?

xPaw:
> Maybe this would even be a good first "real" contribution (code, not just ideas) for someone like me?

Yes!

astorije:
I'm not sure I get the use case for it as I have never needed it, but I'm not using ZNC nor have had issues with ChanServ, so I'm not the best person to be in favor or against this, I'm fine with it if @xPaw is fine with it.

However, reading your description, I don't think `/cycle` is the best terminology for it. I think `/rejoin` would work better (after all, you've used that word 3 times in your description :D). And once #1430 gets a fix, `/rejoin [<channel>]` could allow an optional channel name to part-and-join.

And yup, totally open to a contribution :)

dgw:
> However, reading your description, I don't think `/cycle` is the best terminology for it. I think `/rejoin` would work better

I stole the command name from Textual too :grin:, what if both commands are exposed? The Lounge is the most Textual-like IRC experience I've found, either for web or non-Mac in general, so if I'm gravitating toward it for that reason I can't be the only one.

> Yes! You can figure out how part/join commands work and combine them.

Presumably by using parts of the internals from each command, not calling their functions directly? I'll give the command code a closer look soon-ish.

xPaw:
> I don't think /cycle is the best terminology for it. I think /rejoin would work better

While true, `/cycle` is a decade old command and we shouldn't have our own name for it. https://www.google.com/search?q=irc+cycle+command

astorije:
> While true, /cycle is a decade old command and we shouldn't have our own name for it.

I don't think that's the right way to look at it. The Lounge is not a decade-old client, and is not trying to look like a decade-old one either. People have reported switching to The Lounge for its modernity, we shouldn't think like that.
If something is more logical and practical to use, then it should be what we go for.

Anyway, in this very case, we should definitely have a `/cycle` alias, but I still believe `/rejoin` makes more sense.

YaManicKill:
[...]

PR #1449, opened 2017-08-24
Add /cycle command (a.k.a. /rejoin)

dgw:
First pass at implementing #1436, no idea if I'm doing this right but gotta start somewhere! I expect this branch will be force-pushed approximately 5,821 times before being merged, because I still don't actually understand how commands work.

Tested locally under Windows inside Git Bash (which doesn't terminate `node.exe` on <kbd>Ctrl</kbd>+<kbd>C</kbd> :disappointed:). Line endings should be fine because Git's `core.autocrlf` setting is enabled, but the linter also reports unused arguments (not sure if I can define the new function with just `network` and `chan`).

For a first pass it works pretty great, just as expected except the part message isn't used (doesn't appear to be used by `/part` itself either; when I tested, the part message sent was `#channelname` for either command). I'd like to make it so `/cycle`ing a channel doesn't move it to the bottom of the list, but so many PRs end up with bikeshedding that I didn't want to start out by bikeshedding myself. :laughing:


src/plugins/inputs/cycle.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
```
astorije: Mind calling the file `rejoin.js` please? :) I'd like this to be the canonical name, and `/cycle` seen as an alias.

astorije:
If you look into `index.html`, you'll see a help section with commands, could you add your `/rejoin` command + `/cycle` alias there, please? You'll find example on other commands with aliases there.

src/plugins/inputs/cycle.js
dgw: I disagree, on the basis that `/cycle` is a long-standing name for the function (and is even a builtin on some networks, e.g. [SwiftIRC](https://wiki.swiftirc.net/wiki/Other_commands#Cycle.2C_Hop)). "rejoin" is usually attached to rejoining channels upon being kicked.

Input from other maintainers (@xPaw, @YaManicKill?) will be the deciding factor here, though. If there's consensus to canonically call this `rejoin`, I won't fight it. But it is my strongly held belief that calling it `rejoin` instead of `cycle` is more or less breaking from tradition for the sake of doing so.

dgw:
I will update the docs once the issue of canonical name has been settled. (And I'll also port my changes to a PR on the docs repo, once this is merged.)

astorije:
> except the part message isn't used (doesn't appear to be used by `/part` itself either; when I tested, the part message sent was `#channelname` for either command)

Interestingly, I'm encountering the same issue when testing #1448. Not sure what's going on, but I wouldn't worry about that for now.

> (And I'll also port my changes to a PR on the docs repo, once this is merged.)

If you're talking about [this page](https://thelounge.github.io/docs/client/commands.html), don't worry about it, it's deprecated and was removed from the new docs (WIP). Help window is the source of truth.

src/plugins/inputs/cycle.js
astorije: I understand where you're coming from, and I don't disagree with the rationale in terms of usage which is why I'm happy with a `/cycle` alias, but when it comes to the filenames and such, this is The Lounge codebase territory.
We don't have other command-related filenames with `de`/`re` suffixes yet, but we do have commands that have their counterparts prefixed with `de` (`voice` vs. `devoice`, etc.). This file should then be `rejoin.js`, with support for `cycle` command.

src/plugins/inputs/cycle.js
```diff
@@ -0,0 +1,31 @@
...
+
+var Msg = require("../../models/msg");
+var Chan = require("../../models/chan");
+
+exports.commands = ["cycle", "rejoin"];
+
+exports.input = function(network, chan, cmd, args) {
```
astorije: Yep, you can just do:

```js
exports.input = function(network, chan) {
```

dgw:
Haven't squashed yet, but I added the help and renamed the command file.

Still want to figure out how to make it cycle without destroying the existing channel and losing scrollback.

src/plugins/inputs/rejoin.js
```diff
@@ -0,0 +1,31 @@
...
+var Msg = require("../../models/msg");
+var Chan = require("../../models/chan");
+
+exports.commands = ["cycle", "rejoin"];
+
+exports.input = function(network, chan) {
+	if (chan.type !== Chan.Type.CHANNEL) {
+		chan.pushMessage(this, new Msg({
+			type: Msg.Type.ERROR,
+			text: "You can only cycle channels."
```
xPaw: s/cycle/rejoin

src/plugins/inputs/rejoin.js
```diff
@@ -0,0 +1,31 @@
...
+	});
+	network.irc.part(chan.name, "Rejoining");
+
+	this.emit("join", {
+		network: network.id,
+		chan: chan
+	});
+	network.irc.join(chan.name);
+
+	this.save();
```
xPaw: Don't need this either.

dgw:
Squashed

src/plugins/inputs/rejoin.js
```diff
@@ -0,0 +1,31 @@
...
+exports.input = function(network, chan) {
+	if (chan.type !== Chan.Type.CHANNEL) {
+		chan.pushMessage(this, new Msg({
+			type: Msg.Type.ERROR,
+			text: "You can only cycle channels."
+		}));
+		return;
+	}
+
+	this.emit("part", {
```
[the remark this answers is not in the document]
astorije: We can't keep the chat window open without losing history like @dgw was trying to achieve, right?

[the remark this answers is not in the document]
xPaw: Not sending emits to client *will* keep history.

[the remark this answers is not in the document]
astorije: Odd, I tested this PR and it didn't.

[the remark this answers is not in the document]
astorije: @dgw, could you confirm this? When I tested, it killed the window and history then recreated it.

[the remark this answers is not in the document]
dgw: Sadly I cannot confirm that it keeps history. Even after making sure git said I'm up-to-date with `origin/cycle-command`, making sure the `emit`s were not present in `rejoin.js`, and running a fresh `npm run build`, starting up a test instance and checking the command still cleared the channel history.

I'm confused. Is the server perhaps catching the `PART` command and destroying the channel on its end, which then propagates to the client and clears the history?

client/index.html
```diff
@@ -787,6 +787,19 @@ <h1 class="title">Help</h1>
 
 						<div class="help-item">
 							<div class="subject">
+								<code>/rejoin</code>
+							</div>
+							<div class="description">
+								<p>
+									Equivalent to doing <code>/part</code> followed immediately
+									by <code>/join #channel</code>.
```
astorije: I'd prefer something like "Reload the current channel by leaving it then immediately re-joining it."

xPaw:
Damn yeah I forgot that receiving part from the server will close the channel.

https://github.com/thelounge/lounge/blob/6ba2b8a3d1440d46252e70e3fc11db622669eef8/src/plugins/irc-events/part.js#L14-L20

We might also hit an issue here if we ignore the part:
https://github.com/thelounge/lounge/blob/6ba2b8a3d1440d46252e70e3fc11db622669eef8/src/plugins/irc-events/join.js#L25

astorije:
Is there anything we can do about this?

YaManicKill:
> Is there anything we can do about this, you think?

Well yes, it depends how you want it to work. 

We could add a command (or an extra bit to the part command) that will part a channel, but not hide it or remove history. Could be useful.

dgw:
I noticed that receiving a kick doesn't destroy the channel. It does clear the list of users, but that's part of the point of this command (one use is to get ChanServ to sync up your modes). I might play with that, though it will probably lead to nothing.

client/index.html
dgw: I've been toying with wordings. Do any of these strike your fancy?

````
Quickly part and rejoin the current channel. Effectively
reloads the user list and syncs modes, at the expense of
losing channel scrollback.
----
Shortcut to leave and then immediately rejoin the current
channel, reloading users/modes/etc.
----
Leaves and immediately rejoins the current channel. Useful
to quickly get op from ChanServ in an empty channel, for example.
````

I'm not convinced that the verbiage about losing history is needed, but I think it's helpful unless a way to keep history can be implemented. :\

Comment (astorije) [reply client/index.html]: Good stuff @dgw (sorry I missed that notification)! 👏

I like a mix of first and last:

```
Leave and immediately rejoin the current channel. Useful
to quickly get op from ChanServ in an empty channel, for example.
```

> I'm not convinced that the verbiage about losing history is needed

I agree with that part, let's not.

dgw:
Updated help text and re-squashed.

Is this in merge-now-and-improve-later status, or in blocked-on-improvements-to-part-command status?

YaManicKill:
I think we should merge, as it just simplifies the `/part` and `/join #chan` process, and we can look into making it not lose scrollback later. It's still an improvement, it doesn't break any existing functionality.

YaManicKill:
@astorije Any objections to this being merged in the current state and adding an issue to look at keeping scrollback?

astorije:
> @astorije Any objections to this being merged in the current state and adding an issue to look at keeping scrollback?

Not at all, that was my feeling too :)

> If merged as-is, I would like to create the followup issue myself. Taking ownership of the "mess" I made, kinda. 😀

That's a great mindset, thank you!! 👏 👏

PR #1451, opened 2017-08-24
Do not hide time from events in condensed block

xPaw:
@astorije broke it.


astorije:
> @astorije broke it.

Hey not _exactly_, that's so unfair!

PR #1452, opened 2017-08-24
Completely refactor how date markers are inserted

### LINKED ISSUE #1450: Wrong order and date marker when using collapsed events
Steps to reproduce: 
* make sure you've enabled condensed events in settings
* Join a channel with at least one other user and ~no activity
* wait until the day changes and create an event, for example op the other user
* optional: type something to get one more date marker

The result looks like this:
image

The dates are actually correct, I joined on day 1 and everything else happened on day 2. The lines are just out of order.

It survives a page refresh.

The collapsed event group doesn't actually have expandable content (because the join and +o are already visible):
image

This is what it looks like without condensing enabled:
image
(Note: I have +o in this screenshot because I did not actually join, I started Lounge which connects to my bnc, so I was already there and had +o)

I've also noticed strange behaviour here:
image

There are two issues: The date marker is two lines "too late" and the two lines above it belong below the first line that comes after the date marker (look at the timestamps). I've only noticed because I was wondering how tia could join twice.
Note that everything above the date marker belongs to an expanded event group. Initially it was collapsed and only the stuff below the date marker was visible.

It looks to me that this problem occurs whenever a collapsed event group "wants to" group events across a day change.

xPaw:
Fixes #1450 and most likely other subtle bugs.

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -15,57 +15,85 @@ const sidebar = $("#sidebar");
 module.exports = {
 	appendMessage,
 	buildChannelMessages,
-	buildChatMessage,
 	renderChannel,
-	renderChannelMessages,
 	renderChannelUsers,
 	renderNetworks,
 };
 
-function buildChannelMessages(data) {
-	return data.messages.reduce(function(docFragment, message) {
-		appendMessage(docFragment, data.id, data.type, message.type, buildChatMessage({
-			chan: data.id,
-			msg: message
-		}));
+function buildChannelMessages(chanId, chanType, messages) {
+	return messages.reduce((docFragment, message) => {
+		appendMessage(docFragment, chanId, chanType, message);
 		return docFragment;
 	}, $(document.createDocumentFragment()));
 }
 
-function appendMessage(container, chan, chanType, messageType, msg) {
-	// TODO: To fix #1432, statusMessage option should entirely be implemented in CSS
-	if (constants.condensedTypes.indexOf(messageType) === -1 || chanType !== "channel" || options.statusMessages !== "condensed") {
-		container.append(msg);
+function appendMessage(container, chanId, chanType, msg) {
+	const renderedMessage = buildChatMessage(chanId, msg);
+
+	// Check if date changed
+	let lastChild = container.find(".msg").last();
+	const msgTime = new Date(msg.time);
+
+	// It's the first message in a window,
+	// then just append the message and do nothing else
+	if (lastChild.length === 0) {
+		container
+			.append(templates.date_marker({msgDate: msgTime}))
+			.append(renderedMessage);
+
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


client/js/render.js
```diff
@@ -24,37 +24,73 @@ module.exports = {
 
 function buildChannelMessages(data) {
 	return data.messages.reduce(function(docFragment, message) {
-		appendMessage(docFragment, data.id, data.type, message.type, buildChatMessage({
```
astorije: Just curious, why did you move these from arguments to an object? Also, any interest in using parameter destructuring in `appendMessage` (`function f({ a, b }) { ... }`)?

astorije:
Hey @xPaw, unfortunately...

double_date_marker

xPaw:
@astorije I refactored it even further, fixed your bug and found another bug which I also fixed.

client/js/render.js
```diff
@@ -15,57 +15,85 @@ const sidebar = $("#sidebar");
...
 };
 
-function buildChannelMessages(data) {
-	return data.messages.reduce(function(docFragment, message) {
-		appendMessage(docFragment, data.id, data.type, message.type, buildChatMessage({
-			chan: data.id,
-			msg: message
-		}));
+function buildChannelMessages(chanId, chanType, messages) {
+	return messages.reduce(function(docFragment, message) {
```
astorije: Nitpick: `return messages.reduce((docFragment, message) => {`

client/js/render.js
```diff
@@ -15,57 +15,85 @@ const sidebar = $("#sidebar");
...
-function appendMessage(container, chan, chanType, messageType, msg) {
+function appendMessage(container, chanId, chanType, msg) {
+	const renderedMessage = buildChatMessage(chanId, msg);
+
+	// Check if date changed
+	let lastChild = container.find(".msg").last();
+	const msgTime = new Date(msg.time);
+
+	// It's the first message in a window,
+	// then just append the message and do nothing else
```
astorije: An idea for a future PR would be to display the date marker if it's the first message in a window **and** there is no "more" button available. Reason for this would be to remove that odd situation that #1318 shows (I know it's not creating it, just making it more obvious).

client/js/render.js
```diff
@@ -15,57 +15,85 @@ const sidebar = $("#sidebar");
...
-		}));
+function buildChannelMessages(chanId, chanType, messages) {
+	return messages.reduce(function(docFragment, message) {
+		appendMessage(docFragment, chanId, chanType, message);
 		return docFragment;
 	}, $(document.createDocumentFragment()));
 }
 
-function appendMessage(container, chan, chanType, messageType, msg) {
+function appendMessage(container, chanId, chanType, msg) {
```
astorije: This method could use a test or 2, but not blocking this PR obviously.

PR #1465, opened 2017-08-26
Remove the "Stay signed in" checkbox at login

astorije:
This option is less and less the norm on modern webapps, it is fair to assume this is the default behavior. In fact, we were making it the default.

But more importantly, coming soon is the ability of remotely logging out of your other sessions, which is well handled through token deletion. That means we need to know about said tokens, which are not sent in no-"Stay signed in" version.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -203,7 +203,7 @@ function index(req, res, next) {
 	res.render("index", data);
 }
 
-function initializeClient(socket, client, generateToken, token) {
+function initializeClient(socket, client, token) {
 	socket.emit("authorized");
 
 	socket.on("disconnect", function() {
@@ -376,7 +376,7 @@ function initializeClient(socket, client, generateToken, token) {
 		});
 	};
 
-	if (generateToken) {
+	if (!Helper.config.public && token === null) {
 		client.generateToken((newToken) => {
 			token = newToken;
 
@@ -457,7 +457,7 @@ function performAuthentication(data) {
 	const socket = this;
 	let client;
 
-	const finalInit = () => initializeClient(socket, client, !!data.remember, data.token || null);
+	const finalInit = () => initializeClient(socket, client, data.token || null);
 
 	const initClient = () => {
 		client.ip = getClientIp(socket.request);
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1353,17 +1353,6 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	margin-top: 10px;
 }
 
-#sign-in .remember {
-	float: left;
-	font-size: 14px;
-	margin-top: 12px;
-}
-
-#sign-in .remember input {
-	float: left;
-	margin: 3px 10px 0 0;
-}
-
 #sign-in .btn {
 	margin-top: 25px;
 }
--- a/client/index.html
+++ b/client/index.html
@@ -97,12 +97,6 @@
 								<input class="input" type="password" name="password">
 								</label>
 							</div>
-							<div class="col-xs-12">
-								<label class="remember">
-									<input type="checkbox" name="remember" value="on" id="sign-in-remember" checked>
-									Stay signed in
-								</label>
-							</div>
 							<div class="col-xs-12 error" style="display: none;">
 								Authentication failed.
 							</div>
--- a/client/themes/crypto.css
+++ b/client/themes/crypto.css
@@ -78,11 +78,6 @@ a:hover,
 	font-size: 12px;
 }
 
-#sign-in .remember {
-	font-size: 12px;
-	line-height: 30px;
-}
-
 #sidebar .chan:first-child {
 	color: #00ff0e;
 }


src/server.js
```diff
@@ -351,25 +351,21 @@ function initializeClient(socket, client, generateToken, token) {
 		});
 	};
 
-	if (generateToken) {
```
xPaw: This `if` should be kept and replaced with `if (token === null)`. We only want to generate a token if there isn't one.

astorije: Done, better?

Regarding your `data.token || null` then this `if (token === null)`, we could just rely on `undefined` and not bother about that `|| null`, and do `if (!token)`, etc. Did you do that just to be more explicit? (which is fine by me of course, just checking)

astorije:
Tests are failing, I'm not sure what's going on with them.

src/server.js
xPaw: Yes explicit null is better.

src/server.js
```diff
@@ -364,12 +364,10 @@ function initializeClient(socket, client, generateToken, token) {
...
 				}
 			});
-
-			sendInitEvent(token);
 		});
-	} else {
-		sendInitEvent(null);
 	}
+
+	sendInitEvent(token);
```
xPaw: Keep the order, `client.generateToken` has a callback, so it calls `sendInitEvent` inside of it.

astorije: Good catch, thanks

astorije:
@xPaw, I have fixed the couple issues. Also used a promise to simplify calling of `sendInitEvent`, let me know what you think.

src/server.js
```diff
@@ -365,11 +367,11 @@ function initializeClient(socket, client, generateToken, token) {
...
 
-			sendInitEvent(token);
-		});
-	} else {
-		sendInitEvent(null);
+			return newToken;
+		}).catch((error) => log.error(error));
 	}
+
+	Promise.resolve(tokenPromise).then((t) => sendInitEvent(t));
```
xPaw: You could move `sendInitEvent` code here.

src/server.js
```diff
@@ -351,8 +351,10 @@ function initializeClient(socket, client, generateToken, token) {
 		});
 	};
 
-	if (generateToken) {
-		client.generateToken((newToken) => {
+	let tokenPromise = null;
+
+	if (Helper.config.private && token === null) {
```
xPaw: There's no such config as `private`.

src/server.js
astorije: 2 cases:

- `tokenPromise` is `null` so `Promise.resolve(null).then(...)` resolves right away
- `tokenPromise` is a promise of `token` so `Promise.resolve(tokenPromise).then(...)` resolves when `tokenPromise` resolves (because spec allows for `Promise.resolve([a promise])`.

src/server.js
astorije: omg when will I stop

src/server.js
astorije: Done.

src/server.js
astorije: Fixed.

src/server.js
```diff
@@ -365,11 +357,19 @@ function initializeClient(socket, client, generateToken, token) {
...
 
-			sendInitEvent(token);
-		});
-	} else {
-		sendInitEvent(null);
+			return newToken;
+		}).catch((error) => log.error(error));
 	}
+
+	Promise.resolve(tokenPromise).then((tokenToSend) => {
```
astorije: 2 cases:

- `tokenPromise` is `null` so `Promise.resolve(null).then(...)` resolves right away
- `tokenPromise` is a promise of `token` so `Promise.resolve(tokenPromise).then(...)` resolves when `tokenPromise` resolves (because spec allows for `Promise.resolve([a promise])`.

src/client.js
```diff
@@ -295,13 +295,14 @@ Client.prototype.connect = function(args) {
...
-		if (err) {
-			throw err;
-		}
-
-		callback(buf.toString("hex"));
+Client.prototype.generateToken = function() {
+	return new Promise((resolve, reject) => {
+		crypto.randomBytes(48, (err, buf) => {
+			if (err) {
+				reject(err);
```
xPaw: You're going to reject and then resolve it, you need a `return` in here.

astorije: I'm pretty sure once it's rejected it can't be resolved anymore, but fixed anyway.

astorije:
Okay, I can `rm` the second commit if you no like 😢
I'll let @YaManicKill review first before removing in case he feels strongly about this, but I'm okay removing it.

If `async`/`await` was a Node.js v6 thing, I'd urge more to start the switch to promises (which is one step towards `async`/`await`), but since it's only a v8 thing, we won't be able to use them before a looong time.

astorije:
I like that `generateToken()` actually returns the token (in a promise) rather than takes a callback. Conceptually, it makes sense: `generateToken` generates a token and spits out an actual value.

Also, I'd argue it's a bit cleaner as there is some de-duplication involved.

But I'm totally fine removing it, that was more of a suggestion so I'll remove it.

astorije:
Extracted the promisifying of `generateToken` in #1483 in case we want it back, as an archive.

PR #1468, opened 2017-08-28
Combine unread notifications into a single one

xPaw:
Turns this:
photo_2017-08-28_15-35-55

Into this:

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -26,7 +26,7 @@ function Chan(attr) {
 		type: Chan.Type.CHANNEL,
 		firstUnread: 0,
 		unread: 0,
-		highlight: false,
+		highlight: 0,
 		users: []
 	});
 }
@@ -78,7 +78,7 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 		}
 
 		if (msg.highlight) {
-			this.highlight = true;
+			this.highlight++;
 		}
 	}
 };
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -106,20 +106,29 @@ module.exports = function(irc, network) {
 
 		// Do not send notifications for messages older than 15 minutes (znc buffer for example)
 		if (highlight && (!data.time || data.time > Date.now() - 900000)) {
-			let title = data.nick;
+			let title = chan.name;
+			let body = Helper.cleanIrcMessage(data.message);
 
+			// In channels, prepend sender nickname to the message
 			if (chan.type !== Chan.Type.QUERY) {
-				title += ` (${chan.name}) mentioned you`;
-			} else {
-				title += " sent you a message";
+				body = `${data.nick}: ${body}`;
+			}
+
+			// If a channel is active on any client, highlight won't increment and notification will say (0 mention)
+			if (chan.highlight > 0) {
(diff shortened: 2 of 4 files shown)


astorije:
Will there be a way to apply this same code for regular desktop notifications as well?

xPaw:
@astorije Perhaps in the future, but desktop notifications expire automatically so it's less of a problem.

astorije:
Fair. Don't push notifications also expire on desktop though? I need to play with this to get a better sense of what happens then.

src/plugins/irc-events/message.js
```diff
@@ -109,16 +109,14 @@ module.exports = function(irc, network) {
 			let title = data.nick;
 
 			if (chan.type !== Chan.Type.QUERY) {
-				title += ` (${chan.name}) mentioned you`;
-			} else {
-				title += " sent you a message";
+				title += ` (${chan.name})`;
```
astorije: Any chance you can/want to use the same wording than for desktop notifications? "says:", etc.

astorije:
So desktop notifications do expire. It's a bit weird on laptop, but oh well.
Left a comment about consistent wording with desktop notifications though, what do you think @xPaw?

src/plugins/irc-events/message.js
xPaw: The shorter it is, the better, imo. The way it groups messages looks a little messy, so we might something like bullet points.

astorije: Alright then.

xPaw:
I ran this PR for a couple of days myself now, and found it was a subpar experience. So now I group notifications per channel/target and count mentions correctly on the server (it will say `1 new message` for queries and `1 mention` for channels).

It will also only display latest message per notification instead of combining them on new lines (it was really busy and hard to read).

astorije:
When receiving multiple messages from a same person, if you only display the latest one, it's lacking context and is therefore not really helpful. What about showing the message if only one or "X new messages" if more than one (what telegram does), or another could be "<first message> and X other messages" type of thing.

xPaw:
@astorije I have a hard time understanding your last sentence, but the implementation does mirror what Telegram does as far as I know.

astorije:
> or another could be " and X other messages" type of thing.

Alright so what I meant is that the combined notification popup could be:

> [First of the N messages received]
> *... and [N-1] other messages*

Better now?

xPaw:
@astorije Would N-1 be just highlights/mentions or all unread messages for that channel?

astorije:
Just what would usually trigger notification popups, so I guess only highlights/mentions.

xPaw:
Added that.

src/models/chan.js
```diff
@@ -25,7 +25,7 @@ function Chan(attr) {
 		type: Chan.Type.CHANNEL,
 		firstUnread: 0,
 		unread: 0,
-		highlight: false,
+		highlight: 0,
```
astorije: Kind of a nitpick, but could you rename this into `highlights` or `highlightCount` then?

xPaw: This requires finding all the instances (client included), I'd rather not.

astorije: Apart from those you have modified in this PR, the only other occurrence of this that I found is https://github.com/thelounge/lounge/blob/master/client/views/chan.tpl#L3

xPaw: There is definitely more.

astorije: All others that I found are for `Msg.highlight`.

PR #1469, opened 2017-08-29
Sensationalize feature list

MaxLeiter:
Unsure if I like this, just thought I'd put it out there for feedback/opinions.


starquake:
It sounds less technical that way. That might be a pro.

"What else could you ask for?" made me cringe though.

astorije:
In case @thelounge/maintainers stop by, I was okay with most of this (and agree with @starquake too).

PR #1471, opened 2017-08-29
Automatic client reconnection

xPaw:
Instead of simply redrawing all channel windows, it only draws new ones, and for existing ones it updates the topic and only new messages (on auth, client tells the server latest message id it has seen).

Client will force reload the client page if server restarted between connections (server generates a random number on start), this is to help with buffer problems (e.g. last message id going backwards) and Lounge updates.

The red disconnected bar also reports all the status changes.

- [x] Disable input bar correctly.
- [x] Test removing user and reconnecting (should refresh into sign in page).
- [x] Cleanup error message handling

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1483,6 +1483,15 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 }
 
 #connection-error {
+	font-size: 12px;
+	line-height: 36px;
+	font-weight: bold;
+	letter-spacing: 1px;
+	word-spacing: 3px;
+	text-transform: uppercase;
+	background: #e74c3c;
+	color: #fff;
+	text-align: center;
 	display: none;
 }
 
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -588,9 +588,6 @@ $(function() {
 	}
 	setTimeout(updateDateMarkers, msUntilNextDay());
 
-	// Only start opening socket.io connection after all events have been registered
-	socket.open();
-
 	window.addEventListener("popstate", (e) => {
 		const {state} = e;
 		if (!state) {
@@ -604,4 +601,7 @@ $(function() {
 			});
 		}
 	});
+
+	// Only start opening socket.io connection after all events have been registered
+	socket.open();
 });
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -35,6 +35,10 @@ function buildChannelMessages(chanId, chanType, messages) {
 }
 
 function appendMessage(container, chanId, chanType, msg) {
+	if (utils.lastMessageId < msg.id) {
+		utils.lastMessageId = msg.id;
+	}
+
 	let lastChild = container.children(".msg, .date-marker-container").last();
 	const renderedMessage = buildChatMessage(msg);
 
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -8,6 +8,7 @@ const utils = require("./utils");
 const sorting = require("./sorting");
 const constants = require("./constants");
 const condensed = require("./condensed");
+const helpers_parse = require("./libs/handlebars/parse");
 
 const chat = $("#chat");
 const sidebar = $("#sidebar");
@@ -27,11 +28,11 @@ module.exports = {
 	renderNetworks,
 };
 
-function buildChannelMessages(chanId, chanType, messages) {
+function buildChannelMessages(container, chanId, chanType, messages) {
 	return messages.reduce((docFragment, message) => {
 		appendMessage(docFragment, chanId, chanType, message);
 		return docFragment;
-	}, $(document.createDocumentFragment()));
+	}, container);
 }
 
 function appendMessage(container, chanId, chanType, msg) {
@@ -121,7 +122,7 @@ function renderChannel(data) {
 }
 
 function renderChannelMessages(data) {
-	const documentFragment = buildChannelMessages(data.id, data.type, data.messages);
+	const documentFragment = buildChannelMessages($(document.createDocumentFragment()), data.id, data.type, data.messages);
 	const channel = chat.find("#chan-" + data.id + " .messages").append(documentFragment);
 
 	const template = $(templates.unread_marker());
@@ -168,7 +169,7 @@ function renderChannelUsers(data) {
 	}
 }
 
-function renderNetworks(data) {
+function renderNetworks(data, singleNetwork) {
 	sidebar.find(".empty").hide();
 	sidebar.find(".networks").append(
 		templates.network({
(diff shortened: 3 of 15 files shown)


xPaw:
This is ready.

client/css/style.css
```diff
@@ -1495,6 +1495,15 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
 #connection-error {
+	font-size: 12px;
+	line-height: 36px;
+	font-weight: bold;
+	letter-spacing: 1px;
+	word-spacing: 3px;
+	text-transform: uppercase;
+	background: #e74c3c;
+	color: #fff;
+	text-align: center;
```
astorije: Looks like a button, is not a button 😅
When testing a couple days ago, I ended up in a situation where the banner would show up, but the app would never reload. I had to close it and re-open it. Should this banner stay a button?

client/js/lounge.js
```diff
@@ -814,4 +811,7 @@ $(function() {
 			});
 		}
 	});
+
+	// Only start opening socket.io connection after all events have been registered
+	socket.open();
```
astorije: Any specific reason for this change, or is it just to be totally in agreement with the comment above?

client/js/render.js
```diff
@@ -8,6 +8,7 @@ const utils = require("./utils");
 const sorting = require("./sorting");
 const constants = require("./constants");
 const condensed = require("./condensed");
+const helpers_parse = require("./libs/handlebars/parse");
```
astorije: Underscore case oO
Just `parse` instead?

client/js/render.js
```diff
@@ -72,6 +73,10 @@ function appendMessage(container, chanId, chanType, msg) {
 }
 
 function buildChatMessage(msg) {
+	if (utils.lastMessageId < msg.id) {
+		utils.lastMessageId = msg.id;
+	}
```
astorije: It's kinda horrible to see such side-effect inside the function itself. Can it go where the message gets appended to the container instead?

client/js/render.js
```diff
@@ -158,23 +163,59 @@ function renderChannelUsers(data) {
 	}
 }
 
-function renderNetworks(data) {
+function renderNetworks(data, singleNetwork) {
```
astorije: I'm not sure to see what `singleNetwork` is for, could you explain?

client/js/socket-events/auth.js
```diff
@@ -3,15 +3,34 @@
...
 const socket = require("../socket");
 const storage = require("../localStorage");
+const utils = require("../utils");
 
 socket.on("auth", function(data) {
+	// If we reconnected and serverHash differs, that means the server restarted
+	// And we will reload the page to grab the latest version
+	if (utils.serverHash > -1 && data.serverHash > -1 && data.serverHash !== utils.serverHash) {
+		socket.disconnect();
+		$("#connection-error").text("Reloading the page…");
```
astorije: Nitpick but just `Reloading…` or `Refreshing…` is enough IMO.

client/js/socket.js
```diff
@@ -3,53 +3,46 @@
...
 const io = require("socket.io-client");
 const path = window.location.pathname + "socket.io/";
+const status = $("#loading-page-message, #connection-error");
 
 const socket = io({
 	transports: $(document.body).data("transports"),
 	path: path,
 	autoConnect: false,
-	reconnection: false
+	reconnection: !$(document.body).hasClass("public")
```
astorije: I get that we wouldn't want page refresh on public mode, but is this completely disabling client reconnection on public mode? If so, why?

client/themes/crypto.css
```diff
@@ -65,14 +65,6 @@ a:hover,
...
 }
 
-.btn-reconnect {
-	background: #f00;
-	color: #fff;
-	border: 0;
-	border-radius: 0;
-	margin: 0;
-}
-
```
astorije: Don't themes need to define their own values for `#connection-error`?

astorije:
Pretty solid! Left a few comments/questions.

client/css/style.css
xPaw: Are you sure about not reloading? socket.io will try to reconnect for an infinite amount of time.

client/js/lounge.js
xPaw: Align with the comment.

client/js/render.js
xPaw: I can move it to `appendMessage`.

client/js/render.js
xPaw: Gotta do different logic when rendering channels on init when it's reconnecting, so that it won't render duplicate channels.

client/js/socket.js
xPaw: Because reconnecting in public mode is effectively useless, it will destroy all the state.

client/themes/crypto.css
xPaw: I moved it to main style file.

client/css/style.css
astorije: I'm honestly not sure what happened precisely no. It acts super weirdly when running server and client from my laptop (not your PR's fault, it's the case on `master` as well) and that's all I tested so far. It could very well be macOS trying to be too smart.
I need to test with multiple devices and see how things go.

client/js/render.js
astorije: That would be better I think yeah.

client/js/render.js
astorije: I'm still not a fan of the implementation though. It's essentially 2 different functions in 1 with a boolean to switch implementation so that's why I think it reads a bit bizarre.

I'll try to think of other ways as I continue with the review, but on the top of my head, what do you think of either: 2 different functions, calling the correct one when needed (with potentially a common function/helper for shared logic); or a `renderNetworks` that calls multiple times a `renderNetwork` (what we do when building channels and messages), calling the former one on init and the latter one when it's reconnecting?

client/js/socket.js
astorije: So if client loses connection in public mode (in contrast with willingly closing the tab or logging out), it's gone forever? I get that we're doing that currently, but with automatic reconnection possible, that's bit harsh lol

client/themes/crypto.css
astorije: Alright. They were in the theme files on purpose for future-proof theming, but that's fine as well.

Leave the `#f00` background override in `crypto.css` though.

client/js/socket.js
xPaw: If we start adding logic on the server to not destroy the connection as soon as socket closes the connection, we most likely will be opening ourselves to abuse.

client/js/render.js
xPaw: That will effectively be duplicate code, and it doesn't solve the problem.

client/js/socket-events/auth.js
xPaw: What about `Server restarted, reloading…` as it's better to explain why it's reloading.

astorije: Works for me

client/js/render.js
xPaw: This matches how we do it in other files.

client/css/style.css
xPaw: It looks like our server close cleanup now causes the client to stop reconnecting (io server disconnect), I will look into it.

xPaw:
@astorije Fixed reconnection not working when server shuts down (now that we cleanup on ctrl+c).

client/js/socket.js
astorije: Agreed, though the more I think about it, the more I see public mode as short-lived sessions with no authentication/synchronization across devices possible. Absolutely not something for this PR, but I'd be curious to see experiments around this to see how it looks and plays.

astorije:
Cool, thanks @xPaw, I'll give it another shot soon!

client/css/style.css
YaManicKill: It only looks like a button because it previously was a button. It actually doesn't really look like a button on it's own.

client/js/lounge.js
YaManicKill: To be fair, the statement in the comment is binary. You can't be partially in agreement with the comment. It is either true or it isn't.

PR #1475, opened 2017-08-30
Implement condensed messages option entirely with CSS

### LINKED ISSUE #1432: Switching between condensed/expanded options for status message does nothing before reloading the app

xPaw:
Fixes #1432.

This also has a side effect of condensing single status messages too *(1 user joined)*.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -820,24 +820,33 @@ kbd {
 	flex-basis: 100%;
 }
 
-#chat .condensed-text {
+#chat .condensed-summary .content {
+	display: block;
 	cursor: pointer;
 	transition: opacity 0.2s;
 }
 
-#chat .condensed-text:hover {
-	opacity: 0.6;
-}
-
-#chat .condensed-text .toggle-button:hover {
-	opacity: 1;
-}
-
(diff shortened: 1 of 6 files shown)


astorije:
I'm fine with condensed for single messages! Makes it look consistent when you explicitly want condensed statuses. @YaManicKill, you're okay with that?

YaManicKill:
Well, I implemented that initially and was asked to remove it.

astorije:
> Well, I implemented that initially and was asked to remove it.

Really? Was it me lol? I don't remember that if so. I am fine with both really so if you're okay with it, I am too.

PR #1476, opened 2017-08-30
Take an optional argument in /part

### LINKED ISSUE #1430: Can't /part other channel than active one
I was confused earlier when I did `/close nickname` to close an inactive query window, and The Lounge instead parted the currently active channel.

Then I tested also `/part #backgroundchan`, and it again left the current channel instead.

For channels, this can be worked around with `/quote PART #channel`, but I'm now not aware of any way to close a query without switching to it or using the sidebar context menu.

eliemichel:
Fix #1430 

`/part #foo` or `/close #foo` closes the chan `#foo` even if it is not the active one.

`/part` and `/close` without an argument keep their previous meaning.

### DIFF
--- a/src/plugins/inputs/part.js
+++ b/src/plugins/inputs/part.js
@@ -9,7 +9,17 @@ exports.commands = ["close", "leave", "part"];
 exports.allowDisconnected = true;
 
 exports.input = function(network, chan, cmd, args) {
-	if (chan.type === Chan.Type.LOBBY) {
+	let target = args.length === 0 ? chan : _.find(network.channels, {name: args[0]});
+	let partMessage = args.length <= 1 ? Helper.config.leaveMessage : args.slice(1).join(" ");
+
+	if (typeof target === "undefined") {
+		// In this case, we assume that the word args[0] is part of the leave
+		// message and we part the current chan.
+		target = chan;
+		partMessage = args.join(" ");
+	}
+
+	if (target.type === Chan.Type.LOBBY) {
 		chan.pushMessage(this, new Msg({
 			type: Msg.Type.ERROR,
 			text: "You can not part from networks, use /quit instead."
@@ -17,18 +27,17 @@ exports.input = function(network, chan, cmd, args) {
 		return;
 	}
 
-	network.channels = _.without(network.channels, chan);
-	chan.destroy();
+	network.channels = _.without(network.channels, target);
+	target.destroy();
 	this.emit("part", {
-		chan: chan.id
+		chan: target.id
 	});
 
-	if (chan.type === Chan.Type.CHANNEL) {
+	if (target.type === Chan.Type.CHANNEL) {
 		this.save();
 
 		if (network.irc) {
-			const partMessage = args[0] ? args.join(" ") : Helper.config.leaveMessage;
-			network.irc.part(chan.name, partMessage);
+			network.irc.part(target.name, partMessage);
 		}
 	}
 
--- a/client/index.html
+++ b/client/index.html
@@ -787,10 +787,13 @@
 
 						<div class="help-item">
 							<div class="subject">
-								<code>/part</code>
+								<code>/part [channel]</code>
 							</div>
 							<div class="description">
-								<p>Close the current channel or private message window.</p>
+								<p>
+									Close the specified channel or private message window, or the
+									current channel if <code>channel</code> is ommitted.
+								</p>
 								<p>Aliases: <code>/close</code>, <code>/leave</code></p>
 							</div>
 						</div>


src/plugins/inputs/part.js
```diff
@@ -9,26 +9,43 @@ exports.commands = ["close", "leave", "part"];
 exports.allowDisconnected = true;
 
 exports.input = function(network, chan, cmd, args) {
-	if (chan.type === Chan.Type.LOBBY) {
+	let target, partedChan;
+	if (args.length === 0) {
+		target = chan.name;
```
YaManicKill: Could we simplify this and only have 1 variable, because we are just storing the name twice essentially.

So something like
```
const target = args.length === 0 ? chan : _.find(network.channels, {name: args[0]});
```

And then when we want to refer to the name, we can just do `target.name`. Does that make sense?

eliemichel: All right!

src/plugins/inputs/part.js
```diff
@@ -9,26 +9,36 @@ exports.commands = ["close", "leave", "part"];
...
-		chan: chan.id
+		chan: target.id
 	});
 
-	if (chan.type === Chan.Type.CHANNEL) {
+	if (target.type === Chan.Type.CHANNEL) {
 		this.save();
 
 		if (network.irc) {
 			const partMessage = args[0] ? args.join(" ") : Helper.config.leaveMessage;
```
xPaw: If you provide a target into the command, part message will be wrong here?

src/plugins/inputs/part.js
```diff
@@ -9,26 +9,36 @@ exports.commands = ["close", "leave", "part"];
 exports.allowDisconnected = true;
 
 exports.input = function(network, chan, cmd, args) {
-	if (chan.type === Chan.Type.LOBBY) {
+	const target = args.length === 0 ? chan : _.find(network.channels, {name: args[0]});
+
+	if (typeof target === "undefined") {
+		chan.pushMessage(this, new Msg({
+			type: Msg.Type.ERROR,
+			text: "Cannot part channel " + target.name + ": Channel not found."
```
xPaw: `target.name` is undefined here.

src/plugins/inputs/part.js
YaManicKill: Oh yeah, good point.

https://en.wikipedia.org/wiki/List_of_Internet_Relay_Chat_commands#PART

Spec says `PART <channels> [<message>]` (where channels is a csv). So I guess we want to check if the chan exists, the nthe part message is the rest of the args. If the chan doesn't exist, then it's a part message. Although that could get messy.

How do other clients deal with this? Do they support `/part` without a chan argument?

src/plugins/inputs/part.js
eliemichel: Ah yeah I remember now why I did not go for the ternary at first..

src/plugins/inputs/part.js
eliemichel: Mmh does the lobby chan have a name? If not, it will never by `/part`ed from anywhere but itself.

xPaw: It does, the same name you see in the UI.

eliemichel: Ok, but what do you expect to happen when somebody does a `/part Freenode` for instance? For know, it responds `You cannot part from networks`, which does not seem unreasonable to me. Though, I'm ready to change this behavior if needed.

eliemichel: Oh well I missed the point about the part message. I just tested in weechat: it checks that the first argument starts with a `#`. If so, it parts this chan. Otherwise, it considers the argument to be the part message. The drawback is that it prevents one from parting a private chan (`/part SomeNickname` does not work and just parts the current window).

YaManicKill: That sounds like a sensible way to do things, I think.

xPaw: You could just check if first argument is CHANNEL or QUERY and it'll work just fine.

YaManicKill: There's a problem with that, @xPaw 

Say I do `/part thelounge is annoying me` or something similar, then we check if `thelounge` is a channel, and it is, how do we know if the user wants to part that channel, or whether it's part of the part message.

xPaw: There's not much we can do about that. But channels do have to be a full name (so `#thelounge` in this case).

YaManicKill: If they have to include the `#`, then I think that's fine. I wouldn't want it to work without the `#`, otherwise we could get into some problems.

eliemichel: The problem mainly concerns query windows. If I `/part YaManicKill is annoying me` while I'm looking at the `#thelounge` window, but have a query window openned in background with YaManicKill, which one should I quit?

YaManicKill: Well, a part message makes no sense for a query window.

So I'd say that if there is more than 1 arg and we are trying to part a query, don't let the user do it.

eliemichel: I propose the following:

 * `/part #foo bar` quits `#foo` with message "bar".
 * `/part foo bar` quits the **current window** with message "foo bar" even if "foo" is a valid query window..
 * `/part foo` with "foo" a valid query window quits the query window `foo` with no message.
 * `/part foo` with "foo" **not** being a valid query window quits the current window with message "foo".

As a by-product, parting a query window with a leave message is still possible, but only by switching to this window and `/part some message`-ing.

xPaw: Sounds fine to me.

> parting a query window with a leave message is still possible, 

Since part is only sent for channels, it's effectively discarded.

YaManicKill: Yup, sounds fine to me also.

eliemichel: also, `/part #foo bar` displays an error message and do nothing if `#foo` does not exist as a window. It does not part the current window as it would do for `/part foo bar` (with no `#`)

src/plugins/inputs/part.js
```diff
@@ -9,26 +9,49 @@ exports.commands = ["close", "leave", "part"];
 exports.allowDisconnected = true;
 
 exports.input = function(network, chan, cmd, args) {
-	if (chan.type === Chan.Type.LOBBY) {
+	// Whether args[0] is to be considered as a target to part or the first word
+	// of leave message.
+	const targetArgument = args.length > 0 && (_.startsWith(args[0], "#") || args.length === 1);
```
MaxLeiter: This is on the server, so `args[0].startsWith` should work.

MaxLeiter: ignore this, see @xPaw's comment

dgw:
I agree with @xPaw. Checking for whether `args[0]` is an existing channel is an easy solution that leaves room for progressive enhancement (e.g. supporting multiple comma-separated channels, checking potential channel names against `CHANTYPES` instead of hardcoding `#`, etc.).

Overall, I'm excited to be able to part channels without switching to them! :+1:

MaxLeiter:
Can you squash commits, @eliemichel? Great work!

astorije:
Do we need to update https://github.com/thelounge/lounge/blob/master/client/index.html#L788-L796 ?

eliemichel:
Yey sorry guys I'm kind of in a rush for a deadline now, ping me back if I haven't done this end of next week. ;)

astorije:
@eliemichel, since you're busy, I took the liberty to update the help section, and squash your commits. Everything else in the code, I left it untouched :)

I'll let @xPaw or @YaManicKill merge it if they agree with the help changes.

eliemichel:
@astorije Thx!

YaManicKill:
> @astorije Thx!

No, thank you. This is a great addition, thanks for it :-)

PR #1477, opened 2017-08-30
Gracefully quit on Ctrl+C

### LINKED ISSUE #268: Handle SIGTERM and SIGINT
We suggest using Ctrl+C to shutdown the process, but we never actually handle it to do any cleanup.

We should be disconnecting all the connected IRC networks and client sockets.

xPaw:
Fixes #268

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -490,7 +490,7 @@ Client.prototype.names = function(data) {
 	});
 };
 
-Client.prototype.quit = function() {
+Client.prototype.quit = function(signOut) {
 	const sockets = this.sockets.sockets;
 	const room = sockets.adapter.rooms[this.id];
 
@@ -499,7 +499,10 @@ Client.prototype.quit = function() {
 			const socket = sockets.connected[user];
 
 			if (socket) {
-				socket.emit("sign-out");
+				if (signOut) {
+					socket.emit("sign-out");
+				}
+
 				socket.disconnect();
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/server.js
```diff
@@ -115,6 +115,31 @@ module.exports = function() {
...
+
+		// Handle ctrl+c and kill gracefully
+		let suicideTimeout = null;
+		const exitGracefully = function() {
+			if (suicideTimeout !== null) {
+				return;
+			}
+
+			// Forcefully exit after 3 seconds
+			suicideTimeout = setTimeout(() => process.exit(1), 3000);
```
astorije: Maybe another `log.info` in there, like "Something went wrong, force-quitting now"? Also, could do `5000` just in case, but whatever.

astorije:
@xPaw feel free to address comment or not, then to merge or let @YaManicKill review first. I'm good either way.

src/server.js
xPaw: I don't think it's necessary to add a message.

src/client.js
```diff
@@ -499,7 +499,10 @@ Client.prototype.quit = function() {
 			const socket = sockets.connected[user];
 
 			if (socket) {
-				socket.emit("sign-out");
+				if (signOut) {
```
YaManicKill: This feels like an unrelated fix.

xPaw: It's not, we don't want to sign out users on server close.

YaManicKill: Oh woops, I missed that we were already sending sign-out. Sounds good.

PR #1478, opened 2017-08-31
Implement a proper LDAP authentication process

eliemichel:
This replaces #980 with a cleaner commit history.


src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,132 @@
...
+function ldapAuthCommon(user, bindDN, password, callback) {
+	const config = Helper.config;
+
+	const ldapclient = ldap.createClient({
+		url: config.ldap.url,
+		tlsOptions: config.ldap.tlsOptions
+	});
+
+	ldapclient.on("error", function(err) {
+		log.error("Unable to connect to LDAP server", err);
```
astorije: This actually produces ugly stacks on the output. Use this instead:

```js
log.error(`Unable to connect to LDAP server: ${err}`);
```

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,132 @@
...
+	if (!user) {
+		return callback(false);
+	}
+
+	const config = Helper.config;
+
+	const userDN = user.replace(/([,\\/#+<>;"= ])/g, "\\$1");
+	const bindDN = config.ldap.primaryKey + "=" + userDN + "," + config.ldap.baseDN;
+
+	log.info("Auth against LDAP ", config.ldap.url, " with provided bindDN ", bindDN);
```
astorije: Use template literals:

```js
log.info(`Auth against LDAP ${config.ldap.url} with provided bindDN ${bindDN}`);
```

Same for the logging strings below.

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,132 @@
...
+
+	const ldapclient = ldap.createClient({
+		url: config.ldap.url,
+		tlsOptions: config.ldap.tlsOptions
+	});
+
+	const base = config.ldap.searchDN.base;
+	const searchOptions = {
+		scope: config.ldap.searchDN.scope,
+		filter: "(&(" + config.ldap.primaryKey + "=" + userDN + ")" + config.ldap.searchDN.filter + ")",
```
astorije: Are you missing a space after the first closing parenthesis or does this not matter?
Also, would this be clearer to read (it is to me)?

```js
`(&(${config.ldap.primaryKey}=${userDN}) ${config.ldap.searchDN.filter})`
```

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,132 @@
...
+		callback(valid);
+	}
+
+	let auth;
+	if ("baseDN" in Helper.config.ldap) {
+		auth = simpleLdapAuth;
+	} else {
+		auth = advancedLdapAuth;
+	}
+	return auth(user, password, callbackWrapper);
```
astorije: It seems like you don't need a function for that and the following would be enough, but that's just a suggestion, I'm fine either way. @xPaw?

```js
return auth(user, password, (valid) => {
	if (valid && !client) {
		manager.addUser(user, null);
	}
	callback(valid);
});
```

src/plugins/auth/local.js
```diff
@@ -0,0 +1,42 @@
...
+
+			callback(matching);
+		}).catch((error) => {
+			log.error(`Error while checking users password. Error: ${error}`);
+		});
+}
+
+module.exports = {
+	auth: localAuth,
+	isEnabled: () => true
```
astorije: Is it expected that local auth is enabled even when LDAP is enabled?

src/plugins/auth/ldap.js
```diff
@@ -0,0 +1,132 @@
...
+}
+
+function isLdapEnabled() {
+	return !Helper.config.public && Helper.config.ldap.enable;
+}
+
+module.exports = {
+	auth: ldapAuth,
+	isEnabled: isLdapEnabled
```
astorije:
```js
isEnabled: () => !Helper.config.public && Helper.config.ldap.enable
```

Actually, does it even have to be a function? Can it be:

```js
enabled: !Helper.config.public && Helper.config.ldap.enable
```

And you call it with `ldapAuth.enabled`. Double-check that it works though.

src/plugins/auth/local.js
astorije: Looking at code below, because it's in an `else if` that will be ignored when LDAP auth is enabled, this gets ignored. I would prefer if explicit:

```js
isEnabled: !Helper.config.public && !Helper.config.ldap.enable
```

(Assuming not being a function works, `() => ...` otherwise)

test/plugins/auth/ldap.js
```diff
@@ -0,0 +1,151 @@
...
+		if (req.filter.matches(obj.attributes)) {
+			// TODO: check req.scope if ldapjs does not
+			res.send(obj);
+		}
+
+		res.end();
+	});
+
+	server.listen(serverPort, function() {
+		console.log("LDAP server listening at %s", server.url);
```
astorije: Not sure we want to `console.log` in tests, especially since this is not something you can visit to test or something.

test/server.js
```diff
@@ -37,6 +37,11 @@ describe("Server", () => {
 	describe("WebSockets", () => {
 		let client;
 
+		before((done) => {
+			Helper.config.public = true;
```
astorije: What is this for (I mean, apart from fixing broken tests)? And isn't that already set [here](https://github.com/thelounge/lounge/blob/e292ef2bed8017eb913e48c6a5548753d2f633a6/test/fixtures/.lounge/config.js#L5)?

src/plugins/auth/ldap.js
eliemichel: Doesn't matter, `&(foo)(bar)` filters for `foo` AND `bar`.

src/plugins/auth/ldap.js
eliemichel: All right, I'm gonna switch everything to these strings.

src/plugins/auth/ldap.js
eliemichel: Yeah it does not really matter to me, but I wanted to highlight that in the end this mechanism does not belong in here. Whatever calls the auth plugins should just consider that as soon as the plugin accepted the auth, we must create the client if it did not already. I did not write it this way, though, because I want to be sure not to break something in the local auth.

src/plugins/auth/ldap.js
eliemichel: I'm designing what may become the general auth plugin API here, so I'd rather be carefull and make this check dynamic. We may eventually have an auth system whose availability is not known statically from the config but, say, by checking first for some web API or so (e.g. OAuth)..

src/plugins/auth/local.js
eliemichel: This is intented to be a fallback mechanism. The LDAP auth plugin would be the first tested, then if it is not available the system falls back to the local one. There might be an arbitrary list of auth plugin to test, and the local auth plugin is intended to be the ultimate one, that should hence always be available. So, `isEnabled: () => true` sounds semantically more appropriate to me.

eliemichel: I'll replace the if else by the mechanism I just described to make things clearer.

test/plugins/auth/ldap.js
eliemichel: That's right.

test/server.js
eliemichel: WebSocket tests do not intend to test for authentication, so they expect to operate on a public instance.

The problem is that I had to override the fixture settings for the LDAP tests, which are by definition incompatible with public instances. I don't know what's best, maybe it'd be better to have a different fixture for this? Any way, I saw that some other tests put their config assumptions in their `before()` hook, so I did the same. I think this is reasonable: as soon as a tests assumes some config lines, it specifies them.

src/plugins/auth/local.js
eliemichel: @astorije See https://github.com/eliemichel/lounge/commit/dfa5bffff6c3dfb37cc0c8a7806e3a28c18e1463

src/plugins/auth/ldap.js
xPaw: It should be fine to keep it the way it is.

src/plugins/auth/ldap.js
xPaw: I agree with @eliemichel, function is fine for future.

src/server.js
```diff
@@ -491,11 +441,16 @@ function performAuthentication(data) {
...
 
 	// Perform password checking
-	if (!Helper.config.public && Helper.config.ldap.enable) {
-		ldapAuth(client, data.user, data.password, authCallback);
-	} else {
-		localAuth(client, data.user, data.password, authCallback);
+	let auth = () => {
+		log.error("None of the auth plugins is enabled");
+	};
+	for (let i = 0 ; i < authPlugins.length ; ++i) {
```
xPaw: Can you remove the spaces before semilicons please.

src/plugins/auth/ldap.js
astorije: I like that, thanks for clarifying!
In general, I prefer keeping things simple and augmenting capabilities when it's needed, but it's not like it's adding a ton of boilerplate so that's totally fine.

src/plugins/auth/local.js
astorije: I get what you're coming from, and I agree with most of it, except with the idea of an ultimate default one. I think that if you set up The Lounge to use LDAP, and it actually uses local auth as a fallback, it could be bad.
If you want to use an auth mechanism and only this one, but it fails for some reason, then The Lounge should be failing instead. It's unlikely that you have set up local auth with correct users if you are intending to use LDAP anyway.

But we're thinking way ahead so I'm fine with the current state :)
There is no notion of "not available" for LDAP in this PR, I think, so it doesn't matter.

src/server.js
eliemichel: You should add `semi-spacing: [2, {"before": false, "after": true}]` to the `.eslintrc` then. ;) ([doc](https://eslint.org/docs/rules/semi-spacing))

eliemichel:
Squashed down to 9 commits, including the merge commit, and `803cff9` and `32e1a36` that I want to keep independent (not polluted with lint related things). Ready for merging.

MaxLeiter:
Fantastic work @eliemichel -- thanks! 👏

gramakri:
Just wanted to confirm that LDAP changes work great with Cloudron. We have updated the Cloudron package to use the searchDN mechanism.

astorije:
That's awesome, good to know @gramakri!

astorije:
Hey @eliemichel, we have sticker packs for our contributors now!

PR #1479, opened 2017-08-31
Add support for binding to unix sockets

### LINKED ISSUE #686: Make it possible to listen to UNIX socket file
Offering services to the world by listening some TCP high port is not particularly secure, since a hostile unprivileged user can replace the server with a hostile server that steals user's passwords shall the actual server crash.

While this can be prevented with SELinux, it is quite a heavy-weight solution. My proposal is to allow listening of UNIX sockets. Then the reverse proxy or xinetd could be configured to redirect connections to that socket.

xPaw:
Fixes #686 and #691. cc @ilmaisin 

Cleanup is handled in #1477.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -83,18 +83,30 @@ module.exports = function() {
 		}, app);
 	}
 
-	server.listen({
-		port: config.port,
-		host: config.host,
-	}, () => {
-		const protocol = config.https.enable ? "https" : "http";
-		var address = server.address();
+	let listenParams;
 
-		log.info(
-			"Available at " +
-			colors.green(`${protocol}://${address.address}:${address.port}/`) +
-			` in ${colors.bold(config.public ? "public" : "private")} mode`
-		);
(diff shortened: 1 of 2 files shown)


src/server.js
```diff
@@ -83,18 +83,30 @@ module.exports = function() {
...
+		listenParams = config.host.replace(/^unix:/, "");
+	} else {
+		listenParams = {
+			port: config.port,
+			host: config.host,
+		};
+	}
+
+	server.listen(listenParams, () => {
+		if (typeof listenParams === "string") {
```
astorije: It would be a bit clearer if both types were an object with a property `type`, so that you could do `if (listenParams.type === ...)`.
Alternatively, both objects with different properties and `if (listenParams.socket)`.

xPaw: Ideally yes, but `server.listen` can't accept an object like that.

astorije: Ah, gotcha. Somehow I read this object as something different but I see now. All good.

PR #1485, opened 2017-09-01
Always create condensed wrapper

xPaw:
It would not create a wrapper for new channels and when the date changed.

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -35,37 +35,17 @@ function buildChannelMessages(chanId, chanType, messages) {
 }
 
 function appendMessage(container, chanId, chanType, msg) {
+	let lastChild = container.children(".msg, .date-marker-container").last();
 	const renderedMessage = buildChatMessage(chanId, msg);
 
 	// Check if date changed
-	let lastChild = container.find(".msg").last();
 	const msgTime = new Date(msg.time);
-
-	// It's the first message in a window,
-	// then just append the message and do nothing else
-	if (lastChild.length === 0) {
-		container
-			.append(templates.date_marker({msgDate: msgTime}))
-			.append(renderedMessage);
-
-		return;
-	}
-
-	const prevMsgTime = new Date(lastChild.attr("data-time"));
-	const parent = lastChild.parent();
-
-	// If this message is condensed, we have to work on the wrapper
-	if (parent.hasClass("condensed")) {
-		lastChild = parent;
-	}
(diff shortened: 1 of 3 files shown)


client/js/render.js
```diff
@@ -28,22 +28,11 @@ function buildChannelMessages(chanId, chanType, messages) {
...
 
 function appendMessage(container, chanId, chanType, msg) {
+	let lastChild = container.find(".msg").last();
 	const renderedMessage = buildChatMessage(chanId, msg);
 
 	// Check if date changed
-	let lastChild = container.find(".msg").last();
 	const msgTime = new Date(msg.time);
-
-	// It's the first message in a window,
```
YaManicKill: Ah yes, this would be an issue after removing the logic for not creating a condense if there is only 1 message, as it woudl never be created until we had a normal message.

client/js/render.js
```diff
@@ -54,11 +43,8 @@ function appendMessage(container, chanId, chanType, msg) {
 
 	// Insert date marker if date changed compared to previous message
 	if (prevMsgTime.toDateString() !== msgTime.toDateString()) {
-		lastChild.after(templates.date_marker({msgDate: msgTime}));
-
-		// If date changed, we don't need to do condensed logic
-		container.append(renderedMessage);
-		return;
+		lastChild = $(templates.date_marker({msgDate: msgTime}));
```
YaManicKill: Yeah, this could be the issue. It's certainly not a great plan just now.

xPaw: What do you mean?

YaManicKill: Sorry, that's a very confusing sentence.

I do this the code that was here could be the issue we were seeing with @Jay2k1 and what we were doing doesn't actually make any sense to do, because we are already checking if the last thing is a condensed message anyway. 

But I have just realised why this won't work actually.

https://github.com/thelounge/lounge/blob/98e3bd0ca278fbcd8b120fbdaf6dfe4ec58633e7/client/views/date-marker.tpl

The date marker isn't a ".msg". So we'll miss the date marker, condense into the previous condensed, and the order will be wrong, no?

xPaw: That is possible.. I will look further into it later.

xPaw:
@YaManicKill I think I fixed the issue, and simplified the code a bit, so there's no longer need to check for parent element.

astorije:
Nevermind, fixed that for ya.

@YaManicKill, feel free to merge if your test appears conclusive.

PR #1490, opened 2017-09-02
Bump default image prefetch limit

astorije:
512kb is fine for thumbnails but really low for images so bumping it to 4MB by default.


dgw:
Yes, 512kB is small for photos, but 4MB might be going a bit on the high side. I've been happy with 2MB myself.

xPaw:
I agree, 4MB is too much for a default. High res 4k wallpaper in png is less than 2mb.

astorije:
Most of my jpg are 3mb and my DSLR is not recent but that's fine, 2 is still better than 0.5. @xPaw, mind amending my PR to 2mb please? I'm without a laptop for a few days

PR #1491, opened 2017-09-02
Focus a channel by joining it, refactor user commands #1189

realies:
This should work for focusing channels by joining them, the switch statement might need further refactoring. Thoughts?


client/js/lounge.js
```diff
@@ -194,19 +195,27 @@ $(function() {
...
-		}
-
-		if (text.indexOf("/expand") === 0) {
-			$(".chan.active .toggle-button:not(.opened)").click();
-			return;
+		if (text.indexOf("/") === 0) {
+			const separatorPos = text.indexOf(" ");
+			const cmd = text.substring(1, separatorPos > 1 ? separatorPos : text.length);
+			const parameters = separatorPos > text.indexOf(cmd) ? text.substring(text.indexOf(cmd) + cmd.length + 1, text.length) : "";
+			switch (cmd) {
```
xPaw: This can be simplified like `if (cmd in modules)` or `if (typeof modules[cmd] === "function")`

client/js/modules.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
```
YaManicKill: Why is this stuff not just in utils?

client/js/lounge.js
```diff
@@ -194,19 +195,27 @@ $(function() {
...
-			$(".chan.active .toggle-button.opened").click();
-			return;
-		}
-
-		if (text.indexOf("/expand") === 0) {
-			$(".chan.active .toggle-button:not(.opened)").click();
-			return;
+		if (text.indexOf("/") === 0) {
+			const separatorPos = text.indexOf(" ");
+			const cmd = text.substring(1, separatorPos > 1 ? separatorPos : text.length);
```
YaManicKill: `.split(" ")` is probably nicer.

client/js/modules.js
```diff
@@ -0,0 +1,35 @@
...
+	utils.clear();
+}
+
+function collapse() {
+	$(".chan.active .toggle-button.opened").click();
+}
+
+function expand() {
+	$(".chan.active .toggle-button:not(.opened)").click();
+}
```
astorije: These are used only once, I'd rather not have the extra level of indirection.

client/js/modules.js
```diff
@@ -0,0 +1,35 @@
...
+module.exports = {
+	clear,
+	collapse,
+	expand,
+	join
+};
+
+function clear() {
+	utils.clear();
+}
```
astorije: This seems like overkill 😅

client/js/modules.js
astorije: ~I agree.~ Actually, all those are just used once, I think this entire file/refactor is not wanted.

client/js/modules.js
```diff
@@ -0,0 +1,35 @@
...
+	$(".chan.active .toggle-button:not(.opened)").click();
+}
+
+function join(channel) {
+	var chan = utils.findCurrentNetworkChan(channel);
+
+	if (chan.length) {
+		chan.click();
+		return true;
+	}
```
astorije: Okay so _this_ is what this PR/commit is about :)
Whenever possible, try to separate refactors from actual changes in different commits, that makes it for a better history and easier review :)

astorije:
@realies, are you willing/able to pursue work on this? :)

client/js/lounge.js
realies: Please elaborate.

client/js/lounge.js
```diff
@@ -184,14 +184,23 @@ $(function() {
...
-		}
-
-		if (text.indexOf("/expand") === 0) {
-			$(".chan.active .toggle-preview:not(.opened)").click();
-			return;
+		if (text.indexOf("/") === 0) {
+			const args = text.substr(1).split(" ");
+			const cmd = args.shift().toLowerCase();
+			if (typeof utils.inputCommands[cmd] === "function") {
+				if (cmd === "join") {
```
xPaw: Just pass the args into the command function, don't do the if check.

Join command can just use `args[0]`.

client/js/lounge.js
```diff
@@ -184,14 +184,14 @@ $(function() {
...
-		}
-
-		if (text.indexOf("/expand") === 0) {
-			$(".chan.active .toggle-preview:not(.opened)").click();
-			return;
+		if (text.indexOf("/") === 0) {
+			const args = text.substr(1).split(" ");
+			const cmd = args.shift().toLowerCase();
+			if (typeof utils.inputCommands[cmd] === "function") {
+				if (utils.inputCommands[cmd](args)) {
```
xPaw: You can combine the ifs with `&&` to reduce nesting.

client/js/lounge.js
YaManicKill: Sorry, I never did elaborate, but it looks like you figured out what I meant.

client/js/utils.js
```diff
@@ -29,6 +43,27 @@ function forceFocus() {
 	input.trigger("click").focus();
 }
 
+function collapse() {
```
YaManicKill: (I know I said the opposite of this previously, so this is not for this PR...it's for the future)

If we get more than this, we will probably want to split them up out of `utils.js`, but not into just another file. Probably have a directory with a file per command. But as I say, not for this PR.

PR #1492, opened 2017-09-03
One line server startup errors

xPaw:
Fixes this:

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -94,6 +94,8 @@ module.exports = function() {
 		};
 	}
 
+	server.on("error", (err) => log.error(`${err}`));
+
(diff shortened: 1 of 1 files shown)


astorije:
Hey @xPaw, could you post an "after" screenshot for this error? Curious what it'd look like.

xPaw:
`timestamp [ERROR] Error: listen EADDRINUSE :::9000`

astorije:
Does this correctly kill process when such error happens though?

xPaw:
It dies on its own, yes. The only thing keeping the server running is `server.listen`, so if that fails it just quits.

astorije:
Sounds good!

PR #1495, opened 2017-09-04
Only change nick autocompletion when receiving a message

xPaw:
Loading more history would mess with autocompletion order (it would also sort for each message when loading the page, which is completely unnecessary).

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -36,7 +36,7 @@ function buildChannelMessages(chanId, chanType, messages) {
 
 function appendMessage(container, chanId, chanType, msg) {
 	let lastChild = container.children(".msg, .date-marker-container").last();
-	const renderedMessage = buildChatMessage(chanId, msg);
+	const renderedMessage = buildChatMessage(msg);
 
 	// Check if date changed
 	const msgTime = new Date(msg.time);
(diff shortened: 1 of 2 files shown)


client/js/render.js
```diff
@@ -44,7 +44,7 @@ function appendMessage(container, chanId, chanType, msg) {
 
 	// Insert date marker if date changed compared to previous message
 	if (prevMsgTime.toDateString() !== msgTime.toDateString()) {
-		lastChild = $(templates.date_marker({msgDate: msg.time}));
+		lastChild = $(templates.date_marker({time: msg.time}));
```
xPaw: Missed this in previous PR.

client/js/render.js
```diff
@@ -64,11 +64,7 @@ function appendMessage(container, chanId, chanType, msg) {
 	}
 
 	// Always create a condensed container
-	const newCondensed = buildChatMessage(chanId, {
-		type: "condensed",
-		time: msg.time,
-		previews: []
-	});
+	const newCondensed = $(templates.msg_condensed({time: msg.time}));
```
xPaw: There's no need to build a complete message since it uses a separate template so we can just optimize it.

client/js/render.js
```diff
@@ -77,12 +73,6 @@ function appendMessage(container, chanId, chanType, msg) {
 
 function buildChatMessage(chanId, msg) {
```
YaManicKill: ChanId should be removed... which will create a fun diff. Feel free to leave it out of this PR for that reason.

xPaw: It's called in a single place, so it's easy.

PR #1503, opened 2017-09-06
Handle browser history when opening/closing image preview

### LINKED ISSUE #1497: Make image viewer push state to history
When using the built-in image viewer on the phone, I always instinctively press the back button trying to close it, but the image stays open, I just go to the previous channel underneath. I don't know if I am the only one thinking the back button should close the image and take me _back_, but I think it's a good idea.

astorije:
Fixes #1497.

~This doesn't work too great, and I'm not sure why. This code itself seems to work fine, but when interacting with other history-aware elements, or when reaching initially-loaded page, things go wrong.
Any suggestions?~
All fixed.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -336,8 +336,6 @@ $(function() {
 			state.clickTarget = `#footer button[data-target="${target}"]`;
 		}
 
-		state.clickTarget += ", #image-viewer";
-
 		if (history && history.pushState) {
 			if (data && data.replaceHistory && history.replaceState) {
 				history.replaceState(state, null, null);
@@ -809,8 +807,23 @@ $(function() {
 			return;
 		}
 
-		const {clickTarget} = state;
+		let {clickTarget} = state;
+
 		if (clickTarget) {
+			// This will be true when click target corresponds to opening a thumbnail,
+			// browsing to the previous/next thumbnail, or closing the image viewer.
+			const imageViewerRelated =
+				clickTarget === "#image-viewer" ||
+				clickTarget.includes(".toggle-thumbnail");
+
+			// If the click target is not related to the image viewer but the viewer
+			// is currently opened, we need to close it.
+			if (!imageViewerRelated && $("#image-viewer").hasClass("opened")) {
+				clickTarget += ", #image-viewer";
+			}
+
+			// Emit the click to the target, while making sure it is not going to be
+			// added to the state again.
 			$(clickTarget).trigger("click", {
 				pushState: false
 			});
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -336,6 +336,8 @@ $(function() {
 			state.clickTarget = `#footer button[data-target="${target}"]`;
 		}
 
+		state.clickTarget += ", #image-viewer";
+
 		if (history && history.pushState) {
 			if (data && data.replaceHistory && history.replaceState) {
 				history.replaceState(state, null, null);
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -331,7 +331,7 @@ $(function() {
 		const state = {};
 
 		if (self.hasClass("chan")) {
-			state.clickTarget = `.chan[data-id="${self.data("id")}"]`;
+			state.clickTarget = `#sidebar .chan[data-id="${self.data("id")}"]`;
 		} else {
 			state.clickTarget = `#footer button[data-target="${target}"]`;
 		}
@@ -812,9 +812,7 @@ $(function() {
 		if (clickTarget) {
 			// This will be true when click target corresponds to opening a thumbnail,
 			// browsing to the previous/next thumbnail, or closing the image viewer.
-			const imageViewerRelated =
-				clickTarget === "#image-viewer" ||
-				clickTarget.includes(".toggle-thumbnail");
+			const imageViewerRelated = clickTarget.includes(".toggle-thumbnail");
 
 			// If the click target is not related to the image viewer but the viewer
 			// is currently opened, we need to close it.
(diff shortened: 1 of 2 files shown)


astorije:
Well, I do know why actually, just not set on how to fix it yet: when hitting Back, no state is being added to specify image viewer is closed, so things switch in the background.

I know I have several ways to address this, but not sure what's the best way. Should I add a `imageViewerOpen` boolean to the history state? Should I make sure `#image-viewer` gets clicked every time something else is clicked?

astorije:
Okay, after lengthy discussions with @xPaw, I think I can finally call this ready. It was way more painful than anticipated...

xPaw:
@astorije I tested this, and I definitely think that closing the imageviewer should just pop it out of history.

1. Open image viewer
2. Close it (press x)
3. Switch to another channel
4. Press back
   1. This appears to do nothing (bug?)
5. Press back again
   1.  It opens the image viewer

astorije:
> i. This appears to do nothing (bug?)

Yeah it was, because essentially closing the image viewer would not record where we were before. I fixed it, it's acting much more like you'd expect now.

> I definitely think that closing the imageviewer should just pop it out of history.

This is a horrible, horrible experience and I really don't want that. This is why:

- You are on channel A
- Switch to a channel B with multiple thumbnails thumbnails
- Open image 1 on this channel
- Click on the "Next image" button <kbd>></kbd> to view the next image
- You want to go back to the previous image, so naturally click on the "Back" button (especially true on mobiles, where this <kbd>◁</kbd> system button is always available, but also on desktop where clicking "Back" for previous item comes naturally).
- The image viewer is being closed instead of sending you to the previous image.
- Because you pressed "Back" and it did something you didn't expect, you naturally try to hit "Forward" but there is no "Forward" button since we popped. If you hit Back thinking that might bring the image back, that takes you from channel B back to channel A.
- You throw your phone across your living room, cursing at The Lounge 😅 

Essentially, for an intuitive and predictable experience when it comes to "Back"/"Forward" buttons, there are 2 things we should respect:

- **"Back" != "Close"**: If we're talking about a modal, and that modal has only one "step", then yes, going back roughly means closing. But if your modal has a "Continue"/"Next »" kind of action and user clicks it, going back means going to the previous step and not closing the modal like "Close" buttons would do.
- **"Back" ➡️ "Forward"**: If I click "Back", I should be able to revert this by clicking "Forward". It is very frustrating to click "Back" then realize the "Forward" button is disabled and I cannot replay to where I was.

client/js/lounge.js
```diff
@@ -331,7 +331,7 @@ $(function() {
 		const state = {};
 
 		if (self.hasClass("chan")) {
-			state.clickTarget = `.chan[data-id="${self.data("id")}"]`;
+			state.clickTarget = `#sidebar .chan[data-id="${self.data("id")}"]`;
```
astorije: This is just a safeguard as I realize that `.chan[data-id="${self.data("id")}"]` actually refers to 2 things, the sidebar element and the chat window element. Triggering a click on the chat window element does nothing so we're safe for now, but if this changes someday we'd have some weird behavior to debug.
Also there is no reason to click on it so might as well be selective.

YaManicKill:
Here's a question...

Why does the image view support going back and forward anyway? Does anyone actually use this? For me, the vast majority of interactions are as follows:

1. Click on image I want to view
2. Click Esc (if on desktop), click back (if on mobile)
3. That's it

Like, there are even very few circumstances where I would actually see multiple images on the chat, nevermind ones that are at all related. 

The main thing is that if I press back on mobile, it takes me away from the image viewer, and the only way I can get back to the image viewer is by clicking on the image again (or, maybe clicking forward in the menu). All the other things I've seen you mention are not things I would expect to happen.

astorije:
> Why does the image view support going back and forward anyway?

Because of my last 2 points in the previous comment: back means back and not close. I totally understand why/how you use it on mobile (and good news is that what you describe works exactly like that with this PR), but yeah, it's an OS-wide "Back" button, not "Close".

Forward is less often used on mobile than on desktop (in fact, you can't even forward on Android when The Lounge is run outside of the mobile browser), but it is an anti-pattern to provide not silence a "Forward" action after clicking "Back". There are tons of use cases I could cite, but the common rule is to not surprise the user by altering history.
It is like when an "Undo" action is not done correctly and the "Redo" actually does something else (I'm looking at you, Atlassian!). So in a nutshell, A ➡️ Back (n times) ➡️ Forward (n times) should bring you back to A, just like "A Given Text" ➡️ Undo (n times) ➡️ Redo (n times) should aways gets you back on "A Given Text".
Sorry if I'm being pedantic with the examples, I just want to make sure I'm not being confusing.

> Like, there are even very few circumstances where I would actually see multiple images on the chat, nevermind ones that are at all related.

:man_shrugging: 
There are lots of reasons to go back-and-forth between images, like a photography channel, or a team channel where you share Before/After screenshots of a feature, etc.
I know I'm using the Previous/Next buttons quite a bit!

> The main thing is that if I press back on mobile, it takes me away from the image viewer

That's what this PR does, unless you went browsing Previous/Next pictures (because of the Back vs. Close distinction).

> the only way I can get back to the image viewer is by clicking on the image again (or, maybe clicking forward in the menu)

That's what this PR does, clicking the "Forward" button reopens the image viewer (unless you click somewhere else, but that's just standard behavior).

> All the other things I've seen you mention are not things I would expect to happen.

I'm not sure, to me it looks like this PR does pretty much what you describe :))

---

Anyway, I hope this clarifies things, @YaManicKill.
I really _really_ think this is the most optimal experience, and more importantly, the most predictable and natural way to browse Back. At the very least, it solves the main issue which is: open an image, then press the Back button to close it.
I'd very much like to put this one behind me/us 😅

PR #1508, opened 2017-09-06
Render link previews in browser idle event

### LINKED ISSUE #1504: Link previews do not always get inserted on master
I am currently on 65e4e386da7cd1d16e4c00abb0c44cd4e2e517db
as soon as I updated I noticed that some link previews fail. The preview div is there, but empty. No toggle button. No errors.
I am not too sure how to reproduce it, it could be cause by having the window open but the tab unfocused. When I was in another channel, the preview in the first channel loaded just fine. When I myself send a link, that preview also works. 
This is also unrelated to the browser, as it appeared on both my personal (chrome) and my work pc (firefox).

xPaw:
Fixes #1504

### DIFF
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -3,17 +3,12 @@
 const $ = require("jquery");
 const socket = require("../socket");
 const render = require("../render");
+const utils = require("../utils");
 const chat = $("#chat");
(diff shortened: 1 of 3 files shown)


astorije:
Small bugfix, merging.

PR #1511, opened 2017-09-09
Checkered background for transparent images in image viewer

xPaw:


astorije:
So, I get why, but I don't know, I really like that it keeps the transparent backgrounds transparent. It's like a good use case for actually having a transparent background.

screen shot 2017-09-09 at 10 44 43

@YaManicKill, thoughts?

xPaw:
@astorije Keep in mind that black images on transparent background do not show up at all in the image viewer.

astorije:
Yeah, so @xPaw just shared https://openclipart.org/image/2400px/svg_to_png/151735/1311627658.png with me on IRC and in this case, yeah, it feels like we have to do that.
Ah, too bad, I did like quite a lot the black background :man_shrugging:

PR #1512, opened 2017-09-09
Fix image viewer turning black sometimes

xPaw:


astorije:
Are you sure this fixes it? I have a hard time thinking this is what's causing it. But I haven't found a way to consistently reproduce this bug yet, just happens whenever it wants. Did you?

xPaw:
@astorije `#image-viewer { transition: opacity 2s; }` or `transition: none` causes it quite consistently.

astorije:
> @astorije #image-viewer { transition: opacity 2s; } or transition: none causes it quite consistently.

Indeed, I can repro almost every time now, good catch.
But I'm still not sure how this fixes it. Mind explaining? Does it have to do with https://github.com/thelounge/lounge/blob/master/client/js/renderPreview.js#L180 ?

xPaw:
> Does it have to do with 

Precisely with that. It somehow race conditions into being called when opacity finishes animating to 1.

astorije:
Yeah that makes sense and I pretty much figured it out right after asking lol. Mind adding a comment her so we don't wonder why the heck we needed this a few months from now?

xPaw:
@astorije Done.

MiniDigger:
as I originally reported this issue (I had transition: none in my theme) I can confirm this fixed it.

PR #1517, opened 2017-09-12
Move unread marker when loading more history

### LINKED ISSUE #835: Show unread marker in correct place when we load more
Realised there isn't an issue for this.

When the last unread message is further back than what we get sent to the client, we get the unread marker at the top of the page, but when we click "load more", the marker doesn't go any further back.

Main reason being that as soon as we change channel, the server marks everything as read. So we need to have a way to know what message in the past is unread.

xPaw:
Fixes #835

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -371,6 +371,7 @@ $(function() {
 
 		lastActiveChan
 			.find(".unread-marker")
+			.data("unread-id", 0)
 			.appendTo(lastActiveChan.find(".messages"));
 
 		var chan = $(target)
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -120,19 +120,25 @@ function renderChannelMessages(data) {
 	const documentFragment = buildChannelMessages(data.id, data.type, data.messages);
 	const channel = chat.find("#chan-" + data.id + " .messages").append(documentFragment);
 
-	if (data.firstUnread > 0) {
-		const first = channel.find("#msg-" + data.firstUnread);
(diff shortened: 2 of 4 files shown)


client/js/lounge.js
```diff
@@ -371,6 +371,7 @@ $(function() {
 
 		lastActiveChan
 			.find(".unread-marker")
+			.data("unread-id", 0)
```
astorije: This is an id, not a counter, correct? If so, any risk of the id actually being `0` and leading to weird bugs? Wondering if we couldn't use `-1` or `null` or similar exceptional value instead then.

astorije:
So, the unread marker still shows up at the "incorrect" position before the messages are loaded and the marker gets moved. Is there a way to _not_ show it in this case then?

YaManicKill:
> So, the unread marker still shows up at the "incorrect" position before the messages are loaded and the marker gets moved. Is there a way to not show it in this case then?

The good thing about showing it at the top is that it lets us know these are all unread messages. That's how most other chat apps that I know do it. We need some indication that everything in the channel is new, and the easiest way is to show it at the top and pop it up higher when we get more messages. Otherwise you might think you missed it.

I know this is less of an issue with the auto-fetching when you scroll to the top, but you will still see it for a short time when you get to the top. I use this feature in slack all the time.

client/js/lounge.js
xPaw: It checks for `> 0` when using this id. If the message is 0, there can't be any older messages by definition, so nothing will happen.

PR #1518, opened 2017-09-12
Remove unnecessary end() calls

xPaw:
It does absolutely nothing if it's the last call in the chain.

### DIFF
--- a/client/js/socket-events/auth.js
+++ b/client/js/socket-events/auth.js
@@ -34,13 +34,9 @@ socket.on("auth", function(data) {
 		return;
 	}
 
-	$("#footer").find(".sign-in")
+	$("#footer")
+		.find(".sign-in")
 		.trigger("click", {
 			pushState: false,
-		})
-		.end()
-		.find(".networks")
-		.html("")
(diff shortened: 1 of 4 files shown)


client/js/socket-events/auth.js
```diff
@@ -34,13 +34,9 @@ socket.on("auth", function(data) {
...
 	}
 
-	$("#footer").find(".sign-in")
+	$("#footer")
+		.find(".sign-in")
 		.trigger("click", {
 			pushState: false,
-		})
-		.end()
-		.find(".networks")
```
xPaw: There's no way this worked, there's no `.networks` in `#footer`.

astorije: Yeah, this seems to be an artifact of https://github.com/thelounge/lounge/commit/f90c355c8e5356673edb724e8ea29087950e6644#diff-d1e1df35c42cc0890274f60defa80028R35, before `#sidebar` got removed. Not sure what the consequences are, but as you pointed out, if this is buggy, then it's buggy on `master` as well.

PR #1519, opened 2017-09-12
Fix requesting last messages when no message id is known

### LINKED ISSUE #1460: "show older messages" does not work if the buffer is empty
Use `/clear` and click the button, it won't work.

xPaw:
Fixes #1460.

However there's another bug. When doing `/clear`, the show more button jumps into view, thus IntersectionObserver calls observe on the button, and it loads last history again.

### DIFF
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -56,12 +56,13 @@ socket.on("more", function(data) {
 });
 
 chat.on("click", ".show-more-button", function() {
-	var self = $(this);
-	var lastMessage = self.parent().next(".messages").children(".msg").first();
-	if (lastMessage.is(".condensed")) {
-		lastMessage = lastMessage.children(".msg").first();
+	const self = $(this);
+	const lastMessage = self.closest(".chat").find(".msg").first();
+	let lastMessageId = -1;
+
+	if (lastMessage.length > 0) {
+		lastMessageId = parseInt(lastMessage[0].id.replace("msg-", ""), 10);
 	}
-	var lastMessageId = parseInt(lastMessage[0].id.replace("msg-", ""), 10);
 
 	self
 		.text("Loading older messages…")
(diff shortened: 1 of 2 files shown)


YaManicKill:
> When doing /clear, the show more button jumps into view, thus IntersectionObserver calls observe on the button

Hahahaha. That's brilliant.

But that does lead us to a question ... what is the purpose of clear?

I would say either it would clear both client and server side, and therefore have no way to get the stuff back, or it is a pointless feature. Thoughts?

xPaw:
I don't know either, it sounds like it should clear server buffer too.

astorije:
> what is the purpose of clear?

`/clear` really acts like `clear`/<kbd>ctrl</kbd>+<kbd>L</kbd> in the terminal: it hides / empties the screen, whose content can be brought back by scrolling up.
I actually use it (infrequently) on The Lounge, but I would say a good 60% of the usage I'm making of it can now be achieved by typing `/collapse`.

I have tried a few things to reproduce this, and nothing worked. Ideally, we would want to be able to set the `scrollTop` past the maximum, so that latest message is right before the visible block but as far as I know, it's not possible on the web. I have seen a couple "clearable" apps that did that nicely, but they were all native.

Because this is not doable (unless someone has a genius idea), I'd be in favor of removing it.
In fact, even though I use the command myself, I would not like its action to be changed from "hide the latest message" to "destroy all messages in the current channel". That would be a harsh change of an existing feature 😅.
There should be a way to wipe out all the logs for a given channel (in fact, there might be an open issue about this) in a context menu entry for example, but this is a very destructive action that shouldn't be possible with an unfortunate shortcut/command accident lol.

So yeah, until then, I think we should just remove it if there is no way to fix it without drastically changing its behavior.

xPaw:
Regardless of this clear command bug, this PR fixes the issue on its own, the fact it loads history on clear is a different story, so this can be merged.

astorije:
That's fair. Is the `do not merge` label still relevant?

PR #1520, opened 2017-09-12
Add touch-action to Messages, Sidebar and User list

### LINKED ISSUE #1035: Look into adding touch-action to styles
https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners
https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action

This could improve performance and remove click delays. For example message and user list containers can only scroll up and down, so there is no need to handle vertical panning gestures.

Our `viewport` has `user-scalable="no"`, so we could get rid of `pinch-zoom` too.

iamstratos:
This commit fixes #1035 

On the issue, there is a reference to passive events for JS addEventListeners. I've checked all touchend-touchstart events on Master and passive events are already there. So, I only added touch-action css on some parts that have scroll.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -493,6 +493,7 @@ kbd {
 #sidebar .network,
 #sidebar .network-placeholder {
 	margin-bottom: 30px;
+	touch-action: pan-y;
 }
 
 #sidebar .empty {
@@ -935,6 +936,7 @@ kbd {
 
 #chat .messages {
 	padding: 10px 0;
+	touch-action: pan-y;
 }
 
 #chat .msg {
@@ -1296,6 +1298,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	position: absolute;
 	top: 48px;
 	width: 100%;
+	touch-action: pan-y;
 }
 
 #chat .names-filtered {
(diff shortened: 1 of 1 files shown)


xPaw:
@iamstratos Thanks for this!

Have you confirmed the impact of this change in DevTools? Do you think we need `touch-action: none` on body or `*`?

client/css/style.css
```diff
@@ -493,6 +493,7 @@ kbd {
 #sidebar .network,
 #sidebar .network-placeholder {
 	margin-bottom: 30px;
+	touch-action: pan-y;
```
xPaw: Shouldn't this be on `.networks`, or `.sidebar` even?

iamstratos: It may help to add this on `.networks`, it has no effect on `.sidebar` (it didn't when I was testing at least) but it's nice to have there to.

iamstratos:
@xPaw I've tested it with DevTools on desktop & mobile (android). I've read about touch-action on body and tested again with `touch-action: none` on body. It had no negative effects to any of my tests. On the other hand, `touch-action: none` on * had every touch disabled (disabled scroll everywhere too).

astorije:
@xPaw, mind taking review for this?

astorije:
Thanks!

astorije:
Hey @iamstratos, we have sticker packs for our contributors now!

Issue #1522, opened 2017-09-13
Wrong kicker mode displayed


dgw:
> +Kaede has kicked flameboi  (dgw done killed ya!)

In the kick message, the kicker is displayed as having the mode of the target (while the target has no mode character). I tested on `master` (at 2d61c01, specifically) on my PC, as well as with 2.4.0 in production. If I kick a voiced user, `+dgw has kicked _`. Kicking an op: `@dgw has kicked _`. As channel owner, I can also kick admins (`+a`): `&dgw has kicked _`.

I will experiment to try and find a way to have it use the correct mode, but I will submit a PR to "fix" this (along the lines of #1352) soon-ish if I don't hit on any solutions tonight. At the very least, the next release should stop displaying the wrong mode, if it can't display the correct one.

PR #1523, opened 2017-09-13
Remove mode from user_name displayed for kick

dgw:
Quick bodge over #1522 until the proper mode can be displayed. Better to show nothing than wrong data.


dgw:
Superseded by proper fix.

PR #1525, opened 2017-09-13
Correct mode in kick message

xPaw:
@dgw mind testing?


dgw:
The `from_mode` appears to be always empty with this patch.

FWIW I have a patch that displays both modes correctly in all tests so far. I can push it up tonight.

PR #1528, opened 2017-09-14
Move user log function where it belongs

xPaw:


src/userLog.js
```diff
@@ -18,7 +18,7 @@ module.exports.write = function(user, network, chan, msg) {
 	var format = Helper.config.logs.format || "YYYY-MM-DD HH:mm:ss";
 	var tz = Helper.config.logs.timezone || "UTC+00:00";
 
-	var time = moment().utcOffset(tz).format(format);
+	var time = moment(msg.time).utcOffset(tz).format(format);
```
xPaw: Fixes `server-time`.

astorije:
@xPaw, just curious, does this make #663 any more convoluted?

xPaw:
@astorije Should be an easy rebase.

src/userLog.js
YaManicKill: Hmmm, does it? Won't this still use the weird utc offset (which defaults to UTC)?

xPaw: It fixes the fact it always used current time when writing log instead of the server time. So connecting to znc and receiving buffer would log 500 lines all having current time, instead of actual time when message was sent.

YaManicKill: Ah right, that was a different thing than what I thought you were saying it fixed. Cool, good catch.

PR #1529, opened 2017-09-14
Simple search backend in sqlite

florentsolt:
Hi,

I did a simple implementation of a search feature using sqlite.
It's quite straightforward, the sqlite database is created on the the fly and the indexing process (which start with the server) will add new log entries every 10s (to prevent to much stress).
So yes, it's linked to log feature.

There is also a handy command line interface to search.

It seems working properly but did not had the opportunity to test it expect for my personal use case.

Happy to have any feedbacks.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -37,6 +37,7 @@
 		</aside>
 		<footer id="footer">
 			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Sign in"><button class="icon sign-in" data-target="#sign-in" aria-label="Sign in"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Search"><button class="icon search" data-target="#search-query" aria-label="Search"></button></span>
 			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Connect to network"><button class="icon connect" data-target="#connect" aria-label="Connect to network"></button></span>
 			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Client settings"><button class="icon settings" data-target="#settings" aria-label="Client settings"></button></span>
 			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Help"><button class="icon help" data-target="#help" aria-label="Help"></button></span>
@@ -197,6 +198,49 @@
 							</div>
 						</div>
 					</form>
+				</div>
+				<div id="search-query" class="window">
+					<div class="header">
+						<button class="lt" aria-label="Toggle channel list"></button>
+					</div>
+					<form class="container" method="post" action="">
+						<div class="row">
+							<div class="col-sm-12">
+								<h1 class="title">Search</h1>
+							</div>
+							<div class="col-sm-12">
+								<h2>Query</h2>
+							</div>
+							<div class="col-sm-4">
+								<label for="text" class="sr-only">User</label>
+								<input type="text" id="nick" name="nick" class="input" placeholder="Everyone">
+							</div>
+							<div class="col-sm-8">
+								<label for="text" class="sr-only">Enter text to match</label>
+								<input type="text" id="text" name="text" class="input" placeholder="Enter text to match">
+							</div>
+							<div class="col-sm-12">
+								<label for="network-channel-select" class="sr-only">Network and Channel</label>
+								<select id="network-channel-select" name="network-channel" class="input">
+									<option disabled selected value="">Loading...</option>
+								</select>
+							</div>
+							<div class="col-sm-4">
+								<label for="date-from" class="sr-only">From</label>
+								<input type="text" id="date-from" name="from" class="input" placeholder="From: 7 days">
+							</div>
+							<div class="col-sm-4">
+								<label for="date-to" class="sr-only">To</label>
+								<input type="text" id="date-to" name="to" class="input" placeholder="To: now">
+							</div>
+							<div class="col-sm-4">
+								<button type="submit" class="btn pull-right">Search</button>
+							</div>
+						</div>
+					</form>
+					<div class="results" id="chat">
+						<div class="channel messages"></div>
+					</div>
 				</div>
 				<div id="settings" class="window" data-type="settings">
 					<div class="header">
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -25,6 +25,7 @@ require("./clipboard");
 $(function() {
 	var sidebar = $("#sidebar, #footer");
 	var chat = $("#chat");
+	var searchQuery = $("#search-query");
 
 	$(document.body).data("app-name", document.title);
 
@@ -329,6 +330,10 @@ $(function() {
 		}
 	});
 
+	sidebar.on("click", "button.search", function() {
+		socket.emit("search-networks-channels");
+	});
+
 	sidebar.on("click", ".chan, button", function() {
 		var self = $(this);
 		var target = self.data("target");
@@ -485,7 +490,14 @@ $(function() {
 		container.html(templates.user_filtered({matches: result})).show();
 	});
 
-	var forms = $("#sign-in, #connect, #change-password");
+	searchQuery.on("click", ".show-more-button", function() {
+		$(this).text("Loading older messages…").prop("disabled", true);
+		var query = searchQuery.find("form").data("lastQuery");
+		query.offset = searchQuery.find(".highlight").length;
+		socket.emit("search-query", query);
+	});
+
+	var forms = $("#sign-in, #connect, #change-password, #search-query");
 
 	windows.on("show", "#sign-in", function() {
 		$(this).find("input").each(function() {
@@ -533,6 +545,8 @@ $(function() {
 			event = "conn";
 		} else if (form.closest("div").attr("id") === "change-password") {
 			event = "change-password";
+		} else if (form.closest("div").attr("id") === "search-query") {
+			event = "search-query";
 		}
 
 		var values = {};
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -11,17 +11,19 @@ const condensed = require("./condensed");
 const helpers_parse = require("./libs/handlebars/parse");
 
 const chat = $("#chat");
+const searchQuery = $("#search-query");
 const sidebar = $("#sidebar");
 
 require("intersection-observer");
 
 const historyObserver = window.IntersectionObserver ?
 	new window.IntersectionObserver(loadMoreHistory, {
-		root: chat.get(0)
+		root: $("#main").get(0)
 	}) : null;
 
 module.exports = {
 	appendMessage,
+	appendSearchResults,
 	buildChannelMessages,
 	renderChannel,
 	renderChannelUsers,
@@ -74,6 +76,14 @@ function appendMessage(container, chanId, chanType, msg) {
 	condensed.updateText(newCondensed, [msg.type]);
 	newCondensed.append(renderedMessage);
 	container.append(newCondensed);
+}
+
+function appendSearchResults(container, results) {
+	container.append(templates.search_results({results: results}));
+
+	if (historyObserver) {
+		historyObserver.observe(searchQuery.find(".show-more").get(0));
+	}
 }
 
 function buildChatMessage(msg) {
--- a/client/js/socket-events/search.js
+++ b/client/js/socket-events/search.js
@@ -1,0 +1,42 @@
+"use strict";
+
+const $ = require("jquery");
+const socket = require("../socket");
+const render = require("../render");
+const messages = $("#search-query .results .messages");
+const form = $("#search-query form");
+const select = $("#search-query form select");
+
+socket.on("search-networks-channels", function(data) {
+	var groups = {};
+	select.empty();
+	select.append($("<option></option>").attr("value", "*/*").append("** All networks and channels **"));
+	data.forEach(function(entry) {
+		if (!groups[entry.network]) {
+			groups[entry.network] = [];
+		}
+		if (entry.network !== entry.chan) {
+			groups[entry.network].push(entry.chan);
+		}
+	});
+	Object.keys(groups).forEach(function(network) {
+		var group = $("<optgroup></optgroup>").attr("label", network);
+		group.append($("<option></option>").attr("value", network + "/*").append("** All channels **"));
+		group.append($("<option></option>").attr("value", network).append("** Lobby **"));
+		groups[network].forEach(function(chan) {
+			group.append($("<option></option>").attr("value", network + "/" + chan).append(chan));
+		});
+		select.append(group);
+	});
+});
+
+socket.on("search-results", function(query, results) {
+	if (!query.offset) {
+		messages.empty();
+		form.data("lastQuery", query);
+	} else {
+		messages.find(".show-more").remove();
+	}
+	form.find(".btn").prop("disabled", false);
+	render.appendSearchResults(messages, results);
+});
--- a/client/views/search_results.tpl
+++ b/client/views/search_results.tpl
@@ -1,0 +1,21 @@
+{{#each results}}
+
+	<div class="date-marker">
+		<span class="date-marker-text" data-label="{{network}}"></span>
+		<span class="date-marker-text" data-label="{{chan}}"></span>
+		<span class="date-marker-text" data-label="{{friendlydate time}}"></span>
+	</div>
+	{{#each before}}
+		{{> msg}}
+	{{/each}}
+
+	{{> msg}}
+
+	{{#each after}}
+		{{> msg}}
+	{{/each}}
+{{/each}}
+
+<div class="show-more {{#equal results.length 20}}show{{/equal}}">
+	<button class="show-more-button">Show older messages</button>
+</div>
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -506,5 +506,5 @@ module.exports = {
 		// @default  false
 		//
 		raw: false,
-	},
+	}
 };
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "bcryptjs": "2.4.3",
+    "bluebird": "3.5.0",
     "cheerio": "0.22.0",
     "colors": "1.1.2",
     "commander": "2.11.0",
@@ -47,6 +48,7 @@
     "fs-extra": "4.0.2",
     "irc-framework": "2.9.1",
     "ldapjs": "1.0.1",
+    "line-by-line": "0.1.5",
     "lodash": "4.17.4",
     "moment": "2.19.1",
     "package-json": "4.0.1",
@@ -57,7 +59,8 @@
     "spdy": "3.4.7",
     "ua-parser-js": "0.7.17",
     "urijs": "1.19.0",
-    "web-push": "3.2.4"
+    "web-push": "3.2.4",
+    "sqlite3": "3.1.11"
   },
   "devDependencies": {
     "babel-core": "6.26.0",
--- a/src/command-line/search.js
+++ b/src/command-line/search.js
@@ -1,0 +1,48 @@
+"use strict";
+
+const program = require("commander");
+const Utils = require("./utils");
+const Search = require("../search");
+const space = "   ";
+
+program
+	.command("search")
+	.option("-i, --index", "index all available log files")
+	.option("-f, --from <date>", "search after the <date> in the index")
+	.option("-t, --to <date>", "search before the <date> in the index")
+	.option("-n, --nick <text>", "search for <nick> in the index")
+	.option("-q, --text <text>", "search for <text> in the index")
+	.option("-u, --user <name>", "restrict search to a user")
+	.option("-n, --network <name>", "restrict search to a network")
+	.option("-c, --chan <name>", "restrict search to a chan")
+	.description("Manage search")
+	.on("--help", Utils.extraHelp)
+	.action((options) => {
+		if (options.index) {
+			Search.indexAll();
+		} else {
+			Search.query(options)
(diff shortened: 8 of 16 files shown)


client/js/render.js
```diff
@@ -19,8 +20,14 @@ const historyObserver = window.IntersectionObserver ?
 		root: chat.get(0)
 	}) : null;
 
+const searchObserver = window.IntersectionObserver ?
+	new window.IntersectionObserver(loadMoreHistory, {
```
xPaw: Shouldn't that have been `appendSearchResults`?

Why do you need a separate observer anyway?

client/views/search_results.tpl
```diff
@@ -0,0 +1,69 @@
...
+				<span class="text">{{{parse text}}}</span>
+
+				{{#each previews}}
+					<div class="preview" data-url="{{link}}"></div>
+				{{/each}}
+			</span>
+		</div>
+	{{/each}}
+
+	<div class="msg highlight" data-time="{{time}}" data-id="{{id}}">
```
xPaw: Can you just use `msg` template?

defaults/config.js
```diff
@@ -506,4 +506,12 @@ module.exports = {
 		//
 		raw: false,
 	},
+
+	search: {
+		enable: true,
+		driver: "sqliet3",
```
xPaw: Typo.

YaManicKill:
Hi, @florentsolt,

Thanks for this. We'll take a look at it soon. I have just had a quick look over the code, and it looks like you are calling the logs themselves to get the data to store into your database. Would it be better to have an event that the server sends out when it receives a message, and the database can store that without having to read the log files. (also in the future we could also expand that out to having the file logger using the same events).

Just a thought. Might be more performant.

xPaw:
I agree with @YaManicKill, this code can be simplified quite a bit by removing the "indexing" part, and only implementing sqlite storage going forward. 

We *can* have a one-time conversion that would convert logs into sqlite, but this does not need to happen every server start.

florentsolt:
Thanks all for your feedbacks.

for @xPaw: 
- I'll improve the observer code
- I'll use the msg template
- Fix the typo (remove that btw, it was an idea I had before, useless now) and add some comments
- Any idea how I can avoid declaring a second div#chat to inherit css rules?
- About privacy, yes the user asking for search results is locked when querying the database (see server.js `socket.on("search-query")`)

for @YaManicKill & @xPaw about indexing:
- Yes it seems a better idea, so I'll do a one-time indexing process (based on logs) and index new messages on the fly.

I'll come back soon and update that PR.

xPaw:
@florentsolt Version of `sqlite` you are using does not work on node 4/5, and as we still support these version it has to be changed. That's why CI builds are failing too.

florentsolt:
@xPaw this is annoying 👎 I'll have to use sqlite3 directly and forget this package 😢

xPaw:
@florentsolt You should be able to use `sqlite/legacy` from the same package.

florentsolt:
@xPaw Yep I've see that, I'll git it a try. Thanks

florentsolt:
@xPaw & @YaManicKill I think I've managed all your feedbacks. I didn't yet squashed my commits for you to see them all (tell me if you prefer all squashed). In the end I prefer to go for sqlite3 alone, I prefer reduce the number of modules.

Anything more ?

client/index.html
```diff
@@ -198,6 +199,49 @@ <h1 class="title">
...
+							<div class="col-sm-4">
+								<label for="date-from" class="sr-only">From</label>
+								<input type="text" id="date-from" name="from" class="input" placeholder="From: 7 days">
+							</div>
+							<div class="col-sm-4">
+								<label for="date-to" class="sr-only">To</label>
+								<input type="text" id="date-to" name="to" class="input" placeholder="To: now">
+							</div>
+							<div class="col-sm-4">
+									<button type="submit" class="btn pull-right">Search</button>
```
xPaw: extra tab

client/views/msg.tpl
```diff
@@ -6,6 +6,9 @@
 		{{#if from}}
 			{{> user_name nick=from}}
 		{{/if}}
+		{{#if nick}}
+			{{> user_name nick=nick}}
```
xPaw: Did you add this because your search results send `nick` instead of `from`? If so, can you modify it so it sends `from`?

YaManicKill:
> I didn't yet squashed my commits for you to see them all (tell me if you prefer all squashed). I

We will want them squashed, but yeah, no need to do it until we are ready to merge.

I think there's a few things that I'll point out but it's mainly just code style stuff really. Pulling out the sql commands into constants would be good. But generally it looks pretty good, thanks for this. I'll hopefully have some time to do a more in-depth review of this later tonight, or this weekend sometime.

Just to give you a heads up, we don't plan to get this merged for 2.5.0, just because that is nearly finished and this is quite a big PR so late in the 2.5.0 cycle, so we'll probably have to wait until 2.5.0 is finished before we merge.

defaults/config.js
```diff
@@ -506,4 +506,32 @@ module.exports = {
...
+		// @default  false
+		//
+		indexLobby: false,
+
+		// Enables indexing only for messages (skip everything else) from channels.
+		//
+		// @type     boolean
+		// @default  false
+		//
+		indexOnlyChanMessage: true,
```
xPaw: I feel like neither of these `index` options are necessary. Just store everything, we would also eventually want to reload messages from sqlite when restarting the server.

defaults/config.js
```diff
@@ -506,4 +506,32 @@ module.exports = {
...
+
+		// How many messages to display around matches.
+		//
+		// @type     int
+		// @default  4
+		//
+		proximity: 4
```
xPaw: This probably could be changed in the UI when searching?

package.json
```diff
@@ -41,6 +41,7 @@
   },
   "dependencies": {
     "bcryptjs": "2.4.3",
+    "bluebird": "3.5.0",
```
xPaw: Any particular reason you need this?

src/search.js
```diff
@@ -0,0 +1,231 @@
...
+}
+
+var format = Helper.config.logs.format || "YYYY-MM-DD HH:mm:ss";
+
+var schema = [
+	"CREATE TABLE IF NOT EXISTS logs (user TEXT, network TEXT, chan TEXT, time INTEGER, nick TEXT, text TEXT)",
+	"CREATE VIRTUAL TABLE IF NOT EXISTS texts USING fts4(content=\"logs\", nick, text)",
+	"CREATE TRIGGER IF NOT EXISTS logs_bu BEFORE UPDATE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_bu BEFORE UPDATE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_bd BEFORE DELETE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
```
xPaw: Aren't these 3 triggers doing exact same thing?

src/search.js
```diff
@@ -0,0 +1,231 @@
...
+} catch (e) {
+	log.error("Unable to create log directory", e);
+	return;
+}
+
+var format = Helper.config.logs.format || "YYYY-MM-DD HH:mm:ss";
+
+var schema = [
+	"CREATE TABLE IF NOT EXISTS logs (user TEXT, network TEXT, chan TEXT, time INTEGER, nick TEXT, text TEXT)",
+	"CREATE VIRTUAL TABLE IF NOT EXISTS texts USING fts4(content=\"logs\", nick, text)",
```
xPaw: Any reason you need a virtual table? I think just `logs` should be fine.

src/search.js
```diff
@@ -0,0 +1,231 @@
...
+	"CREATE TRIGGER IF NOT EXISTS logs_bu BEFORE UPDATE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_bu BEFORE UPDATE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_bd BEFORE DELETE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_ai AFTER INSERT ON logs BEGIN INSERT INTO texts(docid, nick, text) VALUES(new.rowid, new.nick, new.text); END"
+];
+
+var db = new Sqlite.Database(path.join(Helper.getLogsPath(), "search.db"));
+db.serialize(() => schema.forEach((line) => db.run(line)));
+
+var indexAll = () => {
```
xPaw: Since these index functions are only used from CLI, could you move them there?

src/search.js
```diff
@@ -0,0 +1,231 @@
...
+				}
+			}
+			cb(null, line);
+		}))
+		.on("error", (err) => reject(err))
+		.on("end", () => resolve());
+});
+
+var index = (user, network, chan, time, nick, text) => {
+	console.log(user, network, chan, time, nick, text);
```
xPaw: remove

src/search.js
```diff
@@ -0,0 +1,231 @@
...
+	return Promise.resolve(db)
+		.then(() => Sqlite.run("INSERT INTO logs(user, network, chan, time, nick, text) VALUES(?, ?, ?, ?, ?, ?)",
+			user, network, chan, time.getTime() / 1000, nick, text));
+};
+
+// TODO: this may require indexes with a big database
+var getNetworksChannels = (user) => new Promise((resolve, reject) => {
```
xPaw: Use `function` for function definitions please.

src/search.js
```diff
@@ -0,0 +1,231 @@
+"use strict";
+
+var path = require("path");
```
xPaw: Use `const` instead of `var`.

client/js/socket-events/search.js
```diff
@@ -0,0 +1,42 @@
+"use strict";
+
+const $ = require("jquery");
+const socket = require("../socket");
+const render = require("../render");
+const messages = $("#search-query .results .messages");
+const form = $("#search-query form");
+const select = $("#search-query form select");
+
+socket.on("search-networks-channels", function(data) {
```
xPaw: I can see how this could be useful to search logs for channels that you are no longer a part of, but I also think such a niche feature warrants code this convoluted (which also runs a DISTINCT on the database).

This could be simplified a lot by simply taking networks/channels from the sidebar UI.

src/search.js
```diff
@@ -0,0 +1,231 @@
...
+			cb(null, line);
+		}))
+		.on("error", (err) => reject(err))
+		.on("end", () => resolve());
+});
+
+var index = (user, network, chan, time, nick, text) => {
+	console.log(user, network, chan, time, nick, text);
+	return Promise.resolve(db)
+		.then(() => Sqlite.run("INSERT INTO logs(user, network, chan, time, nick, text) VALUES(?, ?, ?, ?, ?, ?)",
```
xPaw: Is the code correct here, should it actually be `Sqlite.run`?

src/search.js
```diff
@@ -0,0 +1,231 @@
...
+var schema = [
+	"CREATE TABLE IF NOT EXISTS logs (user TEXT, network TEXT, chan TEXT, time INTEGER, nick TEXT, text TEXT)",
+	"CREATE VIRTUAL TABLE IF NOT EXISTS texts USING fts4(content=\"logs\", nick, text)",
+	"CREATE TRIGGER IF NOT EXISTS logs_bu BEFORE UPDATE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_bu BEFORE UPDATE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_bd BEFORE DELETE ON logs BEGIN DELETE FROM texts WHERE docid=old.rowid; END",
+	"CREATE TRIGGER IF NOT EXISTS logs_ai AFTER INSERT ON logs BEGIN INSERT INTO texts(docid, nick, text) VALUES(new.rowid, new.nick, new.text); END"
+];
+
+var db = new Sqlite.Database(path.join(Helper.getLogsPath(), "search.db"));
```
xPaw: Rename to `database`, and use `const`.

xPaw:
I like this quite a bit, we will also eventually be able to also reload logs from sqlite when restarting the server.

src/search.js
florentsolt: This is not a "normal" virtual table, but more the way to use full text search capabilities for ex. you can use "*" in the text and nick input fields.

See https://www.sqlite.org/fts3.html for more infos.

src/search.js
florentsolt: A commit is missing, sorry.

client/views/msg.tpl
florentsolt: Yep, "from" as a field in sql database is not super handy :)
Also, I'm using "from" as time conditions from/to.
I'll see what I can do.

src/search.js
florentsolt: I rather prefer keep code that have access to the "db" object (and do queries) in the same file.

client/js/socket-events/search.js
florentsolt: So you would be able to filter only on the still opened networks, channels, queries... and If you search for something that is not opened anymore (for ex. a past chat you had in a query window) you will have to select the network entry or the "all network / all channels" entry? 

I'm also not a huge fan of the DISTINCT query (and this event).

Everybody is fine with that?

package.json
florentsolt: I'm using Promise.each & Promise.map which are quite handy.
Do you have alternatives?

xPaw:
@florentsolt These index functions are only needed for backwards compatibility, right? If so, I would prefer to keep the main search/database code free from that.

florentsolt:
For @xPaw:

1. index() function will be used every time a new msg is issued.
1. indexAll() (which call indexFile()) is used to import all past log files. IMHO it's quite nice to be able to import past conversations. Also, if anything happen to the sqlite database, it's possible to regenerate it.

But now I get your point. I read too fast one of your past comment. You plan to move from flat files to sqlite. Wow! :) In that case, I will for ex. do a cll "import" and move all the code in here. 

Are we in sync?

xPaw:
>  You plan to move from flat files to sqlite. Wow! :) In that case, I will for ex. do a cll "import" and move all the code in here.

We don't plan it, but my idea is that having sqlite support means messages are automatically logged into it, and the only time we would need to read text logs would to import old logs into the database. Which is why I think the import related code would be in command file.

jakehm:
I'd rather logs be binaries or text files.

There is going to be a huge performance hit when the program is making all these insert statements.

xPaw:
@jakehm The insert is not that much worse comparing to just writing a text file.

jakehm:
Plus log maintenance will be harder.  Some of us have this running on rpi.

xPaw:
@jakehm What are you arguing for? Text logs are not going anywhere, not in this PR anyway.

jakehm:
OK so several inserts a second into a sqlite db is not going to blow up my raspberry pi?

YaManicKill:
@jakehm The point is that sqlite will never be required to run lounge. It's an option that you can enable.

xPaw:
@florentsolt Are you able to bring this up to date? Do keep in mind that we've removed `event-stream` dependency.

florentsolt:
No I don't mind, we just a need another efficient way to read all the log files, but I'm sure there is other ways.

client/js/lounge.js
```diff
@@ -25,6 +25,7 @@ require("./clipboard");
 $(function() {
 	var sidebar = $("#sidebar, #footer");
 	var chat = $("#chat");
+	var searchQuery = $("#search-query");
```
MaxLeiter: change this and the above two lines to `const`/`let`?

client/js/lounge.js
```diff
@@ -485,7 +490,14 @@ $(function() {
 		container.html(templates.user_filtered({matches: result})).show();
 	});
 
-	var forms = $("#sign-in, #connect, #change-password");
+	searchQuery.on("click", ".show-more-button", function() {
+		$(this).text("Loading older messages…").prop("disabled", true);
+		var query = searchQuery.find("form").data("lastQuery");
```
MaxLeiter: let

client/js/lounge.js
```diff
@@ -485,7 +490,14 @@ $(function() {
...
+		query.offset = searchQuery.find(".highlight").length;
+		socket.emit("search-query", query);
+	});
+
+	var forms = $("#sign-in, #connect, #change-password, #search-query");
```
MaxLeiter: let

client/js/socket-events/search.js
```diff
@@ -0,0 +1,42 @@
...
+	var groups = {};
```
MaxLeiter: let

client/js/socket-events/search.js
```diff
@@ -0,0 +1,42 @@
...
+	data.forEach(function(entry) {
+		if (!groups[entry.network]) {
+			groups[entry.network] = [];
+		}
+		if (entry.network !== entry.chan) {
+			groups[entry.network].push(entry.chan);
+		}
+	});
+	Object.keys(groups).forEach(function(network) {
+		var group = $("<optgroup></optgroup>").attr("label", network);
```
MaxLeiter: let

src/command-line/search.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+
+var program = require("commander");
```
MaxLeiter: `const`

MaxLeiter: same for rest of file

src/search.js
```diff
@@ -0,0 +1,244 @@
...
+};
+
+try {
+	fsextra.ensureDirSync(Helper.getLogsPath());
+} catch (e) {
+	log.error("Unable to create log directory", e);
+	return;
+}
+
+var format = Helper.config.logs.format || "YYYY-MM-DD HH:mm:ss";
```
MaxLeiter: `let`/`const` for this and rest of file

src/server.js
```diff
@@ -267,6 +268,31 @@ function initializeClient(socket, client, token, lastMessage) {
...
+			Search.getNetworksChannels(client.name)
+				.then((data) => socket.emit("search-networks-channels", data));
+		}
+	);
+
+	socket.on(
+		"search-query",
+		function(query) {
+			if (!query.network || !query.chan) {
+				var split = query["network-channel"].split("/");
```
MaxLeiter: `let`/`const`

client/js/lounge.js
florentsolt: Sorry but there is also others variables referring to elements below. I'm do not know all their use cases nor scopes. So I did like the others. See with the team what to do globally. My PR is maybe not the right place to do it.

xPaw: In new code I generally use let/const unless `var` is required.

client/js/lounge.js
florentsolt: Eslint advise to use const in that case, what's your opinion?

client/js/lounge.js
florentsolt: This was here before my PR, please reach the initial author.

client/js/socket-events/search.js
florentsolt: Eslint advise to use const in that case, what's your opinion?

client/js/lounge.js
xPaw: We have a lot of old code that still has `var` unfortunately.

client/js/socket-events/search.js
florentsolt: Eslint advise to use const in that case, what's your opinion?

src/command-line/search.js
florentsolt: Done.

src/search.js
florentsolt: Done. (no let, as advised by Eslint)

src/server.js
florentsolt: Done.

client/js/socket-events/search.js
astorije: You should go with `const` then :)

astorije:
Salut @florentsolt!
Sorry we fell behind on your PR... Do you think you could rebase on `master` and fix conflicts? 🙏

florentsolt:
@astorije sorry I'm lacking of free time these days. I've waited for quite a while, and now I'm super busy. Plus, the UI has changed a lot recently, it's quite difficult to know where to merge my changes.

xPaw:
@florentsolt I started rewriting sqlite logging here: https://github.com/thelounge/lounge/compare/sqlite

I want to just get the message logging and reloading part right. As you might see I generate random uuid per user network to avoid collisions and easier querying over all (doesn't matter if hostname changes, or there are multiple networks connecting to same hostname).

florentsolt:
@xPaw it looks quite nice!

McInkay:
@florentsolt Would you have time to rebase this on top of @xPaw's PR? It'll hopefully be merged in soon and then we can get this one in. If you won't have time to do that, feel free to close this PR and we'll stop bugging you about it.

florentsolt:
Hi, @McInkay no sorry, I'm really to far behind and have no free time. I saw you are going forward with a sqlite backend, I'm glad I did inspire this move. Looking forward to the next release :)

McInkay:
Well thanks for the contribution anyway, sorry we couldn't get your code in. If you ever feel like you want to contribute again you are welcome to come back. No hard feelings, we all run out of free time sometimes.

PR #1535, opened 2017-09-17
Fix references to undefined `this`

astorije:
https://github.com/thelounge/lounge/commit/79eb83d82ff48a58aac7bc1bd6a44ece5ef85dcc introduces a couple uses of `this` in a function that doesn't know about the channel.

I preferred to pass the channel instead of making this part of the prototype (and therefore exporting it) but I can switch to that otherwise.

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -59,7 +59,7 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 	this.messages.push(msg);
 
 	if (client.config.log === true) {
-		writeUserLog(client, msg);
+		writeUserLog.call(this, client, msg);
 	}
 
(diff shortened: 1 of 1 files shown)


xPaw:
@astorije I think prototype is better suited here (think private function in a class). Also, oops.

astorije:
> think private function in a class

What do you mean, I don't think it's possible to have a private member, is it?

xPaw:
It is possible in an actual class, I just suggested that prototype is better here than passing references in the same file.

astorije:
Actually, nothing `.call()` can't fix 😁
That way, still private, can still use `this`. What do you think?

It looks like private methods are an ES7 thing, and just not very pretty syntactic sugar over `.call` (same code, just `this::writeUserLog(client, msg)`).

PR #1536, opened 2017-09-17
Implement session list and allow signing out other clients

xPaw:


MaxLeiter:
screen shot 2017-09-18 at 1 12 22 pm

`::1` is a side-effect of localhost being used, maybe we should clean that up client-side? Links to: https://ipinfo.io/::1

xPaw:
@MaxLeiter That's not a side effect, that's the actual IP. And the ipinfo page says it's reserved for localhost.

astorije:
@xPaw, I'm not a fan of the disabled <kbd>Current</kbd> button. Could you do like I did in #986 (totally stolen from Telegram lol)?

screen shot 2017-04-30 at 03 23 18

I know it's not very helpful, but just so you don't have to go dig in that PR, markup was:

```html
<h2>Active sessions</h2>

<h3>Current session</h3>
<p id="current_connection_info"></p>

<h3>Other sessions</h3>
<div id="connection_list"></div>
```

xPaw:
@astorije Ok, will do.

client/js/lounge.js
```diff
@@ -404,6 +405,11 @@ $(function() {
 			socket.emit("names", {target: self.data("id")});
 		}
 
+		if (type === "settings") {
+			$("#session-list").text("Loading…");
+			socket.emit("sessions:get");
```
astorije: Are you planning for live updates? A bit gross but at the very least we could setInterval when settings are open? Meh

xPaw: If you want live updating, then you would emit the event from the server whenever anything related to sessions changes. But I feel like would complicate it more than necessary (for now at least).

astorije: Yeah we'd need `socket.emit("sessions:live");` and `socket.emit("sessions:stop");` when leaving the window or something like that. I get that it's more complicated, but it would also be much more valuable. #986 supported that (though not optimal there, I know) :/
I understand if you don't want to bother with this for this PR.

client/views/session.tpl
```diff
@@ -0,0 +1,15 @@
+<p class="clearfix">
+{{#if current}}
+	<strong>{{agent}}</strong>
+	<a href="https://ipinfo.io/{{ip}}" target="_blank" rel="noopener">{{ip}}</a>
```
astorije: Nitpick but any specific reason why you used ipinfo.io over geoiptool.com I used in #986? I liked that it shows a more precise location on the map + zoomable. ipinfo.io isn't doing too well (though I know this is not a perfect metric, might as well display it in a useful way).

client/views/session.tpl
```diff
@@ -0,0 +1,15 @@
...
+	<br>
+	{{localetime lastUse}}
+{{else}}
+	<button class="btn pull-right remove-session" data-token="{{token}}">Disconnect</button>
+
+	<strong>{{agent}}</strong>
+	<a href="https://ipinfo.io/{{ip}}" target="_blank" rel="noopener">{{ip}}</a>
+	<br>
+	{{localetime lastUse}} {{#if active}}<i>(currently active)</i>{{/if}}
```
astorije: What do you think of prefixing with `Last seen on ...`  or something like that?

client/views/session.tpl
```diff
@@ -0,0 +1,15 @@
+<p class="clearfix">
+{{#if current}}
+	<strong>{{agent}}</strong>
+	<a href="https://ipinfo.io/{{ip}}" target="_blank" rel="noopener">{{ip}}</a>
+	<br>
+	{{localetime lastUse}}
```
astorije: You can remove this one, it will just always give the current time or something like that lol, not super useful I think.

client/views/session.tpl
xPaw: I have used ipinfo.io before, I think it's a nicer service overall.

geoiptool.com reports US for my IPv6 address, so it's that good.

client/js/lounge.js
xPaw: I don't see how it's absolutely necessary to have live refreshing here, this can be a future enhancement.

astorije: It's not, can totally be a future enhanceement.

client/views/session.tpl
astorije: Fair enough :+1:

client/css/style.css
```diff
@@ -246,6 +246,15 @@ kbd {
 	color: #7f8c8d;
 }
 
+.session-list strong {
+	display: block;
+}
+
+.session-list p + p {
+	padding-top: 10px;
+	border-top: 1px solid #cdd3da;
```
astorije: We don't really use lines as separators anywhere else in the app (except under second-level titles) so I think we could do without. But I'm not strongly opinionated either, it's fine with it as well.

client/index.html
```diff
@@ -198,16 +198,15 @@ <h1 class="title">
...
 					</div>
 					<div class="container">
+						<h1 class="title">Settings</h1>
+
 						<div class="row">
 							<div class="col-sm-12">
-								<h1 class="title">Settings</h1>
-							</div>
-							<div class="col-sm-12">
 								<h2>Messages</h2>
```
astorije: Do you want to take this title out of the `row`+`col-md-12` as well?

client/js/socket-events/sessions_list.js
```diff
@@ -0,0 +1,27 @@
...
+	data.sort((a, b) => b.lastUse - a.lastUse);
+
+	let html = "";
+	data.forEach((connection) => {
+		if (connection.current) {
+			$("#session-current").html(templates.session(connection));
+			return;
+		}
+
+		html += templates.session(connection);
```
astorije: I find this way harder to read than just `if`+`else` but nbd.

client/views/session.tpl
```diff
@@ -0,0 +1,17 @@
...
+	<strong>{{agent}}</strong>
+	<a href="https://ipinfo.io/{{ip}}" target="_blank" rel="noopener">{{ip}}</a>
+{{else}}
+	<button class="btn pull-right remove-session" data-token="{{token}}">Disconnect</button>
+
+	<strong>{{agent}}</strong>
+	<a href="https://ipinfo.io/{{ip}}" target="_blank" rel="noopener">{{ip}}</a>
+	<br>
+	{{#if active}}
+		<b>Currently active</b>
```
astorije: I would use `em` instead of `b` to contrast it with the user agent line but again nbd if you don't want to.

client/index.html
```diff
@@ -380,6 +379,18 @@ <h1 class="title">Settings</h1>
 								<textarea class="input" name="userStyles" id="user-specified-css-input" placeholder="You can override any style with CSS here"></textarea>
 							</div>
 						</div>
+
+						{{#unless public}}
+							<div class="session-list">
+								<h2>Sessions</h2>
+								
+								<h3>Current session</h3>
```
astorije: I think the `session-list` id is not set where you want to set it, because we get:

screen shot 2017-09-23 at 20 36 58

Probably want to hide both "Current session" and "Other sessions" when loading, then display them and hide the loading statement when it's available, no?

astorije: Or I'm fine with no loading at all if you don't want to bother, it's pretty fast anyway. Just that at the moment it feels odd, I'm wondering if it's not leftovers from your previous version where you didn't have separate sections.

client/index.html
```diff
@@ -380,6 +379,18 @@ <h1 class="title">Settings</h1>
...
+								<div id="session-current"></div>
+								
+								<h3>Other sessions</h3>
+								<div id="session-list"></div>
```
astorije: At the moment, this section is empty when there is no other known session, yet the title is still present. What about the following?

screen shot 2017-09-23 at 20 44 23

Achieved by:

```diff
─────────────────────────
modified: client/index.html
─────────────────────────
@ index.html:386 @
                        {{#unless public}}
                            <div class="session-list">
                                <h2>Sessions</h2>
-
+
                                <h3>Current session</h3>
                                <div id="session-current"></div>
-
+
                                <h3>Other sessions</h3>
-                               <div id="session-list"></div>
+                               <div id="session-list">
+                                   <p>
+                                       <em>You are not currently logged in to any other device.</em>
+                                   </p>
+                               </div>
                            </div>
                        {{/unless}}
                    </div>
─────────────────────────
modified: client/js/lounge.js
─────────────────────────
@ lounge.js:409 @ $(function() {
        }

        if (type === "settings") {
-           $("#session-list").text("Loading…");
            socket.emit("sessions:get");
        }

─────────────────────────
modified: client/js/socket-events/sessions_list.js
─────────────────────────
@ sessions_list.js:20 @ socket.on("sessions:list", function(data) {
        html += templates.session(connection);
    });

-   $("#session-list").html(html);
+   if (html.length > 0) {
+       $("#session-list").html(html);
+   }
```

(I'm not pushing for removal of `Loading…`, it was just done quickly to not have it delete the added placeholder)

src/client.js
```diff
@@ -304,11 +304,19 @@ Client.prototype.updateSession = function(token, ip, request) {
 		friendlyAgent += ` on ${agent.os.name} ${agent.os.version}`;
 	}
 
-	client.config.sessions[token] = _.assign({
+	client.config.sessions[token] = _.assign(client.config.sessions[token], {
 		lastUse: Date.now(),
 		ip: ip,
 		agent: friendlyAgent,
-	}, client.config.sessions[token]);
+	});
```
astorije: Do you remember why it was in the other order previously? I get the change, but not the previous state.

src/client.js
```diff
@@ -304,11 +304,19 @@ Client.prototype.updateSession = function(token, ip, request) {
...
+
+	client.manager.updateUser(client.name, {
+		sessions: client.config.sessions
+	}, (err) => {
+		if (err) {
+			log.error("Failed to update sessions for", client.name, err);
```
astorije: Isn't that going to log multiple lines like we've seen in other places? If so, you might have to use something like this:

```js
log.error(`Failed to update sessions for ${color.green(client.name)} (${err})`);
```

src/server.js
```diff
@@ -378,8 +377,32 @@ function initializeClient(socket, client, token, lastMessage) {
 		client.unregisterPushSubscription(token);
 	});
 
-	socket.on("sign-out", () => {
-		delete client.config.sessions[token];
+	const sendSessionList = () => {
+		const sessions = _.map(client.config.sessions, (session, sessionToken) => ({
+			current: sessionToken === token,
+			active: typeof _.find(client.attachedClients, (u) => u.token === sessionToken) !== "undefined",
```
astorije: I know the whole reasoning behind that `typeof ... !== "undefined"` but it is frankly overrated, `... !== undefined` is perfectly fine (I read an article not long ago that I can't find, which was dismantling this pretty well; in particular the author was saying than no one ever applies this amount of strictness outside of this case).

In fact, we use `===/!== undefined` in many places already :sweat_smile:

src/server.js
```diff
@@ -389,12 +412,25 @@ function initializeClient(socket, client, token, lastMessage) {
...
+				return;
+			}
+
+			const socketToRemove = manager.sockets.of("/").connected[socketId];
+
+			socketToRemove.emit("sign-out");
+			socketToRemove.disconnect();
+		});
+
+		sendSessionList(client);
```
astorije: Why do you need to send the list when logging out?

src/client.js
xPaw: It was just incorrect before.

src/server.js
xPaw: So that client sees the updated session list.

client/index.html
xPaw: All the headings are in inside. This isn't related to the PR either way.

xPaw:
@astorije Should be done.

client/views/session.tpl
```diff
@@ -0,0 +1,17 @@
+<p class="clearfix">
```
astorije: I played a bit with the markup and it seems to me like you don't need `clearfix` here. At least I couldn't see any difference with / without at different resolutions.

client/js/socket-events/sessions_list.js
```diff
@@ -0,0 +1,31 @@
...
+	});
+
+	if (html.length === 0) {
+		html = "<em>You are not currently logged in to any other device.</em>";
```
astorije: This needs to be in a `<p></p>` to respect font size (and also a good idea to match session blocks in case margin or else gets changed).

btw much better idea than what I posted yesterday lol #shame

src/server.js
```diff
@@ -378,23 +377,56 @@ function initializeClient(socket, client, token, lastMessage) {
...
 			}
+
+			const socketToRemove = manager.sockets.of("/").connected[socketId];
+
+			socketToRemove.emit("sign-out");
+			socketToRemove.disconnect();
 		});
 
-		socket.emit("sign-out");
+		sendSessionList(client);
```
astorije: Actually, I'm still confused, sorry. `sendSessionList` doesn't take an argument. But if I read the code correctly, all connected clients should receive the updated list of sessions, no? In practice, only the one where click on "Disconnect" happened.

xPaw: Only the client that clicked disconnect will receive the updated list, that is correct.

astorije: I see. But the argument is still superfluous here, I believe.

client/js/lounge.js
```diff
@@ -404,6 +405,11 @@ $(function() {
 			socket.emit("names", {target: self.data("id")});
 		}
 
+		if (type === "settings") {
+			$("#session-list").text("Loading…");
```
astorije: Uuuh I feel bad that I missed that but this should be `$("#session-list").html("<p>Loading…</p>");` (or `<p><em>Loading…</em></p>`). I can make the change myself if you give me the go, sorry!

src/server.js
```diff
@@ -378,23 +377,56 @@ function initializeClient(socket, client, token, lastMessage) {
...
+		}));
+
+		socket.emit("sessions:list", sessions);
+	};
+
+	socket.on("sessions:get", sendSessionList);
+
+	socket.on("sign-out", (tokenToSignOut) => {
+		// If no token provided, sign same client out
+		if (!tokenToSignOut) {
```
YaManicKill: Could do `tokenToSignOut = tokenToSignOut || token` but not a massive thing. Just a little neater.

src/server.js
```diff
@@ -378,23 +377,56 @@ function initializeClient(socket, client, token, lastMessage) {
...
 			}
+
+			const socketToRemove = manager.sockets.of("/").connected[socketId];
+
+			socketToRemove.emit("sign-out");
+			socketToRemove.disconnect();
 		});
 
-		socket.emit("sign-out");
+		sendSessionList();
```
YaManicKill: We'll want to wrap this in a check for logging out the current session, otherwise we'll try to send the session list to the session we just disconnected.

xPaw: Fixed.

PR #1538, opened 2017-09-18
Enable `no-console` and `no-alert` ESLint rules 

astorije:
We almost slipped in a few `console.log` statements so this should make it safer. Outside of the current occurrences we have, we shouldn't need any more in the code, or at least not without explicitly acknowledging them.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -9,54 +9,54 @@ env:
   node: true
 
 rules:
-  arrow-body-style: 2
-  arrow-parens: [2, always]
-  arrow-spacing: 2
-  block-scoped-var: 2
-  block-spacing: [2, always]
-  brace-style: [2, 1tbs]
-  comma-dangle: 0
-  curly: [2, all]
-  dot-location: [2, property]
-  dot-notation: 2
-  eol-last: 2
-  eqeqeq: 2
-  handle-callback-err: 2
-  indent: [2, tab]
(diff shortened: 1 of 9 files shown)


client/js/socket.js
```diff
@@ -35,8 +35,6 @@ const socket = io({
 		});
 		// Hides the "Send Message" button
 		$("#submit").remove();
-
-		console.error(data);
```
astorije: We really shouldn't be using `console.xxx` in the browser...

webpack.config.js
```diff
@@ -79,8 +79,6 @@ if (process.env.NODE_ENV === "production") {
 		sourceMap: true,
 		comments: false
 	}));
-} else {
-	console.log("Building in development mode, bundles will not be minified.");
```
astorije: I don't believe this is really helpful at this point, might as well not have it.

.eslintrc.yml
```diff
@@ -9,54 +9,54 @@ env:
...
+  semi-style: [error, last]
+  semi: [error, always]
+  space-before-blocks: error
+  space-before-function-paren: [error, never]
+  space-in-parens: [error, never]
+  space-infix-ops: error
+  spaced-comment: [error, always]
+  strict: error
+  template-curly-spacing: error
+  yoda: error
```
YaManicKill: Is "2" equal to "error"?

xPaw: Yes.

PR #1539, opened 2017-09-18
Create lounge install command

### LINKED ISSUE #1530: Create `lounge install` command for packages
Example:

`lounge install lounge-theme-solarized`

This should install lounge-theme-solarized into `$LOUNGE_HOME/packages`, check that it is, infact, a lounge package, and then ideally tell lounge that a theme has been added (maybe outside of the scope of the first PR).

Example 2:

`lounge install lounge-package-name`

The hsould install lounge-package-name `$LOUNGE_HOME/packages`, check that it is, infact, a lounge package. Same with regards to telling lounge about the new package.

xPaw:
Fixes #1530

### DIFF
--- a/src/command-line/install.js
+++ b/src/command-line/install.js
@@ -1,0 +1,82 @@
+"use strict";
+
+const colors = require("colors/safe");
+const program = require("commander");
+const Helper = require("../helper");
+const Utils = require("./utils");
+
+program
+	.command("install <package>")
+	.description("Install a theme or a package")
+	.on("--help", Utils.extraHelp)
+	.action(function(packageName) {
+		const fs = require("fs");
+		const fsextra = require("fs-extra");
+		const path = require("path");
+		const child = require("child_process");
+		const packageJson = require("package-json");
+
+		if (!fs.existsSync(Helper.CONFIG_PATH)) {
+			log.error(`${Helper.CONFIG_PATH} does not exist.`);
+			return;
+		}
+
+		packageJson(packageName, {
+			fullMetadata: true
+		}).then((json) => {
+			if (!("lounge" in json)) {
+				log.error(`${colors.red(packageName)} does not have The Lounge metadata.`);
+
+				process.exit(1);
+			}
+
+			log.info(`Installing ${colors.green(packageName)}...`);
+
+			const packagesPath = Helper.getPackagesPath();
+			const packagesParent = path.dirname(packagesPath);
+			const packagesConfig = path.join(packagesParent, "package.json");
+
+			// Create node_modules folder, otherwise npm will start walking upwards to find one
+			fsextra.ensureDirSync(packagesPath);
+
+			// Create package.json with private set to true to avoid npm warnings
+			fs.writeFileSync(packagesConfig, JSON.stringify({
+				private: true,
+				description: "Packages for The Lounge. All packages in node_modules directory will be automatically loaded.",
+			}, null, "\t"));
+
+			const npm = child.spawn(
+				process.platform === "win32" ? "npm.cmd" : "npm",
+				[
+					"install",
+					"--production",
+					"--no-save",
+					"--no-bin-links",
+					"--no-package-lock",
+					"--prefix",
+					packagesParent,
+					packageName
+				],
+				{
+					stdio: "inherit"
+				}
+			);
+
+			npm.on("error", (e) => {
+				log.error(`${e}`);
+				process.exit(1);
+			});
+
+			npm.on("close", (code) => {
+				if (code !== 0) {
+					log.error(`Failed to install ${colors.green(packageName)}. Exit code: ${code}`);
+					return;
+				}
+
+				log.info(`${colors.green(packageName)} has been successfully installed.`);
+			});
+		}).catch((e) => {
+			log.error(`${e}`);
+			process.exit(1);
+		});
+	});
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "ldapjs": "1.0.1",
     "lodash": "4.17.4",
     "moment": "2.18.1",
+    "package-json": "4.0.1",
     "read": "1.0.7",
     "request": "2.82.0",
     "semver": "5.4.1",
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -32,6 +32,7 @@ require("./add");
 require("./remove");
 require("./reset");
 require("./edit");
+require("./install");
 
 program.parse(process.argv);
 


src/command-line/install.js
```diff
@@ -0,0 +1,63 @@
...
+			fullMetadata: true
+		}).then((json) => {
+			if (!("lounge" in json)) {
+				log.error(`${colors.red(packageName)} does not have The Lounge metadata.`);
+
+				process.exit(1);
+			}
+
+			const npm = child.spawn(
+				path.sep === "\\" ? "npm.cmd" : "npm",
```
YaManicKill: Is this the best way to know if it is windows? Would it be better to use https://stackoverflow.com/questions/8683895/how-do-i-determine-the-current-operating-system-with-node-js

src/command-line/install.js
```diff
@@ -0,0 +1,63 @@
...
+				[
+					"install",
+					"--global",
```
YaManicKill: Hmm, doesn't `--global` break this, as discussed in #1266 ?

xPaw: It worked in my testing.

src/command-line/install.js
xPaw: Can use `process.platform`

xPaw:
On linux I am getting this:
```
$ node index.js install lounge-theme-solarized
npm WARN enoent ENOENT: no such file or directory, open '/home/xpaw/.lounge/packages/package.json'
npm WARN packages No description
npm WARN packages No repository field.
npm WARN packages No README data
npm WARN packages No license field.
```

It installs anyway, but we should prevent the warnings.

YaManicKill:
> EDIT: Apparently these are not shown with --global, but @astorije had issues with that? Needs more testing.

Yeah, when you install with global, it puts it in `packages/lib/node_modules` rather than `packages/node_modules`. Which might be worth it to remove the errors?

xPaw:
@YaManicKill This might be OS dependant behaviour, because it didn't create a lib/ folder on Windows for me.

YaManicKill:
Of course, the other option is to just hide the stdout in the child exec then.

src/command-line/install.js
```diff
@@ -0,0 +1,70 @@
...
+					"--prefix",
+					packagesPath.replace(/node_modules$/, ""),
+					packageName
+				]
+			);
+
+			npm.on("error", (e) => {
+				log.error(`${e}`);
+				process.exit(1);
+			});
```
astorije: On the channel earlier, when you said "I hid npm stdout", I didn't realize it meant that you simply don't redirect it (actually, for the warnings we could see, it's from `stderr`). I thought you were actively silencing the output or something, hence my suggestion.

If we ever need a clean `stderr` (for example if this block doesn't report everything it should), I checked and having a `package.json` with only `{ "private": true }` in it silences all the warnings :man_shrugging: 

Do you know what's the difference between `npm.on("error", ...)`  and `npm.stderr.on("data", ...)`? Warnings are sent to the latter, and installing a package that doesn't exist goes through your `catch`, but I'm not sure what this one does.

src/command-line/install.js
```diff
@@ -0,0 +1,70 @@
+"use strict";
+
+const colors = require("colors/safe");
+const program = require("commander");
+const Helper = require("../helper");
+const Utils = require("./utils");
+
+program
+	.command("install <package>")
+	.description("Install a package")
```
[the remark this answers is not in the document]
YaManicKill: I thought the point of "package" was that it includes themes as well as all other types of packages. Wasn't that your suggestion?

src/command-line/install.js
xPaw: I reverted hiding of stdout, and create package.json now.

src/command-line/install.js
```diff
@@ -0,0 +1,70 @@
...
+		const packageJson = require("package-json");
+
+		if (!fs.existsSync(Helper.CONFIG_PATH)) {
+			log.error(`${Helper.CONFIG_PATH} does not exist.`);
+			return;
+		}
+
+		packageJson(packageName, {
+			fullMetadata: true
+		}).then((json) => {
```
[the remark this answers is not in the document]
xPaw: I really don't see the need to write it manually, at least right now.

If you're going to hardcode the registry (which I don't think is a good idea), then you will also need to pass it into `npm install` command.

src/command-line/install.js
YaManicKill: Oh no, can we please not do this. We can check whether the error code is 0 and if not we can print out the stdout, but creating a package.json when there isn't a package feels like a horrible solution.

xPaw: `package.json` is in `.lounge/packages/` folder, so there's just one for all the packages (not for each).

YaManicKill: I get that, but it's still misleading because packages/ isn't a node package in itself, so shouldn't have a package.json.

xPaw: Why does it matter?

YaManicKill: Why is it better than muting stderr? I'm not keen on having a file hanging around that is unnecessary (and could be misleading). Especailly one that says "private:true" which makes no sense for the packages dir. 

Realistically, does it matter? No. But I don't see why this is better than muting stderr. I won't stand in the way of the PR because of this but I don't like it.

xPaw: I don't like having `node_modules` folder for example, but since this is the easiest way of doing it with npm, we kept it. Ideally we would just have `packages/lounge-theme-xxx` and no extra stuff around it, and it would just work.

But since we chose to use npm in such a way, we have to deal with consequences.

src/command-line/install.js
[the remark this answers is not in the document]
astorije: Right, internally, everything is a package yeah. From a user standpoint, installing a package or a theme is done the same way through `lounge install`, but they are still 2 different things. Just like themes and packages in Atom, they're technically the same thing, but not to the user.

More pragmatically, for this very comment, what I meant is if an admin wants to install a theme and looks at the CLI help, `Install a theme or package` makes it clear for them to use this.

Does that make sense?

src/command-line/install.js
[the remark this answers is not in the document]
astorije: Makes sense re: `npm install`. Welp, even going with (4) would be better as it has one dependencies vs. the many deps `package-json` has, but if you think we should go with `package-json` then :man_shrugging:

src/command-line/install.js
```diff
@@ -0,0 +1,82 @@
...
+			const packagesParent = path.dirname(packagesPath);
+			const packagesConfig = path.join(packagesParent, "package.json");
+
+			// Create node_modules folder, otherwise npm will start walking upwards to find one
+			fsextra.ensureDirSync(packagesPath);
+
+			// Create package.json with private set to true to avoid npm warnings
+			fs.writeFileSync(packagesConfig, JSON.stringify({
+				private: true,
+				description: "Packages for The Lounge. All packages in node_modules directory will be automatically loaded.",
```
astorije: Just for the record, `description` is not needed, but I get why you add it. I'm fine with it.

src/command-line/install.js
astorije: I agree with @xPaw, embracing the npm way is probably better on the long run than trying to go against it 😅

Also having a `package.json` might be a good thing, I can already see one thing or 2 that this could allow us in the future.

src/command-line/install.js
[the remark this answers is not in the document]
YaManicKill: I guess, I'm not really bothered either way.

src/command-line/install.js
YaManicKill: I disagree, but fine, I won't stand in the way. I really would rather we just muted the output from the npm command, and log it if we need. 

> embracing the npm way

But this isn't the npm way, is it? This is us hacking npm to use in our own way. Npm is designed for packages in your own package (which this is not) or global modules for running commands (also this is not), so that's not an argument.

But as I say, I won't stand in the way if you both think this is the better way to do it.

xPaw: Let's keep the output for now, if it's annoying we can remove it in the future.

PR #1540, opened 2017-09-18
Use native font stack

xPaw:
I don't know if you guys prefer this or not, but I think this might be better over all.

This also replaces monospace font in chat with same font as used on `body`.

Reference: https://v4-alpha.getbootstrap.com/content/reboot/#native-font-stack


MaxLeiter:
Works fine on OS X 10.12.6 (Sierra), Safari and Chrome

astorije:
This looks very slick and feels very native, especially on mobile. Only hiccup is that text is now much smaller, effectively canceling out #1153. We shouldn't hold this PR for this, but I think we should bump up the font size a little bit before 2.5.0.

Overall, it's great that default theme(s) use native fonts, and since custom themes currently have limited ways to change the font (outside of Google Fonts), I think that it's a good idea that at the very least by default they use what's nicest to the user and not our choice of Lato.

That's also a performance boost on our end for built-in themes, as fonts can be tough to load on mobile.

> This also replaces monospace font in chat with same font as used on body.

🎉 

> Let's look into themes providing font files as well and this'll be fine (maybe we should wait till that's implemented so people can still use these fonts if they want).

Actually I think that this PR is a good thing for custom themes. That gives sensitive defaults, and the way to extend fonts would be the same anyway. Not sure I'm making sense, but essentially I don't see the harm with this, unless a custom theme really wanted to use Lato and not rely on Google Fonts. But I think that's niche, relying on system font is a much more sensible choice if you don't want to care about fonts for your theme.

PR #1542, opened 2017-09-19
Print compatibility theme setting only once warning on startup

xPaw:


src/server.js
```diff
@@ -197,15 +197,7 @@ function index(req, res, next) {
...
-	var data = _.merge(
-		pkg,
-		Helper.config
-	);
-	if (!data.theme.includes(".css")) { // Backwards compatibility for old way of specifying themes in settings
-		data.theme = `themes/${data.theme}.css`;
-	} else {
-		log.warn(`Referring to CSS files in the ${colors.green("theme")} setting of ${colors.green(Helper.CONFIG_PATH)} is ${colors.bold("deprecated")} and will be removed in a future version.`);
-	}
+	const data = _.cloneDeep(Helper.config);
```
xPaw: Changed to clone because otherwise it would modify the config object in place, I ran into this issue in changelog PR.

PR #1543, opened 2017-09-19
Remove duplicate keybindings help

xPaw:


client/js/lounge.js
```diff
@@ -168,6 +168,10 @@ $(function() {
 		});
 	}
 
+	if (navigator.platform.match(/(Mac|iPhone|iPod|iPad)/i)) {
+		$(document.body).addClass("is-osx");
```
astorije: Technically speaking, there is no OSX terminology anymore so `is-apple` or `is-apple-os` or `is-apple-product` would be more accurate, and maybe more consistent / better for codebase search but up to you, no big deal.

dgw: Is there such a thing as <kbd>Cmd</kbd> on iP{hone|od|ad)? I thought the keyboard shortcuts were only usable on "desktop" devices.

MaxLeiter:
I'm curious if the keybinds work with lightning keyboards -- I assume not

astorije:
@dgw, @MaxLeiter, the question of is there or not a <kbd>Ctrl</kbd>/<kbd>cmd</kbd> key available is totally legitimate, but also more complicated (and maybe even impossible with the current capabilities we have?) than what this PR achieves.
I'd say this is already an improvement over what's in `master`, so while I'm totally happy with discussing this, I think we should still merge this PR.

MaxLeiter:
@astorije : we're just wondering if we should even show key shortcuts on iOS

astorije:
Totally get that :) I just wanted to make it clear that we should still merge this PR and actually find a way to hide shortcuts on iOS/Android in a further PR (but happy with discussion happening here, or in a dedicated issue).

dgw:
Actually, I only even thought about it because of the thing about using `is-osx` vs `is-apple`-something. :grin:

client/js/lounge.js
xPaw: Bluetooth keyboards.

PR #1545, opened 2017-09-19
Match emoji in their textual representation as well

xPaw:
> To match emoji in their textual representation as well (i.e. emoji that are not Emoji_Presentation symbols and that aren’t forced to render as emoji by a variation selector), require the other regex

Test: ♥


xPaw:
Nevermind, it matches shit like `#`.

PR #1549, opened 2017-09-20
Fix opening first channel on server start

### LINKED ISSUE #1547: When running master for the first time, there is no default channel
When running from master, on first launch before there is any active channel, The Lounge just displays no channel, which appears broken on a mobile browser, and may not be clear on a desktop/larger screen. Once a channel is manually selected, it works fine going forward as it remembers what was selected. astorije mentioned they saw the same issue on first launch.

xPaw:
Fixes #1547. Auto reconnect PR broke this behaviour, so I'm bringing it back.

### DIFF
--- a/client/js/socket-events/init.js
+++ b/client/js/socket-events/init.js
@@ -50,7 +50,7 @@ socket.on("init", function(data) {
 });
 
 function openCorrectChannel(clientActive, serverActive) {
-	let target;
+	let target = $();
 
 	// Open last active channel
 	if (clientActive > 0) {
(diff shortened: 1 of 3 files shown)


client/js/socket-events/init.js
```diff
@@ -50,25 +50,30 @@ socket.on("init", function(data) {
...
+	let target = $();
 
 	// Open last active channel
 	if (clientActive > 0) {
 		target = sidebar.find("[data-id='" + clientActive + "']");
 	}
 
 	// Open window provided in location.hash
-	if (!target && window.location.hash) {
+	if (!target.length > 0 && window.location.hash) {
```
astorije: === 0
Same below

xPaw: oops

astorije: Instead of !...> 0 I mean

xPaw: Fixed it.

Issue #1551, opened 2017-09-20
Split style.css into subcomponents and preprocess css as a build step


xPaw:
Our style.css is over 2000 lines long, and it's a mess. We definitely need to split it into some smaller components like we did with javascript code.

And then as a build step have it assembled and minified (auto prefixed too!) into a single file.

xPaw:
We are able to move styles into their vue components now, we should do that.

bumpylumps:
If this is still open I'd be happy to break up that CSS!

MaxLeiter:
@bumpylumps currently I'd wait for #4559, and then if you want to update #4494 feel free.

bumpylumps:
> @bumpylumps currently I'd wait for #4559, and then if you want to update #4494 feel free.

Sounds good! I'll keep an eye on it

PR #1552, opened 2017-09-21
Use monospace font in custom CSS textarea

astorije:
This is definitely the best CSS ever (far from it really) but when tackling #1551 this should fall into place. The end result is much nicer than not having this anyway, and not specifying placeholder makes for a weird look when empty.

screen shot 2017-09-21 at 01 59 26 screen shot 2017-09-21 at 02 00 01

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -83,7 +83,8 @@ button {
 }
 
 code,
-kbd {
+kbd,
+textarea#user-specified-css-input {
 	font-family: Consolas, Menlo, Monaco, "Lucida Console", "DejaVu Sans Mono", "Courier New", monospace;
 }
 
--- a/client/index.html
+++ b/client/index.html
@@ -376,7 +376,7 @@
 								<h2>Custom Stylesheet</h2>
 							</div>
 							<div class="col-sm-12">
-								<textarea class="input" name="userStyles" id="user-specified-css-input" placeholder="You can override any style with CSS here"></textarea>
+								<textarea class="input" name="userStyles" id="user-specified-css-input" placeholder="/* You can override any style with CSS here */"></textarea>
 							</div>
 						</div>
 


xPaw:
Can we change placeholder to look like a comment `/* Put custom css here */` instead of messing with fonts?

MaxLeiter:
I agree with @xPaw about the placeholder, but also think the custom CSS box should be monospaced

xPaw:
@MaxLeiter that is the idea to use monospaced.

astorije:
So essentially this:

screen shot 2017-09-22 at 01 58 16

I don't like the inconsistency between placeholders.

I prefer the result given in the PR description, even if that means having multiple definition of the font stack. Again, once #1551 is taken care of, that won't be an issues anymore I believe.

Thoughts?

xPaw:
I really don't like the fact you're adding 3 copypasted definitions of the font stack just for a minor change here. The screenshot you showed looks completely fine.

YaManicKill:
I agree with @xPaw. If it looks different with text in, what's wrong with it looking different in the placeholder as well? I'm not a fan of the placeholder being a different font from the normal text.

astorije:
Alright then, since I'm outnumbered lol, I've made the changes:

Before this PR | After this PR
--- | ---
screen shot 2017-09-22 at 23 48 33 <p align="center">✅</p> | screen shot 2017-09-22 at 23 47 31 <p align="center">❌</p>
screen shot 2017-09-22 at 23 48 23 <p align="center">❌</p> | screen shot 2017-09-22 at 23 48 11 <p align="center">✅</p>

Though I gotta say, I really hate the inconsistency between the placeholders. I know you guys all have custom CSS, and I actually don't, so I wish I didn't open this PR 😅
But here we are and I think the benefits of monospaced font when there is text trump the style of the placeholder. I hope however that we'll be okay to re-evaluate this when using variables in our CSS.

astorije:
Actually, I have just tried something that I think is pretty cool using CSS variables: it lets us _not_ set the same font stack multiple times, keeps both ✅ screenshots above (🎉), and users can just override variables in custom CSS, which will be more robust than overriding selectors on the long run:

screen shot 2017-09-23 at 00 03 57

Also, once we go tackle #1551, the auto-prefixer will take care of the multiple placeholder overrides.



> If it looks different with text in, what's wrong with it looking different in the placeholder as well?

Sorry @YaManicKill I forgot to answer. My 2c was that the page looked more cohesive when placeholders all share the same style. But both sides of the argument are opinions, I'm not saying there is a right and a wrong :)

astorije:
@xPaw, reason why I originally set 2.5.0 milestone is because I thought the textarea used to be monospaced but stopped doing so after removing monospace on messages, for some reason.
Since it's already like that in 2.4.0 actually, I'm fine not having it in 2.5.0.

astorije:
Thoughts on this?

xPaw:
@astorije I'd merge the first commit for 2.5, but not the second.

YaManicKill:
Yeah, I'd be happy with the same as @xPaw 

I still don't like the placeholder being a different font to the text. I get what you are saying about wanting the placeholders to be all the same, but when you have custom text, they won't look the same...

dgw:
> Can we change placeholder to look like a comment `/* Put custom css here */` instead of messing with fonts?

I'm on board with the placeholder for custom CSS looking like CSS, not like the other non-code placeholders. To me, it's odder for an element's font to change when I start typing in it than for the placeholder to be different.

astorije:
Okay then, will remove.

astorije:
Done.

PR #1553, opened 2017-09-21
Increase font size on desktops and mobiles

### LINKED ISSUE #841: Small default font size

astorije:
Fixes #841.

This increases fonts that use to be monospace before #1540 and ended up being smaller as a side effect.

- Increase by a little bit on message window (except on message input, unfortunately, but that was adding a scrollbar to the input, then fixing this would misalign borders, etc. not for this PR).
- Increase a bit more on tablets and phones. The result is really great there, finally not tiny texts anymore, and this + native font stack really makes it look like a native chat app.
- Remove the font size for extra large screens now that message window size is bigger.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -855,8 +855,12 @@ kbd {
 #windows .header .topic,
 .messages .msg,
 .sidebar {
-	font-size: 13px;
+	font-size: 14px;
 	line-height: 1.4;
+}
+
+#windows #form .input {
+	font-size: 13px;
 }
 
 #windows #chat .header {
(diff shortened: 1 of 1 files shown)


xPaw:
https://github.com/thelounge/lounge/blob/929812f85d75a921e050caed727e6af7e7b4c465/client/css/style.css#L2054-L2060

Is this unnecessary now?

astorije:
Good catch @xPaw, will remove!

astorije:
Done!

christer88:
I'm not so sure I fancy larger text on phones. It's already a small screen with not much of the channel showing. Larger text would mean less communication is showing and more scrolling. 👎

astorije:
You can always decrease the font-size using custom CSS to your liking, but by default right now the font size is waaaay too small, much smaller than any other chat app (or app at all) I've been encountering.

PR #1556, opened 2017-09-22
Toggle user list tooltip shouldn't appear on mobile

Gilles123:
Minor thing, but all other tooltips are suppressed on mobile, and the "Toggle user list" one isn't. I think it should be.

### DIFF
--- a/client/views/chat.tpl
+++ b/client/views/chat.tpl
@@ -3,7 +3,7 @@
 	<div class="header">
 		<button class="lt" aria-label="Toggle channel list"></button>
 		{{#equal type "channel"}}
-			<span class="rt-tooltip tooltipped tooltipped-w" aria-label="Toggle user list">
+			<span class="rt-tooltip tooltipped tooltipped-w tooltipped-no-touch" aria-label="Toggle user list">
 				<button class="rt" aria-label="Toggle user list"></button>
 			</span>
 		{{/equal}}


astorije:
So the reason why we disabled _some_ tooltips was because of their behavior with regard to their context. For example, on the channel list, clicking a footer icon would bring up the tooltip for a very short period of time until the animated menu is closed. The Send Message icon was kind of annoying too.
Why do you think we should remove this? Any inconvenience?
At some point we actually discussed about _not_ disabling the tooltips anymore (see https://github.com/thelounge/lounge/pull/824#discussion_r110471542) but that was never changed.

Gilles123:
I don't see any other tooltip that appears on mobile, and on mobile (at least on iOS), it usually results in having to click the button twice (first time displays the tooltip, second time to actually use the button). Not good UX in my opinion.

astorije:
> Not good UX in my opinion.

That I agree, but it seems to be a bug, not a conscious choice.
And the reason why I am reticent to merge this PR as-is is because it is removing the symptom while we probably have a real issue somewhere. You shouldn't have to click twice to open the channel list.

> at least on iOS, it usually results in having to click the button twice (first time displays the tooltip, second time to actually use the button).

Alright, let's see if we can find the root cause (I don't have an iOS device so bear with me 😅).
There was a tooltip on the "Send message" button but it's been removed because the tooltip stayed open each time the button was clicked, which is really bad on the message input (see https://github.com/thelounge/lounge/pull/612).
However, tooltips on the nick editor are still here.

Could you:

- Try to change your nick to see if you need to click twice before you can edit it?
- Re-enable the "Send message" tooltip (i.e. remove `tooltipped-no-touch` [here](https://github.com/thelounge/lounge/blob/master/client/index.html#L76)) and see if you need to click twice to send a message?

As you probably noted, I'm trying to see if this is an issue on all our tooltips, or something only on the channel list button.

---

The reality is that Primer Tooltips are not good on mobile. In fact, [they have told us that responsive was not handled yet](https://github.com/primer/primer-tooltips/issues/4#issuecomment-261139740). I'm wondering if we should work with them on that, or if we should look into another library, or if patching up here and there is still the most logical action we can take.

Gilles123:
Turns out the behaviour is not exactly as I thought it was. It's not that I was clicking twice, it's that the clickable area that causes the tooltip to appear is slightly larger than the one that causes the button to function. If I click right on the button, it works every time, but slightly to the right of it causes the tooltip to appear and not trigger the button. Sorry, I should've tested this more before submitting. Looking at the code, I'm not sure anything could be done about that as it's probably just a rendering quirk.

While I was testing the other tooltips though, I realized that the nick change function doesn't appear to work on iOS. The buttons change and all, but the nick does not appear to be editable. That'd be for a different bug though, and I'll test that some more before submitting it.

astorije:
> It's not that I was clicking twice, it's that the clickable area that causes the tooltip to appear is slightly larger than the one that causes the button to function.

Nice catch! That might be fixable though. Do you want to give it a shot?
Might want to close this PR and open a new PR for this though, to keep code history as an archive type of thing.

> Sorry, I should've tested this more before submitting.

No worries at all, that's perfectly fine, and we all do that at some point. Discussing about it is the best way to identify the cause, and opening an early issue or PR is how to discuss about it 😁 

> I realized that the nick change function doesn't appear to work on iOS.

Great catch too! I'm on Android where it works fine, and don't have an iOS device to test (for now!) but it looks like someone else could reproduce. Anyway, thanks for opening that other PR!

astorije:
> It's not that I was clicking twice, it's that the clickable area that causes the tooltip to appear is slightly larger than the one that causes the button to function.

Nice catch! That might be fixable though. Do you want to give it a shot?
Let's close this PR and I'll let you open a new PR for this if you are willing to give it a shot though, to keep code history as an archive type of thing.

> Sorry, I should've tested this more before submitting.

No worries at all, that's perfectly fine, and we all do that at some point. Discussing about it is the best way to identify the cause, and opening an early issue or PR is how to discuss about it 😁 

> I realized that the nick change function doesn't appear to work on iOS.

Great catch too! I'm on Android where it works fine, and don't have an iOS device to test (for now!) but it looks like someone else could reproduce. Anyway, thanks for opening that other PR!

Issue #1558, opened 2017-09-23
Fonts are huge after 3eaf12


Zbob750:
[This commit bumped font sizes](https://github.com/thelounge/lounge/commit/3eaf12cc36400f6698286dda8c9200cf546d6104) and, in my opinion, you've gone to the extreme.
On Desktop it feels just a little big, but really not anything that should cause problems. However, on mobile this change impacts usability. It has diminished the amount of content that fits on screen. 

<img src=https://i.imgur.com/cdgdv3F.png width=300>

In my opinion, this change is unnecessary on the desktop and hurtful to mobile.
I would like to see the change just reverted. I realize that CSS can be used to alter this for my own personal tastes, however I really don't think I should need to maintain a theme to make the mobile client show more information. I also think that many people who upgrade from the current release to the upcoming release will find this change jarring.

Thanks

astorije:
> However, on mobile this change seriously impacts readability. The change has seriously diminished the amount of content that fits on screen.

I understand about the reduced amount of content (though 25 lines as shown on your screenshot is still more than what Telegram would show you for example) but readability _is_ better as it is right now, as in it's less hard to read.

I'm going to run some more tests to tweak the current size, but I am certainly against a simple revert.

astorije:
Resolved by #1561.

PR #1561, opened 2017-09-23
Reduce message size on mobile and increase a bit UI text

astorije:
I didn't intend to mix those together, but I happened to work in the same branch, and the results kind of depend on each other anyway.

### Reduce message size on mobile and increase a bit UI text

Some feedback on the channel and on #1558 mentioned that the messages on mobile are too big.
I don't entirely agree as they respect current trends (for example, Telegram ships by default with a 16px size on Android) but I understand that the gap between the previous state and what is currently on `master` was too abrupt and threw people off, so I set `15px` as a compromise.

There were discussions to make this configurable, but it's further down the road and this just tries to find a common ground right now.

Also make the rest of the app on mobile a tad bigger to match the message size and to make it more usable. The jump here is very small compared to messages so it will be more subtle.

Also removed some `font-size` overrides that were not needed anymore.

### Make channel list links clickable on full width, simplify code 

 - Close button and badge are now positioned using the same method. Previously one was float and the other was absolutely positioned.
- Increase the close button size to match 2.4.0, as it was made smaller when moving to the native font stack.
- Removed link margins on left/right of the channel links, and the `1px` gap between links.
- Removed hack/fix for drag-and-drop now made unnecessary

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -475,7 +475,7 @@ kbd {
 }
 
 #sidebar .networks {
-	padding: 20px 30px 0;
+	padding-top: 20px;
 	touch-action: pan-y;
 }
 
@@ -498,17 +498,9 @@ kbd {
 #sidebar .chan,
 #sidebar .chan-placeholder {
 	display: block;
-	margin: 1px -10px;
-	padding: 6px 10px 8px 36px;
-	position: relative;
-	text-align: left;
+	padding: 8px 25px;
 	transition: color 0.2s;
-	width: 180px;
-	left: auto !important; /* Fix for drag'n'drop not recalculating left position */
-}
-
-#sidebar .chan-placeholder {
-	padding-bottom: 10px;
+	width: 100%;
 }
 
 #sidebar .chan:first-child {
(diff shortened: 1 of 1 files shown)


YaManicKill:
Oh goodness, more discussions about text sizes. Boy do I love that.

astorije:
It's also a cleanup ;)

client/css/style.css
```diff
@@ -2004,8 +1999,20 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+	#windows .header .topic,
+	#settings .error,
+	#help .help-item,
+	#help .about,
+	#loading,
+	#context-menu,
+	.textcomplete-menu,
 	.messages .msg {
-		font-size: 16px;
+		font-size: 15px;
```
xPaw: Since there are so many overrides, which elements keep the 16px font?

astorije: None, 16px was set to `.messages .msg`, which is now set to 15px. Everything else is set to 15px on mobile (they are all 14px on desktop).

xPaw: So why aren't you just setting the font-size on body or something?

astorije: Because at the moment, aaaaaall those things are overriding stuff everywhere. In the long run, I'd like to do that, body or base text be at a given size, and then do specific sizes for other elements (titles, etc., even better if all in `rem`) so that only overriding the base would be enough.
But I won't be able to do this before v2.5.0, and it's a rather extensive change, so I'm getting this out of the way and we can improve later.

astorije: Font sizes are a real mess right now. I don't intend to fix that with this PR :)

PR #1562, opened 2017-09-23
Remove package-json dependency in favor of registry-url + custom code

astorije:
This is just me playing with a POC, not an actual PR ready to be merged.

@xPaw, would it make sense to do something like this? `registry-url` could be replaced with code easily but not `registry-auth-token`, and unless I'm wrong, if we support custom registries, we should probably keep that latter one.

This is totally up for debate/discussion, and I know it has about 10 times more chances to be closed than merged, I'm cool with that, but this version has a couples benefits:

- More optimized as it only downloads `latest` JSON (until we support specifying a version, though `package-json` doesn't seem to work with `package@version` format, IIUC)
- Better control over error messages (when a package is not found for example)

I don't know, this is probably stupid, but I wanted to give it a try. Feel free to close if it is.


astorije:
This PR does add have few advantages, listed above (on top of removing a dependency with extra other dependencies), but it is not a complete necessity right now.
Closing the PR for now, and if we need it in the (distant) future, it will be archived in here.

PR #1565, opened 2017-09-23
Fix wrongly positioned menu when opening it and switching to landscape

### LINKED ISSUE #1564: Opening network menu and resizing the window results in empty space
Opening the menu adds `menu-open`  class, which opens it on mobile window, but resizing the window to desktop size moves it even further.

astorije:
Fixes #1564.

This could be reproduced on tablet or large phone (as long as portrait mode shows a collapsible menu and landscape mode shows normal menu) or when resizing the browser on desktop.

I simply moved the definition of those collapsible-menu-related attributes only where it behaves that way. This also has the advantage of not animating the menu anymore when changing orientation for example.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -353,26 +353,6 @@ kbd {
 	overflow: hidden;
 }
 
-#viewport {
-	height: 100%;
-	transition: transform 160ms, -webkit-transform 160ms;
-	-webkit-transform: translateZ(0);
-	transform: translateZ(0);
-}
-
-#viewport.menu-open {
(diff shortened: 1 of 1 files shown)


xPaw:
Ideally we would allow the menu to work on desktop correctly too.

astorije:
I'm not sure it's a good idea to have collapsible menu on desktop. Most chats I know don't. But if we do, there is a bit of work (and possibly trick) involved, we want it to persist in refresh, and it should be open by default on desktop, closed by default on mobile.

PR #1567, opened 2017-09-24
Abort image prefetch if Content-Length exceeds limit

### LINKED ISSUE #1563: Image prefetch: Abort if Content-Length exceeds maximum
Discussion in #1554 got me to look at the link prefetch code for images, which I had assumed would not try to fetch images in excess of the configured `prefetchMaxImageSize`. Apparently, the code instead fetches image data until either the response ends (the image size is below the limit) or the buffered data exceeds the limit (at which point the request is aborted).

With the default maximum image size now being 2MB (as of #1490), that means potentially 2MB of "wasted" network traffic every time a link to an image bigger than the limit is received. I'm not sure when the response headers become available (and #1554 will change how the code works anyway), but I do think it would be sensible to abort the request as soon as possible if the `Content-Length` header indicates that the image exceeds the limit, in the interest of using bandwidth as efficiently as possible on both The Lounge's end and on the image host's end.

_Edited to add:_ I'm interested in tackling this after #1554 lands (and likely after 2.5.0 stabilizes), if the idea is approved. I'm not just idly demanding it :laughing:

dgw:
Does adding this beg for the addition of tests? If so, someone will probably have to guide me in doing that. :grin:

Closes #1563.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -159,7 +159,14 @@ function fetch(uri, cb) {
 	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/^image\/.+/.test(res.headers["content-type"]))) {
+			if (/^image\/.+/.test(res.headers["content-type"])) {
+				// response is an image
+				// if Content-Length header reports a size exceeding the prefetch limit, abort fetch
+				const contentLength = parseInt(res.headers["content-length"], 10) || 0;
+				if (contentLength > limit) {
+					req.abort();
+				}
+			} else {
 				// if not image, limit download to 50kb, since we need only meta tags
 				// twitter.com sends opengraph meta tags within ~20kb of data for individual tweets
 				limit = 1024 * 50;


src/plugins/irc-events/link.js
```diff
@@ -163,6 +163,13 @@ function fetch(uri, cb) {
 				// if not image, limit download to 50kb, since we need only meta tags
 				// twitter.com sends opengraph meta tags within ~20kb of data for individual tweets
 				limit = 1024 * 50;
+			} else {
+				// response is an image
+				// if Content-Length header reports a size exceeding the prefetch limit, abort fetch
+				const contentLength = parseInt(res.headers["content-length"], 10) || NaN;
```
xPaw: parseInt returns NaN on it's own, I would change it to 0 though, then you don't need to the AND check below.

src/plugins/irc-events/link.js
```diff
@@ -163,6 +163,13 @@ function fetch(uri, cb) {
 				// if not image, limit download to 50kb, since we need only meta tags
 				// twitter.com sends opengraph meta tags within ~20kb of data for individual tweets
 				limit = 1024 * 50;
+			} else {
```
xPaw: Can you invert the if above, it will look cleaner.

src/plugins/irc-events/link.js
dgw: Testing in Chrome's dev console with a simple literal object, running parseInt on a key that didn't exist set the test var to `undefined`. But using 0 instead is sensible. I went with NaN and the `&&` in case a negative limit (i.e. no limit) is/will be possible.

src/plugins/irc-events/link.js
dgw: I'm not sure what you're asking me to do, but my interpretation is to make the first if test for whether it is an image, rather than not, and swap the code blocks.

xPaw: Correct.

dgw: Done.

src/plugins/irc-events/link.js
```diff
@@ -159,7 +159,14 @@ function fetch(uri, cb) {
 	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/^image\/.+/.test(res.headers["content-type"]))) {
+			if ((/^image\/.+/.test(res.headers["content-type"]))) {
+				// response is an image
+				// if Content-Length header reports a size exceeding the prefetch limit, abort fetch
+				const contentLength = parseInt(res.headers["content-length"], 10) || NaN;
+				if (contentLength && contentLength > limit) {
+					res.req.abort();
```
xPaw: This can be `res.abort`.

src/plugins/irc-events/link.js
```diff
@@ -159,7 +159,14 @@ function fetch(uri, cb) {
 	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/^image\/.+/.test(res.headers["content-type"]))) {
+			if ((/^image\/.+/.test(res.headers["content-type"]))) {
+				// response is an image
+				// if Content-Length header reports a size exceeding the prefetch limit, abort fetch
+				const contentLength = parseInt(res.headers["content-length"], 10) || NaN;
```
xPaw: Do the 0 as I suggested.

src/plugins/irc-events/link.js
```diff
@@ -159,7 +159,14 @@ function fetch(uri, cb) {
 	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/^image\/.+/.test(res.headers["content-type"]))) {
+			if ((/^image\/.+/.test(res.headers["content-type"]))) {
```
xPaw: You can remove extra parenthesis.

src/plugins/irc-events/link.js
dgw: OK, I was copying the code in the pipe handler below because I couldn't find a reference to `abort()` in the docs for `request`. For future reference, where is it documented that you can call `abort()` directly on the `response` passed to this event handler?

src/plugins/irc-events/link.js
dgw: This and the other comments in the batch are done.

src/plugins/irc-events/link.js
```diff
@@ -159,7 +159,14 @@ function fetch(uri, cb) {
 	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/^image\/.+/.test(res.headers["content-type"]))) {
+			if (/^image\/.+/.test(res.headers["content-type"])) {
+				// response is an image
+				// if Content-Length header reports a size exceeding the prefetch limit, abort fetch
+				const contentLength = parseInt(res.headers["content-length"], 10) || 0;
+				if (contentLength > limit) {
+					res.abort();
```
astorije: @xPaw, in #1554, we went for `req.abort()`. I wonder if both do the same thing, but should we use the same thing then? Which do you prefer?

xPaw: abort calls abort on req internally, so yes.

src/plugins/irc-events/link.js
```diff
@@ -159,7 +159,14 @@ function fetch(uri, cb) {
 	var limit = Helper.config.prefetchMaxImageSize * 1024;
 	req
 		.on("response", function(res) {
-			if (!(/^image\/.+/.test(res.headers["content-type"]))) {
+			if (/^image\/.+/.test(res.headers["content-type"])) {
+				// response is an image
+				// if Content-Length header reports a size exceeding the prefetch limit, abort fetch
+				const contentLength = parseInt(res.headers["content-length"], 10) || 0;
```
astorije: I know you guys had some discussion about this so sorry if I missed something, but it looks to me that you don't even need `|| 0` (which I find more misleading than explicit):

```js
> const limit = 2048
undefined
> parseInt({}["content-length"], 10)
NaN
> NaN > limit
false
> null > limit
false
> undefined > limit
false
> 4096 > limit
true
```

So whatever `parseInt` returns, if the value is not an integer, the check below will be `false`.

src/plugins/irc-events/link.js
astorije: @dgw, mind switching to `req.abort()` then?

src/plugins/irc-events/link.js
xPaw: It's not misleading as it will force an integer type.

astorije: It's not necessary, I don't think we should do it.

src/plugins/irc-events/link.js
dgw: There's no `req` parameter passed to this handler, only `res`.

src/plugins/irc-events/link.js
dgw: I don't like comparing the wrong data type if I can avoid it. Passing NaN into anything number-related feels kinda dirty…

astorije: Fair enough, I guess.

src/plugins/irc-events/link.js
astorije: You have access to `req` from L160. It's also being used at R179 :)

dgw: What. Has variable scoping has changed since I learned JS (over a decade ago, I feel old :grin:)? L160 is outside the function definitions in both cases; how does _that_ work? :confused:

(Also, L179 seems silly with the `req.response.req` thing… It's not related to this PR but now I really want to make it just `req.abort()`…)

xPaw: `res.abort` is correct here, don't change it.

PR #1568, opened 2017-09-24
Fix nick change on Safari for Mac and iOS

Gilles123:
It is currently impossible to edit the nick text after pressing the edit button on Safari for iOS and Mac because user-select is set to None on Safari by default. This fixes that issue. Tested that it is not editable when it shouldn't be and that it doesn't break anything on Chrome or Firefox.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1498,6 +1498,14 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	outline: none;
 }
 
+[contenteditable="true"] {
+	-webkit-user-select: text;
+	-moz-user-select: text;
+	-ms-user-select: text;
+	user-select: text;
+	cursor: text;
+}
+
 /* Nick editor */
 
 #form #nick {


client/css/style.css
```diff
@@ -1496,6 +1496,8 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 [contenteditable]:focus {
 	outline: none;
+	-webkit-user-select: text;
+	user-select: text;
```
astorije: If this is the right fix, you'll want to use the full set of prefixes we use for example [here](https://github.com/thelounge/lounge/blob/master/client/css/style.css#L161-L167):

```css
-webkit-user-select: text;
-moz-user-select: text;
-ms-user-select: text;
user-select: text;
```

While you're at it, could you add `cursor: text;` please? Thanks!

astorije:
Thanks for this, @Gilles123!

I know this is broken in v2.4.0 as well, but I'm marking this for v2.5.0 milestone, because fix seems easy, and the bug is bad enough :)

@MaxLeiter, could you give this a go please?

client/css/style.css
Gilles123: I didn't include those because as far as I can tell, only WebKit/Safari uses "none" as its default and the other browsers worked as expected. For certainty's sake, we can do that though.

astorije: Yeah, let's be consistent, and whenever we tackle #1551, we'll let the auto-prefixer get on with it :)

xPaw:
Shouldn't this be set on `[contenteditable]` without `:focus`?

Gilles123:
It works either way. I changed it to be without focus.

xPaw:
Approved, but can you squash your commits?

MaxLeiter:
Works for me 👍

client/css/style.css
```diff
@@ -1498,6 +1498,14 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	outline: none;
 }
 
+[contenteditable] {
```
astorije: Actually, make this `[contenteditable="true"]` so it's only selectable + correct cursor in edit mode.

xPaw: I don't think that's necessary in lounge?

astorije: Why not? That's what we do everywhere else in the app except messages, i.e. UI not selectable. With this addition, it stays like that, except when the nick is being edited, when it can be selected and have the text cursor. Makes sense?

astorije:
(Hey @Gilles123, do you happen to have a Twitter account to give you a shout-out after we release v2.5.0? 🙂)

astorije:
Hey @Gilles123, we have free sticker packs for our contributors now!

PR #1578, opened 2017-09-27
Handle `CHGHOST` cap

xPaw:
This needs testing on a network that supports it.

### DIFF
--- a/client/js/condensed.js
+++ b/client/js/condensed.js
@@ -29,6 +29,9 @@ function updateText(condensed, addedTypes) {
 			case "back":
 				strings.push(obj[type] + (obj[type] > 1 ? " users have come back" : " user has come back"));
 				break;
+			case "chghost":
+				strings.push(obj[type] + (obj[type] > 1 ? " users have changed hostname" : " user has changed hostname"));
+				break;
 			case "join":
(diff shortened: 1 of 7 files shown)


MaxLeiter:
Testing on `testnet.inspircd.org` by changing my mode to `+x` on one client and watching with another.

Don't currently have time to debug but the host being changed isn't reported at all.

xPaw:
@MaxLeiter https://github.com/kiwiirc/irc-framework/commit/c725576be81c9b32c394ded5ea3fc25d4a02cd26 this?

xPaw:
I just tested it with the fix applied, it shows the new hostname correctly.

MaxLeiter:
Works for me with fix applied, but is this really something we want users to see? I think `CHGHOST` is more useful for state management (i.e. updating `user` objects), not necessarily for notifying users.

astorije:
Still a do not merge, @xPaw?

xPaw:
@astorije Needs a irc-fw update

astorije:
@xPaw, now that `irc-framework` has been bumped, does this need changes before we can merge?

astorije:
Also, should this be added to the IRCv3 GitHub project? (I believe that would be https://github.com/thelounge/lounge/projects/1#card-242870)

astorije:
Rebased on `master` and fixed ESLint issue.

PR #1579, opened 2017-09-27
Add info message before calling `packageJson`

xPaw:
I had some network hiccup, so the CLI seemingly didn't do anything until the packageJson callback got called and the message that package is installing displayed.

### DIFF
--- a/src/command-line/install.js
+++ b/src/command-line/install.js
@@ -21,6 +21,8 @@ program
 			return;
 		}
 
+		log.info("Retrieving information about the package...");
+
 		packageJson(packageName, {
 			fullMetadata: true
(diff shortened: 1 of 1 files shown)


src/command-line/install.js
```diff
@@ -21,6 +21,8 @@ program
 			return;
 		}
 
+		log.info("Checking requested package data on npm...");
```
astorije: What about `Retrieving information about the package, please wait...`?

xPaw: Perhaps `Retrieving information about the package...` ?

astorije: 👍

PR #1580, opened 2017-09-28
Use service worker to display notifications if available

xPaw:
Displays notifications if the client is active on mobile devices (new Notification fails there)

### DIFF
--- a/client/js/libs/handlebars/ircmessageparser/cleanIrcMessage.js
+++ b/client/js/libs/handlebars/ircmessageparser/cleanIrcMessage.js
@@ -1,0 +1,4 @@
+"use strict";
+
+// TODO: This does not strip hex based colours - issue #1413
+module.exports = (message) => message.replace(/\x02|\x1D|\x1F|\x16|\x0F|\x03(?:[0-9]{1,2}(?:,[0-9]{1,2})?)?/g, "").trim();
--- a/.npmignore
+++ b/.npmignore
@@ -7,9 +7,10 @@
 !.lounge_home
 
 # Ignore client folder as it's being built into public/ folder
-# except for the findLinks.js file which is used by the server
+# except for the specified files which are used by the server
 client/**
 !client/js/libs/handlebars/ircmessageparser/findLinks.js
+!client/js/libs/handlebars/ircmessageparser/cleanIrcMessage.js
 
 public/js/bundle.vendor.js.map
 coverage/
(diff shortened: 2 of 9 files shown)


src/cleanIrcMessage.js
```diff
@@ -0,0 +1,4 @@
+"use strict";
+
+// TODO: This does not strip hex based colours
+module.exports = (message) => message.replace(/\x02|\x1D|\x1F|\x16|\x0F|\x03(?:[0-9]{1,2}(?:,[0-9]{1,2})?)?/g, "").trim();
```
MaxLeiter: This fails on Node 8:

https://travis-ci.org/thelounge/lounge/jobs/280780940#L2204

astorije: I sure am glad we run production builds in our CI!

According to https://stackoverflow.com/questions/43888474/unexpected-token-operator-from-uglifyjs, Webpack relies on a version that doesn't support ES6 yet, so hopefully by the time this PR gets merged, they will be done (?).
I'm really surprised why this arrives only now though, since that's not the first fat-arrow function we use on the server...

xPaw: That was because `src` folder wasn't included in babel transpiling.

astorije: Do we want to add the `src` though? I don't know, I feel odd about that. We have client code we use on the server and now server code we use on the client :/

Also does that mean more gets added to the client bundle? I compared and it doesn't seem to any extra kb heavier.

xPaw: I could move this file to client (and reference it on the server like we already do with findLinks) instead, this will at least keep the consistency in this regard.

astorije: That seems doing yeah

astorije: Fine*

Damn, 1.5month to notice my typo...

client/js/libs/handlebars/ircmessageparser/cleanIrcMessage.js
```diff
@@ -0,0 +1,4 @@
+"use strict";
+
+// TODO: This does not strip hex based colours - issue #1413
+module.exports = (message) => message.replace(/\x02|\x1D|\x1F|\x16|\x0F|\x03(?:[0-9]{1,2}(?:,[0-9]{1,2})?)?/g, "").trim();
```
astorije: You need to add this file to [`.npmignore`](https://github.com/thelounge/lounge/blob/master/.npmignore#L12) now, right?

astorije:
Rebased on master and fixed conflicts (ESLint stuff only).

client/js/libs/handlebars/ircmessageparser/cleanIrcMessage.js
xPaw: Done.

PR #1583, opened 2017-09-29
Use `thelounge` instead of `lounge` when it comes to external packages

astorije:
We keep having confusions between `lounge`  and `thelounge` everywhere (sorry.....).

When v3 shows up, we should be able to get rid of the inconsistency entirely. In the meantime, new stuff, especially external packages that will rely on fairly stable contracts, should honor that to make the transition easier. Apart from those, nothing should be problematic.

### DIFF
--- a/src/plugins/themes.js
+++ b/src/plugins/themes.js
@@ -62,11 +62,11 @@ function getModuleInfo(packageName) {
 		log.warn(`Specified theme ${colors.yellow(packageName)} is not installed in packages directory`);
 		return;
 	}
-	if (!module.lounge) {
+	if (!module.thelounge) {
 		log.warn(`Specified theme ${colors.yellow(packageName)} doesn't have required information.`);
 		return;
 	}
-	return module.lounge;
+	return module.thelounge;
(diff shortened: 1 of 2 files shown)


src/plugins/themes.js
```diff
@@ -62,11 +62,11 @@ function getModuleInfo(packageName) {
 		log.warn(`Specified theme ${colors.yellow(packageName)} is not installed in packages directory`);
 		return;
 	}
-	if (!module.lounge) {
+	if (!module.thelounge) {
 		log.warn(`Specified theme ${colors.yellow(packageName)} doesn't have required information.`);
 		return;
 	}
-	return module.lounge;
+	return module.thelounge;
```
astorije: Unrelated to this PR, and also I realize that it doesn't hurt to be on the safe side, but now that the `install` command takes care of checking the metadata, maybe we should assume that if a package reaches this point, it has the correct metadata, no?

xPaw: It's better to be on the safe side.

PR #1588, opened 2017-09-29
Expand user model

zchrykng:
Captures more data in the User model. Is needed for other features I am working on. Such as automatically banning using accounts when the server supports it.

### DIFF
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -10,6 +10,10 @@ function User(attr, prefixLookup) {
 		away: "",
 		mode: "",
 		nick: "",
+		host: "",
+		account: "",
+		ident: "",
+		gecos: "",
 		lastMessage: 0,
 	});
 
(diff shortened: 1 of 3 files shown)


MaxLeiter:
Providing this data to the client (maybe not for this PR) will allow functionality like #356

YaManicKill:
I'm quite a strong proponent of only putting code in when we actually need it. Why would we merge this in without anything actually using that code? I'd suggest leaving it until something actually needs that. Having extra data lying around when we don't need it feels a little unnecessary.

astorije:
I'm with @YaManicKill on that one. Unused code is dead code, and since it's unused (and untested), it's likely that one of us will break it silently. Then good luck fixing that 😅

@zchrykng, maybe you could instead build on your branch/commits, adding your feature there, and when you're done, the whole PR would make more sense?

xPaw:
I agree with both. Going to close this for now.

zchrykng:
Yep makes sense. When I have time to finish it I will do a pull request for the finished feature.

PR #1589, opened 2017-09-29
Update express to version 4.16.0

xPaw:


dgw:
Greenkeeper slacking?

astorije:
Express doesn't set the correct dist-tags on prereleases, which we avoid to stay in stable versions, then does not re-submit stable versions...

PR #1593, opened 2017-09-30
Ensure passwords are reloaded when updated via CLI

### LINKED ISSUE #598: lounge reset - reset password command requires a server restart.
In testing the tools for lounge.  The 'lounge reset' command does change the hash in the json file for the user but the password change doesn't take effect until the server is stopped and started again.  I've noticed in the clientManager.js that it checks for users being added or removed but from what I can tell it doesn't check for changes in the json file.

RJacksonm1:
As title, fixes #598

I didn't update the entire `config` object, cause I haven't the slightest clue what possible consequences that could have.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -51,8 +51,8 @@ ClientManager.prototype.autoloadUsers = function() {
 			log.info(noUsersWarning);
 		}
 
-		// New users created since last time users were loaded
-		_.difference(updatedUsers, loaded).forEach((name) => this.loadUser(name));
+		// Reload all users. Existing users will only have their passwords reloaded.
+		updatedUsers.forEach((name) => this.loadUser(name));
 
 		// Existing users removed since last time users were loaded
 		_.difference(loaded, updatedUsers).forEach((name) => {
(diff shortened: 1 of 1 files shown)


xPaw:
This generally sounds fine, but this will start triggering `loadUser` every time Lounge itself updates the user file, which might be undesired.

astorije:
Any alternative you can think of?

RJacksonm1:
@xPaw Is the user config modified enough for that to be a performance concern?

src/clientManager.js
```diff
@@ -78,12 +78,25 @@ ClientManager.prototype.loadUser = function(name) {
...
+		this.clients.push(client);
+	}
+
+	/**
+	 * If we happen to reload an existing client, make super duper sure we have their latest password.
+	 * We're not replacing the entire config object, because that could have undesired consequences.
+	 * 
+	 * @see https://github.com/thelounge/lounge/issues/598
+	 */
+	if (json.password !== undefined) {
```
xPaw: Make this an `else if`

MaxLeiter: bump, @RJacksonm1?

xPaw: I'd wait for #1636.

astorije:
@RJacksonm1, could you rebase this on master and fix conflicts please? Let us know if you get stuck

astorije:
Hey @RJacksonm1, would you mind rebasing on `master` and fixing the conflicts, please? 🙏 
Would love to get your PR in! :)

astorije:
Alright, I've rebased/fixed conflicts for this.

@xPaw, logs show `Password for user xxx was reset.` many times now, as user config gets reloaded over and over. Should we add something like `&& client.config.password !== userConfig.password` in that `if`? Do you think there is a better solution?

xPaw:
@astorije Well obviously you want to only print that message if the password actually changed. However "shows many times" sounds like a bug, as I mentioned in first comment, lounge writing the file will trigger autoload.

astorije:
> Well obviously you want to only print that message if the password actually changed. However "shows many times" sounds like a bug

Yes, I agree that we'll need to print the message if the password actually changed, but I left it that way for now specifically because of your second part: it does sound like a bug. Do you mind testing this locally and see if you see something obvious?

astorije:
@xPaw, this should be ready now.

astorije:
Hey @RJacksonm1, we have sticker packs for our contributors now!

Issue #1581, opened 2017-09-28
CSS: Separator between nick and message is a bit off


Jay2k1:
I just updated to da31317 and noticed that using the morning theme, said separator is not a continuous line anymore, hence messages start at different x offsets:

(increased contrast to make separator more visible)
![image](https://user-images.githubusercontent.com/2070503/30969356-688ea71c-a462-11e7-93de-6e94610132c7.png)

Apparently this happens due to different timestamp widths (if it contains many 1's, it's narrower, making the following elements (nick, message) more indented or less, especially visible here:
![image](https://user-images.githubusercontent.com/2070503/30969714-82cd1824-a463-11e7-935b-3aba01a829de.png)

This can be verified by looking at the computed element dimensions of the timestamp element in web inspector.

Tested on Chrome, Firefox and Vivaldi on Mac.

astorije:
This was introduced by #1540 and I remember now why I stopped when I played with native font stack a while ago. Lato had fixed width for numbers if we had kept it (and I'm not saying we should have, I'm very happy with native stack!).
Actually, are you using a custom font, @Jay2k1?

Apart from giving a fixed width to the time column, I'm not sure what alternatives we have. Thoughts?
By the way, totally a candidate for v2.5.0 :/

Jay2k1:
Nope, just using the morning theme, no custom font. I actually disabled my custom CSS to test before filing the bug, just to be sure.

And I agree that this should be fixed somehow before release.

astorije:
@Jay2k1, by "native font stack", we refer to using the system font of your OS, see [this](https://github.com/thelounge/lounge/pull/1540/files#diff-97db1f70168fb5f12457b238ff6052b5R18).
This also means width will slightly differ based on OS, so relying on fixed width is a bit clumsy...

@xPaw, do you have any idea?

xPaw:
@astorije set width on `.time` like it is set on `.from`.

PR #1597, opened 2017-10-01
Give chat timestamp a fix width to fix message alignment issues

astorije:
Fixes https://github.com/thelounge/lounge/issues/1581.

I gave 5px headroom to the timestamp width I get on my laptop. Tested with/without seconds, on Chrome/Safari/Firefox on Mac, Chrome on Android. Would be nice if someone on Windows and/or Linux could test, just to be sure.

Before | After
--- | ---
screen shot 2017-10-01 at 01 12 41 | screen shot 2017-10-01 at 01 12 22

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1005,6 +1005,11 @@ kbd {
 #chat .time {
 	color: #ddd;
 	padding-left: 10px;
+	width: 55px;
+}
+
+#chat.show-seconds .time {
+	width: 75px;
 }
 
 #chat .from {
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -113,6 +113,7 @@ settings.on("change", "input, select, textarea", function() {
 		chat.find(".msg > .time").each(function() {
 			$(this).text(tz($(this).parent().data("time")));
 		});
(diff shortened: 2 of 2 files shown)


dgw:
Yay for fixing this! But is there a way to fix the width relative to the font size? I can't get in and play around right now, however it seems like something around 5-6em would do unless I'm forgetting something important about CSS sizing…

astorije:
That could work. What do you think, @xPaw?

astorije:
@dgw, after chatting with @xPaw, we decided to go for pixels as a quick fix for #1581 because we're not 100% sure how that would behave under different fonts, sizes, etc.

That being said, if you or anyone wants to give it a try, we'd totally consider it in the next release while keeping the quickfix for this release.
I'm merging that now to be able to release `v2.5.0-rc.2` right now, but feel free to object and we could totally discuss it before `v2.5.0` happens if that seems unwise to you!

dgw:
Eh, I don't think there's anything wrong with pixels unless font sizes are going to change a lot. This isn't something that would be worth holding back `v2.5.0` stable for. If I (or anyone) get around to playing with it in time, fine; if not, also fine!

PR #1601, opened 2017-10-01
Remove user key from user config 

xPaw:
It is completely unnecessary and confusing. It's never used, changing it does not change the actual username.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -114,7 +114,6 @@ ClientManager.prototype.addUser = function(name, password, enableLog) {
 		}
 
 		var user = {
-			user: name,
 			password: password || "",
 			log: enableLog,
 			awayMessage: "",


astorije:
@xPaw, doesn't that correspond to the following when connecting to a network?

screen shot 2017-10-01 at 11 30 48

xPaw:
It does not.

src/clientManager.js
```diff
@@ -85,6 +85,8 @@ ClientManager.prototype.loadUser = function(name) {
 			json
 		));
 	}
+
+	return client;
```
astorije: Did you mean something like this?

```js
	let client = this.findClient(name);
	if (!client) {
		client = new Client(
			this,
			name,
 			json
		));
		this.clients.push(client);
	}
	return client;
```

Otherwise `client` is undefined and ESLint barks.

xPaw:
Gonna make a new PR.

PR #1605, opened 2017-10-01
Expand channel topic when clicked

### LINKED ISSUE #276: Topic is cut off
If a topic is too long, I can't see the entire thing on the top bar. http://screencloud.net/v/6EhZ

I would suggest making it expand when you hover over it or click on it.

PolarizedIons:
Closes #276. I liked @astorije's idea in the issue better, so I opted for doing that

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -752,6 +752,24 @@ kbd {
 	overflow: hidden;
 }
 
+#windows .header.expanded {
+	height: auto;
+	overflow-y: scroll;
+	min-height: 48px; /* Don't shrink if topic is short */
+	max-height: 96px; /* Double original height */
+	line-height: 24px; /* 48px line height looks weird when there are multiple lines */
+	padding-top: 12px; /* this is so the top text doesn't shift down when changing the line-height */
+}
+
+#windows .header.expanded button {
+	/* Fix buttons shifting down because of padding above */
+	margin-top: -6px; /* From the button's styling: margin: 6px -12px 0 0 */
+}
+
+#windows .header.expanded .rt-tooltip {
+	margin-right: -8px; /* Stop buttons from shifting left when header is expanded */
+}
+
 #windows .header .title {
 	font-size: 15px;
 }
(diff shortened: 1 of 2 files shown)


xPaw:
I tested it quickly, it does need tweaking (padding, how the text overflows, etc).

PolarizedIons:
Oh derp, I just tested it for me, but since we use native fonts now, I probably can't rely on the paddings and stuff I did. 

I need to remove those "magic numbers" from the css used to position everything right for me, but not really sure how to fix the shifting another way 

This is what it looks like for me btw: https://gfycat.com/downrightquestionableeasternglasslizard

astorije:
Hey, this is cool, great stuff!

I tested this and this is my take on it:

- Bottom padding needs to match top padding
- The situation when clicking on a link is a bit clumsy at the moment: the whole line has a 👆 cursor, but clicking on a link opens the page in a new tab + expands the topic.
  - I'd say let's not expand when clicking on a link
  - Maybe change cursor to give a visual clue that clicking on topic or URL act differently? There is no perfect cursor for this, but we could consider `zoom-in`/`zoom-out` or `ns-resize` or `row-resize`? Just an idea
- It seems a tiny bit too small on mobile, it's about 3.25 lines for me right now. Maybe give this as many lines as the message input can get?
- Because it's not clear that it can be scrolled when opened, add a fading gradient when not at the bottom in expanded mode? Something like this (but in the current context, so closer to our channel list on long channel names):
  gradient
- Maybe if there is nothing to expand, cursor should not be 👆?
- And a `...` in collapsed mode or something?

PolarizedIons:
I've been trying a bit - I don't think **I** can do this properly, so closing this. Hope someone else (or me, I don't know) can pick it up later

astorije:
Huuh, that's too bad, @PolarizedIons! Honestly, your stuff was really close to be a first pass :)
If you can just fix the bottom padding stuff, I think we should merge this and improve later.
Would you be able to give it a shot?

astorije:
Hey @PolarizedIons, would you be willing to give this another try? :) Would love to help!

PR #1606, opened 2017-10-01
Store visibility status of sidebar menus in localstorage

### LINKED ISSUE #51: Store visibility status of sidebar menus in settings object
For example:
1. Hide user list.
2. Restart the client.
3. User list is now visible, but if the visibility state was saved, it would stay hidden.

PolarizedIons:
Closes #51. The example only talked about the userlist, but keeping with the title, I did both that and the network list, split into separate commits if the latter isn't needed/wanted.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -37,7 +37,8 @@ $(function() {
 	$("#main").on("click", function(e) {
 		if ($(e.target).is(".lt")) {
 			sidebarSlide.toggle(!sidebarSlide.isOpen());
-		} else if (sidebarSlide.isOpen()) {
+		} else if (sidebarSlide.isOpen() && viewport.outerWidth() < 768) {
+			// If mobile, sidebar is open, and we click on something else - close the sidebar
 			sidebarSlide.toggle(false);
 		}
 	});
@@ -366,8 +367,6 @@ $(function() {
 		if (sidebar.find(".highlight").length === 0) {
 			utils.toggleNotificationMarkers(false);
 		}
-
-		sidebarSlide.toggle(false);
 
 		var lastActive = $("#windows > .active");
 
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -30,7 +30,7 @@ $(function() {
 
 	var windows = $("#windows");
 	var viewport = $("#viewport");
-	var sidebarSlide = slideoutMenu(viewport[0], sidebar[0]);
+	var sidebarSlide = slideoutMenu(viewport[0], sidebar[0], "networklist-expanded");
 	var contextMenuContainer = $("#context-menu-container");
 	var contextMenu = $("#context-menu");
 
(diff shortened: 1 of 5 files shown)


client/js/lounge.js
```diff
@@ -411,6 +422,18 @@ $(function() {
...
 
 		focus();
+
+		// Toggle userlist if its "openness" isn't what it should be (what is stored in localstorage)
+		const userlistShouldBeExpanded = storage.get("userlist-expanded") === "true"; // stored string -> boolean
+		if (userlistShouldBeExpanded !== utils.isUserlistVisible()) {
+			viewport.find(".chan.active .rt").click();
+		}
+
+		// Toggle networklist if its "openness" isn't what it should be (what is stored in localstorage)
```
xPaw: You don't really need to sync network list here (just on page load should be enough?)

PolarizedIons: ...right. I didn't put much thought into where I placed this one, was just thinking "put this next to this one" :+1:

client/js/utils.js
```diff
@@ -85,3 +86,7 @@ function requestIdleCallback(callback, timeout) {
 		callback();
 	}
 }
+
+function isUserlistVisible() {
+	return $(".chan.active .sidebar").first().css("right") === "0px";
```
PolarizedIons: I know this isn't the most foolproof way of doing this, and it works. Does anyone have any objections/suggestions for this?

xPaw:
@PolarizedIons Do you want to also look into making network hiding work on desktop? Because the button is currently hidden and you can't hide it.

PolarizedIons:
Alright, I first tried just doing the same as on mobile, but that moves part of the channel off of the screen, so I opted for just shrinking the channel when it's open, and expanding it when closed. The shrinking didn't look very good on mobile, so I kept that the same.

I also noticed a bug with how I was previously detecting the userlist opening (via the 'transitionend' event) when dragging: when you drag the menu open entirely (vs. just flicking it open) that event never gets triggered, so I moved that storage part into `libs/slideout.js`.

I also added a bugfix I noticed a lot while testing - the userlist opening/closing when going from "desktop" to "mobile" widths. Let me know if I should remove this.

YaManicKill:
So, I did a bit of investigating, and I've at least found how to get the 2 to show at the same time. If you move the network list toggle check that's on line 44-48, into init.js after openCorrectChannel, and you put the `$("#loading").remove();` straight after it, then the loading screen is remove basically at the same time as the 2 sidebars are expanded.

With your code:



With this change:



I've put the changes up at [this commit](https://github.com/thelounge/lounge/commit/dc6add7b8581a7f1eae03e2bf456f4c71d7dbfcc) if you want to try them out. I think this would be a good addition.

It might be nice to make them both display none, then show them if we want to, then make them display and hide the loading screen at the same time, because then you won't get the animation on the first load, and all you'll see is the loading screen disappear and the page behind appear. I think it'll be much nicer. If I have time tomorrow I might investigate that. It might need some z-index changes as well.

McInkay:
Hey, @PolarizedIons, would you have any time to address the comments we left, and rebase/fix conflicts?

astorije:
Closing this, superseded by @xPaw's #2235. Thanks for starting this, @PolarizedIons!! 🙏

PR #1608, opened 2017-10-02
Remove --progress from webpack

xPaw:
Webpack does not detect if it's run in CI and it's rather spammy to see in all the builds. It probably slows down the builds too.


astorije:
Milestone, 2.5.0 / 2.6.0 ? Seems to me that it can go whenever regardless of release.

Issue #1147, opened 2017-05-08
Autocomplete really should be mentioned in the help section


Cldfire:
I've used lounge for a year now and did not realize that 2.3.0 introduced @ nick auto-completion until @MaxLeiter told me in IRC. In fact, the only auto-completion strategy I discovered on my own without expending effort digging up the autocomplete PR was the emoji auto complete (triggered via `:`). If it weren't for my use of other non-IRC-related chat programs, I wouldn't have known about that, either. I can easily see someone who uses IRC exclusively only figuring out that such an auto-completion exists after a chance typing of `:`.

Despite the fact that a PR was just merged to remove autocomplete from the help section, I am filing this issue. It feels extremely odd to assume that users will discover these things on their own; it's certainly not hurting anyone to dedicate a small paragraph to it in the help section. If I tell a new lounge user to check out the help page, it is only logical to want them to discover auto-completion options on their own when reading it, rather than having to manually add "oh, and don't forget about autocomplete..."

YaManicKill:
I agree, actually. I have read the rational from @astorije in #1125 but I think that we should bring it back.

I accept that autocomplete should be intuitive, but I see no reason why we can't still have this documented in the help page. Plenty of people won't find ":", and it's not that uncommon for people not to automatically "@" a user. So, I say we bring it back. @astorije do you have any strong objections to this?

PolarizedIons:
I agree with this. Especially after #1356, the discovering of `:` for emoji completion is much harder.

astorije:
> @astorije do you have any strong objections to this?

Nah that's totally fair.
Put back in context, when this was removed, we didn't have color completion, and I thought the help was too blunt to be really helpful.

But I'm cool with an improved version that @dgw is working on (thanks!). It makes sense, and clearly there is a need for it considering the number of 👍 (@xPaw and I agreeing to remove it was probably experience bias 😅).

Issue #1433, opened 2017-08-18
Tab-completion of channels unavailable with auto-complete disabled


dgw:
I disabled auto-complete for the time being because I kept ending up with emoji instead of simple emoticons (see #1355), and am consistently surprised that I can tab-complete nicks, but not channels with auto-complete turned off.

Personally, I would like tab-completion to be available for both nicks and channels, even with the full auto-complete functionality disabled—hence the issue title—but at the very least I think this should be consistent (both nicks and channels completed on <kbd>Tab</kbd> even with auto-complete disabled, or neither).

The only somewhat relevant issue I found before posting this was #1316, which is about rotating the completed value on subsequent presses of <kbd>Tab</kbd>, but it's not even possible to worry about rotating the value if it won't complete in the first place. :)

xPaw:
Nicks can still be completed because it's handled by a different library (we have two at the moment, one for just the tab key and the other for @ and the UI). Ideally your issue should be handled whenever we get rid of the secondary library.

PR #1609, opened 2017-10-02
Restore autocompletion help

dgw:
See #1147. With so many thumbs-up, including two from maintainers, I thought this a reasonable step.

I took the original help section removed in #1125 and rewrote it a bit to emphasize that <kbd>Tab</kbd>-completion also exists, and to mention that menu-completion can be disabled in settings.

The commit message doesn't mention the issue number because this PR will undoubtedly undergo many force-pushes and/or rebases before being merged, and I don't want to spam it. I'll add the issue number to the commit message itself if/when this PR is approved.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -419,8 +419,14 @@
 							<div class="description">
 								<p>
 									Mark any text typed after this shortcut to be colored. After
-									hitting this shortcut, enter an integer in the
-									<code>0—15</code> range to select the desired color.
+									hitting this shortcut, enter an integer in the range
+									<code>0—15</code> to select the desired color, or use the
+									autocompletion menu to choose a color name (see below).
+								</p>
+								<p>
+									Background color can be specified by putting a comma and
+									another integer in the range <code>0—15</code> after the
+									foreground color number (autocompletion works too).
 								</p>
 								<p>
 									A color reference can be found
@@ -465,6 +471,53 @@
 									Mark all text typed after this shortcut to be reset to its
 									original formatting.
 								</p>
+							</div>
+						</div>
+
+						<h2>Autocompletion</h2>
+
+						<p>
+							To auto-complete nicknames, channels, commands, and emoji, type one of the characters below to open
+							a suggestion list. Use the <kbd>↑</kbd> and <kbd>↓</kbd> keys to highlight an item, and insert it by
+							pressing <kbd>Tab</kbd> or <kbd>Enter</kbd> (or by clicking the desired item).
+						</p>
+						<p>
+							Autocompletion can be disabled in settings.
+						</p>
+
+						<div class="help-item">
+							<div class="subject">
+								<code>@</code>
+							</div>
+							<div class="description">
+								<p>Nickname</p>
+							</div>
+						</div>
+
+						<div class="help-item">
+							<div class="subject">
+								<code>#</code>
+							</div>
+							<div class="description">
+								<p>Channel</p>
+							</div>
+						</div>
+
+						<div class="help-item">
+							<div class="subject">
+								<code>/</code>
+							</div>
+							<div class="description">
+								<p>Commands (see list of commands below)</p>
+							</div>
+						</div>
+
+						<div class="help-item">
+							<div class="subject">
+								<code>:</code>
+							</div>
+							<div class="description">
+								<p>Emoji (note: requires two search characters, to avoid conflicting with common emoticons like <code>:)</code>)</p>
 							</div>
 						</div>
 


dgw:
@MaxLeiter Oh, I forgot that change was merged. :+1: Not sure whether to put it in the blurb at the top or in the emoji section below. If that top blurb is going to talk about minimum character counts I might want to rewrite it again.

In case anyone didn't notice, this PR has two different wording options for the bit at the top in separate commits, because I couldn't decide which to use. There might be other variants too. I figure since force-pushing is probably inevitable, I might as well include alternatives and squash them all out later once a decision happens. :grin:

Anyway, will play with wording to integrate the change to emoji completion and add as another variant, probably tomorrow night.

client/index.html
```diff
@@ -468,6 +468,53 @@ <h1 class="title">Help</h1>
 							</div>
 						</div>
 
+						<h2>Autocompletion</h2>
+
+						<p>
+							Pressing <kbd>Tab</kbd> will autocomplete nicknames, channels, and <kbd>/</kbd>-commands.
+							Users with previous IRC experience will find this familiar; most clients implement tab-completion.
```
astorije: I find this a bit unclear. In fact, to select an entry in the autocompletion menu, simply hit <kbd>Tab</kbd> or <kbd>Enter</kbd>, they act the same way when it's open.
The only specific case is the nick one, because it doesn't require a prefix (like # or @), and therefore won't open a dropdown menu when pressing <kbd>Tab</kbd>.

client/index.html
```diff
@@ -468,6 +468,53 @@ <h1 class="title">Help</h1>
...
+						</p>
+						<p>
+							You can also start typing one of the following to get a menu of suggestions.
```
astorije: Right, so this is why I find this a bit confusing: to complete on <kbd>Tab</kbd> press (again, putting aside nicks which has its own logic), you need to start typing one of the mentioned characters which triggers the completion.

I get why you're saying that I think, because _technically_ pressing <kbd>Tab</kbd> has its logic/use case besides dropdown+<kbd>Enter</kbd>, but from a UX standpoint, these are the same thing.
I would structure the doc like this, let me know if it makes sense (of course, ignore my horrible English/wording):

- To open the autocomplete suggestions, type one of the following trigger characters
- To send the highlighted item, press Enter or Tab
- Note: Nicks can also be completed by hitting Tab after a sequence of characters
- Note: You can turn off the autocompletion menu in the settings (I would remove "if unwanted"). Nicks completion when Tab will remain effective.

(By the way, did you mean to add `characters` after `following`?)

client/index.html
```diff
@@ -468,6 +468,53 @@ <h1 class="title">Help</h1>
...
+						</div>
+
+						<div class="help-item">
+							<div class="subject">
+								<code>:</code>
+							</div>
+							<div class="description">
+								<p>Emoji</p>
+							</div>
+						</div>
```
astorije: Add foreground/background color completion?

astorije:
Hey @dgw, thanks for taking this. I understand why people want this information to be in the help section and I'm okay with that.

Re: the two-character required thingy, could that become cognitive overload? I'm torn on this: on one end, I understand how someone could get crazy if they expect the menu to open on first character, but on the other hand, if it feels natural enough, we might not even need to specify it. I'm fine either way, but if we add it, let's make sure it's lightweight 😅 

Also @thelounge/maintainers, should this make it into 2.5.0 or do we want that for the next release?

MaxLeiter:
I worry people will complain it doesn't work, @astorije

astorije:
That's fair, then maybe just a lightweight note in the `:` section of the trigger characters then?

client/index.html
YaManicKill: How about we just don't mention the old style tab completion? That was only left in for people who have that muscle memory, so it's not something we'd expect people to want to find.

astorije: That would be fine too, though could it be confusing when autocompletion is disabled yet still available on nicks with <kbd>Tab</kbd>? Maybe we should disable all autocomplete strategies when "Disable autocompletion" is checked...

dgw: > Note that when autocompletion is disabled, only nicks can be tab-completed.

> Maybe we should disable all autocomplete strategies when "Disable autocompletion" is checked...

Yes, I was meaning to test that because I thought it was the case. For my part, I have disabled the menu-based autocompletion strategy because I find it far too busy/intrusive. The <kbd>Tab</kbd> strategy doesn't have that issue, and I use it often for the items it supports (as mentioned, currently only nicknames).

I don't think the answer is to remove tab-completion when the menu strategy is disabled, though. For missing channel tab-completion, #1433 exists. Should I open another issue to track missing `/`-command tab-complete? Or maybe what's needed is a more general "Unify Auto-Completion Behavior" issue to replace #1433 et al, for which I'd be happy to write an initial proposal to get the discussion started on that front.

client/index.html
dgw: As mentioned in another review thread, I personally have the menu strategy disabled. I'll be playing with it shortly on a local test instance to see if I can get better wording added to this PR.

client/index.html
dgw: Yes, definitely in the next revision. :smile:

dgw:
> maybe just a lightweight note in the `:` section of the trigger characters

That was my thought. I'll try integrating that in the next revision.

dgw:
I'll have the next revision pushed up in a moment.

Adding the color completion turned out to be trickier than I expected. I wound up revising the Keyboard Shortcuts section a bit to reference that autocompletion is available for colors. Honestly, I'm not sure that I like including <kbd>Ctrl</kbd>+<kbd>K</kbd> in the Autocompletion section. It's easy to snip out if the documentation in the Keyboard Shortcuts section is sufficient.

I'm quite certain that the wording in general is still not perfect, but thanks to @astorije's feedback I think it's better now.

client/index.html
YaManicKill: I don't think it'll be confusing because it's generally not something people will find unless they expect it. I think the way it currently works is good, and I think just ignoring it from the documentation is best.

YaManicKill:
> Also @thelounge/maintainers, should this make it into 2.5.0 or do we want that for the next release?

I'd say leave it till 2.6.0, unless we are happy with it before 2.5.0 is released, but that'll happen pretty soon. 

I think my issue with adding things to a release after RC is that RC is "release candidate". That shouldn't be "What we might release if we don't think of any new features we want in". It means "We want to release this, but we need to check there are no showstopping bugs before we do". Bugs are the only things that should get merged into a release after RC.

The issue is that our releases are so big they take so long to get out, so we want ot shove as much into them as possible.

I think we should release on a regular basis, and if something doesn't get in, fine, it'll get in the next one. We keep pushing them back and pushing them back because we want more stuff in them because releases take ages. It's a vicious cycle.

astorije:
Totally fine with that @YaManicKill. Since we fail at releasing often, I'll make sure from now on to release `pre.N` versions often, and `rc.N` versions when we are close to release the stable version. /PR-hijack-off :)

dgw:
This isn't a release blocker, but hopefully I can get it up to snuff in time for 2.5.0. It'd be nice to have.

Are there any comments I'm missing? I don't think there was any feedback given on the latest revision, so I'm directionless for the moment, haha.

client/index.html
```diff
@@ -468,6 +474,64 @@ <h1 class="title">Help</h1>
...
+								<code>:</code>
+							</div>
+							<div class="description">
+								<p>Emoji (note: requires two search characters, to avoid conflicting with common emoticons like <code>:)</code>)</p>
+							</div>
+						</div>
+
+						<div class="help-item">
+							<div class="subject">
+								<kbd>Ctrl</kbd> + <kbd>K</kbd>
```
xPaw: Make this `<kbd class="key-all">Ctrl</kbd><kbd class="key-apple">⌘</kbd>`

dgw: Fixup pushed.

client/index.html
```diff
@@ -468,6 +474,64 @@ <h1 class="title">Help</h1>
 							</div>
 						</div>
 
+						<h2>Autocompletion</h2>
+
+						<p>
+							The Lounge offers automatic completion of nicknames, channels, <kbd>/</kbd>-commands, emoji, and text colors.
```
MaxLeiter: I don't think `<kdb>-</kbd> commands` is necessary; just `commands` is sufficient

dgw:
Pushed a fixup removing `<kbd>/</kbd>-` per @MaxLeiter. Depending on consensus, I can either include it or drop it during final rebase (hopefully soon).

MaxLeiter:
👍

dgw:
Interpreted the long silence as "There's nothing left to change." Rebased, squashed, etc. onto current master for hopeful inclusion in 2.6.0 (which, looking at the milestone, is _so close_ :soon:).

astorije:
Hey @dgw, sorry, been reeaaaally busy IRL the last few weeks. Taking a break to look at your changes now :)

Good stuff overall, but we could be just as informative while being shorter, therefore easier to follow.

Would you mind changes with the following please?

```diff
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: client/index.html
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ index.html:480 @
                        <h2>Autocompletion</h2>

                        <p>
-                           The Lounge offers automatic completion of nicknames, channels, commands, emoji, and text colors.
-                           Type one of the characters below to get a menu of suggestions, and filter the list by typing part of the channel
-                           name, nickname, emoji, command, or color you want. Use the up and down arrow keys to highlight a menu item.
-                           Insert the highlighted suggestion by pressing <kbd>Tab</kbd> or <kbd>Enter</kbd>, or click the desired item.
+                           To auto-complete nicknames, channels, commands, and emoji, type
+                           one of the characters below to open a suggestion list. Choose with
+                           <kbd>↑</kbd> and <kbd>↓</kbd>, and select by hitting
+                           <kbd>Tab</kbd> or <kbd>Enter</kbd>, or by clicking the desired
+                           item.
                        </p>
                        <p>
-                           Autocompletion can be disabled in settings. (Typing part of a nickname and pressing <kbd>Tab</kbd> will always
-                           complete it, regardless of setting state.)
+                           Autocompletion can be disabled in settings.
                        </p>

                        <div class="help-item">
@ index.html:513 @
                                <code>/</code>
                            </div>
                            <div class="description">
-                               <p>Commands (see list of commands below)</p>
+                               <p>Commands (See list of commands below)</p>
                            </div>
                        </div>

@ index.html:522 @
                                <code>:</code>
                            </div>
                            <div class="description">
-                               <p>Emoji (note: requires two search characters, to avoid conflicting with common emoticons like <code>:)</code>)</p>
-                           </div>
-                       </div>
-
-                       <div class="help-item">
-                           <div class="subject">
-                               <kbd class="key-all">Ctrl</kbd><kbd class="key-apple">⌘</kbd> + <kbd>K</kbd>
-                           </div>
-                           <div class="description">
-                               <p>Text color. Type <kbd>,</kbd> after selecting foreground color to add an optional background color.</p>
+                               <p>
+                                   Emoji (Requires two search characters, to avoid conflicting
+                                   with common emoticons like <code>:)</code>)
+                               </p>
                            </div>
                        </div>
```

This gives:

Before | After
--- | ---
screen shot 2017-11-05 at 14 13 25 | screen shot 2017-11-05 at 14 24 48

I'll be 👍 after these changes :)

Note that on top of reducing the size, I removed the part about colors. This is because it's an exact duplicate of the "Keyboard Shortcuts" section above. I realize it's debatable where to put it and how to present it, but I think having both repeated slightly differently is a bit confusing.
I don't have a strong opinion about this, we can improve/reword in a latter PR.

Thanks!!

dgw:
@astorije Fair point about the duplication of color documentation. I was not entirely sold on including it in the autocompletion section myself.

Is the new wording still clear enough? Especially the bit about inserting the selected item, which I tried to improve by using a strategic pair of parentheses instead of another comma…

dgw:
> What do you think?

I think that was missed in my amended commit. :) Hazard of late-night vimming/GitHubbing…

astorije:
Merging. @MaxLeiter's review at the top of the page is still ✖️, but he did 👍 the PR in a comment above.

PR #1615, opened 2017-10-04
Mark channels as read when receiving self-messages

### LINKED ISSUE #911: Mark buffer as read when receiving self-/echo-message
The Lounge makes a great "always-on" client for ZNC that I can load up if I'm away from a computer, without having to keep a client running on my phone all the time. But the experience is a little frustrating when it comes to unread statuses for channels and PMs. Things I've sent messages to from another client still appear as unread in The Lounge, and I have to remember whether I've actually read them.

Since #270 added support for the `echo-message` extension of IRCv3 (and ZNC's version of it, `znc.in/self-message`), I wonder if it wouldn't be possible for The Lounge to mark buffers as read on incoming echoed messages.

Granted, I haven't yet seen any other client do this…but I'd argue that more clients should, as it's much better UX for bouncer users—especially given The Lounge's explicit design goal of being "always on", unlike a traditional desktop IRC client.

dgw:
Resets highlight counter and last unread message ID on server, and clears the badge on the active client when the message is rendered.

I've been poking at it for a good half-hour, so hopefully I haven't missed any edge cases. (I also hope I'm understanding the unread-message tracking correctly and not failing to reset some value I missed.)

Resolves #911 (issue mention omitted from commit description for the same reason as my other PR: avoiding spam from force-pushes).

Edit: Glad to see the tests pass on CI, as I was pretty sure they would. Had tests fail locally but seemed to be an environment problem (maybe outdated eslint or something).

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -72,7 +72,11 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 		}
 	}
 
-	if (!msg.self && !isOpen) {
+	if (msg.self) {
+		// reset counters/markers when receiving self-/echo-message
+		this.firstUnread = 0;
+		this.highlight = 0;
+	} else if (!isOpen) {
 		if (!this.firstUnread) {
 			this.firstUnread = msg.id;
 		}
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -61,6 +61,11 @@ function processReceivedMessage(data) {
 			.appendTo(container);
 	}
 
+	// Clear unread/highlight counter if self-message
+	if (data.msg.self) {
+		sidebar.find("[data-target='" + target + "'] .badge").removeClass("highlight").empty();
+	}
+
 	// Message arrived in a non active channel, trim it to 100 messages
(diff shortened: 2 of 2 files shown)


src/models/chan.js
```diff
@@ -80,6 +80,10 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 		if (msg.highlight) {
 			this.highlight++;
 		}
+	} else if (msg.self && this.firstUnread > 0) {
```
xPaw: I don't think you need the `this.firstUnread > 0`. You could rewrite this if as `if(msg.self)` and move the if above to just `else if (!isOpen)`

dgw: Updated. Theory was that it'd be more performant to only update the variable if it needs to be updated, but that's probably me being silly. JS vars probably don't even work like that.

dgw:
I'm not sure if this is ready to be rebased/squashed, but let me know.

xPaw:
@dgw Can you squash?

dgw:
Done & pushed. :checkered_flag:

PR #1619, opened 2017-10-06
Allow custom css from plugins

YaManicKill:
Example:

On thelounge export, put 

```
run: function (thelounge) {
    thelounge.addStylesheet("file.css");
}
```

`file.css` is relative to the root of the package. (so `file.css` would be in `packagefolder/`)

- [x] Allow other files other than css files


src/server.js
```diff
@@ -64,6 +67,12 @@ module.exports = function() {
 		return res.sendFile(theme);
 	});
 
+	app.get("/packages/:package/:style.css", (req, res) => {
```
xPaw: I don't really see a reason to hardcode this for `.css`, just let plugins specify which files they want to serve.

YaManicKill: Hmmm alright, that makes it a little more complicated, but I'll look into it.

YaManicKill: Alright, changing this to be for any type of file was quite easy, I've got that working. However, combining it with the 1 above (along with keeping the backwards compatibility of the themes) is quite difficult.

xPaw: What's the trouble?

If you mean it's difficult for default themes because they don't have `package.json`, I wanted to eventually create them.

YaManicKill: Alright, well I've combined this with the themes if we can assume that all of the old ways of defining themes was just getting it from "themes/something.css" if we can assume that, I can keep the backwards compatibility. I'll upload this evening hopefully.

YaManicKill: Actually no scratch all of that, I missed something. Combining the 2 is quite hard when we have to support the local themes as well. For just now, I'll keep it separate. Until we move the themes into packages.

xPaw: Will this work for packaged themes?

YaManicKill: Yeah, I can combine them for packaged themes easily, but not for the local ones.

src/server.js
```diff
@@ -477,6 +488,14 @@ function getClientConfiguration() {
 	return config;
 }
 
+function getServerConfiguration() {
+	const config = Helper.config;
+
+	config.customCss = packages.getCustomCss();
```
xPaw: You're assigning this to `Helper.config` object.

YaManicKill: Yuuuup, good point. I shall clone it.

YaManicKill:
Anyway, this should be good now. It would be good to deprecate (and remove) local themes at some point, but that's probably not anytime soon (as we don't even have the themes in packages yet).

The packages will need some code abstracted out a bit when I add more, but I'll do that when we add more stuff. I think this is a good start for package support and we can build on more features when we add them.

src/server.js
YaManicKill: > What's the trouble?

YaManicKill: > What's the trouble?

Oh sorry, I missed this before. It's a bit complicated to explain, but here we go. 

It's a combination of the default theme config, allowing the old way of defining things, and combining the `/themes/theme.css` and `/packages/package/file` endpoints. It obviously makes sense to remove `/themes/theme.css` and just have `/packages/package/file` as the themes are just files.

When you are doing this, you have to have a package name, which we would either do as the name of the theme (possibility) or something like `localTheme` or something horrible. Not so difficult just now, I guess. You just define the local theme path as `/packages/localTheme/default.css` and then we switch on the package, if it's called `localTheme` then we go get the theme file from the themes folder.

So, the default theme can be either `name` or `path/to.css`. Now, if it's a name, we have to know whether or not it's a local theme, but that's easy because we have a list of themes and can get the appropriate filepath for the theme (we built up a `localTheme` object for each local theme at the start). However, the complication comes when we have a path.

It's simple if we can define them all as `themes/default.css` or similar, because then we can pull the name out of there. But we can't guarantee that, as we support any path to any css file (currently). So we could have something like `themes/folder/test.css`, or even something like `otherfolderhere/something.css`. 

The best way to do this is probably to remove the support for specifying a css file, only support local files in the themes folder, and then we can combine the 2.

Does that make sense? Have I missed something obvious?

YaManicKill: And then on top of that, deprecating local themes, so we can remove them in the future and only allow the default theme built into the lounge. We could even automatically install the other themes, and as we will have removed path support for css files, then the name would map to the package rather than the local theme.

So I would suggested:

1. Ship this PR in 2.7.0 with the 2 backwards compatible ways of dealing with things, but deprecated support for local themes (I can add that into the PR before it gets merged, just letting the user know if they have a local theme that isn't 1 of the themes we ship with and telling them to get it packaged)
2. Package up the 4 themes (we could just package up the 3 non-default themes, but that would still leave some hardcoded stuff for the default theme)
3. Have the lounge automatically install the default theme and the 3 non-default themes (we could just ship with the 1 theme, but it would be nice to the users if we didn't)
4. In 3.0 Remove the deprecated support for css file paths
5. After 3.0, we can then remove the support for local themes (if we were automatically installing them, then it wouldn't be a breaking change as long as the css file paths support has already been removed)

After all those things have been done, we can very easily combine the 2, because all the themes are packaged up, and we don't have to do any weird stuff.

Thoughts?

src/plugins/packages.js
```diff
@@ -0,0 +1,61 @@
+"use strict";
+
+const fs = require("fs");
+const colors = require("colors/safe");
+const Helper = require("../helper");
+const themes = require("./themes");
+
+const customCss = [];
```
xPaw: Can we simply call this `stylesheets` (and all the other occurances of `customCss`)

src/plugins/packages.js
```diff
@@ -0,0 +1,61 @@
...
+
+function getModuleInfo(packageName) {
+	let module;
+	try {
+		module = require(Helper.getPackageModulePath(packageName));
+	} catch (e) {
+		log.warn(`Specified package ${colors.yellow(packageName)} is not installed in packages directory`);
+		return;
+	}
+	if (!module.thelounge && !module.lounge) {
```
xPaw: You can remove `module.lounge`, it's not there on master.

YaManicKill: We would have deprecated this first because we publicised it as this. I'll add it back in here with deprecation warning and we can remove it in 3.0.

src/plugins/packages.js
YaManicKill: Yeah, sir

src/plugins/packages.js
astorije: Nah, you can just remove it: we never made a release without it, it was in `master` for like 24 hours. It's silly to add something directly deprecated :/

YaManicKill: Alrighty

YaManicKill: Done

src/plugins/packages.js
YaManicKill: Done

YaManicKill:
Alrighty, think I've cleared up the last bits of this, it should be good to go now.

src/plugins/packages.js
```diff
@@ -0,0 +1,64 @@
...
+function getStylesheets() {
+	return stylesheets;
+}
+
+function loadPackages() {
+	fs.readdir(Helper.getPackagesPath(), (err, packages) => {
+		if (err) {
+			return;
+		}
+		packages
```
xPaw: unnecessary newline

src/plugins/themes.js
```diff
@@ -44,7 +41,6 @@ function getFilename(module) {
 		return themes.get(module).filename;
 	}
 }
-
```
xPaw: keep the new line

src/plugins/packages.js
xPaw: Also store the a map of loaded packages

src/server.js
```diff
@@ -51,6 +54,13 @@ module.exports = function() {
 		return res.sendFile(theme);
 	});
 
+	app.get("/packages/:package/:filename", (req, res) => {
+		const packageName = req.params.package;
```
xPaw: With stored map of loaded packages, verify requested package against that map here.

src/server.js
```diff
@@ -51,6 +54,13 @@ module.exports = function() {
...
+		const fileName = req.params.filename;
+		const packagePath = Helper.getPackageModulePath(packageName);
+		return res.sendFile(path.join(packagePath, fileName));
```
xPaw: I think we should also verify that `fileName` is in fact exposed by the package, e.g. the `stylesheets` array.

src/server.js
```diff
@@ -469,6 +479,7 @@ function getClientConfiguration() {
 	config.version = pkg.version;
 	config.gitCommit = Helper.getGitCommit();
 	config.themes = themes.getAll();
+	config.stylesheets = packages.getStylesheets();
```
xPaw: This is not needed.

src/plugins/packages.js
YaManicKill: For what purpose? What's the reason for storing them until we use it for something?

xPaw: Verifying that requested package exists, see comment down below.

YaManicKill: Ah right, I just saw that. Makes sense.

> Also will need it to grab manifest.json as we discussed.

Sure, but this isn't in this PR, so it wouldn't be a reason for storing them yet. But yeah, I'll do it for verifying.

src/server.js
YaManicKill: Oh yeah, added this in when testing how the new UI worked. Forgot to remove it.

src/server.js
YaManicKill: Yeah, I guess there's a minor security concern if we don't check that. Added.

src/server.js
YaManicKill: I'm just checking that it exists

src/plugins/themes.js
YaManicKill: Yeah, this was an accidental deletion.

src/plugins/packages.js
```diff
@@ -0,0 +1,70 @@
...
+			return;
+		}
+		packages.forEach((packageName) => {
+			const packageFile = getModuleInfo(packageName);
+			if (!packageFile) {
+				return;
+			}
+			packageMap.set(packageName, packageFile);
+			if (packageFile.type === "theme") {
+				themes.addtheme(packageName, packageFile);
```
xPaw: `addTheme`?

YaManicKill: `addtheme: addTheme,`

Oh boy, sometimes I wonder why I bother :-P

src/plugins/themes.js
```diff
@@ -2,38 +2,35 @@
 
 const fs = require("fs");
 const Helper = require("../helper");
-const colors = require("colors/safe");
 const path = require("path");
 const _ = require("lodash");
 const themes = new Map();
 
 module.exports = {
+	addtheme: addTheme,
```
astorije: If you call the exported value `addTheme`, then with ES6 this whole thing could be: 

```js
{
	addTheme
  	getAll,
	getFilename,
	loadLocalThemes,
}
```

:)

YaManicKill: Yeah, I knew about this but I had assumed it didn't work in node 4. Turns out it does, TIL.

YaManicKill:
I hope to finish this off tonight, is it just the addTheme and the exports to sort out?

YaManicKill:
I keep running out of time to do this. It's so close I want it in 2.7.0, so I will get it updated this week.

YaManicKill:
Alright, updated.

YaManicKill:
> My only concern is that we're still only allowing stylesheets to be served from the server (it checks against stylesheets map).

Yeah, I think this is something that should be added, but I'd like to add that in a second PR. There are a lot of plugin PRs that I have in mind but all of them rely on this because this has a bunch of the underlying plugin functionality.

astorije:
@YaManicKill, could you add some example usage of this API in PR description? It would help me review, and as I'm currently writing a very short doc on custom CSS, I will also mention this as well.

Thanks!

YaManicKill:
@astorije Done, hope that's enough.

astorije:
Alright, I'm not gonna bikeshed this, promise :)

> - [x] Allow other files other than css files

Shouldn't this actually _only_ allow CSS files? What use cases are you thinking about?
Trying to think of the implications of accidentally or intentionally (trying to exploit a vulnerability or something) specifying say an image, script, etc. but I don't think there is any apart from "it doesn't work but doesn't complain either".

---

At the moment, this is server-side only, correct? As in, if admin installs this or that custom CSS package, user _has_ to use it? Not requesting in this PR of course, just making sure we're on the same page.

---

In terms of implementation, we can improve that over time, but I think we should nail the interface as part of this PR, so we don't immediately break APIs later (or break them mildly :D). So here are my 2 thoughts:

- `run` should be called something like `onServerStart` (or better if you have). `run` feels like a placeholder, though I'm not making any assumptions, maybe you had a perfectly valid reason to call it like that and I simply missed it. My way of looking at it is that, based on how you architected this, I see a the beginning of a package lifecycle: this API runs at startup, but we could have some that run at server shutdown (plugins that create and delete storage or whatnot), or packages that run before/after authentication, after channels are displayed, etc.

- The way I envision our package interfaces is through a bunch of APIs, that we'll create or augment over time, a bit like https://atom.io/docs. So right now you're working on the Stylesheets API (that our theme management system could leverage at some point, for better dogfooding), you mentioned context menu abilities, so that could be a ContextMenu API, we could have Settings API, etc.
  In terms of documentation, it's great because it's very easy to process, based on what you want to do as part of your package. And a package can use multiple APIs at the same time.
  So instead of `(thelounge) => { thelounge.addStylesheet("file.css"); }`, I would do something like `(apis) => { apis.Stylesheets.addFile("file.css"); }` or similar. Once #1727 makes it, it will be even nicer, because you can then declare which APIs you want to use directly: `(Stylesheets, [...]) => { Stylesheets.addFile("file.css"); }`. Ideally, it would be even nicer if we could do `const Stylesheets = require("thelounge").Stylesheets;` but I'm not entirely sure yet how to do that. I just realized this is exactly what Atom does, not sure how.
  Note that I names the API `Stylesheets`, but this is very arbitrary (and it's 1am+ 🙂), so that can totally be tweaked.

YaManicKill:
> Shouldn't this actually only allow CSS files? What use cases are you thinking about?

So, there's 2 things to talk about this:

1. It doesn't actually currently allow any file other than the css files, because it checks that the user has specified them. I chatted with @xPaw about this, and we are probably going to add a way for the plugin to add a "web accessible file" which can be referenced by the client-side portion of their plugin. Then we can check against that list as well and let those files through. But basically, the url previously hard-coded the css file which it doesn't just now, but it still only allows css files to be sent to the client
2. The use cases are fonts that they might want to bundle, or, say, background image. That sort of thing.

> At the moment, this is server-side only, correct? As in, if admin installs this or that custom CSS package, user has to use it?

Yeah, at some point I want to add settings to turn them off. But I'm waiting on #1851 and a couple of others before we do that. This will cover like 90% of the functionality that people will want from custom css, and that'll be good enough for this PR, I think.

> run should be called something like onServerStart

I hadn't thought of any use cases of having functions run at other times, but I guess this makes sense. Will do this.

> (apis) => { apis.Stylesheets.addFile("file.css"); }

Sure, I guess that could make sense. I'm not sure how that's any cleaner than having just "apis.addStylesheet" but I don't care enough to argue :-P

YaManicKill:
Alrighty, I've refactored things slightly, changed `run` to `onServerStart` and I've added the sub-api as suggested, however, instead of "Stylesheets" I've done "Themes" because I think it makes sense to combine themes and custom css.

So now we have this passed to the :

```
apis: {
  Themes: {
    addTheme: function,
    addStylesheet: function,
  }
}
```

I'm not 100% sold on `addStylesheet` as the function for a non-theme css file, but I also don't like "custom" nor does "addFile" make sense, as we want to distinguish between themes and other css. Themes should be considered something that you only have 1 of, whereas stylesheets you can have many.

I was thinking maybe "override" might be a better term, but ... nah, I don't like it.

I think we are now at the point of just arguing over naming, so let's get that done :-P

astorije:
> It doesn't actually [...] but it still only allows css files to be sent to the client

Makes sense, thanks for the explanation 👍 

> The use cases are fonts that they might want to bundle, or, say, background image. That sort of thing.

Right, agreed, but I thought you meant "allow other types than CSS for `addStylesheet`" and was confused. Absolutely agree serving fonts and images and such will be something to tackle.

> Yeah, at some point I want to add settings to turn them off. But I'm waiting on #1851 and a couple of others before we do that. This will cover like 90% of the functionality that people will want from custom css, and that'll be good enough for this PR, I think.

Yes, that makes sense too, just wanted to make sure we were talking about the same thing. I'll make sure that when announcing and/or documenting to specify that, admins be careful, if you install a custom stylesheet package, it will be enabled by default for all your users and there is nothing you can do about this at the moment except uninstalling it (which we don't have right now :D)

> I'm not sure how that's any cleaner than having just "apis.addStylesheet" but I don't care enough to argue :-P

The idea that we'll soon have all sorts of APIs and capabilities for our packages, and making them compartmentalized is nice for mental model, documentation, etc. If you write a package to add a dumb entry to the context menu, you only need to read that part of the documentation and not have to read a whole page with dozens of functions, etc.
I know it's early, but I'd rather avoid breaking our APIs if we can (though we'll probably have to at some point), just future-proofing a bit is all.

> instead of "Stylesheets" I've done "Themes" because I think it makes sense to combine themes and custom css.

I'm not sure about that, because if you write a theme, or if you browse themes, it's not about packages, and the other way around. I definitely think that in the future our theme management can use this very API to register stylesheets, fonts, background images, etc., but I see themes as a higher-level mechanism than the stylesheets part, which is lower-level.
For example, it's likely that our codebase itself consumes this API at some point, while the theme part would be reserved to actual themes, and therefore the codebase would not make direct use of it.

But I guess I'm diverging here, and we can go with that for now. I do think however that we'll have to change this soon, which is why I went for said suggestion directly.

> I'm not 100% sold on addStylesheet as the function for a non-theme css file, but I also don't like "custom" nor does "addFile" make sense, as we want to distinguish between themes and other css. Themes should be considered something that you only have 1 of, whereas stylesheets you can have many.

Yes, I think I exactly see what you mean.

In fact, I don't think there should be a package API for themes: if you develop a theme (i.e. you have the `thelounge-theme` keyword), what you export is the theme, whether it's a filename string or something else. And if you want to take the easy route (80% cases), just use your package.json to register your theme CSS.

What I would like to avoid is make our package API a gray area that makes it too easy to publish themes hidden in packages, or packages hidden in themes, if that makes sense.

But again, we can easily say this API is in beta for now, and refine over time.

Ah, also, `register` could work instead of `add`. Doesn't matter, it's our first API, we'll get better as we go :)

astorije:
@xPaw, what do you think of the latest changes? Still 👍 ?

astorije:
> there is nothing you can do about this at the moment except uninstalling it (which we don't have right now :D)

#1938 :)

YaManicKill:
> The idea that we'll soon have all sorts of APIs and capabilities for our packages, and making them compartmentalized is nice for mental model, documentation, etc.

Yeah I guess

> In fact, I don't think there should be a package API for themes

Disagree, I can definitely see people wanting to build a package that includes a theme and some other things. I think we absolutely should allow a package to register a theme. I don't think there's any disadvantage to that.

However, I'm happy to either rename `Themes` to something else (I dunno, like `Visuals` or something) or separate `Themes` from `Styles` but I don't see any reason to delete the theme addition from the API.

> I know it's early, but I'd rather avoid breaking our APIs if we can

> But again, we can easily say this API is in beta for now, and refine over time.

Oh boy, that was a *wild ride*, @astorije 

Anway, jokes aside. I legitimately don't care about the naming.

xPaw:
I looked through the code, and I'm gonna agree with @astorije that `addTheme` is not a good idea to add (not in the current state anyway). The way that function currently works does not make it nice for 

I suggest don't expose that just yet, to allow us to iterate more that in a separate PR (I see things that can be refactored further in theme management).

Good idea to separate into sub components by the way, we don't want to have a ton of functions under a global namespace.

astorije:
> Disagree, I can definitely see people wanting to build a package that includes a theme and some other things. I think we absolutely should allow a package to register a theme. I don't think there's any disadvantage to that.

I'm strongly 👎 on that. Main disadvantage is confusion and obscurity for the user: when installing a package, they should only get a package; when installing a theme, they should only install a theme. Even though the underlying implementation is consuming a shared API, it's clearer what happens.
I'm planning on doing `/packages` and `/themes` pages for re-docs àla Atom docs and that would fit really well. Since packages and themes are 2 different types in our keywords (and probably the only types we'll have), I think it makes sense to separate them semantically: themes will only change the visible shape of TL, while packages may affect features.

Also, like @xPaw said, maybe we can revise this once things get better, but now it's not ideal.

> Oh boy, that was a _wild ride_, @astorije

I know, but my rationale goes as follows: I did think (still do) that I'd rather avoid breaking our APIs if we can, _assuming we 100% agree at this stage_. Since we don't (we might have diverging visions for packages for now, and that's fine considering how early in the process this is, we're not yet all clear on what we all want and think), we can either hold on this and iterate until we do, or we can merge with the official statement that it will probably change.
You don't realize it, but I was trying to be nice to you and make a compromise to help you merge this ASAP instead of arguing and iterating forever pre-merge 😁 

> Good idea to separate into sub components by the way, we don't want to have a ton of functions under a global namespace.

Exactly. Since our internal codebase will over time evolve to make use of the same APIs packages have at their disposal, it's nice to think of those APIs as core APIs and not a single packages file somewhere :)

BTW, I'll come up with a doc structure and maybe fake contents to show what I'm planning on doing for package doc. I didn't think of that before, but that way we can have the doc structure prepared for future external APIs. Then with some kind of JSDoc or else, we can even have this filled for us 👌.

astorije:
I'm sorry to be say that to you, but I think we should not ship this as part of v2.7 as they're are clearly changes we should make, even to get this early-stage ready.
However, I'm in favor or merging it soon after v2.7 is released, and we can iterate on that until v3 release.

Thoughts?

YaManicKill:
I think you are artificially creating a difference between themes and packages where there's no need to create one, but I don't care enough so I'll remove it. 

I didn't mean anything negative by my other comment, it was a joke. 

Anyway, as I've said on the IRC, I don't care whether this gets into 2.7, I just want it merged in soon because all the other plugin prs rely on it. 

Having said that, I'm not sure of your reasoning for keeping it from 2.7. The only changes I can see you wanting are removing the themes function and renaming the namespace. Am I missing something? That doesn't seem like much. 

Obviously, there is a lot to be improved elsewhere, but this in itself is fine, imo.

christer88:
I suggest YaManicKill does the changes suggested by astorije and then merge this one, together with the other PR which is left for 2.7. Let's get 2.7 out. We can make changes to this in 3.0 if needed be, and continue with the discussion regarding what packages may contain.

YaManicKill:
I have removed the addTheme, and renamed Themes to Stylesheets, as originally suggested by @astorije 

If you still don't want it in 2.7.0, then that's fine, but it's there.

astorije:
> I didn't mean anything negative by my other comment, it was a joke.

I know I know, I was not offended or something 😅  You had a point though, and I just wanted to clarify my thought process :)

PR #1620, opened 2017-10-06
Add tests for invalid urls

xPaw:


test/client/js/libs/handlebars/ircmessageparser/findLinks.js
```diff
@@ -103,4 +103,49 @@ describe("findLinks", () => {
...
+			link: "http://www.example.com"
+		}, {
+			end: 42,
+			start: 16,
+			link: "ssh://-oProxyCommand=whois"
+		}];
+
+		const actual = findLinks(input);
+
+		expect(actual).to.deep.equal(expected);
```
astorije: This should be its own "should add http protocol to urls starting with www" or something, no?

xPaw: Example doesn't matter here, the test is for the second URL.

astorije: Mmh ok

PR #1623, opened 2017-10-10
Make sure channel list close button is really absent when channel is not selected

astorije:
This removes the ghost close button when a channel with long name is not selected, or when there is a badge for it.

The first part of this bug is already present in v2.4.0 (it is however more noticeable now that font is bigger), but not the second, most likely introduced by https://github.com/thelounge/lounge/pull/1561/commits/4dc3769b18c509ec631cfc312250773ae7570e40.

&nbsp; | When channel is selected (no difference) | When channel is not selected | When badge is shown
--- | --- | --- | ---
Before | screen shot 2017-10-10 at 01 43 43 | screen shot 2017-10-10 at 01 43 53 | screen shot 2017-10-10 at 01 45 12
After | screen shot 2017-10-10 at 01 44 04 | screen shot 2017-10-10 at 01 44 13 | screen shot 2017-10-10 at 01 44 43

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -556,7 +556,7 @@ kbd {
 	font-size: 10px;
 	margin-top: 1px;
 	padding: 3px 6px;
-	transition: opacity 0.2s, background-color 0.2s, color 0.2s;
+	transition: background-color 0.2s, color 0.2s;
 }
 
 #sidebar .badge:empty {
@@ -568,17 +568,11 @@ kbd {
 	color: #49505a;
 }
 
-#sidebar .chan.active .badge,
-#sidebar .badge:empty {
-	opacity: 0;
-}
-
 #sidebar .close {
 	border-radius: 3px;
 	width: 18px;
 	height: 18px;
-	visibility: hidden;
-	opacity: 0;
+	display: none;
 	transition: opacity 0.2s, background-color 0.2s;
 }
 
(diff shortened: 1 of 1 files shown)


client/css/style.css
```diff
@@ -556,7 +556,7 @@ kbd {
 	font-size: 10px;
 	margin-top: 1px;
 	padding: 3px 6px;
-	transition: opacity 0.2s, background-color 0.2s, color 0.2s;
+	transition: background-color 0.2s, color 0.2s;
```
astorije: There is no opacity change for badges, they are just displayed or not.

client/css/style.css
```diff
@@ -568,17 +568,11 @@ kbd {
 	color: #49505a;
 }
 
-#sidebar .chan.active .badge,
-#sidebar .badge:empty {
-	opacity: 0;
-}
```
astorije: `#sidebar .badge:empty` is already `display: none` (see L562 above), and `#sidebar .chan.active .badge` is a moot point, because [badge content is being emptied on opening](https://github.com/thelounge/lounge/blob/09878a5568e909d0bbcec3a7c6080251424be386/client/js/socket-events/open.js#L10).

client/css/style.css
```diff
@@ -568,17 +568,11 @@ kbd {
...
-
 #sidebar .close {
 	border-radius: 3px;
 	width: 18px;
 	height: 18px;
-	visibility: hidden;
-	opacity: 0;
+	display: none;
 	transition: opacity 0.2s, background-color 0.2s;
```
astorije: No opacity/visibility transition on click anymore (switching from opacity/visibility to display in order to hide/show), but there is still an opacity/bg-color transition on hover as opposed to the badge.

Issue #1534, opened 2017-09-16
Tab-complete common network service names


dgw:
I was thinking of other little quality-of-life things that my favorite desktop clients do, and one of them is tab-completing the names of network services (NickServ, ChanServ, MemoServ, etc.) along with nicks from the channel. (Here's an [example](https://github.com/Codeux-Software/Textual/blob/2c02fa4dcd5d87d1c4e828c5a9bdab8502e97a1b/Classes/Library/TLONicknameCompletionStatus.m#L278-L284) from Textual.)

This could be my next little project for The Lounge. I'd have service names always be at the end of the autocomplete sort order, so nicks in the channel are always completed first. Thoughts from the maintainers? Yay/nay?

xPaw:
I don't think it's a good idea to auto complete *nicknames* of these services. However adding shortcut commands like `/ns` is a better idea.

dgw:
Lite version of this PR'd with the shortcut commands. Still happy to work on the nickname part unless it's been vetoed outright.

LFlare:
@xPaw, close this?

PR #1627, opened 2017-10-12
Add service aliases to command list for completion

dgw:
Includes known aliases for AdminServ (`/as`), BotServ (`/bs`), ChanServ (`/cs`), HelpServ (`/hs`), HostServ (`/hs`; `/ho` on some IRCds), MemoServ (`/ms`), NickServ (`/ns`), OperServ (`/os`), and RootServ (`/rs`).

Not all IRCds will have all of these services available, but it's not trivial to filter out the ones that don't exist on a particular network.

This is the #1534-lite that @xPaw suggested. I still believe it would be useful to have service nickname completion available, but these shortcuts should be added anyway whether that happens or not. :grinning:

Feel free to point out if I've forgotten a useful alias. The most comprehensive reference I found was for unrealircd, and though most networks I've used tend to agree on the aliases I still might have left out a couple that are unique to another daemon.


xPaw:
Could also throw `/znc` in there.

dgw:
Given this some thought and I'm not sure `/znc` should be included. It _can_, yes, but I've not seen other clients complete it. It's a user-side thing, not network-side like the rest of the aliases this adds.

PR #1631, opened 2017-10-14
Don't send messages on init

xPaw:
I don't know what problems can arise from doing this, but this fixes a major problem where loading the client may take a lot of time due to huge amount of channels.

This might screw `lastMessageId` on reconnects.

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -139,7 +139,7 @@ Chan.prototype.removeUser = function(user) {
 Chan.prototype.toJSON = function() {
 	var clone = _.clone(this);
 	clone.users = []; // Do not send user list, the client will explicitly request it when needed
-	clone.messages = clone.messages.slice(-100);
+	clone.messages = clone.messages.slice(-1);
 	return clone;
 };
 
--- a/client/views/chat.tpl
+++ b/client/views/chat.tpl
@@ -12,7 +12,7 @@
 		<span title="{{topic}}" class="topic">{{{parse topic}}}</span>
 	</div>
 	<div class="chat">
-		<div class="show-more {{#equal messages.length 100}}show{{/equal}}">
+		<div class="show-more{{#if messages.length}} show{{/if}}">
 			<button class="show-more-button" data-id="{{id}}">Show older messages</button>
 		</div>
 		<div class="messages"></div>


src/models/chan.js
```diff
@@ -129,7 +129,7 @@ Chan.prototype.getMode = function(name) {
 Chan.prototype.toJSON = function() {
 	var clone = _.clone(this);
 	clone.users = []; // Do not send user list, the client will explicitly request it when needed
-	clone.messages = clone.messages.slice(-100);
```
MaxLeiter: could lowering this be a better solution?

dgw: Lowering the number of messages would postpone the issue until a user has more channels, but it wouldn't actually _solve_ the problem. Having the client fetch messages when a channel is activated, just like it does with user lists, provides the simplest solution to having a big fat blob of init data that has to be parsed all in one go. Lazy-loading messages for inactive channels in the background can come later, right @xPaw?

xPaw: I could probably send one message for each chananel (that'll fix the possible `lastMessageId` issue). Also I think there is no need to lazily send messages after login.

kode54:
#1658 appears to be fixed by changing this to:

```
	clone.messages = clone.messages.slice(-1);
```

kode54:
I tested this now. Now resync prevents the client from getting more than a single message, ever. The client doesn't know it needs to fetch more messages, nor does it display the "load more" button.

xPaw:
You mean reconnect?

kode54:
Any time you connect, it fetches one message for a buffer, then never shows a way to fetch the backlog prior to that one message, until at least 99 other messages come in post connect.

astorije:
@xPaw, where are you with this PR? Does it need more changes? Do you want to get it merged or was it an experiment?

kode54:
It should either send no messages, or send only one message, and have a mechanism in the client to request more messages when the window is selected. I tried one message only, and the client stopped requesting further backlog, and didn't display the button to load more messages, either.

xPaw:
@kode54 It should be doing what you are describing, did you build correctly when testing?

kode54:
I always `npm install` and `NODE_ENV=production npm run build` every time I change anything.

xPaw:
@kode54 I just tested it again, it's working as expected for me.

astorije:
@xPaw, when I load initially, the server window has everything (MOTD, etc.) but when I reload, I only get the last line:

screen shot 2017-11-21 at 22 13 46

kode54:
Exactly the issue I had with the single message feed version. That change needs the rest of the client to be altered to handle this change, so it knows to fetch up to 99 messages to fill in after the buffer is selected.

astorije:
Mmmh after re-`npm i && npm run build && npm start`-ing, I can't repro anymore...

I notice it's quite _abrupt_ to load the page, hard transitions from the loading page, then kind of a blink of the "Show more" buttons that instantly disappears, etc.
It would be nice to smooth things like this, add animations, ... Not for this PR, just an idea to keep in mind.

xPaw:
I still haven't run into this problem you're both describing. I am currently running this PR on my instance, and neither desktop or mobile have this problem.

xPaw:
Gonna close this because I don't like this as a solution (clicking on a channel flashes new messages in which is rather silly).

kode54:
I much prefer the solution where I can't even connect if my connection or device is too slow to receive 100 messages for my 120+ buffers.

xPaw:
Hm yeah, that's still a problem. Have you tested this pr once more, did it work?

I might reopen this, but make active channel send 100 messages so there's less flashing.

PR #1636, opened 2017-10-15
Cleanup client manager functions

### LINKED ISSUE #1592: Verify username consistency
I've noticed that in some places there are checks that check if the file name `{username}.json` exist, but there is also a `username` field inside of the file that is also used. There may be inconsistency in that.

xPaw:
- Fixes #1592
- Checks if a file exists directly instead of reading the whole `users` folder and checking in the array.
- Some error handling cleanup

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -1,10 +1,11 @@
 "use strict";
 
-var _ = require("lodash");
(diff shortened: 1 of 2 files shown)


MaxLeiter:
👍

Issue #1621, opened 2017-10-07
Too much padding on settings page titles on mobile


MaxLeiter:
![image](https://sr.ht/Ukmq.png)

brainloc:
Has this issue been resolved?

MiniDigger:
@brainloc yes, see #1637, this can be closed.

PR #1637, opened 2017-10-15
Reduced padding around page titles

Swapnull:
Removed excess padding around page titles as per issue #1621.
**Before**
image

**After**
image

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -66,10 +66,6 @@ h3 {
 	margin: 0;
 }
 
-h1.title {
-	margin-bottom: 10px;
-}
-
 input {
 	outline: 0;
 }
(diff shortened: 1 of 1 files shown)


client/css/style.css
```diff
@@ -67,7 +67,7 @@ h3 {
 }
 
 h1.title {
-	margin-bottom: 10px;
+	margin: -30px 0 -20px 0;
```
xPaw: Can this be fixed in a way that doesn't require negative margins?

Swapnull: Sure, I have removed the padding top of the element that was setting it rather than having negative margin that was effectively just negating it.

MaxLeiter:
👏

xPaw:
@Swapnull Can you squash your commits into one?

Swapnull:
@xPaw I have squashed the commits into one :)

PR #1650, opened 2017-10-19
Fix loading app with autocomplete disabled

### LINKED ISSUE #1647: Page load fails if autocomplete disabled
After updating from 2.4.0 to 2.5.0 on my production instance, the app simply failed to load after login, stalling on "Loading the app". The JS console showed me a nice error trace:

````
Uncaught TypeError: $(...).textcomplete is not a function
    at HTMLInputElement.<anonymous> (options.js:121)
    at HTMLDivElement.dispatch (jquery.js:5206)
    at HTMLDivElement.elemData.handle (jquery.js:5014)
    at Object.trigger (jquery.js:8201)
    at HTMLInputElement.<anonymous> (jquery.js:8269)
    at Function.each (jquery.js:362)
    at jQuery.fn.init.each (jquery.js:157)
    at jQuery.fn.init.trigger (jquery.js:8268)
    at Object.<anonymous> (options.js:125)
    at __webpack_require__ (bootstrap c22410d6bdbfb4c08f65:54)
````

_(I pulled the above trace later, from my local debug instance checked out at tag `v2.5.0`. The same bug manifests on current `master`.)_

Once I found the error in the console, I kind of just jumped on a hunch and tried clearing LocalStorage/cookies, then reloading—the app connected fine. I put all my settings back the way they had been—failure. So I cleared again and started reloading the app after every setting change.

It failed to load right after I unchecked "Enable autocomplete". Conclusion: The Lounge v2.5.0 cannot be run without autocomplete enabled on the client. I'm guessing not many of the active contributors have autocomplete turned off, or this would have been caught during testing of whatever PR broke it. (Even though I worked on several PRs during the incubation of 2.5.0, I never bothered turning off autocomplete on my debug instance… If I had, I probably would have caught this _before_ release. :disappointed:) It seems there's no test coverage of the setting, either. (I can't find any test coverage of the client at all, actually, not that I'm super familiar with the tests file layout.)

The code referenced in `options.js:121` hasn't been touched in 4 months (it was the same in 2.4.0 stable), so I thought something must have changed elsewhere. If I'm lucky, I'll come up with a PR later this week, because a full course of `git bisect` pointed me to 1e2d35f—a starting point for fixing the regression. (Not tonight, because I already stayed up longer than I should have just writing this bug report! :laughing:)

dgw:
Adds a client-side require() that was missing from 1e2d35f. Fixes #1647.

Bare `require()` because it's a `jquery` plugin that doesn't need to be referenced directly, but if there should be a `const` on there for some other reason, let me know.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const $ = require("jquery");
+require("jquery-textcomplete");
 const escapeRegExp = require("lodash/escapeRegExp");
 const settings = $("#settings");
 const userStyles = $("#user-specified-css");


YaManicKill:
Should we release a 2.5.1 with this fix in?

xPaw:
@YaManicKill It would have to be done separately and not from master branch. If @astorije wants to do it..

YaManicKill:
Yeah, my question was whether or not it is important enough to do it. I think it probably is (unless we are going to do 2.6.0 much sooner than expected) because it can literally crash the whole client for people with this setting on.

astorije:
I'm okay releasing current master as 2.5.1. What's in master is mostly greenkeeper stuff, fixes and internals. Are you okay with that? If not... 2.6.0 it is 😅

xPaw:
@astorije There's some changes in master that I'd definitely want to be tested more, so let's not rush into releasing it.

YaManicKill:
Let's be radical and just feature freeze 2.6.0, test it for a couple of days and release.

dgw:
Why not cherry-pick 72e5844, and maybe 50504ed, into a bugfix branch? Don't even need to include dependency updates, just fixes for known things that could cause a crash. This type of situation is precisely what patch-versions are for.

PR #1656, opened 2017-10-21
Adding ability to color usernames based on their hash value

DarkMio:
Simple HSL color rotation based on user hash. Generates 359 unique colors per hash, looks more diverse than the current color limitation.

The settings are a bit wonky and could use unified styling.

Screenshot: https://i.imgur.com/w7HtUt3.png  
Demo Video: https://a.pomf.cat/yacrsp.mp4

PR is probably work in progress until the layout doesn't look so odd anymore.

Edit: Also I take some nice username suggestions, other than the ones in the video or "JohnDoe", etc.
Additionally I'll look into supplying (or just plain calculating) SL values based on the themes usage of that.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -268,6 +268,18 @@
 										</option>
 									{{/each}}
 								</select>
+								<div class="col-sm-12">
+									<h3>Username Coloring</h3>
+								</div>
+								<label for="saturation">Saturation</label>
+								<input id="sat-select" name="saturation" type="range" min="0" max="100" value="70" class="input">
+								<label for="lightness">Lightness</label>
+								<input id="light-select" name="lightness" type="range" min="0" max="100" value="35" class="input">
+								<div id="user-example-coloring">
+									<span id="example-user-1" class="user" style="color: hsl(0, 50%, 50%)">DankMio</span><br>
+									<span id="example-user-2" class="user" style="color: hsl(120, 50%, 50%)">xPaw</span><br>
+									<span id="example-user-3" class="user" style="color: hsl(240, 50%, 50%)">Alram</span>
+								</div>
 							</div>
 							{{#if prefetch}}
 							<div class="col-sm-12">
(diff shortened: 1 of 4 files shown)


MaxLeiter:
Lightness -> brightness

DarkMio:
It's lightness. Brightness is perceived lightness, which varies by a roundtrip on a certain lightness value.

MaxLeiter:
I've never seen the term lightness regarding UI elements and think it may cause confusion

DarkMio:
I don't really care about the final wording - brightness works for most users, too, I guess.

McInkay:
I'm not sure about this, it feels quite niche to put into the settings, tbh.

If there are issues with username colours on particular themes, those should be fixed in the themes, and if individuals want to change the colours of certain usernames, that can be done with css.

Similarly, once we have options implemented in plugins (I'll be doing that once #1851 is merged) this could be implemented as a plugin.

Thoughts, @thelounge/maintainers ?

astorije:
I'm with @McInkay, it feels quite niche and totally a good candidate for packages when they're able to support this.

Also, if I understand it correctly, this entirely removes color control from themes, which isn't very nice. For example, one could write a theme bleu-monochrome, and choose to have nicks show accordingly (that would limit to fewer colors, but that's beside the point). This PR would enforce colors to such theme.

If this PR were to solve accessibility issues directly and/or would be the only accessibility alternative, we would probably approach it with a different angle, but we do have the option to disable colored nicknames (which solves any contrast issue). Also note that we used to have auto-generated colors before switching to the current solution, because it had its entire set of issues.

Like @McInkay said, thank you very much for your contribution still, and we'll try to remember to ping this PR when packages allow you to do this. Feel free to open an issue next time if you have an idea you'd like to work on, to check what we feel about it :)

McInkay:
Alright, closing then.

PR #1661, opened 2017-10-24
Fix url query parameters in public mode

MaxLeiter:


dgw:
I was confused what this was doing based on the PR title, but looking at the code I see what it's doing. Good change :+1:

PR #1662, opened 2017-10-24
Allow run-pr script to pass arguments to npm start

MaxLeiter:
Allows `./scripts/run_pr.sh 1000 --public`

### DIFF
--- a/scripts/run-pr.sh
+++ b/scripts/run-pr.sh
@@ -13,4 +13,5 @@ git rebase master
 npm install
 NODE_ENV=production npm run build
 npm test || true
-npm start
+shift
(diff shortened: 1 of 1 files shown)


scripts/run-pr.sh
```diff
@@ -13,4 +13,4 @@ git rebase master
 npm install
 NODE_ENV=production npm run build
 npm test || true
-npm start
+npm start -- "${@:2}"
```
astorije: This is bash specific, not sh. We could tell the file to use bash, or use "shift", which seems more common to me.

MaxLeiter:
done, @astorije

Issue #282, opened 2016-04-26
A logo for the project


xPaw:
cc @thelounge/maintainers.

We should think of a logo for our project.
- github org logo
- favicons for the client
- optionally logo on the site?

This could be a good inspiration https://www.google.ee/search?q=lounge+logo&tbm=isch

MaxLeiter:
I think a couch/lounge chair would be fitting, something like ![this](https://cdn2.iconfinder.com/data/icons/windows-8-metro-style/256/sofa.png)

YaManicKill:
I think something liek that but incorporated with something liek a speach bubble. A way to make it clear that it's about a place to congregate to discuss, not just to like .. look at each other or something.

maxpoulin64:
@YaManicKill like that?

![blah](https://cloud.githubusercontent.com/assets/5481612/14830325/29b4873e-0bbe-11e6-8300-ff13b410676c.jpg)

xPaw:
:ship:

astorije:
@guibbs, if you still want to offer your help, now is a good time to speak up :P
FYI, he is a former colleague who happens to be really good with logos. @guibbs, do you have some example of your recent work? I don't have them on hand.

astorije:
I'm adding the `2.0.0` milestone for now, it would be truly great if version 2 comes with its very own logo! :-)
I know @guibbs has been looking at options the last few days. Feel free to post any preliminary work here, but no pressure!

MaxLeiter:
After some discussion on IRC, here's my personal favorite:

![https://news.ycombinator.com/item?id=12574302](https://sr.ht/ac7t.png)

xPaw:
@simonv3 also proposed an interesting idea:
![screenshot 2016-10-15 08 50 55](https://cloud.githubusercontent.com/assets/613331/19417289/15296088-93b2-11e6-8429-9ed9a97d0c97.png)

elioqoshi:
What do you think about this?
Font is Exo 2, Open Font License of course.
![draft-01](https://cloud.githubusercontent.com/assets/5436686/19417470/1598620c-93ae-11e6-95dd-7c94620489e9.jpg)

elioqoshi:
![1-01](https://cloud.githubusercontent.com/assets/5436686/19417556/4a9acd1c-93b0-11e6-8166-5ebf8f41c5a0.jpg)

Here another one, which works as a couch and message pun, so to speak

MaxLeiter:
[...]

PR #1668, opened 2017-10-28
Center loading screen and add logo

xPaw:
| Before | After |
| - | - |
|  |  |

Javascript message was broken in 1d47290ada0a01cc0144444da6760bfac12821c2


client/index.html
```diff
@@ -95,7 +88,7 @@ <h1 class="title" id="loading-title">The Lounge is loading…</h1>
 	<div id="image-viewer"></div>
 
 	<script src="js/bundle.vendor.js"></script>
-	<script src="js/bundle.js"></script>
+	<script src="js/bundlez.js"></script>
```
MaxLeiter: accidental `z`?

client/css/style.css
```diff
@@ -1498,22 +1497,33 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+	flex-direction: column;
+	text-align: center;
 }
 
 #loading p {
 	margin-top: 10px;
 }
 
+#loading img {
+	max-width: 256px;
```
astorije: Would you mind adding a `max-height` or something (Chrome tells me the image shows as 256px x 171.48px)? On load there is a bump until the image is loaded, which is right in the middle of loading (at least in my case).
I'm playing with slow network and it does that too.

astorije:
This is so nice!

I noticed that on screens that aren't very high, like landscape mode on mobile, it tends to do this:

screen shot 2018-02-28 at 00 56 32

How expensive would it be to fix this?

xPaw:
@astorije Fixed it up by further using flexbox instead of absolute positioning.

I don't know how to handle image jump when it loads though. We'd have to define exact image size for that not to happen.

client/index.html.tpl
```diff
@@ -49,22 +49,15 @@
...
-								<p id="loading-slow">
-									This is taking longer than it should, there might be
-									connectivity issues.
-								</p>
-								<button id="loading-reload" class="btn">Reload page</button>
-								<script async src="js/loading-error-handlers.js"></script>
-							</div>
-						</div>
+					<div id="loading-status-container">
+						<img src="img/logo-vertical-transparent-bg.svg" alt="The Lounge" width="256" height="170">
```
astorije: I often forget, but we allow server config to specify a custom theme, which is then used on loading screen. When theme is dark, result is pretty horrible:

screen shot 2018-02-28 at 23 42 19

What about having a second image here ([this one](https://github.com/thelounge/thelounge.github.io/blob/astorije/logos/assets/logos/logo/TL_White%26Yellow_Vertical_logotype_Transparent_Bg/TL_White%26Yellow_Vertical_logotype_Transparent_Bg.svg)) hidden with CSS, then custom themes can hide the normal one and display inverted logo only?

(Alternatively, get rid of server config for theme and only have official theme on page load... :trollface: I know, I know, this idea is not liked 😅)

astorije:
Very nice fixes @xPaw, good job!

client/index.html.tpl
McInkay: > I know, I know, this idea is not liked

It's not liked because those of us using dark themes don't like a bright flash when loading the lounge. If we were to get custom themes to sync from client to server, then I'd be fine with removing the server theme. Only then.

Also, bear in mind that even if you did remove the server theme, the client theme takes over before this screen goes away sometimes, so this still needs to be dealt with.

astorije: That was not a real suggestion, just a bit of trolling :D
Real suggestion is the 2 images, but there are probably others (css background?)

McInkay: I know you were trolling, I'm just making sure people know the reasons for it :-P 

As for the actual feature...I'm not sure which I prefer. The hidden image is easiest for theme makers, but also most feels like a hack. Maybe just making both images available on the server and themes can set css background, as you say. Shouldn't be too difficult?

Other option (this is bigger than this PR I think) is we can have a way for themes to set themselves as "light" or "dark" and they can inherit things like this from defaults.

xPaw: We could do some color changing dynamically (e.g. from manifest or background color) as it's `svg`.

McInkay: That could also work.

xPaw:
Okay I've embedded the SVG in html itself, this allows themes to individually change colors of the logo.

dgw:
How does letting themes change the logo colors square with the brand guidelines given by the logo designer in #282? Specifically, "Don't alter or change the colors".

astorije:
What @dgw said is on point. Let's not do this, it would be a recipe to disaster.
If someone's theme does not work with a given background, then they should hide the logo instead of altering it.

Let's go with either having 2 images, or choosing the image via CSS background like @McInkay mentioned. I don't have a preference over which of those solutions though.

xPaw:
Brand guideline or not, I think this is a slightly more elegant solution, as opposed to having to switch images (either there are 2 img tags in the page, or change background-image).

astorije:
That requires theme authors to set up our specific colors as opposed to pick an image, I don't find that elegant at all from a theme author perspective. Also not very elegant to embed the SVG code within the markup IMO.
And I am 💯% sure that themes will go nuts on the logo colors.

I think we really shouldn't do that.

xPaw:
@McInkay which method do you prefer? Coloring individual parts via CSS (svg is embedded in html) or changing background image via css?

I don't want to put 2  tags because then the browser will start downloading both images.

McInkay:
Euch.

I think I'm probably with @astorije - we have strict guidelines around the logo, so we shouldn't encourage breaking the guidelines. So probably background css.

xPaw:
Mind you, changing image url won't stop theme designers from switching to any other logo they've designed (so they could copy the svg and change the colors there)...

astorije:
All things considered, I would prefer having both `` elements over background CSS, and way way over SVG + custom colors. Optimized SVG file is 2.62 KB, 475 bytes transferred over the network (according to Chrome devtools). I think we can live with the overhead :)

PR #1676, opened 2017-10-31
handle empty client queries

realies:


xPaw:
It's probably better to combine this with the previous `if` so that `/query` without spaces also displays a message.

And there are linting errors in your PR.

src/plugins/inputs/query.js
```diff
@@ -7,11 +7,15 @@ var Msg = require("../../models/msg");
 exports.commands = ["query"];
 
 exports.input = function(network, chan, cmd, args) {
-	if (args.length === 0) {
+	var target = args[0];
+	if (args.length === 0 || target.length === 0) {
+		chan.pushMessage(this, new Msg({
+			type: Msg.Type.ERROR,
+			text: "You can not open a query window without an argument."
```
astorije: Nitpick but can you change `can not` into `cannot` please? 💅

astorije:
Awesome @realies, thanks! Could you squash your commits?

PR #1683, opened 2017-11-04
Split index.html into components

### LINKED ISSUE #1055: Split index.html into multiple files

### LINKED ISSUE #1672: Make index.html static, send defaults over WS
Currently index.html is rendered on the server with handlebars, which is rather unnecessary, this also leaks `defaults` from config in private mode.

Making index.html static would also completely remove client dependency on the server.

Related to #1055

xPaw:
- Fixes #1672 - Send configuration variables (default network, git hash of current build) over socket.io after authentication to improve privacy
- Fixes #1055 - Split `index.html` into multiple views that are rendered on the client (this makes #1271 a lot easier)
- Sets `theme-color` in `index.html` to value from `manifest.json`
- Removes `express-handlebars` (goes back to how it was done in Shout with lodash templates)
- Disables `X-Powered-By` in Express


xPaw:
@thelounge/maintainers I think this is done, needs testing though.

src/server.js
```diff
@@ -33,27 +32,15 @@ module.exports = function() {
 (node ${colors.green(process.versions.node)} on ${colors.green(process.platform)} ${process.arch})`);
 	log.info(`Configuration file: ${colors.green(Helper.CONFIG_PATH)}`);
 
-	if (!fs.existsSync("public/js/bundle.js")) {
-		log.error(`The client application was not built. Run ${colors.bold("NODE_ENV=production npm run build")} to resolve this.`);
-		process.exit();
-	}
-
 	var app = express()
+		.disable("x-powered-by")
```
astorije: Seriously Express...

Unrelated to this PR, but it might be worth looking into https://expressjs.com/en/advanced/best-practice-security.html#use-helmet. What do you think?

astorije:
Rebased on master to fix conflicts (dangling comma stuff).

client/index.html
```diff
@@ -9,23 +9,23 @@
 	<link rel="preload" as="script" href="js/bundle.js">
 	<link rel="stylesheet" href="css/bootstrap.css">
 	<link rel="stylesheet" href="css/style.css">
-	<link id="theme" rel="stylesheet" href="{{ theme }}">
+	<link id="theme" rel="stylesheet" href="<%- theme %>">
```
astorije: Why do we go back to Lodash templates? It was overall a good thing that we switch from that to Handlebars, since we use Handlebars for anything else 😕

src/server.js
xPaw: It doesn't provide any advantage.

client/index.html
xPaw: Since most templates moved to client, they are still being compiled by handlebars. But this leaves just a couple of variables that need to be set in index.html, and it felt like a good idea to drop a somewhat heavy runtime dependency from the server.

astorije: > it felt like a good idea to drop a somewhat heavy runtime dependency from the server.

Yep, absolutely!

Overall, shouldn't we be serving a totally static `index.html` and all values would be set by the client (using that same `getClientConfiguration`)? That would solve this along with a few other things, no?

xPaw: It would be nice, yes, but these variables that I kept have to be set before JS loads (especially `transports` which is used for connecting socket.io), and the initial delay of loading the theme.

astorije:
Hey @xPaw! Testing it, it seems to work fine, so far I've only noticed this issue (but more testing is in order): defaults don't seem to apply on the client. See:

On `master` | On this PR
--- | ---
screen shot 2017-11-19 at 12 40 15 | screen shot 2017-11-19 at 12 28 43

xPaw:
@astorije Fixed.

client/index.html
```diff
@@ -79,805 +79,10 @@ <h1 class="title">The Lounge is loading…</h1>
...
-
-							<a href="https://thelounge.github.io/" target="_blank" rel="noopener">Website</a><br>
-							<a href="https://thelounge.github.io/docs/" target="_blank" rel="noopener">Documentation</a><br>
-							<a href="https://github.com/thelounge/lounge/issues/new" target="_blank" rel="noopener">Report a bug</a>
-						</p>
-					</div>
-				</div>
+				<div id="sign-in" class="window"></div>
+				<div id="connect" class="window"></div>
+				<div id="settings" class="window" data-type="settings"></div>
```
astorije: I don't think `data-type="settings"` is being used anymore. Not sure if you want to kill it in this PR or later, but something to keep in mind.

PR #1687, opened 2017-11-07
Update user from CLI

### LINKED ISSUE #1686: Change add <user> to allow updates, if user already exists
When updates are posted to thelounge, I run an automated script to create a new user. It will be useful to modify the add user command with the following,

* pass along password as an optional CLI argument (similar to mysql -u -p command)
* just update password if the user already exists, but leave other options intact

NOTE: I have a patch for this and will be updating this soon.

abrahamvarricatt:
Resolves #1686.

This change will allow provisioning systems to create/update users
without propting for input from STDIN. A nice benefit is that if
the user already has some networks configured, they should still
remain intact.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -100,19 +100,20 @@ ClientManager.prototype.addUser = function(name, password, enableLog) {
 
 	const userPath = Helper.getUserConfigPath(name);
 
+	var user = {
+		password: password || "",
+		log: enableLog || false,
+		awayMessage: "",
+		networks: [],
+		sessions: {},
+	};
+
 	if (fs.existsSync(userPath)) {
+		// If user already exists, update their config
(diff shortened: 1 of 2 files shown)


astorije:
Hey @abrahamvarricatt, do you have an example on how this gets used that you could post here?

xPaw:
I already commented on the issue, but I will mirror here:
> This seems like a very niche use case, you should update your scripts to do what you want instead of modifying the add command. There is a reset command which does what you want -- change the password.

astorije:
I agree with @xPaw.

@abrahamvarricatt, I'll be closing this PR for now per what @xPaw said, but I am still interested at use case and example usage for this. It would be great if you could post one.
Maybe you have a perfectly valid use case that we're not seeing, in which case we can always re-open your PR :)

Thanks for your effort and contribution either way! 👋

PR #1691, opened 2017-11-08
fix gap between #sidebar and #footer

realies:
image

Currently #sidebar has `bottom: 48px` and the #footer has `height: 45px`. This should fix the gap.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -431,7 +431,7 @@ kbd {
 }
 
 #sidebar {
-	bottom: 48px;
+	bottom: 45px;
 	left: 0;
 	overflow: auto;
 	overflow-x: hidden;


astorije:
Actually, I looked at this quickly, and because of themes, this is not as straightforward. I'll take a closer look and get back to you soon @realies. Good catch though!!

xPaw:
@astorije I looked through the themes, and it's not an issue anywhere, so it's fine.

astorije:
I was also having a hard time figuring what I meant, but it's actually the "Example" theme that I was talking about: because of [this line](https://github.com/thelounge/lounge/blob/master/client/themes/example.css#L14), the fix in this PR doesn't apply to "Example".

@realies, am I assuming correctly that your screenshot is for Morning? The fix works for this theme indeed.
Could you update `example.css#L14` to be `49px` instead of 52? (I tried and anything under 49 will make sidebar/footer overlap).

xPaw:
@astorije I thought the margin around footer in example theme is fine because footer has margin on all sides, not just top.

astorije:
Aaah, that's a very fine point. I was missing it because all I was playing with... was the window zoomed at the maximum :) Perfect illustration of the expression "missing the big picture" :D

Anyway, that's perfect then. Thanks @realies, sorry it took me so long to react and then notice my mistake!

PR #1698, opened 2017-11-10
Refactor how user object is sent to the client

xPaw:
This was started by `kick` event where users are sent as `User` objects instead of putting nick/mode in the message itself.

This changes all `var` to use `const` and `let` in `irc-events` folder. This also changes events like join/part/quit to send message to client first before updating the user list (just a minor improvement).


astorije:
Rebased on master, fixed comma conflict shit...

PR #1700, opened 2017-11-11
Display in query when user disconnects and re-connects using ISON, closes #337

### LINKED ISSUE #337: User join/leave/disconnect not shown in query window
I suggest query windows should show when the user disconnects or joins one of the channel we're both on.

**Example:**
As a persistent service, a user is often away form the IRC window for a long time, thus when another user opens a query with the user, it's good to know whether that other user is still online or not when eventually getting the message.

**Implications:**
These notifications are only caught when join/leave/disconnect events for a user happen in a channel we are also in. Kinda hackish.

Suggestions (taken from conversation about that issue on #thelounge irc channel):
- Using the topic bar to display something like "Online on: #thelounge #freenode" or "Last seen 20 minutes ago on #thelounge"( @maxpoulin64 )  , add hostmask as well ( @xPaw )

MaxLeiter:
WIP, almost done.

Need a better way to create the messages - ideally they'd be created on the server (so they could be assigned a msgid), however due to the nature of `ison` that's a PITA. The lack of an ID causes the unread marker to act a little odd. @xPaw, suggestions?

Also, should somehow compare hostnames between when the user was last connected and if they are now. Again, due to `ison` primitiveness, not sure best way to do this. Suggestions appreciated.

screen shot 2017-11-11 at 12 07 23 am

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -116,6 +116,10 @@ function enableHistory() {
 	inputTrap.bind("enter", function() {
 		position = 0;
 
+		if (input.data("autocompleting")) {
+			return false;
+		}
+
 		const text = input.val();
 
 		if (text.length === 0) {


client/js/lounge.js
```diff
@@ -312,6 +312,15 @@ $(function() {
 		if (!target) {
 			return;
 		}
+
+		const targetTitle = self.data("title");
+		if (targetTitle.indexOf("#") === -1 && !self.hasClass("lobby")) { // no # or .lobby; assume it's a user query
```
xPaw: This entire thing can be done on the server inside `open`, and it can check if open channel type is query.

xPaw:
I don't see a reason why all this code is on the client, you say it's a PITA to do on the server, but what's the reason because I fail to see one. You can add `state` (or something like that) on `Chan` object on the server to send updates to client correctly.

I would also modify `quit` event to show quits for common channels in query.

client/js/constants.js
```diff
@@ -72,6 +72,8 @@ const actionTypes = [
 	"away",
 	"back",
 	"ban_list",
+	"connect",
```
xPaw: Reusing quit/join could probably work just fine.

src/plugins/irc-events/ison.js
```diff
@@ -0,0 +1,49 @@
...
+module.exports = function(irc) {
+	const client = this;
+	irc.on("users online", (data) => {
+		if (data.nicks.length === 1) {
+			const target = client.find(client.lastActiveChannel);
+			const chan = target.chan;
+			if (chan.type === Chan.Type.QUERY) {
+				const messages = chan.messages;
+				let lastMessageType;
+				for (let i = 0; i < (messages.length > 10 ? 10 : messages.length - 1); i++) {
```
xPaw: Why not simply add `isOnline` to `chan` object and compare that?

src/plugins/irc-events/ison.js
```diff
@@ -0,0 +1,49 @@
+"use strict";
+
+const Msg = require("../../models/msg");
+const Chan = require("../../models/chan");
+
+module.exports = function(irc) {
+	const client = this;
+	irc.on("users online", (data) => {
+		if (data.nicks.length === 1) {
+			const target = client.find(client.lastActiveChannel);
```
xPaw: This is gonna be weird with multiple clients open. As we discussed on IRC, the better way is most likely querying all names with `ison` at once.

MaxLeiter: and just compare the array of open queries with the returned array and find what's missing?

xPaw: Pretty much, yeah.

astorije:
Hey @MaxLeiter, do you think you'll get back to this? :)

MaxLeiter:
yeah, want to finish up nicks in links first

astorije:
Cool! It seems to me that this other PR is almost done, so that's great!

xPaw:
@MaxLeiter want to revisit?

astorije:
@MaxLeiter, @xPaw, do you know what's the status of this?

MaxLeiter:
@astorije: requires a full rewrite on how it works. It's going to have an array of users + their status and refresh everyone n minutes (or something similar), instead of when you click on the query.

At least, that's what I remember. @xPaw?

xPaw:
I believe we talked about using MONITOR for this and as a fallback use quit/part/join messages (and probably ISON on open like it is now).

And instead of iterating last X messages use `Chan.State` to store whether user is currently online.

astorije:
@MaxLeiter, it looks like the pr auto closed, you can reopen it by pushing new commits to the branch

MaxLeiter:
Yeah I’m working locally before commiting.

PR #1706, opened 2017-11-12
Fix local theme folder

xPaw:


src/plugins/themes.js
```diff
@@ -49,7 +49,7 @@ function makeLocalThemeObject(css) {
 	const themeName = css.slice(0, -4);
 	return {
 		displayName: themeName.charAt(0).toUpperCase() + themeName.slice(1),
-		filename: path.join(__dirname, "..", "client", "themes", `${themeName}.css`),
```
xPaw: How did this work before, I am not sure. The final path would be `lounge/src/client/themes` where `src/` is incorrect.

PR #1707, opened 2017-11-12
Add a script to pre-generate changelog entries

astorije:
I have been using and improving that script for some time now, at least all 2.5 and 2.6 releases. 👌

### DIFF
--- a/scripts/.eslintrc.yml
+++ b/scripts/.eslintrc.yml
@@ -1,0 +1,5 @@
+---
+
+# Necessary to support async/await... grumble grumble...
+parserOptions:
+  ecmaVersion: 8
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -50,7 +50,11 @@ rules:
   semi-style: [error, last]
   semi: [error, always]
   space-before-blocks: error
-  space-before-function-paren: [error, never]
+  space-before-function-paren:
+    - error
+    - anonymous: never
+      named: never
+      asyncArrow: always # Otherwise requires `async()`
   space-in-parens: [error, never]
   space-infix-ops: error
   spaced-comment: [error, always]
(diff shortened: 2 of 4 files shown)


scripts/.eslintrc.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+# Necessary to support async/await... grumble grumble...
+parserOptions:
+  ecmaVersion: 8
+
+rules:
+  space-before-function-paren:
```
xPaw: Can you modify eslintrc in root to match this?

scripts/.eslintrc.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+# Necessary to support async/await... grumble grumble...
+parserOptions:
+  ecmaVersion: 8
```
xPaw: You can do this with a eslint comment within the `changelog.js` file instead of making a config file.

astorije: It seems to me that `parserOptions` configs cannot be part of an ESLint comment 💩
Or at least I couldn't find how... Any idea?

scripts/.eslintrc.yml
astorije: Sure thing! It's kinda clumsy though in the current shape, do you mind if I change overall into `space-before-function-paren: error` (i.e. `always`)? It's seems way more common to me and ESLint's default.
Plus I can do that automatically with `--fix` to not break anything.

Will do in another PR if you're okay with this.

xPaw: Oh, that's gonna affect so much that I wouldn't recommend doing it.

astorije: Yeah, about 400 `function()`... I really hate those, but whatever.

Anyway, moved this to the root config!

astorije:
Merging this now as I'm going to use it to draft the v2.6.0 release. See you on the other side!

PR #1708, opened 2017-11-12
Switch CLI tool from `lounge` to `thelounge`, deprecate `lounge`

astorije:
This is in preparation of The Lounge v3 which will make `thelounge` (aliased `tl`) uniform across the entire project. No more confusion!

screen shot 2017-11-19 at 03 18 09

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "version": "2.6.0",
   "preferGlobal": true,
   "bin": {
-    "lounge": "index.js"
+    "lounge": "index.js",
+    "thelounge": "index.js"
   },
   "repository": {
     "type": "git",
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -39,6 +39,13 @@ require("./reset");
 require("./edit");
 require("./install");
 
+// TODO: Remove this when releasing The Lounge v3
+if (process.argv[1].endsWith(`${require("path").sep}lounge`)) {
+	log.warn(`The ${colors.red("lounge")} CLI is ${colors.bold("deprecated")} and will be removed in v3.`);
+	log.warn(`Use ${colors.green("thelounge")} instead.`);
+	process.argv[1] = "thelounge";
+}
+
 program.parse(process.argv);
 
 if (!program.args.length) {
--- a/README.md
+++ b/README.md
@@ -40,13 +40,13 @@ Run this in a terminal to install (or upgrade) the latest stable release from
 When installation is complete, run:
 
 ```sh
-lounge start
+thelounge start
 ```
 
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -7,7 +7,9 @@
   },
   "preferGlobal": true,
   "bin": {
-    "lounge": "index.js"
+    "lounge": "index.js",
+    "thelounge": "index.js",
+    "tl": "index.js"
```
astorije: I do not feel strongly about the alias. I thought it would be a nice-to-have to make the transition nicer, as I'm sure lots of people will not like having to type extra characters. I don't mind keeping it or removing it depending on what people think.

xPaw: I don't see a need to have this alias.

astorije: Meh, not a need, but a convenience. Anyway, removed.

astorije:
Without the `tl` alias:

screen shot 2017-11-19 at 03 19 00

src/command-line/index.js
```diff
@@ -39,6 +39,13 @@ require("./reset");
 require("./edit");
 require("./install");
 
+// TODO: Remove this when releasing The Lounge v3
+if (process.argv[1].endsWith("/lounge")) {
```
YaManicKill: Why `endswith`? Wouldn't that make it match "thelounge"? Also, is the `/` a typo?

xPaw: Slash is specifically so it doesn't match with `thelounge`. But does that slash have to match OS default one?

astorije: @YaManicKill, the idea is that `process.argv[1]` always contains the full path, so I can't test for equality, hence the forward slash.

@xPaw, that's a good point! Would you mind running `console.log(process.argv[1])` for me and let me know what it tells you? I don't know if Node normalizes those or not.

I also don't want to do a `!.endsWith("thelounge")` because that would match when running `npm start`.
If `/` is an issue, I'll go with a `/(^the)lounge$/` regex (but I suck at regex so that might just be wrong, I'd need to play) or `.endsWith("lounge") && !.endsWith("thelounge")`. Gross, but it's a temporary check anyway so whatever.

xPaw: Use `${path.sep}lounge`.

astorije: Ah, awesome! Done.

dgw:
Thoughts on renaming this repository from `thelounge/lounge` to `thelounge/thelounge` when the CLI tool is renamed? Or for v3?

astorije:
@dgw, we are planning to do that for v3 yep! Thanks for the reminder, I forgot to create an issue for it. Actually, I forgot if I created an issue for it or not... 🤔

astorije:
@xPaw, @YaManicKill, are you good with this now?

PR #1709, opened 2017-11-14
 Link nicks mentioned in messages 

MaxLeiter:
Not sure how I like this. kiwi does it. Can disable in settings (but requires reload to change sent messages, which is fine imo). Setting probably needs better wording.

Should be good to go.

screen shot 2017-11-14 at 6 50 54 pm

relies on #1712

### DIFF
--- a/client/js/libs/handlebars/ircmessageparser/findNames.js
+++ b/client/js/libs/handlebars/ircmessageparser/findNames.js
@@ -1,0 +1,17 @@
+"use strict";
+
+function findNames(text, users) {
+	const result = [];
+	let index = 0;
+	users.forEach((nick) => {
+		index = text.indexOf(nick, index);
+		result.push({
+			start: index,
+			end: index + nick.length,
+			nick: nick,
+		});
+	});
+	return result;
+}
+
+module.exports = findNames;
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -6,7 +6,9 @@ const anyIntersection = require("./ircmessageparser/anyIntersection");
 const findChannels = require("./ircmessageparser/findChannels");
 const findLinks = require("./ircmessageparser/findLinks");
 const findEmoji = require("./ircmessageparser/findEmoji");
+const findNames = require("./ircmessageparser/findNames");
 const merge = require("./ircmessageparser/merge");
+const colorClass = require("./colorClass");
 
 // Create an HTML `span` with styling information for a given fragment
 function createFragment(fragment) {
@@ -47,9 +49,14 @@ function createFragment(fragment) {
 	return escapedText;
 }
 
-// Transform an IRC message potentially filled with styling control codes, URLs
-// and channels into a string of HTML elements to display on the client.
-module.exports = function parse(text) {
+// Transform an IRC message potentially filled with styling control codes, URLs,
+// nicknames, and channels into a string of HTML elements to display on the client.
+module.exports = function parse(text, users) {
+	// if it's not the users we're expecting, but rather is passed from Handlebars (occurs when users passed to template is null or undefined)
+	if (users && users.hash) {
+		users = [];
+	}
+
 	// Extract the styling information and get the plain text version from it
 	const styleFragments = parseStyle(text);
 	const cleanText = styleFragments.map((fragment) => fragment.text).join("");
@@ -62,11 +69,13 @@ module.exports = function parse(text) {
 	const channelParts = findChannels(cleanText, channelPrefixes, userModes);
 	const linkParts = findLinks(cleanText);
 	const emojiParts = findEmoji(cleanText);
+	const nameParts = findNames(cleanText, (users || []));
 
 	// Sort all parts identified based on their position in the original text
 	const parts = channelParts
 		.concat(linkParts)
 		.concat(emojiParts)
+		.concat(nameParts)
 		.sort((a, b) => a.start - b.start || b.end - a.end)
 		.reduce((prev, curr) => {
 			const intersection = prev.some((p) => anyIntersection(p, curr));
@@ -77,7 +86,7 @@ module.exports = function parse(text) {
 			return prev.concat([curr]);
 		}, []);
 
-	// Merge the styling information with the channels / URLs / text objects and
+	// Merge the styling information with the channels / URLs / nicks / text objects and
 	// generate HTML strings with the resulting fragments
 	return merge(parts, styleFragments).map((textPart) => {
 		// Create HTML strings with styling information
@@ -92,6 +101,9 @@ module.exports = function parse(text) {
 			return `<span class="inline-channel" role="button" tabindex="0" data-chan="${escapedChannel}">${fragments}</span>`;
 		} else if (textPart.emoji) {
 			return `<span class="emoji">${fragments}</span>`;
+		} else if (textPart.nick) {
+			const nick = Handlebars.Utils.escapeExpression(textPart.nick);
+			return `<span role="button" class="user ${colorClass(nick)}" data-name="${nick}">${fragments}</span>`;
 		}
 
 		return fragments;
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -4,6 +4,7 @@ const Chan = require("../../models/chan");
 const Msg = require("../../models/msg");
 const LinkPrefetch = require("./link");
 const cleanIrcMessage = require("../../../client/js/libs/handlebars/ircmessageparser/cleanIrcMessage");
+const nickRegExp = /[\w[\]\\`^{|}-]{4,}/gi;
 
 module.exports = function(irc, network) {
 	const client = this;
@@ -88,6 +89,14 @@ module.exports = function(irc, network) {
 			highlight = network.highlightRegex.test(data.message);
 		}
 
+		const users = [];
+		let match;
+		while ((match = nickRegExp.exec(data.message))) {
+			if (chan.findUser(match[0])) {
+				users.push(match[0]);
+			}
+		}
+
 		const msg = new Msg({
 			type: data.type,
 			time: data.time,
@@ -95,6 +104,7 @@ module.exports = function(irc, network) {
 			text: data.message,
 			self: self,
 			highlight: highlight,
+			users: users,
 		});
 
 		// No prefetch URLs unless are simple MESSAGE or ACTION types
--- a/test/client/js/libs/handlebars/parse.js
+++ b/test/client/js/libs/handlebars/parse.js
@@ -243,6 +243,49 @@ describe("parse Handlebars helper", () => {
 		expect(actual).to.deep.equal(expected);
 	});
 
+	it("should find nicks", () => {
+		const testCases = [{
+			users: ["MaxLeiter"],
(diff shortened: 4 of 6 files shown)


christer88:
Big thumbs up for this!

MaxLeiter:
I think this is good to go

client/js/options.js
```diff
@@ -18,14 +18,15 @@ const options = {
...
+	linkNicks: true,
 	motd: true,
 	notification: true,
 	notifyAllMessages: false,
 	showSeconds: false,
 	statusMessages: "condensed",
 	theme: $("#theme").attr("href").replace(/^themes\/(.*).css$/, "$1"), // Extracts default theme name, set on the server configuration
 	thumbnails: true,
-	userStyles: userStyles.text(),
+	userStyles: userStyles.text()
```
astorije: Revert this change

client/js/libs/handlebars/ircmessageparser/findNames.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
+
+function findNames(text) {
```
astorije: Assuming this PR gets accepted overall, these 2 files + changes to `parse` could totally use some test cases :)
Don't worry about it for now though, let's get the rest of the PR settled first.

client/index.html
```diff
@@ -221,6 +221,12 @@ <h1 class="title">Settings</h1>
 									Show seconds in timestamp
 								</label>
 							</div>
+							<div class="col-sm-7">
+								<label class="opt">
+									<input type="checkbox" name="linkNicks">
+									Make mentioned names clickable
+								</label>
+							</div>
```
astorije: I know this is gonna add up to my rep, but I'm not at all in favor of a user setting for this.
If this PR works and we want it (I personally do!), let's make some UX decision and simply make this a thing.
Code will be simpler, less branching (i.e. less things to make sure we do not break), and it's a good addition overall. 👍

astorije:
This is a very fine addition, I haven't tested but love it already 👏 👏 Awesome job @MaxLeiter!

Last time this was attempted (client-side only), @xPaw has some performance concerns. The way you did it seems to me like you do not have this issue anymore. I'll let @xPaw give a first pass at this for this reason.

MaxLeiter:
I'm all for no setting 👍

client/js/options.js
MaxLeiter: Fixed

client/js/libs/handlebars/ircmessageparser/findNames.js
MaxLeiter: 👍

client/js/libs/handlebars/ircmessageparser/parseName.js
```diff
@@ -22,31 +20,26 @@ module.exports = function(orig) {
...
-				suffix = "'s";
-				strip++;
-			}
+	const prefixes = ["@", "+", "%", "!"];
+	prefixes.map((e) => {
+		if (orig.startsWith(e)) {
+			cleaned = cleaned.substring(e.length, cleaned.length);
+			prefix = e;
 		}
+	});
```
RJacksonm1: This prefix and suffix logic could be simplified in another way, without multiple `startsWith` checks:

```
const prefixes = ["@", "+", "%", "!"];

let prefix = "";
let matchedPrefixPos = prefixes.indexOf(orig[0]);

if (matchedPrefixPos !== -1) {
    prefix = prefixes[matchedPrefixPos];
    cleaned = cleaned.substring(0, cleaned.length - prefix.length);
}
```

src/plugins/irc-events/message.js
```diff
@@ -87,14 +87,25 @@ module.exports = function(irc, network) {
 			highlight = network.highlightRegex.test(data.message);
 		}
 
+		const users = [];
+		const splitMessage = data.message.split(" ");
+		splitMessage.forEach((val, index) => {
+			const cleanedNick = val.replace(":", "").replace(",", "").replace("@", "").replace("'s", ""); // common nick prefixes/suffixes when chatting
+			const user = chan.findUser(cleanedNick);
```
xPaw: This doesn't sound like a great idea to do it the channel has thousands of users and its highly active.

MaxLeiter: Does the server have a concept of rendered messages on the client? I.e. when you click “show more”/on init? Could do it there

client/js/libs/handlebars/ircmessageparser/parseName.js
MaxLeiter: won't work if the suffix/prefix we're looking for is length > 1 (due to `indexOf(orig[0])`)

astorije:
@MaxLeiter, could you rebase on master please?

astorije:
@MaxLeiter, since #1712 was merged, I took the liberty to rebase that for you and fix conflicts (which were just dangling comma shit) so I can test it with most recent master.

src/models/msg.js
```diff
@@ -13,6 +13,7 @@ class Msg {
 			text: "",
 			type: Msg.Type.MESSAGE,
 			self: false,
+			users: [], // {start, end, nick}
```
xPaw: Hm, I don't think we need to set a default? `parse` should be able to deal with it just fine.

Also, it might be beneficial to rename this to `nicks` to be more specific.

astorije: I'm getting the following, which might be related to your comment:

```
lounge/src/models/chan.js:124
	return this.users.get(nick.toLowerCase());
	                          ^

TypeError: Cannot read property 'toLowerCase' of undefined
    at Chan.findUser (/Users/jeremie/Perso/lounge/src/models/chan.js:124:28)
    at Chan.getUser (/Users/jeremie/Perso/lounge/src/models/chan.js:128:14)
    at IrcClient.<anonymous> (/Users/jeremie/Perso/lounge/src/plugins/irc-events/topic.js:18:15)
    at IrcClient.emit (/Users/jeremie/Perso/lounge/node_modules/eventemitter3/index.js:129:35)
    at /Users/jeremie/Perso/lounge/node_modules/irc-framework/src/client.js:185:20
    at next (/Users/jeremie/Perso/lounge/node_modules/middleware-handler/index.js:42:18)
    at MiddlewareHandler.handle (/Users/jeremie/Perso/lounge/node_modules/middleware-handler/index.js:64:3)
    at IrcCommandHandler.<anonymous> (/Users/jeremie/Perso/lounge/node_modules/irc-framework/src/client.js:179:34)
    at IrcCommandHandler.emit (/Users/jeremie/Perso/lounge/node_modules/eventemitter3/index.js:130:35)
    at IrcCommandHandler.emit (/Users/jeremie/Perso/lounge/node_modules/irc-framework/src/commands/handler.js:95:23)
```

xPaw: Doesn't appear related to the PR.

astorije: It's not happening on `master`, so I'm guessing something gets passed wrong somewhere up the stack.

astorije:
@xPaw, did you see this? https://travis-ci.org/thelounge/lounge/jobs/304356874#L2268
I'm getting this when testing this PR locally as well.

client/js/libs/handlebars/parse.js
```diff
@@ -83,6 +91,8 @@ module.exports = function parse(text) {
 			return `<span class="inline-channel" role="button" tabindex="0" data-chan="${escapedChannel}">${fragments}</span>`;
 		} else if (textPart.emoji) {
 			return `<span class="emoji">${fragments}</span>`;
+		} else if (textPart.nick) {
+			return templates.user_name({nick: textPart.nick}).trim();
```
xPaw: Unfortunately to fix tests (at least right now), the easiest solution would be copying user_name template here.

Tests fail because .tpl files aren't being compiled when required from mocha.

Also probably should be using `fragments` here like the others above.

src/models/msg.js
MaxLeiter: yeah, this works fine without the default

client/js/libs/handlebars/parse.js
MaxLeiter: Done

astorije:
@MaxLeiter, @xPaw, I believe I have something to fix the failing test. I need to play more with that though to make it merge-ready. I'll try to do that after work tonight. Thus ignore it for now.

client/js/libs/handlebars/parse.js
```diff
@@ -83,6 +91,9 @@ module.exports = function parse(text) {
 			return `<span class="inline-channel" role="button" tabindex="0" data-chan="${escapedChannel}">${fragments}</span>`;
 		} else if (textPart.emoji) {
 			return `<span class="emoji">${fragments}</span>`;
+		} else if (textPart.nick) {
+			const nick = fragments.trim();
```
xPaw: Why is it trimmed?

client/js/libs/handlebars/parse.js
```diff
@@ -83,6 +91,9 @@ module.exports = function parse(text) {
...
+			return `<span role="button" class="user ${colorClass(nick)}" data-name="${nick}">${nick}</span>`;
```
xPaw: `fragments` may contain HTML, so you probably want to use `textPart.nick` in `colorClass`/`date-name` (and escape it?)

client/js/libs/handlebars/parse.js
MaxLeiter: Was an issue from before with handlebars, need to test if necessary now

client/js/libs/handlebars/parse.js
MaxLeiter: Don't think I need to escape?

MaxLeiter:
@Bonuspunkt could you take a look? This is broken with nicks, colors, bold, etc

Bonuspunkt:
i'm guessing rel #1718 ?
what are the expected results?

MaxLeiter:
Right now duplicates are produced (www.MaxLeiter.com with nick MaxLeiter in Chan renders www.MaxLeiter.comMaxLeiter. Each word is correct (link is a link, nick is nick))

astorije:
FYI, just tested it works fine with copy/paste. Can't wait for this feature!

xPaw:
@MaxLeiter Can you add some tests for this? Like `http://example.com` with ` example` being an user, and some others?

test/client/js/libs/handlebars/parse.js
```diff
@@ -243,6 +243,49 @@ describe("parse Handlebars helper", () => {
 		expect(actual).to.deep.equal(expected);
 	});
 
+	it("should find nicks", () => {
+		const testCases = [{
+			users: ["MaxLeiter"],
+			input: "test, MaxLeiter",
+			expected:
+				"test, " +
+				"<span role=\"button\" class=\"user color-12\" data-name=\"MaxLeiter\">" +
```
astorije: Can you use single quotes instead to avoid escaping? (See my PR I just merged)

xPaw: This was merged before yours. You should have enforced this with eslint rule by the way.

astorije:
Yeah sorry I was on phone and didn't realize it was merged.
I don't think there is a way to enforce it, just allow it, no?
-- 
Jérémie Astori
https://jeremie.astori.fr

PR #1712, opened 2017-11-16
Convert users list to map

xPaw:
Main reason is that `findUser` is now O(1) instead of being O(n).

This also no longer sorts user list on every update, only when the client requests the user list.

### DIFF
--- a/test/models/chan.js
+++ b/test/models/chan.js
@@ -1,12 +1,31 @@
 "use strict";
 
-var expect = require("chai").expect;
-
-var Chan = require("../../src/models/chan");
-var Msg = require("../../src/models/msg");
-var User = require("../../src/models/user");
+const expect = require("chai").expect;
+const Chan = require("../../src/models/chan");
+const Msg = require("../../src/models/msg");
+const User = require("../../src/models/user");
 
 describe("Chan", function() {
+	const network = {
+		network: {
+			options: {
+				PREFIX: [
+					{symbol: "~", mode: "q"},
+					{symbol: "&", mode: "a"},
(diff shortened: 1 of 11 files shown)


astorije:
@xPaw, I rebased your PR to fix conflicts. I really f*ed this PR with my #1711 😁

test/models/chan.js
```diff
@@ -27,89 +46,123 @@ describe("Chan", function() {
...
+			chan.setUser(new User({nick: "(foo)"}, prefixLookup));
+			chan.setUser(new User({nick: "{foo}"}, prefixLookup));
+			chan.setUser(new User({nick: "<foo>"}, prefixLookup));
+			chan.setUser(new User({nick: "_foo"}, prefixLookup));
+			chan.setUser(new User({nick: "@foo"}, prefixLookup));
+			chan.setUser(new User({nick: "^foo"}, prefixLookup));
+			chan.setUser(new User({nick: "&foo"}, prefixLookup));
+			chan.setUser(new User({nick: "!foo"}, prefixLookup));
+			chan.setUser(new User({nick: "+foo"}, prefixLookup));
+			chan.setUser(new User({nick: "Foo"}, prefixLookup));
```
astorije: For this and all others above, could you keep the condensed version instead?

```js
[
	"[foo", "]foo", "(foo)", "{foo}", "<foo>", "_foo", "@foo", "^foo",
	"&foo", "!foo", "+foo", "Foo",
].forEach((nick) => {
  chan.setUser(new User({nick: nick}, prefixLookup));
});

xPaw: Done.

Issue #533, opened 2016-07-29
Disable whois/messaging users when clicking their nicks


zachbr:
This is a UX change request, not a critical feature, however it is one of those "slight annoyance" things.

Clicking on a user's nickname should not open a new tab with their whois information. This is only slightly annoying on desktop, however on mobile it is infuriating. I would much prefer a more traditional, and much more explicit, context menu option to whois/message them, rather than just a single click, which seems a bit unexpected at best.

Thanks, enjoying the lounge overall.

EDIT: even an options toggle or something, just some acknowledgement that this may not be the best design. Nor should it assume I want to whois people. Why not autofill their name? or do nothing at all?

metsjeesus:
This is one of the small things what might need huge overhaul on system to fix it. I might be wrong though.

You can remove whois command, but after that it wont open private channel. Problem is, every channels has some ID on it, so client and server can figure out on what channel they exchange messages. So to create channel,  you have to initiate it somehow. Moment you either send whois, a message or get a private message to create a client-server pair of a channel. Whois does not send usually anything to the target channel, so they keep using it.

How to fix? Because channel is paired with a ID in client and server, you need to emit a message what does not push any IRC commands trough. Moment it seems this type of messages wont exist yet and probably needs huge overhaul to do it.

maxpoulin64:
It would actually be relatively easy to add, it's just currently not worth the time as we're busy on much more important things. PRs totally welcome to fix tho, we already have context menus in place. It's about just a matter of adding a command for it.

I'm not sure why people are annoyed by the whois tho. It doesn't hurt, just gives you a bit of info in the window but that's it. It's easy to ignore. I find it cool to be greeted with the person's details when opening a PM window.

Zbob750:
I don't really even want it to open a PM window, or do anything. Which is half of the issue.
May look into this at some point, but I'm not very familiar with the way this is written, or JS at all.

metsjeesus:
Imagine you got 2 interfaces to the same client, web and mobile. Generally you want to sync them up, so any action on either of those interfaces will be trasfered to all of them. Clicking on username moment is done with a easy solution. Clicking username send whois command to node server, what sends whois command to irc, gets result back and sends result to every interface connected. Works. 

So if you got 2 interfaces, what should happen on clicking on nicks? Because channels are saved in node server, should it create private channel on every interface? That means if you click on web interface on a nick, it creates channel in web interface and in mobile interface?

[...]

PR #1722, opened 2017-11-21
Add actions to user context menu

### LINKED ISSUE #1674: Clicking on a username in a query window outputs userinfo again as new messages.
## Steps 
1. Click on a user in a regular window for which no query window is open.
2. After the query window opens click on the username. 

## Expected 
Nothing happens as we already have a query window open so the conversation is already ongoing. 

## Actual 
The user information is shown again. 

image

## Impact 

I'd say the impact is rather low, it mostly confused me on mobile as it is easier to accidentally click on a user in chat there than on desktop. 
When this happens in a channel and it opens up a window with the user it is clear what happend. 
In this case when that window is already open it just is confusing, I personally thought at first that I found a bug with how the backscroll is loaded.

creesch:
Resolves #1674.

This adds: 

- Whois
- Conversation 

And if the lounge user is also Op in the channel it adds: 

- Kick 
- Ban 

image

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -224,6 +224,9 @@ kbd {
 .context-menu-chan::before { content: "\f292"; /* http://fontawesome.io/icon/hashtag/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 .context-menu-list::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
+.context-menu-action-whois::before { content: "\f05a"; /* http://fontawesome.io/icon/info-circle/  */ }
+.context-menu-action-query::before { content: "\f0e6"; /* http://fontawesome.io/icon/comments-o/ */ }
+.context-menu-action-kick::before { content: "\f05e"; /* http://fontawesome.io/icon/ban/ */ }
 
 .context-menu-network::before,
 #sidebar .chan.lobby::before,
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -79,9 +79,34 @@ $(function() {
 		if (target.hasClass("user")) {
 			output = templates.contextmenu_item({
 				class: "user",
+				action: "whois",
 				text: target.text(),
 				data: target.data("name"),
 			});
+			output += templates.contextmenu_divider();
+			output += templates.contextmenu_item({
+				class: "action-whois",
+				action: "whois",
+				text: "User information",
+				data: target.data("name"),
+			});
+			output += templates.contextmenu_item({
+				class: "action-query",
+				action: "query",
+				text: "Direct messages",
+				data: target.data("name"),
+			});
+
+			const channel = target.closest(".chan");
+			if (utils.isOpInChannel(channel) && channel.data("type") === "channel") {
+				output += templates.contextmenu_divider();
+				output += templates.contextmenu_item({
+					class: "action-kick",
+					action: "kick",
+					text: "Kick",
+					data: target.data("name"),
+				});
+			}
 		} else if (target.hasClass("chan")) {
 			let itemClass;
 
@@ -95,6 +120,7 @@ $(function() {
 
 			output = templates.contextmenu_item({
 				class: itemClass,
+				action: "focusChan",
 				text: target.data("title"),
 				data: target.data("target"),
 			});
@@ -102,12 +128,14 @@ $(function() {
 			if (target.hasClass("lobby")) {
 				output += templates.contextmenu_item({
 					class: "list",
+					action: "list",
 					text: "List all channels",
-					data: target.data("target"),
+					data: target.data("id"),
 				});
 			}
 			output += templates.contextmenu_item({
 				class: "close",
+				action: "close",
 				text: target.hasClass("lobby") ? "Disconnect" : target.hasClass("channel") ? "Leave" : "Close",
 				data: target.data("target"),
 			});
@@ -121,7 +149,11 @@ $(function() {
 		return false;
 	}
 
-	viewport.on("contextmenu", ".user, .network .chan", function(e) {
+	viewport.on("contextmenu", ".network .chan", function(e) {
+		return showContextMenu(this, e);
+	});
+
+	viewport.on("click contextmenu", ".user", function(e) {
 		return showContextMenu(this, e);
 	});
 
@@ -285,20 +317,6 @@ $(function() {
 		$(this).closest(".msg.condensed").toggleClass("closed");
 	});
 
-	chat.on("click", ".user", function() {
-		var name = $(this).data("name");
-		var chan = utils.findCurrentNetworkChan(name);
-
-		if (chan.length) {
-			chan.click();
-		}
-
-		socket.emit("input", {
-			target: chat.data("id"),
-			text: "/whois " + name,
-		});
-	});
-
 	sidebar.on("click", ".chan, button", function(e, data) {
 		// Pushes states to history web API when clicking elements with a data-target attribute.
 		// States are very trivial and only contain a single `clickTarget` property which
@@ -447,24 +465,60 @@ $(function() {
 		return false;
 	});
 
-	contextMenu.on("click", ".context-menu-item", function() {
-		switch ($(this).data("action")) {
-		case "close":
-			$(".networks .chan[data-target='" + $(this).data("data") + "'] .close").click();
-			break;
-		case "chan":
-			$(".networks .chan[data-target='" + $(this).data("data") + "']").click();
-			break;
-		case "user":
-			$(".channel.active .users .user[data-name='" + $(this).data("data") + "']").click();
-			break;
-		case "list":
+	const contextMenuActions = {
+		close: function(itemData) {
+			$(`.networks .chan[data-target="${itemData}"] .close`).click();
+		},
+		focusChan: function(itemData) {
+			$(`.networks .chan[data-target="${itemData}"]`).click();
+		},
+		list: function(itemData) {
 			socket.emit("input", {
-				target: chat.data("id"),
+				target: itemData,
 				text: "/list",
 			});
-			break;
-		}
+		},
+		whois: function(itemData) {
+			const chan = utils.findCurrentNetworkChan(itemData);
+
+			if (chan.length) {
+				chan.click();
+			}
+
+			socket.emit("input", {
+				target: $("#chat").data("id"),
+				text: "/whois " + itemData,
+			});
+
+			$(`.channel.active .users .user[data-name="${itemData}"]`).click();
+		},
+		query: function(itemData) {
+			const chan = utils.findCurrentNetworkChan(itemData);
+
+			if (chan.length) {
+				chan.click();
+			}
+
+			socket.emit("input", {
+				target: $("#chat").data("id"),
+				text: "/query " + itemData,
+			});
+		},
+		kick: function(itemData) {
+			socket.emit("input", {
+				target: $("#chat").data("id"),
+				text: "/kick " + itemData,
+			});
+		},
+	};
+
+	contextMenuActions.execute = (name, ...args) => contextMenuActions[name] && contextMenuActions[name](...args);
+
+	contextMenu.on("click", ".context-menu-item", function() {
+		const $this = $(this);
+		const itemData = $this.data("data");
+		const contextAction = $this.data("action");
+		contextMenuActions.execute(contextAction, itemData);
 	});
 
 	chat.on("input", ".search", function() {
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const $ = require("jquery");
+const escape = require("css.escape");
 const input = $("#input");
 
 var serverHash = -1;
@@ -19,6 +20,7 @@ module.exports = {
 	toggleNickEditor,
 	toggleNotificationMarkers,
 	requestIdleCallback,
+	isOpInChannel,
 };
 
 function findCurrentNetworkChan(name) {
@@ -35,6 +37,15 @@ function findCurrentNetworkChan(name) {
 
 function resetHeight(element) {
 	element.style.height = element.style.minHeight;
+}
+
+// Given a channel element will determine if the lounge user is Op in that channel
+function isOpInChannel(channel) {
+	const channelID = channel.data("id");
+	const network = $("#sidebar .network").has(`.chan[data-id="${channelID}"]`);
+	const ownNick = network.data("nick");
+	const isOP = channel.find(`.users .user-mode.op .user[data-name="${escape(ownNick)}"]`).length;
+	return isOP;
 }
 
 // Triggering click event opens the virtual keyboard on mobile
--- a/client/views/contextmenu_item.tpl
+++ b/client/views/contextmenu_item.tpl
@@ -1,3 +1,3 @@
-<li class="context-menu-item context-menu-{{class}}" data-action="{{class}}"{{#if data}} data-data="{{data}}"{{/if}}>
+<li class="context-menu-item context-menu-{{class}}" data-action="{{action}}"{{#if data}} data-data="{{data}}"{{/if}}>
 	{{text}}
 </li>
--- a/src/plugins/inputs/query.js
+++ b/src/plugins/inputs/query.js
@@ -48,5 +48,6 @@ exports.input = function(network, chan, cmd, args) {
 	this.emit("join", {
 		network: network.id,
 		chan: newChan.getFilteredClone(true),
+		shouldOpen: true,
 	});
 };


creesch:
Added ab80d640fc988a26d9f1f20cf4cc2309bb5251a3 this fixes everything in #1674 as it will now show the context menu when you click a user instead of doing a `whois`.  As far as UX goes this is more in line of other IRC clients and should make The Lounge somewhat easier to use. 

On #thelounge there was some discussion about the ban `action` since it now, lacking better information,  defaults to banning based on nickname. I'll remove that in a future commit.

client/js/lounge.js
```diff
@@ -72,6 +72,15 @@ $(function() {
 		return offset;
 	}
 
+	function isOpInChannel(channel) {
```
MaxLeiter: Could we move this to a new file? Doesn’t seen to fit into lounge.js

creesch: Move to utils.js maybe?

client/views/chan.tpl
```diff
@@ -1,5 +1,5 @@
 {{#each channels}}
-<div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
+<div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}} chanid-{{id}}">
```
xPaw: Why did you add this?

client/views/user_name.tpl
```diff
@@ -1 +1 @@
-<span role="button" class="user {{colorClass nick}}" data-name="{{nick}}">{{mode}}{{nick}}</span>
+<span role="button" class="user {{colorClass nick}} nick-{{nick}}" data-name="{{nick}}">{{mode}}{{nick}}</span>
```
xPaw: And this?

client/views/chan.tpl
creesch: To keep [this function]( https://github.com/creesch/lounge/blob/ab80d640fc988a26d9f1f20cf4cc2309bb5251a3/client/js/lounge.js#L75-L82) from turning into a convoluted spaghetti mess. 
All the information I needed was already available but I needed a reliable way to cross reference it.

client/views/user_name.tpl
creesch: Same as the other one.

client/views/chan.tpl
xPaw: You can already find it by `data-id`.

client/views/user_name.tpl
xPaw: Same here, you can find it by `data-name`.

client/views/chan.tpl
creesch: That doesn't allow me to find it in all instances as jquery `data()` is used after the client is initialized. So when a nickname is changed it might not match the attribute anymore. 

For channels it appears to be relatively safe to use an attribute selector but I could not be certain so opted for including a more reliable selector in both instances.

creesch:
@MaxLeiter moved the function to utils.js 
@xPaw changed the selectors to data-attributes as discussed on irc.

client/js/lounge.js
```diff
@@ -442,15 +453,49 @@ $(function() {
 	});
 
 	contextMenu.on("click", ".context-menu-item", function() {
-		switch ($(this).data("action")) {
+		const $this = $(this);
+		const thisData = $this.data("data");
+		let chan;
+		switch ($this.data("action")) {
 		case "close":
```
YaManicKill: I feel like we should be starting to use the command pattern with these actions now rather than just increasing the case statement over time.

client/js/utils.js
```diff
@@ -17,12 +17,22 @@ module.exports = {
...
 function resetHeight(element) {
 	element.style.height = element.style.minHeight;
 }
 
+// Given a channel element will determine if the lounge user is Op in that channel
+function isOpInChannel(channel) {
+	const channelID = channel.data("id");
+	const network = $("#sidebar .network").has(".chan[data-id=\"" + channelID + "\"]");
+	const ownNick = network.data("nick");
+	const isOP = channel.find(".users .user-mode.op .user[data-name=\"" + ownNick + "\"]").length;
```
xPaw: Use string interpolation and `CSS.escape` (`const escape = require("css.escape");` for `ownNick`.

client/js/lounge.js
```diff
@@ -82,6 +82,27 @@ $(function() {
...
 			});
+			output += templates.contextmenu_divider();
+			output += templates.contextmenu_item({
+				class: "action-whois",
+				text: "whois",
+				data: target.data("name"),
+			});
+			output += templates.contextmenu_item({
+				class: "action-query",
+				text: "conversation",
```
xPaw: Capitalise these strings.

client/js/lounge.js
creesch: Can you elaborate, not sure what you mean by this.

YaManicKill: I don't really know what to say to elaborate.

Instead of having a case statement going through all of these actions, which can get very messy and long, we should have a command for each of them that are registered with a click action. These can live in separate places. If you don't know what the command pattern is, go [read up on it](https://en.wikipedia.org/wiki/Command_pattern). It's a very useful design pattern to know.

creesch: Ah, gotcha. Was indeed not familiar with it.  So something like this then?

```Javascript
var contextActions = {
    close: function(thisData) {
        $(`.networks .chan[data-target="${thisData}"] .close`).click();
    },
    chan: function(thisData) {
        $(`.networks .chan[data-target="${thisData}"]`).click();
    },
    whois: function(thisData) {
        const chan = utils.findCurrentNetworkChan(thisData);

        if (chan.length) {
            chan.click();
        }

        socket.emit("input", {
            target: $("#chat").data("id"),
            text: "/whois " + thisData,
        });

        $(`.channel.active .users .user[data-name="${thisData}"]`).click();
    },
    query: function(thisData) {
        const chan = utils.findCurrentNetworkChan(thisData);

        if (chan.length) {
            chan.click();
        }

        socket.emit("input", {
            target: $("#chat").data("id"),
            text: "/query " + thisData,
        });
    },
    kick: function(thisData) {
        socket.emit("input", {
            target: $("#chat").data("id"),
            text: "/kick " + thisData,
        });
    }
};

contextActions.execute = function(name) {
    return contextActions[name] && contextActions[name].apply(contextActions, [].slice.call(arguments, 1));
};
```
Somewhere tucked away. And then instead of the switch statement do something like

```Javascript

contextMenu.on("click", ".context-menu-item", function() {
    const $this = $(this);
    const thisData = $this.data("data");
    const contextAction = $this.data("action");
    contextActions.execute(contextAction, thisData)
});
```

YaManicKill: Almost, yeah. That would be a good start (probably good enough  for this PR) but at some point each of the contextActions should be in a separate file (1 for kick, 1 for query, etc). But as there aren't a huge number of them, probably not necessary yet, but wouldb egood at some point.

creesch:
Did some searching and if accepted this would also fix #533 and partially #347.

MaxLeiter:
@creesch can you squash commits? 👏

client/js/utils.js
```diff
@@ -37,6 +39,15 @@ function resetHeight(element) {
 	element.style.height = element.style.minHeight;
 }
 
+// Given a channel element will determine if the lounge user is Op in that channel
+function isOpInChannel(channel) {
```
MaxLeiter: I wonder if this should be `function isOpInChannel(user, channel)`?

creesch: Doesn't make sense in the context of this commit as I specifically moved all of that (including the nick detection) to that function and would mean that I would have to move the first three lines back to [here](https://github.com/creesch/lounge/blob/46dfad0a0045ceede19718cc4859575a9675b7e6/client/js/lounge.js#L101)  where I explicitly removed it as to keep that area clean and readable.

creesch:
Sure can do. 

image

Done :)

client/css/style.css
```diff
@@ -220,12 +220,21 @@ kbd {
 #chat button.menu::before { content: "\f142"; /* http://fontawesome.io/icon/ellipsis-v/ */ }
 
 .context-menu-user::before { content: "\f007"; /* http://fontawesome.io/icon/user/ */ }
+
```
astorije: Could you remove this extra line?

client/css/style.css
```diff
@@ -220,12 +220,21 @@ kbd {
...
+.context-menu-action-query::before { content: "\f0e6"; /* http://fontawesome.io/icon/comments-o/ */ }
+
+.context-menu-action-kick::before { content: "\f05e"; /* http://fontawesome.io/icon/ban/ */ }
+
+
 #sidebar .chan.lobby::before,
 #chat .lobby .title::before { content: "\f0a0"; /* http://fontawesome.io/icon/hdd-o/ */ }
 
+
```
astorije: Extra line here too.

client/css/style.css
```diff
@@ -220,12 +220,21 @@ kbd {
...
 .context-menu-chan::before { content: "\f0f6"; /* http://fontawesome.io/icon/file-text-o/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 
+.context-menu-action-whois::before { content: "\f29c"; /* http://fontawesome.io/icon/question-circle-o/ */ }
+
+.context-menu-action-query::before { content: "\f0e6"; /* http://fontawesome.io/icon/comments-o/ */ }
+
+.context-menu-action-kick::before { content: "\f05e"; /* http://fontawesome.io/icon/ban/ */ }
+
+
```
astorije: Remove new lines between all those, also remove the one at R226.

client/js/lounge.js
```diff
@@ -79,18 +79,45 @@ $(function() {
...
+			output += templates.contextmenu_item({
+				class: "action-whois",
+				action: "whois",
+				text: "Whois",
+				data: target.data("name"),
+			});
+			output += templates.contextmenu_item({
+				class: "action-query",
+				action: "query",
+				text: "Conversation",
```
astorije: Let's do "Send a direct message".

client/js/utils.js
```diff
@@ -37,6 +39,15 @@ function resetHeight(element) {
...
+	const channelID = channel.data("id");
+	const network = $("#sidebar .network").has(`.chan[data-id="${channelID}"]`);
+	const ownNick = network.data("nick");
+	const isOP = channel.find(`.users .user-mode.op .user[data-name="${escape(ownNick)}"]`).length;
```
astorije: @xPaw, isn't there a better way to determine if the current user is an op? If I read this correctly, this reads the info from the HTML of the user list.

client/js/lounge.js
```diff
@@ -79,18 +79,45 @@ $(function() {
 		if (target.hasClass("user")) {
 			output = templates.contextmenu_item({
 				class: "user",
+				action: "whois",
 				text: target.text(),
 				data: target.data("name"),
 			});
+			output += templates.contextmenu_divider();
+			output += templates.contextmenu_item({
+				class: "action-whois",
```
astorije: Why are we duplicating this action? If this is for clarity sake (which would be a valid argument I think), then really we should make the first entry non-clickable. Visually different line for the name, à la Slack, minus the avatar picture:

screen shot 2017-12-10 at 23 43 08

client/js/lounge.js
```diff
@@ -79,18 +79,45 @@ $(function() {
...
+				action: "whois",
+				text: "Whois",
```
astorije: "Information about this user"

client/js/lounge.js
```diff
@@ -104,7 +131,11 @@ $(function() {
 		return false;
 	}
 
-	viewport.on("contextmenu", ".user, .network .chan", function(e) {
+	viewport.on("contextmenu", ".network .chan", function(e) {
+		return showContextMenu(this, e);
+	});
+
+	viewport.on("click contextmenu", ".user", function(e) {
```
astorije: I was very worried about the dropdown menu on left click, but it actually works very nicely. Nice job!
Especially like the fact that it's super snappy. On Slack it's not only slightly delayed but there is also an animation. Frustrating experience.

client/css/style.css
```diff
@@ -220,12 +220,21 @@ kbd {
 #chat button.menu::before { content: "\f142"; /* http://fontawesome.io/icon/ellipsis-v/ */ }
 
 .context-menu-user::before { content: "\f007"; /* http://fontawesome.io/icon/user/ */ }
+
 .context-menu-chan::before { content: "\f0f6"; /* http://fontawesome.io/icon/file-text-o/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 
+.context-menu-action-whois::before { content: "\f29c"; /* http://fontawesome.io/icon/question-circle-o/ */ }
```
astorije: Could you use http://fontawesome.io/icon/info-circle/ for this instead please? (`f05a`)

client/js/lounge.js
```diff
@@ -427,18 +444,56 @@ $(function() {
...
+		kick: function(itemData) {
+			socket.emit("input", {
+				target: $("#chat").data("id"),
+				text: "/kick " + itemData,
+			});
+		},
+	};
+
+	contextMenuActions.execute = function(name) {
+		return contextMenuActions[name] && contextMenuActions[name].apply(contextMenuActions, [].slice.call(arguments, 1));
```
astorije: Why the `.apply` and `[].slice.call`?
It seems to me you could add an `itemData` argument to the `execute` function and pass it to `contextMenuActions[name](itemData)`, no?

client/js/utils.js
MaxLeiter: this was discussed on IRC, that's basically the only way at the moment.

astorije: Gotcha. If there is a better way that could be possible when refactoring this or that, do we want to add a TODO note here?

client/js/lounge.js
creesch: I rather keep whois here, as that is the commonly known irc terminology.

client/js/lounge.js
creesch: That is misleading, you are effectively starting a conversation which is different from direct message functionality imho.

client/js/lounge.js
creesch: Left the duplicate activity because long time lounge users might expect that action. Also it doesn't hurt.

client/js/lounge.js
astorije: What about "Direct messages" then (like Slack)? "Conversation" kind of means everything and anything is my issue.

client/js/lounge.js
astorije: And I'd... rather not 😅
New users would have no idea what this is, while experienced users would feel free to type `/whois` anyway.
You could also do "Information about this user (`whois`)" if you reaaally want to keep "whois" in there. I'd be okay with that.

client/js/lounge.js
astorije: Well, it's going literally 5mm lower, so it's not too confusing either :)
I think we should explore better UI ideas for this, but it doesn't have to be done in your PR.

client/js/lounge.js
creesch: Simply future prooving and OOP oriented. If the handling method would change it would require all objects directly accessing these methods  to also be modified. By abstracting it away further like this we accept any named methods that can be performed on the object, passing along any data that might be used.

client/js/lounge.js
creesch: How about "private conversation"?

astorije: "Conversation" still carries this confusing meaning. I prefer "messages" over "conversation". "Private" or "Direct" work, though "private" kinda gives this vibe or privacy, i.e. OTR or something, which we don't do, so I prefer "Direct", but I'm fine with either.

client/js/lounge.js
astorije: Could you try this? It might work and is more explicit/simpler (also FP > OOP 😁):

```js
contextMenuActions.execute = (name, ...args) => contextMenuActions[name] && contextMenuActions[name](args);
```

creesch: Hrm I guess that could work, didn't think of using a rest parameter.

astorije: Actually, that would be `contextMenuActions[name](...args)` too, rather.

client/js/lounge.js
creesch: Would you be okay with just "User information"?  Fits much better with the width of the menu.

YaManicKill: How is "whois" confusing? It's literally saying it in the name what it is.

I'm really not a fan of these long sentences you want in the context menu.

How about "Who is this user?" or something? Then you get the whois in, but it's a bit more of a sentence for new users.

client/js/lounge.js
YaManicKill: What's the problem with Conversation?

client/js/lounge.js
YaManicKill: (I'd definitely be fine with "User Information", that works fine for me)

client/js/lounge.js
creesch: Still curious about this as well, for all intends and purposes you are starting a conversation as it does open the query window. Direct message more or less implies a fire and forget sort of thing more like mail or private messaging on forums. 

I don't care too strongly either way, matter of fact already changed it but am still curious.

creesch:
@astorije should be all of it. Just need to squash stuff....

client/css/style.css
```diff
@@ -224,6 +224,9 @@ kbd {
 .context-menu-chan::before { content: "\f292"; /* http://fontawesome.io/icon/hashtag/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 .context-menu-list::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
+.context-menu-action-whois::before { content: "\f05a"; /* http://fontawesome.io/icon/info-circle/  */ }
+.context-menu-action-query::before { content: "\f0e6"; /* http://fontawesome.io/icon/comments-o/ */ }
+.context-menu-action-kick::before { content: "\f05e"; /* http://fontawesome.io/icon/ban/ */ }
```
astorije: FYI, I just removed an extra line L237. No big deal, I just didn't want to bother you again with something minor.

client/js/lounge.js
astorije: It was 2am so maybe I wasn't clear, sorry :)
Issue with this is that "Conversation" is too broad. Actually, let me give you what I expected: when I first ran your PR, I had not looked at the code yet, I wanted to see the look-and-feel without a bias first. When I saw "Conversation", seeing that there was also "Whois" to open a window (with whois information, granted), I was totally expecting that it would bring the nick to the current conversation by auto-completing its name (it's not such a far-fetched idea, it had been brought up by a few people in the past).
I was miles away expecting it would open a new window without the whois :man_shrugging:

client/js/lounge.js
astorije: User information is great, thanks! And yeah, "whois" doesn't say much unless you already know what a whois is in the context of IRC. "User information" (I was also fine with "User information (whois)") is way clearer :D

Re: sentences, I totally understand. It's difficult to give a proper explanation in 3 words max, but yeah my suggestion was too long. Thanks all!

Issue #1718, opened 2017-11-20
Sending channel and link together generates duplicate text


xPaw:
Sending `#test-https://google.com` generates `#test-https://google.comhttps://google.com`

HTML:
```html
<span class="inline-channel" role="button" tabindex="0" data-chan="#test-https://google.com">#test-https://google.com</span><a href="https://google.com" target="_blank" rel="noopener">https://google.com</a><button class="toggle-button toggle-preview  opened" data-url="https://google.com" aria-label="Toggle website preview"></button>
```

cc @Bonuspunkt, any idea?

Bonuspunkt:
yes

- channel and url pattern matches.
- the simple concat is non sufficient for this case ( https://github.com/thelounge/lounge/blob/master/client/js/libs/handlebars/parse.js#L66-L68 )

what would be your expected result?

xPaw:
I guess simpler solution would be consider entire thing a channel and not detect the url (intersection?), but this still has to work for colors and stuff like that.

Bonuspunkt:
https://github.com/Bonuspunkt/ircmessageparser/commit/23986736f52444fc24a2f6d9bdf08eeb0fe2eb51

xPaw:
@Bonuspunkt Could you make a PR with that please?

PR #1723, opened 2017-11-21
Sending channel and link together generates duplicate text

Bonuspunkt:
#1718

### DIFF
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -2,6 +2,7 @@
 
 const Handlebars = require("handlebars/runtime");
 const parseStyle = require("./ircmessageparser/parseStyle");
+const anyIntersection = require("./ircmessageparser/anyIntersection");
 const findChannels = require("./ircmessageparser/findChannels");
 const findLinks = require("./ircmessageparser/findLinks");
 const findEmoji = require("./ircmessageparser/findEmoji");
@@ -68,9 +69,11 @@ module.exports = function parse(text) {
 		.concat(emojiParts)
 		.sort((a, b) => a.start - b.start || b.end - a.end)
 		.reduce((prev, curr) => {
-			const intersection = prev.some(p => anyIntersection(p, curr));
+			const intersection = prev.some((p) => anyIntersection(p, curr));
 
-			if (intersection) return prev;
+			if (intersection) {
+				return prev;
+			}
 			return prev.concat([curr]);
 		}, []);
 


client/js/libs/handlebars/parse.js
```diff
@@ -66,7 +66,13 @@ module.exports = function parse(text) {
 	const parts = channelParts
 		.concat(linkParts)
 		.concat(emojiParts)
-		.sort((a, b) => a.start - b.start);
+		.sort((a, b) => a.start - b.start || b.end - a.end)
+		.reduce((prev, curr) => {
+			const intersection = prev.some(p => anyIntersection(p, curr));
```
astorije: `p` ➡️ `(p)`

client/js/libs/handlebars/parse.js
```diff
@@ -66,7 +66,13 @@ module.exports = function parse(text) {
...
+
+			if (intersection) return prev;
```
astorije: Add `{}` and a new line.

client/js/libs/handlebars/parse.js
astorije: `anyIntersection` is not defined, you need to `require()` it.

MaxLeiter:
Doesn't completely work, `#chan-www.website.com` embeds a website and links as a chan (fine behavior, imo), but the preview arrow doesn't show to collapse. (`/collapse` also doesn't work)
screen shot 2017-11-21 at 8 16 05 pm

Bonuspunkt:
> Also, any chance this could use a new test case? 
no, because it seems you don't really care about it, since previous and upcoming changes doesn't have to include tests

xPaw:
@Bonuspunkt We try to add more tests as we go along, and since you already wrote a test for this it's not a terrible idea to include it.

@MaxLeiter That happens because server uses just `findLinks` and not the full `parse` function, not sure what we should do about that.

astorije:
> no, because it seems you don't really care about it, since previous and upcoming changes doesn't have to include tests

I'm not sure what led you to think that, because we did include tests in changes to the existing code (for example in https://github.com/thelounge/lounge/pull/1446 and https://github.com/thelounge/lounge/pull/1100, the only real 2 changes since your original work). We also try to ask people to add tests when possible, and particularly in this already-well-tested area (I can't find examples, but I know I've asked).

But if you noticed something different, please let us know and we will try to do better!

astorije:
Oh, you do have a test case for it, I didn't know before @xPaw mentioned it: https://github.com/Bonuspunkt/ircmessageparser/pull/12/files#diff-e6c9a6bca996bc454cc244d17bfeda5cR203
@Bonuspunkt, mind bringing it here? :)

astorije:
Hey, closing this PR in favor of https://github.com/thelounge/lounge/pull/1747 to attempt bringing the test. Since you deleted the repo, we can't edit the PR anymore :/
Thanks for your help!!

astorije:
Hey @Bonuspunkt, we have free sticker packs for our contributors now!

Issue #1724, opened 2017-11-22
LDAP Auth succeeds with blank password


keegan:
When a user inputs no password, but gives a valid username in LDAP, they can authenticate successfully.

This should be a very simple fix, this block:
```
if (!user) {
    return callback(false);
}
```
can simply be replaced by this:
```
if (!user || !password) {
    return callback(false);
}
```
in the beginning of the different LDAP auth methods

astorije:
@eliemichel, @thisisdarshan, @gramakri, @lindskogen, you've all been very helpful in the past regarding our LDAP support, could you advise on this? :)

gramakri:
I checked quickly in the Cloudron.io app and LDAP auth does not succeed with an empty password. You can try this on the demo at https://my-demo.cloudron.me (username: cloudron password: cloudron).

Quickly reading the code, it just passes on the password to the bind logic at https://github.com/thelounge/lounge/blob/master/src/plugins/auth/ldap.js#L19 . @keegan can you explain the code flow here on your LDAP server?

astorije:
> I checked quickly in the Cloudron.io app and LDAP auth does not succeed with an empty password.

@gramakri, there was a recent refactor of the LDAP support at https://github.com/thelounge/lounge/pull/1478. Is the Cloudron.io app running the previous version, or the current one?
Reason why I'm asking is because it might have been working before the refactor, then a bug could have been introduced in the refactor.

ollipa:
I can reproduce this on my instance which is running 2.6.0. Authentication fails when using an incorrect password but is successful if password field is left blank. Hopefully this gets fixed fast...

dgw:
Relevant? https://security.stackexchange.com/a/6714/601

Excerpt:

> For LDAP, however, the common case is that the server allows logins with **any username and an empty password**. The user will end up as being anonymous. But the login returns a status code indicating success.
> 
> […]
> 
> So, now imagine that the login is not done by the user directly. But the user logs into some other software, and this software takes the username and password and does an LDAP login behind the scenes.
> 
> If this software blindly trust the "success" result from the LDAP server **without having a special case handling for empty passwords, boom**. (emphasis in original)

As a side note, a Google search for "ldap password empty" turned up a fair number of tickets and discussions about similar issues in various software products and frameworks.

[...]

PR #1725, opened 2017-11-22
Stop LDAP Auth from succeeding without password

keegan:
Fixing issue #1724

### DIFF
--- a/src/plugins/auth/ldap.js
+++ b/src/plugins/auth/ldap.js
@@ -23,7 +23,7 @@ function ldapAuthCommon(user, bindDN, password, callback) {
 }
 
 function simpleLdapAuth(user, password, callback) {
-	if (!user) {
+	if (!user || !password) {
 		return callback(false);
 	}
 
@@ -41,7 +41,7 @@ function simpleLdapAuth(user, password, callback) {
  * LDAP auth using initial DN search (see config comment for ldap.searchDN)
  */
 function advancedLdapAuth(user, password, callback) {
-	if (!user) {
+	if (!user || !password) {
 		return callback(false);
 	}
 


astorije:
@eliemichel, @thisisdarshan, @gramakri, @lindskogen, pinging you here as well, sorry for the spam! 🙏

xPaw:
Sounds rather odd, isn't it your LDAP server that doesn't return an error if the password is empty?

Could you also add a test case for this change?

dgw:
It's default behavior for some LDAP servers to allow anonymous binds (`unauthenticated`, in spec terms). The spec also [specifically warns](https://tools.ietf.org/html/rfc4513#section-6.3.1) against clients trusting the LDAP server's bind result without checking if the password is empty:

> Operational experience shows that clients can (and frequently do) misuse the unauthenticated authentication mechanism of the simple Bind method (see Section 5.1.2).  For example, a client program might make a decision to grant access to non-directory information on the basis of successfully completing a Bind operation.  LDAP server implementations may return a success response to an unauthenticated Bind request.  This may erroneously leave the client with the impression that the server has successfully authenticated the identity represented by the distinguished name when in reality, an anonymous authorization state has been established.  Clients that use the results from a simple Bind operation to make authorization decisions should actively detect unauthenticated Bind requests (by verifying that the supplied password is not empty) and react appropriately.

tl;dr this check for empty passwords in LDAP authentication mode should be merged and released ASAP.

gramakri:
On one hand, I am tempted to call this a server configuration issue. As per http://www.openldap.org/doc/admin23/security.html, you can disable anonymous binds on the server side using `disallow bind_anon` in slapd.conf.

On the other hand, most people don't know of this feature and it's very dangerous. The change looks good but please add a comment as to why we are making that check. Maybe a link to https://tools.ietf.org/html/rfc4513#section-5.1.1

eliemichel:
Mmh, I wasn't aware of this being a common issue. My prod server does not allow empty passwords and the unit test code does not have this this problem neither. But I totally agree that if it is a typical error that administrators do, it is part of our responsibility to warn them/protect them.

To add a test for this, you'll simply have to `edit startLdapServer::authorize` in `test/plugins/auth/ldap.js` in order to make the fake test server behave like misconfigured one (i.e. authorizing empty password).

astorije:
Hey hey, does anyone want to help @keegan write a test for this? We can make sure commit authorship is kept if so :)

astorije:
@gramakri, @eliemichel, you've both been very helpful to us and our LDAP support. Would you like to be added to our GitHub organization and official reviewers for the LDAP integration?

You don't have to commit to code anything obviously, it's mainly to help us triage bugs, review/test PRs, etc. 🙏

gramakri:
@astorije Yeah, sure. I can help review LDAP related code. Thanks!

astorije:
@gramakri, awesome! I added you to the org 😊 

Does anyone want to help adding a test for this? 🙏

astorije:
@gramakri, would you be able to add a test for this case? 🙏 🙏 🙏

astorije:
Hey @keegan, we have sticker packs for our contributors now!

PR #1726, opened 2017-11-22
Instrument client code before running tests

astorije:
Despite being a bit gross to look at, this brings a few advantages:

- Tests are now closer to what actually runs, so more likely to find broken stuff.
- We can start using things that were so far Webpack-only or browser-only, like ES6 imports, loading Handlebars templates, etc.
- We open ourselves to browser testing (there is some work to do, but that would be a necessary step).
- We improve the client/server separation, by making it possible to run them independently

I do some extra steps around coverage: now we have 2 reports (client + server), so I have an extra step to combine them (the `nyc report` part). This is strictly to keep feature parity (the coverage report of this code is effectively the same as before), but in the near future, we might want to keep both reports separate, for example to continue separating client/server.
Another reason would be to use something like Codecov, which I believe has the ability to have multiple reports. This is down the road though, our coverage is not good enough to make hosting them somewhere be useful (I think).

A few other extras with this:

- Coverage summary is displayed when tests are run (this is not slowing down tests)
- Tests check for leaks (see https://mochajs.org/#--check-leaks)
- Tests now output with the `dot` reporter. This is nice as `npm test` runs in parallel, the whole output holds in a few lines instead of spanning over multiple screens.


.nycrc
```diff
@@ -1,13 +0,0 @@
...
-    "public/",
-    "test/",
-    "webpack.config.js"
-  ],
-  "reporter": [
-    "lcov",
-    "text",
-    "text-summary"
-  ]
-}
```
astorije: Because I have multiple of them now, I moved them to the `test/` folder.

client/js/libs/handlebars/parse.js
```diff
@@ -7,6 +7,9 @@ const findLinks = require("./ircmessageparser/findLinks");
 const findEmoji = require("./ircmessageparser/findEmoji");
 const merge = require("./ircmessageparser/merge");
 
+// FIXME: This is just for testing
+const templates = require("../../../views/chat.tpl"); // eslint-disable-line
```
astorije: This was temporary to force something that would break on `master`, see if tests pass here. I'll remove before calling this PR ready.

package.json
```diff
@@ -12,14 +12,17 @@
   },
   "homepage": "https://thelounge.github.io/",
   "scripts": {
-    "coverage": "nyc mocha",
+    "coverage": "rm -rf .nyc_output/ && npm-run-all coverage:* && cp test/.nycrc-report .nycrc && nyc report json && rm .nycrc",
+    "coverage:mocha": "cp test/.nycrc-mocha .nycrc && nyc npm run test:mocha && rm .nycrc",
+    "coverage:mocha-webpack": "cp test/.nycrc-mocha-webpack .nycrc && nyc npm run test:mocha-webpack && rm .nycrc",
```
astorije: Please ignore the stupid `cp`/`rm` dance for now. It will be removed soon. This is because at the moment, `.nycrc` path cannot be custom. Thankfully, this is just about to change: https://github.com/istanbuljs/nyc/pull/724

package.json
```diff
@@ -12,14 +12,17 @@
   },
   "homepage": "https://thelounge.github.io/",
   "scripts": {
-    "coverage": "nyc mocha",
+    "coverage": "rm -rf .nyc_output/ && npm-run-all coverage:* && cp test/.nycrc-report .nycrc && nyc report json && rm .nycrc",
```
astorije: The way this works, I tell `nyc` to not clean up temporary JSON files, empty the folder when starting up, and at the end build a report combining both. In the future, we might want to skip this step, and just maintain 2 reports. I'm debating doing it now or not.

test/.nycrc-mocha-webpack
```diff
@@ -0,0 +1,8 @@
+{
+  "all": true,
+  "include": [
+    "client/"
+  ],
+  "reporter": ["json"],
+  "clean": false
+}
```
astorije: Because these 2 are very similar, I was hoping to have only one, then specify the `exclude`/`include` at the CLI level in `package.json`, but for some reason I have not figured out, I could not make these work... :(

test/.nycrc-report
```diff
@@ -0,0 +1,7 @@
+{
+  "reporter": [
+    "lcov",
+    "text",
+    "text-summary"
+  ]
+}
```
astorije: Similarly, `"text-summary"` reporter does _not_ work when passed as a CLI argument (⁉️) so doing it in an extra file for now...

.travis.yml
```diff
@@ -18,6 +18,11 @@ cache:
 before_script:
   - NODE_ENV=$BUILD_ENV npm run build
```
astorije: I had removed this at first, thinking that all client stuff would be built as part of the `mocha-webpack` command,  but that gave me `Error: Cannot find module '../public/manifest.json'` 💩 

That means the server still relies on client-side build. Also that time saved on Travis CI won't be as nice. But surely that's something we can fix later :)

astorije:
Not sure why the `BUILD_ENV=production` version fails on Travis CI, I saw [this](https://github.com/webpack/webpack/issues/1612#issuecomment-160265914) but I don't believe this is causing it (because it is still running Babel). I'll have to investigate, but ideas welcome.

astorije:
@xPaw, any chance we could remove the dependency on `npm run build` from the server-side standpoint? It seems to me that [this](https://github.com/thelounge/lounge/blob/a10f147c3627605d25b693ecd4ab4ac32c6fc65e/src/helper.js#L82) is the only place that forces it.

Also, this PR means that we can now use `import`/`export` like https://github.com/thelounge/lounge/pull/973/files. Do you prefer switching to that as part of this PR or in a later one?

xPaw:
> any chance we could remove the dependency on npm run build from the server-side standpoint? It seems to me that this is the only place that forces it.

If you want to hard code the theme color in server code, then yes, it can be removed. The idea of that code is to allow overriding theme-color by themes in the future, and fallback to lounge default.

astorije:
Re: ES6 module commit, I was really excited as I was bringing this, but after converting stuff, I'm not so convinced we want that. On the other hand, I don't know if we want to stick with the `require` potentially bad practices we are doing.
I don't feel strongly on this, and either way it will be taken out of this PR, which already does too many things. No need for such a broad conversion at the same time.

client/js/libs/handlebars/parse.js
```diff
@@ -10,6 +10,9 @@ const findNames = require("./ircmessageparser/findNames");
 const merge = require("./ircmessageparser/merge");
 const colorClass = require("./colorClass");
 
+// FIXME: This is just for testing
+const templates = require("../../../views/chat.tpl"); // eslint-disable-line
```
astorije: ❌ Reminder to remove this before merging.

astorije:
@thelounge/maintainers, after waaaaaiting on dependencies to publish releases, this PR is _finally_ ready! 💦

Because of that wait, I ended up packing quite a few things in here, but I'm actually pretty happy with the result and, while not perfect, it's a step forward: it makes client testing much more comfortable (because it now webpacks before testing, so no more issue where tests could not run code that `require("[...].tpl")`), and I believe should eventually simplify browser testing (maybe browser tests in v3 :crossed_fingers:).

Anyway, there are indeed a few things, and I'm more than happy to walk you through confusing parts and add comments where necessary. Feel free to add questions in comments and I'll do my best to address them.

.travis.yml
```diff
@@ -17,6 +17,10 @@ cache:
...
 
+script:
+  # Running front-end tests on the oldest Node version we support
+  - if [ "${TRAVIS_NODE_VERSION}" = "4" ]; then npm test; else npm run travis; fi
```
xPaw: No v4 anymore

package.json
```diff
@@ -13,15 +13,18 @@
...
     "watch": "webpack --watch",
     "test": "npm-run-all --aggregate-output --parallel --continue-on-error test:* lint:*",
-    "test:mocha": "mocha --colors",
+    "travis": "npm-run-all --aggregate-output --parallel --continue-on-error test:mocha lint:*",
+    "test:mocha": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
+    "test:mocha-webpack": "nyc --nycrc-path=test/.nycrc-mocha-webpack mocha-webpack --colors --opts=test/mocha-webpack.opts",
     "lint:js": "eslint . --report-unused-disable-directives --color",
-    "lint:css": "stylelint --color \"client/**/*.css\""
+    "lint:css": "stylelint --color \"client/**/*.css\"",
+    "start:mocha": "webpack-dev-server --config=webpack.config-browser.js"
```
xPaw: This seems like an unrelated change?

.travis.yml
astorije: I've removed this, as other things caused it to bring nothing (we saved 4 seconds, yay...)

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- 9 # Current stable 
+- 9 # Current stable
```
xPaw: Want to skip this in the pr?

astorije: What do you mean? It's my editor which removed a trailing space on save, it's a pain to exclude every time...

xPaw: You just skip the file when you rebase, as there are no other changes in the file.

Yeah my editor does that too, but this file is untouched in this PR.

astorije: Done, thanks for pointing this out.

package.json
```diff
@@ -12,13 +12,14 @@
...
   "scripts": {
-    "coverage": "nyc mocha",
+    "coverage": "rm -rf .nyc_output/ && npm-run-all test:* && nyc --nycrc-path=test/.nycrc-report report",
     "start": "node index start",
     "start-dev": "npm-run-all --parallel watch start",
     "build": "webpack",
     "watch": "webpack --watch",
     "test": "npm-run-all --aggregate-output --parallel --continue-on-error test:* lint:*",
-    "test:mocha": "mocha --colors",
+    "test:mocha": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
```
xPaw: Why did you change normal tests to go through `nyc`? I tried `npm test` locally, and it replaces test output with coverage output (which is printed twice).

test/.nycrc-mocha
```diff
@@ -1,15 +1,16 @@
 {
   "all": true,
   "exclude": [
```
xPaw: Perhaps change this to `include: [ "src/" ]`?

package.json
astorije: I wanted coverage numbers to show up when running tests, it's a nice metric to have, and I removed all the noise that shows up with `npm run coverage` to keep it to the minimum when we run `npm test`.
Reason why it is printed twice is because client and server coverage are split, and `npm run coverage` is combining them for good measure.

I think it's pretty nice as is, and definitely worth it. I'd even prefer if it only printed one number in each coverage summary (so 1 number for client and 1 for server), but I don't know if `nyc`/Istanbul supports this.

Mind giving this a try, and we can revise/improve if this appears to not be optimal?

test/.nycrc-mocha
astorije: Yeah, I didn't want to miss files / future additions though.
If it works, what about `include src/ and /*.js (at root level) + exclude webpack.config*.js`?

astorije: @xPaw, I've done that in 47eb0382. I'm not convinced of the result, you tell me which one you prefer and I'm fine either way. I guess it's easier to miss with a whitelist than an exclude-list, but no big deal.

Issue #1663, opened 2017-10-25
Drop node v4 support for v3.0


xPaw:
Increasing minimum supported version to v6 brings many new JS features: https://blog.risingstack.com/whats-new-in-node-v6/

v4 is currently in maintenance, while v6 is active LTS.

PR #1727, opened 2017-11-22
Switch to Node v6 as the minimal supported version

astorije:
Follow-up of https://github.com/thelounge/lounge/pull/1715. Trying to be a good citizen and open the PRs that get rid of the deprecations as I add deprecations 😁
Resolves https://github.com/thelounge/lounge/issues/1663.

First commit is mainly to look nice and consistent. This is how the warning looks with this PR, assuming everything is fine:

screen shot 2017-11-21 at 22 51 42

When there is an issue, it gracefully degrades:

Problem with logging | Problem with colors | Problem with both
--- | --- | ---
screen shot 2017-11-21 at 22 52 10 | screen shot 2017-11-21 at 22 52 32 | screen shot 2017-11-21 at 22 52 55

---

The last commit is a non-exhaustive set of changes to things that Node v6 brings.
@xPaw, if you know or notice places where an ES6 feature not supported in Node v4 would be better, then this commit is meant for that. That way, we can already get better alternative without having to wait and remember.
Until v3 gets out, that leaves some time to make this part better. I'll do the rebases/fix conflicts when it will be necessary.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -42,13 +42,17 @@ rules:
   no-trailing-spaces: error
   no-unsafe-negation: error
   no-useless-computed-key: error
+  no-useless-constructor: error
   no-useless-return: error
   no-use-before-define: [error, {functions: false}]
   object-curly-spacing: [error, never]
   padded-blocks: [error, never]
   prefer-const: error
+  prefer-rest-params: error
+  prefer-spread: error
   quote-props: [error, as-needed]
   quotes: [error, double, avoid-escape]
+  rest-spread-spacing: error
   semi-spacing: error
   semi-style: [error, last]
   semi: [error, always]
--- a/index.js
+++ b/index.js
@@ -5,14 +5,31 @@
 process.chdir(__dirname);
 
 // Perform node version check before loading any other files or modules
-// Doing this check as soon as possible allows us to avoid ES6 parser errors or other issues
+// Doing this check as soon as possible allows us to avoid ES6 parser errors or
+// other issues
+// Try to display warnings nicely, but gracefully degrade if anything goes wrong
 var pkg = require("./package.json");
 if (!require("semver").satisfies(process.version, pkg.engines.node)) {
-	/* eslint-disable no-console */
-	console.error("=== WARNING!");
-	console.error("=== The oldest supported Node.js version is", pkg.engines.node);
-	console.error("=== We strongly encourage you to upgrade, see https://nodejs.org/en/download/package-manager/ for more details\n");
-	/* eslint-enable no-console */
+	let colors;
+	let log;
+
+	try {
+		colors = require("colors/safe");
+	} catch (e) {
+		colors = {};
+		colors.green = colors.red = colors.bold = (x) => x;
+	}
+
+	try {
+		log = require("./src/log");
+	} catch (e) {
+		log = {};
+		log.warn = (msg) => console.error(`[WARN] ${msg}`); // eslint-disable-line no-console
+	}
+
+	log.warn(`The Lounge requires Node.js ${colors.green(pkg.engines.node)} (current version: ${colors.red(process.version)})`);
+	log.warn(colors.bold("We strongly encourage you to upgrade Node.js"));
+	log.warn("See https://nodejs.org/en/download/package-manager/ for more details");
 }
 
 require("./src/command-line");
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@ process.chdir(__dirname);
 // Perform node version check before loading any other files or modules
 // Doing this check as soon as possible allows us to avoid ES6 parser errors or
 // other issues
-// Try to display warnings nicely, but gracefully degrade if anything goes wrong
+// Try to display messages nicely, but gracefully degrade if anything goes wrong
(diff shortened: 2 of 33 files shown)


index.js
```diff
@@ -5,14 +5,31 @@
 process.chdir(__dirname);
 
 // Perform node version check before loading any other files or modules
-// Doing this check as soon as possible allows us to avoid ES6 parser errors or other issues
+// Doing this check as soon as possible allows us to avoid ES6 parser errors or
+// other issues
+// Try to display warnings nicely, but gracefully degrade if anything goes wrong
```
xPaw: I would just add `process.exit(1)` if the version is not satisfied.

package.json
```diff
@@ -35,7 +35,7 @@
   ],
   "license": "MIT",
   "engines": {
-    "node": ">=4.2.0"
+    "node": ">=6"
```
xPaw: I looked through the news, and LTS starts at 6.9.0. https://nodejs.org/en/blog/release/v6.9.0/

YaManicKill: We don't want to specify the LTS version as the one we require. I'm pretty sure we required 4.2.0 because it had a feature that we needed, didn't it?

It feels silly to change our requirements based on LTS and not "what we require to run".

xPaw: Aren't we basing our support on LTS versions?

YaManicKill: I think we are saying that we will definitely support the latest LTS version, but why break earlier versions if we don't have a reason to? So, if there's a feature that 6 gives us that we want to use, then that's fine, but if we aren't using something that isn't in 4, there's no point in crashing out (regarding your comment about just exiting). And so, what's the point in saying we only support 6.9.0 if there's no functional difference that we don't use with 6.0.0? People should obviously make sure their version is up to date with 6.9.0, but ... it seems arbitrary to say "6.9.0" unless there's a feature in that that isn't in earlier 6 versions that we use.

xPaw: That's fair.

index.js
astorije: I thought about it, wondered why we didn't do that already. I guess it comes down to: do we keep the third commit or not?
If so, then yes, app is broken under Node 4, so might as well `exit`.
If not, then it will be compatible until it's not, and we don't have to `exit`.

I am fine with exiting, and we have until v3 to change our mind :)

xPaw: I'd rather exit because I wouldn't want it to crash randomly because we make use of a newer feature.

astorije: Done:

screen shot 2017-11-22 at 15 56 05

Switched to errors instead of warnings.

package.json
astorije: Yeah, so the way I understand Node's LTSes, 6.0 to 6.8 were released during before they entered the LTS schedule, and 6.9+ were released during the LTS schedule. But overall, v6 is the LTS version.

But yeah, if The Lounge is compatible with 6.0 to 6.8, we might as well not require 6.9. I see at https://github.com/thelounge/lounge/pull/280 that we already had this discussion lol, but the key difference is that v4.2 was released 1 month after 4.0, while there is a 6-month gap between 6.0 and 6.9.
This kinda tells me we need an extra build for `6.0` (on top of `6`, the latest LTS), but meh.

Anyway, I think 6.0.0 is reasonable, and we can bump this if The Lounge is not compatible with this one.

index.js
```diff
@@ -5,14 +5,33 @@
...
 if (!require("semver").satisfies(process.version, pkg.engines.node)) {
-	/* eslint-disable no-console */
-	console.error("=== WARNING!");
-	console.error("=== The oldest supported Node.js version is", pkg.engines.node);
-	console.error("=== We strongly encourage you to upgrade, see https://nodejs.org/en/download/package-manager/ for more details\n");
-	/* eslint-enable no-console */
+	let colors;
+	let log;
+
+	try {
```
xPaw: I'd really just keep using `console.error` here, I don't see the need to have this elaborate code just to print a message that node needs to be updated.

astorije: Ooh come on, I had fun doing it, don't take away my fun! :)
I made it extra safe to pass your gates! 🙏

index.js
YaManicKill: Well, the third commit shouldn't be part of this PR. It has nothing to do with setting our recommended version of node, it's changing the actual code. I don't have anything against it in general, but I don't think it should be in this PR.

astorije: I can move this commit to another PR when the Node setting is ready to go (i.e. when we're prepping for v3), but until then I'll keep it in there if you don't mind, as it's easier to test without having to duplicate the other commits, and I don't want to create yet another noisy PR in the list 😅

.eslintrc.yml
```diff
@@ -42,13 +42,17 @@ rules:
   no-trailing-spaces: error
   no-unsafe-negation: error
   no-useless-computed-key: error
+  no-useless-constructor: error
   no-useless-return: error
   no-use-before-define: [error, {functions: false}]
   object-curly-spacing: [error, never]
   padded-blocks: [error, never]
   prefer-const: error
+  prefer-rest-params: error
```
astorije: This will cause an ESLint failure, on code that is removed in #1760.

PR #1717, opened 2017-11-19
Deprecate `LOUNGE_HOME` env var in favor of `THELOUNGE_HOME`, `.lounge_home` file in favor of `.thelounge_home`

astorije:
This is one more step towards uniformization of lounge/thelounge due with The Lounge v3.
Hopefully I did not introduce any breaking changes.

screen shot 2017-11-19 at 13 13 42
screen shot 2017-11-19 at 13 18 59
screen shot 2017-11-19 at 13 19 49

PR #1734, opened 2017-11-22
Switch default home location to `~/.thelounge`

astorije:
This is a follow-up of #1717 meant to actually change the location. This is not a deprecation, but definitely a breaking change, so making it a v3 PR.


astorije:
@xPaw, I've added a commit to this PR to stop reading `.lounge_home`. You might want to give it another review. I thought I did that in another of the v3 PRs, but apparently not.

PR #1739, opened 2017-11-23
Do not keep scroll to bottom in inactive channels

xPaw:
Should prevent unnecessary and expensive layout/paint triggers.

This needs to keep `handleImageInPreview` logic, but only call `keepToBottom` for active channel.


astorije:
This is supposed to be a no-op for the user, right? Meaning, when testing this, I should not expect any visual change. If so then this acts as expected but I did very little testing, feel free to do more before merging.

PR #1742, opened 2017-11-25
Call callback on fetch error

xPaw:
With storage enabled, it was possible for thumbnails to retain their original urls if fetch failed (usually a timeout).


MaxLeiter:
👍

PR #1681, opened 2017-11-02
Include all available whois info

creesch:
This fixes #560 by including all the information [irc-framework provides](https://github.com/kiwiirc/irc-framework/blob/16728e3165d04afce5502ec6c8cd848638b2c4b7/docs/events.md). 

Specifically it adds the following information when available: 

- actual host (used in cases of webirc)
- registered nick
- modes
- operator
- helpop
- bot 
- logon time

PR #1743, opened 2017-11-25
Fix whois template parsing real name field

MaxLeiter:
Links and etc will now render correctly


astorije:
Good catch, this was done in https://github.com/thelounge/lounge/pull/1689 but lost in https://github.com/thelounge/lounge/pull/1681/files.

PR #1750, opened 2017-11-26
Parallelize `npm test`

astorije:
### Parallelize `npm test`

After this, tests run in ~7s locally vs. ~14s without. ¯\\\_(ツ)_/¯
One caveat is that coloring must be forced now. See why [here](https://github.com/mysticatea/npm-run-all/blob/HEAD/docs/npm-run-all.md#known-limitations).
~Let's hope this looks okay on Travis.~ It does.

### Make stylelint report unnecessary disables
 
From the man:

```
    --report-needless-disables, --rd

      Report stylelint-disable comments that are not blocking a lint warning.
      If you provide the argument "error", the process will exit with code 2
      if needless disables are found.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "test": "npm-run-all --aggregate-output --parallel --continue-on-error test:* lint:*",
     "test:mocha": "mocha --colors",
     "lint:js": "eslint . --report-unused-disable-directives --color",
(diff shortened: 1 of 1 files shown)


astorije:
Result looks fine on Travis, and on AppVeyor... that weird bug that says mocha path can't be found anyway (but the test report does show all test passing) so who cares. I think it's a decent addition.

PR #1760, opened 2017-11-27
Get rid of `Object.assign` polyfill

astorije:
As far as I know, this is only required for IE at this point, and we have already made changes that removed support of any version of IE.

Not sure if this is a v2.7.0 or v3.0.0 PR. @xPaw, what do you think?

### DIFF
--- a/client/js/libs/handlebars/ircmessageparser/merge.js
+++ b/client/js/libs/handlebars/ircmessageparser/merge.js
@@ -3,26 +3,13 @@
 const anyIntersection = require("./anyIntersection");
 const fill = require("./fill");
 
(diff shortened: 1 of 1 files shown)


MaxLeiter:
I don’t see any need for this to be in 3.0 if it’s not a new deprecation (as in, IE is already not supported)

astorije:
I think @xPaw put it under v3 "just in case", but I might be wrong. I would really like to do browser testing before going v3 as this specific change could be easily confirmed (it's on a tested path), but it's unlikely I'll be able to get to that soon enough. If someone wants to give it a try... :)

PR #1761, opened 2017-11-27
Mark slow tests as such to reduce noise on test report

astorije:
This is somewhat related to another PR I'm preparing (#1726) but makes sense in its own PR.

Before | After
--- | ---
screen shot 2017-11-27 at 18 48 10 | screen shot 2017-11-27 at 18 48 50

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -41,7 +41,9 @@ describe("Server", () => {
 		});
 	});
 
-	describe("WebSockets", () => {
+	describe("WebSockets", function() {
+		this.slow(300);
+
 		let client;
 
 		before((done) => {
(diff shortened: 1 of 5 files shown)


test/server.js
```diff
@@ -41,7 +41,9 @@ describe("Server", () => {
 		});
 	});
 
-	describe("WebSockets", () => {
+	describe("WebSockets", function() {
```
astorije: This is related to https://mochajs.org/#arrow-functions.
It should be fixed across the repo, but it will make sense to do this once https://github.com/lo1tuma/eslint-plugin-mocha/issues/141 gets implemented (and we use that plugin).

PR #1764, opened 2017-11-28
Fix data-from on messages

xPaw:


client/views/msg.tpl
```diff
@@ -1,4 +1,4 @@
-<div class="msg {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}}" id="msg-{{id}}" data-time="{{time}}" data-from="{{from}}">
+<div class="msg {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}}" id="msg-{{id}}" data-time="{{time}}"{{#if from}} data-from="{{from.nick}}"{{/if}}>
```
astorije: In what situation can `from|` be null, compared to before?

astorije: Or does this mean that we had some `data-from="null"` before?

astorije: (Note that if https://github.com/thelounge/lounge/pull/1773 makes it, `{{#if from}}` will need to be changed to `{{#if from.id}}`)

astorije: Let's go with `{{#if from.nick}}` once I update #1773.

PR #1767, opened 2017-11-28
Fix nicks and modes for messages being unintentionally updated later on

### LINKED ISSUE #1765: Nick and modes for users when a message is sent, isn't kept
Everything looks fine if lounge is open while receiving messages, but after a reload, messages display the user's *current* mode and nick, even if they didn't have that nick or mode when the message was sent.

Running master (c9e340e)

before reload:

after reload:

PolarizedIons:
Fixes #1765

I probably would have just stuck in a `_.clone` in there, but @xPaw made the point that it should probably only have those values.
```
<PolarizedIons>  hmm :\ thought it was an easy fix, and you could just put in a _.clone in there, but then modes don't get updated
<+xPaw>  Msg has to store a clone, yes
<+xPaw>  but getUser must not return one
<+xPaw>  and ideally with just 2 keys, nick+mode, and not whatever else is in user obj
```

I had to do the check that it is an object, because if `from` is a string, and I `_.pick` from it, it would return an empty object, which would crash on the client when trying to render the (undefined) nickname.

I did a quick test to see if it was working, and it doesn't look like the other fields in `from` were being used

### DIFF
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -30,3 +30,10 @@ User.prototype.toJSON = function() {
 		lastMessage: this.lastMessage,
 	};
 };
+
+User.prototype.value = function() {
+	return {
+		nick: this.nick,
+		mode: this.mode,
+	};
+};
--- a/src/plugins/irc-events/away.js
+++ b/src/plugins/irc-events/away.js
@@ -18,7 +18,7 @@ module.exports = function(irc, network) {
 				type: away ? Msg.Type.AWAY : Msg.Type.BACK,
 				text: away || "",
 				time: data.time,
-				from: user,
+				from: user.value(),
 			});
 
 			chan.pushMessage(client, msg);
--- a/src/plugins/irc-events/ctcp.js
+++ b/src/plugins/irc-events/ctcp.js
@@ -15,7 +15,7 @@ module.exports = function(irc, network) {
 		const msg = new Msg({
 			type: Msg.Type.CTCP,
 			time: data.time,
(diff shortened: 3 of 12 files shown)


dgw:
CI checks are failing because of code style. @PolarizedIons, change the single-quoted strings you added to double quotes and delete one of the spaces before the `//` line comment, and it should be good to go.

PolarizedIons:
xPaw further pointed out that other fields in the `Msg` object could also be `User` objects, not just the `from` key (eg. the invite message has an `invited` user key), so I updated this completely. 

Not the biggest fan of the `User.prototype.value` function's name, but that's the best we could come up with

---

Thanks, @dgw. I completely changed the pr so those are mute now, but I forgot to run tests anyway :man_shrugging:

astorije:
Also, I wonder if we could use the spread operator for this `...attr.from` or if it's missing from Node v4 (I think one form is available, one is not)

dgw:
Spread operator is available in Node v4 [with a flag](http://node.green/#ES2015-syntax-spread-------operator).

astorije:
Hey @PolarizedIons, so after your attempt, I decided to give this a shot in #1771. I believe these effectively both fix the underlying issue, but I think it's safer that way.
Let me know if you see anything obvious that I missed (and yeah, I went on cloning the whole object rather than picking nick and mode).

Thanks a lot for bringing this up!!!

PolarizedIons:
superseded by #1771

PR #1769, opened 2017-11-29
Increase timeout of server tests

astorije:
I've noticed this in https://github.com/thelounge/lounge/pull/1726 and thought it was my fault, but `master` actually has that issue from time to time, and so did https://github.com/thelounge/lounge/pull/1767.

I don't know if we want to be okay with the tests timing out or not, so here it is. Also, I didn't want to enable this globally for all tests, because it's only about one test right now. I'd rather avoid allow this globally :man_shrugging:

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -5,7 +5,8 @@ const expect = require("chai").expect;
 const request = require("request");
 const io = require("socket.io-client");
 
-describe("Server", () => {
+describe("Server", function() {
+	this.timeout(5000);
 	let server;
 
 	before(() => {


xPaw:
I don't like the idea of increasing timeout here, that's just hiding the problem.

astorije:
Absolutely agree in general. In this case, I figured this was like network-related or something, as we do start a server. I can see a few good reasons why this would range from 1 to 4 seconds to start (network, FS, Travis CPU, etc.).

@xPaw, feel free to close if you think this isn't right or if you know a better way to do that :) I agree, just don't know what can be done.

xPaw:
I don't think it hurts anyhow if the test is detect as slow? If not, just close this.

astorije:
It hurts when the server takes > 2s to start, CI fails. Unless I misunderstood your comment.

xPaw:
Oh, that was the issue here. In that case I'll merge it, but we probably should look into why that happens.

astorije:
> but we probably should look into why that happens.

Agreed. Can you open an issue for it?

PR #1771, opened 2017-11-29
Clone instances of `User` in `Msg` to avoid unintentional mutations

### LINKED ISSUE #1765: Nick and modes for users when a message is sent, isn't kept

astorije:
Fixes #1765.

This supersedes #1767. I find this way preferable because we have better ways to ensure this security. Also, test case!

Ah, if only we were using https://github.com/facebook/immutable-js/ :trollface:

### DIFF
--- a/src/models/msg.js
+++ b/src/models/msg.js
@@ -6,6 +6,18 @@ var id = 0;
 
 class Msg {
 	constructor(attr) {
+		// Some properties need to be copied in the Msg object instead of referenced
+		if (attr) {
+			["from", "target"].forEach((prop) => {
+				if (attr[prop]) {
+					this[prop] = {
+						mode: attr[prop].mode,
+						nick: attr[prop].nick,
+					};
+				}
+			});
+		}
+
 		_.defaults(this, attr, {
 			from: "",
 			id: id++,
--- a/client/views/actions/invite.tpl
+++ b/client/views/actions/invite.tpl
@@ -3,7 +3,7 @@ invited
 {{#if invitedYou}}
 	you
 {{else}}
-	{{> ../user_name invited}}
+	{{> ../user_name target}}
 {{/if}}
 to
 {{{parse channel}}}
--- a/src/plugins/irc-events/invite.js
+++ b/src/plugins/irc-events/invite.js
@@ -16,7 +16,7 @@ module.exports = function(irc, network) {
 			type: Msg.Type.INVITE,
 			time: data.time,
 			from: chan.getUser(data.nick),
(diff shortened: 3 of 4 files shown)


src/models/msg.js
```diff
@@ -6,8 +6,17 @@ var id = 0;
...
+			["from", "invited", "target"].forEach((prop) => {
+				if (attr[prop]) {
+					this[prop] = _.cloneDeep(attr[prop]);
+				}
+			});
+		}
+
 		_.defaults(this, attr, {
-			from: "",
+			from: {},
```
astorije: This should fix some other edge cases we missed IMO.

astorije:
@PolarizedIons, could you test this PR against the bug you reported in #1765?

src/models/msg.js
```diff
@@ -6,6 +6,15 @@ var id = 0;
 
 class Msg {
 	constructor(attr) {
+		// Some properties need to be copied in the Msg object instead of referenced
+		if (attr) {
+			["from", "invited", "target"].forEach((prop) => {
```
xPaw: If we're going this route, change invited to target in invite messages so that it's somewhat consistent.

src/models/msg.js
```diff
@@ -6,6 +6,15 @@ var id = 0;
...
+				if (attr[prop]) {
+					this[prop] = _.cloneDeep(attr[prop]);
+				}
```
xPaw: As with previous PR, store only nick and mode because we don't want to copy all the other unnecessary stuff (away message etc)

PolarizedIons:
I can confirm this also fixes the issue. I have no problems that if want to take over the issue :) I guess I'll close mine in the meantime.

src/models/msg.js
astorije: Gotcha, will do.

src/models/msg.js
astorije: Fair enough. I figured it would be better for future sake to store the whole object, but I guess we don't have to at the moment.

src/models/msg.js
astorije: Done! Would you mind testing a bit to make sure I didn't break invite stuff? I can't seem to be able to display "xxx invited yyy", only "xxx invited you".

src/models/msg.js
astorije: Done.

PR #1772, opened 2017-11-29
Fix nick changes wrongly reported

astorije:
Before | After
--- | ---
screen shot 2017-11-29 at 01 41 36 | screen shot 2017-11-29 at 01 40 28

⚠️ Note that this fix will only work after #1771 gets merged. The bug this PR fixes is due to `user.nick` being set too early **and** `msg.from` being modified after instantiating `msg`.

### DIFF
--- a/src/plugins/irc-events/nick.js
+++ b/src/plugins/irc-events/nick.js
@@ -32,10 +32,6 @@ module.exports = function(irc, network) {
 				return;
 			}
 
-			chan.removeUser(user);
-			user.nick = data.new_nick;
-			chan.setUser(user);
-
 			msg = new Msg({
 				time: data.time,
 				from: user,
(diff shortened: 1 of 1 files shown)


src/plugins/irc-events/nick.js
```diff
@@ -32,20 +32,19 @@ module.exports = function(irc, network) {
...
 				new_nick: data.new_nick,
 				self: self,
 			});
-			chan.pushMessage(client, msg);
 
+			chan.removeUser(user);
 			user.nick = data.new_nick;
+			chan.setUser(user);
+
+			chan.pushMessage(client, msg);
```
xPaw: You could keep this line where it was.

astorije: Done!

MaxLeiter:
I wonder if we want to do something to link old nicks to new nicks though? I.e. in `astorije2 is now known as astorije3` if we could link `astorije2` to `astorije3`

xPaw:
@MaxLeiter I'd rather not, you'd have to link further nick changes, and if user rejoins the channel the whole thing falls apart.

PR #1778, opened 2017-11-30
Heavily improve performance of "init" event

### LINKED ISSUE #1145: Lazily send channel messages on login (page load)
Currently they are all sent in the `init` event. But we could change init event to only send messages for active channel, and then send `more` buffer for other channels with some delay.

### LINKED ISSUE #1575: "init" does heavy cloning on the server
Every chan object is cloned in `toJSON`. And on reconnection the entire network is deep cloned to implement lastMessage.

https://github.com/thelounge/lounge/blob/da31317156047000819fa0363c435005104aa572/src/server.js#L435-L437
https://github.com/thelounge/lounge/blob/da31317156047000819fa0363c435005104aa572/src/models/chan.js#L129-L130

xPaw:
- Fixes #1145: Send 100 messages for active channel, and 1 message for all others on init.
- Fixes #1575: Do not perform heavy cloning of network and channel objects on init (especially when reconnecting)

Getting rid of lodash methods allows us to fine tune performance instead of blindly cloning entire object and then dropping values we don't need (heavy offenders being users and message arrays).

### DIFF
--- a/client/views/chat.tpl
+++ b/client/views/chat.tpl
@@ -12,7 +12,7 @@
 		<span title="{{topic}}" class="topic">{{{parse topic}}}</span>
 	</div>
 	<div class="chat">
-		<div class="show-more {{#equal messages.length 100}}show{{/equal}}">
+		<div class="show-more{{#if messages.length}} show{{/if}}">
 			<button class="show-more-button" data-id="{{id}}">Show older messages</button>
 		</div>
 		<div class="messages"></div>
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -139,7 +139,7 @@ Chan.prototype.removeUser = function(user) {
 Chan.prototype.toJSON = function() {
 	var clone = _.clone(this);
 	clone.users = []; // Do not send user list, the client will explicitly request it when needed
-	clone.messages = clone.messages.slice(-100);
+	clone.messages = clone.messages.slice(-1);
 	return clone;
 };
 
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -136,11 +136,39 @@ Chan.prototype.removeUser = function(user) {
 	this.users.delete(user.nick.toLowerCase());
 };
 
-Chan.prototype.toJSON = function() {
-	var clone = _.clone(this);
-	clone.users = []; // Do not send user list, the client will explicitly request it when needed
-	clone.messages = clone.messages.slice(-1);
-	return clone;
+/**
+ * Get a clean clone of this channel that will be sent to the client.
+ * This function performs manual cloning of channel object for
+ * better control of performance and memory usage.
+ *
+ * @param {(int|bool)} lastActiveChannel - Last known active user channel id (needed to control how many messages are sent)
+ *                                         If true, channel is assumed active.
+ * @param {int} lastMessage - Last message id seen by active client to avoid sending duplicates.
+ */
+Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
+	return Object.keys(this).reduce((newChannel, prop) => {
+		if (prop === "users") {
+			// Do not send users, client requests updated user list whenever needed
+			newChannel[prop] = [];
+		} else if (prop === "messages") {
+			// If channel is active, send up to 100 last messages, for all others send just 1
+			// Client will automatically load more messages whenever needed based on last seen messages
+			const messagesToSend = lastActiveChannel === true || this.id === lastActiveChannel ? -100 : -1;
+
+			// If client is reconnecting, only send new messages that client has not seen yet
+			if (lastMessage > -1) {
+				newChannel[prop] = this[prop]
+					.filter((m) => m.id > lastMessage)
+					.slice(messagesToSend);
+			} else {
+				newChannel[prop] = this[prop].slice(messagesToSend);
+			}
+		} else {
+			newChannel[prop] = this[prop];
+		}
+
+		return newChannel;
+	}, {});
 };
 
 function writeUserLog(client, msg) {
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -7,6 +7,17 @@ module.exports = Network;
 
 let id = 1;
 
+/**
+ * @type {Object} List of keys which should not be sent to the client.
+ */
+const filteredFromClient = {
+	awayMessage: true,
+	chanCache: true,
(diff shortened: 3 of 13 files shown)


src/server.js
```diff
@@ -440,6 +434,8 @@ function initializeClient(socket, client, token, lastMessage) {
 			networks: networks,
 			token: tokenToSend,
 		});
+
+		console.timeEnd(`${client.name} - time to emit to socket.io`);
```
astorije: Debugging leftovers.

src/models/chan.js
```diff
@@ -136,11 +136,27 @@ Chan.prototype.removeUser = function(user) {
...
+
+			// If channel is active, send up to 100 last messages, for all others send just 1
+			// Client will automatically load more messages whenever needed based on last seen messages
+			newChannel[prop] = this[prop].slice(this.id === lastActiveChannel ? -100 : -1);
+		} else {
+			newChannel[prop] = this[prop];
+		}
+
+		return newChannel;
+	}, {});
```
astorije: Could you add tests for all those please? I'm really uncomfortable shipping convoluted code like this (convoluted for a reason, I know) without having it tested. Also, they're both part of the models, so that's not too tricky to test (i.e. there is no `irc` or `socket` object involved...).

src/server.js
xPaw: I left it in to measure perf.

astorije: Right but we don't want to merge with this, right?

xPaw: I will remove it later before merging.

src/models/chan.js
xPaw: Add tests.

client/views/chat.tpl
```diff
@@ -12,7 +12,7 @@
 		<span title="{{topic}}" class="topic">{{{parse topic}}}</span>
 	</div>
 	<div class="chat">
-		<div class="show-more {{#equal messages.length 100}}show{{/equal}}">
+		<div class="show-more{{#if messages.length}} show{{/if}}">
```
astorije: What does that mean really? Show the button if the channel has messages?!

xPaw: Yep, because it's gonna send 1 message per channel, and 100 just for the active.

astorije: I see, thanks!

src/models/chan.js
```diff
@@ -136,11 +136,29 @@ Chan.prototype.removeUser = function(user) {
 	this.users.delete(user.nick.toLowerCase());
 };
 
-Chan.prototype.toJSON = function() {
-	var clone = _.clone(this);
-	clone.users = []; // Do not send user list, the client will explicitly request it when needed
-	clone.messages = clone.messages.slice(-100);
-	return clone;
+Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
```
astorije: Could you add a comment to describe what those arguments `lastActiveChannel` and `lastMessage` are to the function? Just looking at the signature, I'm not seeing what they would do.

src/models/chan.js
```diff
@@ -136,11 +136,29 @@ Chan.prototype.removeUser = function(user) {
...
+	// Perform manual cloning of channel object for better control of performance and memory usage
+	return Object.keys(this).reduce((newChannel, prop) => {
+		if (prop === "users") {
```
astorije: Does this need a `if <...>.hasOwnProperty` or whatnot?
Also, any specific reason to use this way instead of `for ... in`? I'd expect `for ... in` to be sliiightly faster, but you most likely know better.

src/models/chan.js
```diff
@@ -136,11 +136,29 @@ Chan.prototype.removeUser = function(user) {
...
+			// If client is reconnecting, only send new message that client has not seen yet
+			if (lastMessage > -1) {
+				newChannel[prop] = this[prop].filter((m) => m.id > lastMessage);
+			} else {
+				newChannel[prop] = this[prop];
+			}
+
+			// If channel is active, send up to 100 last messages, for all others send just 1
+			// Client will automatically load more messages whenever needed based on last seen messages
+			newChannel[prop] = newChannel[prop].slice(lastActiveChannel === true || this.id === lastActiveChannel ? -100 : -1);
```
astorije: Could this be combined with the previous operation? :man_shrugging:

xPaw: I wanted to, but this ends up being duplicated over both if/else.

src/models/chan.js
xPaw: `for in` should work.

src/models/chan.js
xPaw: Done.

src/models/chan.js
astorije: I figured, didn't know if you thought the potential perf improvement was worth the duplication. Thanks!

xPaw: I ended up changing it :)

src/models/chan.js
astorije: I could have sworn that `for ... in` + `hasOwnProperty` + whatever necessary would have been faster than `Object.keys(...).reduce` (because we loop only once).

src/models/network.js
```diff
@@ -63,14 +74,27 @@ Network.prototype.setNick = function(nick) {
...
+ *
+ * Both of the parameters that are accepted by this function are passed into channels' getFilteredClone call.
+ *
+ * @see {@link Chan#getFilteredClone}
+ */
+Network.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
+	return Object.keys(this).reduce((newNetwork, prop) => {
+		if (prop === "channels") {
+			// Channels objects perform their own cloning
+			newNetwork[prop] = this[prop].map((channel) => channel.getFilteredClone(lastActiveChannel, lastMessage));
```
astorije: It seems super weird that `lastActiveChannel` can be either an int or a boolean (essentially bypassing the int). Wouldn't make more sense to do this at the caller's level, something like `channel.getFilteredClone(channel.id === lastActiveChannel, lastMessage)`?

xPaw: This is effectively emulating overloaded methods since JS doesn't have these.

astorije: Meh, I really don't like this. This is not a construct the language has, it's odd to try to emulate it and can lead to bugs. But oh well. Glad this is merged either way.

PR #1794, opened 2017-12-03
Handle javascript errors while loading

xPaw:
This is not a fully fledged error reporting solution, but it does solve cryptic loading issues if there is an issue while Lounge scripts are still loading.

### DIFF
--- a/client/js/loading-slow-alert.js
+++ b/client/js/loading-slow-alert.js
@@ -27,10 +27,21 @@ window.g_LoungeErrorHandler = function LoungeErrorHandler(error) {
 	title = document.getElementById("loading-page-message");
 	title.textContent = "An error has occured that prevented the client from loading correctly.";
 
-	var element = document.createElement("p");
-	element.contentEditable = true;
-	element.textContent = error instanceof ErrorEvent ? error.message : error;
-	title.parentNode.insertBefore(element, title.nextSibling);
+	var summary = document.createElement("summary");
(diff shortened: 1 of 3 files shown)


client/js/loading-slow-alert.js
```diff
@@ -19,3 +19,18 @@ setTimeout(function() {
...
+window.g_LoungeErrorHandler = function LoungeErrorHandler(error) {
+	var title = document.getElementById("loading-title");
+	title.textContent = "An error has occured";
+
+	title = document.getElementById("loading-page-message");
+	title.textContent = "An error has occured that prevented the client from loading correctly.";
+
+	var element = document.createElement("p");
+	element.contentEditable = true;
+	element.textContent = error instanceof ErrorEvent ? error.message : error;
```
MaxLeiter: should be monospace imo

astorije: I would actually not display the error as part of the UI. I think we should always provide human-friendly error messages (yes, that requires `if`/`switch` statements we improve over time), and if necessary, open dev console.
99% of the time, it won't be completely random errors, but things we know (network issue, bundle issue, etc.) and therefore we can advise on.

Just checking, if you don't display `error.message`, it still appears in dev console, right?

And yes, I know mobile will not allow this. An alternative could be what Chrome does: default error is very human-readable, and a "More info" displays technical details. We could use `details`/`summary` for a rudimentary solution.

Issue #1793, opened 2017-12-03
Reverse style (0x16) is not actually implemented


xPaw:
It's being parsed in `parseStyle` but nothing actually happens or gets reversed, it's silently ignored.

https://modern.ircdocs.horse/formatting.html#reverse-color

PR #1797, opened 2017-12-04
#1793 - implement reverse style

grissly-man:
Reverses the fg and bg colors, if defined. If only fg is defined, currently sets background to the foreground color, and returns foreground to the default style

### DIFF
--- a/test/client/js/libs/handlebars/parse.js
+++ b/test/client/js/libs/handlebars/parse.js
@@ -244,6 +244,27 @@ describe("parse Handlebars helper", () => {
 		input: "\x11monospace",
 		expected: '<span class="irc-monospace">monospace</span>',
 	}, {
+		name: "reverse with foreground and background colors",
+		input: "\x0313,1fg and bg \x16and reversed",
+		expected:
+			'<span class="irc-fg13 irc-bg1">fg and bg </span>' +
+			'<span class="irc-fg1 irc-bg13">and reversed</span>',
+	}, {
+		name: "toggle reverse with foreground and background colors",
+		input: "\x0305,11text \x16reversed and \x16back again and \x16reversed",
+		expected:
+			'<span class="irc-fg5 irc-bg11">text </span>' +
+			'<span class="irc-fg11 irc-bg5">reversed and </span>' +
+			'<span class="irc-fg5 irc-bg11">back again and </span>' +
+			'<span class="irc-fg11 irc-bg5">reversed</span>',
+	}, {
+		name: "escape reverse when new colors are applied",
+		input: "\x0311,02text \x16 reversed \x0304,05 and new style",
+		expected:
+			'<span class="irc-fg11 irc-bg2">text </span>' +
+			'<span class="irc-fg2 irc-bg11"> reversed </span>' +
+			'<span class="irc-fg4 irc-bg5"> and new style</span>',
+	}, {
 		name: "resets",
 		input: "\x02bold\x038yellow\x02nonBold\x03default",
 		expected:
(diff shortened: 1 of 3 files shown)


xPaw:
This PR is a bit screwed since you based it on the same branch as your previous PR which brings in incorrect commits here.

client/js/libs/handlebars/parse.js
```diff
@@ -16,11 +16,20 @@ function createFragment(fragment) {
 	if (fragment.bold) {
 		classes.push("irc-bold");
 	}
-	if (fragment.textColor !== undefined) {
-		classes.push("irc-fg" + fragment.textColor);
-	}
-	if (fragment.bgColor !== undefined) {
-		classes.push("irc-bg" + fragment.bgColor);
+	if (fragment.reverse) {
```
xPaw: I think it's better to implement these in `parseStyle` so the fragments can be optimized correctly.

client/js/libs/handlebars/parse.js
```diff
@@ -39,10 +48,18 @@ function createFragment(fragment) {
 	const escapedText = Handlebars.Utils.escapeExpression(fragment.text);
 
 	if (fragment.hexColor) {
-		attributes += ` style="color:#${fragment.hexColor}`;
+		if (fragment.reverse) {
```
xPaw: I don't think we need to reverse hex colors for this. http://modern.ircdocs.horse/formatting.html#reverse-color

client/views/windows/help.tpl
```diff
@@ -37,6 +37,18 @@
 			</p>
 		</div>
 	</div>
+	
+	<div class="help-item">
+		<div class="subject">
+			<kbd class="key-all">Ctrl</kbd><kbd class="key-apple">⌘</kbd> + <kbd>R</kbd>
```
xPaw: Ctrl+R is reload page which is useful, overall I don't think keybind is needed for this at all.

client/js/libs/handlebars/parse.js
grissly-man: Happy to do this, but if we implement these in `parseStyle` doesn't that render the `reverse` flag in the fragment moot? Namely all logic/testing can be done via the `textColor` and `bgColor` and `reverse` seems somewhat redundant

xPaw: > doesn't that render the reverse flag in the fragment moot?

I believe so, yeah. Feel free to remove it.

test/client/js/libs/handlebars/parse.js
```diff
@@ -243,6 +243,18 @@ describe("parse Handlebars helper", () => {
 		name: "monospace",
 		input: "\x11monospace",
 		expected: '<span class="irc-monospace">monospace</span>',
+	}, {
+		name: "reverse with foreground and background colors",
+		input: "\x0313,1fg and bg \x16and reversed",
```
xPaw: I would add tests like `\x0313,01test\x16test\x16test`, `\x0313,01test\x16test\x01,13test` to see what kind of html it generates.

grissly-man:
@xPaw So currently this is set to toggle reverse until a new color formatting is specified. At that point it will parse the new formatting as normal, and not as though it is reversed. Is this the behavior you want implemented?

Also, the keybinding and help document have been reverted

xPaw:
I think that is expected, yeah.

Can you rebase correctly upon thelounge's master? Your PR currently has unnecessary commits in it.

grissly-man:
@xPaw Yessir

astorije:
Hey @grissly-man, we have free sticker packs for our contributors now!

PR #1799, opened 2017-12-05
Insert user-configurable string when autocompleting nicks

xPaw:
@thelounge/maintainers this is ready to go with no regressions!

### DIFF
--- a/client/views/windows/settings.tpl
+++ b/client/views/windows/settings.tpl
@@ -56,6 +56,13 @@
 			</label>
 		</div>
 		<div class="col-sm-12">
+			<label class="opt">
+				<label for="nickPostfix" class="sr-only">Nick autocomplete postfix (e.g. <code>, </code>)</label>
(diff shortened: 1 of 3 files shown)


client/views/windows/settings.tpl
```diff
@@ -55,6 +55,13 @@
 				Enable autocomplete
 			</label>
 		</div>
+		<div class="col-sm-12">
+			<label class="opt">
+				<label for="nickPostfix">Nick autocomplete postfix (e.g. <code>, </code>)</label>
+				<input type="text" id="nickPostfix" name="nickPostfix" class="input" placeholder="Nick autocomplete postfix">
```
astorije: Could you add `class="sr-only"` to the `label`, and ` (e.g. ", ")` to the `placeholder`, to match the custom highlights fields please?
It's not an idea look, I know, but to keep things consistent for now.

astorije:
Very nice. Minor comment but apart from that, all good.

client/views/windows/settings.tpl
xPaw: Did that.

PR #1800, opened 2017-12-05
Update textcomplete library and rewrite tabcomplete

### LINKED ISSUE #1061: Remove tabcomplete plugin in favor of textcomplete
*This will be relevant once #787 is merged*

We don't need two implementation of the same feature, I think we just need to figure an elegant way of triggering testcomplete's completion when there is no trigger (like `@`).

### LINKED ISSUE #1126: Combine @-less nick completion with new autocompletion dropdown
Right now, typing `@` opens the nick autocompletion dropdown while typing a set of character + hitting <kbd>Tab</kbd> completes like it used to.

When hitting a set of characters then <kbd>Tab</kbd>, the nick autocompletion dropdown should open with that set of characters as the `term` in completion strategy.

Essentially, typing `xyz` then hitting <kbd>Tab</kbd> should open a completion dropdown with the exact same suggestions than typing `@xyz`.

Thoughts?

### LINKED ISSUE #1208: jquery-textcomplete is no longer maintained
superseded by https://github.com/yuku-t/textcomplete

### LINKED ISSUE #1433: Tab-completion of channels unavailable with auto-complete disabled

xPaw:
- Fixes #1208 by updating to `textcomplete`
- Fixes #1433, correctly disable tab completion if option is turned off
- Fixes #1126 and fixes #1061 by rewriting tabcomplete within the autocomplete code and calling into `nicksStrategy.replace` to allow for features like #1799
- Fixes autocompletion being registered/enabled twice unintentionally
- Should have better performance as `completeNicks` grep is no longer performed within `input` event
- Replaces `<space>|\n` regex with `\s` to better match all whitespace
- The code is simpler over all for better maintainability


astorije:
Noticed a bug when testing this, typing `:` (or `#`) + <kbd>Tab</kbd> auto-completes with `undefined`. It doesn't happen with `@`.

xPaw:
@astorije Fixed.

PR #1802, opened 2017-12-06
Add 'Channel list' to network context menu

MaxLeiter:
No clue if we want this or not, I'm a fan. 🤷‍♂️ 

I just think some commands should be more accessible from the UI

image

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -90,6 +90,13 @@ $(function() {
 				data: target.data("target"),
 			});
 			output += templates.contextmenu_divider();
+			if (target.hasClass("lobby")) {
+				output += templates.contextmenu_item({
+					class: "list",
+					text: "List all channels",
+					data: target.data("target"),
+				});
+			}
 			output += templates.contextmenu_item({
 				class: "close",
 				text: target.hasClass("lobby") ? "Disconnect" : target.hasClass("channel") ? "Leave" : "Close",
@@ -441,6 +448,12 @@ $(function() {
 			break;
 		case "user":
 			$(".channel.active .users .user[data-name='" + $(this).data("data") + "']").click();
+			break;
+		case "list":
+			socket.emit("input", {
+				target: chat.data("id"),
+				text: "/list",
+			});
 			break;
 		}
 	});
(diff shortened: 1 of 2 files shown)


astorije:
> I just think some commands should be more accessible from the UI

💯

client/js/lounge.js
```diff
@@ -89,6 +89,13 @@ $(function() {
 				text: target.data("title"),
 				data: target.data("target"),
 			});
+			if (target.hasClass("lobby")) {
+				output += templates.contextmenu_item({
+					class: "list",
+					text: "Channel list",
+					data: target.data("target"),
+				});
+			}
```
astorije: Could you move this under the divider?

client/js/lounge.js
```diff
@@ -89,6 +89,13 @@ $(function() {
 				text: target.data("title"),
 				data: target.data("target"),
 			});
+			if (target.hasClass("lobby")) {
+				output += templates.contextmenu_item({
+					class: "list",
+					text: "Channel list",
```
astorije: I'm debating if it would not be more user-friendly to have something like "List channels on this server" or something.

MaxLeiter: I think that’s too long... maybe “List existing channels”?

YaManicKill: I agree, it's too long.

Maybe "List Channels" or "List all channels" or "List of Channels" or something.

MiniDigger: I don't really see whats wrong with Channel List, its pretty straight forward. List Channels would fit better to disconnect tho since both are an action.

YaManicKill: I guess the issue with "Channel List" or "List Channels" is ... what channels? But none of our solutions fix that problem, really. I'd be fine with any of these other than "List channels on this server"

client/js/lounge.js
MaxLeiter: Done

client/js/lounge.js
astorije: screen shot 2017-12-06 at 18 36 45

:man_shrugging: 

Let's go with "List all channels" then.

client/js/lounge.js
```diff
@@ -90,6 +90,13 @@ $(function() {
 				data: target.data("target"),
 			});
 			output += templates.contextmenu_divider();
+			if (target.hasClass("lobby")) {
+				output += templates.contextmenu_item({
+					class: "list",
+					text: "Channel list",
```
astorije: "List all channels"

astorije:
Adding to v2.7.0 milestone as it's super trivial.

MaxLeiter:
Should be good to go

PR #1804, opened 2017-12-06
Enable no-use-before-define rule

xPaw:


client/js/options.js
```diff
@@ -127,56 +160,11 @@ module.exports.initialize = () => {
...
 			}
+		} else if (name === "desktopNotifications") {
+			if ($(this).prop("checked") && Notification.permission !== "granted") {
+				Notification.requestPermission(updateDesktopNotificationStatus);
+			}
 		}
 	}).find("input")
 		.trigger("change");
-
-	$("#desktopNotifications").on("change", function() {
```
xPaw: Lines 134-148 are now handled by `if (name === "desktopNotifications")`.

PR #1806, opened 2017-12-06
Support audio file previews

MaxLeiter:
example

To test you can use https://demo.thelounge.chat/audio/pop.ogg

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -23,7 +23,7 @@ const options = {
 	showSeconds: false,
 	statusMessages: "condensed",
 	theme: $("#theme").attr("href").replace(/^themes\/(.*).css$/, "$1"), // Extracts default theme name, set on the server configuration
-	thumbnails: true,
+	media: true,
 	userStyles: userStyles.text(),
 };
 let userOptions = JSON.parse(storage.get("settings")) || {};
@@ -44,7 +44,7 @@ userOptions = null;
 module.exports = options;
 
 module.exports.shouldOpenMessagePreview = function(type) {
-	return (options.links && type === "link") || (options.thumbnails && type === "image");
+	return (options.links && type === "link") || (options.media && (type === "image" || type === "audio"));
 };
 
 module.exports.initialize = () => {
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -106,6 +106,22 @@ function parse(msg, preview, res, client) {
 
 		break;
 
+	case "audio/midi":
+	case "audio/mpeg":
+	case "audio/mpeg3":
+	case "audio/ogg":
+	case "audio/wav":
+	case "audio/x-mid":
+	case "audio/x-midi":
+	case "audio/x-mpeg":
+	case "audio/x-mpeg-3":
+		if (!preview.link.startsWith("https://")) {
+			break;
+		}
+		preview.type = "audio";
+		preview.res = res.type;
+		break;
+
 	default:
 		return;
 	}
@@ -170,6 +186,8 @@ function fetch(uri, cb) {
 				if (contentLength > limit) {
 					req.abort();
 				}
+			} else if (/^audio\/.+/.test(res.headers["content-type"])) {
+				req.abort(); // ensure server doesn't download the audio file
 			} else {
 				// if not image, limit download to 50kb, since we need only meta tags
 				// twitter.com sends opengraph meta tags within ~20kb of data for individual tweets
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1235,6 +1235,11 @@ kbd {
 	display: inline-flex !important;
 }
 
+#chat audio {
+	width: 600px;
+	max-width: 100%;
+}
+
 /* Do not display an empty div when there are no previews. Useful for example in
 part/quit messages where we don't load previews (adds a blank line otherwise) */
 #chat .preview:empty {
--- a/client/views/msg_preview.tpl
+++ b/client/views/msg_preview.tpl
@@ -5,15 +5,22 @@
 			<img src="{{thumb}}">
 		</a>
 	{{else}}
-		{{#if thumb}}
-			<a class="toggle-thumbnail" href="{{link}}" target="_blank" rel="noopener">
-				<img src="{{thumb}}" class="thumb">
+		{{#equal type "audio"}}
+			<audio controls preload="metadata">
+				<source src="{{link}}" type="{{res}}">
+				Your browser does not support the audio element.
+			</audio>
+		{{else}}
+			{{#if thumb}}
+				<a class="toggle-thumbnail" href="{{link}}" target="_blank" rel="noopener">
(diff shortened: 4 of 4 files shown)


src/plugins/irc-events/link.js
```diff
@@ -106,6 +108,23 @@ function parse(msg, preview, res, client) {
...
 
+	case "audio/mpeg3":
+	case "audio/x-mpeg-3":
+	case "audio/mpeg":
+	case "audio/x-mpeg":
+	case "audio/midi":
+	case "audio/x-mid":
+	case "audio/x-midi":
+	case "audio/ogg":
+		if (res.size > (Helper.config.prefetchMaxImageSize * 4028)) {
```
MaxLeiter: need to change this

src/plugins/irc-events/link.js
```diff
@@ -65,6 +66,7 @@ function parse(msg, preview, res, client) {
 			|| $('meta[name="twitter:image:src"]').attr("content")
 			|| $('link[rel="image_src"]').attr("href")
 			|| "";
+		preview.type = res.type;
```
xPaw: unnecessary?

MaxLeiter: yeah, but may as well populate if it's there?

xPaw: It's already set to `link` above.

MaxLeiter: fixed

src/plugins/irc-events/link.js
dgw: Why not a new `prefetchMaxAudioSize` defaulted to 4-8 MB?

Or deprecate the existing setting in favor of a `prefetchMaxMediaSize`.

src/plugins/irc-events/link.js
```diff
@@ -106,6 +106,23 @@ function parse(msg, preview, res, client) {
...
+	case "audio/midi":
+	case "audio/mpeg":
+	case "audio/mpeg3":
+	case "audio/ogg":
+	case "audio/wav":
+	case "audio/x-mid":
+	case "audio/x-midi":
+	case "audio/x-mpeg":
+	case "audio/x-mpeg-3":
+		if (res.size > (Helper.config.prefetchMaxImageSize * 4028)) {
```
xPaw: How do browsers deal with preloading audio data? Does it download the entire audio file as soon as it sees the `audio` tag or downloads just enough to display the duration?

MaxLeiter: Browsers should default to fetching metadata, but I can add a ‘preload’ attribute to specify

xPaw: My point is, if browsers just fetch the metadata, don't really need to limit the size.

You do need to abort the request after content-type was fetched by the way.

MaxLeiter: MDN says browsers should should default to metadata, and says that the preload attribute acts as a hint to the browser — should we trust the browser?

astorije: MDN mentions:

> If not set, its default value is browser-defined (i.e. each browser may have its own default value). The spec advises it to be set to `metadata`.

I think it's fair to have `preload="metadata"`.

xPaw: What if we went with `preload=none` to avoid unnecessary requests? When I tried this PR Chrome didn't display duration half the time anyway.

MaxLeiter: Time won’t be loaded then, and I think that’s pretty important

MaxLeiter: added `preload="metadata"`

astorije:
@MaxLeiter, without going all-in with styling as of yet, could you make this use the whole container? Something like this:

screen shot 2017-12-07 at 13 27 25

xPaw:
@astorije Don't do that, link previews only take width of their content not parent. It will look rather dumb on any big screen.

MaxLeiter:
removed `prefetchMaxImageSize` test, as it's at the users discretion if the audio file should be downloaded or not. Ready to go, unless I should do something with storage, I think?

astorije:
Makes sense @xPaw, but at least 1.5 or 2x Chrome's default width (max out at parent width for mobile), it's very small as-is and makes seeking through progress rather hard.

src/plugins/irc-events/link.js
```diff
@@ -106,6 +106,21 @@ function parse(msg, preview, res, client) {
...
+	case "audio/midi":
+	case "audio/mpeg":
+	case "audio/mpeg3":
+	case "audio/ogg":
+	case "audio/wav":
+	case "audio/x-mid":
+	case "audio/x-midi":
+	case "audio/x-mpeg":
+	case "audio/x-mpeg-3":
+		if (!`/^https:\/\//`.test(url)) {
```
xPaw: Why do you have string interpolation here?

This could be done without regex: `url.startsWith("https://")`

MaxLeiter:
should be set

MaxLeiter:
@astorije bumped the width

astorije:
Looks great @MaxLeiter!

I've noticed a couple issues, trying to make sense out of the first one:

- https://mdn.github.io/learning-area/html/multimedia-and-embedding/video-and-audio-content/viper.ogg works fine, but http://techslides.com/demos/samples/sample.mp3 that you posted in the screenshot does not load in the client (it does load fine if I open a new tab). Did you change something?

- When posting an image URL, the message shows up with the image open (along with the 🔽 caret), but when posting an audio URL, it keeps it closed (along with the ➡️ caret).

MaxLeiter:
First link is https, second is not

No clue on second point, will investigate

astorije:
> First link is https, second is not

Ah right, I forgot that we added this since your original commit, thanks!

> will investigate

Cool, thanks!

It's very cool you took this btw, nice job 👏

xPaw:
It doesn't open by default because of this: https://github.com/thelounge/lounge/blob/d0f5d5025e4c5184ace4d98d250643d37e1d0c92/client/js/options.js#L46-L48

We most likely need to switch `image` to a more generic `media`.

MaxLeiter:
I think thats part of a larger issue with embedding images/links (maybe it should just be (`expand media` or something)).

For now I'll just have it embed if options.links?

client/js/options.js
```diff
@@ -44,7 +44,7 @@ userOptions = null;
 module.exports = options;
 
 module.exports.shouldOpenMessagePreview = function(type) {
-	return (options.links && type === "link") || (options.thumbnails && type === "image");
+	return (options.links && type === "link") || (options.media && (type === "image" || type === "audio"));
```
astorije: Off-topic/Not for this PR: @xPaw, @MaxLeiter, what about merging both under `options.preview` or something?

MaxLeiter: I think there's a slight difference, in that links will be low-bandwidth, but media could vary. At least, I think that's how @xPaw and I see it at the moment.

astorije: That's fair, thanks!

PR #1807, opened 2017-12-07
Do not load user commands or display them on `--help` if public mode or using LDAP

astorije:
Also move user commands to their own subfolder (which should unfold even more nicely in v3 I think).

Commands like `list` do not make sense when using LDAP (or public) as they load the user config files, and commands like `add` only make sense in private, non-LDAP mode.
In the future, we should allow plugins to add their own commands, in which case LDAP integration could itself have meaningful commands there. In the meantime, my understanding is that it's noise.

@gramakri, could you give your 👍 to confirm this PR?

(You can see I'm working on the re-docs for the CLI because I'm going to open a couple PRs on this 😅)

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -51,7 +51,9 @@ Helper.setHome(home);
 
 require("./start");
 require("./config");
-require("./users");
+if (!Helper.config.public && !Helper.config.ldap.enable) {
+	require("./users");
+}
 require("./install");
 
 // TODO: Remove this when releasing The Lounge v3
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -51,11 +51,7 @@ Helper.setHome(home);
 
 require("./start");
 require("./config");
-require("./list");
-require("./add");
-require("./remove");
-require("./reset");
-require("./edit");
+require("./users");
 require("./install");
 
 // TODO: Remove this when releasing The Lounge v3
--- a/src/command-line/users/add.js
+++ b/src/command-line/users/add.js
@@ -17,11 +17,6 @@ program
 		}
 
 		const ClientManager = require("../../clientManager");
-
-		if (Helper.config.public) {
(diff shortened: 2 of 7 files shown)


src/command-line/index.js
```diff
@@ -51,11 +51,9 @@ Helper.setHome(home);
 
 require("./start");
 require("./config");
-require("./list");
-require("./add");
-require("./remove");
-require("./reset");
-require("./edit");
+if (!Helper.config.public && !Helper.config.ldap.enable) {
```
xPaw: The only problem here is that CLI overrides (e.g. `--public` and later on `--config` will have no effect on this.

astorije: That's a good point.
A bit niche right now I think though, because the only situation where it could hurt is if your config file says `public: true` but you intend to run `thelounge --private` and try to run `thelounge add <user>`. That seems very specific and unlikely to me (as opposed to port and stuff, or even run the other way around: private in config file and `--public`).

One solution I can see is moving `-c` at `thelounge` level. So `thelounge -c public=false start`. That way, you can do `thelounge -c public=false add <user>` and it would work. What do you think?
In general, it might be better to do that anyway, because in a world where we have package, we'll want to give the ability to packages to register their own commands, and they would benefit from overrides just as much.
I'm starting to really prefer this idea actually.

xPaw: In your other PR, just do `parseConfigOptions` after doing `setHome` in `index.js` and then it will work as you are describing.

astorije: Yes exactly. I just wanted to make sure you were on board with that. It's a breaking change from what you describe, but such a niche one that it doesn't matter, and `-c` for 2.7.0 brings way more benefits.

Only thing is `thelounge start -c public=false` will not work anymore, options have to be passed before start (I could duplicate in `index.js` AND `start.js` but that starts to be too intricate).

astorije:
Gonna merge this now so I can test with improved #1820 correctly.

PR #1808, opened 2017-12-07
Do not expand paths when displaying them to the user

astorije:
This refactor has a few benefits, for example there cannot be a rogue update of `Helper.CONFIG_PATH` or something.

It kinda bugged me that the everything but the defaults location (last line of `thelounge -h`) was showing an unexpanded version, hated the inconsistency :)

Before | After
--- | ---
screen shot 2017-12-06 at 22 59 18 | screen shot 2017-12-06 at 22 56 14
screen shot 2017-12-06 at 22 59 33 | screen shot 2017-12-06 at 22 57 00
screen shot 2017-12-06 at 22 59 43 | screen shot 2017-12-06 at 22 57 29
screen shot 2017-12-06 at 23 00 19 | screen shot 2017-12-06 at 22 58 02

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -43,7 +43,7 @@ ClientManager.prototype.autoloadUsers = function() {
 
 	users.forEach((name) => this.loadUser(name));
 
-	fs.watch(Helper.USERS_PATH, _.debounce(() => {
+	fs.watch(Helper.getUsersPath(), _.debounce(() => {
 		const loaded = this.clients.map((c) => c.name);
 		const updatedUsers = this.getUsers();
 
@@ -88,7 +88,7 @@ ClientManager.prototype.loadUser = function(name) {
 
 ClientManager.prototype.getUsers = function() {
 	return fs
-		.readdirSync(Helper.USERS_PATH)
+		.readdirSync(Helper.getUsersPath())
 		.filter((file) => file.endsWith(".json"))
 		.map((file) => file.slice(0, -5));
 };
--- a/src/command-line/add.js
+++ b/src/command-line/add.js
@@ -11,8 +11,8 @@ program
 	.description("Add a new user")
 	.on("--help", Utils.extraHelp)
 	.action(function(name) {
-		if (!fs.existsSync(Helper.USERS_PATH)) {
-			log.error(`${Helper.USERS_PATH} does not exist.`);
+		if (!fs.existsSync(Helper.getUsersPath())) {
+			log.error(`${Helper.getUsersPath({expanded: false})} does not exist.`);
 			return;
 		}
 
--- a/src/command-line/config.js
+++ b/src/command-line/config.js
@@ -9,20 +9,20 @@ const Utils = require("./utils");
 
 program
 	.command("config")
-	.description(`Edit configuration file located at ${colors.green(Helper.CONFIG_PATH)}.`)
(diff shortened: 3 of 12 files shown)


xPaw:
Yeah, I'm not a fan of this. I'm pretty sure this will break on Windows too since `~` is not a thing there.

You made `getHomePath` return non expanded path by default, which is the entire point of that function. I would very much prefer to keep it as it is instead of introducing new potential bugs for slightly nicer UI display (which I would argue it is more useful to see the full path rather than guessing).

astorije:
😭 

> I'm pretty sure this will break on Windows too since `~` is not a thing there.

AppVeyor passing is kind of a good validation that this works on Windows when it comes to this. Besides, `~` will only appear in display, not in calls.

> You made getHomePath return non expanded path by default, which is the entire point of that function.

Ah, yes, that's my mistake. It was expanded by default in my previous attempt, but breaking on Node 4, and I messed up the changes. Let's assume for the sake of argument that it stays expanded by default :D

astorije:
> You made `getHomePath` return non expanded path by default

This is fixed now.
Is this still something you don't want?

Besides the expanded vs. non-expanded versions, I think relying on variables (`HOME` vs. `CONFIG_PATH` is bound to backfire, as one can be changed and not keep the others in sync. I think we should use functions for all those.

xPaw:
> Is this still something you don't want?
I think adding `expanded` is more bug prone than having `HOME` variables.

> I think relying on variables (HOME vs. CONFIG_PATH is bound to backfire, as one can be changed and not keep the others in sync. I think we should use functions for all those.

You still kept some variables. I don't making these helper functions bubble all the way up to `expandHome` is a good idea either. Keeping home folder in a variable and using that is better. (we have setHome for a reason).

astorije:
> You still kept some variables.

Mmmh, no I didn't. I obviously meant "variables that can be mutated outside of this file", not "we should not use variables"...
The last one of those in the helper is `config`, which I didn't touch altogether.

> I don't making these helper functions bubble all the way up to `expandHome` is a good idea either.

I'm not sure what you mean by that.
My point is that on top of having `setHome`, those variables (`HOME`, `CONFIG_PATH`, etc.) can also be modified on their own and that would break a lot of things. We are using those in a lot of places, I'm not sure why we should keep this liability, which has room for bugs.

> (we have setHome for a reason).

Yes, and that reason is exactly why we shouldn't be able to edit those variables. Also, the inconsistency between `Helper. CONFIG_PATH`, `Helper.USERS_PATH`, `Helper. getStoragePath()`, `Helper. getPackagesPath()`, etc. does not make sense to me. These should all be `Helper.getXxxPath()`.

xPaw:
I agree that we can make all the paths functions, but internally within the helper file keep the variables so we're not reformatting the home string every time the function is called (this pretty much just applies to the HOME).

astorije:
That's fair, and I only needed that to expand `~` within the helper function. But since you don't like this anyway, I guess I don't need it anymore... 😭

xPaw:
> But since you don't like this anyway, I guess I don't need it anymore... 😭

In my opinion it's an unnecessary increase in complexity. And I haven't really seen other software print paths within help outputs like we do. And the ones that print paths, it's the full one.

astorije:
Alright then, I'll remove the part that doesn't expand on display.

astorije:
Closing this as superseded by https://github.com/thelounge/lounge/pull/1811. I opened a different PR to keep the discussion relevant with its context.

PR #1809, opened 2017-12-07
Switch CLI from `thelounge --version` to `thelounge version`

astorije:
I was looking forward to having no top-level options because they're confusing:

- What does `thelounge add -v` means compared to `thelounge -v`?
- Why does `thelounge add -h` and `thelounge -h add` differ?
- `-v` often means "verbose"
- Message for `-v` is auto-generated, we do not control the wording / casing

I was planning to have the following syntax:

```
thelounge version
thelounge help
thelounge help add
```

... and so on.

~This makes my docs reboot around CLI look nicely organized...~ (Actually, I've changed the format a bit, and it doesn't matter that much anymore, but I still think the other arguments make `thelounge version` better)
Unfortunately, while `thelounge version` is possible (it's even in one of their examples), there is currently no way to remove `-h, --help` from commander.js :(

So until then, this will only "convert" `thelounge version`, which is already a win. I'll make the docs work with `--help` until a better solution arises.

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -14,7 +14,7 @@ if (require("semver").lt(process.version, "6.0.0")) {
 	log.warn("Please upgrade to Node.js v6 or more recent.");
 }
 
-program.version(Helper.getVersion(), "-v, --version")
+program
 	.option("--home <path>", `${colors.bold("[DEPRECATED]")} Use the ${colors.green("THELOUNGE_HOME")} environment variable instead.`)
 	.on("--help", Utils.extraHelp)
 	.parseOptions(process.argv);
@@ -57,6 +57,7 @@ require("./remove");
 require("./reset");
 require("./edit");
 require("./install");
+require("./version");
 
 // TODO: Remove this when releasing The Lounge v3
 if (process.argv[1].endsWith(`${require("path").sep}lounge`)) {
--- a/src/command-line/version.js
+++ b/src/command-line/version.js
@@ -1,0 +1,11 @@
+"use strict";
+
+const program = require("commander");
+const Helper = require("../helper");
+
+program
+	.command("version")
+	.description("Output the version number")
+	.action(() => {
+		console.log(Helper.getVersion()); // eslint-disable-line no-console
+	});


MaxLeiter:
not a huge fan. It's conventional to have parameters start with `--` for full and `-`for shorthand. People expect anything without `-` or `--` to open a prompt. `program --version` and `program -v` are incredibly common.

astorije:
> It's conventional to have parameters start with `--` for full and `-`for shorthand.

I agree, but my point is that this is not a parameter, it's a command. Passing it to a command does not affect the command.

> People expect anything without `-` or `--` to open a prompt.

Unless I misunderstood... no 😅

> `program --version` and `program -v` are incredibly common.

This is true, but so is `program version` (especially in programs with commands like `thelounge`/`git`/etc.), as well as `program -v` for verbosity. You'll notice that I marked this for v3, I think it would be nice to keep `-v` available in case we want/need verbosity level in the v3 life cycle.

> I don't think people expect `program command` to terminate instantly.

Most programs I use on a daily basis do, so this is not an esoteric concept. And in our specific case, `thelounge list` does.

I hear you, but overall I still think this switch makes sense. Also it makes the new docs marginally cleaner and easier to follow (haven't finished the page yet), which I think is good.

YaManicKill:
> Why does thelounge add -h and thelounge -h add differ?

Oh no, I definitely disagree with this one. I would expect `thelounge add -h` would give you the detailed help around specifically the add command, whereas `thelounge -h` gives you higher level info around the whole thing.

I can accept the version one.

astorije:
> I would expect `thelounge add -h` would give you the detailed help around specifically the add command, whereas `thelounge -h` gives you higher level info around the whole thing.

Definitely agree for those 2, but the less obvious case is `thelounge -h add`, which gives the high-level info and not detailed help on `add`. It effectively ignores `add` without letting the user know or without failing or whatever.
I think that if you are debuting with The Lounge, and more broadly with hosting, it's a bit confusing. `thelounge help [command]` (like git does) would be a nice way to lift this confusion. Except, well, it's not possible :man_shrugging:

xPaw:
:-1: for removing `--version`. As far as I know this is a lot more common than a sub-command.

astorije:
Argh, you guys are killing me. Closing then.
(Also, I find a sub-command more common than a flag when a CLI is sub-command-based, like `git`, Kubernetes CLIs, etc.)
I guess if we ever need a verbosity flag, we'll re-assign `-v` to that and keep `--version` then.

PR #1811, opened 2017-12-08
Clean up path helpers, expand defaults location in `thelounge --help`, add tests for `expandHome`

astorije:
Supersedes #1808.

- First commit only does the cleanup that we discussed around path getters.

- Second commit actually does the opposite of what #1808 did: instead of solving the inconsistency by "tildifying" everything, it expands the only path that is _not_ expanded, the default location of the home folder. In the end, it makes sense, because in practice The Lounge does not load from `~/.lounge`, it loads from `<my home dir>/.lounge`.
  Arguments in favor of why we should display expanded paths were better explained by @xPaw on #1808 and I came to agree (as long as that inconsistency is fixed :D).
  See screenshots below for result.

- Third commit adds a few tests. I was trying to find a way to expand `~userB/tmp` into `<userB's home dir>/tmp`, but couldn't find a way. Oh well, not really common in our case I think.

Before | After
--- | ---
screen shot 2017-12-06 at 22 59 18 | screen shot 2017-12-07 at 23 00 41

### DIFF
--- a/test/src/helperTest.js
+++ b/test/src/helperTest.js
@@ -1,0 +1,39 @@
+"use strict";
+
+const expect = require("chai").expect;
+const os = require("os");
+const Helper = require("../../src/helper");
+
+describe("Helper", function() {
+	describe("#expandHome", function() {
+		it("should correctly expand a Unix path", function() {
+			expect([`${os.homedir()}/tmp`, `${os.homedir()}\\tmp`])
+				.to.include(Helper.expandHome("~/tmp"));
+		});
+
+		it("should correctly expand a Windows path", function() {
+			expect(Helper.expandHome("~\\tmp")).to.equal(`${os.homedir()}\\tmp`);
+		});
+
+		it("should correctly expand when not given a specific path", function() {
+			expect(Helper.expandHome("~")).to.equal(os.homedir());
+		});
+
+		it("should not expand paths not starting with tilde", function() {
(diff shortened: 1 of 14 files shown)


astorije:
So, one of the failures AppVeyor reports is this:

screen shot 2017-12-07 at 23 43 12

I don't have a Windows machine to test, but I'm curious: @xPaw, how come `~/.lounge` (default location) works fine for us on Windows? Is it because Node doesn't care and interprets it in current OS format?

astorije:
Yeah, nevermind, that's what it does. `path.resolve` will read Unix formats and "convert" them to Windows format. Interesting, I never caught that before.

test/src/helperTest.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
+
+const expect = require("chai").expect;
+const os = require("os");
+const Helper = require("../../src/helper");
+
+describe("Helper", function() {
+	describe("#expandHome", function() {
+		it("should correctly expand a Unix path", function() {
+			expect([`${os.homedir()}/tmp`, `${os.homedir()}\\tmp`])
```
xPaw: This test seems completely useless, did you miss something?

astorije: Yes I did :)
Good catch, will fix, thanks!

astorije: Fixed.

PR #1818, opened 2017-12-09
Remove dead code in tests, and fix a link test

astorije:
I used `npm run coverage` while *not* excluding the test folder to detect dead code in our test folder, it is actually pretty useful to do so (as a one-shot, not to do that in our config).

Only remaining unreached path is [L40 in `test/plugins/auth/ldap.js`](https://github.com/thelounge/lounge/blob/a8fb892873721a9ac8eb138533c999180f333313/test/plugins/auth/ldap.js#L40), but it does seem to me that it might be useful in case of failures, so I preferred to leave it there.

Proof of no more dead code (read this as non-dead code coverage and not test coverage, of course):

screen shot 2017-12-09 at 19 07 49

**Leftovers:**

screen shot 2017-12-09 at 19 05 55 screen shot 2017-12-09 at 19 06 16

### DIFF
--- a/test/fixtures/env.js
+++ b/test/fixtures/env.js
@@ -1,11 +1,6 @@
 "use strict";
 
-global.log = {
-	error: () => console.error.apply(console, arguments), // eslint-disable-line no-console
-	warn: () => {},
-	info: () => {},
-	debug: () => {},
-};
+global.log = require("../../src/log.js");
 
 var home = require("path").join(__dirname, ".lounge");
 require("../../src/helper").setHome(home);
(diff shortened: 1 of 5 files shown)


test/fixtures/env.js
```diff
@@ -1,11 +1,6 @@
 "use strict";
 
-global.log = {
-	error: () => console.error.apply(console, arguments), // eslint-disable-line no-console
-	warn: () => {},
-	info: () => {},
-	debug: () => {},
-};
+global.log = require("../../src/log.js");
```
astorije: I think it's wiser to do that.
In fact, while running this change on some code/test I am writing right now, it showed me that I was generating a unhandled warning in one of my tests and I completely missed it because it was being silenced.

Issue #419, opened 2016-06-19
Add a CLI param to override config variables


xPaw:
Something like `-c https.enable=true`.

This would remove the need to have ip/port/public/private params.

toXel:
Should the ip/port/public/private params be removed in the same PR? I'll look into this issue

xPaw:
@toXel We'd want to deprecate them as not to break backwards compatibility for now.

astorije:
@xPaw, I'm a bit on the fence about this idea. What's wrong with having explicit `--port`, `--private` etc. instead of `-c port`, `-c private=true`, etc.?
It might result in argument parsing duplication, but it's nicer from a CLI user point of view.

I'm not close to the idea, I just want to discuss the outcome before someone spends time on this :-)

toXel:
The thing is: at the moment the parameters are not working correctly. For example:
`lounge start --port 1234` or `lounge start -P 1234` starts Lounge on default port 9000. Same with `lounge --port 1234 start` (That's the way `lounge -h` describes the usage)
To make it work you have to do `lounge --port 1234`, `lounge -- start --port 1234` or `lounge -- --port 1234 start` which is not the way it should be, I think.
The whole CLI params parsing implementation has a problem.

astorije:
Thinking back at this, and after fiddling with command line code and discovering the madness there, I agree with this.

A few random thoughts about the CLI, some of them being related to this very issue:

- `lounge <subcmd>` is a nice thing, àla git. Problem is, it is very poorly handled at the moment. This should be improved by respecting [this](https://github.com/tj/commander.js#git-style-sub-commands) (we use `cmd`, just not correctly 😅)
- `config` subcommand is pointless and should be deprecated. Config will most likely always be this config file and can be edited with any text editor (as opposed with user management which might evolve and subcommands for that are just a nice abstraction over that).
- `private`/`public`/`host`/`bind`/`port` are indeed not necessary here for a normal use case. Esoteric use cases can use your suggested option.
- Some options are still good to have: location of the configuration (currently `home`, but I'm not too happy with how it's handled), `--help`, `--version`, and maybe `--debug`? (even though this latter one can be enabled with your suggestion, there might be some benefits in having it at this level, like debugging configuration file issues, etc.)

So anyway, 👍 on the idea with some suggestions.

PR #1820, opened 2017-12-10
Deprecate existing options of `thelounge start` and add a generic `--config` override

astorije:
Resolves https://github.com/thelounge/lounge/issues/419.

Removes the `--public` / `--private` / `--host` / `--bind` / `--port` options for `thelounge start` in favor of a new slightly-more-intricate-but-much-more-powerful `--config` / `-c` option.

Example on how to use this (note that I have successfully tested all examples given in this description):

```sh
thelounge start -c public=true \
                -c lockNetwork=true \
                -c defaults.port=6667 \
                -c defaults.tls=false
```

(Multiline for GitHub readability but it obviously works on a single line)

There are a bunch of different cases I handle, and tests for all situations I could think of.

Some more example of what is possible:

```sh
-c https.enable=true    # Supports booleans
-c port=9001    # See below for caveats on integers
-c debug.raw=true    # Supports dot-notation for objects (also comes with array-notation out-of-the-box, but not very useful in our case
-c transports=[websocket,polling]    # Supports arrays
-c "transports=[websocket, polling]"    # Supports spaces assuming option is wrapped in quotes
-c logs.format="DD MMMM YYYY HH:mm:ss"   # Can also quote just the value
-c ldap.searchDN.rootDN="cn=thelounge,ou=system-users"   # Supports values containing `=`
-c ldap.searchDN.filter='(objectClass=person)(memberOf=ou=accounts,dc=example,dc=com)'   # Complex value
```

I expect we will discover weird edge cases later, but I think this covers maybe 95% of the needs if not more.

One thing this does _not_ support is passing JavaScript code, like functions, for example for `webirc`. This would require using `eval()` which is a no-no for me considering this can affect a whole instance. I will argue that if you need to pass complex JavaScript code through the CLI, you should definitely define your own config file! (which is also valid for some complex cases, like LDAP, but it's easier and less risky to support)

Integers are not parsed as such. I have checked and tested almost all usages of integer options in our configuration files (maybe expect `identd`), and they all work fine. In some cases, we `parseInt`, but also note that strings are coerced into numbers in almost all cases in JS. For example, `"5" * 3` is `15` (as an integer) and `"5.1" * 3` is `15.3` (as a float) so all `<`/`>` comparisons work fine.
If we ever discover an issue with that, I can think of at least 3 ways of handling this, from easy to rather complex (the simplest would be to automatically `parseInt` if value is integer-looking, though I'm sure there are other pitfalls), but I think we should do this simply right now and fix as we go.
After all, this Just Works™ :)

---

About the first commit: _Define a raw logger to avoid using `console.log`, use it in extra help for environment variables, and add a test for this_

As said in commit message, this has a few benefits:

- Respects the "Do not mock what you do not own" principle, instead we mock `log.raw` when necessary
- Lets us not re-assign `console.log`, which breaks as Mocha uses `console.log` as well
- Save and restore initial `log.raw` in test hooks (before/after), otherwise this would break Mocha/Chai

This allowed me to write extra tests :)

---

Note: This PR currently fails on Node v4 but will pass once #1727 gets merged.

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -2,6 +2,7 @@
 
 global.log = require("../log.js");
 
+const _ = require("lodash");
 const fs = require("fs");
 const path = require("path");
 const program = require("commander");
@@ -16,6 +17,11 @@ if (require("semver").lt(process.version, "6.0.0")) {
 
 program.version(Helper.getVersion(), "-v, --version")
 	.option("--home <path>", `${colors.bold("[DEPRECATED]")} Use the ${colors.green("THELOUNGE_HOME")} environment variable instead.`)
+	.option(
+		"-c, --config <key=value>",
+		"override entries of the configuration file, must be specified for each entry that needs to be overriden",
+		Utils.parseConfigOptions
+	)
 	.on("--help", Utils.extraHelp)
 	.parseOptions(process.argv);
 
@@ -48,6 +54,9 @@ if (!home) {
 }
 
 Helper.setHome(home);
+
+// Merge config key-values passed as CLI options into the main config
+_.merge(Helper.config, program.config);
 
 require("./start");
 require("./config");
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -11,12 +11,12 @@ const Helper = require("../helper");
 const Utils = require("./utils");
 
 if (require("semver").lt(process.version, "6.0.0")) {
-	log.warn(`Support of Node.js v4 is ${colors.bold("deprecated")} and will be removed in The Lounge v3.`);
+	log.warn(`Support of Node.js v4 is ${colors.bold.red("deprecated")} and will be removed in The Lounge v3.`);
 	log.warn("Please upgrade to Node.js v6 or more recent.");
 }
 
 program.version(Helper.getVersion(), "-v, --version")
-	.option("--home <path>", `${colors.bold("[DEPRECATED]")} Use the ${colors.green("THELOUNGE_HOME")} environment variable instead.`)
+	.option("--home <path>", `${colors.bold.red("[DEPRECATED]")} Use the ${colors.green("THELOUNGE_HOME")} environment variable instead.`)
 	.option(
 		"-c, --config <key=value>",
 		"override entries of the configuration file, must be specified for each entry that needs to be overriden",
@@ -26,7 +26,7 @@ program.version(Helper.getVersion(), "-v, --version")
 	.parseOptions(process.argv);
 
 if (program.home) {
-	log.warn(`${colors.green("--home")} is ${colors.bold("deprecated")} and will be removed in The Lounge v3.`);
+	log.warn(`${colors.green("--home")} is ${colors.bold.red("deprecated")} and will be removed in The Lounge v3.`);
 	log.warn(`Use the ${colors.green("THELOUNGE_HOME")} environment variable instead.`);
 }
 
@@ -43,7 +43,7 @@ if (!fs.existsSync(path.join(
 }
 
 if (process.env.LOUNGE_HOME) {
-	log.warn(`${colors.green("LOUNGE_HOME")} is ${colors.bold("deprecated")} and will be removed in The Lounge v3.`);
+	log.warn(`${colors.green("LOUNGE_HOME")} is ${colors.bold.red("deprecated")} and will be removed in The Lounge v3.`);
 	log.warn(`Use ${colors.green("THELOUNGE_HOME")} instead.`);
 }
 
@@ -67,7 +67,7 @@ require("./install");
 
 // TODO: Remove this when releasing The Lounge v3
 if (process.argv[1].endsWith(`${require("path").sep}lounge`)) {
-	log.warn(`The ${colors.red("lounge")} CLI is ${colors.bold("deprecated")} and will be removed in v3.`);
+	log.warn(`The ${colors.red("lounge")} CLI is ${colors.bold.red("deprecated")} and will be removed in v3.`);
 	log.warn(`Use ${colors.green("thelounge")} instead.`);
 	process.argv[1] = "thelounge";
 }
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -22,12 +22,13 @@ program.version(Helper.getVersion(), "-v, --version")
 		"override entries of the configuration file, must be specified for each entry that needs to be overriden",
 		Utils.parseConfigOptions
 	)
-	.on("--help", Utils.extraHelp)
-	.parseOptions(process.argv);
+	.on("--help", Utils.extraHelp);
+
+// Parse options from `argv` returning `argv` void of these options.
+const argvWithoutOptions = program.parseOptions(process.argv);
 
 if (program.home) {
-	log.warn(`${colors.green("--home")} is ${colors.bold.red("deprecated")} and will be removed in The Lounge v3.`);
-	log.warn(`Use the ${colors.green("THELOUNGE_HOME")} environment variable instead.`);
+	log.warn(`${colors.bold("--home")} is ${colors.bold.red("deprecated")} and will be removed in The Lounge v3. Use the ${colors.bold("THELOUNGE_HOME")} environment variable instead.`);
 }
 
 // Check if the app was built before calling setHome as it wants to load manifest.json from the public folder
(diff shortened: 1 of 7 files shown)


src/command-line/utils.js
```diff
@@ -51,6 +52,54 @@ class Utils {
...
+		const parseValue = (value) => {
+			if (value === "true") {
+				return true;
+			} else if (value === "false") {
+				return false;
+			} else if (value === "undefined") {
+				return undefined;
+			} else if (value === "null") {
+				return null;
+			} else if (/^\[.*\]$/.test(value)) { // Arrays
```
xPaw: Could we just `JSON.parse` arrays and objects?

xPaw:
Should we ship this in 2.x (minus the removal of existing arguments) to test it and give deprecation notice?

src/command-line/utils.js
astorije: I guess. The `if` would be the same though (otherwise I have to `try`/`catch` for strings, no thanks). I'll try with that, not sure if I like that yet. It's silly to re-write my own parser when `JSON.parse` exists, but I also don't want to give too much power to `-c`.

Also, I don't really want to support objects. Arrays are useful (for example for `transports`), but if you want to pass an object, use the dot-notation keys. I don't want to support everything and spend my life dealing with edge cases lol.

astorije:
> Should we ship this in 2.x (minus the removal of existing arguments) to test it and give deprecation notice?

Good call, I'll update my PR to do so.

src/command-line/utils.js
xPaw: I was thinking JSON.parse would be a single line/call compared to what you are doing now.

astorije: You can't, because `JSON.parse` explodes on strings, except when they are `true`, `false`, or `null`. Ah, JavaScript...

astorije:
I'll make my PR work with Node v4 then.

src/command-line/utils.js
astorije: Alright, so I've played with this, and unfortunately I can't use `JSON.parse` for that for a few reasons:

- When passing `-c foo=true`, `JSON.parse(value)` correctly returns `true`. But if I pass `-c foo=bar`, it explodes. If I go the safe route and add quotes to all values, then ``JSON.parse(`"${value}"`)`` works for the latter, but the former is treated as `"true"` so I end up... having to re-check all special cases like `true`/`false`/etc. and even arrays lol (only I don't have to check for `null`, yay).
- I can't expect users to pass valid, quoted JSON for strings such as `-c foo="bar"` because, well first off it's lame and too easy to forget (programs never require that, but also  when I get the config in the code, quotes have been removed at shell level, so I get `foo=bar`. Back to square one. We might have to require users to do `-c foo=\"bar\"` which is just horrible.
- I was hoping to be able to do that for arrays at the very least but... `-c foo=[bar,baz]` has the exact same issue, `[bar,baz]` is not valid JSON.

So in the end, the only thing I can do is make sure all cases are correctly tested, which I think is what I'm doing :man_shrugging: Let me know if you think I forgot cases.

astorije:
screen shot 2017-12-10 at 15 58 35

screen shot 2017-12-10 at 15 58 12

astorije:
OMG I am spending my entire weekend on this PR...

@xPaw, managed to almost make the `-c` option solve the problem you described in #1807...

**`thelounge -h` with a config file saying `public: true`:**

screen shot 2017-12-10 at 16 08 09

**`thelounge -h -c public=false` with a config file saying `public: true`:**

screen shot 2017-12-10 at 16 17 28

Buuuuuuuut I have one more issue.
As you can see, running `thelounge -h -c public=false` says:

```js
2017-12-10 21:07:58 [WARN] Configuration key public was already specified, igoring...
```

`-c public=false` gets parsed twice. Not just that, but the second time it's parsed, the `memo` object contains previous values.
I tried sooo many things and I am not sure what's going on here. The easy solution could be to remove that warning, and just silently do nothing when giving `-c public=true -c public=false`, but I'd rather not do that. I'll spend 10 more minutes on that but then I give up, it's driving me crazy. If you have any idea of what's going on (I've been looking at commander's code for far too long), I'd love to hear it lol.

astorije:
Wooooot I finally fixed it.
We were running `.parseOptions(process.argv)`, parsing options a first time, then `parse(process.argv)`, also parsing options. It turns out `parseOptions` [returns `argv` without the options already parsed](https://github.com/tj/commander.js/blob/fefda77f4632922b6b6fa68202b7b021e9bdd270/index.js#L687-L688), but unfortunately for me it [splits the result in 2, arguments and options](https://github.com/tj/commander.js/blob/fefda77f4632922b6b6fa68202b7b021e9bdd270/index.js#L762) so I had to re-concatenate them...

With the problematic example above (`thelounge -h -c public=false`, this is how things look like:

```js
# process.argv
[ '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/node',
  '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/thelounge',
  '-h',
  '-c',
  'public=false' ]

# argvWithoutOptions.args.concat(argvWithoutOptions.unknown)
[ '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/node',
  '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/thelounge',
  '-h' ]
```

With a more complex example  (`thelounge -c public=false -h --home /tmp --foo bar start -P 1234`):

```js
# process.argv
[ '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/node',
  '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/thelounge',
  '-c',
  'public=false',
  '-h',
  '--home',
  '/tmp',
  '--foo',
  'bar',
  'start',
  '-P',
  '1234' ]

# argvWithoutOptions.args.concat(argvWithoutOptions.unknown)
[ '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/node',
  '/Users/jeremie/.nvm/versions/node/v6.11.4/bin/thelounge',
  'start',
  '-h',
  '--foo',
  'bar',
  '-P',
  '1234' ]
```

so we should be fine.

I have wasted soooo much time on this PR. I need to not be in front of a computer now :)
@xPaw, I hope you like how things look now! ❤️

src/command-line/utils.js
```diff
@@ -51,6 +52,55 @@ class Utils {
...
+
+		// Note: If passed `-c foo="bar=42"` (with single or double quotes), `val`
+		//       will always be passed as `foo=bar=42`, never with quotes.
+		const position = val.indexOf("="); // Only split on the first = found
+		const key = val.slice(0, position);
+		const value = val.slice(position + 1);
+		const parsedValue = parseValue(value);
+
+		if (_.has(memo, key)) {
+			log.warn(`Configuration key ${colors.bold(key)} was already specified, igoring...`);
```
dgw: `s/igoring/ignoring/` assuming that this will be in the final merge.

astorije:
Great catch, will fix, thanks!
-- 
Jérémie Astori
https://jeremie.astori.fr

src/command-line/utils.js
astorije: Fixed, thanks!

src/command-line/index.js
```diff
@@ -58,12 +68,12 @@ require("./install");
...
 // TODO: Remove this when releasing The Lounge v3
 if (process.argv[1].endsWith(`${require("path").sep}lounge`)) {
-	log.warn(`The ${colors.red("lounge")} CLI is ${colors.bold("deprecated")} and will be removed in v3.`);
+	log.warn(`The ${colors.red("lounge")} CLI is ${colors.bold.red("deprecated")} and will be removed in v3.`);
 	log.warn(`Use ${colors.green("thelounge")} instead.`);
 	process.argv[1] = "thelounge";
 }
 
-program.parse(process.argv);
+program.parse(argvWithoutOptions.args.concat(argvWithoutOptions.unknown));
```
xPaw: Add a comment here, it's not very clear where `unknown` comes from.

astorije: Done, thanks for pointing this out!

astorije:
FYI, if anyone is interested, I have written the docs around `--config` in the new docs I'm working on. It's WIP, but can be seen at https://github.com/thelounge/thelounge.github.io/blob/680e9aec389215b7120491f9f38189e07b4ca4f6/_docs/usage.md#configuring-the-lounge (no website styling but GitHub does not support Pages on PRs/branches, and I still haven't configured an alternative way to do so).

PR #1829, opened 2017-12-12
Use better icons for channel/query list and context menu actions

astorije:
Follow-up of #1824.

I did not do a great job at all with #1824:

- Some people pointed out the duplication of the hash character ("**#** #thelounge") in channel list and window title. Even more so with channels wtarting with `##`. It also is unrealistic to strip the first `#` of the channel because it becomes even more confusing, with all those different channels: `#channel`, `##channel`, `&channel`, etc.

- The :hash: and 👤 icons do not have the same width, so channel names and query names are now misaligned in the channel list. Good job me 👏

- I mistakenly attempted to make the 👤 icon consistently used to represent a user and a query window, but that just doesn't work. #1722 showed that even more clearly, because 👤 is used for the user name, and 💬 is used for the query window, which is correct.

Anyway, I'm making a new attempt with this PR. Channel windows are consistently using http://fontawesome.io/icon/comment/, query windows are consistently using http://fontawesome.io/icon/comment-o/, and users are still correctly using http://fontawesome.io/icon/user/. Results below:

Channel window / action | User menu | Query window / action
--- | --- | ---
screen shot 2017-12-12 at 00 51 05 screen shot 2017-12-12 at 00 51 15 | screen shot 2017-12-12 at 00 51 22 | screen shot 2017-12-12 at 00 51 32 screen shot 2017-12-12 at 00 51 40

I think it works better than what we originally had (and was buggy in context menus), and definitely better than my attempt in #1824. I'm open to suggestions obviously if people think these icons are 💩.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -221,22 +221,23 @@ kbd {
 #chat button.menu::before { content: "\f142"; /* http://fontawesome.io/icon/ellipsis-v/ */ }
 
 .context-menu-user::before { content: "\f007"; /* http://fontawesome.io/icon/user/ */ }
-.context-menu-chan::before { content: "\f292"; /* http://fontawesome.io/icon/hashtag/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 .context-menu-list::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
 .context-menu-action-whois::before { content: "\f05a"; /* http://fontawesome.io/icon/info-circle/  */ }
-.context-menu-action-query::before { content: "\f0e6"; /* http://fontawesome.io/icon/comments-o/ */ }
 .context-menu-action-kick::before { content: "\f05e"; /* http://fontawesome.io/icon/ban/ */ }
 
 .context-menu-network::before,
 #sidebar .chan.lobby::before,
 #chat .lobby .title::before { content: "\f0a0"; /* http://fontawesome.io/icon/hdd-o/ */ }
 
+.context-menu-query::before,
+.context-menu-action-query::before,
 #sidebar .chan.query::before,
-#chat .query .title::before { content: "\f007"; /* http://fontawesome.io/icon/user/ */ }
+#chat .query .title::before { content: "\f0e5"; /* http://fontawesome.io/icon/comment-o/ */ }
 
+.context-menu-chan::before,
 #sidebar .chan.channel::before,
-#chat .channel .title::before { content: "\f292"; /* http://fontawesome.io/icon/hashtag/ */ }
+#chat .channel .title::before { content: "\f086"; /* http://fontawesome.io/icon/comments/ */ }
 
 #sidebar .chan.special::before,
 #chat .special .title::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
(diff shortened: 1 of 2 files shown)


christer88:
I'm not 100 % sure about the icons for channel and pm's. They are a bit too similar to each other. Would be good to separate them a bit more. Besides this, thumbs up!

astorije:
I was actually thinking that their visual proximity had a nice effect to it, and the plain vs. empty was making them visually very distinct (plus no risk to have different widths!) but you have a point and as said I am not married to those.

Is there anything in http://fontawesome.io/icons/ that you might find appropriate?

christer88:
I suggest http://fontawesome.io/icon/hashtag/ for channels and http://fontawesome.io/icon/comments/ for pm. Hashtag is what most think of regarding IRC channels, and the comments icon shows theres two speech bubbles, indicating there's a communication between two people.

dgw:
> I suggest http://fontawesome.io/icon/hashtag/ for channels

That is literally the first problem @astorije listed with the last icon revision. Having `#` for an icon right next to `#channel` doesn't look very good.

astorije:
Yeah, I don't mean to bikeshed or overthink this. Whatever icons get picked, I don't care that much, I just wanted to fix the alignment issue and the iconsistencies/errors in icons between the context menus.

@YaManicKill or @xPaw, would be nice if you can run this PR on top of your instance (I know you use ZNC) and see if it bothers you. I'm happy to change to anything, as long as the icons don't have that width issue. Your call :)

YaManicKill:
> @YaManicKill or @xPaw, would be nice if you can run this PR on top of your instance (I know you use ZNC) and see if it bothers you.

I actually don't see the icons, I think solarized theme hides them, and I don't miss them, so...

dgw:
The one change I would make is using http://fontawesome.io/icon/comments/ for channels instead of http://fontawesome.io/icon/comment/. It would be less visually confusing than differentiating channels and queries solely based on whether the icon shape is filled in.

astorije:
screen shot 2017-12-15 at 01 10 54 vs. screen shot 2017-12-15 at 01 11 07 vs. screen shot 2017-12-15 at 01 11 44

<p align=center>
@dgw<br>
@thelounge/maintainers<br>
❓ 
</p>

dgw:
@astorije Option 1 for me, maybe even use `comment` instead of `comment-o` for queries.

astorije:
<p align=center>screen shot 2017-12-15 at 01 47 15<br>
❓ </p>

dgw:
@astorije Hmm… Seeing it mocked up, the `comment-o` is probably better. `comment` is too heavy and draws attention to the query even though it's not focused.

Another idea: Use the `-o` icons for both, but switch to the solid variant for the active window. Might be too busy, though… We already have a highlight color to indicate `.active`.

astorije:
Alright then, option 1 is it. FYI @xPaw and @MaxLeiter also +1ed this one.

> Another idea: Use the -o icons for both, but switch to the solid variant for the active window. Might be too busy, though… We already have a highlight color to indicate .active.

Mmmh I don't know. Definitely not for this PR, I'll let someone else try if they want :D Right now this is good enough, and it fixes the current issues so we can move forward :)

Thanks for your help @dgw!

PR #1817, opened 2017-12-09
Support video file previews

MaxLeiter:
will rebase on #1806 if it's merged in first (assume it will be)

screen shot 2017-12-09 at 3 25 59 pm
screen shot 2017-12-09 at 3 25 52 pm

PR #1832, opened 2017-12-12
Rename thumbnail/image option to more general `media`

MaxLeiter:
For #1806 and #1817, mainly.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -44,7 +44,7 @@ userOptions = null;
 module.exports = options;
 
 module.exports.shouldOpenMessagePreview = function(type) {
-	return (options.links && type === "link") || (options.media && (type === "image" || type === "audio"));
+	return type === "link" ? options.links : options.media;
 };
 
 module.exports.initialize = () => {
--- a/client/views/windows/settings.tpl
+++ b/client/views/windows/settings.tpl
@@ -74,8 +74,8 @@
 		</div>
 		<div class="col-sm-6">
 			<label class="opt">
-				<input type="checkbox" name="thumbnails">
-				Auto-expand images
+				<input type="checkbox" name="media">
+				Auto-expand media
 			</label>
 		</div>
 		<div class="col-sm-6">


defaults/config.js
```diff
@@ -87,15 +87,15 @@ module.exports = {
...
 	// If prefetch is enabled, The Lounge will only display content under the maximum size.
-	// Specified value is in kilobytes. Default value is 2048 kilobytes.
+	// Specified value is in kilobytes. Default value is 4096 kilobytes.
 	//
 	// @type     int
-	// @default  2048
+	// @default  4096
 	//
-	prefetchMaxImageSize: 2048,
+	prefetchMaxMediaSize: 4096,
```
astorije: Instead of removing it, add `prefetchMaxMediaSize` on top of `prefetchMaxImageSize`. If `prefetchMaxImageSize` is present in the config, use it and warn that it is deprecated, otherwise use `prefetchMaxMediaSize`.
That way we don't break BC until v3, and we encourage people to switch :)

astorije: I know it's slightly more complicated, but that's the cost of stability 😁

src/server.js
```diff
@@ -58,6 +58,10 @@ module.exports = function() {
 		log.warn("Server is public and set to use LDAP. Set to private mode if trying to use LDAP authentication.");
 	}
 
+	if (config.prefetchMaxImageSize) { // TODO: Remove in v3.0.0
+		log.warn("prefetchMaxImageSize has been removed as of v2.7.0. Use the new prefetchMaxMediaSize option instead.");
```
astorije: Just to try to be more consistent (it's useful to search-entire-project for "deprecated", though I know there are 2 cases that don't match the other 10ish lol) , what do you think of the following?

```js
log.warn(`$colors.bold("prefetchMaxImageSize") is ${colors.bold.red("deprecated")} and will be removed in The Lounge v3. Use ${colors.bold("prefetchMaxMediaSize")} option instead.`);
```

defaults/config.js
astorije: Either that or the 2.7.0 PR contains just your first commit, but I think it's nice to give users some time to adapt peacefully lol, + it's just a matter of `config.prefetchMaxMediaSize = config.prefetchMaxImageSize;` within your `if` in `src/server.js`.

MaxLeiter: well, it should still be removed from default config.js, right? Will make the swap elsewhere, though.

astorije: Yes indeed, good point. So to keep bc you just need setting max media if there is a max image set

src/server.js
MaxLeiter: Done

defaults/config.js
MaxLeiter: Done

MaxLeiter:
Changes made, @astorije

defaults/config.js
xPaw: I don't really see a reason to rename this since it's still going to be used only by images. Let's keep it this way until we actually need it for other purposes.

client/js/options.js
```diff
@@ -44,7 +44,7 @@ userOptions = null;
 module.exports = options;
 
 module.exports.shouldOpenMessagePreview = function(type) {
-	return (options.links && type === "link") || (options.thumbnails && type === "image");
+	return type === "link" ? options.links : options.media;
```
YaManicKill: Oh that's much cleaner. The former was so complicated I wasn't sure this did what it was meant to do :-P

src/server.js
```diff
@@ -58,6 +58,11 @@ module.exports = function() {
 		log.warn("Server is public and set to use LDAP. Set to private mode if trying to use LDAP authentication.");
 	}
 
+	if (config.prefetchMaxImageSize) { // TODO: Remove in v3.0.0
+		log.warn(`${colors.bold("prefetchMaxImageSize")} is ${colors.bold.red("deprecated")} and will be removed in The Lounge v3.0.0. Use the ${colors.bold("prefetchMaxMediaSize")} option instead.`);
+		config.prefetchMaxMediaSize = config.prefetchMaxImageSize;
```
YaManicKill: Should we maybe do

`config.prefetchMaxMediaSize = config.prefetchMaxMediaSize || config.prefetchMaxImageSize`

Just incase any users have the media size set as well as the image size? I'm not sure what the best thing to do it, here.

defaults/config.js
YaManicKill: It's used by the audio and video in #1806 and #1817

xPaw: It's actually not used by these because we came to the conclusion it's not need for 2 reasons:

1. Server aborts connection right after getting content-type.
2. Client only preloads metadata about the media so size doesn't matter in these cases.

YaManicKill: Oh, hmm, I missed that bit happening, but yeah, you are right.

Alright, I'm happy with either this going in or not going in. I guess it could be misleading if we have it as "media" when it's only images. But maybe users would think a gif isn't an image.

src/server.js
MaxLeiter: That seems good. Will do

xPaw: See above, there's currently no reason to change the server option (and it's misleading to change since it still will only affect images).

astorije:
@YaManicKill, I removed the second commit per @MaxLeiter's instructions and fixed conflicts. Feel free to merge when you're happy with it so we can fix `master`.

YaManicKill:
Alright, so we are now just changing the frontend option to expand audio/video as well. Gotcha.

PR #1833, opened 2017-12-13
Remove lobby close button

Cldfire:
Where `remove` means `display: none`.

Prompted by discussion on IRC.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -444,9 +444,9 @@ $(function() {
 		}
 	});
 
-	sidebar.on("click", ".close", function() {
+	function closeChan(chan) {
 		var cmd = "/close";
-		var chan = $(this).closest(".chan");
+
 		if (chan.hasClass("lobby")) {
 			cmd = "/quit";
 			var server = chan.find(".name").html();
@@ -463,11 +463,15 @@ $(function() {
 			opacity: 0.4,
 		});
 		return false;
+	}
+
+	sidebar.on("click", ".close", function() {
+		closeChan($(this).closest(".chan"));
 	});
 
 	const contextMenuActions = {
 		close: function(itemData) {
-			$(`.networks .chan[data-target="${itemData}"] .close`).click();
+			closeChan($(`.networks .chan[data-target="${itemData}"]`));
 		},
 		focusChan: function(itemData) {
 			$(`.networks .chan[data-target="${itemData}"]`).click();
--- a/client/views/chan.tpl
+++ b/client/views/chan.tpl
@@ -1,7 +1,7 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-	<button class="close" aria-label="Close"></button>
+	{{#equal type "lobby"}}{{else}}<button class="close" aria-label="Close"></button>{{/equal}}
 	<span class="name" title="{{name}}">{{name}}</span>
 </div>
 {{/each}}


MaxLeiter:
Am a 👎 . No way to close lobby's via GUI on mobile.

astorije:
I suggested the idea so I'm obviously 👍, but as I suggested it, I don't feel like I have a right to approve the PR :)

Rationale behind this change is:

- As opposed to channels and queries where it's not uncommon to open/close at will, disconnecting from a network is a much more rare operation.
- It still happens to hit close by accident. We drastically reduced that risk when we made this icon not reactive to mouse hover, but the effects of disconnecting from a vs. closing a channel/query are much more annoying (also, when loading messages happens, closing/re-joining a channel will be almost seamless.
- It is still possible to disconnect from a network via the lobby actions:
  screen shot 2017-12-13 at 01 13 55
- @MaxLeiter is currently experimenting with adding a "join" button in the GUI, and the best results he has so far gets me concerned about the proximity in location and shape of the "join a channel" and "disconnect from network" buttons:
  image

YaManicKill:
I'm generally positive about the idea (the idea that you can so easily disconnect from the network is ... crazy).

But why are we using display:none rather than just ... not creating the thing in the first place?

Cldfire:
Well, the client creates a single close button for channels / queries / lobbies (as seen here), so it was the quick and easy solution.

If you would prefer for the template to be modified, I could take a stab at doing that sometime.

YaManicKill:
> If you would prefer for the template to be modified, I could take a stab at doing that sometime.

If you don't mind, I would. It should just be a case of only printing it if the type isn't lobby.

xPaw:
I think dropdown "disconnect" searches for that close button, so that might break.

astorije:
> It should just be a case of only printing it if the type isn't lobby.

I honestly don't know why I didn't think of that... Sorry for misleading you, @Cldfire :man_facepalming:

astorije:
> I think dropdown "disconnect" searches for that close button, so that might break.

Indeed.

[This](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L138) defines an action when clicking on the dropdown entry.
Then what happens when triggering this action [is done here](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L469-L471). It emits a click on the <kbd>&times;</kbd> button, which is then handled [by this snippet](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L447-L466).

@Cldfire, as a solution, I would try this:

1. Move the content of this handler into its own function:

  ```js
  function closeChan() {
    var cmd = "/close";
    var chan = $(this).closest(".chan");
    /* ... */
    return false;
  }
  ```

2. Update the handler to be `sidebar.on("click", ".close", closeChan);`

3. Update [the action](https://github.com/thelounge/lounge/blob/master/client/js/lounge.js#L470) to be `closeChan();` so that it doesn't try to click on the button anymore.

I might be wrong, but that should work and be decent enough for what you are trying to achieve. We have done such conversion (from emitting a click to calling a function) in the past actually. Not sure which one anymore, I feel like it was joining a channel or something.

Of course, I totally understand that it can be frustrating that a seemingly one-liner becomes much more than that, so no worries if you don't want to bother with this, someone else will pick it up ;)
Regardless of what you decide, thank you very much for bringing this up and opening a PR about it!

Cldfire:
> I honestly don't know why I didn't think of that... Sorry for misleading you, @Cldfire 🤦‍♂️

No problem 😃 

> if you don't want to bother with this, someone else will pick it up ;)

I should be able to take care of this tonight or tomorrow 😄 thanks for the instructions!

Cldfire:
Okay, I think that should do the trick (or at least, it does locally).

I'm sure I probably missed something or did something in a non-idiomatic way, so criticize away 👍

client/js/lounge.js
```diff
@@ -444,9 +444,16 @@ $(function() {
 		}
 	});
 
-	sidebar.on("click", ".close", function() {
+	function closeChan(itemData) {
```
astorije: Call this `target`. At this point in the code (i.e. not in generic `contextMenuActions` actions anymore), it's fine to re-specialize and makes the code easier to follow.

client/views/chan.tpl
```diff
@@ -1,7 +1,7 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-	<button class="close" aria-label="Close"></button>
+	{{#if needsCloseButton}}<button class="close" aria-label="Close"></button>{{/if}}
```
astorije: Test that type is lobby instead, that way you don't have to add `needsCloseButton` everywhere on the server-side (it's a client/front-end implementation detail, so the server doesn't need to know about it).

client/js/lounge.js
```diff
@@ -444,9 +444,16 @@ $(function() {
...
 		var cmd = "/close";
-		var chan = $(this).closest(".chan");
+		var chan;
+
+		if (typeof itemData === "undefined") {
+			chan = $(this).closest(".chan");
```
astorije: In which situation can this be `undefined` actually?
(Also, we have a bunch of `=== undefined` in the code, I think you can go for that. I know, I know, what about if `undefined` gets reassigned, I've heard the urban legends :D)

astorije:
> I'm sure I probably missed something or did something in a non-idiomatic way, so criticize away 👍

No worries! And I'm not criticize, it's already super nice of you to help! :) Hope my comments make sense

client/js/lounge.js
Cldfire: Whoops, looks like it won't be, you're right.

I've refactored it locally to simply pass the channel into the function:

```js
close: function(itemData) {
	closeChan($(`.networks .chan[data-target="${itemData}"]`));
},
```

I have one bug to track down before I commit, though, and I don't have time to do that right now, so it'll have to wait.

client/views/chan.tpl
Cldfire: Yeah, that's the first way I tried doing it. When I looked it up, however, I read that handlebars templates are supposed to be logicless and therefore there are no equality operators, so I couldn't find a way to do that.

If you could point me in the right direction that would be great 😃

astorije: I think that something like [this](https://github.com/thelounge/lounge/blob/master/client/views/chat.tpl#L5) (except not `"channel"`) would work.

Cldfire: Ah, alright, that makes sense. Thanks!

I should be able to wrap this up on Sunday.

client/js/lounge.js
astorije: If you want help with your bug, feel free to commit and describe it so I can reproduce + debug :)

Cldfire:
Comments have been addressed.

client/js/lounge.js
Cldfire: I figured it out 👍

Cldfire:
Squashed 👍

astorije:
@xPaw or @YaManicKill are you good with this?

client/views/chan.tpl
```diff
@@ -1,7 +1,7 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-	<button class="close" aria-label="Close"></button>
+	{{#equal type "lobby"}}{{else}}<button class="close" aria-label="Close"></button>{{/equal}}
```
YaManicKill: Euch, can we drop in another handlebars helper that is the opposite of equal so we don't have to have this weird thing?

YaManicKill:
Just 1 small comment.

client/views/chan.tpl
astorije: @YaManicKill, agreed, but do you mind if we merge as-is instead of asking @Cldfire to add a Handlebars helper for that?
If you're okay with that, then after we merge I'll add such helper with tests for it, also add tests for `equal`. I can do that today, right after merging.

I thought `diff` was meant for that but it's actually for... something else...
At some point, we might want to use something like https://github.com/helpers/handlebars-helpers but right now we don't need it so we can look into that when we start adding more generic helpers. Also I couldn't make it work with Webpack lol.

astorije: @YaManicKill, done at #1874. Whichever PR goes first, I'll adapt the other accordingly. Feel free to review/approve both and I'll remember to do the conversion before merging.

YaManicKill: Alrighty

Issue #1219, opened 2017-06-11
no UI for joining a new channel


lelutin:
Hello,

the lounge's UI is generally quite enjoyable. there's one thing that I'm sure can be unsettling for users that aren't already used to IRC: there's no UI for joining a new channel.

I'd like to suggest either having some + button in the left section near/under a network's channel list, or have a "join channel" option in the channel menu shown under three vertical dots (top right).

this addition would make users that are not used to typing commands with / more comfortable to start using the lounge for most basic uses.

PR #1836, opened 2017-12-13
Join channel UI

MaxLeiter:
Resolves https://github.com/thelounge/lounge/issues/1219.
screen shot 2017-12-20 at 4 43 14 pm
screen shot 2017-12-20 at 4 43 09 pm

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -222,6 +222,7 @@ kbd {
 #viewport .rt::before { content: "\f0c0"; /* http://fontawesome.io/icon/users/ */ }
 #chat button.menu::before { content: "\f142"; /* http://fontawesome.io/icon/ellipsis-v/ */ }
 
+.context-menu-join::before { content: "\f067"; /* http://fontawesome.io/icon/plus/ */ }
 .context-menu-user::before { content: "\f007"; /* http://fontawesome.io/icon/user/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 .context-menu-list::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
@@ -457,7 +458,6 @@ kbd {
 	will-change: transform;
 }
 
-#sidebar button,
 #sidebar .chan,
 #sidebar .sign-out,
 #sidebar .empty {
@@ -551,6 +551,7 @@ kbd {
 }
 
 #sidebar .badge,
+#sidebar .add-channel,
 #sidebar .close {
 	float: right;
 	margin-left: 5px;
@@ -592,6 +593,32 @@ kbd {
 	text-align: center;
 	content: "×";
 	color: #fff;
+}
+
+#sidebar .lobby .add-channel {
+	border-radius: 3px;
+	width: 18px;
+	height: 18px;
+	transition: opacity 0.2s, background-color 0.2s;
+}
+
+#sidebar .lobby .add-channel::before {
+	font-size: 20px;
+	font-weight: normal;
+	display: inline-block;
+	line-height: 16px;
+	text-align: center;
+	content: "+";
+	color: #fff;
+}
+
+#sidebar .lobby .add-channel {
+	opacity: 0.4;
+}
+
+#sidebar .lobby .add-channel:hover {
+	background-color: rgba(0, 0, 0, 0.1);
+	opacity: 1;
 }
 
 #sidebar .chan.active .close {
@@ -685,7 +712,7 @@ kbd {
 	font-size: 14px;
 }
 
-#windows .input {
+.input {
 	background-color: white;
 	border: 1px solid #cdd3da;
 	border-radius: 2px;
@@ -892,6 +919,35 @@ kbd {
 	width: 180px;
 	transition: right 0.4s;
 	touch-action: pan-y;
+}
+
+
+/**
+  * Toggled via JavaScript
+  */
+#sidebar .join-form {
+	display: none;
+}
+
+#sidebar .join-form .input {
+	display: block;
+	margin-left: auto;
+	margin-right: auto;
+	margin-top: 5px;
+	margin-bottom: 5px;
+	width: 80%;
+}
+
+#sidebar .join-form .btn {
+	display: block;
+	width: 80%;
+	padding: 1px;
+	margin: auto;
+	height: 29px;
+}
+
+#sidebar .add-channel-tooltip {
+	float: right;
 }
 
 #chat .show-more {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -598,7 +598,8 @@ kbd {
 	border-radius: 3px;
 	width: 18px;
 	height: 18px;
-	transition: opacity 0.2s, background-color 0.2s;
+	opacity: 0.4;
+	transition: opacity 0.2s, background-color 0.2s, transform 0.2s;
 }
 
 #sidebar .lobby .add-channel::before {
@@ -611,12 +612,12 @@ kbd {
 	color: #fff;
 }
 
-#sidebar .lobby .add-channel {
-	opacity: 0.4;
-}
-
 #sidebar .lobby .add-channel:hover {
 	opacity: 1;
+}
+
+#sidebar .lobby .add-channel.opened {
+	transform: rotate(45deg);
 }
 
 #sidebar .chan.active .close {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -578,7 +578,6 @@ kbd {
 }
 
 #sidebar .close {
-	border-radius: 3px;
 	width: 18px;
 	height: 18px;
 	display: none;
@@ -617,7 +616,6 @@ kbd {
 }
 
 #sidebar .lobby .add-channel:hover {
-	background-color: rgba(0, 0, 0, 0.1);
 	opacity: 1;
 }
 
@@ -627,7 +625,6 @@ kbd {
 }
 
 #sidebar .chan.active .close:hover {
-	background-color: rgba(0, 0, 0, 0.1);
 	opacity: 1;
 }
 
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -483,35 +483,6 @@ $(function() {
 		closeChan($(this).closest(".chan"));
 	});
 
-	sidebar.on("click", ".add-channel", (e) => {
-		const id = $(e.target).data("id");
-		const joinForm = $(`#join-channel-${id}`);
-		joinForm.toggle();
-		joinForm.find(".input[name='channel']").focus();
-		return false;
-	});
-
-	sidebar.on("submit", ".join-form", function() {
-		const form = $(this);
-		const channel = form.find("input[name='channel']");
-		const channelString = channel.val();
-		const key = form.find("input[name='key']");
-		const keyString = key.val();
-		const chan = utils.findCurrentNetworkChan(channelString);
-		if (chan.length) {
-			chan.click();
-		} else {
-			socket.emit("input", {
-				text: `/join ${channelString} ${keyString}`,
-				target: form.prev().data("id"),
-			});
-		}
-		channel.val("");
-		key.val("");
-		form.hide();
-		return false;
-	});
-
 	const contextMenuActions = {
 		join: function(itemData) {
 			$(`#join-channel-${itemData}`).show();
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -20,6 +20,7 @@ const utils = require("./utils");
 require("./webpush");
 require("./keybinds");
 require("./clipboard");
+const JoinChannel = require("./join-channel");
 
 $(function() {
 	var sidebar = $("#sidebar, #footer");
@@ -485,7 +486,8 @@ $(function() {
 
 	const contextMenuActions = {
 		join: function(itemData) {
-			$(`#join-channel-${itemData}`).closest(".network").find("button.add-channel").click();
+			const network = $(`#join-channel-${itemData}`).closest(".network");
+			JoinChannel.openForm(network);
 		},
 		close: function(itemData) {
 			closeChan($(`.networks .chan[data-target="${itemData}"]`));
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -485,7 +485,7 @@ $(function() {
 
 	const contextMenuActions = {
 		join: function(itemData) {
-			$(`#join-channel-${itemData}`).show();
+			$(`#join-channel-${itemData}`).closest(".network").find("button.add-channel").click();
 		},
 		close: function(itemData) {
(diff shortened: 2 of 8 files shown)


client/views/chan.tpl
```diff
@@ -1,7 +1,13 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
+	{{#equal type "lobby"}}
+		<button class="add" aria-label="Add channel" data-id="{{id}}"></button>
```
xPaw: use something less generic than `add`

client/views/join_channel.tpl
```diff
@@ -0,0 +1,5 @@
+<div id="join_channel_{{id}}" class="join_form">
+	<input type="text" class="input" id="joinchan_{{id}}:channel" name="channel" placeholder="#channel">
```
xPaw: You don't really need all these ids, since the event handler you can just do `$(this).find("input[name=...")` (or on classname if you add one)

client/js/lounge.js
```diff
@@ -465,6 +465,23 @@ $(function() {
 		return false;
 	});
 
+	sidebar.on("click", ".add", (e) => {
+		e.preventDefault();
+		const id = $(e.target).data("id");
```
xPaw: Same here with ids, you can just go to parent and find the relevant form. I'm suggesting this because we don't use this `id` approach anywhere else.

client/js/lounge.js
```diff
@@ -465,7 +471,35 @@ $(function() {
...
+		const form = $(this);
+		const channel = form.find("input[name='channel']").val();
+		const key = form.find("input[name='key']").val();
+		const chan = utils.findCurrentNetworkChan(channel);
+		if (chan.length) {
+			chan.click();
+		} else {
+			socket.emit("input", {
+				text: `/join ${channel} ${key}`,
+				target: $("#chat").data("id"),
```
xPaw: Don't you need to find the id of the network where the form is?

client/js/lounge.js
```diff
@@ -465,7 +471,35 @@ $(function() {
...
+			});
+		}
+		form.toggle();
```
xPaw: Use `hide` instead.

client/js/lounge.js
```diff
@@ -465,7 +471,35 @@ $(function() {
...
+		return false;
+	});
+
 	const contextMenuActions = {
+		join: function(itemData) {
+			$(`#join_channel_${itemData}`).toggle();
```
xPaw: And `show` here.

client/js/lounge.js
```diff
@@ -465,7 +471,35 @@ $(function() {
...
 
+	sidebar.on("click", ".add-channel", (e) => {
+		e.preventDefault();
+		const id = $(e.target).data("id");
+		const joinForm = $(`#join_channel_${id}`);
+		joinForm.toggle();
+	});
+
+	sidebar.on("submit", ".join_form", function(e) {
+		e.preventDefault();
```
xPaw: You don't need this as it's being done by `return false` in jquery.

client/js/lounge.js
MaxLeiter: When you press the add button multiple times it feels natural for it to toggle it imo

xPaw: This is submit handler, not the `+` button click.

client/views/chan.tpl
```diff
@@ -1,7 +1,14 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
+	{{#equal type "lobby"}}
+		<button class="add-channel" aria-label="Add channel" data-id="{{id}}"></button>
```
astorije: Could you add a tooltip that says "Join a channel..." here? I think join is more explicit than add, and ellipsis sort of because of [this](https://medium.com/@jsaito/the-mighty-ellipsis-6c2c00ddc864).

client/views/join_channel.tpl
```diff
@@ -0,0 +1,6 @@
+<form id="join_channel_{{id}}" class="join_form" method="post" action="" style="display: none;">
+	<input type="text" class="input" name="channel" placeholder="#channel" pattern="[^\s]+" maxlength="200" title="Should be a valid channel name">
+	<input type="password" class="input" name="key" placeholder="channel key" pattern="[^\s]+" title="Should be a valid channel key" maxlength="200" >
```
astorije: Can you add "(optional)" in the placeholder? New people might be wondering what a channel key is when they want to join.
Also I would just do "Name" (instead of "#channel") and "Key (optional)".

client/views/join_channel.tpl
```diff
@@ -0,0 +1,6 @@
+<form id="join_channel_{{id}}" class="join_form" method="post" action="" style="display: none;">
```
astorije: Instead of inline CSS, you should put that in style.css.

client/js/lounge.js
```diff
@@ -465,7 +471,38 @@ $(function() {
 		return false;
 	});
 
+	sidebar.on("click", ".add-channel", (e) => {
```
astorije: When I click on "+", that opened the network lobby, marked it as read, and therefore hovered the cursor on <kbd>&times;</kbd> to disconnect the network. Scary stuff lol

#1833 addressed the last point, but I think clicking "+" should not open the lobby. What do you think?

Also, could you try switching the badge and "+" button? I think I will prefer that the button doesn't move based on unread badge. Will see after trying out.

client/views/join_channel.tpl
```diff
@@ -0,0 +1,6 @@
...
+	<input type="text" class="input" name="channel" placeholder="#channel" pattern="[^\s]+" maxlength="200" title="Should be a valid channel name">
+	<input type="password" class="input" name="key" placeholder="channel key" pattern="[^\s]+" title="Should be a valid channel key" maxlength="200" >
+	<button type="submit" class="btn joinchan:submit" data-id="{{id}}">Join</button>
```
astorije: The button text does not have the correct color. I think we need to remove `#sidebar button` at https://github.com/thelounge/lounge/blob/master/client/css/style.css#L457 (just the line, not the whole block) but I just did a super quick test, nothing too involved.

client/js/lounge.js
dgw: > but I think clicking "+" should not open the lobby.

Right. It shouldn't. Probably as simple as an `e.preventDefault()` or however that works now.

xPaw: `return false` will do ;)

client/css/style.css
```diff
@@ -893,6 +920,30 @@ kbd {
 	touch-action: pan-y;
 }
 
+#sidebar .join_form {
+	display: none;
+}
+
+#sidebar .join_form .input {
```
astorije: I think most of this will become unnecessary if you remove `#window` from `#window .input` L713.

client/css/style.css
```diff
@@ -893,6 +920,30 @@ kbd {
...
+	margin-left: auto;
+	margin-right: auto;
+	margin-top: 5px;
+	margin-bottom: 5px;
+	width: 80%;
+	border: 1px solid #ddd;
+}
+
+#sidebar .join_form .btn {
+	border-radius: 2px;
```
astorije: Remove this.

client/css/style.css
```diff
@@ -893,6 +920,30 @@ kbd {
...
+	display: block;
+	width: 80%;
+	padding: 1px;
+	margin: auto;
+}
```
astorije: Can you add `height: 29px` (height of `.input`)?

client/js/lounge.js
```diff
@@ -132,6 +132,12 @@ $(function() {
 					text: "List all channels",
 					data: target.data("id"),
 				});
+				output += templates.contextmenu_item({
+					class: "join",
+					action: "join",
+					text: "Join channel",
```
astorije: `Join a channel...`

client/js/lounge.js
```diff
@@ -465,7 +471,34 @@ $(function() {
 		return false;
 	});
 
+	sidebar.on("click", ".add-channel", (e) => {
+		const id = $(e.target).data("id");
+		const joinForm = $(`#join_channel_${id}`);
+		joinForm.toggle();
+		return false;
+	});
```
astorije: Could you focus on the first input when user clicks on <kbd>+</kbd>?

client/js/lounge.js
```diff
@@ -465,7 +471,34 @@ $(function() {
 		return false;
 	});
 
+	sidebar.on("click", ".add-channel", (e) => {
+		const id = $(e.target).data("id");
+		const joinForm = $(`#join_channel_${id}`);
```
astorije: It doesn't matter much but it looks like most (all? That seems unlikely) of our ids/etc. use dashes instead of underscores.
Also, if you don't want to make this a dynamic id, you could use the id of the parent object and a class on this object. Up to you.

client/views/join_channel.tpl
astorije: Did you miss this comment? `Name`/`Key (optional)`

client/views/chan.tpl
```diff
@@ -1,7 +1,15 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
+	{{#equal type "lobby"}}
+		<span class="tooltipped tooltipped-w tooltipped-no-touch" aria-label="Join channel...">
```
astorije: `Join a channel...`

client/views/chan.tpl
```diff
@@ -1,7 +1,15 @@
...
+			<button class="add-channel" aria-label="Add channel" data-id="{{id}}"></button>
```
astorije: Same here in the ARIA label.

xPaw: I don't think the button needs aria label since the span above has one.

client/js/lounge.js
MaxLeiter: done

client/js/lounge.js
MaxLeiter: fixed

client/js/lounge.js
MaxLeiter: done

client/js/lounge.js
MaxLeiter: done

client/css/style.css
MaxLeiter: done

client/css/style.css
MaxLeiter: done

client/css/style.css
MaxLeiter: done

MaxLeiter:
@astorije: requested changes made. see new images

client/js/lounge.js
```diff
@@ -465,7 +471,35 @@ $(function() {
...
+		const chan = utils.findCurrentNetworkChan(channel);
+		if (chan.length) {
+			chan.click();
+		} else {
+			socket.emit("input", {
+				text: `/join ${channel} ${key}`,
+				target: form.prev().data("id"),
+			});
+		}
+		form.hide();
```
astorije: Could you clear the fields after submitting? It currently keeps the former values.

client/views/join_channel.tpl
```diff
@@ -0,0 +1,5 @@
+<form id="join-channel-{{id}}" class="join-form" method="post" action="">
+	<input type="text" class="input" name="channel" placeholder="Channel" pattern="[^\s]+" maxlength="200" title="Should be a valid channel name">
```
astorije: I wondering if we shouldn't add `autocomplete` here. @xPaw?

MaxLeiter: chrome no longer respects `autocomplete` - still worth adding? If so, I'm a big fan.

astorije: wat, seriously? It's part of the HTML5 spec (as opposed to `autocorrect`), so that's kinda surprising...

MaxLeiter: https://bugs.chromium.org/p/chromium/issues/detail?id=468153

astorije:
With @MaxLeiter's permission, I have added a few commits to enhance this PR. Behavior and UI are exactly how @MaxLeiter made it though.

- Moved the join channel form to its own component
- Added a keyboard handler to close the form when hitting <kbd>Escape</kbd>
- Made consistent (used `…` instead of `...`) in tooltips/context menu of join-channel and change-nick
- Changed the content of tooltip to "Cancel" and animate button from <kbd>+</kbd> to <kbd>&times;</kbd> when opening/closing join channel form. I removed the very discreet background on close/join-channel sidebar buttons to make this look good (which was at .1 opacity anyway lol). This will only be nice after #1833.

@MaxLeiter, hope you like it :))

astorije:
Also, I didn't touch your first commit, it just shows my name next to yours due to a rebase.

client/views/join_channel.tpl
```diff
@@ -0,0 +1,5 @@
+<form id="join-channel-{{id}}" class="join-form" method="post" action="" autocomplete="off">
+	<input type="text" class="input" name="channel" placeholder="Channel" pattern="[^\s]+" maxlength="200" title="Should be a valid channel name">
+	<input type="password" class="input" name="key" placeholder="Key (optional)" pattern="[^\s]+" title="Should be a valid channel key" maxlength="200" >
```
YaManicKill: Hmmm, should this be Password instead? I'm not sure "Key" makes it clear what it is.

client/js/lounge.js
```diff
@@ -474,6 +480,9 @@ $(function() {
 	});
 
 	const contextMenuActions = {
+		join: function(itemData) {
+			$(`#join-channel-${itemData}`).closest(".network").find("button.add-channel").click();
```
YaManicKill: Could we change this to ... not do this but call the function on the click handler instead? We keep ending up with issues because we rely on UI elements and then when we change or move those elements it breaks other things.

astorije:
- Rebased on master to bring the latest primer (fixes boldness in tooltip) and the removal of close button.
- Renamed "Key" into "Password" (I amended your commit for that one, @MaxLeiter)
- Did what @YaManicKill suggested, i.e. call a specific function to open the form from context menu instead of relying on click handler. That fixes a bug so it's all good!
- Made the channel field required so the form can't be submitted with no channel given

screen shot 2017-12-22 at 12 19 05

PR #1838, opened 2017-12-14
Try to find og:video and og:audio on html pages

xPaw:
Takes video/audio embeds further. This works on gfycat for example.

Code most likely needs a cleanup.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1245,6 +1245,11 @@ kbd {
 	max-width: 100%;
 }
 
+#chat video {
+	max-width: 640px;
+	max-height: 240px;
+}
+
 /* Do not display an empty div when there are no previews. Useful for example in
 part/quit messages where we don't load previews (adds a blank line otherwise) */
 #chat .preview:empty {
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -8,6 +8,9 @@ const cleanIrcMessage = require("../../../client/js/libs/handlebars/ircmessagepa
 const findLinks = require("../../../client/js/libs/handlebars/ircmessageparser/findLinks");
 const storage = require("../storage");
 
+const mediaTypeRegex = /^(audio|video)\/.+/;
+const linkRegex = /^https?:\/\//;
+
 process.setMaxListeners(0);
 
 module.exports = function(client, chan, msg) {
@@ -19,7 +22,7 @@ module.exports = function(client, chan, msg) {
 	const cleanText = cleanIrcMessage(msg.text);
 
 	// We will only try to prefetch http(s) links
-	const links = findLinks(cleanText).filter((w) => /^https?:\/\//.test(w.link));
+	const links = findLinks(cleanText).filter((w) => linkRegex.test(w.link));
 
 	if (links.length === 0) {
 		return;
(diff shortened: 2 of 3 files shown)


src/plugins/irc-events/link.js
```diff
@@ -198,7 +248,7 @@ function fetch(uri, cb) {
 				if (contentLength > limit) {
 					req.abort();
 				}
-			} else if (/^(audio|video)\/.+/.test(res.headers["content-type"])) {
+			} else if (mediaTypeRegex.test(res.headers["content-type"])) {
 				req.abort(); // ensure server doesn't download the audio file
```
MaxLeiter: This comment should be updated to audio or video file

MaxLeiter: Or more general media

astorije:
@xPaw, let me know when this is ready for review.

src/plugins/irc-events/link.js
```diff
@@ -49,8 +52,50 @@ module.exports = function(client, chan, msg) {
 
 function parse(msg, preview, res, client) {
 	switch (res.type) {
-	case "text/html":
-		var $ = cheerio.load(res.data);
+	case "text/html": {
+		const $ = cheerio.load(res.data);
```
YaManicKill: I know you are still wanting to clean this up, but the one thing I would say is this code would be nice to put ... somewhere else, not in this switch statement.

astorije:
@xPaw, do you have any plans to get back to this? :)

astorije:
👋

src/plugins/irc-events/link.js
```diff
@@ -55,51 +58,108 @@ module.exports = function(client, chan, msg) {
...
-				|| resThumb.size > (Helper.config.prefetchMaxImageSize * 1024)) {
+				// Make sure thumbnail is a valid url
+				if (!linkRegex.test(preview.thumb)) {
 					preview.thumb = "";
 				}
 
-				handlePreview(client, msg, preview, resThumb);
+				// Verify that thumbnail pic exists and is under allowed size
+				if (preview.thumb.length) {
+					fetch(escapeHeader(preview.thumb), client, (resThumb) => {
```
williamboman: `{language: client.language}` here?

xPaw: Just passing `client` should work as they variable has the same name, it gets destructed in the function anyway.

williamboman: Yeah but that introduces an implicit contract between the Client interface and the fetch function signature. This would not be the case if passing an object literal (this was initially pointed out by astorije in the PR where I added language support)

xPaw: Fair.

client/css/style.css
```diff
@@ -1450,6 +1450,11 @@ button.collapse-network:first-child:nth-last-child(3) {
 	max-width: 100%;
 }
 
+#chat video {
+	max-width: 640px;
```
astorije: I'm assuming this width is intentional? Does it max out to 100% on mobile?

xPaw: Yes.

astorije:
@xPaw, any chance you could add tests for the newly detected things?

xPaw:
@astorije It's painful to add a test because we only allow media over https. Even if we start a server with TLS certificates, it's not gonna work due to being untrusted.

PR #1839, opened 2017-12-14
Add sqlite logging and reloading messages

### LINKED ISSUE #442: Add persistent storage

### LINKED ISSUE #1392: Duplicate log entries when using two nicks

xPaw:
This is inspired by #1529 because I wanted to get a simpler backend in, then search can be added separately.

This takes some work from #663 but ultimately supersedes it.
Closes #442.

First commit introduces network UUIDs and "fixes" #1392, this is to prevent logging incorrectly when you are connected to same network under different nicknames (e.g. ZNC). This also makes it simpler to lookup networks and things like that.

Second commit adds creation of sqlite database itself and logging messages into the database. Currently it only writes actual messages and notices. Other message types are not written for one simple reason: backwards compatibility. If we store entire `Msg` object as json in the database, we will have to support any changes in these messages for all the types (good example is when I refactored `nick/mode` into a single `from` field).

Third commit actually adds reading messages back into the channel buffer when these channels are created or joined. `loadMessages` still needs some work to make loading messages a bit nicer.

Fourth commit adds a config variable to enable sqlite (it's disabled by default to not disrupt existing instances, at least right now).


defaults/config.js
```diff
@@ -148,6 +148,14 @@ module.exports = {
 	// @default  null
 	webirc: null,
 
+	//
+	// Storing and reloading messages from sqlite
+	//
+	// @type     boolean
+	// @default  false
+	//
+	sqlite: false,
```
astorije: I saw your note in PR description, but I think that load-from-backlog is such a desperately wanted feature that it would probably be a good thing to enable it by default.

Because v2.7.0 is starting to be in a good shape, we could target this PR for v3, and therefore the blast radius of setting this by default could be mitigated by the major bump.

No strong opinion, just my 2c.

test/models/network.js
```diff
@@ -10,6 +10,7 @@ describe("Network", function() {
 	describe("#export()", function() {
 		it("should produce an valid object", function() {
 			var network = new Network({
+				uuid: "hello world",
```
astorije: Don't we want actually valid UUIDs there, even in test objects?

astorije:
> backwards compatibility

Even in the current form, we're adding a schema, and we will probably want to extend it in the future. Do you have ideas on how we would go about that? Shipping with migrations that run on server start? I don't know, I'm sure there are 5 ways to address it, just wondering.

Also, we don't have to solve it in this PR, but that _will_ arise sooner or later, I just want to know that we have some sort of thoughts on this.

astorije:
Are you planning to add tests for this, by enabling the DB in test config? That would be sweet!

test/models/network.js
xPaw: This is testing that it's being preserved. There's a test down below to test actual generation.

astorije: Right, I just thought it'd be better to keep an actually valid value but whatever, no real argument there.

xPaw:
@astorije I changed it to store full msg objects (except for id and previews), and it stores all types of messages now.

astorije:
@xPaw, I know you were against this at first. What made you change your mind?

xPaw:
There are a lot of small little things that otherwise get lost, like users mode at the time or mentioned nicknames.

astorije:
Any insights on how we're going to evolve that / make sure things don't blow up with older schemas? :S

xPaw:
We will have to be careful about that. And I'm not sure we can do much besides doing migrations (since we store JSON data, that becomes a little silly, will have to select all the data and parse the json).

I'll store schema version in sqlite so it becomes a little easier in the future.

xPaw:
@florentsolt Would you mind giving this code a review? My current concern is that storing JSON objects will make it harder to implement message searching.

MiniDigger:
Been running this for like a week now and it works great. Not having to think about restarting lounge is really great.

few things that I noticed while running it:
* If you restart often, your scrollback is full of topic spam. I don't think it makes sense to print the topic when reloading data from sql. I didn't just join the channel, I am aware of the topic
* Previews for messages that are loaded from log don't get rendered. I guess this is minor too since I actually have seen the preview before restarting lounge, but its kinda confusing to users.
* File size for the db seems large.
```
.rw-r--r--  69M lounge 10 Feb 22:54 messages.db

sqlite> select count(*) from messages;
260447
```

xPaw:
> If you restart often, your scrollback is full of topic spam.

Well the server sends the topic whenever you join the channel, so that makes sense. I wouldn't want to filter topics from being stored either. Not sure what to do about that.

> Previews for messages that are loaded from log don't get rendered.

Yeah I am deliberately filtering previews out for 2 reasons: they can be updated after msg has already been added to sqlite, and with prefetch storage the images would be deleted on server restart.

> File size for the db seems large.

No compression + the fact we store JSON object adds a little boilerplate.

astorije:
> Well the server sends the topic whenever you join the channel, so that makes sense. I wouldn't want to filter topics from being stored either. Not sure what to do about that.

Is there a way to differentiate "someone just set a new topic" and "here is the topic since you just joined"?
If so, I think the latter is of little interest for The Lounge, because it's in the header (set aside the fact that we don't display full topic for now). If we can not render the topic right after joining, that would solve this.

defaults/config.js
```diff
@@ -148,6 +148,19 @@ module.exports = {
...
+	// Logging is also controlled per user individually (logs variable)
+	//
+	// none: Do not store any logs
+	// text: Text file per network/channel in user folder
+	// sqlite: Messages are stored in SQLite, this allows them to be reloaded on server restart
+	//
+	// @type     string
+	// @default  "text"
+	//
+	messageStorage: "text",
```
McInkay: Could this be an array? Allowing someone to specify ["sqlite", "text"] to get the advantage of reloading messages, but also be able to view text logs?

src/models/chan.js
```diff
@@ -183,21 +179,85 @@ Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
...
 	}
 
-	userLog.write(
-		client.name,
-		target.network.host, // TODO: Fix #1392, multiple connections to same server results in duplicate logs
-		this.type === Chan.Type.LOBBY ? target.network.host : this.name,
-		msg
-	);
-}
+	// TODO: Something more pluggable
```
McInkay: This is probably fine for this PR, the changes to allow plugins etc can be done later.

test/models/network.js
```diff
@@ -46,6 +48,15 @@ describe("Network", function() {
 			});
 		});
 
+		it("should generate uuid (v4) for each network", function() {
```
McInkay: Could maybe add a simple test that checks whether we log to text files or not. Don't need too much, but it might be a nice wee test.

src/plugins/sqlite.js
```diff
@@ -0,0 +1,126 @@
...
+	// Schema version #1
+	"CREATE TABLE IF NOT EXISTS options (name TEXT, value TEXT, CONSTRAINT name_unique UNIQUE (name))",
+	"CREATE TABLE IF NOT EXISTS messages (network TEXT, channel TEXT, time INTEGER, type TEXT, msg TEXT)",
+	"CREATE INDEX IF NOT EXISTS network_channel ON messages (network, channel)",
+	"CREATE INDEX IF NOT EXISTS time ON messages (time)",
+];
+
+class MessageStorage {
+	constructor(client) {
+		this.isEnabled = client.config.log && Helper.config.messageStorage === "sqlite";
```
McInkay: Could we just ... not require this if the user doesn't specify sqlite?

src/models/chan.js
```diff
@@ -183,21 +179,85 @@ Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
...
+		.then((messages) => {
+			if (messages.length === 0) {
+				if (network.irc.network.cap.isEnabled("znc.in/playback")) {
+					this.requestZncPlayback(network, 0);
+				}
+
+				return;
+			}
+
+			// TODO: This needs to insert correctly based on time
```
McInkay: What does this mean happens just now?

src/models/chan.js
```diff
@@ -183,21 +179,85 @@ Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
...
+				this.firstUnread = messages[messages.length - 1].id;
+			}
+
+			client.emit("more", {
+				chan: this.id,
+				messages: messages.slice(-100),
+			});
+
+			if (network.irc.network.cap.isEnabled("znc.in/playback")) {
+				const from = Math.floor(messages[messages.length - 1].time.getTime() / 1000);
```
McInkay: So is this from the last message we have in the logs until present?

src/models/chan.js
xPaw: It just prepends the message array. I don't remember what issues this could have caused though (out of sync messages?)

src/models/chan.js
xPaw: Yep.

src/plugins/sqlite.js
xPaw: Yes enabling needs to be moved outside.

defaults/config.js
xPaw: I don't really want to do that because there's no storage interface yet, which I don't plan to add in this PR either.

This could be improved upon in a future PR.

McInkay: I'm not sure the disadvantage of just allowing an array just now, and doing both if thye are both in the array? Sure, it'll need to be expanded and tidied up in other stuff (I'm not suggesting doing the storage interface just now) but your existing logic just needs to change to equal or .includes and change the `else if` to just `if` when setting up logic. Am I missing something?

src/plugins/sqlite.js
McInkay: We can probably leave this till we allow plugins if you want though.

src/models/chan.js
McInkay: But in what case would the messages be out of sync? Is there a race condition when sending messages?

xPaw: I'm not sure anymore to be honest, I don't think it's a such a big issue than I initially though.

astorije: Wasn't that due to some ZNC replay or something? I might be totally mixing this with another PR/issue though 😅

package.json
```diff
@@ -59,6 +59,7 @@
     "thelounge-ldapjs-non-maintained-fork": "1.0.2",
     "ua-parser-js": "0.7.17",
     "urijs": "1.19.1",
+    "uuid": "3.1.0",
```
astorije: This package is now in [v3.2.1](https://github.com/kelektiv/node-uuid/blob/master/CHANGELOG.md#320-2018-01-16).

src/models/network.js
```diff
@@ -41,6 +42,10 @@ function Network(attr) {
 		chanCache: [],
 	});
 
+	if (!this.uuid) {
```
astorije: Why do we need to check for previous ids?

test/models/network.js
astorije: @xPaw, what do you think about this?

src/models/network.js
xPaw: I didn't want to sheets to always call uuid() in default object as it would waste the call when loading an existing network.

package.json
```diff
@@ -56,6 +56,7 @@
     "semver": "5.5.0",
     "socket.io": "2.0.4",
     "spdy": "3.4.7",
+    "sqlite3": "3.1.13",
```
astorije: For good measure, I went through a couple of their (*many*) open issues, and oh boy does it feel like `node-ldapjs` again (if I'm reading the following tickets correctly):

- [Their README](https://github.com/mapbox/node-sqlite3#supported-platforms) doesn't mention Node v8 or Node v9 support, though their CI seems fine with it (also see https://github.com/mapbox/node-sqlite3/issues/939 and https://github.com/mapbox/node-sqlite3/issues/906)
- https://github.com/mapbox/node-sqlite3/issues/902 mentions issues with Node v9
- https://github.com/mapbox/node-sqlite3/issues/903 and https://github.com/mapbox/node-sqlite3/issues/942 talk about issues with the RPi3
- https://github.com/mapbox/node-sqlite3/issues/910 mentions issues with Win32
- The eternal `node-gyp` issues with https://github.com/mapbox/node-sqlite3/issues/909, https://github.com/mapbox/node-sqlite3/issues/901 (I stopped counting projects where people reported errors due to `node-gyp` thingies)

Ouch.
Note that I'm not trying to make some sort of a point with this comment, just documenting findings. Just like when we introduced LDAP support that got us a few install-related tickets, we might have to face issues related to installs, especially if this is not compatible with v8/v9!

Just putting this out there, someone mentioned https://github.com/JoshuaWise/better-sqlite3 as an alternative 🤷‍♂️

Given all those, I think we really want some minimal tests around writing/loading logs, so that all supported Node versions and Linux/Windows support get some confidence that this is working.
Thoughts?

src/client.js
```diff
@@ -10,6 +10,7 @@ const Network = require("./models/network");
 const ircFramework = require("irc-framework");
 const Helper = require("./helper");
 const UAParser = require("ua-parser-js");
+const MessageStorage = require("./plugins/sqlite");
```
astorije: Is it on purpose that `MessageStorage` == `sqlite.js`? I'm fine if so, just a sanity check in case it was forgotten with a rebase or something.

src/client.js
```diff
@@ -78,8 +79,16 @@ function Client(manager, name, config = {}) {
 	});
 
 	const client = this;
-
 	let delay = 0;
+
+	if (!Helper.config.public) {
+		client.messageStorage = new MessageStorage(client);
```
astorije: Your constructor in `plugins.js` does not take an argument.

src/models/chan.js
```diff
@@ -183,21 +179,72 @@ Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
...
 };
 
-function writeUserLog(client, msg) {
-	if (!msg.isLoggable()) {
-		return false;
+Chan.prototype.writeUserLog = function(client, msg) {
+	this.messages.push(msg);
+
+	// Does this user have logs disabled
+	if (!client.config.log) {
```
astorije: I see you already do this check in `src/client.js` L87, should we do it again, or instead check that a message store was enabled?

xPaw: `src/client.js` check is for `sqlite` only, while this check here ensures no logging is performed for any method (sqlite or text). It will most likely become clearer when the logging API is abstracted out.

src/client.js
xPaw: Yeah this is correct.

test/models/network.js
xPaw: I'm not sure why @McInkay said text file logging, which I didn't touch in this PR?

I'll try to add some sqlite tests.

McInkay: You didn't change the text logging, but you changed the logic around when we do that. You've added the option in the config to define when we do what logging.

xPaw: The trouble with testing these code paths is that they want plenty of object references (client, network, message, config helper)

McInkay: Looks like we need to improve the code then.

src/models/chan.js
```diff
@@ -183,21 +179,86 @@ Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
...
-}
+	// TODO: Something more pluggable
+	if (Helper.config.messageStorage.includes("sqlite")) {
+		client.messageStorage.index(target.network.uuid, this.name, msg);
+	}
+
+	if (Helper.config.messageStorage.includes("text")) {
+		userLog.write(
+			client.name,
+			target.network.host, // TODO: Fix #1392, multiple connections to same server results in duplicate logs
```
astorije: Your PR description says the first commit fixes #1392, is it still TODO?

src/models/chan.js
```diff
@@ -183,21 +179,86 @@ Chan.prototype.getFilteredClone = function(lastActiveChannel, lastMessage) {
...
+			}
+		})
+		.catch((err) => log.error(`Failed to load messages: ${err}`));
+};
+
+Chan.prototype.isLoggable = function() {
+	return this.type === Chan.Type.CHANNEL || this.type === Chan.Type.QUERY;
+};
+
+Chan.prototype.requestZncPlayback = function(network, from) {
```
astorije: Does this function need to be exported? If not maybe just have it outside the prototype and pass this.name as argument

src/models/msg.js
```diff
@@ -40,6 +40,7 @@ class Msg {
 
 	isLoggable() {
 		return this.type !== Msg.Type.MOTD &&
```
astorije: I didn't realize we didn't log motds, maybe we should?

src/plugins/sqlite.js
```diff
@@ -0,0 +1,126 @@
+"use strict";
+
+const path = require("path");
+const fsextra = require("fs-extra");
+const sqlite3 = require("sqlite3");
+const Helper = require("../helper");
+const Msg = require("../models/msg");
+
+const currentSchemaVersion = 1;
```
astorije: At first I was thinking of storing the version of The Lounge in the schema, but that would be pretty terrible when running and updating from `master`.

How are we planning to do migrations?
First versions of Rails were storing migration versions as integer like this, but they ended with so many problems that they switched to timestamp and have kept it that way since. I'm no fan of Rails (it's part of my day job), but their migration system is pretty nicely done. I'm expecting candidates in the Node world, do we want to look at this before shipping this feature?

src/plugins/sqlite.js
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	index(network, channel, msg) {
+		if (!this.isEnabled) {
+			return;
+		}
+
+		const clonedMsg = Object.keys(msg).reduce((newMsg, prop) => {
+			// id is regenerated when messages are retrieved
+			// previews are not stored because storage is cleared on lounge restart
```
astorije: Do we have a way to re-trigger previews when they are loaded? (i.e. if user never loads older messages, they wouldn't get re-fetched)

src/plugins/sqlite.js
```diff
@@ -0,0 +1,126 @@
...
+	 *
+	 * @param Network network - Network object where the channel is
+	 * @param Chan channel - Channel object for which to load messages for
+	 */
+	getMessages(network, channel) {
+		if (!this.isEnabled || Helper.config.maxHistory < 1) {
+			return Promise.resolve([]);
+		}
+
+		return new Promise((resolve, reject) => {
```
astorije: I found https://github.com/kriasoft/node-sqlite when looking if node-sqlite3 can return promises. No idea if that's an acceptable candidate, but I also appreciated that it has built-in support for migrations!
It seems to be a simple wrapper around node-sqlite3 that might serve us well maybe?

test/plugins/sqlite.js
```diff
@@ -0,0 +1,102 @@
...
+
+describe("SQLite Message Storage", function() {
+	const expectedPath = path.join(Helper.getHomePath(), "logs", "testUser.sqlite3");
+	let store;
+
+	// Delete database file from previous test run
+	before((done) => {
+		store = new MessageStorage();
+
+		if (fs.existsSync(expectedPath)) {
```
astorije: Could you use [`fs-extra`'s `pathExists`](https://github.com/jprichardson/node-fs-extra/blob/master/docs/pathExists.md)? That way it doesn't have to be async.

(By the way functions of `fs-extra` support promises so we should probably make use of that)

test/plugins/sqlite.js
```diff
@@ -0,0 +1,102 @@
...
+			done();
+		}
+	});
+
+	it("should resolve an empty array when disabled", (done) => {
+		store.getMessages(null, null).then((messages) => {
+			expect(messages).to.be.empty;
+			done();
+		});
+	});
```
astorije: What about using https://github.com/domenic/chai-as-promised?

For example, this test would become:

```js
it("should ...", function () {
	return expect(store.getMessages(null, null)).to.eventually.be.empty;
});
```

test/plugins/sqlite.js
```diff
@@ -0,0 +1,102 @@
...
+const Msg = require("../../src/models/msg");
+const Helper = require("../../src/helper");
+const MessageStorage = require("../../src/plugins/sqlite.js");
+
+describe("SQLite Message Storage", function() {
+	const expectedPath = path.join(Helper.getHomePath(), "logs", "testUser.sqlite3");
+	let store;
+
+	// Delete database file from previous test run
+	before((done) => {
```
astorije: [Mocha recommends](https://mochajs.org/#arrow-functions) against using arrow functions, and sticking to regular functions ([`eslint-plugin-mocha` has `no-mocha-arrows`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-mocha-arrows.md) but it can't be used along with `prefer-callback` yet), mind changing the different `() => {}`s into `function() {}`s?

test/plugins/sqlite.js
```diff
@@ -0,0 +1,102 @@
...
+	});
+
+	it("should create database file", () => {
+		expect(store.isEnabled).to.be.false;
+		expect(fs.existsSync(expectedPath)).to.be.false;
+
+		store.enable("testUser");
+
+		expect(store.isEnabled).to.be.true;
+		expect(fs.existsSync(expectedPath)).to.be.true;
```
astorije: You could use `pathExists` here as well.

test/plugins/sqlite.js
```diff
@@ -0,0 +1,102 @@
...
+	});
+
+	it("should insert schema version to options table", (done) => {
+		store.database.serialize(() =>
+			store.database.get("SELECT value FROM options WHERE name = 'schema_version'", (err, row) => {
+				expect(err).to.be.null;
+
+				// Should be sqlite.currentSchemaVersion,
+				// compared as string because it's returned as such from the database
+				expect(row.value).to.equal("1");
```
astorije: I'm expecting this to be directly handled by whatever migration mechanism we end up using, but in the present case, you could make `schema_version` a column of `options`, which wouldn't be key/value anymore but a single row with as many columns as we have options.
Version could then be the expected type. Also slightly easier to deal with default values, etc. Only caveat is to have to ensure only 1 line ever lives there, but that's worth it IMO.
Thoughts?

test/plugins/sqlite.js
```diff
@@ -0,0 +1,102 @@
...
+
+				done();
+			})
+		);
+	});
+
+	it("should store a message", (done) => {
+		store.index("this-is-a-network-guid", "#ThisIsAChannel", new Msg({
+			time: 123456789,
```
astorije: Add `type` maybe?

test/plugins/sqlite.js
```diff
@@ -0,0 +1,102 @@
...
+			text: "Hello from sqlite world!",
+		}));
+
+		store.database.serialize(done);
+	});
+
+	it("should retrieve previously stored message", (done) => {
+		store.getMessages({
```
astorije: This retrieves a message stored in a previous test?!
It would be better if tests didn't depend on each other. You'll also need to change `before` into `beforeEach` above.

src/models/chan.js
xPaw: It fixes it in the case of sqlite, I didn't change it for text logging because that would change the logging folder (which is breaking, and out of the scope of this PR).

src/models/chan.js
xPaw: If we ever (hopefully) refactor to use classes, then it will make sense to keep it.

src/models/msg.js
xPaw: That would be pretty spammy in the lobby network.

src/plugins/sqlite.js
xPaw: Timestamp is still an integer, so I don't see how that solves anything.

src/plugins/sqlite.js
xPaw: We technically can, but I wouldn't look at it because it's not as simple as iterating all links and fetching previews (would need to queue and process with limits).

test/plugins/sqlite.js
xPaw: The only reason they discourage it is due to:
> If you do not need to use Mocha’s context, lambdas should work.

Which we are aware of and take into consideration.

test/plugins/sqlite.js
xPaw: I went with `options` because I was thinking we could expand that table to store other details (hence K/V). Do you want me to change it?

test/plugins/sqlite.js
xPaw: I really don't see a need to do that (it would be more work than just changing `before`). I understand it's great when each test is self contained but that's not always practical.

test/plugins/sqlite.js
xPaw: That doesn't work as `pathExists` uses `fs.access`, but the file is locked by sqlite.

test/plugins/sqlite.js
xPaw: Doesn't seem worthwhile to reduce 2 lines into one.

src/models/chan.js
astorije: Ok, then we either need to reword the PR description, or remember to re-open the issue after this is merged, as it will auto-close #1392.

src/models/chan.js
astorije: What do you mean? If it's not used externally, it's safe to move it outside of the prototype, no? We have that in a few places as classes don't have private methods (yet).

src/plugins/sqlite.js
astorije: It's not about the type, it's about the value. Each migration id is a timestamp, so there is no conflicts across PRs, order is easier to manage, etc.

But really it doesn't matter that much, because I would _much_ prefer if we don't reinvent the migration wheel and use something off-the-shelf, which would enforce its conventions.

src/plugins/sqlite.js
astorije: Makes sense, I didn't mean in this PR necessarily.

src/plugins/sqlite.js
astorije: It would make the code about promises a tiny bit simpler, but more importantly it would bring support for migrations. I'm still concerned how we're going to deal with this from now on. Doesn't have to be now, but before v3 we need to find a solution that will work for us overtime or we'll regret it 😅

test/plugins/sqlite.js
astorije: They also specify

> However, the result will be more difficult to refactor if the need eventually arises.

I would like us to use `eslint-plugin-mocha` at some point, this is just forward-thinking 🤷‍♂️

test/plugins/sqlite.js
astorije: I'd argue it makes tests easier to comprehend, and it can obviously be used in more than one place, but whatever. We can use `async`/`await` as soon as we drop support for Node v6, post April 2019 :trollface:

test/plugins/sqlite.js
astorije: Argh, that sucks. Oh well...

test/plugins/sqlite.js
astorije: I'm undecided on this. I totally understand the need for arbitrary K/V, but also it's just as common to have one column per option (even future ones) as it is to have random K/V/T (type) rows.

I think if we change our mind in a future version, it would be easier to go from 1-row-multi-columns to K/V rows than the opposite, and that also offers native types, which is cool.
If you're not okay with this, at least add a type column maybe?

test/plugins/sqlite.js
astorije: tbh if it's not practical, there might be a problem somewhere. In my experience, this _will_ bite us.
I understand it's extra work though, so I'm not gonna veto or anything, but I'd be more comfortable if we don't do that (I can lend a hand if you want). @McInkay, what's your take on this?

src/plugins/sqlite.js
xPaw: I'll change `1` to a timestamp. I thought about migrations more, and an existing solution that merely executes SQL statements from a file will not work for us in the future.

Because we store JSON objects, we will surely need to change the structure of it in the future (e.g. when `from` was changed into an object of `{nick, mode}`), which would require iteration in javascript as it would be very impractical trying to write a SQL query to manipulate the JSON.

It's hard to predict what kind of migrations we will have just yet, it's hard to design for it. Good thing is that we're not limited by our implementation until the need arises.

McInkay: > Each migration id is a timestamp, so there is no conflicts across PRs, order is easier to manage, etc.

I mean, there will be a conflict either way because changing the same line in a file will create a git conflict. But either way, this is what we want, because even if the version is a timestamp, it'll need to change if someone else has merged since I started my development, because otherwise my time might be behind master now, and it looks like the schema is in the past.

src/models/msg.js
McInkay: There's no reason to log MOTD. It's meant to be something that is just around for your current connection. On reconnect, you'll get the new MOTD, so the only time you won't have the MOTD is when you are currently disconnected from the IRC server.

src/plugins/sqlite.js
astorije: 👍 to both your comments.

@xPaw, I agree that SQL migrations don't work for us. Rails (I'm just taking this as an example because that's my only experience with migrations, not because I want what they're doing specifically) does migrations in Ruby, and it's very useful to migrate data along with schema.
I can see how JavaScript migrations would be a much, much better solution for us for precisely what you're saying. It's indeed likely our migrations will more often be about the stored JSON than the SQL schema, and JS would give us that for free. [There are many articles on the subject](https://kostasbariotis.com/data-migration-with-nodejs/) for when we need that.
Or... don't store JSON and store things in traditional SQL tables/columns 😅

@McInkay, yes re: conflict, but I meant when multiple PRs do changes to the schema, they have to compete on the number, which is why Rails moved away from that and went for timestamps. And you're right about conflicting in the file itself and wanting this on purpose, [in fact this is exactly what Rails does and for that very reason](http://edgeguides.rubyonrails.org/active_record_migrations.html#schema-dumps-and-source-control).

src/models/msg.js
astorije: I guess I'm expecting servers to update their MOTDs frequently, which is definitely not extremely usual. But for example, I like being able to refer to older Freenode MOTDs, as they display news / info in there (for example about their conference), but it might be too niche and/or not worth it.

McInkay: I feel like the advantages of this for a few people would be outweighed by the confusion of having multiple motds appearing when you start the lounge, personally. But I don't feel about it strongly.

astorije: Yeah that's a good conclusion. I do see benefits in keeping track of older MOTDs, but you're right that the benefits do not outweigh the drawbacks. Another thing that I'm happy to see what feedbacks we get, i.e. if we ever get users confused by this or not, and we can then rethink this.

src/models/chan.js
```diff
@@ -259,6 +259,6 @@ Chan.prototype.isLoggable = function() {
 	return this.type === Chan.Type.CHANNEL || this.type === Chan.Type.QUERY;
 };
 
-Chan.prototype.requestZncPlayback = function(network, from) {
-	network.irc.raw("ZNC", "*playback", "PLAY", this.name, from.toString());
-};
+function requestZncPlayback(channel, network, from) {
```
astorije: Could do `requestZncPlayback({name}, network, from)` or passing `name` directly `requestZncPlayback(name, network, from)` but that is awfully nitpicky :D

src/plugins/sqlite.js
```diff
@@ -6,7 +6,7 @@ const sqlite3 = require("sqlite3");
 const Helper = require("../helper");
 const Msg = require("../models/msg");
 
-const currentSchemaVersion = 1;
+const currentSchemaVersion = 1520239200;
```
astorije: Ahah, I should have been more specific, I was picturing a more human-readable way to do timestamp (e.g. `20180307010400` ...à la Rails... I know I know I should stop mentioning it before it bites back 🙂) but at this point in our implementation, it probably doesn't really matter.

astorije:
I am overall 👍 on this PR, except this test file where each test relies on the one previously run, which is really something I would consider as a last measure if we don't have a choice.
For example, we might soon (or already) benefit from randomized-order tests (to test better) or parallel tests (to test faster, though not at all a problem right now) and that would make those impossible.

What does @McInkay think about this? I'm not blocking this PR though, just not super thrilled about that part, and happy to sit back if @McInkay disagrees :)

src/models/chan.js
xPaw: Just `name` is more confusing because it doesn't tell you what name it should be, passing `channel` is more clearer, imo.

astorije: Fair enough.

src/models/msg.js
McInkay: Yeah, if people find it confusing, happy to revisit.

McInkay:
> except this test file where each test relies on the one previously run, which is really something I would consider as a last measure if we don't have a choice.

Yeah, so ... unit tests shouldn't do this, but it's really hard to test sqlite code in unit tests, and integration tests should be fine for this.

When we change it up to allow plugins for extending, we should make sure we have decent testing around that, and we'll be fine.

We *could* mock out a bunch of stuff with sqlite for each test and make them properly unit tests but ... meh

astorije:
Alright then, 2 against 1, I'll abdicate 😁

xPaw:
You still need to keep maxHistory set, it does not load messages from sqlite when scrolling up just yet.

PR #1842, opened 2017-12-16
 Move the user list client code to its own file and make it possible to navigate with keyboard

astorije:
Note that the context menu does not support keyboard navigation yet, so there is still a click needed to actually do an action. This belongs to separate work however.

<p align=center>
  userlist keyboard navigation
</p>

This could probably use some ARIA attributes but I'm not 100% sure.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1921,7 +1921,8 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 .context-menu-item:hover,
 .textcomplete-item:hover,
-.textcomplete-menu .active {
+.textcomplete-menu .active,
+#chat .users .user.active {
 	background-color: #f6f6f6;
 	transition: none;
 }
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1204,7 +1204,7 @@ kbd {
 	color: #50a656;
 }
 
-#chat .user:hover {
+.chat .user:hover {
 	opacity: 0.6;
 }
 
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -163,6 +163,11 @@ function renderChannelUsers(data) {
 		.sort((a, b) => b.lastMessage - a.lastMessage)
 		.map((a) => a.nick);
 
+	// Before re-rendering the list of names, there might have been an entry
+	// marked as active (i.e. that was highlighted by keyboard navigation).
+	// It is `undefined` if there was none.
+	const previouslyActive = users.find(".active").data("name");
+
 	const search = users
 		.find(".search")
 		.prop("placeholder", nicks.length + " " + (nicks.length === 1 ? "user" : "users"));
@@ -175,6 +180,15 @@ function renderChannelUsers(data) {
 	// Refresh user search
 	if (search.val().length) {
 		search.trigger("input");
+	}
+
+	// If a nick was highlighted before re-rendering the lists, re-highlight it in
+	// the newly-rendered list.
+	if (previouslyActive) {
+		// We need to un-highlight everything first because triggering `input` with
+		// a value highlights the first entry.
+		users.find(".user").removeClass("active");
+		users.find(`.user[data-name="${previouslyActive}"]`).addClass("active");
 	}
 
 	return users;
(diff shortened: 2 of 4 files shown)


client/css/style.css
```diff
@@ -1046,7 +1046,7 @@ kbd {
 	color: #50a656;
 }
 
-#chat .user:hover {
+.chat .user:hover {
```
astorije: After this change, only nicks in the message list window will change opacity on hover, since hovering on the userlist adds a background instead, like our dropdowns.

client/js/render.js
```diff
@@ -173,6 +176,8 @@ function renderChannelUsers(data) {
 	if (search.val().length) {
 		search.trigger("input");
 	}
+
+	return users;
```
xPaw: This appears to be unnecessary?

client/js/userlist.js
```diff
@@ -0,0 +1,108 @@
...
+		} else {
+			userlist = userlists.find(".names-original");
+		}
+
+		const users = userlist.find(".user");
+
+		// Find which item in the array of users is currently selected, if any.
+		// Returns -1 if none.
+		const activeIndex = users.toArray()
+			.findIndex((user) => user.classList.contains("active"));
```
astorije: This is certainly not ideal, but I couldn't find a better way. I could try to improve now that it's committed and working if you prefer (also, suggestions welcome).

client/js/render.js
astorije: Look at R123, I need it to keep a reference on the user list that was just created.

xPaw: Oh my bad, the diff was rendered in such a way I assumed the return was for a wrong function.

astorije: Hey, np, that just shows my change is not as easy to read as I'd hoped :) Do you want me to add a comment for this?

xPaw: Nah it's fine.

client/js/userlist.js
xPaw: Active would reset whenever the user list updates, right? All the user list updates are just full re-renders.

You could try using this: https://api.jquery.com/index/

astorije: > Active would reset whenever the user list updates, right? All the user list updates are just full re-renders.

Good catch, I missed that. It's correctly handled now, highlight will carry on if there is a user list update 🎉
But that was suuuuuch a wild goose chase though, with all the cases, had to keep multiple browsers opened, etc. I _think_ I have all the cases right now, but it would be great if you could give it a quick test to see if I missed something.
I so wish there were ways to write test cases for that, it would have been a breeze to list and fix all the cases. The day we have browser tests, I am so happy.

> You could try using this: api.jquery.com/index

Yeah I tried like different ways with that and couldn't go anywhere with it. I'll give it a quick try again, but I'm not too confident. Would love a hand if you want to give it a shot though.

astorije: Yeah so I'm trying with things like the following, but no dice:

```js
const activeIndex = users.find.index(".active");
const activeIndex = users.index(".user.active");
const activeIndex = userlist.index(".user.active");
const activeIndex = userlist.index(".active");
```

There must be a way, I'm just not seeing it...

xPaw:
@astorije I've rebased it for you. I've tested it and seems fine except for one problem: I can't escape out of context menu after opening it with enter. I'm not sure why it doesn't let me do that.

Context menus need to become keyboard aware too, but that's outside of this PR.

astorije:
Fixed linting issue.

@xPaw, you can't escape from context menu simply because there is no keyboard navigation on context menus, like you mentioned. When pressing <kbd>Enter</kbd>, keyboard focus is not passed to the context menu.

astorije:
I really don't understand why the `Escape` key event gets swallowed when the input has focus...
I feel like I've gone through this in the past, most likely another project, but I don't remember the resolution.

astorije:
Maybe later, I'd like to focus on PR reviews + docs first. But yeah, fixing this bug is useless until keyboard-accessible context menus anyway.

astorije:
btw I wanted to create an issue but forgot there was one already: https://github.com/thelounge/thelounge/issues/1911

PR #1845, opened 2017-12-16
Improve error page at loading time

astorije:
- Display the "Reload page" instantly and not after 5 seconds
- Remove stack trace, buggy anyway
- Wrap the error details so it does not expand beyond boundaries (scrollbar would not show up either)
- Do not show the slow-loading warning on error
- Make error details selectable instead of editable
- Label improvements
- Wrap all script in IIFE
- Rename script to more generic error handling

Before | After
--- | ---
<b>On load:</b><br>screen shot 2017-12-16 at 16 16 43<br><b>After 5 seconds:</b><br>screen shot 2017-12-16 at 16 16 47 | screen shot 2017-12-16 at 16 17 14
screen shot 2017-12-16 at 16 16 57 | screen shot 2017-12-16 at 16 17 20


client/css/style.css
```diff
@@ -1345,10 +1345,30 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+}
+
+#loading pre {
+	white-space: normal;
+
+	-webkit-user-select: text;
+	-moz-user-select: text;
+	-ms-user-select: text;
+	user-select: text;
+	cursor: text;
```
astorije: @xPaw, I'm guessing you made the `pre` content-editable so people could copy-paste. I have instead made it selectable, let me know if you see something wrong in that.

client/js/loading-slow-alert.js
```diff
@@ -8,40 +8,53 @@
...
-setTimeout(function() {
-	var element = document.getElementById("loading-slow");
+function displayReload() {
+	var loadingReload = document.getElementById("loading-reload");
+	if (loadingReload) {
+		loadingReload.style.display = "block";
+	}
+}
+
+var loadingSlowTimeout = setTimeout(function() {
```
xPaw: This creates a global variables (along with the function above). Wrap this entire script in an anonymous function.

client/js/loading-slow-alert.js
```diff
@@ -8,40 +8,53 @@
...
+	message.textContent = "An error has occured that prevented the client from loading correctly.";
 
 	var summary = document.createElement("summary");
 	summary.textContent = "More details";
 
-	if (error instanceof ErrorEvent) {
-		error = error.message + "\n\n" + error.stack + "\n\nView developer tools console for more information and a better stacktrace.";
+	var data = document.createElement("pre");
+	if (e instanceof ErrorEvent) {
+		data.textContent = e.error && e.error.stack ? e.error.stack : e.message;
```
xPaw: I didn't include stacktrace because it's completely useless without sourcemap loaded in, and that only loads in with devtools open before the error happens, so it's effectively very useless and misleading.

astorije: Interesting. You did include `error.stack` and that was always `undefined` (at least for me), so I figured you wanted the stack trace. Want me to remove it?

client/js/loading-slow-alert.js
astorije: Done.

client/js/loading-slow-alert.js
astorije: > that only loads in with devtools open before the error happens

Odd, because I'm trying everything to cause this, even loading in private browsing and such, and it always shows up. Also I have a fallback if the stack is empty.

But I get your overall point. Want me to remove it?

xPaw: I don't understand how it could possibly work since sourcemaps are loaded by devtools. Looking at other error reporters, they either post-process errors on server, or load sourcemap as specified in JS themselves.

So yeah, it's better off removing it since we're not making a fully fledged error reporter here.

client/js/loading-error-handlers.js
```diff
@@ -0,0 +1,62 @@
...
+	};
+
+	var loadingSlowTimeout = setTimeout(function() {
+		var loadingSlow = document.getElementById("loading-slow");
+
+		if (loadingSlow) {
+			loadingSlow.style.display = "block";
+		}
+
+		displayReload();
```
xPaw: You can move this inside the `if` above, if there is no slow message, there won't be a reload button anyway.

client/css/style.css
xPaw: Instead of duplicating these rules, can you add the selector where these are defined instead?

client/js/loading-slow-alert.js
astorije: > I don't understand how it could possibly work since sourcemaps are loaded by devtools.

I don't know, mate, it just works :man_shrugging: 
But you'll notice that on the screenshot it shows the stacktrace of the built JS bundle, not the original source. That makes it less interesting for sure.

Will remove.

client/js/loading-error-handlers.js
astorije: Well, I [moved the button outside the block](https://github.com/thelounge/lounge/pull/1845/files#diff-7f958a715ef0d644371ae8b1b5f18250R59) though.

I'm not 100% sure of why this `if` exists: is it because we're not taking any chances and it could not exist in case of a bug? If so then maybe we should display the button even if the text wasn't found, no?

client/css/style.css
astorije: Done.

client/js/loading-slow-alert.js
astorije: Done:

screen shot 2017-12-17 at 23 53 36

Very glad you added this page overall. I had my concerns at first because of misunderstanding, but it makes the experience more pro and less frustrating 👏

client/js/loading-error-handlers.js
astorije: I didn't do this (at least yet) per my message above. Let me know what you think.

xPaw: The timer is not cancelled when the app loads, so it just does nothing if the element doesn't exist.

astorije: Ah, good catch, will do and add a comment about this.

astorije: Done, thanks!

PR #1846, opened 2017-12-16
Replace the all-red highlighted messages with a dot marker

astorije:
This re-uses the shape and color of the notification dot.

Browser | Mobile
--- | ---
screen shot 2017-12-16 at 17 16 04 | screen shot 2017-12-16 at 17 16 18

This can definitely use some CSS tweaking.
What do people think?

Marking as v3 if it makes it because it's sufficiently disrupting...

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1026,6 +1026,22 @@ kbd {
 	position: relative;
 }
 
+#chat .channel .message.highlight::before {
+	display: inline-block;
+	content: " ";
+	background: #e74c3c;
+	height: 16px;
+	min-width: 16px;
+	border-radius: 8px;
+	margin-top: 5px;
+	margin-left: 10px;
+	margin-right: -5px;
+}
+
+#chat .channel .message.highlight .from {
+	width: 113px; /* 134px original - 5px margin - 16px highlight dot width */
+}
+
 #chat .content {
 	flex: 1 1 auto;
 	min-width: 0;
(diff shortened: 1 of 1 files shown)


client/css/style.css
```diff
@@ -2243,6 +2256,12 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		padding: 5px 10px;
 	}
 
+	#chat .channel .message.highlight::before {
+		margin: 0;
+		height: 12px;
+		min-width: 12px;
```
astorije: I wanted to leave this at 16px, but I get some weird alignment issues beyond 12px (even though the text is supposed to be 18px high, I don't entirely understand).

client/css/style.css
```diff
@@ -1026,6 +1026,22 @@ kbd {
 	position: relative;
 }
 
+#chat .channel .message.highlight::before {
+	display: inline-block;
+	content: " ";
+	background: #e74c3c;
+	height: 16px;
+	min-width: 16px;
```
astorije: Can't use `width`: multiline messages hide half the dot.

client/css/style.css
```diff
@@ -1026,6 +1026,22 @@ kbd {
 	position: relative;
 }
 
+#chat .channel .message.highlight::before {
+	display: inline-block;
```
astorije: `inline-block` so it's compatible with all our responsive width.

williamboman:
I thinks this looks really awkward. I like the idea, but right now it's too obnoxious imho. What about something like this?

screenshot 2018-03-07 at 11 57 16

xPaw:
Gonna close this to keep the discussion in #1847.

astorije:
Superseded by https://github.com/thelounge/thelounge/pull/2526.

PR #1847, opened 2017-12-16
Replace current highlight with a highlight on entire message

astorije:
This is an alternative attempt at #1846.
So far I prefer the other result, but that's a personal preference. Also it seems like there will be more color-related issues with this way of doing so, which does not bode well with accessibility.

Browser | Mobile
--- | ---
screen shot 2017-12-16 at 17 54 44 | screen shot 2017-12-16 at 17 54 59

At the moment the fading-out gradient for long nick is broken with this. I'll try to fix it if that's what we go with.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1174,11 +1174,20 @@ kbd {
 }
 
 #chat .error,
-#chat .error .from,
-#chat .channel .highlight .from,
-#chat .channel .highlight .text,
-#chat .channel .highlight .user {
-	color: #f00;
+#chat .error .from {
+	color: #e74c3c;
+}
+
+#chat .channel .message.highlight {
+	background-color: rgba(231, 76, 60, 0.5);
+}
+
+#chat .channel .message.highlight .from {
+	border-right: 1px solid rgba(231, 76, 60, 0.5);
+}
+
+#chat .channel .message.highlight .from::after {
+	background: transparent;
 }
 
 #chat .toggle-button.opened, /* Thumbnail toggle */
@@ -2243,6 +2252,13 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		padding: 5px 10px;
 	}
 
+	#chat .channel .message.highlight {
+		margin-left: -10px;
+		margin-right: -10px;
+		padding-left: 10px;
+		padding-right: 10px;
+	}
+
 	#chat .msg {
 		display: block;
 		padding: 2px 0;
@@ -2250,6 +2266,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 	#chat .time,
 	#chat .from,
+	#chat .channel .message.highlight .from,
 	#chat .content {
 		border: 0;
 		display: inline;
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -166,11 +166,16 @@ body {
 }
 
 #chat .error,
-#chat .error .from,
-#chat .channel .highlight,
-#chat .channel .highlight .from,
-#chat .channel .highlight .text {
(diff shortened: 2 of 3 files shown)


christer88:
As talked about on IRC, the contrast in colours is a bit much in the example above. Using the colour of the background in a different shade, would in my opinion look very good! (as shown below, where a different shade of grey colour is used)

irccloud

astorije:
> Using the colour of the background in a different shade, would in my opinion look very good!

That would be some kind of light gray on example theme, what do you think?

Also, that would require fixing some contrasts like the one on time.

christer88:
It would yes. Not tested how it would look like.

Including some examples below.

IRCCloud white background:
irccloud_white

Discord white background:
discord_white

xPaw:
Looking at the first screenshot, it's very attention grabbing, I still prefer the way I did it in Solarized theme where only time and nick parts are colored.

astorije:
> Looking at the first screenshot, it's very attention grabbing

I agree, and that's why I prefer #1846, and even more so when we have a way to navigate through the highlights with an element of the UI. ***That*** would be sweet.

I'm really not a fan of the Solarized version though. I get the goal, but I don't think it looks really good, and I think we can do something a bit nicer, but not just with CSS tweaks (i.e. said UI element to navigate).
I'll see if I can come up with something, at least a prototype to validate the idea or trash it.

YaManicKill:
Yeah, I'm not a fan of this one. I think if forced to choose between the status quo, this one and #1846 , I'd go for #1846. I definitely prefer the way @xPaw did it in solarized, but if people don't want that in master, I'm totally fine with it just being in the theme. For default, I can see the reasons for doing #1846 instead, and it's much nicer than what we have. This one is ... a bit hash.

christer88:
I would rather keep the current solution than #1846. #1846 is just not visible enough when scrolling through the chat to see where you was highlighted.

McInkay:
Alright, I think I've changed my mind on this.

I think if I were to choose between this one, status quo and #1846, I'd go for this one now.

I still prefer the one on the solarized theme, but this might be better for the default these. It's definitely more obvious, and that's the point of a highlight.

xPaw:
I'm experimenting further with highlights in Solarized, and find that this approach works quite well:

- Change background color to be lighter
- Small border on the left of the message that is high contrast



I think this approach works quite well for number of reasons:
- Light background color change isn't always visible: low contrast, bad displays
- Changing background color to something substantial: very annoying, hurts text readability
- Using a simple dot/border (like in the other PR) makes it harder to scan messages

McInkay:
I do like it, I think it gives the advantages of all the approaches we've come to, while trying to deal with a couple of the disadvantages of them as well.

The key point is that the side line (or whatever it ends up being, whether it's a dot in default or whatever) needs to be very visible, but the background colour needs to be not *too* different so that everything is still readable.

I notice that you change the colour of the time as well. Is that deliberate?

xPaw:
> I notice that you change the colour of the time as well. Is that deliberate?

Currently, yes. As I said, experimenting with different approaches.

xPaw:
Just to come back to this: I am still using the same style as shown in my screenshot above. I very much enjoy this approach.

astorije:
@xPaw, I like your approach too. I gave it a shot in the bundled themes, using colors given by https://color.adobe.com/ to match our main color, in my branch at #2315 to match it with our actual "new" blue:

screen shot 2018-06-04 at 00 37 44
screen shot 2018-06-04 at 00 44 08

What do you think?

benharri:
@astorije, that looks great!

astorije:
Superseded by https://github.com/thelounge/thelounge/pull/2526.

PR #1850, opened 2017-12-17
Format whois as a definition list

xPaw:

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1483,6 +1483,25 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	display: inline-block;
 }
 
+.whois {
+	display: -ms-grid; /* Edge 15- */
+	display: grid;
+	-ms-grid-template-columns: max-content auto;
+	grid-template-columns: max-content auto;
+	margin: 0;
+}
+
+.whois dt {
+	-ms-grid-column-start: 1;
+	grid-column-start: 1;
+	margin-right: 20px;
+}
+
+.whois dd {
+	-ms-grid-column-start: 2;
+	grid-column-start: 2;
+}
+
 #form {
 	background: #eee;
 	border-top: 1px solid #ddd;
--- a/client/views/actions/whois.tpl
+++ b/client/views/actions/whois.tpl
@@ -1,93 +1,80 @@
-<div>
-	{{> ../user_name nick=whois.nick}}
-	<i class="hostmask">({{whois.user}}@{{whois.host}})</i>
-</div>
-{{#if whois.actualhost}}
-<div>
-	{{> ../user_name nick=whois.nick}}'s actual host is: <i>{{whois.actualhost}}</i> (<i>{{whois.actualip}}</i>)
-</div>
-{{/if}}
-{{#if whois.real_name}}
-<div>
-	{{> ../user_name nick=whois.nick}}'s real name is:
-	<b>{{{parse whois.real_name}}}</b>
-</div>
-{{/if}}
+<p>{{> ../user_name nick=whois.nick}}</p>
+<dl class="whois">
 {{#if whois.account}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	is logged in as <b>{{whois.account}}</b>
-</div>
+	<dt>Logged in as:</dt>
+	<dd>{{whois.account}}</dd>
 {{/if}}
+
+	<dt>Host mask:</dt>
+	<dd class="hostmask">{{whois.user}}@{{whois.host}}</dd>
+
+{{#if whois.actualhost}}
+	<dt>Actual host:</dt>
+	<dd class="hostmask"><a href="https://ipinfo.io/{{whois.actualip}}" target="_blank" rel="noopener">{{whois.actualip}}</a>{{#equal whois.actualhost whois.actualip}}{{else}} ({{whois.actualhost}}){{/equal}}</dd>
+{{/if}}
+
+{{#if whois.real_name}}
+	<dt>Real name:</dt>
+	<dd>{{{parse whois.real_name}}}</dd>
+{{/if}}
+
 {{#if whois.registered_nick}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	{{whois.registered_nick}}
-</div>
+	<dt>Registered nick:</dt>
+	<dd>{{whois.registered_nick}}</dd>
 {{/if}}
-{{#if whois.modes}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	{{whois.modes}}
-</div>
-{{/if}}
-{{#if whois.special}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	{{whois.special}}
-</div>
-{{/if}}
-{{#if whois.operator}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	{{whois.operator}}
-</div>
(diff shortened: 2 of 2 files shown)


MaxLeiter:
I think maybe `Secure` should be `Connection`

client/css/style.css
```diff
@@ -1475,6 +1475,20 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	display: inline-block;
 }
 
+.whois {
+	display: grid;
```
astorije: Look at us, jumping on the grid layout train!
That kills Edge 15 support though, but :man_shrugging: Edge 16 was released in October, so that should be fine.

client/css/style.css
```diff
@@ -1475,6 +1475,20 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+	grid-template-columns: max-content auto;
+}
+
+.whois dt {
+	min-width: 150px;
```
astorije: Is there a way to make the second column start at "<width of longest text in first column> + margin" instead of a fixed width?
Thinking of those with different font family or size for example.

client/views/actions/whois.tpl
```diff
@@ -1,93 +1,79 @@
-<div>
-	{{> ../user_name nick=whois.nick}}
-	<i class="hostmask">({{whois.user}}@{{whois.host}})</i>
-</div>
+<p>{{> ../user_name nick=whois.nick}}</p>
+<dl class="whois">
+	<dt>Host mask</dt>
```
astorije: Could you add a `:` after all the `dt`s?

client/views/actions/whois.tpl
```diff
@@ -1,93 +1,79 @@
...
+	<dd>{{whois.modes}}</dd>
 {{/if}}
+
 {{#if whois.special}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	{{whois.special}}
-</div>
+	<dt>Special</dt>
+	<dd>{{whois.special}}</dd>
```
astorije: What is "special" in practice lol?

client/views/actions/whois.tpl
```diff
@@ -1,93 +1,79 @@
...
+	<dt>Help</dt>
+	<dd>is available for help</dd>
 {{/if}}
 {{#if whois.bot}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	is a bot
-</div>
+	<dt>Bot</dt>
+	<dd>is a bot</dd>
```
astorije: What about "Available for help: Yes/No", "Bot|Is a bot: Yes/No", etc.?

client/views/actions/whois.tpl
```diff
@@ -1,93 +1,79 @@
...
+	<dd>{{whois.special}}</dd>
 {{/if}}
+
 {{#if whois.operator}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	{{whois.operator}}
-</div>
+	<dt>Operator</dt>
+	<dd>{{whois.operator}}</dd>
```
astorije: Is this a boolean or a text value?

client/views/actions/whois.tpl
```diff
@@ -1,93 +1,79 @@
...
+	<dd>is a bot</dd>
 {{/if}}
+
 {{#if whois.channels}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	is on the following channels: {{{parse whois.channels}}}
-</div>
+	<dt>Channels</dt>
+	<dd>{{{parse whois.channels}}}</dd>
```
astorije: It would be cool if we could render this as "channel1, channel2, and channel3" instead of just all joined by a whitespace, but probably for another PR.

client/views/actions/whois.tpl
```diff
@@ -1,93 +1,79 @@
...
 {{/if}}
-{{#if whois.server}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	is connected to {{whois.server}} <i>({{whois.server_info}})</i>
-</div>
+
+{{#if whois.away}}
+	<dt>Away</dt>
+	<dd>{{{parse whois.away}}}</dd>
```
astorije: Isn't this going to always be "Away: true"? If so, could we do "Away: Yes/No" instead?

client/views/actions/whois.tpl
```diff
@@ -1,93 +1,79 @@
...
 {{/if}}
+
 {{#if whois.secure}}
-<div>
-	{{> ../user_name nick=whois.nick}}
-	is using a secure connection
-</div>
+	<dt>Secure</dt>
+	<dd>using a secure connection</dd>
```
astorije: I agree with @MaxLeiter, "Connection: Secure/Not secure" or "Secure connection: Yes/No" would be better IMO.

In general, if we can avoid making the keys optional based on the values (when possible), that would be sweet. I think it's better to see for example that a user is not away rather than not seeing anything if they're not away.

client/views/actions/whois.tpl
dgw: In markup or using `:after` in styles?

astorije: I would do it in markup, because it's relevant for a copy-and-paste.

client/views/actions/whois.tpl
creesch: Afaik it is a text value, you can actually see what is and isn't by looking at the current whois.tpl. When I did the previous PR for that I translated booleans to something userfriendly will directly showing anything else.

client/views/actions/whois.tpl
creesch: Afaik it also shows the away reason.

client/views/actions/whois.tpl
creesch: Whatever pleases the network.

client/css/style.css
creesch: > Edge 16 was released in October, so that should be fine.

Awfully optimistic about that as it is tied in with the fall creators update people have to install. Not to mention people limited by the IT department at their work. 
I am not saying the lounge should support older browser until the end of times, but I'd say this might be a bit soon as far as browser support goes. 

Seems like you can be a bit more in the clear using a -ms- prefix here.

client/views/actions/whois.tpl
xPaw: Displaying all possible fields that aren't set would be a lot of useless information, in my opinion.

client/css/style.css
astorije: Yeah, `-ms-` for grid layout is reasonable. We usually support "last 2 browsers", though we have no test running on browsers to kind of prove it (nor correct tests for that) yet.

client/views/actions/whois.tpl
astorije: It depends, for example secure/insecure connection might be a good one to always display.
I'd favor negative (as in, not `true`) information over no information. For example, I never realized we were showing if the user was a bot or not. I'd have to find a bot to discover, and wonder otherwise. Not a terrific experience :/

client/css/style.css
```diff
@@ -1483,6 +1483,25 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	display: inline-block;
 }
 
+.whois {
+	display: -ms-grid; /* Edge 15- */
```
YaManicKill: Euch, can we add autoprefixer into webpack rather than adding these into the source?

Issue #1825, opened 2017-12-11
Provide optional syncing of client settings.


creesch:
Issue made after discussing it on irc. The idea is to implement client settings sync based on the following principles. 

1. Settings sync is an *optional* setting per client. Explicitly done like this to prevent people loosing their current settings. 
2. All settings will be synced. More fine grained settings sync will be for a follow up. The idea is to start simple and based on the fact that there currently is no sync so this is already an improvement. Previous attempts stranded on trying to go for fine grained from the get go. 

The work flow for this implementation will be roughly something like this. For clients that have settings sync enabled: 

- On client load settings are picked up from the server and applied. (this requires reading the current client settings to see if this is the case.)
- When a setting is changed in a client the settings object will be saved to the server. 
- The server pushes the changed settings to other sync enabled clients that are currently active. To keep things simple settings will not be applied on the fly but the user will be presented with a message giving them the option to reload the client or disable syncing for that client. Something roughly like this 
![image](https://user-images.githubusercontent.com/2502190/33842794-dbe5180e-de9b-11e7-9dd4-980f9209e97a.png)

creesch:
Previous attempt https://github.com/thelounge/lounge/compare/settings

xPaw:
Important to note specific settings like "custom highlights" would have to be always synced to provide correct support for push notifications. Right now these custom highlights are done entirely on the client and not the server.

astorije:
Hey @creesch, thanks for taking that up!

I think however this should be handled in a different way: at least for now, there should be no UI change introduced by this whatsoever.
In general, I don't think the user should be prompted for settings sync. Heck, I'm trying to think really hard on all the apps I use on multiple devices that have settings, and of the dozen-or-so I daily use, I have never seen something like this.

So anyway, this is how I think we should handle this:

- Some settings like the custom highlights, should be always synced. Like @xPaw said, we currently do that on the client, and it's hurting: push notifications do not have a way of knowing what they are.
  There are other things we currently do not handle correctly because of lack of sync, such as the auto-away message: it is currently stored in the user's JSON configuration file. Not user-friendly at all :D This is also probably true for the `commands` array we store in user config

[...]

PR #1851, opened 2017-12-17
Optional syncing of client settings. 

creesch:
This provides syncing as detailed in #1825. Some demonstrations

- [General behavior of syncing and unchecking the box.](http://creesch.com/dump/vid/vid_5a3432fc0f6dc.mp4)
- [Enabling sync for the very first time](https://www.creesch.com/dump/vid/vid_5a3692e0339c4.mp4)

## Note on always syncing some settings. 

This PR includes the ability to make some settings mandatory to sync regardless of syncing in general being on or not. This works fine, however I didn't include any settings as migration is a consideration. The ability was requested in relation to custom highlights, not working for people with push enabled.  So when people have no push enabled then it isn't broken to them.  Meaning that it is likely that some people have explicitly set it differently in different clients, which means that a migration strategy is needed. 

Because the goal of this PR is to implement syncing in general first I decided to not turn mandatory sync yet. I left the ability in as the goal of this PR is to have a basis to work on from here on out. 

## options.js refactoring 
In addition this PR also reworks `client/js/options.js` to be more reliable, easier to understand and generally more robust. Functionality that was all randomly called on has been divided in separate functions: 

- Initialization 
- Processing 
- Saving 
- Applying 

Meaning that settings are now actually only saved and applied when things actually change. Before it could easily happen that when nothing was actually changed in settings there still would be multiple write actions and attempts at applying settings to the client where it was not needed. 

It also changes the `module.exports` object to separate user settings from other information and functions, potentially preventing outside modules from accidentally overwriting settings. 

Making options.js robust enough to be able to handle synced settings was the biggest part of the work that needed to be done.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -2,15 +2,28 @@
 
 const $ = require("jquery");
 const escapeRegExp = require("lodash/escapeRegExp");
-const userStyles = $("#user-specified-css");
 const storage = require("./localStorage");
 const tz = require("./libs/handlebars/tz");
+const socket = require("./socket");
 
-const windows = $("#windows");
-const chat = $("#chat");
+const $windows = $("#windows");
+const $chat = $("#chat");
+const $settings = $("#settings");
+const $theme = $("#theme");
+const $userStyles = $("#user-specified-css");
 
-// Default options
-const options = {
+const noCSSparamReg = /[?&]nocss/;
+
+// Not yet available at this point but used in various functionaly.
+// Will be assigned when `initialize` is called.
+let $syncWarningOverride;
+let $syncWarningBase;
+let $warningUnsupported;
+let $warningBlocked;
+
+// Default settings
+const settings = {
+	syncSettings: false,
 	autocomplete: true,
 	nickPostfix: "",
 	coloredNicks: true,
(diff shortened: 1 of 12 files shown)


client/js/socket-events/init.js
```diff
@@ -4,12 +4,18 @@ const $ = require("jquery");
...
 const storage = require("../localStorage");
 const utils = require("../utils");
 
 socket.on("init", function(data) {
+	if (options.initialized) {
+		return;
+	}
+	options.initialize(data.clientSettings ? data.clientSettings : "");
+	webpush.initialize();
 	$("#loading-page-message, #connection-error").text("Rendering…");
```
xPaw: Put rendering above initializes.

client/js/socket-events/init.js
```diff
@@ -4,12 +4,18 @@ const $ = require("jquery");
...
 const socket = require("../socket");
 const render = require("../render");
+const options = require("../options");
 const webpush = require("../webpush");
 const sidebar = $("#sidebar");
 const storage = require("../localStorage");
 const utils = require("../utils");
 
 socket.on("init", function(data) {
+	if (options.initialized) {
```
xPaw: I'm quite certain doing this breaks reconnections.

src/server.js
```diff
@@ -388,6 +388,42 @@ function initializeClient(socket, client, token, lastMessage) {
 
 	socket.on("sessions:get", sendSessionList);
 
+	socket.on("setting:set", (newSetting) => {
+		// Older user configs will not have the clientSettings property.
+		if (!client.config.hasOwnProperty("clientSettings")) {
+			client.config.clientSettings = {};
+		}
+		// We do not need to do write operations and emit events if nothing changed.
+		if (!client.config.clientSettings.hasOwnProperty(newSetting.name) || client.config.clientSettings[newSetting.name] !== newSetting.value) {
```
xPaw: `hasOwnProperty` is unnecessary here.

src/server.js
```diff
@@ -388,6 +388,42 @@ function initializeClient(socket, client, token, lastMessage) {
...
+			});
+		}
+	});
+
+	socket.on("settings:get", () => {
+		if (!client.config.hasOwnProperty("clientSettings")) {
+			socket.emit("setting:empty");
+		} else {
+			const clientSettings = client.config.clientSettings;
+			for (const setting in clientSettings) {
```
xPaw: Seems very wasteful to emit separate packets for each setting. You're better off doing one emit and handling the loop on the client.

client/js/socket-events/init.js
creesch: This is already done in a very similar way currently, are you sure about that?

xPaw: Your change blocks future `init` calls, which happen after reconnecting. On master it blocks `configuration` which is fine.

src/server.js
creesch: Fair enough, I'll rework that.

src/server.js
```diff
@@ -388,6 +388,42 @@ function initializeClient(socket, client, token, lastMessage) {
...
+			client.emit("setting:new", {
+				name: newSetting.name,
+				value: newSetting.value,
+			});
+		}
+	});
+
+	socket.on("settings:get", () => {
+		if (!client.config.hasOwnProperty("clientSettings")) {
+			socket.emit("setting:empty");
```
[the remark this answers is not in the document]
creesch: Yup, though I'd think it is better to still make it two events 

  - `setting:new` which is already a thing for when individual settings are saved and synced and we know the setting is new.. 
  - `settings:all` for the entire package where the client can figure out looping etc if needed. 

Makes things more explicit and also reduces the amount of looping that needs to be done in general for figuring out what needs to be changed.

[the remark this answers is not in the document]
xPaw: If both of these events will handle settings in the same way (except for the loop), then I don't see a reason to do separate events.

src/server.js
```diff
@@ -388,6 +388,34 @@ function initializeClient(socket, client, token, lastMessage) {
...
+			client.emit("setting:new", {
+				name: newSetting.name,
+				value: newSetting.value,
+			});
+		}
+	});
+
+	socket.on("settings:get", () => {
+		if (!client.config.hasOwnProperty("clientSettings")) {
+			socket.emit("setting:all", {});
```
xPaw: You forgot a return.

creesch: Actually forgot the else, but return works as well. Fixing that now.

client/views/windows/settings.tpl
```diff
@@ -5,6 +5,23 @@
...
+			<h2>
+				Settings synchronisation
+			</h2>
+			<p class="opt">
+				<input type="checkbox" name="syncSettings">
+				Synchronize settings with other clients. <br>
+				<strong>Warning</strong> Checking this box will override current client settings.
+			<p>
+			<p id="master-sync" class="opt">
+				No settings synced yet, sync settings from this client as base set?
```
YaManicKill: I'm not convinced on the value of this?

Under what circumstances would someone want to sync when there are no server settings and *not* set these as the ones currently synced.

client/js/socket-events/configuration.js
```diff
@@ -7,10 +7,9 @@ const options = require("../options");
 const webpush = require("../webpush");
 
 socket.on("configuration", function(data) {
-	if (!options.initialize) {
+	if (options.initialized) {
```
YaManicKill: Hmmm, why this change?

client/js/options.js
```diff
@@ -22,156 +35,263 @@ const options = {
...
-	userStyles: userStyles.text(),
+	userStyles: "",
 };
-let userOptions = JSON.parse(storage.get("settings")) || {};
 
-for (const key in options) {
-	if (userOptions[key] !== undefined) {
-		options[key] = userOptions[key];
+const noSync = ["syncSettings"];
+// Reserved for things like "highlights".
```
YaManicKill: Nitpick: it's not clear which this is referring to. I initially thought it was noSync it was talking about.

client/js/socket-events/configuration.js
creesch: Before it wasn't a boolean and checked something else.

client/views/windows/settings.tpl
creesch: The first time someone does enable syncing there is nothing on the server yet. We could automatically sync the settings of that client but they might not have intended that. [See also this video](https://www.creesch.com/dump/vid/vid_5a3692e0339c4.mp4) for how it functions. 
This will basically give them the opportunity to go "Oh crap, I didn't want to sync this client but my other one". 

Basically done from the same reasoning as how I approached the advanced settings classification the other day :)

client/js/options.js
creesch: The comment you mean?

client/js/socket-events/configuration.js
creesch: To be a bit more clear, before it set `module.exports.initialize` to `null` which works (but a bit overkill) but also resulted in the negative check for initialize which is slightly confusing as it is more of a `true` situation as options are in fact initialized. 

tl;dr readability of code mostly.

client/views/windows/settings.tpl
YaManicKill: I don't think you've convinced me of this one. I think it's a bit of a weird thing, but ... I'll let others say what they think. The rest I really like, I'm just not convinced about this 1 button press for when there are no synced settings and never again.

client/js/socket-events/configuration.js
YaManicKill: Aha, I see

client/js/options.js
YaManicKill: Yeah

client/views/windows/settings.tpl
creesch: As per the discussion on irc I'll remove the manual step and will instead display something like 

>  "No settings have been synced before. Enabling this will make the settings of this client the base settings."

client/js/options.js
```diff
@@ -22,156 +36,269 @@ const options = {
 	notifyAllMessages: false,
 	showSeconds: false,
 	statusMessages: "condensed",
-	theme: $("#theme").attr("href").replace(/^themes\/(.*).css$/, "$1"), // Extracts default theme name, set on the server configuration
+	theme: $theme.attr("href").replace(/^themes\/(.*).css$/, "$1"), // Extracts default theme name, set on the server configuration
```
xPaw: Can you change all `.attr` to `.prop` please, this was done on master.

creesch: Yeah sure. 

Out of curiosity, why has it been changed?

xPaw: https://github.com/thelounge/lounge/pull/2015#discussion_r164976035

creesch: Changed it. Also rebased and squashed it.

McInkay:
I think this PR should be merged (if we want this) reasonably soon. There are a lot of extra pieces of functionality I want to implement that will change code here, and the improvements to `options.js` will be necessary for them.

As a side-note, this can easily be added as an advanced feature if #1860 is merged. It can be done in either PR depending on which is done before, but neither will conflict much with the other, as that one is almost entirely in html (almost).

creesch:
Yup, this is probably one in the advanced section.

src/server.js
```diff
@@ -432,6 +432,35 @@ function initializeClient(socket, client, token, lastMessage) {
 
 	socket.on("sessions:get", sendSessionList);
 
+	socket.on("setting:set", (newSetting) => {
```
xPaw: Don't bind these setting events in public mode.

creesch:
@xPaw made the changes to public mode, not sure why Travis is failing on node9. Seems like a timeout and not an actually failed test.

client/js/options.js
```diff
@@ -2,15 +2,29 @@
...
+const $userStyles = $("#user-specified-css");
 
-// Default options
-const options = {
+const noCSSparamReg = /[?&]nocss/;
+
+// Not yet available at this point but used in various functionaly.
+// Will be assigned when `initialize` is called.
+let $syncWarningOverride;
+let $syncWarningBase;
```
MaxLeiter: I don’t think we prefix element objects anywhere else with $

creesch: It is a good habit to do it with jqueury objects, makes it easier to read code.

client/views/windows/settings.tpl
```diff
@@ -5,6 +5,19 @@
 	<h1 class="title">Settings</h1>
 
 	<div class="row">
+		{{#unless public}}
+		<div class="col-sm-12">
+			<h2>
+				Settings synchronisation
+			</h2>
+			<p class="opt">
+				<input type="checkbox" name="syncSettings">
```
xPaw: Make this a correct `<label>` so the text is also clickable. (look at other checkboxes)

Don't include warnings as part of the label though.

creesch: I explicitly didn't do that as that creates a bunch of issues with settings styling and how things are aligned and you really don't want the text to be part of the label as you don't want to accidentally activate this. 

Also just having the input element in a label doesn't make much sense to me?

creesch: Done

client/views/windows/settings.tpl
```diff
@@ -5,6 +5,19 @@
...
+		<div class="col-sm-12">
+			<h2>
+				Settings synchronisation
+			</h2>
+			<label class="opt">
+				<input type="checkbox" name="syncSettings">
+				Synchronize settings with other clients.
+			</label>
+			<span class="sync-warning-override"><strong>Warning</strong> Checking this box will override the settings of this client with those stored on the server.</span>
+			<span class="sync-warning-base"><strong>Warning</strong> No settings have been synced before. Enabling this will sync all settings of this client as the base for other clients.</span>
```
astorije: Could use `p` elements instead of `span`s? That will fix the extra font size there is right now.

src/server.js
```diff
@@ -441,6 +441,44 @@ function initializeClient(socket, client, token, lastMessage) {
...
+					value: newSetting.value,
+				});
+
+				client.manager.updateUser(client.name, {
+					clientSettings: client.config.clientSettings,
+				});
+			}
+		});
+
+		socket.on("settings:get", () => {
```
astorije: In terms of socket events, there are both `setting` (including `setting:all`) and `settings` like this guy. Do we want to use `settings` everywhere instead?

client/views/windows/settings.tpl
creesch: I actually liked the warning being slightly bigger, but yeah it can also be changed if you prefer so.

astorije: That would be great, thanks :)

astorije:
Hey @creesch, I quickly looked at the code and tested, and didn't notice any bugs. For the full, detailed review, I'll defer to @xPaw and @McInkay, who have been very involved in this so far, so I think I would not bring anything else useful to the table.
Feel free to ping me if you need a hand with, like, final UI or something like this, but I don't think this will massively change.

src/server.js
creesch: Yeah that makes sense, initially I had the singular and plural because one was going to pass multiple settings and the other not. That changed along the way so it makes sense to have consistent naming there.

creesch:
@astorije made those changes. 

At the request of @xPaw also added an extra warning note that looks like this

image

As a bonus, that can in the future also be used for other things.

astorije:
Hey @creesch, we have free sticker packs for our contributors now!

PR #1852, opened 2017-12-18
[WIP] Download log files from the UI

astorije:
Highly WIP state right now. It works pretty well... for Freenode only 😅
Lots of TODOs and FIXMEs I still need to work on.

<p align=center>
screen shot 2017-12-18 at 00 47 07
</p>

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -133,6 +133,16 @@ $(function() {
 					data: target.data("id"),
 				});
 			}
+
+			if (!target.hasClass("lobby")) { // TODO: Download all logs for a network
+				output += templates.contextmenu_item({
+					class: "download-logs",
+					action: "download_logs",
+					text: "Download logs",
+					data: target.data("id"),
+				});
+			}
+
 			output += templates.contextmenu_item({
 				class: "close",
 				action: "close",
@@ -510,15 +520,16 @@ $(function() {
 				text: "/kick " + itemData,
 			});
 		},
+		download_logs: function(channelId) {
+			socket.emit("logs:download", channelId);
+		},
 	};
-
-	contextMenuActions.execute = (name, ...args) => contextMenuActions[name] && contextMenuActions[name](...args);
 
 	contextMenu.on("click", ".context-menu-item", function() {
 		const $this = $(this);
 		const itemData = $this.data("data");
 		const contextAction = $this.data("action");
-		contextMenuActions.execute(contextAction, itemData);
+		contextMenuActions[contextAction](itemData);
 	});
 
 	chat.on("input", ".search", function() {
--- a/src/server.js
+++ b/src/server.js
@@ -384,6 +384,22 @@ function initializeClient(socket, client, token, lastMessage) {
 		client.unregisterPushSubscription(token);
 	});
 
+	socket.on("logs:download", (channelId) => {
+		const result = client.find(channelId);
+		// FIXME: result can be false. Is that possible here? If so, fix
+		const host = result.network.host;
+		const channel = result.chan.name;
+		const logFolder = Helper.getUserLogsPath(client.name, host);
+		// TODO: When supporting entire networks, retrieve entire folder
+		const filename = path.resolve(logFolder, `${channel}.log`);
+		const logContent = fs.readFileSync(filename, "utf8");
+
+		socket.emit("logs:download", {
+			filename: `${channel}.txt`, // TODO: Extension will vary with entire networks
+			content: logContent,
+		});
+	});
+
 	const sendSessionList = () => {
 		const sessions = _.map(client.config.sessions, (session, sessionToken) => ({
 			current: sessionToken === token,
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -223,6 +223,7 @@ kbd {
 .context-menu-user::before { content: "\f007"; /* http://fontawesome.io/icon/user/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 .context-menu-list::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
+.context-menu-download-logs::before { content: "\f0ed"; /* http://fontawesome.io/icon/cloud-download/ */ }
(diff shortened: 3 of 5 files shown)


src/server.js
```diff
@@ -373,6 +373,19 @@ function initializeClient(socket, client, token, lastMessage) {
 		client.unregisterPushSubscription(token);
 	});
 
+	socket.on("logs:download", (data) => {
+		const network = "irc.freenode.org"; // FIXME: Retrieve this from client one way or another
+		const logFolder = Helper.getUserLogsPath(client.name, network);
```
astorije: @xPaw, do you have any suggestion for passing network + channel/query name from the client to the server?
I believe the client doesn't know the URL of the network, just the display name (at least for now), and I need it to browse the correct folder.
I'm thinking that instead of passing string of network+channel, I could instead pass the ID of the channel that the server knows about. Is that possible? Any suggestion?

creesch: Making the host available is fairly easy as it is already send to the client with the "init"  event as can be seen here https://github.com/thelounge/lounge/blob/36be7565a700ab43c3f7e8ef4844d7886f88f498/src/server.js#L431 

So all it takes is to add `data-host="{{host}}"` somewhere in `network.tpl` and it is available. 

The same data is available server side so you could use the id as that simply sees to be an array key+1.

xPaw: I'd wait for https://github.com/thelounge/lounge/pull/1839/commits/d085b0692d4b047fb4ee7ffae005e4ad20c86df3 until making logs downloadable to be honest.

client/js/lounge.js
```diff
@@ -510,15 +520,18 @@ $(function() {
...
 			});
 		},
+		download_logs: function(itemData) {
+			socket.emit("logs:download", {
+				name: itemData,
+			});
+		},
 	};
 
-	contextMenuActions.execute = (name, ...args) => contextMenuActions[name] && contextMenuActions[name](...args);
```
creesch: Any reason you decided to rip this out? You have left it less flexible for any future changes.

creesch: Also this was made a command pattern specifically because of a review done on a early PR...

src/server.js
```diff
@@ -373,6 +373,19 @@ function initializeClient(socket, client, token, lastMessage) {
...
+		// TODO: When supporting entire networks, retrieve entire folder
+		const filename = path.resolve(logFolder, `${data.name}.log`);
+		const logContent = fs.readFileSync(filename, "utf8");
+
+		socket.emit("logs:download", {
```
xPaw: Sending megabytes of data over socket.io is not a great idea by the way, a better idea would be just expose logs with express and some kind of access token for that specific user.

src/server.js
astorije: > I'd wait for d085b06 until making logs downloadable to be honest.

I was thinking about that, but also I'd like to not depend on something else to be able to make progress on this. I don't mind reworking this PR once d085b06 makes it in.
I'm also thinking of just cherry-picking it for the time being, but I feel like this PR might be ready before #1839.

client/js/lounge.js
astorije: For the record if someone stops by, we discussed this in the channel. There is nothing definitive here and I'm not planning on breaking anything :) This PR as a whole is WIP.

src/server.js
astorije: Yes I've considered that, but considering how sensitive this information is, I would like to avoid at all cost exposing it via Express. I can't imagine the blast radius of security fail if we have a bug with exposing this one thing.

I'm also considering always compressing this on the server before sending. Of course, there are a few channels in which that would be an issue, but I think overall this could easily work for like 95% cases. Then I think it doesn't have to be perfect at first attempt, we can refine.

astorije: > Sending megabytes of data over socket.io is not a great idea by the way

What makes you say that, btw? I'm curious as I couldn't find evidence of that somewhere else.
There are however ways (packages and such) to send large files via sockets, or stream them, etc. Again, not thinking all the edge cases for now, just something that works.

src/server.js
astorije: It turns out, this was muuuch easier than I thought, and didn't require any trickery. As long as we're passing the channel id to the server, we can `find` whatever we need.

McInkay:
What's the status of this one, @astorije ? Much left to do? I notice you have put "do not merge" on it. Could this possibly work better as a plugin (which we are nearly able to do, as I've got the code working for context menus, and then we just need to allow plugins to access logs via an api that uses whatever method the user wants).

astorije:
Hey @McInkay thanks for the ping. Yeah, quite a few things left to do, but also this will rather change once #1839 gets merged, and I definitely want that other PR merged before this current feature. I'll close it for now.

I'm not against this living in a package, but only if we can make packages ship by default. It's an important feature to have by default IMO, let users have access to (and control, ultimately) their data.
However, the meat of it is not the context menu part but the storage access and how are the artifacts exposed to the user.
We've mentioned pluggable storage, in which case this PR would become much easier, though that's just translating the complexity somewhere else :)

michael-thayer-vb:
Is this something likely to be taken up again?  If not, how hard would it be for someone unfamiliar with the code base, assuming that you would want that?

astorije:
Probably not too hard to do, but a bit of a learning curve for someone unfamiliar.

This would involve generate a temporary URL that gets invalidated after, say 1 download (or time based, or a combination of both).

michael-thayer-vb:
I take it from the answer that it is no longer something you plan to add yourself.  Any pointers to where in the code to start, or just "read the source"?

astorije:
@michael-thayer-vb, definitely read the entire thread in this PR, there is a lot of insight.

It would be worth making some research around sending large files over HTTP vs. the WebSocket. After briefly reading about it, I am not convinced HTTP is the way to go. It would have been true 1-2 years ago, but I have read opposing arguments saying that the web has evolved and has dedicated ways to send large binary content over a WebSocket efficiently (whether socket.io supports those efficient methods or not is another thing to look into).
It would for sure make things easier and secure/private by design since there would be no manual auth via temporary token to come up with.

PR #1855, opened 2017-12-19
Update to primer on npm

MaxLeiter:
https://github.com/primer/primer/tree/master/modules/primer-tooltips


client/index.html
```diff
@@ -9,6 +9,7 @@
 	<link rel="preload" as="script" href="js/bundle.js">
 	<link rel="stylesheet" href="css/bootstrap.css">
 	<link rel="stylesheet" href="css/style.css">
+	<link rel="stylesheet" href="css/primer-tooltips.css">
```
astorije: I think we want this to be above `style.css`, just in case.

client/index.html
```diff
@@ -36,11 +37,11 @@
...
-			<span class="tooltipped tooltipped-n tooltipped-no-touch" aria-label="Sign out"><button class="icon sign-out" id="sign-out" aria-label="Sign out"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch tooltipped-no-delay" aria-label="Sign in"><button class="icon sign-in" data-target="#sign-in" aria-label="Sign in"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch tooltipped-no-delay" aria-label="Connect to network"><button class="icon connect" data-target="#connect" aria-label="Connect to network"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch tooltipped-no-delay" aria-label="Client settings"><button class="icon settings" data-target="#settings" aria-label="Client settings"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch tooltipped-no-delay" aria-label="Help"><button class="icon help" data-target="#help" aria-label="Help"></button></span>
+			<span class="tooltipped tooltipped-n tooltipped-no-touch tooltipped-no-delay" aria-label="Sign out"><button class="icon sign-out" id="sign-out" aria-label="Sign out"></button></span>
```
astorije: Why did you re-add `tooltipped-no-delay` to all of those? FYI, we used to not have delays on those and added them because:

- It made for a very schizophrenic UI when constantly hovering tooltipped stuff
- On mobile clicking somewhere on the menu would instantly display a tooltip + close the menu, that was very buggy

We went back and forth before we reached the current okay solution, I think we should keep it that way. That is, unless I'm missing something obvious :)

client/views/msg.tpl
```diff
@@ -1,5 +1,5 @@
 <div class="msg {{type}}{{#if self}} self{{/if}}{{#if highlight}} highlight{{/if}}" id="msg-{{id}}" data-time="{{time}}"{{#if from.nick}} data-from="{{from.nick}}"{{/if}}>
-	<span class="time tooltipped tooltipped-e" aria-label="{{localetime time}}">
+	<span class="time tooltipped tooltipped-e tooltipped-no-delay" aria-label="{{localetime time}}">
```
astorije: Yeah so this was particularly horrible without delay, because since times are like a vertical barrier between chat and channel list, you would *always* trigger them...

client/index.html
MaxLeiter: I think the delay was increased in primer, they feel very sluggish without. Can remove

client/views/msg.tpl
MaxLeiter: fair

client/index.html
astorije: Yeah I should have added: I am pretty sure we customized that value in our CSS so we should overwrite it in `style.css`.

In general, we made a few changes between our version and the upstream project. I am building you a diff so we can have an idea of what we're talking about.

astorije:
FYI this is the diff between our modified v0.53 and upstream v1.5.0: https://www.diffchecker.com/XZd59jmz

We should be fine overall.

astorije:
There are sizes and spacing differences, I don't know if we want to tweak those or not:

Our version | Upstream project
--- | ---
screen shot 2017-12-18 at 22 52 06 | screen shot 2017-12-18 at 22 52 15

Text is a bit smaller, tooltip has a bit more height.

xPaw:
We need to compile our CSS already :(

MaxLeiter:
Would still need to directly reference the primer css file, right?

MaxLeiter:
The size is fine to me, up to you and rest of @thelounge/maintainers, astorije

astorije:
Looking at https://www.diffchecker.com/dd47az8a, I'm fine with the bigger paddings because I absolutely do not want to have to maintain our overrides of all margins and such. It's reasonable to rely on the library for that.

The only thing I would do is keep our current font-size. I think it works better in our UI and makes things more readable (as opposed to paddings that are more on the side of habits / preferences). Also, it's a single override, not many lines we have to maintain over the upgrades. Would you mind adding the following in `style.css`?

```css
.tooltipped::after {
	font-size: 12px;
}
```

If @xPaw or @YaManicKill have strong opinions, I'll go with the flow :)

MaxLeiter:
Done

PR #1856, opened 2017-12-19
Add tooltip to channel close button

MaxLeiter:
Maybe this was left without a tooltip intentionally; unsure why if so.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -628,6 +628,11 @@ kbd {
 	color: #49505a;
 }
 
+#sidebar .close-tooltip {
+	float: right;
+}
+
+#sidebar .leave,
 #sidebar .close {
 	width: 18px;
 	height: 18px;
@@ -635,6 +640,7 @@ kbd {
 	transition: opacity 0.2s, background-color 0.2s;
 }
 
+#sidebar .leave::before,
 #sidebar .close::before {
 	font-size: 20px;
 	font-weight: normal;
@@ -677,6 +683,7 @@ kbd {
 	display: unset;
 }
 
+#sidebar .chan.active .leave:hover,
 #sidebar .chan.active .close:hover {
 	opacity: 1;
 }
--- a/client/views/chan.tpl
+++ b/client/views/chan.tpl
@@ -6,7 +6,11 @@
 		</span>
 	{{/equal}}
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-	{{#notEqual type "lobby"}}<button class="close" aria-label="Close"></button>{{/notEqual}}
+	{{#notEqual type "lobby"}}
+		<span class="close-tooltip tooltipped tooltipped-w" aria-label="Close">
+			<button class="close" aria-label="Close"></button>
+		</span>
+	{{/notEqual}}
 	<span class="name" title="{{name}}">{{name}}</span>
 </div>
 {{#equal type "lobby"}}


client/views/chan.tpl
```diff
@@ -1,7 +1,9 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-	<button class="close" aria-label="Close"></button>
+	<span class="close-tooltip tooltipped tooltipped-w tooltipped-no-delay" aria-label="Close">
```
astorije: Mind removing `tooltipped-no-delay`?

client/css/style.css
```diff
@@ -574,6 +574,10 @@ kbd {
 	color: #49505a;
 }
 
+#sidebar .close-tooltip {
+	float: right;
+}
+
```
astorije: Tooltip gets bold on lobbies. I don't know if #1855 fixes it by accident.

MaxLeiter: It's because the lobby name is bold, not sure of best fix

astorije: Yeah I know, but I'm wondering if #1855 does not fix this. Their version and ours differ slightly and it might force to be non-bold.

client/views/chan.tpl
```diff
@@ -1,7 +1,9 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-	<button class="close" aria-label="Close"></button>
+	<span class="close-tooltip tooltipped tooltipped-w" aria-label="Close">
```
astorije: Could you change so that it's "Leave" on channels and "Close" on queries, like the context menu says? Don't worry about lobbies, we're getting rid of the close button there in #1833.

client/views/chan.tpl
```diff
@@ -1,7 +1,16 @@
 {{#each channels}}
 <div data-id="{{id}}" data-target="#chan-{{id}}" data-title="{{name}}" class="chan {{type}} chan-{{slugify name}}">
 	<span class="badge{{#if highlight}} highlight{{/if}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-	<button class="close" aria-label="Close"></button>
+		{{#equal type "channel"}}
+			<span class="close-tooltip tooltipped tooltipped-w" aria-label="Leave">
+				<button class="leave" aria-label="Leave"></button>
```
astorije: Leave this class as `close`, less risk to break some consistency/action IMO. Also, you let the tooltip class as `close-tooltip` for both, which kinda shows what I mean ahah.

MaxLeiter: That was on purpose, as to avoid adding a CSS line and I couldn't think of a more generic classname, but yeah point taken.

astorije: Just use `.close` for both and it'll be fine.

astorije:
@MaxLeiter, could you fix conflicts and address comments on this? It's one of those small and easy PRs we can put behind us :)

MaxLeiter:
Done, @astorije 
screen shot 2018-01-29 at 9 36 48 am

client/css/style.css
astorije: Never closed the loop, but yes, current primer-tooltip (updated after this PR was opened) fixed this.

PR #1858, opened 2017-12-19
Add banlist context menu item if user isn't normal user

YaManicKill:


client/js/lounge.js
```diff
@@ -133,6 +133,14 @@ $(function() {
 					data: target.data("id"),
 				});
 			}
+			if (target.hasClass("channel") && utils.hasUserModeInChannel($(`#chan-${target.data("id")}`))) {
```
YaManicKill: I'm not 100% on when to show this item.

On the one hand, on all the irc servers I use, any user can view the banlist, so should we show it for everyone? On the otherhand, I thought that only people with some sort of privileges in the channel would have a **reason** to view it, so we should probably only show it for them.

Open to suggestions.

client/js/utils.js
```diff
@@ -48,6 +49,15 @@ function isOpInChannel(channel) {
 	return isOP;
```
YaManicKill: As a side note, this only works for "op", doesn't work for "halfop" "owner", "admin" or anything else. So literally it tells you if a user is op, if the user is more privileged than op, it'll return false.

client/js/utils.js
```diff
@@ -48,6 +49,15 @@ function isOpInChannel(channel) {
...
 }
 
+// Given a channel element will determine if the lounge user has a user mode other than norml
+function hasUserModeInChannel(channel) {
```
YaManicKill: We could probably find a nice way to combine these functions (allow the user to pass in a minimum privilege to check for.

client/js/lounge.js
xPaw: Yeah I see no reason to limit this to user modes, just display it to everyone.

client/js/lounge.js
```diff
@@ -133,6 +133,14 @@ $(function() {
...
+				output += templates.contextmenu_item({
+					class: "list",
+					action: "banlist",
+					text: "Show banlist",
```
xPaw: Add a space in banlist.

client/js/lounge.js
creesch: It might be nice to not show it for regular people as makes the functionality more accessible where people now might not be aware of it being possible unless they run a channel. It might result in some soapboxing in some channels if people are made more aware of the banlist. 

Technically there is no reason not to show it indeed.

client/js/utils.js
creesch: Agreed, I probably made the isOp function a it narrow in scope as I had a very specific target in mind. Maybe make it possible to pass an array of possible modes to check for so the result can just be `passedArray.contains(userMode)`

client/js/utils.js
YaManicKill: Sure, but we currently have some support for them, so we should probably deal with them in this somehow.

client/js/lounge.js
YaManicKill: Yeah, my reasoning is 2-fold:

1. It makes the context list busier when there's probably no reason for you to be calling the banlist if you aren't at least voiced
1. Making it more accessible for users can lead to them unnecessarily calling it, and could lead to issues as explained by @creesch 

If people think we should just show it for everything, that's fine. @astorije , what do you think with your UX hat on?

client/js/utils.js
YaManicKill: That could be a nice way to do it.

client/js/lounge.js
YaManicKill: Done

client/js/utils.js
YaManicKill: Alright, I've updated it to this, defaulting to all of the roles (other than normal). Let me know what you think.

creesch: Yup looks good to me, much more versatile.

client/js/lounge.js
astorije: I can see both sides of the argument, but I am in favor of displaying it for normal users. I'll try to list my reasons here (don't see that as patronizing, I'm just too tired to gather thoughts properly and listing helps me compartimentalize :D).

- Users can still send the `/banlist` command, so it's not like we're completely blocking the feature (though we could, but I don't think we should), we're just hiding it.
- Adding a UI element for it helps with discoverability, which is nice.
- The context menu becoming busier is a valid point, but at the moment we only have 1 real entry in there (2 if you count the entry that clicks on the target :) ) so I don't see that as an issue for now. We can for sure re-think that when it becomes a problem, and there are paradigms for that, such as grouping entries, or re-evaluate what we display in there.
- You rightfully mention that this is primarily useful for "power users" or actually "voiced or op'ed users", which is a good reason why this shouldn't be in the main UI (i.e. constantly on the page, like #1836 adds). But this is in a context menu, which is a good place where to put advanced features without burying them completely, I think.
- Though it's clearly useful for voiced/op users, I can see how it's useful for regular users. Just like the question someone asked today "why would someone connect to the same network multiple times" (which is even less obvious IMO ahah), I'm sure there are plenty use cases for not being voiced on a channel you actively contribute/operate, or checking banlists on channels you don't operate, etc. In fact, I was working in a (private) team where no one was ever op or voiced to leave everyone on the same hierarchy level. Workflows :man_shrugging: 
- Re: soapboxing (thanks, I learned a new word :D), I get what you're saying, but I don't think it's our business at least in this case. If we were a SaaS solution or something proprietary like that, I'd probably have a different point of view.

So yeah, I'd just display it :)

client/js/lounge.js
```diff
@@ -133,6 +133,14 @@ $(function() {
 					data: target.data("id"),
 				});
 			}
+			if (target.hasClass("channel") && utils.hasRoleInChannel($(`#chan-${target.data("id")}`))) {
+				output += templates.contextmenu_item({
+					class: "list",
+					action: "banlist",
+					text: "Show ban list",
```
astorije: Could you do "List banned users"? That is a bit more explanatory to me and fits better with the "List all channels" wording we did for #1802.

package.json
```diff
@@ -22,7 +22,7 @@
     "build": "webpack",
     "watch": "webpack --watch",
     "test": "npm-run-all --aggregate-output --parallel --continue-on-error test:* lint:*",
-    "test:mocha": "mocha --colors",
+    "test:mocha": "mocha --colors --require babel-polyfill --require babel-register",
```
astorije: Argh, running into that was bound to happen as we have more tests.
We can't do exactly that because that's babeling everything including server tests, and that means we could easily ship something incompatible with Node 4 or 6 on the server...

... But! I've been working on solving this and I'm almost there: https://github.com/thelounge/lounge/pull/1726. There is one more dependency I rely on that needs releasing (the PR I opened to fix their repo is merged, they _just_ need to release) but after that I'll be all set. That means that our client tests will run with the same config than our font-end builds, while the server tests remain unaffected.
I'm pretty excited about it because it brings a few good things, and opens the door to full-feature browser testing. Also I know it is not 100% ideal yet, but we can always improve from there.

(Also if they don't release before your PR is ready, we can find a workaround to not block your PR on it!)

client/js/lounge.js
YaManicKill: Alrighty, you've convinced me, I'll enable it for everyone. 

Should I remove the changes I made to the op function, or should I put that in a different PR?

package.json
YaManicKill: Yeah, that's a valid reason. 

But ... Surely the easiest solution is just having 2 different run configurations for client and server tests?

client/js/lounge.js
YaManicKill: Makes sense, done

package.json
YaManicKill: (this is now out of this pr and in #1864)

client/js/lounge.js
YaManicKill: I have removed the changes to the op function and put them in #1864

package.json
astorije: > But ... Surely the easiest solution is just having 2 different run configurations for client and server tests?

That is pretty much what #1726 does, with test configuration running behind the (almost) same Webpack config than our client builds.

PR #1859, opened 2017-12-19
Fix possible issues with `from` in messages

xPaw:
Was looking at that code and it looked sketchy. `from` is supposed to be cloned in `Msg` constructor, but it was being assigned later on.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -38,11 +38,13 @@ module.exports = function(irc, network) {
 
 	function handleMessage(data) {
 		let chan;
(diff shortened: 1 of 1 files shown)


src/plugins/irc-events/message.js
```diff
@@ -38,19 +38,20 @@ module.exports = function(irc, network) {
 
 	function handleMessage(data) {
 		let chan;
+		const from = chan.getUser(data.nick);
```
YaManicKill: This breaks the server.

Chan doesn't exist yet.

xPaw: Fuck.

PR #1860, opened 2017-12-19
Add advanced toggle and hide certain settings by default

McInkay:
This should probably not be merged until #1851 is merged because there's some overlap with that one changing how `options.js` works.

Insert bikeshedding about which options are advanced.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -24,6 +24,7 @@ let $warningBlocked;
 // Default settings
 const settings = {
 	syncSettings: false,
+	advanced: false,
 	autocomplete: true,
 	nickPostfix: "",
 	coloredNicks: true,
@@ -155,6 +156,8 @@ function applySetting(name, value) {
 		} else if (!value) {
 			$warningBlocked.hide();
 		}
+	} else if (name === "advanced") {
+		$("#settings [data-advanced]").toggle(settings[name]);
 	}
 }
 
--- a/client/views/windows/settings.tpl
+++ b/client/views/windows/settings.tpl
@@ -5,8 +5,17 @@
 	<h1 class="title">Settings</h1>
 
 	<div class="row">
+		<div class="col-sm-6">
+			<label class="opt">
+				<input type="checkbox" name="advanced">
+				Advanced settings
+			</label>
+		</div>
+	</div>
+
+	<div class="row">
 		{{#unless public}}
-		<div class="col-sm-12">
+		<div class="col-sm-12" data-advanced>
 			<h2>
 				Settings synchronisation
 				<span class="tooltipped tooltipped-n tooltipped-no-delay" aria-label="Note: This is an experimental feature and may change in future releases.">
@@ -71,7 +80,7 @@
 				Enable autocomplete
 			</label>
 		</div>
-		<div class="col-sm-12">
+		<div class="col-sm-12" data-advanced>
 			<label class="opt">
 				<label for="nickPostfix" class="sr-only">Nick autocomplete postfix (e.g. <code>, </code>)</label>
 				<input type="text" id="nickPostfix" name="nickPostfix" class="input" placeholder="Nick autocomplete postfix (e.g. ', ')">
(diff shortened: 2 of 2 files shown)


creesch:
> Insert bikeshedding about which options are advanced.

Any option that sort of can "mess up"  the client imho. So css and probably custom highlights at the moment and possibly the sync feature in the future. 

The extreme other end would be anything besides themes and notifications on/off if you are going for the clean look :P

YaManicKill:
> custom highlights

Hmmm, I agree with the weird "highlight for every message" setting, but custom highlights? Hmmm, maybe so.

creesch:
Just going with the question "can users mess this up in a way the confuses them and makes using the lounge highly annoying"?"  to which the answer is yes :) People only need to input one character for them to get pinged on basically 80% of the messages. 

Basically "does this setting require some forethought or doesn't it really matter if it is changed"

YaManicKill:
Alright, you've convinced me, @creesch . I have added that to advanced. I have also (on this note) added the autocomplete postfix setting to advanced as well.

client/views/windows/settings.tpl
```diff
@@ -4,6 +4,15 @@
 <div class="container">
 	<h1 class="title">Settings</h1>
 
+	<div class="row">
+		<div class="col-sm-6">
+			<label class="opt">
+				<input type="checkbox" name="advanced">
+				Show advanced settings
```
astorije: "Advanced settings" only maybe?
Also maybe display this at the end? If it's advanced and not useful to everyone, I'm thinking it doesn't need to be the first thing everyone sees. But not married to that idea either.

client/js/options.js
```diff
@@ -172,6 +173,8 @@ module.exports.initialize = () => {
 			if ($(this).prop("checked") && Notification.permission !== "granted") {
 				Notification.requestPermission(updateDesktopNotificationStatus);
 			}
+		} else if (name === "advanced") {
+			$(".container").find("[data-advanced]").toggle(options[name]);
```
astorije: Isn't that something we'd want to show/hide with CSS instead?

McInkay:
> I think I'd rather have settings nicely organized over an advanced checkbox, but that's something we can revisit when such reorganization actually happens.

There will always be settings that we want but that we don't want to surface for everyone. For example, custom css is a great setting, but many people can end up completely destroying their app. 

The other option as opposed to having advanced settings hidden is having them in a separate section, which I don't like. I like the concept of "Here are the settings, and advanced isn't a section, but it's a level of involvement in the project". It's how firefox works (or at least used to, no idea if it still does). 

But yeah, we can discuss this when the reorganisation happens, obviously.

client/js/options.js
McInkay: Well, we have to used javascript somewhere, and we either set a class on the window, and have the css work with that, or we do this. I'm not massively bothered as to which, but this involves less code/css total.

client/views/windows/settings.tpl
McInkay: > "Advanced settings" only maybe?

Sure, will fix

> Also maybe display this at the end? If it's advanced and not useful to everyone, I'm thinking it doesn't need to be the first thing everyone sees. But not married to that idea either.

Hmmm, can do. But should it be behind the sessions and the password reset? I was thinking this position is more of a "over all the settings" which can stay when the settings are turned into tabs.

astorije:
This is a longer-term discussion so I already approved your PR, but reason behind my comment is, I was looking at other apps/systems, related or totally unrelated, and realized for example that Telegram or macOS/iOS (I know, wide spectrum) don't have this. In fact, it's a paradigm we see less and less I think.

I agree with you re: advanced settings section, though it can be useful if the only settings you want to hide are _very much_ advanced, like app in debug/safe mode, etc., in short things for your devs/contributors, i.e. never or extremely rarely used by regular-to-advanced users.
Because the everlasting issue we'll have is that we'll often have people thinking this or that setting is/is not advanced. It's a very subjective choice, and impossible to predict without actually enabling this mode, which isn't great.

When it comes to destructive actions (we don't have any _at the moment_, even custom CSS can be disabled with `?nocss`), I'd usually prefer being able to undo, or confirm before acting.

So anyway, just like everything, we'll improve over time, and when we split things by section, we might end up with 1-5 settings per section, in which case it will be a good thing to re-think about.

McInkay:
I'm definitely happy to discuss if it ends up feeling silly when we split up the settings page.

client/js/options.js
xPaw: Change this line to `$("#settings [data-advanced]").toggle(options[name]);`

McInkay: Done

McInkay:
This should be ready to go if you are both still happy. I changed the name of the setting.

xPaw:
:shipit:

PR #1864, opened 2017-12-20
Change isOpInChannel to allow multiple different user roles

YaManicKill:


YaManicKill:
@creesch The place where you use this function, I feel like it should check against half-op and above instead of just op. Thoughts?

creesch:
Yeah probably, didn't do so at first because it also had the ban action.

client/js/constants.js
```diff
@@ -1,4 +1,5 @@
 "use strict";
+const modes = require("./libs/handlebars/modes");
```
xPaw: Can this be done with `import { modes } from ...`?

client/js/constants.js
```diff
@@ -105,11 +106,16 @@ const timeFormats = {
 	msgWithSeconds: "HH:mm:ss",
 };
 
+const allPriviledgeModes = Object.values(modes.modes);
+allPriviledgeModes.splice(-1, 1);
```
xPaw: What's this supposed to do? Remove `normal`?

Add a comment please.

client/js/constants.js
```diff
@@ -105,11 +106,16 @@ const timeFormats = {
...
+
 module.exports = {
+	actionTypes: actionTypes,
```
xPaw: bad rebase?

package.json
```diff
@@ -22,7 +22,8 @@
     "build": "webpack",
     "watch": "webpack --watch",
     "test": "npm-run-all --aggregate-output --parallel --continue-on-error test:* lint:*",
-    "test:mocha": "mocha --colors",
+    "test:client": "mocha --colors --require babel-polyfill --require babel-register test/client",
+    "test:server": "mocha --colors test/src",
```
xPaw: also bad rebase

package.json
```diff
@@ -63,7 +64,9 @@
   "devDependencies": {
     "babel-core": "6.26.0",
     "babel-loader": "7.1.2",
+    "babel-polyfill": "6.26.0",
```
xPaw: What are you polyfilling?

client/js/constants.js
McInkay: Possibly, but do we allow es6 imports just now?

client/js/constants.js
McInkay: Oh goodness, that's a good question. Yeah, I think it was to remove normal, because normal isn't a priviledged mode.

package.json
McInkay: This is in the commit I'm removing anyway

package.json
McInkay: In the commit I'm removing. Superceded by #1726

McInkay:
I have removed the second commit from this PR as it has been superseded by #1726

client/js/constants.js
xPaw: I think we do.

astorije: I believe we can since #1726, except in files that are shared between the client and the server :(

astorije:
Womp womp

```
TypeError: Object.values is not a function
    at Object.<anonymous> (/home/travis/build/thelounge/lounge/.tmp/mocha-webpack/1519216387870/webpack:/client/js/constants.js:16:33)
```

McInkay:
> Womp womp

Yeah, I know. The weird thing is ... I thought your changes from #1726 put them all through babel? So this should have done that as it's client code.

McInkay:
I have removed the default roles from the function as ... well, we don't use them and we got into a big discussion about what was the best way to write code that we don't even use. So I deleted it.

astorije:
Re: `Object.values`, I believe Babel only statically rewrites syntax and does not do runtime transforms (such as that, or adding `Promise`/`Map`/etc. support if it's not there). There are a bunch of reasons why IIRC.
Also a bunch of Babel plugins to do this, but we don't have any because we haven't needed any thus far.

McInkay:
It's fine, I'm not using Object.values anymore now that I've deleted the other code :-P

PR #1867, opened 2017-12-21
Add support for collapsing networks

### LINKED ISSUE #867: collapsible channel list
I find it difficult to navigate when connected to multiple servers and channels. Perhaps collapsing the channel list into a server header would ease that.

> 13:25	realies	anyone up for collapsible server lists?
13:33	MiniDigger	+1
13:33	MiniDigger	make an issue realies
13:34	MiniDigger	when collapsed it should still show new messages and pings, but only on the server label
13:39	realies	MiniDigger, I was thinking that instead of messages amount to be displaying the amount of channels/private messages you've received since the last check

McInkay:
Fixes #867

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -717,6 +717,11 @@ kbd {
 	transition: opacity 0.2s, background-color 0.2s, transform 0.2s;
 }
 
+button.collapse-network:first-child:nth-last-child(3) {
+	/* Hide collapse button if there are no channels/queries */
+	display: none;
+}
+
 #sidebar .network.collapsed .collapse-network {
 	transform: rotate(-90deg);
 }
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -564,7 +564,7 @@ kbd {
 #sidebar .chan,
 #sidebar .chan-placeholder {
 	display: block;
-	padding: 8px 25px;
+	padding: 8px 16px 8px 32px;
 	transition: color 0.2s;
 	width: 100%;
 }
@@ -710,24 +710,20 @@ kbd {
 
 #sidebar .network .collapse-network {
 	border-radius: 3px;
-	width: 25px;
+	width: 32px;
 	height: 34px;
 	float: left;
 	opacity: 0.4;
 	transition: opacity 0.2s, background-color 0.2s, transform 0.2s;
 }
 
-#sidebar .network .collapse-network::before {
-	font-weight: normal;
-	display: inline-block;
-	line-height: 16px;
-	text-align: center;
-	content: "\f0d7"; /* http://fontawesome.io/icon/caret-down/ */
-	color: #fff;
+#sidebar .network.collapsed .collapse-network {
+	transform: rotate(-90deg);
 }
 
-#sidebar .network.collapsed .collapse-network::before {
-	content: "\f0da"; /* http://fontawesome.io/icon/caret-right/ */
+#sidebar .network .collapse-network::before {
+	content: "\f0d7"; /* http://fontawesome.io/icon/caret-down/ */
+	color: #fff;
 }
 
 #sidebar .collapse-network:hover {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -248,7 +248,8 @@ kbd {
 #help .report-issue-link::before,
 #nick button::before,
 #image-viewer .previous-image-btn::before,
-#image-viewer .next-image-btn::before {
+#image-viewer .next-image-btn::before,
+#sidebar .collapse-network::before {
 	font: normal normal normal 14px/1 FontAwesome;
 	font-size: inherit; /* Can't have font-size inherit on line above, so need to override */
 	-webkit-font-smoothing: antialiased;
@@ -523,6 +524,10 @@ kbd {
 	font-size: 14px;
 }
 
+#sidebar .network.collapsed .chan:not(.lobby) {
+	display: none;
+}
+
 #sidebar button,
 #sidebar .chan,
 #sidebar .sign-out {
@@ -701,6 +706,32 @@ kbd {
 #sidebar .lobby .add-channel.opened {
 	/* translateZ(0) enables hardware acceleration, this is to avoid jittering when animating */
 	transform: rotate(45deg) translateZ(0);
+}
+
+#sidebar .network .collapse-network {
+	border-radius: 3px;
+	width: 25px;
+	height: 34px;
+	float: left;
+	opacity: 0.4;
+	transition: opacity 0.2s, background-color 0.2s, transform 0.2s;
+}
+
+#sidebar .network .collapse-network::before {
+	font-weight: normal;
+	display: inline-block;
+	line-height: 16px;
+	text-align: center;
+	content: "\f0d7"; /* http://fontawesome.io/icon/caret-down/ */
+	color: #fff;
+}
+
+#sidebar .network.collapsed .collapse-network::before {
+	content: "\f0da"; /* http://fontawesome.io/icon/caret-right/ */
+}
+
+#sidebar .collapse-network:hover {
+	opacity: 1;
 }
 
 #sidebar .chan.active .close {
--- a/client/views/network.tpl
+++ b/client/views/network.tpl
@@ -6,6 +6,10 @@
 	data-nick="{{nick}}"
 	data-options="{{tojson serverOptions}}"
 >
+	<button class="collapse-network" aria-label="Collapse" data-id="{{id}}"
+		aria-controls="network-{{id}}-chanlist" aria-expanded="true"></button>
+	<div id="network-{{id}}-chanlist" role="region" class="chanlist">
 	{{> chan}}
+	</div>
 </section>
 {{/each}}
--- a/client/views/network.tpl
+++ b/client/views/network.tpl
@@ -6,10 +6,11 @@
 	data-nick="{{nick}}"
 	data-options="{{tojson serverOptions}}"
 >
-	<button class="collapse-network" aria-label="Collapse" data-id="{{id}}"
-		aria-controls="network-{{id}}-chanlist" aria-expanded="true"></button>
 	<div id="network-{{id}}-chanlist" role="region" class="chanlist">
-	{{> chan}}
+		<button class="collapse-network" aria-label="Collapse" data-id="{{id}}"
+			aria-controls="network-{{id}}-chanlist" aria-expanded="true">
+		</button>
+		{{> chan}}
 	</div>
 </section>
 {{/each}}
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -24,6 +24,7 @@ const historyObserver = window.IntersectionObserver ?
 module.exports = {
 	appendMessage,
 	buildChannelMessages,
+	collapseNetwork,
 	renderChannel,
 	renderChannelUsers,
 	renderNetworks,
@@ -284,4 +285,18 @@ function loadMoreHistory(entries) {
 
 		target.trigger("click");
 	});
+}
+
+sidebar.on("click", ".collapse-network", (event) => collapseNetwork($(event.target)));
+
+function collapseNetwork(collapseButton) {
+	$("#sidebar").find(`.network[data-id='${collapseButton.attr("data-id")}']`).toggleClass("collapsed");
+
+	if (collapseButton.attr("aria-expanded") === "true") {
+		collapseButton.attr("aria-expanded", false);
+		collapseButton.attr("aria-label", "Expand");
+	} else {
+		collapseButton.attr("aria-expanded", true);
+		collapseButton.attr("aria-label", "Collapse");
+	}
 }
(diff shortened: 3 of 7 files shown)


astorije:
Very nice, glad you're taking up on that!
The one thing (I promise I won't bikeshed 🙂) about this is that it's rather unusual to trigger an expand/collapse via context menu. In fact, when it's collapsed, there is currently no way to know if a network is collapsed or without a channel.

That said...

> I didn't add a toggle on the actual network because I couldn't make it look nice, so here's a first version.

Yes, I was thinking the same thing. Between the <kbd>&times;</kbd> on lobby being removed, the icons being recently changed, the <kbd>+</kbd> button being added to join a channel, there has been a lot of activity in there.
But things are settling down now, and things are pretty nice, so I thing relying on the commonly used 🔽/▶️ carets would work well. In fact, we're already using that concept on image previews and condensed messages, so that would fit in IMO.

I didn't want to arrive without a suggestion, so I went ahead and made a mockup. What do you think?

Current | Proposed
--- | ---
screen shot 2017-12-22 at 14 11 30 | screen shot 2017-12-22 at 14 05 54

It's not functional obviously, just playing with the UI. Also, don't worry about spacing, that can be tweaked (I'm mocking with `&nbsp;`s lol), but the nice thing is that on this specific screenshot, the network name line is only losing a couple pixels width!
Let me know what you think and I can help :)
(Also, that would reduce your amount of rebase on #1878 :trollface:)

> There are obvious improvements that can be made to this

Lots yes, but a PR with simple expand/collapse like you did is a perfect start.

YaManicKill:
@astorije Did you mock that up in html/css or with an image editor? My issue is trying to get it to look nice actually implementing it (without rearchitecting a whole bunch of how the sidebar currently works)

The problem is that the server icon is in a ::before, which we can't put something inside the chan label **and** before the server icon. We'll have to either:

1. Change how the server icon is shown (either not using a before, or doing a before on the name)
2. Put the collapse icon outside of the label, and just line it up nicely with some css
The problem with this one is that a lot of the things in the sidebar are defined using :first-child which will all need to be changed. This does work, but I'm sure there are things that I've missed in this.
3. Replace the server icon with the carat
Sadly this one also needs a bit of changing how the icon is shown because you can't bind clicks onto pseudo elements, but it would be a lot easier than 1.

I have currently done 2 on my machine, and I **think** it works. I'll update the PR later today when I've done some more tests on it. There's a bit more styling I need to finish off with it.

YaManicKill:
Aight, I've added a new commit (will squash if people are happy with this before merging, obviously) which changes it to have the carat.

Happy to discuss any css changes. It works now. I think I've changed all of the first-child instances that broke with this (they should have just been using .lobby anyway, realistically)

client/css/style.css
```diff
@@ -280,6 +281,10 @@ kbd {
 #sidebar .chan.special::before,
 #chat .special .title::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
 
+.chan::before {
+	padding-top: 3px;
```
YaManicKill: Note, this is here because I noticed the icons for the server and channels were actually slightly misaligned.

|Before|After|
|-|-|
|||

dgw: Hmm, to me the server icon looks misaligned in the before image, but the channel icon looks misaligned in the after image. Good eye though. I didn't notice the alignment at all before.

YaManicKill: Yeah, you might be right. I'll change it to just do the lobby and see how that looks.

YaManicKill: Alright, just pushing up a change to only do this for the lobby icon.

astorije:
> @astorije Did you mock that up in html/css or with an image editor?

HTML and CSS, but not in a way we can use here (i.e. used the `<i>` elements to mock). Also I didn't keep the code for that very reason, I believe lol. But I figured that if you were okay with this, you'd tell me to do it myself, which I would have happily helped with :)

> or doing a before on the name

I was thinking about that, but I'll admit I didn't look too much.

Anyway, your current result looks good IMO. We can surely tweak the alignments to go pixel-perfect, but besides that, it seems fine. I haven't looked at your changes yet, I'll do that when we start merging v3 PRs. 👏

McInkay:
Rebased and fixed conflicts

xPaw:
Tooltip is misaligned:



Arrow target is actually rather small (especially for tapping on a phone), perhaps this can be duplicated in dropdown menu?

McInkay:
> Arrow target is actually rather small (especially for tapping on a phone), perhaps this can be duplicated in dropdown menu?

I originally just had that. I can add it back in, sure.

MaxLeiter:
How can you access drop down from mobile? Not sure it’s a great solution

astorije:
I agree with @MaxLeiter that it doesn't help on mobile. It would also be a weird entry to have in the context menu, as it would be pretty unique to us.

I think the issue is that the hit box is only on the icon itself, and the left/right margins around the icon don't help. If I re-use my mock screenshot above, I think the hit box should look like this:

34310202-16fa44a6-e724-11e7-8a44-d0b598ee175d

At that point, it would be as big as the <kbd>+</kbd> for joining a new channel, if I'm not mistaken.

@McInkay, if you think that's a valid approach, do you want me to help with alignments and stuff? I could add a commit to your PR and see how it goes. Up to you :)

McInkay:
I'll play around with it this morning and see if I can get it working nicely or not.

As a sidenote, part of the issue is probably just that our sidebar is really small, but maybe that's just the density on my phone, I dunno. When I compare it to other apps, it seems quite small on a phone.

dgw:
The Lounge's sidebar does seem small on Android. A quick informal survey to compare its width to the sidebars in native apps (used several of Google's just because, in theory, they follow the platform design guidelines) showed that TL uses 55-60% of the screen width at most for its sidebar, while native apps use more like 80-85% of the available width.

McInkay:
It's not just the width, it's that each entry in the sidebar isn't very tall, leading to issues with clicking on them, etc.

xPaw:
@MaxLeiter holding your finger over a channel shows the context menu.

astorije:
@xPaw, ... except on iOS 😂

Re: sidebar, I agree with @McInkay on mobile. Entries are usually twice as tall when I look at email/SMS apps, Telegram, etc.

I'm hoping to improve the channel list for v3, I'll experiment with this, making it more pleasant to use on mobile without affecting density on desktop. Right now we just show exactly the same thing on mobile and desktop, just with different total width, and we can do something a bit more bespoke and adapted to the different experiences :)

astorije:
But for this PR, @McInkay, I wouldn't worry too much. Just make the button clickable that includes the margins and not just the icon, and we'll be fine, we can improve over time.

McInkay:
Alrighty, I've changed the hitbox to be the outer span, and changed it so it's the whole width between the edge and the name, and the same height as the name.



Also, fixed the tooltip alignment.

McInkay:
> Also wondering, should open state be synced to the server (multi client, server restarts, page reloads)?

Oh I completely missed this comment earlier. Shall we put this PR in first, and we can add more later? @xPaw

xPaw:
@McInkay Sure.

McInkay:
Cool, then I think this PR is ready unless someone finds an issue with it. There are lots of things to improve, but this is a good start.

1. Keep state on client for refreshes
2. Possibly sync it to server for multiple devices (this is a more complicated one and I'm not sure if we definitely want it to happen or not, but we can discuss that elsewhere)
3. Make the sidebar more touch-friendly

I'll make tickets for 1 and 2 once this is merged and we can discuss there. Do we have a ticket for 3 already or does that need made? We've discussed it a little bit but haven't seen a ticket number so not sure if it exists or not yet.

client/views/network.tpl
```diff
@@ -6,6 +6,12 @@
 	data-nick="{{nick}}"
 	data-options="{{tojson serverOptions}}"
 >
+	<span class="collapse-network-tooltip tooltipped tooltipped-e tooltipped-no-touch" aria-label="Collapse network…" data-alt-label="Cancel">
+		<button class="collapse-network" aria-label="Collapse network…" data-id="{{id}}"
```
astorije: Remove `…` from `Collapse network…`. Ellipsis in menus often means that there are further options or required input from the user. I'm sure there are better reads than this one, but for example: https://ux.stackexchange.com/a/9545/90032

We've been doing better at this lately, that wasn't the case before. For example, we correctly have "Join a channel…", "Change nick…" vs. "Send message", etc.

astorije: Also, did you mean "Expand" instead of "Cancel"? I'd actually vote for "Collapse" instead of "Collapse network" but no big deal. Or even no tooltip actually, we don't have one for other collapse/expand buttons (previews, status messages, and in a further extent, "More details" on error loading page).

(Kinda odd that in "alt mode", tooltip `aria-label` will be "Expand" and button `aria-label` will be "Collapse". I know it's not your PR, just an observation, but we ought to fix this someday 😅)

McInkay: Yeah that's... Copy paste error. Good catch, will fix it. Would need to keep the aria but happy to remove the tooltip if you don't want it.

astorije: Yeah I'd say remove it, but if there are strong opinions against that, then I don't mind.

McInkay: Alright, I removed it. Also, updating the aria-label when we expand/collapse now.

astorije:
Hey @McInkay, I had a few nitpicky comments about animation and UI tweaks, so I took the liberty to do that in an extra commit rather than bother you with endless back-and-forth on pixel perfectionism.
I hope you're not annoyed that I did that, and feel free to drop my commit if you don't like it.
What it does:

- Animate the caret similarly to previews and status messages
- Tweak left/right margins to kinda center things (namely the caret and + sign are sort of equal distance from the sides). This does not reduce the available width for channel names.
- Remove unused CSS attributes

Result:

Before | After
--- | ---
screen shot 2018-03-01 at 23 06 57 | screen shot 2018-03-01 at 23 07 06

You'll notice that I wasn't looking for perfection, only improve this a bit. I think the channel list deserves a nice re-design rather than minor tweaks that compromise more than improve, but that's way beyond this.

Also, any chance you could hide the caret when there are no channels/queries? If that's too much of a pain, I'm totally fine if we do this in a further PR, it's an edge case and this PR is pretty much ready to go as it is.

McInkay:
Yeah sure, looks lovely.

McInkay:
> Also, any chance you could hide the caret when there are no channels/queries? If that's too much of a pain, I'm totally fine if we do this in a further PR, it's an edge case and this PR is pretty much ready to go as it is.

I'll take a quick look today to see how difficult it would be, if I don't get it today then we can ship this without it.

McInkay:
Alright, done. I had to move the button into the network div in the sidebar, because with css we can only check number of siblings, not number of children or sibling's children. 

|With channel|Without channel|
|--|--|
|||

McInkay:
I have also added a commit for storing the collapsed state in localstorage.

McInkay:
Removed storing the collapsed state in localstorage as it was buggy, will do a new PR for it when this is merged

client/css/style.css
```diff
@@ -703,6 +708,33 @@ kbd {
...
+#sidebar .network .collapse-network {
+	border-radius: 3px;
+	width: 32px;
+	height: 34px;
+	float: left;
+	opacity: 0.4;
+	transition: opacity 0.2s, background-color 0.2s, transform 0.2s;
+}
+
+button.collapse-network:first-child:nth-last-child(3) {
```
astorije: I can't tell if this is a total hack or brilliant :D Either way, probably the first time I see `:nth-last-child` actually useful so 👏, love it.

client/views/network.tpl
```diff
@@ -6,6 +6,11 @@
 	data-nick="{{nick}}"
 	data-options="{{tojson serverOptions}}"
 >
-	{{> chan}}
+	<div id="network-{{id}}-chanlist" role="region" class="chanlist">
+		<button class="collapse-network" aria-label="Collapse" data-id="{{id}}"
+			aria-controls="network-{{id}}-chanlist" aria-expanded="true">
+		</button>
+		{{> chan}}
+	</div>
```
astorije: (Just fixed the indentation)

client/css/style.css
McInkay: (It's brilliant)

McInkay:
🚢 🎉

realies:
Amazing stuff! Although @xPaw did not see it needed, I think that sliding up/down the channel list at the same speed as the arrow rotation speed would make it look really sleek. Had a few ugly attempts at transforming `scaleY` for `.chan:not(.lobby)`, but think this would be possible if all channels have their own surrounding container. Thoughts?

MiniDigger:
> Removed storing the collapsed state in localstorage as it was buggy, will do a new PR for it when this is merged  

now that this got merged, do you think y o u could give that another try @McInkay

McInkay:
Yeah, I've got the base code for it, will see if I can fix the issue I was having with it over the weekend.

PR #1870, opened 2017-12-21
Only show in active if the document is active

McInkay:
Something I was thinking about when doing #1868 

If the user isn't actively looking at the lounge, we probably don't want to show something in the active window. Just a thought.


dgw:
I'd be against this. If The Lounge is in a background tab and something comes in (like a notice) that would go to the server window, but I left a channel open, I won't necessarily know to check the server window and might miss it.

If things are being sent to the active window (and I'm very excited for that), they should still go there even if the client is open but not focused.

MaxLeiter:
As @astorije and @xPaw both 👍'd the parent comment, I think this can be closed, unless @McInkay has strong feelings about it?

PR #1871, opened 2017-12-21
Add visual feedback on focused buttons, for example when tabbing to it

astorije:
This is nice for keyboard navigation (and accessibility).

Before | After
--- | ---
before | after

Note that this example is not the best because users can (and probably do) hit <kbd>Enter</kbd> to submit, but buttons are not necessarily for submitting forms, in which case it looks like the focus is lost.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -124,7 +124,8 @@ kbd {
 }
 
 .btn:disabled,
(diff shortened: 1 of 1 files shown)


astorije:
Issue to improve all messed up `outline: none` at #1872

Issue #1872, opened 2017-12-21
Fix all uses of `outline: none`


astorije:
At the moment, buttons have no visual cue on focus (there will be one after #1871 but not perfect), inputs have one but not very visible and same as hovering, nick buttons do not have any feedback cue, ...

All this makes keyboard tabbing very hard. This issue for me to remember to fix them.

xPaw:
https://caniuse.com/#feat=css-focus-visible

`:focus-visible` is supported in Firefox and Chrome now.

PR #1873, opened 2017-12-22
Add custom focus outlines for inputs and green buttons

astorije:
Ref https://github.com/thelounge/lounge/issues/1872

We were automatically silencing outlines for all buttons and inputs, even though we really meant `.btn`s and `.input`s where we had some custom UI on hover only. I have added an outline on focus on those specifically, not on everything.

This is clearly not perfect, but I think it's much better. I prioritized on things that get focus naturally, like form inputs, but did not apply this to all keyboard-navigable elements (like font-icon buttons). The most important aspect of it is that now we do not remove outlines by default on elements, so tabbing through font-icon buttons, if not super pretty, is at least visible.

⚠️ It looks ugly on themes at the moment, test on `Example`. If that seems okay to you, I'll apply the changes to the themes.

screen shot 2017-12-21 at 19 12 53
screen shot 2017-12-21 at 19 13 03
screen shot 2017-12-21 at 19 13 10

### TODO

- [ ] Built-in themes
- [ ] Icon buttons (any ideas?)

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -100,10 +100,6 @@ h3 {
 	margin: 0;
 }
 
-input {
-	outline: 0;
-}
-
 button {
 	border: none;
 	background: none;
@@ -172,7 +168,7 @@ kbd {
 	margin-bottom: 10px;
 	padding: 9px 17px;
 	text-transform: uppercase;
-	transition: background 0.2s, border-color 0.2s, color 0.2s;
+	transition: background 0.2s, border-color 0.2s, color 0.2s, box-shadow 0.2s;
 	word-spacing: 3px;
 	cursor: pointer; /* This is useful for `<button>` elements */
 }
@@ -189,8 +185,14 @@ kbd {
 	opacity: 1;
 }
 
+.input:focus,
+.btn:active,
+.btn:focus {
+	outline: 0;
+	box-shadow: 0 0 0 3px rgba(132, 206, 136, 0.5);
+}
+
 .btn:active {
-	box-shadow: none;
 	opacity: 0.8;
 }
 
@@ -877,9 +879,8 @@ background on hover (unless active) */
 	font-size: 14px;
 	margin: 2px 0;
 	margin-bottom: 10px;
-	outline: 0;
 	padding: 0 10px;
-	transition: border-color 0.2s;
+	transition: border-color 0.2s, box-shadow 0.2s;
 	width: 100%;
 	height: 35px;
 	line-height: 35px;
@@ -891,7 +892,7 @@ background on hover (unless active) */
 
 .input:not(:disabled):hover,
 .input:not(:disabled):focus {
-	border-color: #79838c;
+	border-color: #84ce88;
 }
 
 #user-specified-css-input {
@@ -2103,7 +2104,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 #chat .userlist .user.active {
 	background-color: #f6f6f6;
 	transition: none;
-	outline: 0; /* TODO: Handle focus outlines in PR #1873 */
+	outline: 0;
 }
 
 .context-menu-item::before,


client/css/style.css
```diff
@@ -1572,9 +1583,6 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 #help .help-item {
 	display: table-row;
-}
-
-#help .help-item {
```
astorije: This was flagged by my stylelint linter on Atom as duplicate entry. Not sure why my editor reports different things than our actual stylelint configuration...

client/css/style.css
```diff
@@ -149,8 +144,22 @@ kbd {
 	opacity: 1;
 }
 
+input[type="checkbox"],
+input[type="radio"] {
+	transition: box-shadow 0.2s;
+}
+
+#chat .search:focus,
+.input:focus,
```
astorije: I removed the `#window` part of this, because "Join a channel" form is not part of it and therefore was not being picked up.

astorije:
For comparison, same screenshots as mine above but with @xPaw's changes:

screen shot 2017-12-25 at 19 09 37
screen shot 2017-12-25 at 19 09 44
screen shot 2017-12-25 at 19 09 56

I don't have a preference, anything works better than what we currently have.

astorije:
@xPaw, I think we should leave the custom outlines away from radios and checkboxes. It doesn't look too great, at least on Chrome and Firefox, and I'm afraid it's only going to be worse for other browsers:

Chrome | Firefox
--- | ---
screen shot 2017-12-25 at 19 03 28 screen shot 2017-12-25 at 19 03 48 | screen shot 2017-12-25 at 19 04 10 screen shot 2017-12-25 at 19 04 17

On Chrome, checkboxes have a bottom padding that visually adds a white line, and lack of border-radius looks odd on both types.
On Firefox, the default outlines are kept on top of the custom ones (wtf Firefox) and only for those 2 types (i.e. not inputs/buttons, as far as I can tell), it shows kind of a messed up gradient instead of plain box-shadow. So weird.

xPaw:
@astorije remove it from checkbox/radio then. We can tackle it later on by using custom styles of these.

astorije:
@xPaw, done.

client/css/style.css
```diff
@@ -769,12 +771,20 @@ kbd {
...
+	transition: border-color 0.2s, box-shadow 0.2s;
 	width: 100%;
 }
 
+#chat .search:focus,
+.input:focus {
+	border-color: #84ce88;
+}
+
+#windows .input:hover {
```
xPaw: You can combine this with the one above.

astorije: I'll also remove the `#window` in front of it.

astorije: Done.

astorije:
It's pretty gross on icon buttons right now (especially the ones in the sidebar, since it's often used and there are animations). Any ideas on what we could do?

astorije:
I know Slack has made efforts towards tab navigation, and frankly their current solution is really well done:

slack_outline

The outlines are big and visible, but what's good about them is that you never see them unless you actually navigate with your keyboard. That means no visual noise unless this is the expectation.

The current state of this PR does not do this, i.e. outlines on button icons show up even when clicking around (as opposed to using keyboard). That means we get that kind of ugliness here and there:

footer_outline join_channel_outline

Slack does this by having a special class that they add/remove on keyboard tab, while having an `outline: 0` on everything. I don't know how hard it would be to do that ourselves, but it seems like a workaround to the fact that there is no elegant solution to applying `:focus` on keyboard tabbing only. I don't know if we want to go the Slack route, or [whatever that is](https://stackoverflow.com/a/45191208/1935861), or anything else I did not think of.
Maybe for icon buttons, `:focus` = no outline + same style than `:hover`?

McInkay:
Are we pushing this from 3.0.0? Or will you have time to look at it?

astorije:
As discussed on IRC, pushing to 3.1 for now.

xPaw:
@astorije We still want to get this all fixed up, right?

xPaw:
Rebased this PR, and reverted the changes on user search input (looks broken) and context menus (slightly unnecessary as it already changes color).

The issues described in https://github.com/thelounge/thelounge/pull/1873#issuecomment-366868695 about clicking and focus ring showing up are still a concern.

PR #1874, opened 2017-12-22
Add a `notEqual` block helper for Handlebars and tests for `equal`

astorije:
It turns out, we already had a case of `{{equal a b}}{{else}}Something{{/equal}}`.

This will be useful for #1833, which has something similar.

### DIFF
--- a/client/js/libs/handlebars/notEqual.js
+++ b/client/js/libs/handlebars/notEqual.js
@@ -1,0 +1,18 @@
+"use strict";
+
+module.exports = function(a, b, opt) {
+	if (arguments.length !== 3) {
+		throw new Error("Handlebars helper `notEqual` expects 3 arguments");
+	}
+
+	a = a.toString();
+	b = b.toString();
+
+	if (a !== b) {
+		return opt.fn(this);
+	}
+
+	if (opt.inverse(this) !== "") {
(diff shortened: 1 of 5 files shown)


client/js/libs/handlebars/notEqual.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
+
+module.exports = function(a, b, opt) {
+	if (arguments.length !== 3) {
+		throw new Error("Handlebars helper `equal` expects 3 arguments");
```
PolarizedIons: \`equal\` -> \`notEqual\`

astorije: Aaah damn, good eye, thanks! 🙏 
Fixed.

test/client/js/libs/handlebars/equalTest.js
```diff
@@ -0,0 +1,26 @@
...
+const equal = require("../../../../../client/js/libs/handlebars/equal");
+
+describe("equal Handlebars helper", function() {
+	const block = {
+		fn: () => "fn",
+		inverse: () => "inverse",
+	};
+
+	it("should render the first block if both values are equal", function() {
+		expect(equal("foo", "foo", block)).to.equal("fn");
```
YaManicKill: This is a really confusing sentence to read, but I get it :-P

PR #1875, opened 2017-12-22
Fix incorrect documentation url in default config

MiniDigger:
Just something that I found while updating my instance


YaManicKill:
Minor fix, merging.

astorije:
Good catch, will make sure new docs take that into account, thanks!

PR #1878, opened 2017-12-22
Pull context menu code out of lounge.js and make it more generic

McInkay:
This gives us a few advantages:

1. Removes some more code from lounge.js (it was getting big again)
2. Keeps all the context menu code together (separately, we had the functions and the element addition in very different parts of the code, and that was odd
3. Allows any part of the lounge codebase to add elements to the context menu
4. Beginning of support for plugins (that'll come in another PR once this is merged)

### DIFF
--- a/client/js/contextMenu.js
+++ b/client/js/contextMenu.js
@@ -1,0 +1,85 @@
+"use strict";
+const $ = require("jquery");
+const templates = require("../views");
+let contextMenu, contextMenuContainer;
+
+module.exports = class ContextMenu {
+	constructor(contextMenuItems, contextMenuActions, selectedElement, event) {
+		this.contextMenuItems = contextMenuItems;
+		this.contextMenuActions = contextMenuActions;
+		this.selectedElement = selectedElement;
+		this.event = event;
+
+		contextMenuContainer = $("#context-menu-container");
+		contextMenu = $("#context-menu");
+	}
+
+	show() {
+		showContextMenu(this.contextMenuItems, this.selectedElement, this.event);
+		this.bindEvents();
+		return false;
+	}
+
+	bindEvents() {
+		const contextMenuActions = this.contextMenuActions;
+
+		contextMenuActions.execute = (id, ...args) => contextMenuActions[id] && contextMenuActions[id](...args);
+
+		contextMenu.find(".context-menu-item").on("click", function() {
+			const $this = $(this);
+			const itemData = $this.data("data");
+			const contextAction = $this.data("action");
+			contextMenuActions.execute(contextAction, itemData);
+		});
+	}
+};
+
+function showContextMenu(contextMenuItems, selectedElement, event) {
+	const target = $(event.currentTarget);
+	let output = "";
+
+	for (const item of contextMenuItems) {
+		if (item.check(target)) {
+			if (item.divider) {
+				output += templates.contextmenu_divider();
+			} else {
+				output += templates.contextmenu_item({
+					class: typeof item.className === "function" ? item.className(target) : item.className,
+					action: item.actionId,
+					text: typeof item.displayName === "function" ? item.displayName(target) : item.displayName,
+					data: typeof item.data === "function" ? item.data(target) : item.data,
+				});
+			}
+		}
+	}
+
+	contextMenuContainer.show();
+	contextMenu
+		.html(output)
+		.css(positionContextMenu(selectedElement, event));
+}
+
+function positionContextMenu(selectedElement, e) {
+	let offset;
+	const menuWidth = contextMenu.outerWidth();
+	const menuHeight = contextMenu.outerHeight();
+
+	if (selectedElement.hasClass("menu")) {
+		offset = selectedElement.offset();
+		offset.left -= menuWidth - selectedElement.outerWidth();
+		offset.top += selectedElement.outerHeight();
+		return offset;
+	}
+
+	offset = {left: e.pageX, top: e.pageY};
+
+	if ((window.innerWidth - offset.left) < menuWidth) {
+		offset.left = window.innerWidth - menuWidth;
+	}
+
+	if ((window.innerHeight - offset.top) < menuHeight) {
+		offset.top = window.innerHeight - menuHeight;
+	}
+
+	return offset;
+}
--- a/client/js/contextMenuFactory.js
+++ b/client/js/contextMenuFactory.js
@@ -1,0 +1,213 @@
+"use strict";
+const $ = require("jquery");
+const socket = require("./socket");
+const utils = require("./utils");
+const JoinChannel = require("./join-channel");
+const ContextMenu = require("./contextMenu");
+const contextMenuActions = [];
+const contextMenuItems = [];
+
+module.exports = {
+	addContextMenuItem,
+	createContextMenu,
+};
+
+addDefaultItems();
+
+/**
+ * Used for adding context menu items. eg:
+ *
+ * addContextMenuItem({
+ * 		check: (target) => target.hasClass("user"),
+ * 		className: "customItemName",
+ * 	 	data: (target) => target.data("name"),
+ * 	 	displayName: "Do something",
+ * 	 	callback: (name) => console.log(name), // print the name of the user to console
+ * });
+ *
+ * @param opts
+ * @param {function(Object)} [opts.check] - Function to check whether item should show on the context menu, called with the target jquery element, shows if return is truthy
+ * @param {string|function(Object)} opts.className - class name for the menu item, should be prefixed for non-default menu items (if function, called with jquery element, and uses return value)
+ * @param {string|function(Object)} opts.data - data that will be sent to the callback function (if function, called with jquery element, and uses return value)
+ * @param {string|function(Object)} opts.displayName - text to display on the menu item (if function, called with jquery element, and uses return value)
+ * @param {function(Object)} opts.callback - Function to call when the context menu item is clicked, called with the data requested in opts.data
+ */
+function addContextMenuItem(opts) {
+	opts.check = opts.check || (() => true);
+	opts.actionId = contextMenuActions.push(opts.callback) - 1;
+	contextMenuItems.push(opts);
+}
+
+function addContextDivider(opts) {
+	opts.check = opts.check || (() => true);
+	opts.divider = true;
+	contextMenuItems.push(opts);
+}
+
+function createContextMenu(that, event) {
+	return new ContextMenu(contextMenuItems, contextMenuActions, that, event);
+}
+
+function addWhoisItem() {
+	function whois(itemData) {
+		const chan = utils.findCurrentNetworkChan(itemData);
+
+		if (chan.length) {
+			chan.click();
+		}
+
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/whois " + itemData,
+		});
+
+		$(`.channel.active .userlist .user[data-name="${itemData}"]`).trigger("click");
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("user"),
+		className: "user",
+		displayName: (target) => target.data("name"),
+		data: (target) => target.data("name"),
+		callback: whois,
+	});
+
+	addContextDivider({
+		check: (target) => target.hasClass("user"),
+	});
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("user"),
+		className: "action-whois",
+		displayName: "User information",
+		data: (target) => target.data("name"),
+		callback: whois,
+	});
+}
+
+function addQueryItem() {
+	function query(itemData) {
+		const chan = utils.findCurrentNetworkChan(itemData);
+
+		if (chan.length) {
+			chan.click();
+		}
+
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/query " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("user"),
+		className: "action-query",
+		displayName: "Direct messages",
+		data: (target) => target.data("name"),
+		callback: query,
+	});
+}
+
+function addKickItem() {
+	function kick(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/kick " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => utils.hasRoleInChannel(target.closest(".chan"), ["op"]) && target.closest(".chan").data("type") === "channel",
+		className: "action-kick",
+		displayName: "Kick",
+		data: (target) => target.data("name"),
+		callback: kick,
+	});
+}
+
+function addFocusItem() {
+	function focusChan(itemData) {
+		$(`.networks .chan[data-target="${itemData}"]`).click();
+	}
+
+	const getClass = (target) => {
+		if (target.hasClass("lobby")) {
+			return "network";
+		} else if (target.hasClass("query")) {
+			return "query";
+		}
+
+		return "chan";
+	};
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("chan"),
+		className: getClass,
+		displayName: (target) => target.attr("aria-label"),
+		data: (target) => target.data("target"),
+		callback: focusChan,
+	});
+
+	addContextDivider({
+		check: (target) => target.hasClass("chan"),
+	});
+}
+
+function addChannelListItem() {
+	function list(itemData) {
+		socket.emit("input", {
+			target: itemData,
+			text: "/list",
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("lobby"),
+		className: "list",
+		displayName: "List all channels",
+		data: (target) => target.data("id"),
+		callback: list,
+	});
+}
+
+function addBanListItem() {
+	function banlist(itemData) {
+		socket.emit("input", {
+			target: itemData,
+			text: "/banlist",
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("channel"),
+		className: "list",
+		displayName: "List banned users",
+		data: (target) => target.data("id"),
+		callback: banlist,
+	});
+}
+
+function addJoinItem() {
+	function openJoinForm(itemData) {
+		const network = $(`#join-channel-${itemData}`).closest(".network");
+		JoinChannel.openForm(network);
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("lobby"),
+		className: "join",
+		displayName: "Join a channel…",
+		data: (target) => target.data("id"),
+		callback: openJoinForm,
+	});
+}
+
+function addDefaultItems() {
+	addWhoisItem();
+	addQueryItem();
+	addKickItem();
+	addFocusItem();
+	addChannelListItem();
+	addBanListItem();
+	addJoinItem();
+}
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -20,7 +20,8 @@ require("./webpush");
 require("./keybinds");
 require("./clipboard");
 const Changelog = require("./socket-events/changelog");
-const JoinChannel = require("./join-channel");
+const contextMenuFactory = require("./contextMenuFactory");
+const contextMenuContainer = $("#context-menu-container");
 
 $(function() {
 	const sidebar = $("#sidebar, #footer");
@@ -29,8 +30,6 @@ $(function() {
 	$(document.body).data("app-name", document.title);
 
 	const viewport = $("#viewport");
-	const contextMenuContainer = $("#context-menu-container");
-	const contextMenu = $("#context-menu");
 
 	function storeSidebarVisibility(name, state) {
 		if ($(window).outerWidth() < utils.mobileViewportPixels) {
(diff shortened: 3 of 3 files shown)


astorije:
You like to rebase, don't you? 😁

YaManicKill:
@astorije you mean refactor, yeah?

astorije:
No I meant that there are a few open PRs that add stuff to the context menu and pretty much ready to go, so you'll have to rebase this PR when they get merged :/

astorije:
@McInkay, mind rebasing this on master? Quite a few conflicts, sorry about that :S

McInkay:
Alright, rebased. I *think* I've dealt with all the conflicts properly. There were quite a few changes to the context menu, but I tested it against what it does on master, and I went through each commit affecting the code since I last rebased, and I think I caught everything.

Let me know if you see anything wrong with it.

astorije:
Hey @McInkay, thanks for taking up on this!! 👏

I realize this is coming from the legacy all-in-lounge.js approach, but I'm not a huge fan of the current implementation however. Would you mind giving it a more component/class form?
It seems to me that the following would work nicely (but since I haven't done it nor been as thorough as you did, I may obviously be missing some things):

```js
function addDefaultItems(contextMenu) {
  /* ... */
}

function showContextMenu(contextMenu, event) {
  /* ... */
}

function positionContextMenu(contextMenu, event) {
  /* ... */
}

module.exports = class ContextMenu {
  setup() {
    /* ... */

    addDefaultItems(this);
  }
  
  addItem(opts) {
    /* ... */
  }
};
```

Then the calling code would look like:

```js
const ContextMenu = require("./ContextMenu");

const contextMenu = new ContextMenu();
contextMenu.setup();
contextMenu.addItem({
  /* ... */
});
```

In particular, this brings the advantage of not doing any side-effects when `require`ing the file. At the moment, we too often rely on things to happen when calling files with `require`, and I'd like us to avoid doing that.
One advantage is that we make this code closer to testable. I know there are jQuery stuff in there still, but it's easier to test and reason about if the file itself doesn't carry the state, but the instance of the class created does.

Another advantage I see is that in the future, we can have an instance of `ContextMenu` per actual context menu (on channel vs. on nick, on channel list vs. on channel header, etc.).
Currently we shove all items in there, then check at runtime what goes where. You've made it a bit nicer than what we have in `master` with this `check` function, so that's good, but it's still not a great design (not criticizing you, I know we had this in `master`) and it's bound to bite back soon enough. Since this is refactoring and opening the door to package API, we can probably improve this upfront.

McInkay:
Yeah, that sounds good. I'll get that done today.

McInkay:
Ah wait, no. That won't work.

Your structure works fine for when we are only adding context menu items from `contextMenu.js`, but that's not always the case.

Obviously, in future we'll want plugins to add to them (see #1972) but even just in this PR, there is 1 example that we need to add from `lounge.js` because we need something that exists there (and adding that into `contextMenu.js` would get messy).

So, we still need a function on the export to be able to add to all future context menus. We can't have context menus created empty and only add to them when the code instantiating it requests.

For the functionality we want, I don't think we can get away from having a singleton. However, we can try and make it as nice as possible, and the singleton doesn't need to be the contextMenu. We can use the factory pattern.

So, we'll have 2 new files, `contextMenu.js` and also `contextMenuFactory.js`.

`contextMenuFactory` would be the singleton, and something like this:
```
module.exports = {
	addContextMenuItem: () => {},
	createContextMenu: () => {new ContextMenu()... ... addDefaultItems(contextMenu) },
	addDefaultItems: () => {}
}
```

Then we still get a single instance of context menu for every time we create one (which is nice), we remove all of the default items out of the context menu file (which was a horrible thing when I did it, but I was being lazy), and we still remove the side effects from requiring the file, but plugins and other parts of the code can add context menu items to all future context menus as well. `contextMenu` would be pretty much what you asked for above, but without `addDefaultItems`.

It also doesn't stop us having different context menus, as we can either have different methods for creating the different context menus in the future (`createUserContextMenu`, `createChannelContextMenu` etc).

I'll implement this in a separate commit so we can revert if you want, but I think this is probably the nicest we'll get for just now.

Infact, I think I'll pull each of the creation of the default items into it's own functions to make future refactoring easier. I'll do that in another commit as well so you can see what I mean, before we merge.

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,193 @@
...
+		check: (target) => target.hasClass("lobby"),
+		className: "join",
+		display: "Join a channel…",
+		data: (target) => target.data("id"),
+		callback: openJoinForm,
+	});
+}
+
+function addDefaultItems() {
+	addWhoisItem();
```
McInkay: So, this now just calls each of the functions. I'm not 100% on this just now. Ideally, each of these functions are called somewhere else (next to code for this feature) but ... that doesn't really exist just now for most of it. So we'll keep this all here for now, but ... eventually it'll be nicer, I think.

McInkay:
Aight, done. It's much nicer now, and probably the best we'll get at this point in time.

client/js/contextMenu.js
```diff
@@ -0,0 +1,85 @@
...
+module.exports = class ContextMenu {
+	constructor(contextMenuItems, contextMenuActions, selectedElement, event) {
+		this.contextMenuItems = contextMenuItems;
+		this.selectedElement = selectedElement;
+		this.event = event;
+
+		contextMenuContainer = $("#context-menu-container");
+		contextMenu = $("#context-menu");
+
+		contextMenuContainer.on("click contextmenu", function() {
```
xPaw: You're going to be binding these events multiple times, as far as I understand it, as the container is created once unlike this constructor.

McInkay: Good point, I've put this with the other ones in `lounge.js`. It's not ideal, but at least this is a very small amount of code there, which won't grow, unlike the rest of the context menu code.

client/js/contextMenu.js
```diff
@@ -0,0 +1,80 @@
...
+		this.contextMenuItems = contextMenuItems;
+		this.selectedElement = selectedElement;
+		this.event = event;
+
+		contextMenuContainer = $("#context-menu-container");
+		contextMenu = $("#context-menu");
+
+		contextMenuActions.execute = (id, ...args) => contextMenuActions[id] && contextMenuActions[id](...args);
+
+		contextMenu.on("click", ".context-menu-item", function() {
```
xPaw: This probably needs to change to `contextMenu.find(".context-menu-item").on("click")` and be bound after the context menu is drawn. Otherwise you have the same problem as I commented earlier.

McInkay: Alright done.

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,193 @@
...
+			text: "/whois " + itemData,
+		});
+
+		$(`.channel.active .users .user[data-name="${itemData}"]`).click();
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("user"),
+		className: "user",
+		display: (target) => target.text(),
```
xPaw: I didn't want to make a separate PR to reduce the rebase hell for you, but could you change this to `target.data("name")` to match `data`?

If you open context menu on a user where they sent a message as op, their nick will show up as `@xPaw` in the context menu.

McInkay: Done, and squashed.

McInkay:
Rebased and conflicts solved again.

McInkay:
Rebased and solved conflicts again.

astorije:
Rebased and solved conflicts again. 😬
(It was only my object shorthand stuff that conflicted, and I feel like you would have killed me if you had to fix conflicts over that :D)

astorije:
> Your structure works fine for when we are only adding context menu items from `contextMenu.js`, but that's not always the case.

Seeing this, I feel like either 1) you misunderstood me, 2) I'm misunderstanding you, or more than likely, 3) we're misunderstanding each other :D

I have not looked in details your implementation, but my idea was _supposed_ to work just fine even outside the class/file itself, as I indeed understood that you'd want to add entries from outside the file. In fact, that was the whole point of my idea 😅

I'm uneasy at the extra complexity of factory + singleton files here, that seems rather over-engineered for our needs. In fact, even if we need such complexity, there is a common pattern in ES6 classes for singletons, which is, in the same file, declare a class and export it with `modules.exports = new MyClass()`. But overall, I think we don't need that, we do need however to instantiate the class in the server file and pass it around to be able to add elements. This imo would allow for simpler code and potential added testability.
That said, I have not looked in details at your implementation as mentioned above, so I accept that I can be misguided and missing something entirely. I'm planning to look closely at your PR and play with it in the coming days (unlikely before next Wednesday though because of work-related stuff, but soon™ I swear 🤞).

Thanks for your patience and for taking this up in the first place! :))

McInkay:
We had a chat about this on IRC, but I wanted to summarise on here.

1. We need a place to store the elements that will be added to a context menu. 
2. That place needs to be central, because it needs to be a single place, otherwise we'll get confused with where things are stored and some context menus won't contain everything they need.
3. As far as we can tell, there are 2 ways to centralise this. Either in a singleton or in lounge.js. 

That's what we agreed on.

What follows is my opinion on the matter.

I don't want more code added to lounge.js if we can avoid it. My ideal is that lounge.js just ties all the modules together, and there's no logic in there other than requiring components in.

A singleton is useful for centralised data storage. In this case, what possible items on the context menus we have. Where many different places can add items to the context menus, and we don't want context menus to stay in memory when they aren't being used, this singleton works really well. It's not like a singleton is as bad as a global variable (and storing it in `lounge.js` would have some of the disadvantages of global variables) because modules have to require the singleton to have it in the scope, as opposed to  global variables which are forced upon a module.

The Factory pattern is about abstracting away the concept of the actual instantiation of the class you want, and hiding some of the complexity. The complexity in this case being storing the possible entries in the context menu. So, the factory pattern works really well for this.

I think your idea for the ContextMenu class that doesn't include any of the hard coded entries is perfect, and I think it was a great addition to the PR. But I think the factory should stay, for the reasons outlined above.

Moving this data to `lounge.js` gives us no advantages, and makes it more complicated for other parts of the code to add context menu items. 

I understand you want to make sure we get this right first  time. However, as I said on irc to you, none of this actually affects the API which is the hardest thing to change. If we find an issue in the future with how this is architected and would prefer to move it to `lounge.js` then I will be happy to revisit it. But personally, I think this is the best way to store this data.

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,193 @@
...
+addDefaultItems();
+
+/**
+ *
+ * @param opts
+ * @param {function(Object)} [opts.check] - Function to check whether item should show on the context menu, called with the target jquery element, shows if return is truthy
+ * @param {string|function(Object)} opts.className - class name for the menu item, should be prefixed for non-default menu items (if function, called with jquery element, and uses return value)
+ * @param {string|function(Object)} opts.data - data that will be sent to the callback function (if function, called with jquery element, and uses return value)
+ * @param {string|function(Object)} opts.display - text to display on the menu item (if function, called with jquery element, and uses return value)
+ * @param {boolean} [opts.divider] - Whether to put a divider after this option in the menu
```
astorije: I think it would be nicer to have a dedicated `addDivider` function instead.

That would match previous APIs I have seen so far, but mostly, I think it makes for a slightly simpler usage: when adding items conditionally (`if (...) { addItem(...) }`), you'd require extra check to add a divider after the last item, while an extra function would just go at the end.

If this is unclear, and this might very well be, I'm happy to give a proper example.

McInkay: Yeah, that makes sense. Will do that tomorrow.

McInkay: @astorije updated that.

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,204 @@
...
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/whois " + itemData,
+		});
+
+		$(`.channel.active .userlist .user[data-name="${itemData}"]`).trigger("click");
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("user"),
```
astorije: Alright, I'm looking only at the exposed API first, because that's what really important.

One question: this uses a single context menu to which we add everything, and each item has its own `check` that we use at render time to decide if it needs to be displayed or hidden, depending on what the target is, correct?

I think it would make for a nicer API if we exposed multiple context menus depending on the context, and when writing a package, one would add items to their expected locations (`channelListMenu.addItem(...)`, `userMenu.addItem(...)`, etc.).
That would also allow for more "specialized" context menus. Totally random example, but imagine context menus for users that would contain specific styling and organization to include avatars or whatnot (I know this is not possible, I said it was random lol). Such context menu could extend the main context menu, or not, but the API would be the same, just a different container.
I'm sure there are ways to make this work with the current solution, just weighing in options to pick what seems the most sensible.

I totally understand if you don't want to bother with that since you already did the work, and also because this PR is merely a modularization of what we already have, but since this is a rather different API shift, I just wanted to mention that before we ship. Whatever the answer is fine as long as we considered our options :)

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,204 @@
...
+	createContextMenu,
+};
+
+addDefaultItems();
+
+/**
+ *
+ * @param opts
+ * @param {function(Object)} [opts.check] - Function to check whether item should show on the context menu, called with the target jquery element, shows if return is truthy
+ * @param {string|function(Object)} opts.className - class name for the menu item, should be prefixed for non-default menu items (if function, called with jquery element, and uses return value)
```
astorije: You can tell when someone writes React 😁

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,204 @@
...
+ * @param {string|function(Object)} opts.data - data that will be sent to the callback function (if function, called with jquery element, and uses return value)
+ * @param {string|function(Object)} opts.display - text to display on the menu item (if function, called with jquery element, and uses return value)
```
astorije: I often see `displayName` as a wording to call this kind of thing (`label` too). Nitpicky and don't feel like I expect you to change, I just wanted to put this out there in case you were having second thoughts.
(For a second, I wondered if this would be `display` as in `display: none` and thought it could be confusing, but that might be too far fetched)

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,204 @@
...
+ * @param {function(Object)} opts.callback - Function to call when the context menu item is clicked, called with the data requested in opts.data
+ */
+function addContextMenuItem(opts) {
```
astorije: Do you mind giving the very bare minimal example you expect a package to look like? Like, nothing fancy, doesn't have to be a real use case, but I think writing out how an API gets used is a good way to validate said API.

client/js/contextMenuFactory.js
```diff
@@ -0,0 +1,204 @@
...
+	opts.check = opts.check || (() => true);
+	opts.actionId = contextMenuActions.push(opts.callback) - 1;
+	contextMenuItems.push(opts);
+}
+
+function addContextDivider(opts) {
```
astorije: You don't export `addContextDivider`, is that on purpose?

Also, if I understand correctly, adding a divider can be done from a package with `addContextMenuItem({ /* … */, divider: true /* … */ })`, that's not something we want, right?

This is internal and not API anymore so no big deal, but we could have different classes for these things, and the `items.push` stores those objects directly. `instanceof` to know what type it is, or the constructor of those objects sets a `type` property, etc., but that could help with separation of concerns.

client/js/contextMenuFactory.js
McInkay: Yeah, so I had thought about that, but I decided not to go for it because currently we don't define different kinds of context menu, so I think it would probably require quiet a bit more refactoring.

I do, however, think this is a good thing. How would you feel about leaving that till a future PR but not adding this functionality to the plugin API until it is done. I don't want to put too much in this because it's quite a refactor as it is, but it doesn't change anything underlying. I'm perfectly happy with doing what you suggest before we add it to the plugin API, but I think it's probably a separate PR. Does that make sense?

client/js/contextMenuFactory.js
McInkay: Happy to change this.

client/js/contextMenuFactory.js
McInkay: Bear in mind this doesn't include package functionality, but I'm happy to show an example of using the API, sure.

client/js/contextMenuFactory.js
McInkay: > You don't export addContextDivider, is that on purpose?

I didn't do it because it's not used elsewhere, so IDEs complain :-P we can easily add it when we add the package API functionality or call it from elsewhere.

> Also, if I understand correctly, adding a divider can be done from a package

Yeah, that could be done, however it's not like it'll break someone else's code. It's something that someone can do but it would just be them using an undocumented internal API, which we can't always get rid of. 

> This is internal and not API anymore so no big deal, but we could have different classes for these things, and the items.push stores those objects directly. instanceof to know what type it is, or the constructor of those objects sets a type property, etc., but that could help with separation of concerns.

I'm not a real fan of using instanceof because it feels a bit hacky. If we were to do this (we don't have classes at all for the elements) then it would be better to get the classes to generate the html and we just call the function for them to do that. 

I'll have a play around, see what I can do.

client/js/contextMenuFactory.js
astorije: > How would you feel about leaving that till a future PR but not adding this functionality to the plugin API until it is done.

Completely fine. I totally realize that what I'm describing here is a complete shift from what we currently have compared to your PR which is "just" exposing the current capabilities via an API. This would require changing _a lot_ of things underneath, and it's a PR on its own (even if we were not thinking of packages, we'd possibly consider this on its own).

Reason why I brought that up now is because it'll be a rather different API and I wanted to make sure we're on the same page. Our package APIs are probably going to be unstable for a while anyway, so that's fine.

client/js/contextMenuFactory.js
astorije: You mean there is no way to import a package into a TL instance? If so then that's fine, make all the assumptions or simplifications you need, it's just to make sure we are both looking at the same thing, calibrate expectations, if that makes sense.
(Also a poor-man's documentation for whenever someone stops by and wonders about the API)

client/js/contextMenuFactory.js
astorije: > we can easily add it when we add the package API functionality or call it from elsewhere.

Yeah that's fine. I was just checking it was on purpose and not forgotten as part of the change.

> I'm not a real fan of using instanceof because it feels a bit hacky.

Over the years my opinion about using `instanceof` has gone back and forth, now I'm like heh, if it does the job 🤷‍♂️ but I get where you're coming from, and I was requesting to use that specifically.

Idea behind it is to separate what is an "Item" and what is a "Divider" (and whatever comes next) so that, among other things, API doesn't leak one into another.
But as I mentioned, it's internal so I don't really care at this point. And I would prefer if we can make sure package authors cannot create a divider like this, but as you said, undocumented API, you're on your own.

client/js/contextMenuFactory.js
McInkay: > You mean there is no way to import a package into a TL instance?

In this PR, yes. That's in #1972.

client/js/contextMenuFactory.js
McInkay: Yeah, let's leave this out of this PR, but I agree it would be a nice way to go eventually.

McInkay:
I've assigned myself because there are code changes to do, but I've left you assigned, @astorije because I believe you haven't finished reviewing for now. Feel free to unassign yourself when you've finished reviewing and I'll make sure changes are done.

client/js/contextMenu.js
```diff
@@ -0,0 +1,85 @@
...
+		contextMenu.find(".context-menu-item").on("click", function() {
+			const $this = $(this);
+			const itemData = $this.data("data");
+			const contextAction = $this.data("action");
+			contextMenuActions.execute(contextAction, itemData);
+		});
+	}
+};
+
+function showContextMenu(contextMenuItems, selectedElement, event) {
```
astorije: I believe you wanted to make some code changes, you might as well bring the content of this function back into the method with the same name in the class above, no?

McInkay: I guess so...I kinda prefer splitting out the actual code to show the context menu from the logic around binding the events to it, but if you want this in, then sure.

astorije: I mean, I'm not _demanding_ it, if you don't like that, that's fine, I just don't see the point of the extra level of indirection in this case, but maybe I'm just missing something here or in what you're planning to add/change soon after. Your call 🤷‍♂️

McInkay: I mean I won't change it unless you ask me to because I prefer this way :-P but I'm not so attached to it that I will argue too much about it.

The reason being is that I feel like the `open()` function has 2 logical blocks, this is 1 of the logical blocks, and having this logical block inlined would make that function a little harder to read. I like to split things up into smaller blocks because it makes it easier to structure and read.

client/js/lounge.js
```diff
@@ -63,128 +62,8 @@ $(function() {
...
-		contextMenu
-			.html(output)
-			.css(positionContextMenu($(that), e));
-
-		return false;
-	}
-
 	viewport.on("contextmenu", ".network .chan", function(e) {
-		return showContextMenu(this, e);
+		return contextMenuFactory.createContextMenu(this, e).show();
```
astorije: Correct me if I'm wrong: I feel like you'd be better of passing `$(this)` around (the repetition of `$(this)` vs. `this` doesn't warrant abstracting that IMO, I'd rather consistently pass jQuery objects around in your `contextMenuFactory` method and `new ContextMenu`), then in [here](https://github.com/thelounge/thelounge/pull/1878/files#diff-d15cea7cb8a0229dd9b492ffa0d7ac15R59) you can do:

```js
.css(positionContextMenu(selectedElement, event));
```

(i.e. use the jQuery object itself rather than wrapping it with `$()` every time)

client/js/contextMenu.js
astorije: Sure, doesn't matter much anyway.

client/js/lounge.js
McInkay: Yeah, sure

McInkay:
Alright, I think this is done now, ready for you to test @astorije. Should be pretty easy to test, just check there are no missing items compared to master, and make sure all the buttons still do what they are meant to. Make sure I've not missed antyhing, but it looks fine to me now.

astorije:
I reviewed, tested, and approved your PR, so merge away, but I wanted to reply to your message from a few days ago.

> 1. We need a place to store the elements that will be added to a context menu.
> 2. That place needs to be central, because it needs to be a single place, otherwise we'll get confused with where things are stored and some context menus won't contain everything they need.
> 3. As far as we can tell, there are 2 ways to centralise this. Either in a singleton or in lounge.js.

I'm in full agreement with 1 and 2, not exactly with 3 (if I interpret what you said correctly). A pattern I have seen a lot recently (especially in apps with local stores like Redux or the like) is some part of the codebase centralizes the store, but it doesn't have to be on the top-level file, in our case `lounge.js`.
I can see how some files take care of the store, with an instance owned at `lounge.js` level (but that's nothing more than one assignment, at this point) which is then passed around.

> I don't want more code added to lounge.js if we can avoid it. My ideal is that lounge.js just ties all the modules together, and there's no logic in there other than requiring components in.

Absolutely. I think it wouldn't be out of place or against your ideal that the store is one of those modules that `lounge.js` ties together. Just to clarify that we're on the same page: I am not either suggesting to add code to `lounge.js` :)

Overall, I guess what I'm not a fan of is that if we replicate this pattern to other parts that need storing, we'll have, for each, a singleton, a factory, etc. That seems like a lot of boilerplate/overhead to me, and not that much a controlled/centralized environment. I can see it in this PR already, where we have logic spread across files, and with it the stored data.

That said, I totally accept that I might be misunderstanding your vision. Also, these are just words, and as I do not have time to contribute to this part (can't do everything obviously), take them at what they are, just opinions and voiced concerns in case there was something we didn't think of. I just want us to be open to alternatives if that becomes hard to manage, which could be exploring local state solutions outside of the singleton/factory model.

Anyway, thanks a lot for taking care of this, and great job on getting us somewhere :)

McInkay:
You don't have to tell me to merge twice 😛 will respond to your other stuff later, as that's more theoretical for the future rather than this pr being reliant on this just now

McInkay:
> A pattern I have seen a lot recently (especially in apps with local stores like Redux or the like) is some part of the codebase centralizes the store, but it doesn't have to be on the top-level file, in our case lounge.js.
I can see how some files take care of the store, with an instance owned at lounge.js level (but that's nothing more than one assignment, at this point) which is then passed around.

The problem with using this in this case is that there's quite a bit of code around instantiating a context menu object, unless we want to make the context menu object more busy by putting that code in there. So either it needs to go somewhere else, or we will be duplicating 3/4 lines everywhere we want to create a context menu, rather than just the 1 line in the factory. In most other cases, we won't be instantiating something. For example, when we add a new command api, there's no "create commands" littered around our code, so we won't need a factory in this case, we just need the singular place to store the commands. So, this pattern of having 2 files will be unique, or at least very rare.

Does that make sense?

As a second thing, the problem with storing it in `lounge.js` and passing it to other objects is that means `lounge.js` needs to know all the dependencies that all the objects it creates wants, which makes the instantiation in `lounge.js` a bit messier. It's not something I'll fight against too much, but I much prefer the structure of modules requiring their own dependencies. We just aren't doing this for packages because that would not work (they don't know where to look for the dependencies, and we can't have them requiring random internal js files). I'm really not a fan of dependency injection in many cases, as I feel it hides some "magic" and requires something to pass stuff to us. I get that it makes testing slightly easier, but not a huge amount.

> That seems like a lot of boilerplate/overhead to me, and not that much a controlled/centralized environment. I can see it in this PR already, where we have logic spread across files, and with it the stored data.

I mean, there isn't much boilerplate at all. And separating the logic across files is because the `contextMenu` class shouldn't have default objects stored in it. I could see the argument of having the factory method instead of a factory class, but at this point in time, when we have so many default objects, it makes sense (imo) to have a factory class.

As I mentioned above, this won't happen in most cases because we won't need a factory/class structure, because most things won't fit that. Some might, but in those cases, they might not need a factory class, it just depends on each of them. I don't think we are setting a precedent here.

But anyway, now we are just arguing over theoreticals. I'll keep adding more PRs for changing things and we can argue over them 1 by 1, and if you feel like I'm adding too many classes after a couple of PRs then we can see if there's a nicer way to do it, but I honestly don't think this will be used much.

> I just want us to be open to alternatives if that becomes hard to manage, which could be exploring local state solutions outside of the singleton/factory model.

Absolutely, as I've said, if this gets out of control or every new API adds 2 new files because I've set a precedent, then totally happy to revisit.

You know I like refactoring things :-P 

> Anyway, thanks a lot for taking care of this, and great job on getting us somewhere :)

np, thanks for taking the time to help me get this better, I really liked your idea of context menu being a class, that is nice in this case.

PR #1881, opened 2017-12-23
Fix join channel UI opening by itself

astorije:
Follow-up of #1836.

Without this, `target` was matching all of the lobby link, the "Join a channel..." button, and the join form submit button, causing the form to open on page load.
This change restricts to the first one.

### DIFF
--- a/client/js/socket-events/init.js
+++ b/client/js/socket-events/init.js
@@ -59,17 +59,17 @@ function openCorrectChannel(clientActive, serverActive) {
 
 	// Open last active channel
 	if (clientActive > 0) {
-		target = sidebar.find("[data-id='" + clientActive + "']");
+		target = sidebar.find(`.chan[data-id="${clientActive}"]`);
 	}
 
 	// Open window provided in location.hash
 	if (target.length === 0 && window.location.hash) {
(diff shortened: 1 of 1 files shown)


client/js/socket-events/init.js
```diff
@@ -59,17 +59,17 @@ function openCorrectChannel(clientActive, serverActive) {
...
 	// Open window provided in location.hash
 	if (target.length === 0 && window.location.hash) {
-		target = $("#footer, #sidebar").find("[data-target='" + escape(window.location.hash) + "']");
+		target = $("#footer, #sidebar").find(`.chan[data-target="${escape(window.location.hash)}"]`);
 	}
 
 	// Open last active channel according to the server
 	if (serverActive > 0 && target.length === 0) {
-		target = sidebar.find("[data-id='" + serverActive + "']");
+		target = sidebar.find(`.chan[data-id="${serverActive}"]`);
```
astorije: So, I could hit this block (by loading `http://localhost:9000/` which then adds the `#chan-X` fragment) and the previous one (by loading `http://localhost:9000/#chan-X`), but how do you reach the first one (`// Open last active channel`) and the last one (`// Open first available channel`)?
I couldn't find a way.

PR #1882, opened 2017-12-23
Improve wording of join channel UI form restrictions

astorije:
This gives a more actionable message to the user if they do not enter a valid value:

screen shot 2017-12-23 at 03 06 32

### DIFF
--- a/client/views/join_channel.tpl
+++ b/client/views/join_channel.tpl
@@ -1,5 +1,5 @@
 <form id="join-channel-{{id}}" class="join-form" method="post" action="" autocomplete="off">
-	<input type="text" class="input" name="channel" placeholder="Channel" pattern="[^\s]+" maxlength="200" title="Should be a valid channel name" required>
-	<input type="password" class="input" name="key" placeholder="Password (optional)" pattern="[^\s]+" title="Should be a valid channel key" maxlength="200" >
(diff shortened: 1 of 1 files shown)


client/views/join_channel.tpl
```diff
@@ -1,5 +1,5 @@
 <form id="join-channel-{{id}}" class="join-form" method="post" action="" autocomplete="off">
-	<input type="text" class="input" name="channel" placeholder="Channel" pattern="[^\s]+" maxlength="200" title="Should be a valid channel name" required>
-	<input type="password" class="input" name="key" placeholder="Password (optional)" pattern="[^\s]+" title="Should be a valid channel key" maxlength="200" >
+	<input type="text" class="input" name="channel" placeholder="Channel" pattern="[^\s]+" maxlength="200" title="The channel name may not contain spaces" required>
+	<input type="password" class="input" name="key" placeholder="Password (optional)" pattern="[^\s]+" maxlength="200" title="The channel password may not contain spaces">
```
xPaw: I think that type should be changed to `text` from `password` so that autofill plugins don't pick this up as a user/pass field.

astorije: Let's do this in a further PR. I want to fix the title, but I also want to see if we can't find a better way than displaying the password to solve that. I'd hate to have to do that.

Issue #1885, opened 2017-12-23
Drop nickname length restriction for highlighting


realies:
Currently, highlighting works if nickname length is equal or more than 4, usernames shorter than the limit get ignored. The reported reason for this is that common words can clash with short nicknames such as the word `the`, which might be confusing to users. Thinking about it, the chance of a common word clashing with a short nickname seems similar to a common word clashing with a long nickname. Happy to do some statistics if anyone's interested, although think there shouldn't be a limit to the nickname highlighting functionality.

Issue #1776, opened 2017-11-29
Nick detection in messages highlights some messages wrong 


MiniDigger:
lucks test luck (luck is an user, it colors the first luck but not the second, should be the other way around). 
Usernames in actions don't seem to get colored at all.

I'll add more cases when I find them but most of the time when it highlighted it wrong was related to the first case.

xPaw:
If you send same nick multiple times, it also links just the first one.

PR #1886, opened 2017-12-23
Drop nickname length restriction for highlighting

realies:
#1885


xPaw:
1-3 character nicks will definitely be more prone the the bug where client matches wrong instance of the nick in a sentence (#1776), so I think that should be fixed first before we drop the nick length restriction.

xPaw:
Superseded by #1901.

PR #1889, opened 2017-12-24
Display the correct window on page reload

### LINKED ISSUE #1888: Reloading the page while visiting settings/help/etc. now loads a channel
Probably broken by my changes on changelog or something. _Sigh..._

astorije:
This commit packs 2 fixes:

- Fixes #1888: Reloading the page while viewing the settings or help window would now send you to a channel window. This was broken when fixing an issue with the "Join a channel..." form (#1881)
- Reloading the page while viewing the changelog would also send you to a channel window. This is because, on reload, the click emitted was restricted to sidebar and footer, while it had to be emitted from the help page. It was already a bug while working on #1327. It is yet another aspect of our wonky window management and hopefully we can make this much more robust with a better overall solution.

### DIFF
--- a/client/js/socket-events/init.js
+++ b/client/js/socket-events/init.js
@@ -64,7 +64,8 @@ function openCorrectChannel(clientActive, serverActive) {
 
 	// Open window provided in location.hash
 	if (target.length === 0 && window.location.hash) {
-		target = $("#footer, #sidebar").find(`.chan[data-target="${escape(window.location.hash)}"]`);
+		target = $("#footer, #sidebar, #help")
+			.find(`[data-target="${escape(window.location.hash)}"]`);
 	}
 
 	// Open last active channel according to the server


client/js/socket-events/init.js
```diff
@@ -64,7 +64,8 @@ function openCorrectChannel(clientActive, serverActive) {
 
 	// Open window provided in location.hash
 	if (target.length === 0 && window.location.hash) {
-		target = $("#footer, #sidebar").find(`.chan[data-target="${escape(window.location.hash)}"]`);
+		target = $("#footer, #sidebar, #help")
```
xPaw: Would this break anything if you just searched on the entire body?

astorije: I'm not... sure...

_Probably_ not, though [this `data-target` is suspicious](https://github.com/thelounge/lounge/blob/astorije/fix-window-on-load/client/views/chat.tpl#L2).

xPaw: That looks like something that can be removed. Doesn't it point to itself?

astorije:
Alright, will do both of your suggestions in a further PR, to fix the bug and so we can test this correctly.
Touching this code is very much like gambling lol.

PR #1897, opened 2017-12-24
Improve the version checking and changelog features

astorije:
Follow-up of #1327.

- There is no client caching of the changelog/version anymore. Instead, server returns the expiration date of its cache, and that is used by the client as well. I don't know what I was thinking with my multiple caches, but this is much much nicer and more user-friendly.
- There is now a "Check now" button on the client that appears when data is stale. This means that info is fetched only once and never refreshed (it was refreshed every hour before) unless the user explicitly wants to check latest version, which in turn is as stale as server info is, i.e. 15 minutes max.
- Button style is shared with the "Join a channel" feature, `.btn-small` (not `.btn-sm` to be explicit that this is not a Bootstrap thing).
- Version checker content is now centralized in the `version_checker` template, instead of being partially in the checker template, partially in the Help template,  and partially in the code.
- A "Try again" button lets user attempt to fetch info instead of forcing them to reload the page.
- Use Flexbox to display a nicer version checker: icon is slightly bigger, and button is always aligned on the right. Looks much nicer on mobile.
- Changelog logic has been removed from `lounge.js` and moved into the component file.
- Changelog template is only passed what it needs instead of everything the server gives us.
- Public version now displays version checker, since server is caching things.
- Cleaner code overall.
- Make sure all links will open a new tab instead of exiting the application
- Add (semi-working) image viewer support for images shown in changelogs. Semi-working because this does not support Preview/Next. The existing image viewer is very custom to in-channel message previews and expects a certain markup, that needs rework to be more generic (see #1898).

screen shot 2017-12-24 at 16 46 04
screen shot 2017-12-24 at 16 46 08
screen shot 2017-12-24 at 16 51 22
screen shot 2017-12-24 at 16 45 48
^ On mobile, icon is vertically centered.
screen shot 2017-12-24 at 16 46 52

### DIFF
--- a/client/js/socket-events/changelog.js
+++ b/client/js/socket-events/changelog.js
@@ -21,8 +21,12 @@ function requestIfNeeded() {
 socket.on("changelog", function(data) {
 	// 1. Release notes window for the current version
 	$("#changelog").html(templates.windows.changelog(data.current));
+
+	const links = $("#changelog .changelog-text a");
 	// Make sure all links will open a new tab instead of exiting the application
-	$("#changelog .changelog-text a").attr("target", "_blank");
+	links.attr("target", "_blank");
+	// Add required metadata to image links, to support built-in image viewer
+	links.has("img").addClass("toggle-thumbnail");
 
 	// 2. Version checker visible in Help window
 	let status;
--- a/client/js/socket-events/changelog.js
+++ b/client/js/socket-events/changelog.js
@@ -4,19 +4,60 @@ const $ = require("jquery");
 const socket = require("../socket");
 const templates = require("../../views");
 
+module.exports = {
+	requestIfNeeded,
+};
+
+// Requests version information if it hasn't been retrieved before (or if it has
+// been removed from the page, i.e. when clicking on "Check now". Displays a
+// loading state until received.
+function requestIfNeeded() {
+	if ($("#version-checker").is(":empty")) {
+		renderVersionChecker({status: "loading"});
+		socket.emit("changelog");
+	}
+}
+
 socket.on("changelog", function(data) {
-	const container = $("#changelog-version-container");
+	// 1. Release notes window for the current version
+	$("#changelog").html(templates.windows.changelog(data.current));
+
+	// 2. Version checker visible in Help window
+	let status;
 
 	if (data.latest) {
-		container.addClass("new-version");
-		container.html(templates.new_version(data));
+		status = "new-version";
 	} else if (data.current.changelog) {
-		container.addClass("up-to-date");
-		container.text("The Lounge is up to date!");
+		status = "up-to-date";
 	} else {
-		container.addClass("error");
-		container.text("An error has occurred, try to reload the page.");
+		status = "error";
 	}
 
-	$("#changelog").html(templates.windows.changelog(data));
+	renderVersionChecker({
+		latest: data.latest,
+		status,
+	});
+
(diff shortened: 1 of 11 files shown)


client/js/socket-events/changelog.js
```diff
@@ -4,19 +4,70 @@ const $ = require("jquery");
...
+function requestIfNeeded() {
+	if ($("#version-checker").is(":empty")) {
+		renderVersionChecker({status: "loading"});
+		socket.emit("changelog");
+	}
+}
+
 socket.on("changelog", function(data) {
-	const container = $("#changelog-version-container");
+	let status;
```
xPaw: Can't this be done in the template? I believe thats how I did it initially.

client/views/windows/help.tpl
```diff
@@ -13,11 +13,7 @@
 	</h2>
 
 	<div class="about">
-		{{#unless public}}
-			<p id="changelog-version-container" class="changelog-version">
-				Checking for updates...
-			</p>
-		{{/unless}}
+		<div id="version-checker" />
```
xPaw: Write out `</div>` please. We don't do this anywhere else.

client/js/socket-events/changelog.js
astorije: No, because if I do this in template, it recreates the whole div instead of changing the content/class, and that loses the CSS transition.

I would also need to nest 2 divs (one with `id="version-checker"` and one with the correct class). And also what's nice is that I don't have server logic in the template. By server logic I mean `if (data.latest)` and `if (data.current.changelog)` to define the status, template is pretty straightforward that way.

I will however group things together, it's silly right now, I set things for the version checker, then changelog, then version checker.

client/views/windows/help.tpl
astorije: Sure. It's perfectly valid so I don't think we should restrict ourselves though (interestingly, that's extremely common to see that in React-land, not sure about Vue).

astorije:
@xPaw, done.

client/js/socket-events/changelog.js
```diff
@@ -4,19 +4,70 @@ const $ = require("jquery");
...
+	renderVersionChecker({
+		latest: data.latest,
+		status,
+	});
+
+	// When there is a button to refresh the checker available, display it when
+	// data is expired. Before that, server would return same information anyway.
+	if (data.expiresAt) {
+		setTimeout(
+			() => $("#version-checker").find("#check-now").show(),
```
xPaw: This can be a single selector.

client/js/socket-events/changelog.js
```diff
@@ -4,19 +4,70 @@ const $ = require("jquery");
...
+$("#help").on("click", "#check-now", () => {
+	$("#version-checker").empty();
+	requestIfNeeded();
+});
+
+// Given a status and latest release information, update the version checker
+// (CSS class and content)
+function renderVersionChecker({status, latest}) {
+	const container = $("#version-checker");
+	container.attr("class", status);
```
xPaw: You can chain these calls.

client/js/socket-events/changelog.js
astorije: Good catch, done! Also checked that this was still working.

client/js/socket-events/changelog.js
astorije: Done, much simpler now.

PR #1899, opened 2017-12-25
Improve a bit window loading on init

astorije:
Follow-up of https://github.com/thelounge/lounge/pull/1889#discussion_r158595946.

### DIFF
--- a/client/js/socket-events/init.js
+++ b/client/js/socket-events/init.js
@@ -64,8 +64,7 @@ function openCorrectChannel(clientActive, serverActive) {
 
 	// Open window provided in location.hash
 	if (target.length === 0 && window.location.hash) {
-		target = $("#footer, #sidebar, #help")
-			.find(`[data-target="${escape(window.location.hash)}"]`);
+		target = $(`[data-target="${escape(window.location.hash)}"]`).first();
 	}
 
 	// Open last active channel according to the server
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -174,7 +174,7 @@ $(function() {
 	});
 
 	viewport.on("click", "#chat .menu", function(e) {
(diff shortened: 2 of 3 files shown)


xPaw:
Have you looked through the code that `target` is not referenced anywhere?

client/js/socket-events/init.js
```diff
@@ -64,8 +64,7 @@ function openCorrectChannel(clientActive, serverActive) {
 
 	// Open window provided in location.hash
 	if (target.length === 0 && window.location.hash) {
-		target = $("#footer, #sidebar, #help")
-			.find(`[data-target="${escape(window.location.hash)}"]`);
+		target = $("body").find(`[data-target="${escape(window.location.hash)}"]`);
```
xPaw: Either use `document.body` or you can do `$(`[data-target="...` directly.

astorije:
> Have you looked through the code that target is not referenced anywhere?

I did before, but not carefully enough. Will do it better today.

client/js/socket-events/init.js
astorije: You're right, will do the latter. I'll also add a `.first()` too, just in case.

astorije: Done.

astorije:
Actually @xPaw, I think it might be used here: https://github.com/thelounge/lounge/blob/astorije/improve-init/client/js/socket-events/quit.js#L14

What do you think? There are a looot of uses of `target` in our client code, it's hard to see the big picture.

xPaw:
That code works on the sidebar doesnt it? Should be fine.

astorije:
Doesn't it remove the current window when quitting?

astorije:
@xPaw, I confirm that this breaks disconnecting from the lobby context menu (right click context menu still works though) :/
So I guess I'll have to leave it there (the target part).

PR #1901, opened 2017-12-25
Do not match nicknames incorrectly as part of a bigger word

### LINKED ISSUE #1776: Nick detection in messages highlights some messages wrong

### LINKED ISSUE #1885: Drop nickname length restriction for highlighting

xPaw:
Fixes #1776
Fixes #1885

I tried really hard to inject indexes into the style parser as sent by the server but couldn't figure out a solution, so I did the same regex as the server (minus colors) to fix it instead.

### DIFF
--- a/client/js/libs/handlebars/ircmessageparser/findNames.js
+++ b/client/js/libs/handlebars/ircmessageparser/findNames.js
@@ -1,17 +1,25 @@
 "use strict";
 
+const nickRegExp = /([\w[\]\\`^{|}-]+)/g;
+
 function findNames(text, users) {
 	const result = [];
-	let index = -1;
 
-	users.forEach((nick) => {
-		index = text.indexOf(nick, ++index);
-		result.push({
-			start: index,
-			end: index + nick.length,
-			nick: nick,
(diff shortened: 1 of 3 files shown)


astorije:
@MaxLeiter, would you be able to review this?

astorije:
@xPaw, is there any new or improved tests we could add for this?

xPaw:
@astorije Added.

astorije:
You rock 🎸

PR #1905, opened 2017-12-27
Mute disabled inputs

xPaw:


client/css/style.css
```diff
@@ -775,6 +775,10 @@ kbd {
 	width: 100%;
 }
 
+.input:disabled {
+	background-color: #e9ecef;
```
astorije: That would be a new shade of gray (-ish) in our CSS. Could we re-use one maybe instead? `#eee` seems very close and we use it already. Or better if you prefer something else.

client/css/style.css
```diff
@@ -775,6 +775,10 @@ kbd {
 	width: 100%;
 }
 
+.input:disabled {
```
astorije: I think we should disable `:hover` on disabled elements (so our `.input:hover` should become `.input:not(:disabled):hover`, what do you think?

client/css/style.css
xPaw: `#eee` is way too light, on one of my non-IPS displays I can't tell the difference.

PR #1907, opened 2017-12-27
Add log role to message container

### LINKED ISSUE #936: accessibility/feature request: Add aria live region role to the list of messages
Hi,
Just a quick feature request, if the aria live region role could be added to the messages from the channel you're currently in, that would be great, as then most screen readers that support the aria standard will automatically read incoming messages.
-Michael.

xPaw:
Fixes #936

### DIFF
--- a/client/views/chat.tpl
+++ b/client/views/chat.tpl
@@ -21,7 +21,7 @@
 		<div class="show-more{{#if messages.length}} show{{/if}}">
 			<button class="show-more-button" data-id="{{id}}">Show older messages</button>
 		</div>
-		<div class="messages"></div>
+		<div class="messages" role="log" aria-live="polite" aria-relevant="additions"></div>
 	</div>
 	{{#equal type "channel"}}
 	<aside class="sidebar">


client/views/chat.tpl
```diff
@@ -21,7 +21,7 @@
 		<div class="show-more{{#if messages.length}} show{{/if}}">
 			<button class="show-more-button" data-id="{{id}}">Show older messages</button>
 		</div>
-		<div class="messages"></div>
+		<div class="messages" role="log" aria-live="polite" aria-relevant="additions"></div>
```
astorije: I think we need to add `aria-atomic="true"`. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

VoiceOver on El Capitan re-reads the whole chat window at every change, which quickly becomes unmanageable. I was hoping `aria-atomic` would fix this, but it does not. @cptvitamin, do you have any insight?

xPaw: From the page you linked:

> aria-atomic is not set ("false" by default) so that only the added or removed users should be spoken and not the entire roster each time.

xPaw: I can imagine how this could happen when `more` messages are loaded. perhaps these aria attributes have to be managed by JS to avoid that?

astorije: No, it happens with literally every message, starting with the second one in a blank chat.

xPaw:
I wouldn't want to merge this if it's entirely useless for screen readers.

astorije:
Yeah. I'm not sure what to do though, it might just be one of the many weirdnesses from VoiceOver. Not sure how to get help on that.

PR #1913, opened 2017-12-28
Allow keywords as changelog script version argument, e.g. `node scripts/changelog pre`

astorije:
So, yeah, I'm about to publish a pre-release :)


astorije:
FYI I just used this for v2.7.0-pre.4 and it worked well. Makes my life yet a little easier :D

PR #1914, opened 2017-12-28
Separate and improve wording for `pre` and `rc` pre-release versions

astorije:
Planning on using this going forward, to make sure we can pre-release very often without effort. Hope to clarify the difference between `pre` and `rc` to whomever is not familiar with The Lounge (or software lifecycle in general?). Someone from @thelounge/maintainers, what do you think?

screen shot 2017-12-27 at 19 54 05

### DIFF
--- a/scripts/changelog.js
+++ b/scripts/changelog.js
@@ -93,7 +93,14 @@ function prereleaseTemplate(items) {
 
 [See the full changelog](${items.fullChangelogUrl})
 
-This is a release candidate for v${stableVersion(items.version)} to ensure maximum stability for public release.
+${prereleaseType(items.version) === "rc" ?
+		`This is a release candidate (RC) for v${stableVersion(items.version)} to ensure maximum stability for public release.
+Bugs may be fixed, but no further features will be added until the next stable version.` :
+
+		`This is a pre-release for v${stableVersion(items.version)} to offer latest changes without having to wait for a stable release.
+ At this stage, features may still be added or modified until the first release candidate for this version gets released.`
+}
+
 Please refer to the commit list given above for a complete list of changes, or wait for the stable release to get a thoroughly prepared change log entry.
 
 As with all pre-releases, this version requires explicit use of the \`next\` tag to be installed:
@@ -168,6 +175,11 @@ ${printList(items.uncategorized)}
 // 2.5.0-rc.1, etc.), or false otherwise
 function isPrerelease(v) {
 	return v.includes("-");
+}
+
+// Given a version of `x.y.z-abc.n`, returns `abc`, i.e. the type of pre-release
+function prereleaseType(v) {
+	return semver.prerelease(v)[0];
 }
 
 // Returns the stable version that this pre-release version is targeting. For


scripts/changelog.js
```diff
@@ -84,7 +84,14 @@ function prereleaseTemplate(items) {
 
 [See the full changelog](${items.fullChangelogUrl})
 
-This is a release candidate for v${stableVersion(items.version)} to ensure maximum stability for public release.
+${prereleaseType(items.version) === "pre" ?
```
YaManicKill: I would suggest we switch this round.

So, if it's "rc" then it's a release candidate. Otherwise it's a pre-release.

My reasoning is basically we will never have something else refer to a release candidate (as that is 1 very specific thing) whereas a pre-release, we may well add extra terminology for certain other types of pre-release (beta, alpha, etc) and if we do that, then it'll by default show up in here as "pre-release" (which is true) rather than currently it would show as "release candidate" (which is untrue).

Now, I know we probably won't end up doing that because I know what your opinions are on those terms, etc. But I just thought it's better to make the more specific text (rc) only when it's definitely an rc, and the more generic (pre) for all other pre-releases.

scripts/changelog.js
```diff
@@ -161,6 +168,11 @@ function isPrerelease(v) {
 	return v.includes("-");
 }
 
+// Given a version of `x.y.z-abc.n`, returns `abc`, i.e. the type of pre-release
+function prereleaseType(v) {
```
YaManicKill: Instead of putting this into the pre-release template, should we have a pre-template and an rc-template that inherit from the pre-release template? I know this is just a script that isn't part of the product, but all of our checking for which template to use happens outside of the templates, and having something that is switching inside the template is different and feels a little hacky rather than just doing the same way the other bits do.

astorije: I'll do that if/when the wording for both templates differ. Until then, that's good enough honestly. This script doesn't need the same level of details as the rest of our project, it's simply a way to save time when releasing, so any shortcut in doing so is fine.
As long as I'm making releases, I'm bound to be the only one maintaining and using it 😁

scripts/changelog.js
astorije: Suuuure, I was not putting as much thought into it as you, but yes, that makes sense. I like the idea of order of the code being order of releases (pre _then_ rc) but that's about it. Will switch.

scripts/changelog.js
YaManicKill: I know this isn't our main project and it's just a helper script but ... having an rc template that inherits from the pre-release one (my initial idea of 2 inheriting from the pre-release one was silly... pre is the pre-release one, so you'd only need an extra rc one) isn't a lot of work and does make it a bit nicer, IMO.

But sure, if you don't want to do it, that's fine.

astorije: > having an rc template that inherits from the pre-release one

I can do that, yep. I'm not sure what you mean by inheriting a template, mind giving a quick example?

YaManicKill: Hmmm...I hadn't looked at what the template actually was, my bad. It would require quite a bit of refactoring with how it's done, so let's just leave it.

astorije: Happy to improve whenever, let me know if you have ideas/suggestions.

PR #1915, opened 2017-12-28
Turn channel list into an accessible tablist

### LINKED ISSUE #1912: Make channel list an aria "tablist"
From https://www.marcozehe.de/2015/12/07/looking-at-the-accessibility-of-the-irccloud-service/

> The channels and servers as well as each private message nickname are now announced as tabs. This seemed to us the most appropriate mapping, since only one channel, server output or private message is usually visible at a time, but radio buttons didn’t seem really appropriate to use.

xPaw:
Fixes #1912.

### DIFF
--- a/client/index.html.tpl
+++ b/client/index.html.tpl
@@ -32,7 +32,7 @@
 	<body class="signed-out<%- public ? " public" : "" %>" data-transports="<%- JSON.stringify(transports) %>">
 
 	<div id="wrap">
-	<div id="viewport">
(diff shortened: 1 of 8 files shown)


astorije:
@xPaw, getting an exception after rebasing on `master`:

screen shot 2018-02-18 at 01 41 22

I wanted to play with this. Will change anything on UI or something else, or will it only affect screen readers?

xPaw:
Fixed. It's entirely just for screen readers, no UI change.

PR #1916, opened 2017-12-28
views/join_channel: disable autocompletion of form inputs

williamboman:
Browsers tend to be a bit aggressive with automatically filling form fields, even if `autocomplete="off"`. This form in particular resembles a log in form a lot as well. This trick disables autocompletion of both fields (tested in Chrome and iOS Safari).

### DIFF
--- a/client/views/join_channel.tpl
+++ b/client/views/join_channel.tpl
@@ -1,5 +1,5 @@
 <form id="join-channel-{{id}}" class="join-form" method="post" action="" autocomplete="off">
 	<input type="text" class="input" name="channel" placeholder="Channel" pattern="[^\s]+" maxlength="200" title="The channel name may not contain spaces" required>
-	<input type="password" class="input" name="key" placeholder="Password (optional)" pattern="[^\s]+" maxlength="200" title="The channel password may not contain spaces">
+	<input type="password" class="input" name="key" placeholder="Password (optional)" pattern="[^\s]+" maxlength="200" title="The channel password may not contain spaces" autocomplete="new-password">
 	<button type="submit" class="btn btn-small" data-id="{{id}}">Join</button>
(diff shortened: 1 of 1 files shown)


xPaw:
Won't this trick it into offering to save a new password? I would just change the type from password to text. Channel keys can be visible in other places anyway (when using /join, or displayed in /mode).

williamboman:
> Won't this trick it into offering to save a new password? I would just change the type from password to text. Channel keys can be visible in other places anyway (when using /join, or displayed in /mode).

Yeah. I think it's quite nice though:

image

astorije:
> I would just change the type from password to text. Channel keys can be visible in other places anyway (when using /join, or displayed in /mode).

I think we should really really avoid to display it if we can, even if it can be retrieved somewhere else.

As for the solution for this issue, I don't have a better solution (I also don't have the problem myself :man_shrugging:) so this seems fine to me.

Another :man_facepalming:-trick we could use, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion):

> In some cases, the browser will keep suggesting autocompletion values even if the autocomplete attribute is set to off. This unexpected behavior can be quite puzzling for developers. The trick to really forcing the no-autocompletion is to assign a random string to the attribute, for example:
> ```
> autocomplete="nope"
> ```
> Since this random value is not a [valid one](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete), the browser will give up.

@williamboman, could you test if this fixes the problem?

williamboman:
> @williamboman, could you test if this fixes the problem?

I've tried `autocomplete="nope"` - didn't work. This was the only working solution I could find.

PR #1917, opened 2017-12-28
fix animation jitter when clicking join channel button

williamboman:
(these gifs shows the issue very poorly)

**Before**
jitter1

**After**
jitter

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -668,7 +668,8 @@ kbd {
 }
 
 #sidebar .lobby .add-channel.opened {
-	transform: rotate(45deg);
+	/* translateZ(0) enables hardware acceleration, this is to avoid jittering when animating */
+	transform: rotate(45deg) translateZ(0);
 }
 
 #sidebar .chan.active .close {


client/css/style.css
```diff
@@ -667,7 +667,8 @@ kbd {
 }
 
 #sidebar .lobby .add-channel.opened {
-	transform: rotate(45deg);
+	// translateZ(0) enables hardware acceleration, this is to avoid jittering when animiating
```
astorije: How on Earth did that pass our CSS linter lol?!
Even my local linter (also stylelint) reported this, and actually whatever error I make in the CSS, `npm run lint:css` doesn't flag. There is something wrong in there, I'ma look right after...

astorije: `animiating` :)

astorije:
I can't see this issue on my laptop (macOS, Chrome) or phone (iOS, Safari). I'll try with an old phone (Android, Chrome) to see what happens there.

Just curious, what browser are you using? The rotation doesn't seem centered, which is something I'm not seeing either.

client/css/style.css
williamboman: Too used to writing SCSS :), fixing

williamboman:
Fixed comment now

PR #1918, opened 2017-12-28
relay client's preferred language in link preview requests 

### LINKED ISSUE #1440: Send Accept-Language in link preview requests
Some sites prefer to get language by *server* ip, so sending accept-language might help in this case.

Twitter profile descriptions are shown in german for me as my server is hosted in Germany.

williamboman:
Closes #1440.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -38,7 +38,7 @@ module.exports = function(client, chan, msg) {
 	})).slice(0, 5); // Only preview the first 5 URLs in message to avoid abuse
 
 	msg.previews.forEach((preview) => {
-		fetch(preview.link, function(res, err) {
+		fetch(preview.link, {language: client.language}, function(res, err) {
 			if (err) {
 				preview.type = "error";
 				preview.error = "message";
@@ -85,7 +85,7 @@ function parse(msg, preview, res, client) {
 
 		// Verify that thumbnail pic exists and is under allowed size
 		if (preview.thumb.length) {
-			fetch(escapeHeader(preview.thumb), (resThumb) => {
+			fetch(escapeHeader(preview.thumb), {language: client.language}, (resThumb) => {
 				if (resThumb === null
 				|| !(/^image\/.+/.test(resThumb.type))
 				|| resThumb.size > (Helper.config.prefetchMaxImageSize * 1024)) {
@@ -198,7 +198,19 @@ function emitPreview(client, msg, preview) {
 	});
 }
 
-function fetch(uri, cb) {
+function getRequestHeaders(language) {
+	const headers = {
+		"User-Agent": "Mozilla/5.0 (compatible; The Lounge IRC Client; +https://github.com/thelounge/lounge)",
+	};
+
+	if (language !== null) {
+		headers["Accept-Language"] = language;
+	}
+
+	return headers;
+}
+
+function fetch(uri, {language}, cb) {
 	let req;
 
 	try {
@@ -206,9 +218,7 @@ function fetch(uri, cb) {
 			url: uri,
 			maxRedirects: 5,
 			timeout: 5000,
-			headers: {
-				"User-Agent": "Mozilla/5.0 (compatible; The Lounge IRC Client; +https://github.com/thelounge/lounge)",
-			},
+			headers: getRequestHeaders(language),
 		});
 	} catch (e) {
 		return cb(null, e);
--- a/src/server.js
+++ b/src/server.js
@@ -204,6 +204,17 @@ module.exports = function() {
 	return server;
 };
 
+function getClientLanguage(socket) {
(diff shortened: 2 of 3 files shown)


test/plugins/link.js
```diff
@@ -266,4 +266,21 @@ describe("Link plugin", function() {
...
+			res.send();
+			done();
+		});
+
+		const message = this.irc.createMessage({
+			text: "http://localhost:9002/language-check",
+		});
+
+		link(this.irc, this.network.channels[0], message);
+	});
```
astorije: :clap:

astorije:
@xPaw, could you test that this stops German Twitter from your server located in Germany? :D

src/server.js
```diff
@@ -498,6 +505,7 @@ function performAuthentication(data) {
 		socket.emit("configuration", getClientConfiguration());
 
 		client.ip = getClientIp(socket);
+		client.language = getClientLanguage(socket);
```
xPaw: This helper function isn't really needed, you can put `socket.handshake.headers["accept-language"]` directly here.

However I would either perform some kind of validation on the string, or at least limit its length to 50 or something.

williamboman: > However I would either perform some kind of validation on the string

I was considering this but decided not to because the header is set entirely by the browser (I'd like to think that browsers sets the header correctly)

Not sure this is how one wants to determine user's language, hence the helper function.

xPaw: I primarily want to prevent abuse here, so at least limit the strings length. And perhaps only allow ASCII characters in the string.

williamboman: You mean something like this:

```diff
diff --git a/src/server.js b/src/server.js
index 74ac4964..ef78aab7 100644
--- a/src/server.js
+++ b/src/server.js
@@ -181,8 +181,9 @@ module.exports = function() {
 };

 function getClientLanguage(socket) {
-       if ("accept-language" in socket.handshake.headers) {
-               return socket.handshake.headers["accept-language"];
+       const acceptLanguage = socket.handshake.headers["accept-language"]
+       if (typeof acceptLanguage === 'string' && /^[\x00-\x7F]{1,50}$/.test(acceptLanguage)) {
+               return acceptLanguage
        }
        return null;
 }

```

xPaw: That'll most likely do, yeah.

xPaw:
Hmm, I just realized this will fundamentally conflict with #1952. I'm not sure what course of action to take here.

astorije:
@xPaw, I would say that this PR needs to go first, then in #1952 instead of storing with link as the key, it needs to be a pair link+lang.

xPaw:
@astorije That most likely will defeat the purpose because accept-language header have rather high entropy (browser/os/language dependant).

williamboman:
What about using a normalized version of the accept-language header when storing link previews?

astorije:
@xPaw, I understand but on the other hand, I'd rather have correct content and imperfect queuing than perfect queuing and incorrect content :man_shrugging:

That said, @williamboman's suggestion is pretty good. Accept language header does indeed vary a lot, but in the boundaries of a standard that we can use to extract language from.
[This](https://www.npmjs.com/package/locale) and [that](https://www.npmjs.com/package/accept-language) are not what we're looking for exactly because we don't want to provide a list of supported locales I think, but there might be some logic to re-use (or maybe other packages, I haven't looked to far).

I'm still in favor of this PR so we can improve #1952 on top of it.

src/server.js
astorije: @xPaw, helper function could help with unit testing, if we wanted to do that :man_shrugging:

src/plugins/irc-events/link.js
```diff
@@ -170,16 +170,24 @@ function emitPreview(client, msg, preview) {
...
+	const headers = {
+		"User-Agent": "Mozilla/5.0 (compatible; The Lounge IRC Client; +https://github.com/thelounge/lounge)",
+	};
+	if (client.language !== null) {
+		headers["Accept-Language"] = client.language;
+	}
+	return headers;
+}
+
+function fetch(uri, client, cb) {
```
astorije: `client` is only passed to get the language, so it seems like a heavy thing to do (I'd like to avoid passing things like `client`, `clientManager`, etc. too much, it makes code even more coupled than it currently is).

`function fetch(uri, language, cb)` would do, or we could go ES6 (it's gonna be for v3 anyway, where we drop Node v4), and do `function fetch (uri, {language}, cb)`, that you call with `fetch(preview.link, {language: client.language}, function(res) {` (I know `fetch(preview.link, client, function(res) {` would still work, but it's less explicit and still passes the whole thing around).

williamboman: fixed, went with functon parameter destructuring

astorije:
@williamboman, sorry we just introduced this, but could you fix the conflict? It should be pretty simple.

PR #1919, opened 2017-12-28
Show error if image is greater than max prefetch size

McInkay:
Currently, it just shows nothing, which can be quite confusing

### DIFF
--- a/client/js/libs/handlebars/friendlysize.js
+++ b/client/js/libs/handlebars/friendlysize.js
@@ -1,0 +1,10 @@
+"use strict";
+
+const sizes = ["Bytes", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"];
+
+module.exports = function(size) {
+	// Loosely inspired from https://stackoverflow.com/a/18650828/1935861
+	const i = size > 0 ? Math.floor(Math.log(size) / Math.log(1024)) : 0;
+	const fixedSize = parseFloat((size / Math.pow(1024, i)).toFixed(1));
+	return `${fixedSize} ${sizes[i]}`;
+};
--- a/client/views/msg_preview.tpl
+++ b/client/views/msg_preview.tpl
@@ -29,10 +29,14 @@
 		</a>
 	{{/equal}}
 	{{#equal type "error"}}
-		<a class="toggle-text" href="{{link}}" target="_blank" rel="noopener">
-			<div class="head" title="{{head}}">{{head}}</div>
-			<div class="body" title="{{body}}">{{body}}</div>
-		</a>
+		{{#equal error "image-too-big"}}
+			<em>
+				This image is larger than {{friendlysize maxSize}} and cannot be
+				previewed.
+				<a href="{{link}}" target="_blank" rel="noopener">Click here</a>
+				to open it in a new window.
+			</em>
+		{{/equal}}
 	{{/equal}}
 </div>
 {{/preview}}
--- a/client/views/msg_preview.tpl
+++ b/client/views/msg_preview.tpl
@@ -28,5 +28,11 @@
 			<div class="body" title="{{body}}">{{body}}</div>
 		</a>
 	{{/equal}}
+	{{#equal type "error"}}
+		<a class="toggle-text" href="{{link}}" target="_blank" rel="noopener">
+			<div class="head" title="{{head}}">{{head}}</div>
+			<div class="body" title="{{body}}">{{body}}</div>
+		</a>
+	{{/equal}}
 </div>
 {{/preview}}
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -99,11 +99,13 @@ function parse(msg, preview, res, client) {
 	case "image/jpeg":
 	case "image/webp":
 		if (res.size > (Helper.config.prefetchMaxImageSize * 1024)) {
-			return;
+			preview.type = "error";
+			preview.head = "Large image";
+			preview.body = "Image is greater than your max image size. Click to view.";
+		} else {
+			preview.type = "image";
+			preview.thumb = preview.link;
 		}
-
-		preview.type = "image";
-		preview.thumb = preview.link;
 
 		break;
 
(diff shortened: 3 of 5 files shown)


src/plugins/irc-events/link.js
```diff
@@ -98,12 +98,14 @@ function parse(msg, preview, res, client) {
 	case "image/jpeg":
 	case "image/webp":
 		if (res.size > (Helper.config.prefetchMaxImageSize * 1024)) {
-			return;
+			preview.type = "error";
+			preview.head = "Large image";
```
xPaw: I don't think errors need head, just make it body in red color.

YaManicKill: It felt a bit odd without a head, to be honest. I'm not sure why. Let's see... Red text or red background? Thoughts?

|Red text|Red background|
|-|-|
|||

xPaw: Red background draws too much attention for no good reason.

MaxLeiter: Also not sure it should be “your max image size”, as the client may not own the server / instance.

YaManicKill: Yeah, I knew I wouldn't have the correct wording. Not sure of the best way to word this.

We want the user to know that this is an option on the server that can be changed, but if they aren't the admin, we don't want them to be confused.

"Image is larger than the max image size. Click to open"

That doesn't give them any indication as to how to change that (or ask their admin to change it)

Then again

"Image is larger than the server's max prefetch size. Click to open" 

That's clearly too technical. (Also isn't clear which server it's talking about, necessarily).

astorije: I did that some time ago and @xPaw didn't want to display anything on images that were too big. Wish I didn't kill the commit 😢

astorije: > We want the user to know that this is an option on the server that can be changed

I don't agree with this actually. Docs + clear configuration + potentially server logs (probably not in this case) should help server admins as much as possible, but I think any assumption that user might also be the admin will lead to confusing statements. I think a longer-term solution is to have users with an admin role (but this is a different story :D).

Anyway, in the meantime, I propose this:

screen shot 2017-12-28 at 16 30 18

@YaManicKill, I'm adding a commit with changes I made to get there, but this is purely for convenience so you don't have to copy-paste things around. Feel free to trash it if you don't like it (I do, surprising huh? 😁) and squash it if you do.

If/when we have admin role (think creating users from the interface, listing users on a TL instance, messaging all users on this instance, stats, etc.), we can add an extra line that says: "Admin: Current max size is XXX MB, you may change this value in your `config.js` file."

YaManicKill: Yeah sure, looks better.

> If/when we have admin role

I thought you didn't want this :-P I am very much loving the idea of doing this though, for the record.

astorije: > I thought you didn't want this :-P I am very much loving the idea of doing this though, for the record.

I don't remember being against this, no :)
Every time we assume user == admin, I see a use case for it! Also, it would be a convenient way to say "this user also is an admin on the server" (even though it doesn't have to technically be true, we can safely assume this).
I'll draft something for v3, it would be a good paradigm shift to introduce.

Anyway, didn't mean to hijack this PR with an unrelated feature, sorry.

client/js/libs/handlebars/friendlysize.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const sizes = ["Bytes", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"];
+
+module.exports = function(size) {
+	// Loosely inspired from https://stackoverflow.com/a/18650828/1935861
+	const i = size > 0 ? Math.floor(Math.log(size) / Math.log(1024)) : 0;
+	return `${parseFloat((size / Math.pow(1024, i)).toFixed(1))} ${sizes[i]}`;
```
MaxLeiter: I personally think this is a case where templating actually overly complicates the line, but nbd

client/views/msg_preview.tpl
```diff
@@ -28,5 +28,14 @@
 			<div class="body" title="{{body}}">{{body}}</div>
 		</a>
 	{{/equal}}
+	{{#equal type "error"}}
+		{{#equal error "image-too-big"}}
+			<em>
+				This image was too large to be previewed ({{friendlysize size}}).
```
MaxLeiter: `This image is too large to be previewed: {{size}}` or something. should be present-tense and also some sort of puncuation before size

client/js/libs/handlebars/friendlysize.js
astorije: Fair enough, I can get behind that:

```js
`${parseFloat((size / Math.pow(1024, i)).toFixed(1))} ${sizes[i]}`;
// vs.
parseFloat((size / Math.pow(1024, i)).toFixed(1)) + " " + sizes[i];
```

I can switch to second line if @thelounge/maintainers prefer, unless we have the https://eslint.org/docs/rules/prefer-template rule. I know it's rather insignificant here, but that would apply to future changes as well.

McInkay: Wouldn't it be better to pull it out and still use a template string?

```
const fixedSize = parseFloat((size / Math.pow(1024, i)).toFixed(1));
return `${fixedSize} ${sizes[i]}`
```

I feel like just not having it as a template string doesn't make it any more readable. It's still confusing to read.

MaxLeiter: that or @astorije's solution is better than what there is now imo

client/views/msg_preview.tpl
astorije: Good point. I actually went to a different route, but took this suggestion into consideration:

screen shot 2018-02-01 at 21 13 25

The rationale here is that, from a user perspective, it's of little-to-no value to know the size that was rejected. Instead, knowing the limit is much more interesting and actionable (also something user cannot know if the UI doesn't tell them). And no, having both simply because we can would not add anything to the UX 😁

xPaw: Is there any point in adding `click here to open it in a new window`? The link is already there.

astorije: I think it doesn't hurt to have it there, especially since there can be multiple images previewed that can't be loaded (wording isn't ideal in this case, but it's not that bad either).

PR #1920, opened 2017-12-28
Enforce padding-line-between-statements

xPaw:


YaManicKill:
I am against this.

I feel like this is unnecessary consistency, and adds a whole bunch of extra lines that don't necessarily add any readability. Enforcing the extra line after certain things is over the top, IMO.

xPaw:
So the problem here is that these new lines depend on who's writing the code at that time, I've noticed a couple of PR authors that write any new code that simply excludes empty lines. And IMO, when it becomes more of a pain in the ass to look through code without a single empty line when line numbers go into triple digits.

However I can understand not wanting new lines when creating a variable and calling a function on the next. I mainly wanted to tackle no new lines around if statements and returns.

YaManicKill:
See, I think those things are really dependant on whether the code needs them, and enforcing them in eslint feels over the top. I think it's better just to say to people who don't use any whitespace "Hey, make your code more readable" rather than enforcing this on everything. It'll double the length of some functions which doesn't add to the readability.

xPaw:
I've changed it to block(-like) statements only, do you find that more acceptable?

xPaw:
This pr is easy to rebase because it's auto fixable.

astorije:
@xPaw, rebase this? Yes it will cause conflicts on other PRs, but since it's auto-fixable, those other PRs can just ignore the conflicts and go with their own changes, then run `npm run lint:js -- --fix` / `yarn lint:js --fix`.

xPaw:
@astorije done

PR #1924, opened 2017-12-29
Set decoding=async on image previews

xPaw:
https://github.com/whatwg/html/issues/1920
https://github.com/ampproject/amphtml/pull/12145
https://html.spec.whatwg.org/multipage/images.html#image-decoding-hint

This is a new spec addition, and browsers are starting to implement this.

### DIFF
--- a/client/views/image_viewer.tpl
+++ b/client/views/image_viewer.tpl
@@ -9,7 +9,7 @@
 {{/if}}
 
 <a class="image-link" href="{{link}}" target="_blank">
-  <img src="{{image}}" alt="Preview of {{link}}">
+  <img src="{{image}}" decoding="async" alt="">
 </a>
 
 <a class="btn open-btn" href="{{link}}" target="_blank">
--- a/client/views/msg_preview.tpl
+++ b/client/views/msg_preview.tpl
@@ -2,7 +2,7 @@
 <div class="toggle-content toggle-type-{{type}}{{#if shown}} show{{/if}}">
 	{{#equal type "image"}}
 		<a class="toggle-thumbnail" href="{{link}}" target="_blank" rel="noopener">
-			<img src="{{thumb}}">
+			<img src="{{thumb}}" decoding="async" alt="">
 		</a>
 	{{/equal}}
 	{{#equal type "audio"}}
@@ -20,7 +20,7 @@
 	{{#equal type "link"}}
 		{{#if thumb}}
 			<a class="toggle-thumbnail" href="{{link}}" target="_blank" rel="noopener">
-				<img src="{{thumb}}" class="thumb">
+				<img src="{{thumb}}" decoding="async" alt="" class="thumb">
 			</a>
 		{{/if}}
 		<a class="toggle-text" href="{{link}}" target="_blank" rel="noopener">


client/views/image_viewer.tpl
```diff
@@ -9,7 +9,7 @@
 {{/if}}
 
 <a class="image-link" href="{{link}}" target="_blank">
-  <img src="{{image}}" alt="Preview of {{link}}">
```
xPaw: `alt` should describe an image, putting a link there doesn't make sense.

astorije: `alt` is mandatory per spec though. We can be more descriptive such as "Preview of an image posted on this channel, original located at {{link}}"` but we do have to leave something in there.

xPaw: Alt is mandatory to provide text for screen readers or when the image fails to load. We already didn't have alt set on thumbnails. And a long unnecessary string saying the url is just useless for screen readers. Alt is supposed to describe what's on the image, but since we can't do that, we're better of not setting it.

astorije: What about an empty one then:

> Omitting this attribute altogether indicates that the image is a key part of the content, and no textual equivalent is available. Setting this attribute to an empty string (`alt=""`) indicates that this image is not a key part of the content, and that non-visual browsers may omit it from rendering.

I'm not entirely sure the practical difference between no `alt` and `alt=""` though.

xPaw: Done.

chaals: (@astorije **roughly** if there is no `alt` a screenreader will announce the image, and use whatever heuristics it has to try and help, e.g. providing the URL, or filename, or looking for something else. If there is `alt=""` it will happily just skip over the image altogether. YMMV because different screenreaders and different configurations of the same screenreader exist...).

 `alt=''` means "this image isn't actually adding anything necessary", whereas no `alt` means "this image doesn't have important accessibility information". Imagine you are telling someone by phone how to go through the page. When you get to the image, do you need to say anything? If not, `alt=""` is probably right).

astorije: Hey @chaals, thanks for chiming in and for the advice! And good to hear from you, long time no see 🙂

As a bit of context, images we display are link previews from URL people post to the chat. If the link is a webpage, there is a blurb with title + description + small thumbnail (if present in metadata). If the link is an image, we show a preview of that image.
Here is an example of how it looks:

screen shot 2018-02-17 at 19 35 00

When a screen reader would go through these images, they would have gone through the corresponding URLs beforehand.

Based on your description, I _think_ `alt=""` is correct, but I am not 100% sure. What do you think?

chaals: In a better world you would have a way for the URL to get a description, like the

`-> [URL] [some description]` syntax for RRSAgent. And in an ideal world you would meet ATAG and enable a screenshot description too. Although it is unlikely to get massive use.

For the image case, it would be really good to do this, and I would actually be inclined to leave alt out if  no description is provided, but for non-image URLs I think it is OK to have `alt=""`.

astorije: We have no control over the URLs though, because it's a general purpose chat client for IRC, so we have no control or say over how people post their URLs :/

PR #1925, opened 2017-12-29
Preload preview images before appending them to DOM

xPaw:


astorije:
Switching to the Bug label because, based on your inline comment explaining why this change, it seems to respond to an issue to me.

PR #1926, opened 2017-12-30
Store images with correct file extension

xPaw:
This fixes svg image thumbnails, where it would be stored as `xxxxxxx.svg+xml` on disk and express wouldn't serve correct content-type for it.

This also prevents being able to write any user-input string in the file name.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "irc-framework": "2.10.2",
     "ldapjs": "1.0.1",
     "lodash": "4.17.4",
+    "mime-types": "2.1.17",
     "moment": "2.20.1",
     "package-json": "4.0.1",
     "primer-tooltips": "1.5.1",
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -3,6 +3,7 @@
 const cheerio = require("cheerio");
 const request = require("request");
 const url = require("url");
+const mime = require("mime-types");
 const Helper = require("../../helper");
 const cleanIrcMessage = require("../../../client/js/libs/handlebars/ircmessageparser/cleanIrcMessage");
 const findLinks = require("../../../client/js/libs/handlebars/ircmessageparser/findLinks");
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
     "irc-framework": "2.10.2",
     "ldapjs": "1.0.1",
     "lodash": "4.17.4",
+    "mime-types": "2.1.17",
```
xPaw: We already had this under some sub dependency so nothing new.

src/plugins/irc-events/link.js
```diff
@@ -146,7 +147,23 @@ function handlePreview(client, msg, preview, res) {
...
+		if (preview.type === "link") {
+			preview.thumb = "";
+		} else {
+			return;
+		}
+
+		return emitPreview(client, msg, preview);
+	}
+
+	storage.store(res.data, extension, (uri) => {
```
astorije: Any chance you could add a test or 2 in https://github.com/thelounge/lounge/blob/master/test/plugins/link.js ?

src/plugins/irc-events/link.js
```diff
@@ -146,7 +147,23 @@ function handlePreview(client, msg, preview, res) {
...
+	if (!extension) {
+		// For link previews, drop the thumbnail
+		// For other types, do not display preview at all
+		if (preview.type === "link") {
+			preview.thumb = "";
+		} else {
+			return;
+		}
+
+		return emitPreview(client, msg, preview);
```
astorije: Move this in the `if` block, or do

```js
if (preview.type !== "link") {
	return;
}

preview.thumb = "";
return emitPreview(client, msg, preview);
```

test/plugins/storage.js
```diff
@@ -67,4 +74,23 @@ describe("Image storage", function() {
...
 		});
 	});
+
+	it("should lookup correct extension type", function(done) {
+		const message = this.irc.createMessage({
+			text: "http://localhost:9003/svg-preview",
+		});
+
+		this.app.get("/svg-preview", function(req, res) {
+			res.send("<title>test title</title><meta property='og:image' content='http://localhost:9003/logo.svg'>");
```
astorije: It's a detail but in tests that include HTML, I tend to invert the quotes, to always have double-quotes for HTML attributes. It doesn't matter here, but when testing rendering/parsing, this consistency is rather helpful.

PR #1932, opened 2018-01-02
Update README.md

Roang-zero1:
Fix wrong command name in example commands in the README.md.


xPaw:
It's not wrong, git version currently aliases both `thelounge` and `lounge`. v3.0 will be only `thelounge`.

See aca23257edbe5d6878e34b101aaafcc4519ee68f

PR #1933, opened 2018-01-03
Show channel name on channel-related errors

### LINKED ISSUE #1207: Channel join error info doesn't mention channel name
`Cannot join channel (+l) (channel_is_full)` - which channel? No idea, when this happens on reconnection.

RockyTV:
Fixes #1207.

Examples:
```
#rockytv: Cannot join channel (+b) - you are banned (banned_from_channel)
#rockytv: Cannot join channel (+l) - channel is full, try again later (channel_is_full)
#rockytv: Xinayder is already on channel (user_on_channel)
```

~~The only problem I've found is that if you try to join the channel again after receiving an error that prevents you from joining it, Lounge will just switch to the tab instead of sending the command again.~~

~~EDIT: regarding the bug related above, it seems to happen only on the 2.7 development branch.~~

### DIFF
--- a/src/plugins/irc-events/error.js
+++ b/src/plugins/irc-events/error.js
@@ -6,10 +6,18 @@ module.exports = function(irc, network) {
 	const client = this;
 
 	irc.on("irc error", function(data) {
-		let text = data.error;
+		let text = "";
 
-		if (data.reason) {
-			text = data.reason + " (" + text + ")";
+		if (data.channel) {
+			text = `${data.channel}: `;
+		}
+
+		if (data.error === "user_on_channel") {
+			text += `User (${data.nick}) is already on channel`;
+		} else if (data.reason) {
+			text += `${data.reason} (${data.error})`;
+		} else {
+			text += data.error;
 		}
 
 		const lobby = network.channels[0];


src/plugins/irc-events/error.js
```diff
@@ -10,6 +10,16 @@ module.exports = function(irc, network) {
 
 		if (data.reason) {
 			text = data.reason + " (" + text + ")";
+			if (data.channel) {
+				text = data.channel + ": " + text;
```
xPaw: Can you move `data.channel` outside of any other if so that `+=` can be used in text formatting.

src/plugins/irc-events/error.js
```diff
@@ -10,6 +10,16 @@ module.exports = function(irc, network) {
...
+			}
+		}
+
+		if (data.error === "user_on_channel") {
+			text = data.channel + ": User is already on channel (" + data.error + ")";
+			if (data.nick) {
+				text = data.channel + ": " + data.nick + " is already on channel (" + data.error + ")";
```
xPaw: Would be nicer to combine this into same message above, for example `User (${data.nick}) is...`.

You can use `+=` to cut text duplication.

RockyTV:
Just updated with the requested changes, not sure if it's in a suitable way, though.

src/plugins/irc-events/error.js
```diff
@@ -8,8 +8,14 @@ module.exports = function(irc, network) {
...
-		if (data.reason) {
-			text = data.reason + " (" + text + ")";
+		if (data.channel) {
+			text = `${data.channel}: `;
+		}
+
+		if (data.error === "user_on_channel") {
+			text += `User (${data.nick}) is already on channel`;
+		} else {
+			text += (`${data.reason} (${data.error})` || data.error);
```
xPaw: This fallback doesn't make particular sense, did you want to do something like `data.reason ? `${data.reason} (${data.error})` : data.error`?

src/plugins/irc-events/error.js
```diff
@@ -8,8 +8,14 @@ module.exports = function(irc, network) {
 	irc.on("irc error", function(data) {
 		let text = data.error;
```
xPaw: This probably needs to be an empty string, if there is no `data.channel`, the formatted error will look odd.

src/plugins/irc-events/error.js
RockyTV: Is it okay to replace it with `text += (data.reason ? `${data.reason} (${data.error})` : data.error);`?

xPaw: `else if` ends up cleaner here, imo.

PR #1938, opened 2018-01-05
Add a `thelounge uninstall` command to remove themes and packages

astorije:
Triggered by https://github.com/thelounge/lounge/pull/1619.

Because it currently forces usage of plugins and there is no way to disable them from the server or the client, I think having a way to remove packages should make it in the same release as #1619 (currently v2.7).

### DIFF
--- a/src/command-line/uninstall.js
+++ b/src/command-line/uninstall.js
@@ -1,0 +1,53 @@
+"use strict";
+
+const colors = require("colors/safe");
+const path = require("path");
+const program = require("commander");
+const Helper = require("../helper");
+const Utils = require("./utils");
+
+program
+	.command("uninstall <package>")
+	.description("Uninstall a theme or a package")
+	.on("--help", Utils.extraHelp)
+	.action(function(packageName) {
+		const fs = require("fs");
+		const child = require("child_process");
+
+		if (!fs.existsSync(Helper.getConfigPath())) {
+			log.error(`${Helper.getConfigPath()} does not exist.`);
+			return;
+		}
+
+		log.info(`Uninstalling ${colors.green(packageName)}...`);
+
+		const packagesPath = Helper.getPackagesPath();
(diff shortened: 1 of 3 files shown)


astorije:
I added an extra commit to slightly improve output of `install` and `uninstall`.

src/command-line/uninstall.js
```diff
@@ -0,0 +1,63 @@
...
+		const child = require("child_process");
+
+		if (!fs.existsSync(Helper.getConfigPath())) {
+			log.error(`${Helper.getConfigPath()} does not exist.`);
+			return;
+		}
+
+		log.info(`Uninstalling ${colors.green(packageName)}...`);
+
+		const packagesParent = Utils.preparePackagesDir();
```
xPaw: I don't think you need to prepare it in uninstall, just bail out if packages folder does not exist.

astorije: Done. It's overall a wee bit messier than what I would have hoped, but oh well. I've kept my original commit in case you preferred the first version in the end, so if you're happy with this don't merge as is, I'll have to squash the extra commit.

src/command-line/uninstall.js
```diff
@@ -0,0 +1,72 @@
...
+
+program
+	.command("uninstall <package>")
+	.description("Uninstall a theme or a package")
+	.on("--help", Utils.extraHelp)
+	.action(function(packageName) {
+		const fs = require("fs");
+		const child = require("child_process");
+
+		if (!fs.existsSync(Helper.getConfigPath())) {
```
YaManicKill: There's a lot of similar code to install here that could be abstracted out, but I have similar changes in an update command, so I'm happy to wait till that's in if you are.

astorije: Yeah let's have a cleanup-specific PR then, so that feature-PRs don't have to contain lots of not-directly-related changes. Easier to review and to check history later.

PR #1940, opened 2018-01-05
Hash user tokens, increase token entropy

### LINKED ISSUE #1934: Hash user tokens
1. Avoids leaking other tokens via the session list (disconnect button currently needs to know the token to function)
2. Hashed on disk (in user config) because we don't need to store actual token at this point.

Only the users' local storage will have original token stored to perform login. What algorithm should we use for this?

xPaw:
Fixes #1934.

- Increased token entropy from 48 to 64 bytes. This ends up being 128 byte string (same length as a sha512 token).
- Hash tokens with SHA512.

I also included support for converting tokens into hashed ones on server start, and this makes me thing we should include this in 2.7 so we can remove this compat layer in 3.0.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -94,6 +94,27 @@ function Client(manager, name, config) {
 
 	if (typeof client.config.sessions !== "object") {
 		client.config.sessions = {};
+	} else {
+		// TODO: This is just for backwards compatibility. Remove in v3.0.0
+		const newSessions = {};
+		let changed = false;
+
+		_.forOwn(client.config.sessions, (session, key) => {
+			if (key.length !== 128) {
+				key = client.calculateTokenHash(key);
+				changed = true;
+			}
+
+			newSessions[key] = session;
+		});
+
+		if (changed) {
+			log.info(`User ${colors.bold(client.name)} has been updated with new security requirements for tokens.`);
+
+			delete client.config.token;
+			client.config.sessions = newSessions;
+			client.save();
+		}
 	}
 
 	_.forOwn(client.config.sessions, (session) => {
@@ -282,13 +303,17 @@ Client.prototype.connect = function(args) {
 };
 
 Client.prototype.generateToken = function(callback) {
-	crypto.randomBytes(48, (err, buf) => {
+	crypto.randomBytes(64, (err, buf) => {
 		if (err) {
 			throw err;
 		}
 
 		callback(buf.toString("hex"));
 	});
+};
+
+Client.prototype.calculateTokenHash = function(token) {
+	return crypto.createHash("sha512").update(token).digest("hex");
 };
 
 Client.prototype.updateSession = function(token, ip, request) {
--- a/src/server.js
+++ b/src/server.js
@@ -373,7 +373,7 @@ function initializeClient(socket, client, token, lastMessage) {
 	});
 
 	socket.on("push:register", (subscription) => {
-		if (!client.isRegistered() || !client.config.sessions[token]) {
+		if (!client.isRegistered() || !client.config.sessions.hasOwnProperty(token)) {
 			return;
 		}
 
@@ -418,7 +418,7 @@ function initializeClient(socket, client, token, lastMessage) {
 			tokenToSignOut = token;
 		}
 
-		if (!(tokenToSignOut in client.config.sessions)) {
+		if (!client.config.sessions.hasOwnProperty(tokenToSignOut)) {
 			return;
 		}
 
@@ -459,11 +459,11 @@ function initializeClient(socket, client, token, lastMessage) {
 
 	if (!Helper.config.public && token === null) {
 		client.generateToken((newToken) => {
-			client.attachedClients[socket.id].token = token = newToken;
+			client.attachedClients[socket.id].token = token = client.calculateTokenHash(newToken);
 
 			client.updateSession(token, getClientIp(socket), socket.request);
 
-			sendInitEvent(token);
+			sendInitEvent(newToken);
 		});
 	} else {
 		sendInitEvent(null);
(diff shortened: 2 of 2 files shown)


YaManicKill:
> I also included support for converting tokens into hashed ones on server start, and this makes me thing we should include this in 2.7 so we can remove this compat layer in 3.0.

That feels like quite a big change to put into 2.7 at this point. Could we do a 2.8 before 3.0 if it's just that you want to not support non-hashed ones by 3.0?

xPaw:
@YaManicKill The plan was to do 3.0 after 2.7 because we already have enough PRs lined up for it.

I want to ship this before 3.0 specifically to keep backwards compatibility. And I don't want to have any backwards compat stuff left in 3.0 (from 2.x), so quite the dilemma.

YaManicKill:
> The plan was to do 3.0 after 2.7 because we already have enough PRs lined up for it.

Yeah, I know, but maybe this is a reason to not go straight to 3.0. 

The other option is we can build in a migration script that people have to run when upgrading to 3.0. That's not unheard of. Not ideal, but not unheard of. Could do it in an npm script on install.

src/client.js
```diff
@@ -94,6 +94,27 @@ function Client(manager, name, config) {
 
 	if (typeof client.config.sessions !== "object") {
 		client.config.sessions = {};
+	} else {
+		// TODO: This is just for backwards compatibility
```
astorije: Do you want to add a note saying "Remove this in v3" or something?

---

On an unrelated note that, we've kept our deprecated stuff for a while, because we knew v3 was coming soon-ish, but in the future, I have nothing against deprecating things for a couple versions of a few months and then removing the deprecated support as part of a minor version release.
We've mentioned this in the past, but we're not following semver (we've even removed that from our changelogs): we're an end-user project and not a library, so we're not typically installed in `package.json`s with a `~` range or something, and we document these things pretty well in the changelogs.

src/client.js
```diff
@@ -94,6 +94,27 @@ function Client(manager, name, config) {
...
+		const newSessions = {};
+		let changed = false;
+
+		_.forOwn(client.config.sessions, (session, key) => {
+			if (key.length !== 128) {
```
astorije: `>`? Any reason why this would be, I don't know, `null`, `undefined`, `0`, etc.?

astorije:
> (same length as a sha512 token)

Just out of curiosity, is there a specific reason for this?

> I also included support for converting tokens into hashed ones on server start, and this makes me thing we should include this in 2.7 so we can remove this compat layer in 3.0.

Unless @YaManicKill requests changes, I'm totally cool with having this in v2.7. It's rather straightforward, it's not adding any contentious feature or whatnot (i.e. it's a no-op from the user standpoint), it's ready, it works (tested), it's gonna be validated in at least one non-stable release... :man_shrugging:

I understand the reasons not too bring it to v2.7, but I don't think we should hold ourselves to a strict process. The feature freeze is not like enforced by a SOC 2 compliance, or sync required with other teams, etc., it's just so we don't keep postponing releases, and I don't think this PR would be.

I'm not strongly pushing btw, see this as me not caring where it goes :) 
If not v2.7, let's not make a v2.8 _just_ for this though, let's put it in v3.0 and remove the extra code in v3.1 or v3.2. After all, the worst that happens without this transition is users have to re-login after upgrading. nbd

> The other option is we can build in a migration script that people have to run when upgrading to 3.0. That's not unheard of. Not ideal, but not unheard of. Could do it in an npm script on install.

We might have to do that once we update DB schema after #1839 makes it in, but I think we should avoid doing that whenever possible, especially if it has to be a manual step. In this case, it's just a convenience of not having to re-login so I prefer @xPaw's solution.

Apart from that, good stuff in this PR! 👏

src/client.js
xPaw: Non hashed keys are 96 length (48 in binary), and hashed ones are 128 (96).

astorije: Right, and I suggested `if (key.length < 128)` (actually `>` but that was a typo lol), no big deal though, it doesn't make any difference I believe as long as hashed keys are _always_ 128 char long.

src/server.js
```diff
@@ -495,8 +495,9 @@ function getClientConfiguration() {
 function performAuthentication(data) {
 	const socket = this;
 	let client;
+	let token = null;
 
-	const finalInit = () => initializeClient(socket, client, data.token || null, data.lastMessage || -1);
+	const finalInit = () => initializeClient(socket, client, token, data.lastMessage || -1);
```
YaManicKill: Why are we not using data.token anymore?

src/client.js
```diff
@@ -94,6 +94,27 @@ function Client(manager, name, config) {
 
 	if (typeof client.config.sessions !== "object") {
 		client.config.sessions = {};
+	} else {
+		// TODO: This is just for backwards compatibility. Remove in v3.0.0
```
YaManicKill: If I'm reading this correctly, this means that once we delete it, anyone upgrading from 2.6 to whatever version doesn't have it in (3.0.0 currently), their auth won't work?

xPaw: They will simply have to re-login. We've already had this happen before (2.5 I think?)

My suggestion is to tell people to upgrade to `2.x` before going to `3.x` for existing users because we have quite a bit of deprecation warnings.

src/server.js
xPaw: Because it passes in the hashed token now. And only if the session exists for provided token.

src/client.js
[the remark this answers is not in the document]
xPaw: #1958. cba putting it in the comment, I will make sure to make a PR to remove this code after its merged.

astorije: I'll make sure to mention this in changelog(s). And yeah, it only means people will have to re-login, so that's a reasonable worst case solution :)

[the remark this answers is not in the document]
YaManicKill: Hmmmm, alright, I'll give you it this time :-P

YaManicKill: Ah right, yes, of course, this is the session tokens, not the user password. Minor mind-slip there.

src/server.js
YaManicKill: But it's not, it's passing in `null` everytime here. What even is the point in passing in null?

xPaw: It's set lower in the code before this function is called. Kinda the same way `client` is assigned.

YaManicKill: Oh I  completely missed that this was a function, duh. Alright, yup.

PR #1944, opened 2018-01-06
Pass --ignore-scripts to npm install

### LINKED ISSUE #1935: Pass --ignore-scripts to npm install in lounge install command

xPaw:
Fixes #1935.

See http://blog.npmjs.org/post/141702881055/package-install-scripts-vulnerability for more information. We don't need to run scripts for packages.

### DIFF
--- a/src/command-line/install.js
+++ b/src/command-line/install.js
@@ -52,6 +52,7 @@ program
 				[
 					"install",
 					"--production",
+					"--ignore-scripts",
 					"--no-save",
 					"--no-bin-links",
 					"--no-package-lock",


astorije:
When we start having plugins like LDAP that would require running `postinstall` script stuff, we would need to re-evaluate this, correct?

astorije:
@xPaw, I believe this gets superseded by #1987, no? Unless you want to ship this earlier or something

astorije:
FYI, if we want to ship this in v2.7.0, since it's security related, I'm fine with this. It won't delay release.

xPaw:
@astorije It does get superseded. I'll close this in favor of yarn pr.

PR #1948, opened 2018-01-07
Add a CLI option to force colors on command outputs

astorije:
Not sure if we want this, I was trying to help with https://github.com/thelounge/lounge/pull/1943 and the fact that our colors don't show when running piped.

This is similar to what Mocha, ESLint, Stylelint, etc. have to force colors when displaying their output via piping instead of TTY (for example when using `npm-run-all --print-label` or `npm-run-all --aggregate-output`).

I hesitated to make this part of the config file, then use `-c forceColors=true` but it's less obvious than other options as it's specific to CLI, probably a per-command basis and not per configuration i.e. probably something you never want to enable in config file since it can mess up logs, etc.

### DIFF
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -11,15 +11,21 @@ const Helper = require("../helper");
 const Utils = require("./utils");
 
 program.version(Helper.getVersion(), "-v, --version")
+	.option("--colors", "force enabling of colors in command outputs")
 	.option(
 		"-c, --config <key=value>",
-		"override entries of the configuration file, must be specified for each entry that needs to be overriden",
+		"override entries of the configuration file, must be specified for each " +
+		"entry that needs to be overriden",
 		Utils.parseConfigOptions
 	)
 	.on("--help", Utils.extraHelp);
 
 // Parse options from `argv` returning `argv` void of these options.
 const argvWithoutOptions = program.parseOptions(process.argv);
+
+if (program.colors) {
+	colors.enabled = true;
+}
 
 // Check if the app was built before calling setHome as it wants to load manifest.json from the public folder
 if (!fs.existsSync(path.join(
@@ -46,9 +52,9 @@ if (!Helper.config.public && !Helper.config.ldap.enable) {
 require("./install");
 require("./uninstall");
 
-// `parse` expects to be passed `process.argv`, but we need to remove to give it
-// a version of `argv` that does not contain options already parsed by
-// `parseOptions` above.
+// `parse` expects to be passed `process.argv`, but we need to give it a version
+// of `argv` that does not contain options already parsed by `parseOptions`
+// above.
 // This is done by giving it the updated `argv` that `parseOptions` returned,
 // except it returns an object with `args`/`unknown`, so we need to concat them.
 // See https://github.com/tj/commander.js/blob/fefda77f463292/index.js#L686-L763


src/command-line/index.js
```diff
@@ -73,9 +78,9 @@ if (process.argv[1].endsWith(`${require("path").sep}lounge`)) {
 	process.argv[1] = "thelounge";
 }
 
-// `parse` expects to be passed `process.argv`, but we need to remove to give it
-// a version of `argv` that does not contain options already parsed by
-// `parseOptions` above.
+// `parse` expects to be passed `process.argv`, but we need to give it a version
+// of `argv` that does not contain options already parsed by `parseOptions`
+// above.
```
astorije: This is just fixing my shitty English (rather, write a sentence, re-write it, re-write it again, etc.).

src/command-line/index.js
```diff
@@ -27,6 +28,10 @@ program.version(Helper.getVersion(), "-v, --version")
 // Parse options from `argv` returning `argv` void of these options.
 const argvWithoutOptions = program.parseOptions(process.argv);
 
+if (program.colors) {
+	colors.enabled = true;
```
YaManicKill: Is there any need to add this as an option? When would we not want colours? And if there is a situation, surely the default should be colours, and the option should be `--no-colors`?

xPaw: `isTTY` being false would disable the colors.

xPaw:
I just looked at `colors` lib, and looks like it parses arguments on it's own, so we do not need to add this ourselves: https://github.com/Marak/colors.js/blob/master/lib/system/supports-colors.js

So just passing `--color` would work, but Lounge quits because it thinks it's an unknown argument. So just adding that as an option without actually setting in our code is enough to make it work.

There's also `--no-color` and `--color=false`.

astorije:
@YaManicKill, colors are enabled by default, but not _forced_ by default (it's the same for all the CLIs we use for example, but most CLIs really).
The idea is that if it's TTY (screen), default shows color, but if not, it doesn't. Non-TTY is typically log files, so systems that output disable colors/bold/etc. on non-TTY.
Rationale is that an output _on screen without colors_ is annoying, but an output _in a file with color characters_ is just garbage.
Issue is when you pipe outputs (such as prefixed or aggregated outputs in our `package.json`), it's not TTY anymore, and systems don't know if it's in a file or not, so to not take any chances they disable, and one needs to force.

This is why there needs to be an option to force enabling colors when it's not on.
Does that make sense?

@xPaw, so the only addition of this PR would be:

```js
.option("--color", "force enabling of colors in command outputs")
```

and no `if (options.color) { [...]` (minor nitpick, it would **have** to be `--color` and not `--color` 😕).
So basically, we declare it, but implicitly let the underlying library take care of forcing. Are we okay with that? I think I'd rather have it set explicitly, considering it's a one-liner, just to be safe. No strong opinions though, just my preference.

Also, `--no-color`, I wouldn't even add it unless we have a use case. It's just... not useful lol. Some CLIs don't even have that parity, just `--color`. No strong opinions either, your call.

YaManicKill:
> Does that make sense?

Yeah, that makes sense. I understand now.

> minor nitpick, it would have to be --color and not --color

Am I missing the difference here?

> Also, --no-color, I wouldn't even add it unless we have a use case. It's just... not useful lol. Some CLIs don't even have that parity, just --color. No strong opinions either, your call.

Alright, I would say that if there is no usecase for `--no-color`, then there's no reason to add this as a flag. I say just set colour always, and leave it as that.

astorije:
> Am I missing the difference here?

Ahah, yes, I meant that it would force us to use `--color` (singular) instead of `--colors` (plural). No big deal, just worth noting.

xPaw:
Can we make use of no colours in tests? It's currently stripped with regex.

astorije:
Do you have an example of that? I'm looking at https://travis-ci.org/thelounge/lounge/jobs/343684047#L2579 and not seeing anything, so I'm thinking you might not be talking about this.

xPaw:
Package tests that you recently added, they test for console output

astorije:
Ah, duh, [this](https://github.com/thelounge/lounge/blob/7975f4debc4c473ad034ce4cd9476d11c65e485b/test/util.js#L26-L36). Yeah, that would probably work, though we may want to keep this regex for good measure?
Also, looking at the test, this is [triggered here](https://github.com/thelounge/lounge/blob/e4701be7081e97a6a4e243f104e412be8a0a2c58/test/plugins/packages/indexTest.js#L56), not by the CLI, so we couldn't run `--no-colors` here (and we don't want to do that at Mocha level or it would get rid of **all colors**).

xPaw:
Yeah we can probably set `colors.enabled = false`; somewhere in test code directly. That was just the thought I came up with when you asked. So yeah close this PR.

astorije:
Ok. We can always bring it back should we ever need to.

PR #1949, opened 2018-01-07
Only emit "more" history to the client that requested it

xPaw:
Noticed that when opening a channel on mobile also sent `more` to my desktop client.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -258,7 +258,11 @@ function initializeClient(socket, client, token, lastMessage) {
 	socket.on(
 		"more",
 		function(data) {
-			client.more(data);
+			const history = client.more(data);
+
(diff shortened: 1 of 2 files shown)


src/server.js
```diff
@@ -258,7 +258,11 @@ function initializeClient(socket, client, token, lastMessage) {
 	socket.on(
 		"more",
 		function(data) {
-			client.more(data);
+			const history = client.more(data);
+
+			if (history !== null) {
```
astorije: You could have left `return;` above and here do `if (history)` or `if (history !== undefined)`, but that's pretty much the same thing and I get the preference for being explicit.

astorije:
Bugfix, merging it in v2.7.0.

PR #1950, opened 2018-01-08
User Admin role

astorije:
First and only feature of this is to only display the version checker to admin:

screen shot 2018-01-07 at 23 10 33

Ideally this should instead be in an "Admin" window (coming up next), which I'm going to add to list and manage TL users from the UI.

Also changes wording in CLI outputs, such as:

screen shot 2018-01-07 at 23 18 34

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -240,6 +240,7 @@ kbd {
 #chat .action .from::before,
 #chat .toggle-button::after,
 #version-checker::before,
+#version-checker .admin-only::after,
 .context-menu-item::before,
 #help .website-link::before,
 #help .documentation-link::before,
@@ -1659,9 +1660,18 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	padding-bottom: 6px;
 }
 
+#version-checker .admin-only-container {
+	margin-left: 10px;
+}
+
+#version-checker .admin-only::after {
+	font-size: 1.2em;
+	content: "\f132"; /* http://fontawesome.io/icon/shield/ */
+}
+
 #version-checker::before {
-	margin-left: 6px;
-	margin-right: 12px;
+	margin-left: 5px;
+	margin-right: 10px;
 	font-size: 1.2em;
 }
 
--- a/client/js/socket-events/init.js
+++ b/client/js/socket-events/init.js
@@ -12,6 +12,10 @@ const utils = require("../utils");
 socket.on("init", function(data) {
 	$("#loading-page-message, #connection-error").text("Rendering…");
 
+	// Store information about logged user in state, such as its role (admin or
+	// regular user)
+	$("body").data("current-user", data.currentUser);
+
 	const lastMessageId = utils.lastMessageId;
 	let previousActive = 0;
 
--- a/src/command-line/users/add.js
+++ b/src/command-line/users/add.js
@@ -32,33 +32,48 @@ program
 		log.prompt({
 			text: "Enter password:",
 			silent: true,
-		}, function(err, password) {
+		}, (err, password) => {
 			if (!password) {
 				log.error("Password cannot be empty.");
 				return;
 			}
-			if (!err) {
-				log.prompt({
-					text: "Save logs to disk?",
-					default: "yes",
-				}, function(err2, enableLog) {
-					if (!err2) {
-						add(
-							manager,
-							name,
-							password,
-							enableLog.charAt(0).toLowerCase() === "y"
-						);
-					}
-				});
+
+			if (err) {
+				return;
 			}
+
+			log.prompt({
+				text: "Save logs to disk?",
+				default: "yes",
+			}, (err2, enableLog) => {
+				if (err2) {
+					return;
+				}
+
+				log.prompt({
+					text: "Admin?",
+					default: "no",
+				}, (err3, isAdmin) => {
+					if (err3) {
+						return;
+					}
+
+					add(
+						manager,
+						name,
+						password,
+						isAdmin.charAt(0).toLowerCase() === "y",
+						enableLog.charAt(0).toLowerCase() === "y"
+					);
+				});
+			});
 		});
 	});
 
-function add(manager, name, password, enableLog) {
+function add(manager, name, password, isAdmin, enableLog) {
 	var hash = Helper.password.hash(password);
-	manager.addUser(name, hash, enableLog);
+	manager.addUser(name, hash, isAdmin ? "admin" : "", enableLog);
 
-	log.info(`User ${colors.bold(name)} created.`);
+	log.info(`${isAdmin ? "Admin" : "User"} ${colors.bold(name)} created.`);
 	log.info(`User file located at ${colors.green(Helper.getUserConfigPath(name))}.`);
 }
--- a/src/models/local-user.js
+++ b/src/models/local-user.js
@@ -1,0 +1,12 @@
+"use strict";
+
+module.exports = class LocalUser {
+	constructor(name, role) {
+		this.name = name;
+		this.role = role || "";
+	}
+
+	isAdmin() {
+		return this.role === "admin";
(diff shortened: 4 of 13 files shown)


src/models/local-user.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+module.exports = class LocalUser {
```
astorije: I thought that having a model for TL users was a good idea, but I'm not really sure.
It's useful for things like `isAdmin()` helpers, otherwise they have to live in `client.js` otherwise, already pretty bloated.
If this is a good pattern we want to keep, I can see more and more things using `LocalUser` over time, which will be a subset of `client.js`.
Feedback welcome.

client/js/socket-events/init.js
```diff
@@ -12,6 +12,10 @@ const utils = require("../utils");
 socket.on("init", function(data) {
 	$("#loading-page-message, #connection-error").text("Rendering…");
 
+	// Store information about logged user in state, such as its role (admin or
+	// regular user)
+	$("body").data("current-user", data.currentUser);
```
xPaw: `document.body` damn it.

src/command-line/users/add.js
```diff
@@ -32,33 +32,48 @@ program
...
 	});
 
-function add(manager, name, password, enableLog) {
+function add(manager, name, password, isAdmin, enableLog) {
 	var hash = Helper.password.hash(password);
-	manager.addUser(name, hash, enableLog);
+	manager.addUser(name, hash, isAdmin ? "admin" : "", enableLog);
 
-	log.info(`User ${colors.bold(name)} created.`);
+	log.info(`${isAdmin ? "Admin" : "User"} ${colors.bold(name)} created.`);
```
YaManicKill: There are a few times we get this, could we just use the role from the user object?

src/models/local-user.js
astorije: @xPaw, what's your take on this?

client/js/socket-events/init.js
astorije: > document.body damn it.

Goddamn, why do I always do that?! Fixing.

astorije: > Should we wrap the data.currentUser in our local-user object you've just created for the server.

I didn't want to commit too much into this `LocalUser` solution just yet, in case we actually don't want it. I just timidly brought the idea, and if we do that, I'm happy to use it more.

That said, this is tricky: it is actually sent as a `LocalUser` from the server. I'm guessing when it goes through the socket, it's converted to JSON or something. I'm not against re-wrapping it in a `LocalUser` once it hits the client, but that requires some sharing of models code between server and client.
At the moment, we just don't do that (I know you mentioned that in an earlier discussion). I agree, but if we want to do it cleanly (i.e. having a shared location, library-like or something), that requires some refactoring, or we'll end up having some `require("../src/[...]")` on the client, which would just add to the `require("../client/[...]")` 💩  we have on the server 😁 

So in short, totally in favor, but we need to be smart. Maybe for a future PR? I don't know, no strong opinion, I just don't want this PR to solve many issues, be hard to review, rebase, etc.

client/css/style.css
```diff
@@ -1659,9 +1660,18 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+
+#version-checker .admin-only::after {
+	font-size: 1.2em;
+	content: "\f132"; /* http://fontawesome.io/icon/shield/ */
+}
+
 #version-checker::before {
-	margin-left: 6px;
-	margin-right: 12px;
+	margin-left: 5px;
```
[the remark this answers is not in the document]
astorije: Ah, yes, if I remember correctly, the block padding was in multiples of 5 (5px and 10px) which actually looked obvious with the shield icon (but less so with the check icon), so just cleaning things up.

src/command-line/users/add.js
```diff
@@ -32,33 +32,48 @@ program
...
-			if (!err) {
+
+			if (err) {
+				return;
+			}
+
+			log.prompt({
+				text: "Save logs to disk?",
+				default: "yes",
+			}, (err2, enableLog) => {
```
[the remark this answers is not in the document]
astorije: Funny, I actually tried to refactor to promises in this PR, but since we need all intermediary results in the last promise, we need to... nest the promises lol. It's actually common to nest promises when you need intermediary results, so people get frustrated because they think it kills the point of promises that are meant to remove callback hell, [which isn't the point](https://blog.domenic.me/youre-missing-the-point-of-promises/) 😁

So yeah, I ended up with nested promises, and I figured it wasn't worth the effort in this particular PR.

src/command-line/users/add.js
astorije: If we go for the `LocalUser` route, we can surely add helpers in there for that. But since that'd require upper-casing the first letter in some cases, and not in others, we'd have to do  `user.friendlyRole({capitalize: true})` or something silly like that.
If not, we can still do that, but that adds a silly helper to `client.js` so meh :man_shrugging:

src/models/local-user.js
```diff
@@ -0,0 +1,12 @@
...
+	constructor(name, role) {
```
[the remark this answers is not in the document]
astorije: I was thinking about that, but the reason why I didn't is because there is a chain of things that could cause undetected harm: config contains passwords, and then user gets sent as it through the socket in the server code.
So yeah, we'd be ignoring it in the constructor, and we can even blacklist in the server code, but I can see how one day we might end up sending something to the client that shouldn't be sent to the client, and that would be horrific.

That being said, I'm just telling what I was thinking when doing that, I'm not against improving this, again, this was very early thought and I didn't want to invest time if it was getting trashed, but I'm completely happy testing stuff to reduce the future pain without making security bugs easy to make.

client/js/socket-events/init.js
YaManicKill: > I didn't want to commit too much into this LocalUser solution just yet, in case we actually don't want it. 

Yeah, that's fair. But I think we are probably going to go with it.

> or we'll end up having some require("../src/[...]") on the client

I agree it's less than ideal, but surely there's nothing wrong with just going for this in this PR, it'll be really easy to do for you just now, and then we can have another PR to refactor the common code into a different directory or another way of doing it.  I just see that we are doing the same thing in the client as we do in the localuser (to tell if the user is admin or not) and it would be nice if we were to start off without that.

src/command-line/users/add.js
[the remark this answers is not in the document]
YaManicKill: `Promise.all` would solve your problem there. But sure, not in this PR.

src/models/local-user.js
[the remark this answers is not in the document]
YaManicKill: > I was thinking about that, but the reason why I didn't is because there is a chain of things that could cause undetected harm

Right, sure, but we are currently just getting the stuff out of the config before this constructur, yeah. Surely it would be just as easy to do that in here. I am definitely not suggesting storing the entire config in the user (that would be bad) but the constructor takes in the config, we destructure out what we want from the config, and then we store *only those bits*. It's essentially what we are doing now, but it seems clearner to me (otherwise we will end up with this having many, many things in the constructor).

We still have to explicitly take bits out of the config, so there isn't a risk to accidentally adding something that we don't want to. Does that make sense?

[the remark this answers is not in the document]
astorije: Yeah it does, and I think I have reached to the same conclusion by the time I finished my previous message :)
I still have small concerns, but I'll play with this and see if actually doing this address them, which is rather likely. I'll change, commit, and we'll probably decide that it's fine.

astorije:
Sooo I still want this, but there are a bunch of conflicts, and other priorities for now. I'll close this PR / delete the branch to make way for the Vue transition, but I'll get back to this after we introduced Vue. Sorry and stay tuned!

Issue #1614, opened 2017-10-04
Await existing link expander promises


xPaw:
Link previewer calls `fetch` for every url in every message, it may happen that one url is posted multiple times (in different channels), and before the response arrives for the first request, there is no real need to fire another request, something like
```
if (requests[link]) {
    requests[link].then()
}
```

https://github.com/thelounge/lounge/blob/f18d0193837cac829c26cfc90361752de91f9164/src/plugins/irc-events/link.js#L38-L46

tkulpa:
Hey i have question, where information about currently active requests that are not yet fullfiled should be stored in app? Is client object valid option?

xPaw:
It can be an object within the same file.

PR #1952, opened 2018-01-09
Queue fetching previews so each url is only fetched once

crosma:
This solves issue #1614. Rather than fetching a url once per recipient, put each client and msg into an array to be processed after the attempt to fetch completes.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -11,6 +11,8 @@ const storage = require("../storage");
 
 process.setMaxListeners(0);
 
+const fetchRecipients = {};
+
 module.exports = function(client, chan, msg) {
 	if (!Helper.config.prefetch) {
 		return;
@@ -38,17 +40,28 @@ module.exports = function(client, chan, msg) {
 	})).slice(0, 5); // Only preview the first 5 URLs in message to avoid abuse
 
 	msg.previews.forEach((preview) => {
-		fetch(preview.link, function(res) {
-			if (res === null) {
-				return;
-			}
+		const recipient = {
+			msg: msg,
+			client: client,
+		};
 
-			parse(msg, preview, res, client);
-		});
+		if (!fetchRecipients[preview.link]) {
+			fetchRecipients[preview.link] = [recipient];
+
+			fetch(preview.link, function(res) {
+				if (res !== null) {
+					parse(preview, res, fetchRecipients[preview.link]);
+				}
+
+				delete fetchRecipients[preview.link];
+			});
+		} else {
+			fetchRecipients[preview.link].push(recipient);
+		}
 	});
 };
 
-function parse(msg, preview, res, client) {
+function parse(preview, res, recipients) {
 	switch (res.type) {
 	case "text/html":
 		var $ = cheerio.load(res.data);
@@ -85,7 +98,7 @@ function parse(msg, preview, res, client) {
 					preview.thumb = "";
 				}
 
-				handlePreview(client, msg, preview, resThumb);
+				handlePreview(preview, resThumb, recipients);
 			});
 
 			return;
@@ -139,12 +152,12 @@ function parse(msg, preview, res, client) {
 		return;
 	}
 
-	handlePreview(client, msg, preview, res);
+	handlePreview(preview, res, recipients);
 }
 
-function handlePreview(client, msg, preview, res) {
+function handlePreview(preview, res, recipients) {
 	if (!preview.thumb.length || !Helper.config.prefetchStorage) {
-		return emitPreview(client, msg, preview);
+		return emitPreview(preview, recipients);
 	}
 
 	// Get the correct file extension for the provided content-type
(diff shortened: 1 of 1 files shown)


crosma:
Looks like I missed a bug. Can ditch this PR.

crosma:
Found a bug. Will have an update soon.

src/plugins/irc-events/link.js
```diff
@@ -38,17 +40,32 @@ module.exports = function(client, chan, msg) {
...
-				return;
-			}
+		if (!fetch_recipients[preview.link]) {
+			fetch_recipients[preview.link] = [{
+				msg: msg,
+				client: client,
+			}];
+
+			fetch(preview.link, function(res) {
+				if (res === null) {
```
xPaw: You can do not null check and call `parse` inside.

src/plugins/irc-events/link.js
```diff
@@ -11,6 +11,8 @@ const storage = require("../storage");
 
 process.setMaxListeners(0);
 
+const fetch_recipients = {};
```
xPaw: Camel case please.

src/plugins/irc-events/link.js
```diff
@@ -38,17 +40,32 @@ module.exports = function(client, chan, msg) {
 	})).slice(0, 5); // Only preview the first 5 URLs in message to avoid abuse
 
 	msg.previews.forEach((preview) => {
-		fetch(preview.link, function(res) {
-			if (res === null) {
-				return;
-			}
+		if (!fetch_recipients[preview.link]) {
+			fetch_recipients[preview.link] = [{
```
xPaw: Create the object outside of the `if` so it's not defined in two places.

xPaw:
@crosma I like the idea over all, thanks for doing this! I have a question on how would this handle with 200 IRC connections, I imagine there might be enough latency for the link request to be executed multiple times if the link gets fetched before all IRC connections get the link parsed. Would you be able to test this out?

Also, could you confirm that the preview correctly get stored if you reload the page?

crosma:
Currently there is no caching of the results for these previews. It is definitely the case that the request is repeated if a client requests the link after it has been processed, including both because of latency or a page refresh. Or if the link is shared multiple times.  Also the thumbnail would be written to disk each time.

crosma:
I created a new issue to discuss the caching issue: #1957

xPaw:
Page refresh definitely shouldn't re-fetch the link, I was asking if the info stays visible between page refreshes (`preview` gets updated on the `msg` object).

crosma:
Hey, I was wrong about that, it doesn't fetch links when refreshed.

xPaw:
@crosma Hey, could you look into rebasing this? Master was changed to fetch urls and respect user language to provide more correct previews on services such as Twitter.

This should be changed to group requests per url+language now. I'm thinking language should be avoided in public mode to increase cache hit.

astorije:
> I'm thinking language should be avoided in public mode to increase cache hit.

Are you sure it's a good idea? I feel like public instances are directed at broader audiences than private instances, so that could be pretty bad 😅
At the very least if we don't want to want to cache url+language pairs, we should force language to be English to avoid surprises, but tbh I don't think we need that extra optimization, as it gets in the way of usability.

xPaw:
> Are you sure it's a good idea? I feel like public instances are directed at broader audiences than private instances, so that could be pretty bad
> At the very least if we don't want to want to cache url+language pairs, we should force language to be English to avoid surprises, but tbh I don't think we need that extra optimization, as it gets in the way of usability.

That's what I'm saying, basically use the same behaviour it was before.

astorije:
Not exactly, because what we had before (well, now on `master`) would cache the first language that previews, so you may end up with results in Finnish even though neither your client or server is fetching from there.

But really what I'm saying is... language+url pair is fine for both modes.

astorije:
But why go through the extra pain though? It's not going to burn a server, and we're looking at, what, 15-20 languages in the worst, worst case scenario (like, I'd be surprised if we have more than 5 users with that many languages/users, and if that's the case, they have enough resources on their server/network to deal with it).

It just doesn't make sense to me to over-optimize like this, especially if that ends up being more code (even an extra `if` is something we'll break someday lol) and reduces usability 🤷‍♂️

xPaw:
Multiply that by multiple languages per user (that special allows for percentages per languages and other crazy stuff) and browsers (can have minor differences). The point is to protect server and website owners (it's trivial to abuse if wanted, but luckily nothing stupid has happened yet).

astorije:
Hey @crosma, do you think you'd be able to get back to this PR so we can merge it? :)

astorije:
@xPaw, do you think you'd like to take this over, or should we close it until @crosma stops by this project again?

xPaw:
@astorije Don't have interest in taking this over yet.

astorije:
Fair enough. @crosma, let us know if you're interested again and we'll be happy to help you get this merged :)

PR #1954, opened 2018-01-09
Abstract common code between install and uninstall

YaManicKill:
I got so far and then I started seeing things, I have no idea if this actually makes any sense anymore. Good luck.

### DIFF
--- a/src/plugins/packageManager.js
+++ b/src/plugins/packageManager.js
@@ -46,6 +46,7 @@ function getMetadata(json) {
 	}
 	const metadata = json.thelounge;
 	metadata.version = json.version;
+	return metadata;
 }
 
 function execNpm(command, packageName, metadata) {
@@ -92,9 +93,9 @@ function runNpmCommand(command, packageName, metadata) {
 
 function install(packageName) {
 	checkConfig()
-		.then(getPackageJson.bind(this, packageName))
-		.then(getMetadata)
-		.then(runNpmCommand.bind(this, "install", packageName))
+		.then(() => getPackageJson(packageName))
+		.then((json) => getMetadata(json))
+		.then((metadata) => runNpmCommand("install", packageName, metadata))
 		.then(() => log.info(`${colors.green(packageName)} has been successfully installed.`))
 		.catch((e) => {
 			log.error(`${e}`);
@@ -104,7 +105,7 @@ function install(packageName) {
 
 function uninstall(packageName) {
 	checkConfig()
-		.then(runNpmCommand.bind(this, "uninstall", packageName))
+		.then((metadata) => runNpmCommand("uninstall", packageName, metadata))
 		.then(() => log.info(`${colors.green(packageName)} has been successfully uninstalled.`))
 		.catch((e) => {
 			log.error(`${e}`);
--- a/src/plugins/packageManager.js
+++ b/src/plugins/packageManager.js
@@ -56,15 +56,14 @@ function install(packageName) {
 
 function uninstall(packageName) {
 	checkConfig()
-		.then(() => npm.runNpmCommand("uninstall", {packageName, returnStdOut: true}))
+		.then(() => npm.runNpmCommand("list", {packageName, returnStdOut: true, resolveOnError: true}))
 		.then((data) => {
-			if (data.includes("removed")) {
-				log.info(`${colors.green(packageName)} has been successfully uninstalled.`);
-			} else {
-				log.error(`Can't uninstall ${colors.green(packageName)} as it wasn't installed.`);
-				process.exit(1);
+			if (!data.toString().includes(packageName)) {
+				return Promise.reject(`Can't uninstall ${colors.green(packageName)} as it wasn't installed.`);
 			}
 		})
+		.then(() => npm.runNpmCommand("uninstall", {packageName}))
+		.then(() => log.info(`${colors.green(packageName)} has been successfully uninstalled.`))
 		.catch((e) => {
 			log.error(`${e}`);
 			process.exit(1);
--- a/src/plugins/packageManager.js
+++ b/src/plugins/packageManager.js
@@ -2,53 +2,45 @@
 const Helper = require("../helper");
 const colors = require("colors/safe");
 const log = require("../log.js");
+const fs = require("fs-extra");
+const path = require("path");
+const child = require("child_process");
+const packageJson = require("package-json");
+const packagesPath = Helper.getPackagesPath();
+const packagesParent = path.dirname(packagesPath);
+const packagesConfig = path.join(packagesParent, "package.json");
 
 module.exports = {
 	install,
+	uninstall,
 };
 
-function install(packageName) {
-	const fs = require("fs");
-	const fsextra = require("fs-extra");
-	const path = require("path");
-	const child = require("child_process");
-	const packageJson = require("package-json");
+function checkConfig() {
+	return fs.access(Helper.getConfigPath(), fs.constants.R_OK | fs.constants.W_OK)
+		.catch(() => Promise.reject(`${Helper.getConfigPath()} does not exist or is not readable.`));
+}
 
-	if (!fs.existsSync(Helper.getConfigPath())) {
-		log.error(`${Helper.getConfigPath()} does not exist.`);
-		return;
-	}
-
+function getPackageJson(packageName) {
 	log.info("Retrieving information about the package...");
-
-	packageJson(packageName, {
+	return packageJson(packageName, {
 		fullMetadata: true,
-	}).then((json) => {
-		if (!("thelounge" in json)) {
-			log.error(`${colors.red(packageName)} does not have The Lounge metadata.`);
+	});
+}
 
-			process.exit(1);
-		}
+function getMetadata(json) {
+	if (!("thelounge" in json)) {
+		return Promise.reject(`${colors.red(json.name)} does not have The Lounge metadata.`);
+	}
+	const metadata = json.thelounge;
+	metadata.version = json.version;
+}
 
-		log.info(`Installing ${colors.green(packageName)}...`);
-
-		const packagesPath = Helper.getPackagesPath();
-		const packagesParent = path.dirname(packagesPath);
-		const packagesConfig = path.join(packagesParent, "package.json");
-
-		// Create node_modules folder, otherwise npm will start walking upwards to find one
-		fsextra.ensureDirSync(packagesPath);
-
-		// Create package.json with private set to true to avoid npm warnings
-		fs.writeFileSync(packagesConfig, JSON.stringify({
-			private: true,
-			description: "Packages for The Lounge. All packages in node_modules directory will be automatically loaded.",
-		}, null, "\t"));
-
+function execNpm(command, packageName, metadata) {
+	return new Promise((res, rej) => {
 		const npm = child.spawn(
 			process.platform === "win32" ? "npm.cmd" : "npm",
 			[
-				"install",
+				command,
 				"--production",
 				"--no-save",
 				"--no-bin-links",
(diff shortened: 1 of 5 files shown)


src/plugins/packageManager.js
```diff
@@ -0,0 +1,113 @@
...
+const colors = require("colors/safe");
+const log = require("../log.js");
+const fs = require("fs-extra");
+const path = require("path");
+const child = require("child_process");
+const packageJson = require("package-json");
+const packagesPath = Helper.getPackagesPath();
+const packagesParent = path.dirname(packagesPath);
+const packagesConfig = path.join(packagesParent, "package.json");
+const packagesNpmrc = path.join(packagesParent, ".npmrc");
```
astorije: What problem is this solving?

I can see a few downsides of doing this, so I want to make sure I understand the decision completely before having an opinion :)

For example, I don't really like the idea that commands our code runs relies on arguments that are stored in userland (the configuration folder). Just the idea that if we want to change those, we have to migrate the saved file and keep the migration code for a few versions, or tell people to change it themselves is... not great (again, assuming I'm understanding things correctly, that might change after your answer 😁).

src/plugins/packageManager.js
```diff
@@ -0,0 +1,113 @@
...
+
+	return fs.ensureDir(packagesPath) // Create node_modules folder, otherwise npm will start walking upwards to find one
+		.then(() => fs.writeJson(packagesConfig, packageDirJson, {spaces: "\t"})) // Create package.json with private set to true to avoid npm warnings
+		.then(() => fs.writeFile(packagesNpmrc, buildNpmrcString()))
+		.then(execNpm.bind(this, command, packageName, metadata));
+}
+
+function install(packageName) {
+	checkConfig()
+		.then(getPackageJson.bind(this, packageName))
```
astorije: I notice you're using the `thing.bind(this, arg)` syntax in several places in #1619 and this PR. Any reason to that?
I know it's perfectly valid JS, but I think it's a confusing way to call a function, we should just be doing `thing(arg)` in most places.

src/plugins/packageManager.js
```diff
@@ -0,0 +1,113 @@
...
+		.then(() => log.info(`${colors.green(packageName)} has been successfully installed.`))
+		.catch((e) => {
+			log.error(`${e}`);
+			process.exit(1);
+		});
+}
+
+function uninstall(packageName) {
+	checkConfig()
+		.then(runNpmCommand.bind(this, "uninstall", packageName))
```
astorije: Because of that shared `npmrc`, `uninstall` now uses options that are for `install` only. I know npm ignores the extra parameters, but it might not do that forever (or it might not have done that for all the versions we support). 

Also, it's kinda risky because if we add other `npm` calls, it will implicitly add them, and we might not want to. I don't really like the implicit-behind-the-scenes of the rc file, it makes things too magic IMO.

src/plugins/packageManager.js
```diff
@@ -0,0 +1,113 @@
...
+		.then(() => log.info(`${colors.green(packageName)} has been successfully uninstalled.`))
+		.catch((e) => {
+			log.error(`${e}`);
+			process.exit(1);
+		});
```
astorije: This currently loses some functionality: if the package was not installed, it now says "... has been successfully uninstalled". I know this is normal npm behavior, but I added that on purpose, it's good to know that a package was actually not uninstalled, which is why I had the warning.

Also, it now creates the `package.json` file and package folder when running uninstall without having installed anything. I don't have anything against that, but that's something @xPaw [made me change](https://github.com/thelounge/lounge/pull/1938#discussion_r159850803) 😁 Any reason why we're going back to that? Just for code convenience?

astorije:
I would honestly not abstract `install` and `uninstall` together. They're 2 different commands doing 2 different things. Things like config file checking, or other "environment-related" things, sure, but that's about it.
Mixing them together because they look alike though in their early age means that we're going to have troubles when we want to add functionality, like say, checking for things before/after installing, or letting users install specific versions, etc. We're going to have to un-abstract things, so I would not DRY them up too much as of now.

But as said in comments, maybe I missed some justifications, so let me know :)

src/plugins/packageManager.js
YaManicKill: Yeah, that's a fair problem. Thankfully, @xPaw found that you can pass in `--userconfig` with a path to a file and that is used as an npmrc file. So we can keep it in thelounge app folder, and keep it up to date that way. That'll be better.

The reason I'm doing this is so we aren't passing the individual arguments to commands that don't use it. I'll explain that later on because we need to discuss that in a larger way as to why we are doing that.

src/plugins/packageManager.js
YaManicKill: Yeah, I've been getting into a bad habit of doing `getPackageJson.bind(this, packageName)` instead of `() => getPackageJson(packageName)` just because it's shorter in some cases. It's a bad habit, I'll switch.

src/plugins/packageManager.js
YaManicKill: Good catch, I shall add a check to see if the plugin is installed before trying to uninstall it.

YaManicKill:
Right, so ... I think there are 2 discussions to have here (beside the other 2 points that I'll change, they are totally legit and I'll fix them).

1. Should we remove the massive amounts of duplicate code between the 2 commands or not?
2. How should we pass the arguments to npm.

The answer to 1, IMO, is absolutely we should. 

> I would honestly not abstract install and uninstall together. They're 2 different commands doing 2 different things.

Right, sure, but they are mostly the same. Abstracting code isn't about making 2 functions that are exactly the same into 1, it's about merging code that is *mostly* the same, but making it different only where it is. The vast majority of these 2 functions are the same, but they aren't *exactly* the same, so we merge the similar code, and make it work differently depending on what we want.

> Mixing them together because they look alike though in their early age means that we're going to have troubles when we want to add functionality, like say, checking for things before/after installing, or letting users install specific versions

There are a number of things here.

Adding new functionality to them is really easy. For example, I'm just about to add a check to the uninstall that it'll check to see if it's installed already or not. You just add a function, throw it into the promise chain, and that's it. It's so much easier this way, because you are removing all the boiler plate stuff away and it's very  clear where it needs to happen. 

Users can already install specific versions. `thelounge install thelounge-theme-solarized@1.0.12` will specifically install 1.0.12. But even if we want to change how that works (I would argue we don't, but that's for a different discussion) then doing this isn't going to stop that at all. We might have to change the 1 npm function into, where 1 supports a version and one doesn't, but that's *really not hard*. 

> We're going to have to un-abstract things, so I would not DRY them up too much as of now.

This is a really silly argument, IMO. "We might have to change how this code is, so we may as well not make it nicer". I don't get this at all. All the boiler plate around spawning the npm  command...that is all the same and I don't know why you'd want to not combine that? Even if you want to pass in the arguments separately (I understand that, let's argue about that separately) why would you want the actual spawning code twice? 

Right, now, putting aside the discussion of whether we are going to use this or not, let's talk about passing the arguments to npm.

We have 3 options:

1. Using an npmrc file
2. Passing the same CLI arguments to all of them 
3. Passing the CLI arguments into the npm function that we have

I initially did number 2, which would work, but it obviously isn't ideal. If you do this, currently npm will just ignore it, and there is a possibility that they would change what it means. I mean, I'm not sure what any of the install only arguments would *mean* in an uninstall, but that's beside the point. We are passing in arguments on the CLI that we won't use, and that's not great.

So, I then did number 1. The reason behind this is this is the way that npm suggests you use for config that isn't going to change. The npmrc will always ignore commands that do nothing in uninstall because that's what it is designed to do. Even if they do decide to make some of the commands mean something, npm are pretty good at semver. The few times they have changed things, they have given us 1 major release with it deprecated, and then the next major release with it changed. So we would have at least a year to change after they announced it. If that ever happens (which I don't see happening, because none of these would make any sense to do something in an uninstall) then we can deal with it then. The other advantage is this keeps our config around npm in 1 place, not having to have it strewn across the code.

So, number 3 is the other option. It's an option, but it'll mean duplication of some of the commands. I'm not a fan of this, but if you really don't like the other 2 options, then I won't fight it too much. It feels unnecessary when it doesn't change how the code works at all, and we will just be adding more code without needing to.

src/plugins/packageManager.js
YaManicKill: Hmmm, so I was having troubles actually making this work the same way you did it. It turns out, yours doesn't work. npm still outputs something even when it didn't uninstall. So I'm now checking against the output to see whether we uninstalled or not. When not uninstalling it outputs "up to date" and when uninstalling successfully it outputs "removed 1 package". Did you manage to get this working locally, is there perhaps some setting you forgot to push?

> Also, it now creates the package.json file and package folder when running uninstall without having installed anything. I don't have anything against that, but that's something @xPaw made me change 😁 Any reason why we're going back to that? Just for code convenience?

Nope, I just hadn't realised yours didn't do that. I shall remove it :-)

YaManicKill:
Alrighty, so I have added a few commits to:

1. Remove the binds
2. Move the npmrc to the app rather than userland
3. Show an error when the user tries to uninstall something that wasn't already installed 

4. Don't create package.json when we uninstall

In terms of the npm arguments, I think having the npmrc file is a better way of doing things (because it's the expected way of passing lots of arguments like this) but if you really want them passed in the code, then I can do that for each of the commands, it just feels unnecessary.

Also, in terms of code, this is the same amount of code as it was before, but when we add the upgrade command, it'll only add a few lines of code as opposed to 70, so the more commands we have in terms of the package manager, the more this is an advantage.

src/plugins/packageManager.js
astorije: Oh boy...

screen shot 2018-01-13 at 01 15 48

astorije: For the same reason, you can't rely on keywords `removed` or `up-to-date`. Otherwise you'll keep fixing on v8 and breaking v6, then I'll stop by and fix v6, breaking v8. Repeat 😂

astorije: So, I have spent a rather extensive amount of time trying to please nvm v3 to v5 based on output of `npm uninstall` and no luck. In the end, I fixed that more cleanly in https://github.com/thelounge/lounge/pull/1974.
Yeah, it's an extra step, but it's worth it for robustness. We are less subject to npm variations (we still rely on the package name to be part in the output, but that's about it as opposed to my previous version and yours :man_shrugging:).

Also... sorry for the rebase :/

astorije:
> This is a really silly argument, IMO. "We might have to change how this code is, so we may as well not make it nicer". I don't get this at all.

It is not a silly argument. In fact, it's not even mine, premature abstraction is a thing.
Also, come on, I did not say "we may as well not make it nicer". Let's not debate like that :)

What I meant is that it might be worth not abstracting this very new command _just now_. I had the feeling the `uninstall` command would diverge from the `install` one, one way or another. Case in point: #1974
I'd rather have 2 scripts (2, not 20) tailor-made to our needs rather than having to limit ourselves just to make them 1 script. That being said, if the abstracted version is as nice and permitting as you explained it (I did not have that feeling at first review, but I need to look at the code again after your latest write-up), then I have nothing against it. It's just that too many times I have experienced early abstraction because code looks the same in 2 places, then becoming a pain to make it evolve.

> The few times they have changed things, they have given us 1 major release with it deprecated, and then the next major release with it changed. So we would have at least a year to change after they announced it.

Not quite, because over time we will support between 2 and 3 LTS versions of Node, which come with different versions of npm. For example, Node v4 ships with npm v2, Node v6 ships with npm v3, and Node v8 ships with npm v5. It means that once we release TL v3, we still need to be compatible with npm v3 and v5 (for a while), which bit us above in the `uninstall` script.

It's all good here, this is more of an aside than a comment on your PR, I just wanted to clarify that.

---

Re: abstracting the options, I'm fine with that assuming we do it correctly. For example, I'd argue that the one thing we need to abstract the most is the `--prefix` one, because all our package commands should never ever list, install, uninstall, upgrade, etc. anywhere else.
Reciprocally, some should probably stay per-command, especially if they really abstract 1 command at the moment.

src/plugins/packageManager.js
```diff
@@ -0,0 +1,118 @@
...
+		.then(() => fs.writeJson(packagesConfig, packageDirJson, {spaces: "\t"})); // Create package.json to avoid npm warnings
+}
+
+function install(packageName) {
+	checkConfig()
+		.then(() => getPackageJson(packageName))
+		.then((json) => getMetadata(json))
+		.then((metadata) => {
+			ensurePackageJsonExists();
+			return metadata;
```
astorije: Doesn't this need to be:

```js
.then((metadata) => ensurePackageJsonExists().then(() => metadata))
```

?

YaManicKill: Hmmm yes, you are right. I think I want to write a wrapper for passing through parameters anyway, as this isn't exactly nice to look at.

Of course, actually, I want to switch it to async/await because promises are such a hacky way to fix the async issue. But we can't do that until node 6 is no longer LTS. (unless we want to add babel, which I assume we don't for node). Roll on April 2019.

YaManicKill:
> It is not a silly argument. In fact, it's not even mine, premature abstraction is a thing.
Also, come on, I did not say "we may as well not make it nicer". Let's not debate like that :)

Alright, alright, maybe I was being a little unfair. However, I feel like this is a good change and I don't think this *is* premature abstraction. I'm adding an upgrade command which with the old way would mean adding 70-80 lines of code. With this, it would be 5. 

> Case in point: #1974

As I mentioned in that PR, this doesn't make your point as I added this functionality with 2/3 lines of code (can't remember exactly, I'll double check later) so, it actually makes it really nice.

>  rather than having to limit ourselves just to make them 1 script

Where are we limiting ourselves?

> Not quite, because over time we will support between 2 and 3 LTS versions of Node, which come with different versions of npm.

This is a different argument. We *will* have the time to change our code, the problem is that we'll have to maintain for multiple npm versions. That's a fair argument, but it's different. And it's not unique to this solution. It's a problem we'll have supporting npm no matter how we implement it.

And this is beside the point, which of the arguments that we use do you think they could possibly use for anything that would cause us a problem?

> For example, I'd argue that the one thing we need to abstract the most is the --prefix one

Yeah, the problem is that this one can't be stored in an npmrc, sadly, because the prefix that we use can technically change, and that'll be awkward. If the user changes their lounge home, we need to take that into consideration, which...ew.

> Reciprocally, some should probably stay per-command, especially if they really abstract 1 command at the moment.

Right, I'm not sure I agree on this one though. Using an npmrc for the commands is the supported way of making some flags always on. It is *supported* which means that if you run `npm uninstall` and you have an install only flag, it's not going to mean anything to it. This is how npm works. The npmrc is designed for exactly this purpose. 

Also, the advantage of moving this code out of the commands means that we can call them from the app in other places. For example, I'm adding a command to check if there are updates, which we can show next to the app update.

YaManicKill:
Alright, I've updated to move the npm calling into it's own module.

I know the npm function is a little complicated, but I still think this is better because:

1. The npm calling is always complex, this way we move all the complexity into 1 place, rather than having the complexity in multiple places
2. I think the calling of child_spawn is the most complex (And error prone) part of this, and for this reason I believe it's the most important bit to abstract.

I get your arguments against premature abstraction, I really do, but I don't think this is. The npm calling is horrible (I think we can all agree that) and having it written out multiple times, IMO, is going to cause us problems sometime. Adding extra functionality to the 2 calls is still quite easy (check out [my solution for using list[(https://github.com/thelounge/lounge/pull/1954/commits/e2f80050e139fdd3a46b6471e26a9fad0d1c6f37) ... yeah it was a few more lines than 2, but still pretty good, IMO).

I think this is way better than just duplicating the code, and I don't think it stops us having any of the functionality we want. If it does, we can throw it in the bin at that point, but this is definitely (IMO) better than what we have at this point.

YaManicKill:
This is clearly moving too much just now. I'll come back to it once we're more stable.

PR #1955, opened 2018-01-09
Prevent user contextmenu from opening while selecting text

### LINKED ISSUE #1945: Selecting nicknames on Android opens context menu
We should return early in click handler if there is any text selected to fix that.

xPaw:
Fixes #1945

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -170,6 +170,12 @@ $(function() {
 	});
 
 	viewport.on("click contextmenu", ".user", function(e) {
+		// If user is selecting text, do not open context menu
+		// This primarily only targets mobile devices where selection is performed with touch
+		if (!window.getSelection().isCollapsed) {
+			return true;
+		}
+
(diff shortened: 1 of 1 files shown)


client/js/lounge.js
```diff
@@ -218,13 +224,7 @@ $(function() {
...
 		chat.on("click", ".chat", function() {
 			setTimeout(function() {
-				var text = "";
-				if (window.getSelection) {
-					text = window.getSelection().toString();
-				} else if (document.selection && document.selection.type !== "Control") {
-					text = document.selection.createRange().text;
-				}
-				if (!text) {
+				if (window.getSelection().isCollapsed) {
```
astorije: Do you know why we had that extra code before? I feel like it was a browser protection or something, but I'm not sure if removing it might not break something.

xPaw: That was to support older browsers, however we already only use `getSelection` in other parts of the code without polyfilling it.

PR #1956, opened 2018-01-09
Provide correct timestamp to browser notifications

xPaw:
Tried new style notifications in Chrome, and they were showing up as 48 years old.

This is a problem unique to 2.7 because before that timestamp was not set.

### DIFF
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -149,13 +149,15 @@ function notifyMessage(targetId, channel, msg) {
 					body = cleanIrcMessage(msg.text);
 				}
 
+				const timestamp = Date.parse(msg.time);
+
 				try {
 					if (webpush.hasServiceWorker) {
 						navigator.serviceWorker.ready.then((registration) => {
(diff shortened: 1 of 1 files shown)


client/js/socket-events/msg.js
```diff
@@ -166,7 +168,7 @@ function notifyMessage(targetId, channel, msg) {
 							badge: "img/logo-64.png",
 							icon: "img/touch-icon-192x192.png",
 							body: body,
-							timestamp: msg.time,
+							timestamp: timestamp,
```
YaManicKill: So we want to explicitly format this rather than jsut relying on the toString?

xPaw: `msg.time` is a string, not `Date`.

YaManicKill: Sure, but you've turned it into a date.

xPaw: `Date.parse` returns int.

YaManicKill: Oh. Duh.

PR #1970, opened 2018-01-12
Fix local themes not being found in theme selector dropdown, and serve local themes using the themes route instead of the public folder

### LINKED ISSUE #1968: Built-in themes are not available anymore
I was giving one last test before releasing `v2.7.0-rc.1` and in the settings...

screen shot 2018-01-12 at 00 11 19

Built-in themes (Example, Morning, Zenburn, Crypto) do not show there. However, if I install a theme with `thelounge install`, it will correctly show in the dropdown.

No errors or warnings in the browser JS console, the server console, nor the Webpack output, which correctly contains the following:

```
               themes/crypto.css    1.55 kB          [emitted]
              themes/example.css  819 bytes          [emitted]
                   css/style.css      46 kB          [emitted]
              themes/morning.css    4.63 kB          [emitted]
               css/bootstrap.css    18.5 kB          [emitted]
              themes/zenburn.css       5 kB          [emitted]
```

We very recently merged https://github.com/thelounge/lounge/pull/1619 which touches themes, so that might be a good first place to look at.

### LINKED ISSUE #1969: "/themes/:theme.css" route is not actually used for local themes
This one cost me some hair to figure it out...

When going to, say `/themes/zenburn.css`, https://github.com/thelounge/lounge/blob/3eb017c/src/server.js#L51-L58 is not actually called. Instead, it directly loads the static file located at `public/themes/zenburn.css` due to the `.use(express.static("public"))` declaration a bit above.

What this means is that the `filename` value at https://github.com/thelounge/lounge/blob/3eb017c/src/plugins/packages/themes.js#L49 is actually never used. In fact, this value is currently broken (see #1968) but the built-in theme files are correctly found because that route never gets hit.

I'm not 100% sure what's the ideal fix for this: accept the current situation and not attempt to build the `filename` path for built-in (local) themes, or make sure local themes go through the same route than external ones.
I prefer the latter for consistency and avoid surprising of having 2 very similar URLs being handled differently (in fact, I couldn't find out why my `console.log`s were not being printed out and was wondering about cache). If the latter, then we should remove local themes from the `public` folder, since we now serve them through the server and not as static assets.

astorije:
Fixes #1968.
Fixes #1969.
Supersedes #1970, which contains additional details.

This code was moved from `src/plugins/themes.js` to `src/plugins/packages/themes.js` (one level deeper) in https://github.com/thelounge/lounge/pull/1619/files#diff-9f82c1f3ff2c49ecbb24cd28408ee089L15 but the path being built was not changed.

### DIFF
--- a/src/plugins/packages/themes.js
+++ b/src/plugins/packages/themes.js
@@ -14,7 +14,7 @@ module.exports = {
 };
 
 function loadLocalThemes() {
-	fs.readdir(path.join(__dirname, "..", "..", "public", "themes"), (err, builtInThemes) => {
+	fs.readdir(path.join(__dirname, "..", "..", "..", "public", "themes"), (err, builtInThemes) => {
 		if (err) {
 			return;
 		}
@@ -46,7 +46,6 @@ function makeLocalThemeObject(css) {
 	const themeName = css.slice(0, -4);
 	return {
 		displayName: themeName.charAt(0).toUpperCase() + themeName.slice(1),
-		filename: path.join(__dirname, "..", "..", "public", "themes", `${themeName}.css`),
 		name: themeName,
 	};
 }
--- a/src/server.js
+++ b/src/server.js
@@ -48,6 +48,10 @@ module.exports = function() {
 			maxAge: 86400 * 1000,
 		}));
 
+	// This route serves *installed themes only*. Local themes are served directly
+	// from the `public/themes/` folder as static assets, without entering this
+	// handler. Remember this is you make changes to this function, serving of
+	// local themes will not get those changes.
 	app.get("/themes/:theme.css", (req, res) => {
 		const themeName = req.params.theme;
 		const theme = themes.getFilename(themeName);


src/plugins/packages/themes.js
```diff
@@ -46,7 +46,7 @@ function makeLocalThemeObject(css) {
 	const themeName = css.slice(0, -4);
 	return {
 		displayName: themeName.charAt(0).toUpperCase() + themeName.slice(1),
-		filename: path.join(__dirname, "..", "..", "public", "themes", `${themeName}.css`),
+		filename: path.join(__dirname, "..", "..", "..", "public", "themes", `${themeName}.css`),
```
astorije: Note that this line is currently not being used because of https://github.com/thelounge/lounge/issues/1969.

astorije:
Closing this as superseded by #1971, but that's just because the code change happens to be on the same lines. It is indeed 2 different underlying issues :)

YaManicKill:
> This code was moved from src/plugins/themes.js to src/plugins/packages/themes.js (one level deeper) in thelounge/lounge/pull/1619/files#diff-9f82c1f3ff2c49ecbb24cd28408ee089L15 but the path being built was not changed.

Oh, duh. Well done me.

astorije:
@YaManicKill, @xPaw, made the changes. Don't forget to clean up your `public` folder (or remove `public/themes`) if you want to fully test this.

src/plugins/packages/themes.js
```diff
@@ -46,7 +46,6 @@ function makeLocalThemeObject(css) {
 	const themeName = css.slice(0, -4);
 	return {
 		displayName: themeName.charAt(0).toUpperCase() + themeName.slice(1),
-		filename: path.join(__dirname, "..", "..", "public", "themes", `${themeName}.css`),
```
astorije: This is effectively never used, and having it gave a false sense that the [`/themes/:theme.css`](https://github.com/thelounge/lounge/blob/master/src/server.js#L51-L58) route is using it.

PR #1971, opened 2018-01-12
Fix themes not being found in theme selector dropdown, and being incorrectly served from the public folder

### LINKED ISSUE #1968: Built-in themes are not available anymore

### LINKED ISSUE #1969: "/themes/:theme.css" route is not actually used for local themes

astorije:
Fixes #1968.
Fixes #1969.
Supersedes #1970.

This highlights an anti-pattern (that we might have in other places?): Building/Serving files located in the `public` folder means we're asking Express to serve files that are already publicly available, i.e. served. In this case, we're serving themes at `/themes/:theme.css` **and** files that live at `/public/themes/*.css` behind the same route.

So, since we're serving it via Express, we can remove it from the Webpack config.
Reason why I moved the files from the client to the server is because, in the end, that's where we serve them from. Installed themes with `thelounge install` actually have nothing to do with client code, they're much closer to server code, so this makes it more consistent. Additionally, if we were to keep themes in client folder, that means we'd have to load a file located at `../client` from the server, and I think we should avoid these server/client mixes.

I'm not too concerned about location of those themes anyway, because in the future I think we will move them to their own repos, npm packages, and have some or all being shipped/installed by default with TL.


xPaw:
I really don't want to move themes into src folder like you did here. Reason public folder exists is.. precisely that, only files that are publicly accessible are within that folder.

I'd say go with #1970, because local themes need to be eventually turned into proper themes at some point anyway so the local theme functions you modified will go away.

YaManicKill:
I'm with @xPaw on not moving them. Although, I should note that files from packages and themes aren't moved to public, because they can't if we don't want to have to require a build step. The server serves the files directly.

I feel like we should go with #1970 and just not copy the files to the public folder, and serve them the same was as we serve the packaged themes. Or we just keep the fact that they are stored in public, and just use that. Either way, yeah, I'm not a fan of this. Sure, they are served by the server, but they are client code, not server code.

astorije:
> Reason public folder exists is.. precisely that, only files that are publicly accessible are within that folder.

Except themes that way aren't strictly publicly accessible, they're proxied by the server, like `index.html` (which isn't in the `public` folder) or installed themes.

> local themes need to be eventually turned into proper themes at some point anyway so the local theme functions you modified will go away.

👍
Agreed, hence my last paragraph in PR description.

---

Anyway, keeping this in the client folder means the server code will reference stuff in the client folder, which I hate. But oh well.

astorije:
Reopened and updated #1970 accordingly.

Issue #914, opened 2017-02-09
Registration UI


theSoenke:
I would be nice if user could register via the web interface instead that the administrator manually needs to create accounts via the command line. This way it would be possible to just send a link to the running lounge instance

MaxLeiter:
Some discussion and a potential PR can be found here https://github.com/thelounge/lounge/pull/641

theSoenke:
@MaxLeiter Great. Did not saw this before

astorije:
Unfortunately, the author of #641 is too busy to resume work on his PR. If anyone wants to help by picking up where he left off, I'm sure the community would like it!

However, make sure to discuss it here and share your plans otherwise the PR could stay open forever again, and have a lot of back-and-forth with maintainers like #641 did.

Thanks all! 😄

lazy-sunshine:
I would like to work on this issue

astorije:
Hey @lazy-sunshine, that's great! Feel free to look if #1973 needs help. I haven't looked at it yet, planning to do so soon (not sure if other maintainers have either). There are a few considerations and that could go well with #1950 too.

Arkadipta-Chatterjee:
Can i work on this??

brunnre8:
I think as per the linked issues, best would be if you start with a spec.
How do you think it should behave?

PR #1973, opened 2018-01-12
Adding sign-up feature

thisisdarshan:
This PR is to add a sign-up feature.  It is intended to support local and LDAP sign-up both.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -281,7 +281,7 @@ kbd {
 #chat .special .title::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
 
 #footer .sign-in::before { content: "\f023"; /* http://fontawesome.io/icon/lock/ */ }
-#footer .sign-up:before { content: "\f234"; /* http://fontawesome.io/icon/user-plus/ */ }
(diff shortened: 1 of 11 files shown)


christer88:
This one needs to be optional in config file, and default to false. I'm certain not everyone would like a sign-up feature to be open for the public.

dgw:
Needs fixes to linter errors and a proper commit message.

MaxLeiter:
Not a fan of #signup-signin - shouldn’t the sign up button be #signup and the signin button be #signin?

McInkay:
Sorry about the long time it is taking to look at this, @thisisdarshan .

There has been a lot of change recently, is this something you would be able to rebase and fix, or are you not around to do it anymore? 

We will be porting the client side code to vue (#2636) so it might be easier to rewrite than fix this current one. Let me know what you think.

xPaw:
I will close this PR (for now at least) due to inactivity.

Sorry for that, I think we will need to discuss this feature a little more before actually implementing it. Stuff like abuse and rate limiting needs to be considered.

An issue for reference and discussion: #914

PR #1974, opened 2018-01-13
Fix `thelounge uninstall` command

astorije:
- Exit with code 1 when package was not installed
- Run a pre-step to check if package is installed before uninstalling. I have tried every possible way I could think of without that first `npm list` step based on output of `npm uninstall`, but different npm versions produce different outputs, so there is nothing reliable. This is a more robust way to do so anyway.
- Consolidate error handlers

### DIFF
--- a/src/command-line/uninstall.js
+++ b/src/command-line/uninstall.js
@@ -28,45 +28,75 @@ program
 
 		if (!fs.existsSync(packagesConfig)) {
 			log.warn(packageWasNotInstalled);
-			return;
+			process.exit(1);
 		}
 
-		const npm = child.spawn(
-			process.platform === "win32" ? "npm.cmd" : "npm",
+		const npm = process.platform === "win32" ? "npm.cmd" : "npm";
+		const errorHandler = (error) => {
+			log.error(
+				`Failed to uninstall ${colors.green(packageName)}. ` +
+				`${typeof x === "number" ? "Exit code" : "Error"}: ${error}`
+			);
+			process.exit(1);
+		};
+
+		// First, we check if the package is installed with `npm list`
+		const list = child.spawn(
+			npm,
 			[
-				"uninstall",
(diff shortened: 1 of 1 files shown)


src/command-line/uninstall.js
```diff
@@ -28,45 +28,75 @@ program
...
+				//   package was found
+				// - `process.stderr` is ignored to silence `npm ERR! extraneous` errors
+				stdio: [process.stdin, "pipe", "ignore"],
 			}
 		);
 
-		let hasUninstalled = false;
-
-		npm.stdout.on("data", () => {
-			hasUninstalled = true;
```
astorije: This is the key change. No more assumption that if `uninstall` outputted something, then package was uninstalled, because that is not true on npm v5 (shipping with Node v8) anymore.

Instead, we're checking once and for all with `npm list` before uninstalling. I should have done that in the first place, but I was lazy and thought we could get smart about it. It's npm, so we can't lol.

astorije:
Tested this against npm v2, v3 and v5 (coming resp. with Node v4, v6, v8).

xPaw:
At what point will it be easier to just avoid npm completely and download tarballs ourselves? 🤔

astorije:
> At what point will it be easier to just avoid npm completely and download tarballs ourselves? 🤔

I don't think it will be easier, because we'll have another world of checks, commands, etc. to run ourselves. I think we're better off using npm still, at least for now.
I just looked at `apm` to see how they were doing out of curiosity, and they're still using npm actually! With the slight difference that they're using the programmatic API instead of the command, but npm themselves do not recommend this because they call it a private API :man_shrugging: 

On the long run, I think programmatic API would be a better choice than running commands or downloading tarballs ourselves, but as far as I know, I don't think it's planned for npm, and it's not available for yarn (though on their radar apparently).

PR #1701, opened 2017-11-11
Go back to writing user files synchronously

xPaw:
I just had my user file corrupted while testing locally when I ctrl+c'd the server. This might also be the cause of empty files that some users experience.

Hopefully fixes #1690

PR #1978, opened 2018-01-13
Add missing execution of callback in ClientManager.updateUser

merlinthp:
Without this, the change-password message isn't sent back to the client to indicate the result of the operation.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -136,6 +136,9 @@ ClientManager.prototype.updateUser = function(name, opts, callback) {
 
 	if (!user) {
 		log.error(`Tried to update invalid user ${colors.green(name)}. This is most likely a bug.`);
+		if (callback) {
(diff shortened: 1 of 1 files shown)


xPaw:
This was broken in #1701

src/clientManager.js
```diff
@@ -154,6 +154,8 @@ ClientManager.prototype.updateUser = function(name, opts, callback) {
 		log.error(`Failed to update user ${colors.green(name)} (${e})`);
 		throw e;
```
xPaw: Also need to `callback(e)` for failed writes.

merlinthp:
And in the if (!user) statement at the top of the function too?

xPaw:
Technically yes, but that wouldn't happen when changing password.

merlinthp:
No problem :)

PR #1979, opened 2018-01-14
Do not statically serve the index template prior to rendering it

### LINKED ISSUE #1976: Do not serve index.html as a static asset
At the moment, loading `https://example.com/index.html` displays the`index.html` file as-is, i.e. un-rendered. This means that if someone loads this route by mistake, the app appears not working.

I think we should do both of those:

- Rename `index.html` into `index.html.tpl`. The file on its own is not sufficient without [rendering it](https://github.com/thelounge/lounge/blob/master/src/server.js#L255).
- Do not copy it to the `public` folder at Webpack build-time, and exclude `client/index.html.tpl` in the npm ignore config

astorije:
Fixes #1976.

Without this, going to `https://thelounge.example.com/index.html` would return the raw file. This now excludes it from the `public` folder so it cannot be rendered as is.

Renaming the file is for good measure, to indicate that this HTML file must be templated. Because it is a straight rename with no modification, rebasing PRs on it should not be to painful, as git re-applies changes on renamed files.

I know @YaManicKill preferred renaming it to `index.tpl` instead of `index.html.tpl`.
I prefer the latter because it's more explicit what is the expected format of the file and it costs nothing to have the extension. It's also a very common approach, like `.html.erb` files in Ruby, or `nginx.conf.j2` / `file.json.j2` when using Jinja2 for Ansible or equivalent provisioning tools.
Also I'm thinking some tools/editors might use `*.<ext>.tpl` to determine syntax highlighting, but I have no proof of that.
That being said, it's really unimportant so if strong opinion or consensus in the other direction, I'll happily change. I don't care, Atom treats all `*.tpl` files as HTML :)


astorije:
Looking at failures now.

xPaw:
You need to allow this file in npmignore, otherwise release build is gonna break.

astorije:
Good catch, I even mentioned it myself in #1976 :man_facepalming: 
Fixed.

astorije:
@McInkay, what are your thoughts on the template name? I don't want to bypass your opinion on this by merging straight away. Also while I do prefer `.html.tpl` for explicitness/common practice, I don't care _that_ much to fight over it :)

Issue #1762, opened 2017-11-28
Update website URL for 3.0


MaxLeiter:
for tracking purposes

Places link exists:

- [x] GitHub org description
- [x] Repo description/link
- [ ] Rename website repo
- [x] Website repo description
- [x] [Twitter share button on the website](https://github.com/thelounge/thelounge.github.io/blob/f3092e1a8a7a22935995be1ed5085df128b507d3/_layouts/default.html#L49)
- [x] [References in the code](https://github.com/thelounge/lounge/search?utf8=%E2%9C%93&q=https%3A%2F%2Fthelounge.github.io&type=)
  - https://github.com/thelounge/lounge/pull/1980

astorije:
[defaults](https://github.com/thelounge/lounge/blob/c9da29f2fcaffc01de779c8217a4a50f52cef28a/defaults/config.js#L328)

xPaw:
Gonna close this as it was pretty much done. Website repo can't be renamed as that breaks redirects.

PR #1980, opened 2018-01-14
Replace reference to website with new URL

astorije:
One step towards https://github.com/thelounge/lounge/issues/1762 (but not enough to close it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -664,12 +664,12 @@ In the main repository:
 - Fix general spelling errors ([#1458](https://github.com/thelounge/lounge/pull/1458) by [@PolarizedIons](https://github.com/PolarizedIons))
 - Remove duplicate keybindings help ([#1543](https://github.com/thelounge/lounge/pull/1543) by [@xPaw](https://github.com/xPaw))
 
-On the [website repository](https://github.com/thelounge/thelounge.github.io):
+On the [website repository](https://github.com/thelounge/thelounge.chat):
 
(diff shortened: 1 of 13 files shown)


src/helper.js
```diff
@@ -113,6 +113,12 @@ function setHome(newPath) {
 	} else {
 		this.config.theme = `themes/${this.config.theme}.css`;
 	}
+
+	if (this.config.leaveMessage.includes("thelounge.github.io")) {
```
xPaw: I don't really see a need in doing this.

astorije: I'd like to keep this for a couple version. It doesn't cost anything, and it might be useful for whoever started to use The Lounge with versions 2.5 or 2.6, or copied this in their config. There will be a redirect in the URL, but that might definitely help.

xPaw: @McInkay what do you think?

McInkay: I think it would make more sense that if the leave message is the default one, changing it, because they clearly don't care about what the default leave message is.

But I don't think it's really necessary, it redirects, so ...  ¯\\\_(ツ)_/¯

astorije: We can't change the values in the config file ourselves after the first start (i.e. when the file has been copied over). Did that in the past and there were permission issues, I don't want to break any more things 😅

I can remove this if you don't like it, but otherwise we can keep it in v3.0 and remove in v3.1. That will be a reminder for people to change it in their config.

xPaw: I think asking to change Lounge url will just push most people to either not care, or change it to something completely different :)

astorije: Ok, I removed it.

PR #1984, opened 2018-01-15
Add SECURITY guidelines about security vulnerability disclosures, and link them from the CONTRIBUTING guidelines

astorije:


CONTRIBUTING.md
```diff
@@ -3,6 +3,16 @@
...
 
+### I have found a security vulnerability
+
+- If you think you discovered a security vulnerability, **please do not open a
+  public issue about it on GitHub.** Contact us privately first, in a
+  [responsible disclosure](https://en.wikipedia.org/wiki/Responsible_disclosure)
+  manner.
+- On IRC, contact in private message anyone who is a voiced user on our Freenode
+  channel `#thelounge`.
+<!-- - Or send us an email at <mailto:security@thelounge.chat>.-->
```
astorije: this is not set up so I just added it as a template, but it would be nice if this emailed us @thelounge/maintainers. Opt-in of course, if some maintainers do not want to get emails. I'm happy to be on the list, personally.

YaManicKill: Happy to be on the list.

xPaw:
Do we want to put this in `SECURITY.md` instead? I've seen other projects do this, e.g. https://github.com/electron/electron/blob/master/SECURITY.md

YaManicKill:
> Do we want to put this in SECURITY.md instead?

astorije:
@xPaw, @YaManicKill, better? :)

PR #1987, opened 2018-01-17
Switch from npm to yarn

### LINKED ISSUE #1935: Pass --ignore-scripts to npm install in lounge install command

xPaw:
- [x] Let greenkeeper handle lockfile (added GH_TOKEN on travis)
- [x] npmignore works with yarn
- [x] updated gitgnore
- [x] Deal with npmrc (yarn supports it)
- [x] Update documentation to use yarn
- [x] Update lounge (un)install commands to use yarn as an API
- [x] Fixes #1935


.travis.yml
```diff
@@ -11,12 +11,17 @@ matrix:
     - node_js: 8 # Version used to deploy to npm registry
       env: BUILD_ENV=production
 
-cache:
-  directories:
-    - ~/.npm
+cache: yarn
+
+before_install:
+  - yarn global add greenkeeper-lockfile@1
```
astorije: `@1` for good measure?

.travis.yml
```diff
@@ -11,12 +11,17 @@ matrix:
...
 
 before_script:
-  - NODE_ENV=$BUILD_ENV npm run build
+  - greenkeeper-lockfile-update
+  - NODE_ENV=$BUILD_ENV yarn run build
```
astorije: `run` is not necessary with Yarn (one of the many small wins of Yarn, like passing flags without needing `--`, etc.), so `yarn build` works fine.

appveyor.yml
```diff
@@ -15,19 +15,19 @@ environment:
 
 install:
   - ps: Install-Product node $env:nodejs_version
-  - appveyor-retry npm install
-  - npm run build
-  - npm install mocha-appveyor-reporter
+  - yarn
+  - yarn run build
```
astorije: Same thing here.

src/command-line/uninstall.js
```diff
@@ -40,63 +46,37 @@ program
...
-					stdio: [process.stdin, "ignore", process.stderr],
+				if (line.type === "success") {
+					success = true;
 				}
-			);
+			});
+		});
 
-			uninstall.on("error", errorHandler);
+		yarn.on("error", (e) => errorHandler);
```
astorije: I don't think this is correct (also fails linter).

src/command-line/uninstall.js
```diff
@@ -24,14 +24,20 @@ program
 		const packagesPath = Helper.getPackagesPath();
 		const packagesParent = path.dirname(packagesPath);
 		const packagesConfig = path.join(packagesParent, "package.json");
-		const packageWasNotInstalled = `${colors.green(packageName)} was not installed.`;
+		const packageWasNotInstalled = `${colors.green(packageName)} is not installed.`;
```
astorije: Meh :man_shrugging:

src/command-line/install.js
```diff
@@ -38,45 +38,54 @@ program
...
+				]
 			);
 
-			npm.on("error", (e) => {
+			yarn.stdout.on("data", (data) => {
+				data.toString().trim().split("\n").forEach((line) => {
+					line = JSON.parse(line);
+
+					if (line.type === "step") {
+						log.info(line.data.message);
```
astorije: This now displays:

screen shot 2018-01-18 at 01 36 47

i.e. lines 3 to 6 are new. I don't think they add any value to the user (it is all summarized in the second line, ours) so we should silence them.
In fact, I think first line is not very useful either, we might want to get rid of it too. When it comes to fetching metadata, it only matters to know that the package is not a TL package, which we display as an error.

astorije:
Rebased on master and fixed conflicts.

> - [ ] Update documentation to use yarn

You mean on this repo (README, etc.), correct? I'll update the new docs accordingly based on the changes we make here.

BTW, what are you thinking of, complete switch to `yarn` in README, or keeping both?
On one end, I see most repos out there mention how to install with both `yarn` and `npm`, on the other hand, if we don't test/CI it ourselves, there will come a time where it breaks...

src/command-line/install.js
```diff
@@ -38,45 +38,54 @@ program
...
 					private: true,
 					description: "Packages for The Lounge. All packages in node_modules directory will be automatically loaded.",
 				}, null, "\t"));
 			}
 
-			const npm = child.spawn(
-				process.platform === "win32" ? "npm.cmd" : "npm",
+			let success = false;
+			const add = child.spawn(
+				"./node_modules/.bin/yarn",
```
astorije: @xPaw, could you test on Windows with my latest commit?
I am using `./node_modules/.bin/yarn`, which is actually the same as `./node_modules/yarn/bin/yarn.js` and **not** `./node_modules/yarn/bin/yarn`. The `yarn.cmd` one simply calls `yarn.js` (`yarn/bin/yarn` and `yarn/bin/yarn.cmd` are useful to open new terminal on double-click, if I understand correctly) so `.bin/yarn` should actually be cross-platform.

xPaw: It's not classform and still needs needs the `.cmd`. Either that or directly run `node yarn.js`, should be able to get current node binary to run under it.

xPaw: I switched to using `process.execPath` and `yarn.js` to avoid any ambiguity.

src/command-line/install.js
xPaw: Removed.

.travis.yml
xPaw: That's what greenkeeper recommends. And yeah, good idea since we don't know the exact version.

xPaw:
> You mean on this repo (README, etc.), correct?

Yep.

> BTW, what are you thinking of, complete switch to yarn in README, or keeping both?
I think just suggest yarn, and add a note that npm might work but we're not going to support it.

.travis.yml
```diff
@@ -18,7 +18,7 @@ before_install:
 
 before_script:
   - greenkeeper-lockfile-update
-  - NODE_ENV=$BUILD_ENV yarn run build
+  - NODE_ENV=$BUILD_ENV yarn build
```
astorije: you probably wanted to squash this into the first commit, not this one :/

README.md
```diff
@@ -56,12 +65,12 @@ The following commands install and run the development version of The Lounge:
 ```sh
 git clone https://github.com/thelounge/lounge.git
 cd lounge
-npm install
-NODE_ENV=production npm run build
-npm start
+yarn
```
astorije: I know it's not necessary, but I would document it as `yarn install` to be a bit more explicit. nbd though

README.md
```diff
@@ -56,12 +65,12 @@ The following commands install and run the development version of The Lounge:
 ```sh
 git clone https://github.com/thelounge/lounge.git
 cd lounge
-npm install
-NODE_ENV=production npm run build
-npm start
+yarn install
```
Stanzilla: Just `yarn` is enough

xPaw: We know. Opted to keep this way because it's not immediately obvious for people that haven't used yarn.

README.md
```diff
@@ -27,14 +27,23 @@ The Lounge is the official and community-managed fork of [Shout](https://github.
 ## Installation and usage
 
 The Lounge requires [Node.js](https://nodejs.org/) v4 or more recent.
+[Yarn package manager](https://yarnpkg.com/) is also required to install The Lounge.
 
-### Running stable releases from npm (recommended)
+*While using `npm` might work, we no longer officially support this method.*
```
YaManicKill: I'm not sure how I feel about dropping support entirely for npm. This feels drastic.

src/command-line/uninstall.js
YaManicKill: I'm also not sure on this one. Saying "x is not installed" doesn't give the indication that we want, because when you finish installing, of course it's not installed. The key point is that we *didn't uninstall* because there wasn't a package installed to uninstall. At this point, when we print this message, this statement will always be the case, so we aren't telling the user what we want.

YaManicKill: Also, at the *very least* please make the name of the variable the same as the string ;-)

src/command-line/uninstall.js
```diff
@@ -40,63 +46,48 @@ program
...
-					errorHandler(code);
-				}
+		remove.on("close", (code) => {
+			if (!success || code !== 0) {
+				errorHandler(code);
+			}
 
-				log.info(`${colors.green(packageName)} has been successfully uninstalled.`);
-			});
+			log.info(`${colors.green(packageName)} has been successfully uninstalled.`);
```
YaManicKill: Will this run even if there was an error? We shouldn't do that.

src/command-line/uninstall.js
```diff
@@ -24,14 +24,20 @@ program
...
 
 		if (!fs.existsSync(packagesConfig)) {
 			log.warn(packageWasNotInstalled);
 			process.exit(1);
 		}
 
-		const npm = process.platform === "win32" ? "npm.cmd" : "npm";
+		const packages = JSON.parse(fs.readFileSync(packagesConfig, "utf-8"));
+
+		if (!packages.dependencies || !packages.dependencies.hasOwnProperty(packageName)) {
```
YaManicKill: So we are storing the plugins in package.json now?

src/command-line/uninstall.js
xPaw: errorHandler exits the process.

src/command-line/uninstall.js
xPaw: Yep, this is already in master due to npm bugginess.

src/command-line/uninstall.js
xPaw: > At this point, when we print this message, this statement will always be the case, so we aren't telling the user what we want.

It's printed when the package is in fact not installed. What do you suggest the message should be?

YaManicKill: > It's printed when the package is in fact not installed.

Sure, but my point is that that's what we *want*. At the end of this command, it will *always* not be installed. That's the point of uninstalling. However, if it *wasn't* installed before, then there might be an issue with that, there might have been a typo, etc.

I would say that what we had before was probably fine. It's correct, and it makes sense.

src/command-line/uninstall.js
YaManicKill: Oh duh, yeah. I don't like that, but let's not change it as part of this PR.

xPaw: I added a return to make it clear.

README.md
dgw: What? I was following along thinking this was just for the package system (`lounge install` etc.). I'd need two Node package managers installed if this goes through as-is?

astorije: It's not so much about dropping entirely npm support, mainly switching the docs to `yarn`. As you can see in this PR, `package.json` does not change, so everything _should_ work the same way with `npm`.
However, I don't see myself testing everything we do in 2 different folders, with 2 different setups to make sure all new changes (dependencies, etc.) are 100% compatible. So yeah, official documentation to suggest `yarn`, and if people want to keep using `npm`, that's fine, they still can. If a bug arises that's only with `npm`, we can try to fix it, but that's about it, we won't set up 2 CIs, etc.

Also @dgw, I was fairly skeptical about `yarn` not long ago for that exact same reason, needing 2 package managers for Node, but as I started using it, I could see how `yarn` was much much superior to `npm`. But again, you don't have to use `yarn` with TL v3, it's just how it'll be documented.

src/command-line/uninstall.js
astorije: @YaManicKill, I'm confused, probably looking at a wrong line or something, but what don't you like?

README.md
xPaw: I didn't want to switch to Yarn either, but different npm issues and their management just pushed me to switch.

YaManicKill: Right, but we are conflating things here.

We are switch to yarn for our CI and for the internal package manager.

However, how people install our app shouldn't change. We shouldn't be forcing them to change to yarn if they don't want to. I know we aren't purposefully breaking npm, but I very much disagree with saying `While using `npm` might work, we no longer officially support this method.`.

We definitely *should* be supporting npm, and ensuring it still works.

xPaw: We shouldn't be able to break npm because we're only using it for 2 things: installing lounge (out of our control in the first place) and starting lounge. Both of these actions are compatible with yarn. 

Supporting and ensuring npm works would mean testing it on CIs, no one is going to be testing it manually.

YaManicKill: > We shouldn't be able to break npm because we're only using it for 2 things: installing lounge (out of our control in the first place) and starting lounge. Both of these actions are compatible with yarn.

Yeah, exactly, so this line is also unnecessary.

> Supporting and ensuring npm works would mean testing it on CIs, no one is going to be testing it manually.

No, we don't. We support all the features in our app but that doesn't mean they are all tested on CIs. What support means is that if someone finds a bug with something, we'll probably fix it. And we won't use any yarn specific things for what you would need for npm.

Also, of course someone is going to be testing it manually. @dgw has already said that he doesn't want to use yarn. I still use npm for the majority of things, I won't be moving to yarn for the lounge anytime soon.

src/command-line/uninstall.js
YaManicKill: I'm not a fan of the errorHandler exiting the system. But that's not part of this PR.

.travis.yml
```diff
@@ -11,12 +11,17 @@ matrix:
...
-    - ~/.npm
+cache: yarn
+
+before_install:
+  - yarn global add greenkeeper-lockfile@1
 
 before_script:
-  - NODE_ENV=$BUILD_ENV npm run build
+  - greenkeeper-lockfile-update
+  - NODE_ENV=$BUILD_ENV yarn build
```
astorije: It's a bit silly, but according to https://travis-ci.org/thelounge/lounge/jobs/334724018#L531, `npm test` is still being run. We probably want to explicitly define `yarn test` in `script`, and/or let them know that `npm` is incorrectly used.

astorije: Same with https://travis-ci.org/thelounge/lounge/jobs/334724018#L449-L450, they should print `yarn --version` instead.
Okay, these 2 might be worth opening a support ticket or something.

As part of this, it's also worth checking with them if releasing will correctly be done by `yarn` (but still on npm registry).

xPaw: I created https://github.com/travis-ci/travis-build/pull/1299 for that. If that's not accepted, I will change our script.

README.md
```diff
@@ -27,14 +27,21 @@ The Lounge is the official and community-managed fork of [Shout](https://github.
 ## Installation and usage
```
astorije: What about changing the destination of the npm version badge above from https://www.npmjs.org/package/thelounge to https://yarn.pm/thelounge? I find Yarn's package page so much nicer.

(I know the actual destination is https://yarnpkg.com/en/package/thelounge, but the `yarn.pm` one is the one they show on page + I'm wondering if they wouldn't redirect to the language requested by the browser)

README.md
```diff
@@ -56,12 +63,12 @@ The following commands install and run the development version of The Lounge:
...
-npm install
-NODE_ENV=production npm run build
-npm start
+yarn install
+NODE_ENV=production yarn build
+yarn start
 ```
 
-When installed like this, npm doesn't create a `thelounge` executable. Use `npm start -- <command>` to run subcommands.
+When installed like this, npm doesn't create a `thelounge` executable. Use `yarn start -- <command>` or `node index.js <command>` to run subcommands.
```
astorije: With `yarn`, commands can be run with `yarn start <command>` (i.e. no need for `--`).

dgw: Shouldn't this read `yarn doesn't create`?

README.md
xPaw: You can tackle this in your revamped readme PR. I'm fine with yarn link.

README.md
xPaw: `yarn start command` didn't work anyway because `start` already runs a command. I reworded it.

astorije: > `start` already runs a command.

What did you mean by this?

xPaw: It would turn into `node index.js start <command>` so it didn't exactly do what was implied.

.travis.yml
astorije: Update: they mentioned they'll merge this and deploy tomorrow. 👌

README.md
astorije: Damn, didn't realize this has been broken for a year 😂 https://github.com/thelounge/lounge/commit/86ed0b6e59f1e53b9967f3344b4dd68de0cd6ca2#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L16

PR #1993, opened 2018-01-20
Show channels for only active server in autocomplete

### LINKED ISSUE #1381: Autocomplete for channels should maybe only show channels for current server
Current behaviour:
Autocomplete shows the list of channels across all servers that I'm connected to when I type `#`

Is it possible that only the channels for the current server to be displayed (since referencing someone to a channel in _another_ server is not usual I think).

How does this suggestion sound? I might also be able to help out if needed.

milindl:
Fix #1381.

I hope this is correct. It seems to work for me.

### DIFF
--- a/client/js/autocompletion.js
+++ b/client/js/autocompletion.js
@@ -282,7 +282,9 @@ function completeCommands(word) {
 function completeChans(word) {
 	const words = [];
 
-	sidebar.find(".chan")
+	sidebar.find(".chan.active")
+		.parent()
+		.find(".chan")
 		.each(function() {
 			const self = $(this);
 			if (!self.hasClass("lobby")) {


client/js/autocompletion.js
```diff
@@ -282,7 +282,7 @@ function completeCommands(word) {
 function completeChans(word) {
 	const words = [];
 
-	sidebar.find(".chan")
+	sidebar.find(".chan.active")
```
xPaw: This will only find the active channel, not network. You have to go up to the parent.

milindl: I've now fixed it, and the commits are squashed into one as the Contributing guidelines say.

xPaw: Please change spaces to tabs on the lines that have been added. We use tabs in our codebase.

milindl: I am sorry I missed that, I have also fixed that now.

PR #1999, opened 2018-01-23
Fix timestamp tooltips not aligning correctly with timestamps

### LINKED ISSUE #1998: Timestamp tooltip is centered on the whole message, not the timestamp itself
When the message has multiple lines, or a preview in it, the tooltip gets vertically centered on the message, instead of the timestamp text only. Examples of incorrect alignment:

screen shot 2018-01-21 at 23 49 31

screen shot 2018-01-23 at 00 45 05

Probably a relatively easy fix.

### LINKED ISSUE #2000: Border radius of timestamp tooltips is broken on single-line message
This is how it looks on multi-line messages:

screen shot 2018-01-23 at 00 52 06

This is how it looks on single-line messages:

screen shot 2018-01-23 at 00 52 22

In #1999, you can see that the top border radius of the [After screenshot](https://user-images.githubusercontent.com/113730/35260098-82eea884-ffd7-11e7-8da3-dca7e9fc30ec.png) are missing too, once the timestamp is correctly aligned at the top.

This is probably due to an `overflow: hidden` or something.

astorije:
Fixes #1998.
Fixes #2000.

Very quick, in-browser attempt at fixing #1998. Not sure if this breaks anything else.

Before | After
--- | ---
screen shot 2018-01-23 at 00 45 05 | screen shot 2018-01-23 at 00 50 24

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1045,7 +1045,6 @@ kbd {
 	word-break: break-word; /* Webkit-specific */
 	display: flex;
 	align-items: flex-start;
-	overflow: hidden;
 	position: relative;
 }
 
@@ -1146,6 +1145,7 @@ kbd {
 	min-width: 0;
 	padding-left: 10px;
 	padding-right: 6px;
+	overflow: hidden; /* Prevents Zalgo text to expand beyond messages */
 }
 
 #loading a,
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1044,6 +1044,7 @@ kbd {
 	word-wrap: break-word;
 	word-break: break-word; /* Webkit-specific */
 	display: flex;
+	align-items: flex-start;
 	overflow: hidden;
 	position: relative;
 }


astorije:
I ended up adding a commit for #2000 directly to this branch, because it's literally on the same line (all in the browser, didn't even fire my editor :successkid:).

Before | After
--- | ---
screen shot 2018-01-23 at 00 52 22 | screen shot 2018-01-23 at 01 06 02

No idea how I break stuff with this though. I see my second commit is kinda reverting https://github.com/thelounge/lounge/commit/74ca130d514e42e1677ca8aabb71688b08fe6330, which isn't cool. What about applying `overflow: hidden` to `.msg .text` instead of `.msg`?

(btw this is why I hate opening issues: I always end up trying to fix it even when I don't have the time to do so, instead of just recording it not to forget 😅)

client/css/style.css
```diff
@@ -1044,7 +1044,7 @@ kbd {
 	word-wrap: break-word;
 	word-break: break-word; /* Webkit-specific */
 	display: flex;
-	overflow: hidden;
```
xPaw: Need to keep this rule, at least on `.text`. Otherwise stuff like zalgo text is going to overflow which we don't want.

astorije: Yeah, I mentioned that in my previous comment. I moved it to `.content` (it had no effect on `.text`, I'm guessing because it's an inline `span` :man_shrugging:).

Zalgo is now prevented again. Until we want to add a tooltip in a `.content`, for example a status message or something, and we'll be screwed again :D
But until then, that should fix it.

PR #2004, opened 2018-01-26
Refactor client-side commands to mirror server-side 

MaxLeiter:
Not sure how this fits into the scheme of plugins... I imagine it makes it easier to add client-side commands

### DIFF
--- a/client/js/commands/join.js
+++ b/client/js/commands/join.js
@@ -1,0 +1,15 @@
+"use strict";
+
+const utils = require("../utils");
+
+exports.input = function(args) {
+	const channel = args[0];
+
+	if (channel) {
+		const chan = utils.findCurrentNetworkChan(channel);
+
+		if (chan.length) {
+			chan.trigger("click");
+		}
+	}
+};
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -7,6 +7,7 @@ const moment = require("moment");
 
 // our libraries
 require("./libs/jquery/stickyscroll");
+const commands = require("./commands");
 const slideoutMenu = require("./slideout");
 const templates = require("../views");
 const socket = require("./socket");
@@ -126,8 +127,12 @@ $(function() {
 			const args = text.substr(1).split(" ");
 			const cmd = args.shift().toLowerCase();
 
-			if (typeof utils.inputCommands[cmd] === "function" && utils.inputCommands[cmd](args)) {
-				return false;
+			if (commands[cmd]) {
+				const output = commands[cmd].input(args);
+
+				if (output === true) {
+					return false;
+				}
 			}
 		}
 
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -10,7 +10,6 @@ var lastMessageId = -1; // eslint-disable-line no-var
 module.exports = {
 	// Same value as media query in CSS that forces sidebars to become overlays
 	mobileViewportPixels: 768,
-	inputCommands: {collapse, expand, join},
 	findCurrentNetworkChan,
 	serverHash,
 	lastMessageId,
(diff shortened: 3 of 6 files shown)


client/js/commands/join.js
```diff
@@ -0,0 +1,16 @@
...
+
+exports.commands = ["join", "goto"];
+
+exports.input = function(args) {
+	const channel = args[0];
+	if (channel) {
+		const chan = utils.findCurrentNetworkChan(channel);
+		if (chan.length) {
+			chan.click();
+			return true;
```
xPaw: Remove `return`, that's gone in master.

client/js/utils.js
```diff
@@ -23,6 +23,20 @@ module.exports = {
 	isOpInChannel,
 };
 
+const commands = [
+	"collapse",
+	"expand",
+	"join",
+].reduce(function(cmds, name) {
```
xPaw: Perhaps this could be automated like templates are? https://github.com/thelounge/lounge/blob/master/client/views/index.js

client/js/utils.js
```diff
@@ -23,6 +23,20 @@ module.exports = {
...
+	const plugin = require(`./commands/${name}`);
+	plugin.cmds.forEach((command) => cmds[command] = plugin);
+	return cmds;
+}, {});
+
+function inputCommands() {
```
xPaw: Can't you export `commands` directly?

MaxLeiter:
@xPaw comments addressed

client/js/commands/join.js
```diff
@@ -0,0 +1,17 @@
+"use strict";
+
+const utils = require("../utils");
+
+exports.commands = ["join", "goto"];
```
xPaw: Is this export actually used?

client/js/lounge.js
```diff
@@ -261,7 +262,8 @@ $(function() {
 			const args = text.substr(1).split(" ");
 			const cmd = args.shift().toLowerCase();
 
-			if (typeof utils.inputCommands[cmd] === "function" && utils.inputCommands[cmd](args)) {
+			if (typeof commands[cmd].input === "function") {
```
xPaw: Change this to `if (commands[cmd])`, otherwise it can be null.

client/js/lounge.js
```diff
@@ -253,8 +254,9 @@ $(function() {
 			const args = text.substr(1).split(" ");
 			const cmd = args.shift().toLowerCase();
 
-			if (typeof utils.inputCommands[cmd] === "function" && utils.inputCommands[cmd](args)) {
-				return false;
+			if (commands[cmd]) {
+				commands[cmd].input(...args);
+				return;
```
xPaw: Keep the `return false`. Didn't notice last time.

MaxLeiter: done

MaxLeiter:
Not sure how to invalidate a review, but after rebase should probably be reviewed again, @xPaw

client/js/lounge.js
```diff
@@ -130,7 +131,8 @@ $(function() {
 			const args = text.substr(1).split(" ");
 			const cmd = args.shift().toLowerCase();
 
-			if (typeof utils.inputCommands[cmd] === "function" && utils.inputCommands[cmd](args)) {
+			if (commands[cmd]) {
+				commands[cmd].input(...args);
```
xPaw: You changed the logic, as `return false` only happened if the command itself returned `true`.

Because we want `/join` command to be sent even if it's handled.

MaxLeiter: so 

```
			if (commands[cmd]) {
                             if (commands[cmd].input(...args)) {
				return false;
      			     }
			}
```

?

MaxLeiter: So, we never want it to `return false`? As we always want the command to be sent?

xPaw: Well with how the commands work right now, we don't want to send collapse/expand commands to the server (as they're handled and sent separately).

MaxLeiter: Fixed

client/js/lounge.js
```diff
@@ -124,8 +125,12 @@ $(function() {
 			const args = text.substr(1).split(" ");
 			const cmd = args.shift().toLowerCase();
 
-			if (typeof utils.inputCommands[cmd] === "function" && utils.inputCommands[cmd](args)) {
-				return false;
+			if (commands[cmd]) {
+				const output = commands[cmd].input(args);
+
+				if (output === true) {
+					return output;
```
xPaw: This should be `return false`

MaxLeiter: Fixed

client/js/commands/join.js
```diff
@@ -0,0 +1,18 @@
...
+
+exports.input = function(args) {
+	const channel = args[0];
+
+	if (channel) {
+		const chan = utils.findCurrentNetworkChan(channel);
+
+		if (chan.length) {
+			chan.trigger("click");
+			return true; // we're done here, server doesn't need to do anything.
```
xPaw: Actually, this again changed behaviour from what it is on master, and we do always pass `/join` command to the server.

So just remove both `return` in this function.

MaxLeiter: Fixed

xPaw:
Gonna close this as I dealt with this in Vue branch. There's minor but important details to get right with this refactor.

PR #2006, opened 2018-01-26
Stop expanding condensed messages on /expand

### LINKED ISSUE #2005: /expand command expands collapsed items, /collapse doesn't collapse expanded items
`/expand` does collapse them (it just toggles them)

YaManicKill:
Fixes #2005

### DIFF
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -55,12 +55,12 @@ function forceFocus() {
 }
 
 function collapse() {
-	$(".chan.active .toggle-button.opened").click();
+	$(".chan.active .toggle-button.toggle-preview.opened").click();
 	return true;
 }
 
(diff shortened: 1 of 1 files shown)


YaManicKill:
I think we do want it in, yeah. It will mean another RC, but I think it's worth putting it in.

xPaw:
We don't need to do another RC for this, imo.

YaManicKill:
I would be strongly against releasing 2.7.0 with this in it without another RC.

The point of an RC is to say "This is probably what we are going to release, but let's go through the (almost the same) release process to check it all works fine, and make sure there's nothing glaring in this". If we shortcut that for some bugs, then the whole process is pointless, IMO.

PR #2013, opened 2018-01-30
Fix #1959: Do not block /join command from being sent

### LINKED ISSUE #1959: Trying to join a channel after you've been kicked/banned results in nothing
This bug occurs on the next release branch (master). 

How to reproduce: Launch Lounge, join a channel. Get someone to set the channel to invite only and ask them to kick you. Type `/join #channelname` and you'll see that instead of sending the command, Lounge switches to the channel tab instead.

I thought this was an issue with the autocomplete but it's not, if you type a space after the channel name, Lounge will still switch to the channel tab. If you want to actually join the channel, you must close the tab, which will result in the "you are not in channel" error (probably because closing the tab sends a `/PART`).

xPaw:
@RockyTV, @jimmybrancaccio can you guys confirm?

### DIFF
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -70,7 +70,6 @@ function join(args) {
 		const chan = findCurrentNetworkChan(channel);
 		if (chan.length) {
 			chan.click();
-			return true;
 		}
 	}
 }


xPaw:
@McInkay It would allow join command be sent to the network instead of being blocked in the client.

xPaw:
> Do we definitely always want to send the join command?

Sure. This basically reverts to the behaviour we had before.

>  Do we intercept this in our server at all

No, it's sent as is.

PR #2015, opened 2018-01-30
Remove deprecated jQuery calls

### LINKED ISSUE #2011: Remove calls to deprecated features of jQuery v3.3+
According to the [jQuery v3.3.0 release notes](https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/) (that we switched to in #1994), a bunch of event-related methods got deprecated: https://github.com/jquery/jquery/blob/305f193/src/event/alias.js#L10-L12

I noticed at least one occurrence of `.focus()` in our client code, so it might be worth doing a pass at those, to be proactive. We're also using a couple small jQuery plugins, so it would be good to go through those as well.

xPaw:
Fixes #2011.

- Replace `unbind` with `off`
- Replace `click` and `focus` with `trigger`
- Replace `keyup` and `keypress` with `on`
- Replace all `attr`, `removeAttr`, `removeProp` with `prop` https://stackoverflow.com/a/5876747/2200891

### DIFF
--- a/client/js/join-channel.js
+++ b/client/js/join-channel.js
@@ -20,8 +20,8 @@ function toggleButton(network) {
 	// Toggle content of tooltip
 	const tooltip = network.find(".add-channel-tooltip");
 	const altLabel = tooltip.data("alt-label");
-	tooltip.data("alt-label", tooltip.attr("aria-label"));
-	tooltip.attr("aria-label", altLabel);
+	tooltip.data("alt-label", tooltip.prop("aria-label"));
+	tooltip.prop("aria-label", altLabel);
 }
 
 function closeForm(network) {
@@ -44,7 +44,7 @@ function openForm(network) {
 	}
 
 	// Focus the "Channel" field even if the form was already open
-	form.find(".input[name='channel']").focus();
+	form.find(".input[name='channel']").trigger("focus");
 }
 
 sidebar.on("click", ".add-channel", function(e) {
@@ -69,7 +69,7 @@ sidebar.on("submit", ".join-form", function() {
 	const keyString = key.val();
 	const chan = utils.findCurrentNetworkChan(channelString);
 	if (chan.length) {
-		chan.click();
+		chan.trigger("click");
 	} else {
 		socket.emit("input", {
 			text: `/join ${channelString} ${keyString}`,
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -44,7 +44,7 @@ $(function() {
 
 	viewport.on("click", ".rt", function(e) {
 		var self = $(this);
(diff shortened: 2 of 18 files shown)


client/js/join-channel.js
```diff
@@ -20,8 +20,8 @@ function toggleButton(network) {
 	// Toggle content of tooltip
 	const tooltip = network.find(".add-channel-tooltip");
 	const altLabel = tooltip.data("alt-label");
-	tooltip.data("alt-label", tooltip.attr("aria-label"));
-	tooltip.attr("aria-label", altLabel);
+	tooltip.data("alt-label", tooltip.prop("aria-label"));
+	tooltip.prop("aria-label", altLabel);
```
astorije: I've read the SO link you gave, but still unclear/not convinced: why are we changing from `attr` to `prop`? AFAIK `attr` is not deprecated.

client/js/lounge.js
```diff
@@ -297,14 +297,14 @@ $(function() {
 		utils.setNick(sidebar.find(".chan.active").closest(".network").data("nick"));
 	}
 
-	$("#nick-value").keypress(function(e) {
+	$("#nick-value").on("keypress", function(e) {
```
astorije: Might as well use arrow functions while you're at it, but nbd.

client/js/socket-events/topic.js
```diff
@@ -7,6 +7,6 @@ const helpers_parse = require("../libs/handlebars/parse");
 socket.on("topic", function(data) {
 	const topic = $("#chan-" + data.chan).find(".header .topic");
 	topic.html(helpers_parse(data.topic));
-	// .attr() is safe escape-wise but consider the capabilities of the attribute
-	topic.attr("title", data.topic);
+	// .prop() is safe escape-wise but consider the capabilities of the attribute
```
astorije: What does this comment even mean lol?

client/js/join-channel.js
xPaw: As far as I understood it, `prop` is used to directly access DOM properties, which we are doing here, and should generally be faster.

`prop` is basically `element[ property ] = value;` while `attr` is `element.setAttribute` and whole lot of other magic.

I changed it everywhere in the code to be consistent, as we have mixed attrs/props before.

client/js/lounge.js
xPaw: Didn't want to break anything unintentionally.

astorije: Argh yes, I always forget jQuery is funky when it comes to `this` reference.

astorije:
Aaaand conflict :)

McInkay:
More conflicts!

PR #2020, opened 2018-01-30
Wait for server response when parting channels

### LINKED ISSUE #1991: Do not close a channel's window on immediately when a user runs /part
Currently when you type /part in lounge it immediately closes the channel's window client side. Some networks require users to be in a channel and will deny PART commands for that channel. The result is that the user is still in the channel on the server and still receives messages but is unable to properly reopen the channel's window.

Lounge should instead wait until the server sends the PART back before it closes the channel window.

xPaw:
@crosma mind testing this?

### DIFF
--- a/src/plugins/inputs/part.js
+++ b/src/plugins/inputs/part.js
@@ -27,17 +27,17 @@ exports.input = function(network, chan, cmd, args) {
 		return;
 	}
 
-	network.channels = _.without(network.channels, target);
-	target.destroy();
-	this.emit("part", {
-		chan: target.id,
-	});
-	this.save();
-
-	if (target.type === Chan.Type.CHANNEL) {
-		if (network.irc) {
-			network.irc.part(target.name, partMessage);
-		}
+	// If target is not a channel or we are not connected, instantly remove the channel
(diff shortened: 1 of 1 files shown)


astorije:
@xPaw, you already have a conflict 😁

src/plugins/inputs/part.js
```diff
@@ -27,17 +27,17 @@ exports.input = function(network, chan, cmd, args) {
...
+	// Otherwise send part to the server and wait for response
+	if (target.type !== Chan.Type.CHANNEL || !network.irc || !network.irc.connection || !network.irc.connection.connected) {
+		network.channels = _.without(network.channels, target);
+		target.destroy();
+		this.emit("part", {
+			chan: target.id,
+		});
+		this.save();
+	} else {
+		network.irc.part(target.name, partMessage);
```
McInkay: No save for this one?

xPaw: It gets saved in part event handler.

McInkay: Meh, alright.

PR #2022, opened 2018-01-31
Define which message types should not be logged

xPaw:
`whois` log is currently logged as `undefined whois` which isn't at all useful.


src/models/msg.js
```diff
@@ -63,4 +63,9 @@ Msg.Type = {
 	BANLIST: "ban_list",
 };
 
+Msg.NonLoggableTypes = {};
+Msg.NonLoggableTypes[Msg.Type.BANLIST] = true;
+Msg.NonLoggableTypes[Msg.Type.MOTD] = true;
+Msg.NonLoggableTypes[Msg.Type.WHOIS] = true;
```
MaxLeiter:
```
Msg.NonLoggableTypes = { [Msg.Type.BANLIST]: true, ...}
```
maybe?

xPaw: Doesn't work on node 4 afaik.

astorije:
```sh
$ nvm use 4
Now using node v4.8.4 (npm v2.15.11)
$ node
> a = 42
42
> { [a]: 1}
{ '42': 1 }
```

:man_shrugging:

astorije: Why not add a method to the `Msg` class instead that would do:

```js
function isLoggable() {
	return this.type !== Msg.Type.MOTD &&
		this.type !== Msg.Type.WHOIS;
}
```

And in `writeUserLog`:

```js
if (!msg.isLoggable()) {
	return false;
}
```

It seems more idiomatic to me, no?
(I went for `isLoggable` and not `isNotLoggable` because it's fairly common practice to avoid negated boolean functions)

Bonus, you can get rid of `const Msg = require("./msg");` 😁

xPaw:
Changed it to what @astorije suggested.

PR #2030, opened 2018-02-03
Fix messages not being condensed correctly

### LINKED ISSUE #2028: New status messages ignore rendering preferences
Since installing, `thelounge` runs with the `Condense status messages together` option selected. Just noticed that some channels are full of inline status messages that are not condensed and are unaffected by the `Status messages` options. When the webpage is reloaded, previous status messages that were not condensed get rendered correctly. Choosing the `Hide all status messages` option affects only messages received before loading the page. It seems newly received status messages are not filtered by the `Status messages` preference.

Here is a screenshot with a corresponding [DOM code](https://dpaste.de/pYUr).

untitled-2

xPaw:
Fixes #2028.

This was caused by jQuery conversion I did a couple of days ago, and this is the only place where `.data` was not used.

### DIFF
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -58,7 +58,7 @@ function processReceivedMessage(data) {
 	render.appendMessage(
 		container,
 		targetId,
-		channel.prop("data-type"),
+		channel.data("type"),
 		data.msg
 	);
 


xPaw:
@astorije `data-type` is not a property, `.prop("dataset").type` would work there.

astorije:
Gotcha, thanks!

astorije:
(Removing the `skip changelog` tag, I'll group the refs in the #2015 PR changelog entry)

PR #2031, opened 2018-02-03
Improve monospace formatting in messages

astorije:
This effectively styles it just like to `code` syntax used in the Help window, such as in the list of available commands.

Before | After
--- | ---
screen shot 2018-02-03 at 12 49 23 | screen shot 2018-02-03 at 12 49 04

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -86,7 +86,8 @@ textarea#user-specified-css-input {
 	font-family: Consolas, Menlo, Monaco, "Lucida Console", "DejaVu Sans Mono", "Courier New", monospace;
 }
 
-code {
+code,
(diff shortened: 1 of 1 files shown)


xPaw:
@astorije as I mentioned on IRC before, this most likely messes up colors when using IRC colors, so that needs figuring out.

astorije:
Yeah, I'm not sure what we can do about this. I do want to keep the same color as other `code` though (and grey on grey is not good).
@xPaw, is it that big of a deal? IMO this PR is an improvement already.

xPaw:
@astorije Probably not that big of a deal because this formatter is rather new, and I doubt anyone is going to use it too much with colors.

PR #2033, opened 2018-02-04
Fix default theme not being correct

xPaw:
*what did i get myself into*

`prop` returns full URL and not what was specified in html. We should probably replace that with a `data` attribute in v3.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -22,11 +22,13 @@ const options = {
 	notifyAllMessages: false,
 	showSeconds: false,
 	statusMessages: "condensed",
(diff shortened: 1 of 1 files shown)


astorije:
> `prop` returns full URL and not what was specified in html.

Wow, that's nasty. Any other places where this could be an issue?

astorije:
I'm not sure I'm following, `master` is working fine for me, what do I need to do to trigger the bug?

xPaw:
@astorije Change `theme` in config, load Lounge in incognito. If you're in private mode you have to login first (because options initialise is after auth), you will see it will change theme incorrectly.

PR #2035, opened 2018-02-05
Ensure packages loaded are directories

astorije:
Suppresses this warning for example:

screen shot 2018-02-05 at 01 31 38


src/plugins/packages/index.js
```diff
@@ -39,34 +39,40 @@ function loadPackages() {
...
-			if (packageFile.type === "theme") {
-				themes.addTheme(packageName, packageFile);
-			}
-
-			if (packageFile.onServerStart) {
-				packageFile.onServerStart(packageApis(packageName));
-			}
+			fs.stat(Helper.getPackageModulePath(packageName), function(err2, stat) {
+				if (err2 || !stat.isDirectory()) {
+					return;
```
McInkay: Can we stick in a debug warning for package not existing, because it could be useful for some errors people might have at some point.

xPaw:
With Yarn, we could look up installed packages from `package.json`

src/plugins/packages/index.js
astorije: Done.

astorije:
Alright, so I'm done with this _for now_.

This PR ends up doing a bunch of things related to each other.

- Dependencies are now loaded from `package.json`. @xPaw, we don't need to wait for Yarn PR for that, we do save in `package.json`. This effectively solves the original minor thing I was trying to fix (and which frankly affects only those who tested the Yarn PR probably lol). Also this ensures that a package that installs other dependencies will not try to load the package's dependencies as packages for The Lounge.

- Displays logs when a package is correctly loaded or not:
  screen shot 2018-02-08 at 00 35 16

- Add tests for the 3 existing methods of the `packages/index.js` file. As it is, I am not very happy with either the code or the tests. They're both not very clean, way too stateful, rely on side-effects, etc. Next thing I'll do is clean that up as I know what I want to try, I should be able to do that tomorrow, but I didn't want to hold this PR if I'm not, because...

- ... I'm not happy with the tests, but they did highlight a rather substantial bug with the packages interface that I fixed along the way, and that's the only reason why I'm not pulling this PR out of the 2.7.1 milestone:

  Say you have the following package:

  ```js
  // package.json
  {
    "name": "thelounge-package-foo",
    "private": true,
    "main": "index.js",
    "thelounge": {
      "type": "package"
    }
  }
  
  // index.js
  module.exports = {
	  onServerStart(apis) {
		  apis.Stylesheets.addFile("style.css");
	  },
  };
  ```

  On `master`, the `thelounge` property would be searched from the `index.js` file and not `package.json`, so `loadPackages` would consider this package is not for The Lounge and would ignore it and warn about it.
  In this PR, I fixed this by requiring both `package.json` and the package itself (which requires whatever is in the `main` property). In the case of themes, that would load the same file because our docs recommend `"main": "package.json"` but that's fine. Actually, now that I think about it, I am going to allow themes not to set `main` because it's kind of silly anyway.

src/helper.js
```diff
@@ -75,7 +75,7 @@ function setHome(newPath) {
 	configPath = path.join(homePath, "config.js");
 	usersPath = path.join(homePath, "users");
 	storagePath = path.join(homePath, "storage");
-	packagesPath = path.join(homePath, "packages", "node_modules");
```
astorije: `packagesPath` now does not include the `node_modules` part, because that means we need to compute the parent every time we want to access `package.json` of the packages folder. In practice, we should rarely have to rely on the `node_modules` location, `require` should do things nicely in _most_ case.

src/helper.js
```diff
@@ -144,7 +144,7 @@ function getPackagesPath() {
 }
 
 function getPackageModulePath(packageName) {
-	return path.join(Helper.getPackagesPath(), packageName);
+	return path.join(Helper.getPackagesPath(), "node_modules", packageName);
```
astorije: For dependencies loaded in local `node_modules`, it is possible to do `require("foo/package")` and `require("foo/package.json")`, which always returns the `package.json` content so there is no need to reconstruct the path.

I was hoping there was a way to tell `require` to use a different search path (in fact, [it's sorta possible with Webpack-specific `require.context` 😩](https://github.com/thelounge/lounge/blob/14847260b699a940627f03293ffb2cd00fc8ddee/client/views/index.js#L3-L7)) then we could just do `require(packageName)` and ``packageName(`${packageName}/package.json`)`` instead of manually constructing this, but so far no luck.

src/plugins/packages/index.js
```diff
@@ -35,38 +35,43 @@ function getPackage(name) {
...
+
+		if (packageInfo.type === "theme") {
+			themes.addTheme(packageName, packageInfo);
+		}
+
+		if (packageFile.onServerStart) {
+			packageFile.onServerStart(packageApis(packageName));
+		}
+
+		setTimeout(() => log.info(`Package ${colors.bold(packageName)} loaded`));
```
astorije: This `setTimeout` is very much subpar, but I don't want the logs to show first (because they're loaded before anything. In fact, I know how to do that in a much better way (and also address a rather nasty security concern I have right now), but in the meantime, I figured it would do. I'll clean up the tests as well (they need to use `done()` for now, not great).

McInkay: So wait, why are you doing this? You want them to not show before what?

McInkay: Oh wait, I see. You want them logged after the server is up. I'm not sure that's a great idea.

It makes it look like it turns the server on and then loads packages, rather than the other way around, which is misleading.

Also, all we need is 1 more async thing to happen before these and the logs will be messed up again. I say just log them in the order you get them.

PR #2036, opened 2018-02-05
Report server errors when preview fetch fails

xPaw:
This is gonna report everything: server time outs, DNS failures, wrong HTTP codes, etc. This is pretty useful for knowing why fetch failed, but we might need to filter out some specific errors out (e.g. domain lookup failure).

### DIFF
--- a/client/views/msg_preview.tpl
+++ b/client/views/msg_preview.tpl
@@ -28,5 +28,10 @@
 			<div class="body" title="{{body}}">{{body}}</div>
 		</a>
 	{{/equal}}
+	{{#equal type "error"}}
+		{{#equal error "message"}}
+			<em>Server returned an error: {{message}}</em>
+		{{/equal}}
+	{{/equal}}
 </div>
 {{/preview}}
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -38,7 +38,14 @@ module.exports = function(client, chan, msg) {
 	})).slice(0, 5); // Only preview the first 5 URLs in message to avoid abuse
 
 	msg.previews.forEach((preview) => {
-		fetch(preview.link, function(res) {
+		fetch(preview.link, function(res, err) {
+			if (err) {
+				preview.type = "error";
+				preview.error = "message";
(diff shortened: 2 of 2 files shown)


astorije:
I'm not a big fan of shoving errors to the user bluntly like that. It's likely that it won't be non-tech-people-friendly, possibly not actionable, etc.

UI should say something a bit more generic like "This image could not be loaded" or "There was an error when loading this image" or whatever. Then if we want to follow-up on the actual error, we have a few possibilities:

- `console.log`/`console.error` in the client
- Add a `debug` argument in the config
- `log.error` on the server output
- Do like the loading page, i.e. human-friendly message with a "more details" kind of thing.

It might also be a good time to think about error logging on the browser in general. We can easily do that on the backend using `log.error` + we know the expected audience will be admin and most likely a technical person, while on the UI, things are a bit more tricky.

xPaw:
> UI should say something a bit more generic like "This image could not be loaded"

This is really not any better than not displaying an error in the first place.

`console` logging on the client is not exactly the way to go either since it's not a client error (can happen without any clients connected).

`debug.prefetch` is probably a good toggle for this though, just do what this PR does now, but send the `error` only when debug option is enabled.

MiniDigger:
even without debug enabled it should indicate that the prefetch failed, but we can use something that is more targeted at non technical users like "There was an error when loading this image"

astorije:
IMO the best way to handle errors is human-friendly messages that mean something (even better if they're actionable). That means instead of simply dumping an error code or trace, we interpret this into a sentence.
That means knowing in advance errors and handle them individually (or by groups). If we don't want or can't do that, then I think generic error(s) still works better than code/trace dumps. And I think it's also better than not showing anything, because to the user it may look like it's still loading or stuck somewhere.

`debug.prefetch` might be a good way to go, because there is probably more than one thing we want to link to that debug option.

xPaw:
Okay I changed it to say a generic message. It puts the actual error in there as well, but it's hidden with CSS by default. So if it's really needed to be seen it can be achieved with custom css or just by looking in devtools.

PR #2039, opened 2018-02-06
Fix duplicate chat containers and unread markers when reconnecting

### LINKED ISSUE #1721: Multiple 'New message' markers appearing
Running 56a9ad1 when it happened, not sure what caused it, or how to reproduce. It's a pretty recent commit, so @xPaw didn't think it would be fixed yet, so making this issue. 

They appear in the correct position (ie. before new messages), but way too many of them. When clicking on the channel again they are all at the bottom, and not hidden. Will update when I can reproduce it.

image

xPaw:
Fixes #1721.

When reconnecting, it inserted the same existing channels into chat container again, which means selecting channel by id (`#chat-XXX`) returned 2 elements instead of 1, and jQuery being nice, moved unread marker from second container into the first one, causing it to have 2 unread markers in one channel window. After that, the issue just extrapolated with `.unread-marker` selecting 2 markers, duplicating them, and resulting in 4, etc.

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -228,19 +228,21 @@ function renderNetworks(data, singleNetwork) {
 		newChannels = channels;
 	}
 
-	chat.append(
-		templates.chat({
-			channels: channels,
-		})
-	);
+	if (newChannels.length > 0) {
+		chat.append(
+			templates.chat({
+				channels: newChannels,
+			})
+		);
 
-	newChannels.forEach((channel) => {
-		renderChannel(channel);
+		newChannels.forEach((channel) => {
+			renderChannel(channel);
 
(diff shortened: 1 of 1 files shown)


client/js/render.js
```diff
@@ -228,11 +228,13 @@ function renderNetworks(data, singleNetwork) {
...
 
-	chat.append(
-		templates.chat({
-			channels: channels,
-		})
-	);
+	if (newChannels.length > 0) {
+		chat.append(
+			templates.chat({
+				channels: newChannels,
```
xPaw: This line single-handedly caused the issue. `channels` vs `newChannels`.

PR #2047, opened 2018-02-08
Unbind image events after the image is loaded

xPaw:
We don't need to "leak" memory by keeping them around


xPaw:
As you can see, events are bound with names, like `load.preview`

astorije:
So it's an arbitrary name that you give? I'm not sure to understand what that gives us, care to explain?

xPaw:
Allow me to unbind all the events without having to name each one of them.

We already do this in sticky scroll code with `.sticky`.

PR #2049, opened 2018-02-09
Sync network status and security to client UI

xPaw:
Can't find any good icons in font awesome to represent these states.

TODO: Consider `127.0.0.1` connections secure

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -268,6 +268,14 @@ kbd {
 #sidebar .chan.lobby::before,
 #chat .lobby .title::before { content: "\f0a0"; /* http://fontawesome.io/icon/hdd-o/ */ }
 
+#sidebar .not-secure .chan.lobby::before {
+	content: "\f09c"; /* https://fontawesome.com/v4.7.0/icon/unlock/ */
+}
+
+#sidebar .not-connected .chan.lobby::before {
+	content: "\f127"; /* https://fontawesome.com/v4.7.0/icon/chain-broken/ TODO v5: unlink */
+}
+
 .context-menu-query::before,
 .context-menu-action-query::before,
 #sidebar .chan.query::before,
@@ -557,15 +565,33 @@ kbd {
 	width: 100%;
 }
 
-#sidebar .chan:first-child {
+#sidebar .chan.lobby {
 	color: #84ce88;
 	font-size: 15px;
 	font-weight: bold;
 }
 
-#sidebar .chan:first-child:hover,
-#sidebar .chan:first-child.active {
+#sidebar .chan.lobby:hover,
+#sidebar .chan.lobby.active {
 	color: #c0f8c3;
+}
+
+#sidebar .not-secure .chan.lobby {
+	color: #f39c12;
+}
+
+#sidebar .not-secure .chan.lobby:hover,
+#sidebar .not-secure .chan.lobby.active {
+	color: #f8c572;
+}
+
+#sidebar .not-connected .chan.lobby {
+	color: #e74c3c;
+}
+
+#sidebar .not-connected .chan.lobby:hover,
+#sidebar .not-connected .chan.lobby.active {
+	color: #f1978e;
 }
 
 #sidebar .chan::before,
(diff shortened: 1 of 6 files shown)


client/css/style.css
```diff
@@ -268,6 +268,16 @@ kbd {
 #sidebar .chan.lobby::before,
 #chat .lobby .title::before { content: "\f0a0"; /* http://fontawesome.io/icon/hdd-o/ */ }
 
+#sidebar .not-secure .chan.lobby::before {
+	content: "\f09c"; /* http://fontawesome.io/icon/unlock/ */
+	color: #ff1744;
+}
+
+#sidebar .not-connected .chan.lobby::before {
+	content: "\f1e6"; /* http://fontawesome.io/icon/plug/ */
```
astorije: Not a huge fan of this icon. I seems pretty common to see [`chain-broken`](https://fontawesome.com/v4.7.0/icon/chain-broken/) for broken connections/links (even nicer with [its v5 equivalent](https://fontawesome.com/icons/unlink?style=solid)), or we could go with something more generic like [`ban`](https://fontawesome.com/v4.7.0/icon/ban/) or [`exclamation`](https://fontawesome.com/v4.7.0/icon/exclamation/).

client/css/style.css
```diff
@@ -268,6 +268,16 @@ kbd {
 #sidebar .chan.lobby::before,
 #chat .lobby .title::before { content: "\f0a0"; /* http://fontawesome.io/icon/hdd-o/ */ }
 
+#sidebar .not-secure .chan.lobby::before {
+	content: "\f09c"; /* http://fontawesome.io/icon/unlock/ */
+	color: #ff1744;
```
astorije: Those reds are very aggressive lol. Plus it seems to me we don't use this shade of red already. Mind using `#e74c3c` with is a bit more pleasing to the eye (but still red)? We use that color in a few places, like errors in form, connection lost banner, etc.

astorije: Actually, I have 2 suggestions on top of the previous that I think might look better:

- For insecure connection, use the `#f39c12` orange we have. After all, some networks are insecure for a good reason, not an error on the client or something the user didn't do well, so marking it as an error might be too strict (and might be confusing between the 2 states).

- Changing the color of the entire line looks nicer, more homogenous.

That would look like this:

screen shot 2018-02-18 at 00 43 06<br>screen shot 2018-02-18 at 00 44 40

client/views/network.tpl
```diff
@@ -1,5 +1,11 @@
 {{#each networks}}
-<section id="network-{{id}}" class="network name-{{slugify name}} {{#if serverOptions.NETWORK}}network-{{slugify serverOptions.NETWORK}}{{/if}}" data-id="{{id}}" data-nick="{{nick}}" data-options="{{tojson serverOptions}}">
+<section
+	id="network-{{id}}"
+	class="network name-{{slugify name}} {{#if serverOptions.NETWORK}}network-{{slugify serverOptions.NETWORK}}{{/if}} {{#unless status.connected}}not-connected{{/unless}} {{#unless status.secure}}not-secure{{/unless}}"
+	data-id="{{id}}"
+	data-nick="{{nick}}"
+	data-options="{{tojson serverOptions}}"
+>
```
astorije: I think we should have a tooltip (opening towards the right, "east" side) for not-secure and not-connected types, to inform the user about what those icons / colors mean.

xPaw: It would probably be a pain to get tooltips in there correctly since the icons are on `:before` of the entire title.

client/css/style.css
xPaw: Feel free to make a commit with all the changes.

client/css/style.css
xPaw: Yeah the icons are crap because I couldn't find anything good in FA.

For example, for insecure I'd really like to have a lock with strike through line, not whatever it is now.

`unlink` is probably the best we have for disconnected.

astorije:
@xPaw, mind fixing the conflict?

client/views/network.tpl
astorije: Yeah, I tried and quickly gave up...

astorije:
@xPaw, I added a commit, what do you think?

screen shot 2018-02-19 at 01 43 45

It's not ideal in many ways, but at this point that's the best we can do with the current UI imo. That should hold until a broader UI redo.

xPaw:
@astorije Yeah that's probably the best we can currently do.

PR #2055, opened 2018-02-11
Fix border after nickname not taking full height

### LINKED ISSUE #2054: Border between nick and content does not take full height
This is fine in 2.7.0, but broken in 2.7.1, this is caused by 6bfd6ed47308c50f231f18edb15fb82d69159505. cc @astorije

xPaw:
Fixes #2054 caused by 6bfd6ed47308c50f231f18edb15fb82d69159505

I moved border instead of changing `align-items` to prevent this issue:

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1137,7 +1137,6 @@ kbd {
 }
 
 #chat .from {
-	border-right: 1px solid #f6f6f6;
 	color: #b1c3ce;
 	padding-right: 10px;
 	text-align: right;
@@ -1152,6 +1151,7 @@ kbd {
 	min-width: 0;
 	padding-left: 10px;
 	padding-right: 6px;
+	border-left: 1px solid #f6f6f6;
 	overflow: hidden; /* Prevents Zalgo text to expand beyond messages */
 }
 
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -39,7 +39,7 @@ body {
 }
 
 /* Borders */
-#chat .from,
+#chat .content,
 #windows .header,
 #chat .user-mode::before,
 #chat .sidebar {
--- a/client/themes/zenburn.css
+++ b/client/themes/zenburn.css
@@ -65,7 +65,7 @@ body {
 }
 
 /* Borders */
-#chat .from,
+#chat .content,
 #windows .header,
 #chat .user-mode::before,
 #chat .sidebar {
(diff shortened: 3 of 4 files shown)


astorije:
This leaves a hole when the message is empty. For example, Freenode's MOTD does that:

screen shot 2018-02-11 at 16 50 29

Same with InspIRCd test network:

screen shot 2018-02-11 at 16 51 33

I'm guessing there are other cases.
Maybe empty messages should always be rendered with `&nbsp;`, or have a min height or something?

xPaw:
> Maybe empty messages should always be rendered with &nbsp;

Pretty silly to write code to solve our styling problems. We could do `.text:empty:after { content: " " };` which does the trick, but I feel like this is a silly solution.

astorije:
Well, it's not uncommon to let the browser set the correct height based on whitespace content.
I get why it feels silly (especially the in-markup version), but any attempt to style it "correctly" by setting min-height or whatnot ourselves will feel just as silly, and will _never_ end up working correctly, depending on resolution, font, size, etc.

I'm open to alternatives obviously, I just don't see one.

McInkay:
Is the message completely empty, or does it have spaces in it?

xPaw:
@McInkay It's empty, as adding a space with CSS fixes the border.

xPaw:
I trimmed MOTD message to avoid the empty line at the end, as that wasn't really supposed to be printed.

dgw:
For the record, empty "real" messages in a channel or query still cause a gap in the border line.

It's probably fixable with some `min-height` value, I'll have to look into it.

dgw:
`min-height` didn't really go anywhere. On the plus side, I also discovered the trade-off between the line being too short (before) and the line always being as tall as the message:

thelounge-message-border-too-long

Ideally the border would be the color of the highlight only for the height of the `from` text, and the normal color for the rest of the message, but I'm not sure that's even possible so I hesitate to open a new issue for this…

dyyeid:
โดนแฮ้กช่วยด้วย

เมื่อ 26 ก.พ. 2018 6:39 ก่อนเที่ยง dgw <notifications@github.com> เขียนว่า:

min-height didn't really go anywhere. On the plus side, I also discovered the trade-off between the line being too short (before) and the line always being as tall as the message:

[thelounge-message-border-too-long]<https://user-images.githubusercontent.com/164140/36647980-a1beccfe-1a52-11e8-8f9b-52d8ff539811.png>

Ideally the border would be the color of the highlight only for the height of the from text, and the normal color for the rest of the message, but I'm not sure that's even possible so I hesitate to open a new issue for this…

PR #2058, opened 2018-02-13
Track channel state to allow removing channels user is not in

xPaw:
Some discussion in #1959.

We recently changed channel part to await server response, but if you were kicked from a channel, PART would never come and the channel would be left dangling.

This tracks channel state (parted or joined) to allow instantly removing channels we're already parted in.

cc @RockyTV for test

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -15,6 +15,11 @@ Chan.Type = {
 	SPECIAL: "special",
 };
 
+Chan.State = {
+	PARTED: 0,
+	JOINED: 1,
+};
+
 let id = 1;
 
 function Chan(attr) {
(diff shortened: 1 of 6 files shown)


RockyTV:
Works as expected. When, for some reason, you can't join a channel and you set Lounge to auto join it on connect, you can close the tab without it trying to send a `PART` message. Same thing for when you are kicked from the channel.

src/models/chan.js
```diff
@@ -15,6 +15,11 @@ Chan.Type = {
 	SPECIAL: "special",
 };
 
+Chan.State = {
```
xPaw: This can be expanded for queries (#1700) later on to state if user is online or not.

PR #2059, opened 2018-02-13
Render MOTD with a single message

xPaw:
Instead of splitting it on server and creating `Msg` for every line, just let keep the newlines in `parse` and let CSS render it correctly.


client/js/libs/handlebars/ircmessageparser/parseStyle.js
```diff
@@ -19,7 +19,8 @@ const colorRx = /^(\d{1,2})(?:,(\d{1,2}))?/;
 const hexColorRx = /^([0-9a-f]{6})(?:,([0-9a-f]{6}))?/i;
 
 // Represents all other control codes that to be ignored/filtered from the text
-const controlCodesRx = /[\u0000-\u001F]/g;
+// This regex allows line feed character
+const controlCodesRx = /[\u0000-\u0009\u000B-\u001F]/g;
```
astorije: Do you think you could add a test for this regex change? 🙏 🙏

xPaw: Added.

PR #2060, opened 2018-02-13
Fix themes not loading

xPaw:
Broken by e4701be7081e97a6a4e243f104e412be8a0a2c58

`getModuleInfo` returned `info.thelounge` before but it was removed.


astorije:
Yep, I knew I was forgetting to commit something...

PR #2063, opened 2018-02-14
Provide exact version into `npm install` command

xPaw:
Should avoid dumb `package.json` and caching issues.

### DIFF
--- a/src/command-line/install.js
+++ b/src/command-line/install.js
@@ -61,7 +61,7 @@ program
 					"--no-progress",
 					"--prefix",
 					packagesPath,
-					packageName,
+					`${packageName}@${json.version}`,
 				],
 				{
 					// This is the same as `"inherit"` except `process.stdout` is ignored


xPaw:
@McInkay Yeah it's hardcoded to `latest` in `package-json`

astorije:
@xPaw, should we do ```${packageName}@$latest``` here to make sure updating packaging works? I don't know, not sure.

xPaw:
@astorije `json.version` *is* latest as returned by the NPM web api.

astorije:
By latest, I did mean the keyword `latest` as a string, not a version number. Doesn't `json.version` returns the latest version number?

xPaw:
Yeah I understood what you meant. `package-json` which we use [explicitly requests](https://github.com/sindresorhus/package-json/blob/2312d0437c16d7117d354ae76f640c5ac25fe0e0/index.js#L15) `latest` via HTTP already so we will get the exact latest version number, without having to rely on npm CLI to do it.

astorije:
@xPaw, I'm still confused as to why we don't do `@latest` ourselves, instead of relying on what `package-json` does behind the scenes.

PR #2064, opened 2018-02-14
Make channel list slideout trigger when starting dragging on the left half of the screen.

xyzz:
This makes it easier to drag the channel list out on e.g. iOS, whereas the current approach often results in the Safari previous page gesture.

### DIFF
--- a/client/js/libs/slideout.js
+++ b/client/js/libs/slideout.js
@@ -31,7 +31,7 @@ module.exports = function slideoutMenu(viewport, menu) {
 
 		menuWidth = parseFloat(window.getComputedStyle(menu).width);
 
-		if ((!menuIsOpen && touch.screenX < 50) || (menuIsOpen && touch.screenX > menuWidth)) {
+		if (!menuIsOpen || (menuIsOpen && touch.screenX > menuWidth)) {
 			touchStartPos = touch;
 			touchCurPos = touch;
 			touchStartTime = Date.now();


xyzz:
These apps assume users start dragging on the edge, if you do that in a web app like the lounge, it would go to the previous page. If you increase the cutoff, users still can't drag from the edge and have to pixel hunt the right position, ideally I'd remove the cutoff check completely but not sure if that would break anything.

christer88:
As it is today it's difficult to swipe for channel list and not load previous webpage.

I'm in favour of enabling this on the entire screen. Let's make it easy to swipe! 👍

astorije:
If you're using The Lounge on Safari iOS, I really really recommend pinning it to the homescreen. It comes with a few extras, and that swipe-to-previous issue disappears. I'm not arguing against this PR (I need to play with it but it should be fine), just clarifying :)

astorije:
I've tested this and it feels pretty natural, I like it. At this point, I'm in favor of enabling it on the entire screen.

If later improvements are needed, I think they're not related to this specific setting but other things like maybe increasing the distance required to open the channel list, the small width of channel list on mobile, etc.
So in short, yes to this, yes to having it full size, and we can deal with other issues in future work.

xyzz:
The problem with pinning to homescreen is that if you switch to another app, or just go back to the home screen, the app is unloaded from memory completely. So it has to reload everything when you open it again and it gets really annoying.

maxpoulin64:
If I recall correctly, part of why I did it this way originally is that I kept accidentally opening the menu when scrolling up or down fast. When the menu is moving, the script captures all touch events so nothing else on the page responds, which will also cancel your scrolling. I think scrolling the app left-handed in particular was pretty bad.

It looks trivial at first hand to differentiate scrolling the chat and opening the menu, but we don't have any velocity information available at hand. It's not horribly hard but back then I didn't feel like implementing the whole thing required to accumulate data points to get that information. Plus the necessity of multiple data points means latency which I wasn't too fond of and what most menu libraries out there suffered from.

christer88:
Pinning it to homescreen require SSL, and I have been lazy. Also would like to use it with Opera Mini to save some data. :)

astorije:
Re: pinning to homescreen, well, I learned something new today 😁
I run behind HTTPS so I didn't even realize that, and the reload is indeed extremely annoying, but I didn't know it wasn't happening in the browser. That might totally convince me to go back to a browser tab 😕

Hey @maxpoulin64, thanks for the details!

Funny you mention canceling the scroll, because when swiping to open the menu, scrolling is not disabled, so if you start opening the menu and don't go perfectly horizontal, the chat scrolls with it. I've always found that annoying, but never realized you took care of the other direction (canceling opening the menu when scrolling)!

I've been playing with the Slack app on iOS, whose swipe is really good, but after looking, it seems like it's a native app (I thought it was a web app like their desktop client, my bad).
I've noticed that their cut-off between "user is trying to scroll" and "user is trying to open the menu" seems to be a 45% angle. When the gesture is slightly more horizontal than vertical, the menu opens (the whole screen triggers) and vertical scrolling is disabled; when the gesture is slightly more horizontal than vertical, the menu opens and scrolling is inactive.

That's a nice and very natural pattern IMO. If/when we can get to that, that'll be great, but I think it still supports opening the menu from the full width of the screen.

McInkay:
So, having tested this, I'm finding it a bit off putting.

When I'm scrolling down, I find the menu starting to pop out and then going back in, so every time I scroll, the whole screen moves ~5px to the right and back again. It's quite off putting. Maybe it's just how my finger moves? I dunno, but I don't notice that on any other app.

xPaw:
@McInkay that's exactly what @maxpoulin64 mentions which makes me hesitant on this change.

McInkay:
Yeah, I just read that comment. I don't have anything against this in general, but we need to change the angle we watch, as it's really off putting if that happens.

williamboman:
I agree with @McInkay. When scrolling quickly the chat content jitters horizontally. Doesn't happen all the time but often enough that I find it quite annoying.

astorije:
So, this is that lock thing. When scrolling the chat, the menu is locked closed, but when opening the menu, the chat scroll is not locked.
One should be locked when the other triggers, and the angle for deciding which direction is attempted should be 45% to limit errors as much as possible (when I played with Slack, I noticed that 45% angle too).

That said, do we want to block this PR on fixing the slideout menu / lock thingy? I'd say go ahead with this PR and fix the other thing in another PR, but I'll go with the flow :)

christer88:
> Well, my problem is that it actively makes it awkward for me to use on mobile, you know? It's not like a new feature that just has a missing bit that we can add on later, it's something that makes the experience worse for some people.

For me it's the opposite. It's very hard to slideout channel list on iOS because it thinks I'm going back one webpage instead. :/ This PR fixes this problem.

xPaw:
@christer88 You are missing the problem he's describing. See https://github.com/thelounge/thelounge/pull/2064#issuecomment-367582288

christer88:
I understand the problem when swiping you also some times scroll downwards/upwards in the chat at the same time and vice versa, due to either of them doesn't lock the other one. It's just, the current situation makes it so difficult to slideout channel list at all on iOS - so I think the issue described by McInkay and maxpoulin64 does not outweight the other one. This is puerly my opinion though.

McInkay:
Would it be a compromise at this point to just increase the hitbox for the sidebar to a point where it becomes easier for iOS users, but not the whole screen?

astorije:
It's an acceptable compromise, but it's not going to fix anything on the long run, and the issue you mentioned above will still happen when scrolling too close to the left side of the screen.

I understand and agree with both issues, but it seems to me that one is annoying (the menu jittering) while the other one effectively affects usability. While I think both should be fixed, I'm on the side that the latter is more pressing.

@xyzz or @christer88, would you be able to / want to look into fixing the other issue that has been mentioned in this thread? That would be a great improvement to The Lounge!
In terms of solution for this specific issue, we could either go with this PR assuming a fix for the other issue gets worked on, or hold on this PR until said fix happens, or go with @McInkay's compromise until said fix happens while keeping in mind that it's not a "best of both world" type of solution. I'm fine with anything, though obviously would prefer both things to be entirely fixed :) Help welcome!

McInkay:
Yeah, obviously fixing both issues would be better but my compromise was just a temporary thing, making the jittery thing less often an issue, which still making it easier for iOS users to slideout. Definitely not meant as a 100% fix, just temporary.

McInkay:
I have been watching what I've been doing when scrolling and ... I think 1/4 of the page would be a good one to go with. I don't tend to stray into the left 25% when scrolling with either hand.

astorije:
Sounds good. @xyzz, would you mind updating your PR so it takes 1/4 of the width? 🙏

McInkay:
@xyzz Will you have time to do this, or do you want someone else to pick up? Happy either way, but would be good to get this in with the requested change. Also, conflicting now sadly.

xyzz:
I don't really want to settle on a mediocre solution and don't have time to implement a proper one so somebody else will have to work on that, sorry.

McInkay:
> I don't really want to settle on a mediocre solution and don't have time to implement a proper one so somebody else will have to work on that, sorry.

I'm sorry you feel this is mediocre, but it is at least better than what we currently have in master, and hopefully someone else will have time to implement the fix so we can update this to be the whole width one day. 

Thank you for your contribution. 

Whoever picks this up, assign yourself so we don't have duplicated effort. The only things that need done are rebasing/resolving conflicts, and changing the width to 1/4.

xPaw:
Checked on Android real quick, it seems that *any* vertical movement disables sidebar opening.

In our code I think this can be achieved by changing this line:

https://github.com/thelounge/thelounge/blob/e987917c64cbc2e4e40d23b5ade02de6b730b422/client/js/slideout.js#L56

to read
```js
if (!menuIsMoving && touch.screenY !== touchStartPos.screenY) {
```

@xyzz Are you sure you don't want to finish this up?

xPaw:
Okay I implemented this in #2246 (https://github.com/thelounge/thelounge/pull/2246/commits/2d4f6e9faf89dd24e3a53eb351c15aa49f63c416)

Can you guys test that PR? It should work rather consistently and nicely now.

PR #2066, opened 2018-02-15
Display password field when displayNetwork is false

### LINKED ISSUE #509: Passwords as a per-user thing, not a per-server thing.
Original issue: https://github.com/erming/shout/issues/677

Nowadays a lot of IRC server and client software recognize the password field as something that a user uses to authenticate himself. For example, freenode forwards the PASS option to NickServ, and Kiwi allows the user to change the password field even when server connection settings are locked.

Shout/The Lounge currently handles the password field as a "network setting" as if the password is a mandatory field needed to access the network. In my opinion, and likely in the opinion of many others, the password field should be specified as a "user preference", not as a "network setting."

xPaw:
Fixes #509. The lock on password is not enforced, and it's limiting not to display it.

If `displayNetwork` is `false`, password field is added under username. If `true` the password field stays under network (current behaviour).

### DIFF
--- a/client/views/windows/connect.tpl
+++ b/client/views/windows/connect.tpl
@@ -37,12 +37,6 @@
 				</div>
 			</div>
 			<div class="clearfix"></div>
-			<div class="col-sm-3">
-				<label for="connect:password">Password</label>
-			</div>
-			<div class="col-sm-9">
-				<input class="input" id="connect:password" type="password" name="password" value="{{defaults.password}}">
-			</div>
 			<div class="col-sm-9 col-sm-offset-3">
 				<label class="tls">
 					<input type="checkbox" name="tls" {{#if defaults.tls}}checked{{/if}} {{#if lockNetwork}}disabled{{/if}}>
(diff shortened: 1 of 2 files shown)


astorije:
Could we put it at the same location regardless of value of `displayNetwork`? That is, always between "Username" and "Real name".
I would much prefer that compared to having twice the same field in 2 different locations based on `displayNetwork`.

xPaw:
@astorije I thought so too, but I prefer having password under network, otherwise the sections become rather unbalanced.

astorije:
I think that's perfectly fine:

<p align=center>
screen shot 2018-02-16 at 18 56 52
</p>

I agree with #509 that password is more of a user-related thing than network-related, but more importantly I think it's better for consistency (and also we don't have to keep twice the same input, which at some point I'm sure we'll forgot about and we'll have some bug 😅).

xPaw:
@astorije OK, moved.

client/views/windows/connect.tpl
```diff
@@ -69,6 +63,12 @@
 			<input class="input username" id="connect:username" name="username" value="{{defaults.username}}">
 		</div>
 		{{/unless}}
+		<div class="col-sm-3">
+			<label for="connect:password">Password</label>
+		</div>
+		<div class="col-sm-9">
+			<input class="input" id="connect:password" type="password" name="password" value="{{defaults.password}}">
```
astorije: I believe this default needs to be added to server.js's getClientConfiguration() no?

(I don't know if having a default password makes sense but that's a different story)

xPaw: You are right, added.

nerthazrim:
Thanks a lot for that!

Waiting for your next release and I'll test that ;)

Thanks for your work!

PR #2073, opened 2018-02-18
Enforce object literal shorthand syntax with ESLint

astorije:
All changes in this commit were made with `npm run lint:js -- --fix`.

### DIFF
--- a/client/js/sorting.js
+++ b/client/js/sorting.js
@@ -17,18 +17,18 @@ module.exports = function() {
 		forcePlaceholderSize: true,
 		tolerance: "pointer", // Use the pointer to figure out where the network is in the list
 
-		update: function() {
+		update() {
 			const order = [];
+
 			sidebar.find(".network").each(function() {
 				const id = $(this).data("id");
 				order.push(id);
 			});
-			socket.emit(
-				"sort", {
-					type: "networks",
-					order: order,
-				}
-			);
+
+			socket.emit("sort", {
+				type: "networks",
+				order: order,
+			});
 
 			options.settings.ignoreSortSync = true;
 		},
@@ -43,20 +43,20 @@ module.exports = function() {
 		forcePlaceholderSize: true,
 		tolerance: "pointer", // Use the pointer to figure out where the channel is in the list
 
-		update: function(e, ui) {
+		update(e, ui) {
 			const order = [];
 			const network = ui.item.parent();
+
 			network.find(".chan").each(function() {
 				const id = $(this).data("id");
 				order.push(id);
 			});
-			socket.emit(
-				"sort", {
-					type: "channels",
-					target: network.data("id"),
-					order: order,
-				}
-			);
+
+			socket.emit("sort", {
+				type: "channels",
+				target: network.data("id"),
+				order: order,
+			});
 
 			options.settings.ignoreSortSync = true;
 		},
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -47,6 +47,10 @@ rules:
   no-use-before-define: [error, {functions: false}]
   no-var: error
   object-curly-spacing: [error, never]
+  object-shorthand:
+    - error
+    - methods
+    - avoidExplicitReturnArrows: true
   padded-blocks: [error, never]
   padding-line-between-statements:
     - error
--- a/client/js/autocompletion.js
+++ b/client/js/autocompletion.js
@@ -14,7 +14,7 @@ let enabled = false;
 
 module.exports = {
 	enable: enableAutocomplete,
-	disable: () => {
+	disable() {
 		if (enabled) {
 			input.off("input.tabcomplete");
 			Mousetrap(input.get(0)).off("tab", "keydown");
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -79,6 +79,7 @@ const condensedTypes = [
 	"kick",
 	"mode",
 ];
+const condensedTypesQuery = "." + condensedTypes.join(", .");
 
 const timeFormats = {
 	msgDefault: "HH:mm",
@@ -86,9 +87,9 @@ const timeFormats = {
 };
 
 module.exports = {
-	colorCodeMap: colorCodeMap,
-	commands: commands,
-	condensedTypes: condensedTypes,
-	condensedTypesQuery: "." + condensedTypes.join(", ."),
-	timeFormats: timeFormats,
+	colorCodeMap,
+	commands,
+	condensedTypes,
+	condensedTypesQuery,
+	timeFormats,
 };
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -30,9 +30,7 @@ Mousetrap.bind([
 		scrollTop = Math.ceil(scrollTop + offset);
 	}
 
-	container.animate({
-		scrollTop: scrollTop,
-	}, 200);
+	container.animate({scrollTop}, 200);
 
 	return false;
 });
--- a/client/js/libs/handlebars/ircmessageparser/merge.js
+++ b/client/js/libs/handlebars/ircmessageparser/merge.js
@@ -8,12 +8,9 @@ function assign(textPart, fragment) {
 	const fragStart = fragment.start;
 	const start = Math.max(fragment.start, textPart.start);
 	const end = Math.min(fragment.end, textPart.end);
+	const text = fragment.text.slice(start - fragStart, end - fragStart);
 
-	return Object.assign({}, fragment, {
-		start: start,
-		end: end,
-		text: fragment.text.slice(start - fragStart, end - fragStart),
-	});
+	return Object.assign({}, fragment, {start, end, text});
 }
 
 // Merge the style fragments withing the text parts, taking into account
--- a/client/js/localStorage.js
+++ b/client/js/localStorage.js
@@ -1,7 +1,7 @@
 "use strict";
 
 module.exports = {
-	set: function(key, value) {
+	set(key, value) {
 		try {
 			window.localStorage.setItem(key, value);
 		} catch (e) {
@@ -10,10 +10,10 @@ module.exports = {
 			// available. See http://stackoverflow.com/q/14555347/1935861.
 		}
 	},
-	get: function(key) {
+	get(key) {
 		return window.localStorage.getItem(key);
 	},
-	remove: function(key, value) {
+	remove(key, value) {
 		window.localStorage.removeItem(key, value);
 	},
 };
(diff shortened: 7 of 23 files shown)


creesch:
Wouldn't it be better to make this and issue and do a pull request for this after most of the current pull requests have been merged for 3.0 ? Seems like it will only cause a bunch of merge conflicts either on this end or current pull requests that are otherwise ready to go. 

Also, personally I find this less readable as far as code goes as you basically have things not as neatly aligned anymore in object declarations but more of a zig zag.

McInkay:
I don't think this does make it any less readable, as I don't see how they are less aligned? They are just half the width now.

But yeah, it's going to cause lots of merge conflicts. The problem is, there is never a point where it'll not cause them :-\

creesch:
Well without the shorthand all your objects look like this
```javascript 

const object = {
    var1: value1,
    var2: value2,
    var3: var3,
    var4: value4,
}

```

Which now turns into 

```javascript 

const object = {
    var1: value1,
    var2: value2,
    var3,
    var4: value4,
}
```

So it is no longer a given you can see the value on the right. It is a minor thing. 


As far as preventing merge conflicts, I get that. 

Dunno, if I had known this was coming I'd already followed this style in the PR for settings sync. So you could basically do something like this

- Do `eslint --fix` in 3.0 as the last PR. 
- In the review process for any PR that will be for a version after 3.0 ask people to make any code change with that in mind. 

Personally, having to do a lot of merge conflict solving on a big enough PR always leaves me a bit uneasy about the state of code certainly when the PR has been sitting for a while making me go back and forth a bunch of times making sure I actually merge the correct bits of code. 

You are right though, you have to pick some point in time if you want to make changes like this so might as well do it now. You could pick a minor release but then you probably would still have the same issues with PRs ready for the bigger release after the minor one.

McInkay:
> So it is no longer a given you can see the value on the right. It is a minor thing.

Oh yeah, I see what you mean.

I think this is one of thsoe things that maybe we don't want to enforce everywhere, as there are some examples like this that would not benefit from it, but the majority of cases, we just name them all what the variables were called, and so 

```
var test = {
	var1,
	var2
}
```

definitely looks better than

```
var test = {
	var1: var1,
	var2: var2
}
```

So maybe this is the same sort of thing, unnecessary to enforce, but we can just be improving slowly as we go along? It would also solve the merge conflicts issue.

creesch:
Doesn't eslint allow for more fine-grained control here (it sometimes does), as a start it could be moved to warn.

creesch:
Apparently it does, there is '"consistent-as-needed"' as option.

https://eslint.org/docs/rules/object-shorthand

astorije:
If we wait until next release to merge this kind of PRs, we'll have the same issue. After all, we just did a release and it's already a problem :)

Anyway, I'm totally fine waiting to merge this PR until we merge a bunch of others as it's dead easy to fix conflicts here (it's auto-fixable), but I'm not going to stage ESLint settings, because that would still cause conflicts, only several times (also I don't think it's time well spent).

Note if anyone runs into conflicts after this PR gets merged and want a super easy recipe to fix: ignore all the upstream changes (i.e. keep your version of the code), and run `npm run lint:js -- --fix`. Free conflict resolution :)

McInkay:
What do you think about using consistent-as-needed for this, @astorije ? Would be better, imo.

astorije:
I don't see how it's better actually, I'd rather keep `always`.
Sometimes you are introducing one key/value that requires expansion, and `consistent-as-needed` would force rewriting all the keys, so not a fan.

McInkay:
Well, the point is that these eslint rules we use to make the code look better and more readable. This is why this rule is being enforced, correct? Removing unnecessary characters and making it easier to read some object creation.

The argument for consistent-as-needed is that if an object has 1 or more that needs to be expanded, it is more readable for all of them to be expanded, rather than having to run your eyes in multiple directions more than needed. In this case, the expanded characters are no unnecessary characters, but are helping the readability.

astorije:
> The argument for consistent-as-needed is that if an object has 1 or more that needs to be expanded, it is more readable for all of them to be expanded, rather than having to run your eyes in multiple directions more than needed. In this case, the expanded characters are no unnecessary characters, but are helping the readability.

This is what I disagree with. I find `myFunction({ foo, bar: bad, baz })` more readable than `myFunction({ foo: foo, bar: bad, baz: baz })`.

xPaw:
@astorije I find it more confusing to mix it. I say leave the rule to allow it, and we can clean up code to remove the mixing.

creesch:
To be honest I wouldn't write it like that. 

```javascript
myFunction({ 
    foo, 
    bar: bad, 
    baz 
})
```
And 

```javascript
myFunction({ 
    foo: foo, 
    bar: bad, 
    baz: baz 
})
```

Are usually how these functions are written out for readability and there the second one is much clearer. 
Also in the first example it takes a bit more effort to parse and when not paying attention `bad` could be mistaken for fifth parameter rather than the value for `bar`.

astorije:
Obviously it was on a single line for the sake of commenting here...

> there the second one is much clearer.

Again, I disagree with this. It is very subjective, it's not like one of us holds the truth :)

> Also in the first example it takes a bit more effort to parse and when not paying attention `bad` could be mistaken for fifth parameter rather than the value for `bar`.

That is a bit reaching, IMO. The advantage of this notation is calling out differences, so that `bad` and `bar` cannot be mistaken.

Putting it in practical matters, I find [this](https://github.com/thelounge/lounge/blob/3e11d9a9c1548faa6020e3ea12ca52ee5ae03c37/client/js/constants.js#L108-L115) much more readable/parseable than [this](https://github.com/thelounge/lounge/blob/3e11d9a9c1548faa6020e3ea12ca52ee5ae03c37~/client/js/constants.js#L108-L115), cognitive overflow.

**But here is the kicker:** `consistent-as-needed` is not auto-fixable, so it's an all-or-nothing type of deal :)

creesch:
> Obviously it was on a single line for the sake of commenting here...

Not obvious to me. 

> Again, I disagree with this. It is very subjective, it's not like one of us holds the truth :)

Well yeah, a lot of this is but so far we have provided reasoning and an explanation for how we see it not just "I disagree". 

> That is a bit reaching, IMO. 

In a single line instance? I don't think so, one of the most well known jokes about debuging code is that silly comma or other single character you keep overlooking. When there is a clear pattern of `attribute: value` it is easier to look at glance at something than when that pattern is absent and it sometimes might just be the attribute and other times it might the attribute and value. 

 > consistent-as-needed is not auto-fixable, so it's an all-or-nothing type of deal 

Fair enough, guess that makes it an easy call.

astorije:
> so far we have provided reasoning and an explanation for how we see it not just "I disagree".

Yes, and I did that too 😉

astorije:
> I'd rather we clean up the code so that there is no mixing, but that can be done in the future.

I essentially did that on a bunch of places. Thoughts?

McInkay:
> I essentially did that on a bunch of places. Thoughts?

What do you mean by that? This PR just forces everything to be shorthand if it can be, but it doesn't turn anything from different names into the same name so it can be shorthanded.

For example, before this PR we had stuff like:

```
return {
  chan: chan.id,
  messages: messages,
};
```

Your PR turns it into 

```
return {
  chan: chan.id,
  messages,
};
```

What @xPaw Is saying is that it would be better to have:

```
return {
  chan,
  messages,
};
```

overall (bad example, I know, because we already have a chan in this code, but ... let it be for now).

Now, obviously, we can't just change everything in 1 go, because there are a lot of things (like the example I gave above) which can't just be renamed, due to existing variables. But we don't like the mixed nature of the middle example, and so would prefer if, for this PR, we changed it to `consistent-as-needed` so that we get the 3rd option when we can, but when we can't just now, we get the top, as many of us find it more readable. Eventually, we will change everything to be the 3rd, but only as we edit the code around it.

astorije:
> it doesn't turn anything from different names into the same name so it can be shorthanded.

Yes, I've actually updated this PR to do exactly this, [for example here](https://github.com/thelounge/thelounge/pull/2073/files#diff-386df29cd51cc4ce978bd6fee66927ca), but not everywhere, granted, indeed because:

> we can't just change everything in 1 go, because there are a lot of things (like the example I gave above) which can't just be renamed, due to existing variables.

But where it was very sensible to do so, I did :)

> for this PR, we changed it to `consistent-as-needed`

As mentioned above, my main concern is that `consistent-as-needed` (and `consistent`) is not auto-fixable with the `--fix` flag, which means that this will always have to be fixed by hand, including when other PRs will have to rebase on this.
If it was auto-fixable, I would have made the switch already, but I think that, when in doubt, we should go for the auto-fixable solution as it benefits the many IMO. What do you think about this?

astorije:
@McInkay, did you get any chance to look at my last comment? :)

McInkay:
Yeah, my answer to that would be "meh".

I get the issue, and see why you wouldn't want to do that.

But that would lead me to say I'm not a huge fan of this PR at all in that case. You know that I'm not a fan with enforcing too much with eslint. I like the stuff that can cause real programming errors (`==` vs `===`) and to keep certain things consistent (indenting mainly). But some stuff like this ... just feels unnecessary to force, especially when it's less than ideal.

But whatever, I don't really care enough. I've approved it already, just go for it if you both want it.

McInkay:
Fine with methods.

astorije:
Alright, I think we're getting to a compromise here! *But* I did quite a bit of cleanup for this and I would not feel too great to blindly revert all that.
So what about:

- We enforce `methods`
- We don't enforce `properties`
- I keep the cleanup on things such as:
  ```diff
   function settingSetEmit(name, value) {
  -	socket.emit("setting:set", {
  -		name: name,
  -		value: value,
  -	});
  +	socket.emit("setting:set", {name, value});
   }
  
  -const template = $(templates.msg_preview({preview: preview}));
  +const template = $(templates.msg_preview({preview}));
  ```
  These are no-brainers to me (but again, no enforcing) :)
  - I revert the questionable "cleanups", i.e. those where there is an inconsistency between expanded/shorthand, those requiring subpar gaming-the-system kind of thing (there is a `chan = chan.id` somewhere to please the linter...), etc. so that what you guys find less than ideal doesn't make it.

Deal? :)

McInkay:
Yeah sure, that seems like a nice balance.

astorije:
@McInkay, good to go I think. Makes for a shorter PR, less likely to conflict.

client/js/sorting.js
```diff
@@ -17,18 +17,15 @@ module.exports = function() {
 		forcePlaceholderSize: true,
 		tolerance: "pointer", // Use the pointer to figure out where the network is in the list
 
-		update: function() {
+		update() {
+			const type = "networks";
```
xPaw: This PR looks good now, except for this (and line 44). It feels somewhat unnatural to do it like this.

astorije: Ok, reverted both.

PR #2074, opened 2018-02-18
Enfore array/object desctructuring with ESLint

astorije:
These changes were **not** done by autofix.

Do we want this? I think it makes for nicer code on the long-term, but initially there are some weirdnesses.


McInkay:
I'm not sure I want want to enforce destructuring. It can be nicer in some cases, but can be confusing in others.

creesch:
Wouldn't it be better to make this and issue and do a pull request for this after most of the current pull requests have been merged for 3.0 ? Seems like it will only cause a bunch of merge conflicts either on this end or current pull requests that are otherwise ready to go.

creesch:
I have the same concern as with #2073, wouldn't it be better to make this and issue and do a pull request for this after most of the current pull requests have been merged for 3.0 ? Seems like it will only cause a bunch of merge conflicts either on this end or current pull requests that are otherwise ready to go.

astorije:
Yeah, I'm not going to merge this because enforcing this doesn't make sense, as I could see while fixing stuff. But I want to go through the options today and see if something can be taken from it.

When enforced on a project, I've noticed the shape of the code evolves slightly, for the better, so it's less something to apply after the fact (it can't be auto-fixed, even) and more something to have in mind when starting a feature.

astorije:
Meh.

PR #2075, opened 2018-02-19
Allow setting `rejectUnauthorized` per network

### LINKED ISSUE #313: Verify certificates when using TLS connections

xPaw:
This allow the connection to be rejected when the TLS certificated provided by the server is signed by a trusted CA.

Fixes #313


spangborn:
👍🏻Works well! Could possibly use a more informative message when refusing to connect to a non-trusted CA, but this is a good bit of functionality for the more security-aware users.

client/views/windows/connect.tpl
```diff
@@ -43,6 +43,12 @@
 					Enable TLS/SSL
 				</label>
 			</div>
+			<div class="col-sm-9 col-sm-offset-3">
+				<label class="tls">
+					<input type="checkbox" name="rejectUnauthorized" {{#if defaults.rejectUnauthorized}}checked{{/if}} {{#if lockNetwork}}disabled{{/if}}>
+					Reject unauthorized TLS certificates
```
astorije: I feel like there is a more idiomatic way to say this but I can't think of it. "Allow untrusted certificates"?
By the way I would go for "Allow" rather than "Reject", it seems to me that it's more common to see this wording, but also because an unchecked checkbox means "Do not reject" double-negation type 😅
No big deal if you feel strongly against though.

(I realize that `irc-framework` calls it `rejectUnauthorized`, but I don't think that's the best wording from an end-user standpoint)

defaults/config.js
```diff
@@ -233,6 +233,17 @@ module.exports = {
...
+		//
+		// Enable certificate verification
+		//
+		// If true, the server certificate is verified against
+		// the list of supplied CAs by your node.js installation.
+		//
+		// @type     boolean
+		// @default  true
+		//
+		rejectUnauthorized: true,
```
astorije: It's kind of a headache going through the different combinations of our settings (I've rewritten this comment 3 times already), but ideally the checkbox would change to enabled whenever user is changing the settings to a different network.
That's an amount of extra work probably outside the scope of this PR though.

xPaw: I'm not sure what you mean by

> the checkbox would change to enabled whenever user is changing the settings to a different network.

client/views/windows/connect.tpl
xPaw: `rejectUnauthorized` comes from Node, which is why I kept it. https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener

defaults/config.js
astorije: Yeah, I was wondering if that was clear...
What I meant was that as soon as the user starts typing something that is not the default network, we should probably set the checkbox to the highest security setting. If settings value is `false`, then by default the UI suggests the least secure option.

But as mentioned, this is likely more work than it's worth, and I can think of a couple edge cases making this painful.

xPaw: I think I see what you mean. This feels like a lot of effort for little gain there (what we really need is network editing!).

client/views/windows/connect.tpl
astorije: Yeah I get that from the API/internals point of view, but in the UI/UX, it makes more sense to not use a negated version. It's just confusing right now IMO.

xPaw: I'm fine with changing the label. `Verify TLS certificate trust chain`?

client/views/windows/connect.tpl
```diff
@@ -40,7 +40,13 @@
...
 				<label class="tls">
 					<input type="checkbox" name="tls" {{#if defaults.tls}}checked{{/if}} {{#if lockNetwork}}disabled{{/if}}>
-					Enable TLS/SSL
+					Enable TLS
+				</label>
+			</div>
+			<div class="col-sm-9 col-sm-offset-3">
+				<label class="tls">
+					<input type="checkbox" name="rejectUnauthorized" {{#if defaults.rejectUnauthorized}}checked{{/if}} {{#if lockNetwork}}disabled{{/if}}>
+					Verify TLS certificate trust chain
```
astorije: tbh this raises even more questions to me 😅 

Going through the Node doc, I would call it any of: `(Accept|Allow)
 (untrusted|unverified|unauthorized) [TLS] certificate[s]` (I'm not sure "TLS" brings a lot of info when making the decision of checking/unchecking the checkbox).

That would reverse the meaning of checking/unchecking though (yours: checked is secure, unchecked is insecure; mine: checked is insecure, unchecked is secure) and I understand you'd rather keep the current state, so a contrapositive could be: `Accept (trusted|verified|authorized) [TLS] certificate[s] only`.

What do you think?

xPaw: Reworded it to `Allow untrusted certificates`.

PR #2079, opened 2018-02-19
Add titles to all windows

xPaw:
Follow-up of https://github.com/thelounge/lounge/pull/1915

### DIFF
--- a/client/index.html.tpl
+++ b/client/index.html.tpl
@@ -84,11 +84,11 @@
 						</div>
 					</form>
 				</div>
-				<div id="sign-in" class="window" role="tabpanel"></div>
-				<div id="connect" class="window" role="tabpanel"></div>
-				<div id="settings" class="window" role="tabpanel"></div>
-				<div id="help" class="window" role="tabpanel"></div>
-				<div id="changelog" class="window"></div>
+				<div id="sign-in" class="window" role="tabpanel" aria-label="Sign-in"></div>
+				<div id="connect" class="window" role="tabpanel" aria-label="Connect"></div>
+				<div id="settings" class="window" role="tabpanel" aria-label="Settings"></div>
+				<div id="help" class="window" role="tabpanel" aria-label="Help"></div>
+				<div id="changelog" class="window" aria-label="Changelog"></div>
 			</div>
 		</div>
 	</div>


astorije:
@xPaw, I'd be more comfortable with the following changes instead:

```diff
───────────────────────────────
modified: client/index.html.tpl
───────────────────────────────
@ index.html.tpl:87 @
                        </div>
                    </form>
                </div>
-               <div id="sign-in" class="window" role="tabpanel"></div>
-               <div id="connect" class="window" role="tabpanel"></div>
-               <div id="settings" class="window" role="tabpanel"></div>
-               <div id="help" class="window" role="tabpanel"></div>
-               <div id="changelog" class="window"></div>
+               <div id="sign-in" class="window" role="tabpanel" aria-label="Sign-in"></div>
+               <div id="connect" class="window" role="tabpanel" aria-label="Connect"></div>
+               <div id="settings" class="window" role="tabpanel" aria-label="Settings"></div>
+               <div id="help" class="window" role="tabpanel" aria-label="Help"></div>
+               <div id="changelog" class="window" aria-label="Changelog"></div>
            </div>
        </div>
    </div>
─────────────────────────────
modified: client/js/lounge.js
─────────────────────────────
@ lounge.js:399 @ $(function() {
            .trigger("show");

        let title = $(document.body).data("app-name");
-       const chanTitle = chan.attr("aria-label") || "";
+       const chanTitle = chan.attr("aria-label");
        if (chanTitle.length > 0) {
            title = `${chanTitle} — ${title}`;
        }
```

What I don't like about such fallback is that a typo something along the chain would be breaking, but silently, as we would fallback to an empty string.

Additionally, with that diff above, Settings/Help/etc. now do not have a title, just like channel windows, which is pretty nice :)

What do you think?

xPaw:
I didn't know if we wanted to add titles. If we do, then yes your fix is appropriate.

astorije:
I'd love titles yeah. If anything, things like back dropdown on desktop is gaining context.

peteresculley:
I noticed a bug when I first ran The Lounge that because there is no channel title before logging in "chanTitle" is undefined and this if() has an error:
In /client/js/utils.js :
if (chanTitle.length > 0) {
    title = `${chanTitle} — ${title}`;
}

Sorry about formatting, I don't know how to use GitHub formatting yet.

xPaw:
@peteresculley Sounds like you tried an old version or didn't build correctly.

PR #2081, opened 2018-02-20
Consider localhost connections secure

xPaw:
Browsers do similar logic with serviceworkers and other things.

### DIFF
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -111,8 +111,11 @@ Network.prototype.getNetworkStatus = function() {
 		const transport = this.irc.connection.transport;
 
 		if (transport.socket) {
+			const isLocalhost = transport.socket.remoteAddress === "127.0.0.1";
+			const isAuthorized = transport.socket.encrypted && transport.socket.authorized;
+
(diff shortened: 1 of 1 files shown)


src/models/network.js
```diff
@@ -111,8 +111,11 @@ Network.prototype.getNetworkStatus = function() {
 		const transport = this.irc.connection.transport;
 
 		if (transport.socket) {
+			const isLocalhost = transport.socket.remoteAddress === "127.0.0.1";
```
dgw: Shouldn't this also be true for `::1` (IPv6)?

astorije: What about `localhost`, is that considered a valid `remoteAddr`?

xPaw: This is after DNS resolution

astorije: And what about IPv6?

xPaw: I haven't gotten it to resolve localhost into an IPv6 addr myself, so I don't know which form exactly it will use.

astorije: :shipit:

PR #2090, opened 2018-02-21
Allow generating random numbers in nick on page load

xPaw:
Very useful in public mode. No need to rely on `nick in use` to generate random.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -240,17 +240,17 @@ module.exports = {
 		// For example, Guest%%% will become Guest123 on page load.
 		//
 		// @type     string
-		// @default  "lounge-user"
+		// @default  "thelounge%%"
 		//
-		nick: "lounge-user",
+		nick: "thelounge%%",
 
 		//
 		// Username
 		//
 		// @type     string
-		// @default  "lounge-user"
+		// @default  "thelounge"
 		//
-		username: "lounge-user",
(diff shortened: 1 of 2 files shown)


src/server.js
```diff
@@ -521,7 +521,7 @@ function getClientConfiguration() {
 	config.themes = themes.getAll();
 
 	if (config.displayNetwork) {
-		config.defaults = Helper.config.defaults;
+		config.defaults = _.clone(Helper.config.defaults);
```
astorije: We should not call that defaults anymore IMO. Not necessarily for your PR though.

defaults/config.js
```diff
@@ -236,10 +236,13 @@ module.exports = {
...
 		//
+		// Percent sign (%) will be replaced into a random number from 0 to 9.
+		// For example, Guest%%% will become Guest123 on page load.
+		//
 		// @type     string
-		// @default  "lounge-user"
+		// @default  "thelounger%%"
 		//
-		nick: "lounge-user",
+		nick: "thelounger%%",
```
astorije: I'd prefer to have `thelounge-user%%` but whatever.

xPaw: We can do `TheLounge%%`

astorije: lol no thanks. `thelounge%%`?

PR #2094, opened 2018-02-22
Enable in-browser run of client testing

astorije:
This gets us one step closer to cross-browser (unit) testing 🎉

To test this:

- `yarn install`
- `yarn start:mocha`
- Go to http://localhost:8080/ in the browser you want to test, and you will see this:

screen shot 2018-02-22 at 01 43 08

What I'm hoping then is to be able to run these steps in cloud-based browser testing solutions ([BrowserStack](https://www.browserstack.com/)? [Sauce Labs](https://saucelabs.com/)? Else?) and, as we grow our test suite, make sure that what we do is compatible with the browsers we expect compatibility with.
I haven't done this cloud-based browser testing SaaS part in the past, so this is new ground for me 🙏

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "test:mocha": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
     "test:mocha-webpack": "nyc --nycrc-path=test/.nycrc-mocha-webpack mocha-webpack --colors --opts=test/mocha-webpack.opts",
     "lint:js": "eslint . --report-unused-disable-directives --color",
-    "lint:css": "stylelint --color \"client/**/*.css\""
+    "lint:css": "stylelint --color \"client/**/*.css\"",
+    "start:mocha": "webpack-dev-server --config=webpack.config-browser.js"
   },
   "keywords": [
     "lounge",
@@ -80,6 +81,7 @@
     "jquery": "3.3.1",
     "jquery-ui": "1.12.1",
     "mocha": "5.0.1",
+    "mocha-loader": "1.1.3",
     "mocha-webpack": "1.0.1",
     "mousetrap": "1.6.1",
     "npm-run-all": "4.1.2",
@@ -88,6 +90,7 @@
     "stylelint": "9.1.1",
     "stylelint-config-standard": "18.1.0",
     "textcomplete": "0.16.0",
-    "webpack": "3.11.0"
+    "webpack": "3.11.0",
+    "webpack-dev-server": "2.11.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -12,17 +12,17 @@
   },
   "homepage": "https://thelounge.chat/",
   "scripts": {
-    "coverage": "rm -rf .nyc_output/ && npm-run-all test:* && nyc --nycrc-path=test/.nycrc-report report",
-    "start": "node index start",
-    "start-dev": "npm-run-all --parallel watch start",
     "build": "webpack",
-    "watch": "webpack --watch",
-    "test": "npm-run-all --aggregate-output --parallel --continue-on-error test:* lint:*",
-    "test:mocha": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
-    "test:mocha-webpack": "nyc --nycrc-path=test/.nycrc-mocha-webpack mocha-webpack --colors --opts=test/mocha-webpack.opts",
-    "lint:js": "eslint . --report-unused-disable-directives --color",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -21,7 +21,8 @@
     "test:mocha": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
     "test:mocha-webpack": "nyc --nycrc-path=test/.nycrc-mocha-webpack mocha-webpack --colors --opts=test/mocha-webpack.opts",
     "lint:js": "eslint . --report-unused-disable-directives --color",
-    "lint:css": "stylelint --color \"client/**/*.css\""
+    "lint:css": "stylelint --color \"client/**/*.css\"",
+    "start:mocha": "webpack-dev-server --config=webpack.config-browser.js"
```
McInkay: This scripts section needs an overhaul.

1. Not in alphabetical order, which saddens me
2. We are using `:`s and `-` when we should be consistent.

Which do we want to use? I like `:` but we should change `start-dev` to `start:dev` in that case.

However, that's not for this PR. What is for this PR is ... I'm not convinced about this name.

It should probably be a `test:` prefix, not `start:`, I think. And I would call it something like `test:watch` or `test:browser` or something similar.

McInkay: Well, I mean I'm just not a huge fan of using the `start:` prefix for something that isn't running the application, you know?

> The issue with test:* is that it will get picked by the test script :/

The test script is literally just running 2 test scripts, so we can just change it to use those 2. It's not like we are likely to end up having more scripts that run here very often, and it's not like people are having to type out the long command each time, we have it coded in the package.json so it can be verbose if needs be.

I'd rather that, and call the new one `test:` than `start:`.

Fwiw, I also think the existing test scripts would probably be better as `test:client` and `test:server`, because the client tests (although they use webpack) also use mocha, and we don't care what they are using, the important thing is what we are testing.

If I want to run just 1 of the test scripts, I need to try and remember what the initial tool we use to run them is, rather than whether they are client or server tests, which confuses things.

So, I would suggesting `test:server`, `test:client`, `test:watch`, and change `test` to hard code `test:server` and `test:client` instead of `test:*`. I think that would be the cleanest and most understandable solution.




As a side note, xpaw was complaining about changing `start-dev` to `start:dev` as `:` is harder to type. This led me to think ... why do we need the `start:` prefix for dev at all? Why not just call it `dev` and we just leave `start` for the production run.

astorije: > Well, I mean I'm just not a huge fan of using the `start:` prefix for something that isn't running the application, you know?

Yep, gotcha!

> The test script is literally just running 2 test scripts, so we can just change it to use those 2.

Done! (Also changed the `coverage` script)

> I'd rather that, and call the new one `test:` than `start:`.

Done!

> Fwiw, I also think the existing test scripts would probably be better as `test:client` and `test:server`

Done!

> So, I would suggesting `test:server`, `test:client`, `test:watch`, and change `test` to hard code `test:server` and `test:client` instead of `test:*`. I think that would be the cleanest and most understandable solution.

Done! (Except I went with `test:browser` instead of `test:watch` because `test:watch`/`watch:test` is usually used as "re-run `test` at every file changed", see https://mochajs.org/#-w---watch, whaddaya think?)

> Why not just call it `dev` and we just leave `start` for the production run.

Done!

You see, I can be a reasonable person sometimes! 😁
Happy to change any further if there are other things you'd like changed.

package.json
```diff
@@ -12,16 +12,17 @@
...
-    "coverage": "rm -rf .nyc_output/ && npm-run-all test:* && nyc --nycrc-path=test/.nycrc-report report",
-    "start": "node index start",
-    "start-dev": "npm-run-all --parallel watch start",
     "build": "webpack",
-    "watch": "webpack --watch",
-    "test": "npm-run-all --aggregate-output --parallel --continue-on-error test:* lint:*",
-    "test:mocha": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
-    "test:mocha-webpack": "nyc --nycrc-path=test/.nycrc-mocha-webpack mocha-webpack --colors --opts=test/mocha-webpack.opts",
+    "coverage": "rm -rf .nyc_output/ && run-s test:{client,server} && nyc --nycrc-path=test/.nycrc-report report",
+    "dev": "run-p watch start",
```
xPaw: What's `run-p`?

astorije: Yep https://github.com/mysticatea/npm-run-all/blob/master/docs/run-p.md

astorije:
@McInkay I did 😬

PR #2098, opened 2018-02-23
Enable logging for LDAP users

keegan:


src/plugins/auth/ldap.js
```diff
@@ -114,7 +114,7 @@ function ldapAuth(manager, client, user, password, callback) {
 	// auth plugin API
 	function callbackWrapper(valid) {
 		if (valid && !client) {
-			manager.addUser(user, null);
+			manager.addUser(user, null, true);
```
astorije: Could you remove `|| false` [here](https://github.com/thelounge/thelounge/blob/a24ab74c4cca262ba9dc516e1acf9493957e033a/src/clientManager.js#L115) please?
The only place that was not sending it as an argument was this one, so we can have no fallback now.

astorije:
I suggested that change so obviously I'm 👍 though biased 😅

The rationale behind this is as follows:

- As opposed to local-auth users which are created from the command line (which prompts yes/no for disk logs), there is no such thing with LDAP, a new user gets created on first login, and is given a default value for `logs` (currently `false`, `true` after this PR)
- Before this PR, there is no way to auto-set to `true`, all new users have logs disabled. After this PR, it will be the opposite scenario.
- #1839 gets merged, it will be possible to disable logs globally with `messageStore: none`. Without this PR, it will still be impossible to enable logs by default even after #1839.
- In https://github.com/thelounge/thelounge/pull/903, we made the prompt for local-auth default to yes, so this is consistent.

It's pretty late but I hope I clarified the intent.
Once users can edit their settings from the UI, this will be much easier to deal with on a per-user case.

src/plugins/auth/ldap.js
keegan: I have now added this

keegan:
Sorry about that, I have now quashed the commits

astorije:
No need to be sorry, it's totally fine and sometimes makes it easier to review new changes after a first review ;)
I approved, hopefully we can get that merged and in a pre-release soon!

PR #2101, opened 2018-02-23
Word prefetch errors differently

xPaw:
This is printed for any url, not just images, so the wording doesn't make much sense.


astorije:
Follow-up of #2036.

PR #2102, opened 2018-02-23
Back to help wanted label

xPaw:
GitHub does some magic with this label like saying "38 issues need help" on organization page.


McInkay:
Let's not merge without a review from @astorije as well though

PR #2108, opened 2018-02-23
Log socket close error if any

xPaw:

### DIFF
--- a/src/plugins/irc-events/connection.js
+++ b/src/plugins/irc-events/connection.js
@@ -78,7 +78,7 @@ module.exports = function(irc, network) {
 		identSocketId = client.manager.identHandler.addSocket(socket, client.name || network.username);
 	});
 
-	irc.on("socket close", function() {
+	irc.on("socket close", function(error) {
 		if (identSocketId > 0) {
 			client.manager.identHandler.removeSocket(identSocketId);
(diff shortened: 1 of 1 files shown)


astorije:
Errors are not really human-friendly here. Like I almost don't even know what they mean.
What about a nice message like "The connection with the server was lost", and keep the detailed error hidden like we did in that other PR?

xPaw:
@astorije Hiding the error makes that entire message unnecessary (there's a separate disconnected event).

While these errors can be cryptic, they can be searched for further information rather easily so it gives some context.

Chrome for example shows the error code too sometimes:

astorije:
Well yeah, but like your screenshot shows, human-friendly messages are in bigger, bolder fonts, and the specific error code is muted below.

I just think we can do better error reporting than this is doing, and in general across our project, but I'm not going to push hard for my proposition above.

At the very least if you want to keep the detailed message on screen (which is now inconsistent with prefetch error message, but again our error reporting across the project is not really great anyway), it would be great if we could go from "Socket closed: [...]" into something more digestible like "The connection with the server was lost: [...]". Open to better wording of course :)

xPaw:
@astorije I changed it to `Connection closed unexpectedly`

PR #2113, opened 2018-02-23
Refactor channel titles to use flexbox

xPaw:
- Get rid of ugly negative margins
- No floats
- Topic gets proper ellipsis
- Emojis are better aligned in topics

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -452,16 +452,23 @@ kbd {
 #viewport .rt,
 #chat button.menu {
 	color: #ccc;
-	display: none;
-	float: left;
+	display: flex;
 	font-size: 14px;
 	line-height: 1;
 	height: 36px;
-	margin: 6px 12px 0 -12px;
+	width: 36px;
+	margin-top: 6px;
+	flex-shrink: 0;
+}
+
+#viewport .lt::before,
+#viewport .rt::before,
+#chat button.menu::before {
 	width: 36px;
 }
 
 #viewport .lt {
+	display: none;
 	position: relative;
 }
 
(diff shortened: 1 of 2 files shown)


astorije:
Man you're awesome. This is really cool.
The only thing I can find is that emoji are now not as neatly (vertically centered) aligned:

Before | After
--- | ---
screen shot 2018-02-24 at 00 03 11 | screen shot 2018-02-24 at 00 03 24

I know you mentioned it's better after, but I do prefer before ahah. That said, I noticed the same (mis)alignment in messages and I think we should vertically center them with the text in both topic and messages.

client/css/style.css
```diff
@@ -1948,7 +1949,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 }
 
 .topic .emoji {
-	vertical-align: middle;
+	vertical-align: text-bottom;
```
astorije: I would prefer if we leave this at `middle` and actually change `.topic .emoji` into `.emoji`.
(I just noticed that emoji in messages are aligned with `bottom` except in the "The topic is:" action messages where this rule takes over `.emoji`).

xPaw: So this either depends on font or something else, because on my end  `middle` looks too low and `text-bottom` makes it more centered.

astorije: That's weird!

astorije: To be fair I expect middle to be vertically centered, it feels odd that text-bottom centers.

astorije:
Another benefit of this PR is that it's now pretty inexpensive to expand the header and add a scroll on the topic, so if anyone wants to work on UI to see the full topic, it should be much easier than before.

PR #2114, opened 2018-02-25
Allow connecting/disconnecting from networks in UI, closes #631

### LINKED ISSUE #631: When disconnected, network menu still shows "disconnect" option
When a network is disconnected, lounge tells you to use `/connect` to reconnect, so the menu should provide a connect option too instead of disconnect.

MaxLeiter:
Few thoughts that need to be worked out:

1. Not sure how I feel about icons
<table>
  <tr>
    <td>
screen shot 2018-03-17 at 2 06 24 pm
</td>
    <td>
screen shot 2018-03-17 at 2 06 29 pm

</td>
  </tr>
</table>

2. Should status be saved server-side? Add a `connected` field to `Network`?
3. More I'm sure I forgot
4. Need to deal with add-channel UI when disconnected. Thoughts?

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -300,49 +300,8 @@ $(function() {
 	$("#help").on("click", "#view-changelog, #back-to-help", openWindow);
 	$("#changelog").on("click", "#back-to-help", openWindow);
 
-	function closeChan(chan) {
-		let cmd = "/close";
-
-		if (chan.hasClass("lobby")) {
-			cmd = "/quit";
-			const server = chan.find(".name").html();
-
-			if (!confirm("Disconnect from " + server + "?")) { // eslint-disable-line no-alert
-				return false;
-			}
-		}
-
-		socket.emit("input", {
-			target: chan.data("id"),
-			text: cmd,
-		});
-		chan.css({
-			transition: "none",
-			opacity: 0.4,
-		});
-		return false;
-	}
-
 	sidebar.on("click", ".close", function() {
-		closeChan($(this).closest(".chan"));
-	});
-
-	const getCloseDisplay = (target) => {
-		if (target.hasClass("lobby")) {
-			return "Disconnect";
-		} else if (target.hasClass("channel")) {
-			return "Leave";
-		}
-
-		return "Close";
-	};
-
-	contextMenuFactory.addContextMenuItem({
-		check: (target) => target.hasClass("chan"),
-		className: "close",
-		displayName: getCloseDisplay,
-		data: (target) => target.attr("data-target"),
-		callback: (itemData) => closeChan($(`.networks .chan[data-target="${itemData}"]`)),
+		utils.closeChan($(this).closest(".chan"));
 	});
 
 	$(document).on("visibilitychange focus click", () => {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -265,6 +265,8 @@ kbd {
 .context-menu-user::before { content: "\f007"; /* http://fontawesome.io/icon/user/ */ }
 .context-menu-close::before { content: "\f00d"; /* http://fontawesome.io/icon/times/ */ }
 .context-menu-list::before { content: "\f03a"; /* http://fontawesome.io/icon/list/ */ }
+.context-menu-disconnect::before { content: "\f0c1"; /* https://fontawesome.com/icons/unlink?style=solid */ }
+.context-menu-connect::before { content: "\f127"; /* https://fontawesome.com/icons/link?style=solid */ }
 .context-menu-action-whois::before { content: "\f05a"; /* http://fontawesome.io/icon/info-circle/  */ }
 .context-menu-action-kick::before { content: "\f05e"; /* http://fontawesome.io/icon/ban/ */ }
 .context-menu-action-op::before { content: "\f1fa"; /* http://fontawesome.io/icon/at/ */ }
--- a/client/js/contextMenuFactory.js
+++ b/client/js/contextMenuFactory.js
@@ -108,6 +108,82 @@ function addQueryItem() {
 	});
 }
 
+function addCloseItem() {
+	function getCloseDisplay(target) {
+		if (target.hasClass("lobby")) {
+			return "Remove";
+		} else if (target.hasClass("channel")) {
+			return "Leave";
+		}
+
+		return "Close";
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("chan"),
+		className: "close",
+		displayName: getCloseDisplay,
+		data: (target) => target.attr("data-target"),
+		callback: (itemData) => utils.closeChan($(`.networks .chan[data-target="${itemData}"]`)),
+	});
+}
+
+function addConnectItem() {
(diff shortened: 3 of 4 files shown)


christer88:
I guess this PR will close https://github.com/thelounge/thelounge/issues/1112 because we now have "Disconnect" to only disconnect, and "Remove" to remove network from json config file? If so, thanks @MaxLeiter!

McInkay:
My poor #1878

creesch:
> Need to deal with add-channel UI when disconnected. Thoughts?

Leave channels with logging in place, disable input field for channels on disconnected networks.

xPaw:
There's no need to disable the input because lounge will join it after connecting.

MaxLeiter:
Ready to go?

McInkay:
Oh please... Mine first.

MaxLeiter:
Fine with me, will be a simple rebase

McInkay:
Well let's see what @astorije says about mine after my refactor earlier today. It's just rebasing mine on top of changes does ideas the likelihood of regressions. I'm not going to ask you to wait weeks.

Haarolean:
Would be nice to be able to change network settings via UI as well. 
https://github.com/thelounge/thelounge/issues/179 is related.

McInkay:
@Haarolean This is unrelated to the PR, please keep relevant discussion to their own tickets, as you have commented on the other issue. Thanks.

astorije:
I don't think "Disconnect" is intentional here, right?

screen shot 2018-03-04 at 20 35 28

astorije:
Do we need both entries "Disconnect" and "Remove" (which also disconnects)?

MaxLeiter:
@astorije: Remove deletes, disconnect disconnects

astorije:
Yes I understand that. Remove also disconnects (right?). But, do we actually need that?

I sort of understand the need for `Connect` (actually, `Reconnect`?) because connection with server can be lost from time to time, and there are cases when we don't automatically reconnect (though we should limit this as much as we can), but why "Disconnect" on top of "Remove"?
It seems to me that the main purpose of TL is to stay connected at all time, reconnect whenever possible, etc. so I'm just trying to understand the need behind this, in case we're trying to solve the wrong problem.

MaxLeiter:
There are times where you want to disconnect so you can reconnect, and there are times when you want to delete the network. Sometimes you want to save the network info but temporally leave it. Pretty common in clients, afaik

astorije:
What about a "Reconnect" instead that disconnects + reconnects? That seems to me to be the end-goal here.

McInkay:
> What about a "Reconnect" instead that disconnects + reconnects? That seems to me to be the end-goal here.

It's the end goal for some of it, sure, but there is definitely a use case for an irc client that you maybe don't always want to be connected to, and making "Disconnect" and "Connect" instead of "Reconnect" allows both use cases.

christer88:
I agree with @McInkay!

MaxLeiter:
Updated links to FA 5, rebased to master, updated screenshots

McInkay:
This will need to be refactored now that #1878 is merged, and use the new structure for adding context menu items. If you have any questions about how this works, or anything is unclear about what I've changed then give me a shout @MaxLeiter

MaxLeiter:
@McInkay: working on updating to master.
2 questions
1) Where should I set `connected`? Currently, I set it in `check()`.
2) `displayName` does not seem to be updating. (As in, always shows "Connected"). Fixing number 1 properly may fix this inherently (like if I make two different `ContextMenuItem`'s)


``` javascript
	let connected;
	contextMenuFactory.addContextMenuItem({
		check: (target) => {
			// not ideal to have the line below here. Not sure where else to place it.
			connected = !target.closest("section").hasClass("not-connected");
			console.log(target.closest("section").hasClass("not-connected"));
			return target.hasClass("lobby");
		},
		className: connected ? "disconnect" : "connect",
		displayName: connected ? "Disconnect" : "Connect",
		data: (target) => target.data("id"),
		callback: (itemData) => {
			if (!connected) {
				socket.emit("input", {
					target: itemData,
					text: "/connect",
				});
			} else {
				socket.emit("input", {
					target: itemData,
					text: "/disconnect",
				});
			}
		},
	});
```

McInkay:
@MaxLeiter 

> Where should I set connected? Currently, I set it in check()

I think you'll need to set it onto the object, otherwise the scope won't work? Maybe I'm wrong.

> displayName does not seem to be updating.

Yeah, this is because you are passing through the string, not a function. Change it to this:

```
className: () => connected ? "disconnect" : "connect",
displayName: () => connected ? "Disconnect" : "Connect",
```

This way the class and display name are recalculated everytime it creates a menu item, rather than just when you create it.

Thanks for the advice, I'll make sure to put this more clearly into the docs when I write them.

Hope that makes sense.

MaxLeiter:
Rebased + updated, running into a slight issue with `className` though -- should not need arrow notation for it to function

McInkay:
> running into a slight issue with className though -- should not need arrow notation for it to function

Yes, you absolutely should.

If you are passing through `connected ? "Disconnect" : "Connect"` to the function, that is processed by Javascript and sent as a String. Therefore it'll always be the same. You need to send a function for it to ever change.

client/js/lounge.js
```diff
@@ -330,14 +330,48 @@ $(function() {
...
+		return "Connect";
+	};
+
+	let clickedNetwork;
+	contextMenuFactory.addContextMenuItem({
+		check(target) {
+			clickedNetwork = target;
+			return target.hasClass("lobby");
+		},
+		className: () => isConnected(clickedNetwork) ? "connect" : "disconnect", // eslint-disable-line no-confusing-arrow
```
astorije: I'm not fan of disabling stuff like this. Use the following instead:

```js
className() {
  return isConnected(clickedNetwork) ? "connect" : "disconnect";
},
```

MaxLeiter:
Yeah just haven’t got to it yet

MaxLeiter:
Should be good for reviews

astorije:
Thanks @MaxLeiter!

client/js/lounge.js
```diff
@@ -330,14 +330,50 @@ $(function() {
...
+	const getToggleConnectionDisplay = (target) => {
+		if (isConnected(target)) {
+			return "Disconnect";
+		}
+
+		return "Connect";
+	};
+
+	let clickedNetwork;
+	contextMenuFactory.addContextMenuItem({
```
astorije: @xPaw raised a good point, could you split this into 2 items? I believe you'd have your `isConnected(target)` in `check`. That would simplify things like `className` 😅

xPaw: Also move all this code into context menu factory.

MaxLeiter: That complicates things like `isConnected()`, will need to play with this

MaxLeiter:
Okay, split up connect/disconnect and moved to factory.

Also moved other context code left-over from lounge.js -> factory, but that required moving `closeChan` to `utils.js` so the Factory could use it. I don't care if we leave the close chan context item in `lounge.js` or not, just let me know.

PR #2115, opened 2018-02-25
Channel leave button tooltip to correspond with right click context menu

christer88:
Currently X button tooltip for channels says "Close", while right click context menu says "Leave". It sounds more normal to leave a channel, and therefor updating the X button tooltip to say leave as well. This makes it more consistent.

capture

Thanks to @MaxLeiter on IRC for the help to make this PR.


astorije:
Thanks and congrats in your first PR @christer88! 🎉👏

PR #2122, opened 2018-02-26
Revert "Allow scaling the page"

astorije:
Reverts thelounge/thelounge#1910
Fixes https://github.com/thelounge/thelounge/issues/2120

It's pretty unfortunate that we have to do this, but on iOS this is not going well at all.
I'm open to trying other settings if anyone wants to help, but I think first step is to revert to go back to an OK state, and then we can iterate.

### DIFF
--- a/client/index.html.tpl
+++ b/client/index.html.tpl
@@ -3,7 +3,7 @@
 	<head>
 
 	<meta charset="utf-8">
-	<meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


astorije:
Nah, it's not a critical bug either. Next v3-pre will do just fine imo

McInkay:
> Nah, it's not a critical bug either.

I mean, it depends on your definition of critical, but alright.

PR #2130, opened 2018-02-28
Don't render mode in usernames on userlist

MaxLeiter:
I'd really like to explore methods of removing IRC-specific jargon out of the UI. I'm almost tempted to also remove modes from the message list. 

I'm sure this will get pushback, but thoughts?

</img>

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -172,6 +172,7 @@ function renderChannelUsers(data) {
 		.find(".search")
 		.prop("placeholder", nicks.length + " " + (nicks.length === 1 ? "user" : "users"));
 
+	data.users.forEach((user) => user.dontRenderMode = true);
 	users
 		.data("nicks", nicks)
 		.find(".names-original")
--- a/client/views/user_name.tpl
+++ b/client/views/user_name.tpl
@@ -1,1 +1,1 @@
-<span role="button" class="user {{colorClass nick}}" data-name="{{nick}}">{{mode}}{{nick}}</span>
+<span role="button" class="user {{colorClass nick}}" data-name="{{nick}}">{{#unless dontRenderMode}}{{mode}}{{/unless}}{{nick}}</span>


dgw:
The only objection I would have to this initiative is, if you eventually remove all of the "jargon" like mode symbols from the UI, how are users supposed to learn it or quickly tell who's a channel administrator/operator/owner versus just a random person talking?

It does look cleaner without the symbols in the user list, but users of The Lounge as a client should be given the opportunity to learn what the mode symbols mean. They're meaningful information when displayed in chat next to people's names when they talk.

astorije:
I support this, and across all displays of nicks (thus getting rid of your need for `dontRenderMode`). Instead, it would be a nice addition in the user context menu: "ChanServ (operator)", "astorije (Voiced user)", etc.

I'm not worried about losing the prefixes because as far as I understand, they're here for display only, so if we provide a better display, that's even better UX, IMO.

dgw:
@astorije I don't think user access level should be hidden away in the user context menu. When a chanop (for example) says "Stop <disruptive thing here>", it should be immediately obvious that that person has some special status, versus a peer saying the same thing.

I don't want to get too involved on this, and I'm not going to fight it if removing mode characters happens because it's just not worth the time and energy. But I do disagree with removing them, for the record.

Jay2k1:
I'm with dgw here. In this very example, nicklist fields have status headers like "Operator" and so on, so as long as the prefixes don't get removed from nicks in the chat, well, it could be worse. There's a redundancy here, although if it was for me, I'd rather remove these headers instead.
The other clients I use don't do that though (removing prefixes from nicks in nicklist), they don't even offer to display it like that afaik. And seeing the prefixes in the chat but not in the user list might be confusing.

The very first sentence directly hits a nerve though.

> I'd really like to explore methods of removing IRC-specific jargon out of the UI

Yeah, because after all, TL is an IRC client, so why should it look like one, right? 
Excuse my snarkiness, but this is a recurring topic here, and I've already ranted about it a couple times. Some people seem to think for whatever reason that this IRC client should hide the fact that it's an IRC client wherever possible, and I don't get why.

McInkay:
I am ... mixed on this.

I definitely like how we are making a more usable client, and removing some of the complexities of irc. However, we are not completely removing any of the functionality or well known parts of irc, just creating more user friendly versions of them. We have buttons for kicking users, but you can still do `/kick user`, stuff like that.

However, I'm not sure about removing prefixes entirely. We should remove IRC complexity, but not IRC entirely. We are not just a client for people new to irc (or who don't want to learn) we are also (some would argue primarily) a client for power users who just like modern UIs. In this case, making someone open the context menu (or scan the user list) to see what a user is could be confusing.

I also agree that removing from userlist and not from chat would be confusing, as at least just now you can look at the userlist to find out what a prefix means.

I am totally for making things more usable, and don't have anything against adding the role to the context menu. I think that would be a great thing. But ... I dunno, prefixes infront of usernames is so fundamental to how some people use IRC, I'm not convinced there is a benefit to this.

xPaw:
> But ... I dunno, prefixes infront of usernames is so fundamental to how some people use IRC, I'm not convinced there is a benefit to this.

Agreed. I would also argue that prefixing nicks with modes in user list and having the headers (Operators, and every user is prefixed with `@`) shows/teaches about modes.

MaxLeiter:
What about if the modes were added to the user list section titles? 

‘Operator (@)’

McInkay:
That could possibly be a way to go.

astorije:
I am not convinced by the arguments so far, and I am still with @MaxLeiter overall. Mode prefixes are an antiquated way to represent something when only text interfaces were available, and it feels to me like this concept has not been challenged much since.

Overall, it seems to me that mode prefixes need to stay because people need to be educated about mode prefixes, which feels like a snake eating its own tail :)
I am in favor of educating newcomers about IRC concepts, I just don't think adding a prefix character in front of all nicks is the correct medium for this. For a concrete example, in the current model, I'm not convinced seeing `@nick` (op) or `^nick` (no special mode) educates anyone.

When it comes to the context menu suggestion (which shows up on any click, so hard to miss), I don't see this as hiding: for example, we've moved the ability to open a query window to the context menu, and it seems to me that sending a direct message is more often useful. Not trying to compare this, but interestingly this is how Slack displays user roles.

That said, I am not going to argue any further, especially if this is one of those one-vs-all debates. I'm just a bit uneasy about directly 👎-ing something because _that's how IRC clients do it_ or _we've always done like this in IRC_. We have challenged quite a few things so far, and based on feedback that's for the better, I don't want to see immediate push back when proposing a novel idea (and no, we're not trying to hide the fact that this is an IRC client whenever possible...).

McInkay:
> Overall, it seems to me that mode prefixes need to stay because people need to be educated about mode prefixes, which feels like a snake eating its own tail :)

No, my reasoning can be summarised as:

1. Mode prefixes allow us a quicker way to see people in authority
2. When this is missing (in slack for example), I find it awkward trying to find who admins are

If I'm overruled, then so be it, but it's definitely not "doing it for the sake of doing it".

dgw:
I know I said I wouldn't spend energy on this, but the way the discussion has been running so far leaves me extremely frustrated with the process, so here goes.

As his very first words earlier this morning, @astorije said:

> I am not convinced by the arguments so far [in opposition to this PR]

But I would bet real money that those of us who are against this PR have not been convinced by arguments in favor of this change, for the simple reason that there haven't been any. At least, not that I can find in this thread on the issue tracker… I'll admit that I don't keep up with much of the chat in The Lounge's own IRC channel, and perhaps this has come up in discussions there. It's not that we wouldn't be open to convincing. It's that nobody's presented any real arguments.

I say this fully aware that my arguments (at least) in this thread have not been particularly rigorous. "Users should be given the chance to learn what the IRC symbols mean" is a non-starter for many of the collaborators and contributors here, and that's fine. But removing the mode symbols _should_ serve a concrete purpose other than "removing IRC-specific jargon out of the UI".

In the process of writing this comment, I've determined probably the biggest source of my frustration: design changes in The Lounge being justified not by how they affect the IRC experience, but by e.g. "that's how Slack does it". At the end of the day, _The Lounge is an IRC client._ It's not a Slack client, a Discord client, a HipChat client, or a client any other chat service but IRC. Not all of its UI design goals can be served by simply imitating how those other services look.

Jay2k1:
I have to admit, my view comes from someone who has been using IRC since 2005 and with a handful of clients (Opera's built-in, mIRC for the most time, irssi, Textual, AndroIRC and TL). I am a sysadmin, a hacker and do web stuff for a hobby. I feel like home on CLIs. To sum it up, I am definitely "oldschool" and a power user.

When I look at a new IRC client, it should look like one to me. I think a rather new client like TL should be modern, it doesn't need to be a console client with an added nicklist. Things like an interface with swiping and media/link embedding absolutely enrich IRC and I am grateful for having them in TL. I am not against bringing IRC into today's "modern age". These features build upon IRC and put additional value on top.

What I am against is removing IRC specific things from the UI to either look more like chat clients for other protocols or be "less confusing" for people who don't know what IRC is. I am **strongly** against this in particular when this is done without an added setting to toggle this behaviour. A more drastic example would be not showing hostmasks of users anymore, like (afair) it used to be with Shout, or hiding the hash symbol from channel names, or (which was also discussed at some point) removing the username field.
In contrast to adding things to the IRC experience (what I talked about earlier), this is about removing things, and I think that's a bad idea.

@astorije said,
> Mode prefixes are an antiquated way to represent something when only text interfaces were available, and it feels to me like this concept has not been challenged much since.

and I agree. They just didn't have anything else back then, that's why we have @ and + and sometimes %. Many things we have today have a similar reason (e.g. the save icon being a floppy disk, while the kids today don't even know what that is).
The question is, does it need to be challenged? What would you gain by doing it, and what would you lose? 

I like how Textual does it. Mikey found a way to make clear the symbol is not part of the nickname, but to people who know how IRC works it's still obvious who is op, voice etc:
image
Add to that the headers of TL ('Operator' etc) and it should become pretty clear to everyone.

RockyTV:
I agree with @jay2k1. If the reason behind this PR is to remove stuff that have been around for years, why not remove everything then? Why not completely hide channel modes, user prefixes and masks, and other stuff that have been around for years and people are used to it? 

I pretty much think that if someone sees `<@user> hello, world!`, they will associate it with IRC. No other protocol has this, as far as I know. It's part of IRC's identity. 

Honestly, what I understood is that TL is trying to turn an IRC client into a standard Slack-era client (really, every modern chat platform is trying to mimic Slack's UI). We don't have to adapt to Slack. If you keep comparing to Slack and what it does/has, why not completely abandon the idea of being a IRC client and just become a Slack client instead?

I didn't want to sound harsh but doing this in a popular client would reduce the number of people using IRC. They would look at TL and ask what protocol it usee, then proceed to use another protocol and ask "what IRC has that Slack doesn't? Why people don't use Slack? Both clients have similar interfaces and behaviors". And by the way, if we are comparing to Slack-era chat services, Discord differentiates users based on their roles, Turbo have an icon next to their name in the chat window and user list, even server owners have a crown icon to show who owns the server.

xPaw:
Well this steamrolled into a big discussion. Let's make one thing clear: we're not going to hide user modes completely. But investigating how they can be displayed differently is something that should be done (e.g. see @Jay2k1's Textual screenshot), because let's be honest displaying mode as part of the nickname is confusing (especially when `@` is also used to auto complete names).

It's obvious that users with mode set need to be differentiated (Discord uses text color, that doesn't work in our case as we have colored nicknames, and comes with a bunch of accessibility problems).

Jay2k1:
I feel like this "thread" turned into more than just discussing the pros and cons of removing mode prefixes from the nicklist, and I think I am to blame.

To try and bring it back to discussing this very PR though, it comes down to this:

- If you remove the mode prefixes from the nicklist but not in the chat, it would be inconsistent and people would wonder why someone has an @ in the chat but not in the nicklist
- If you remove it from the chat too, user modes wouldn't be obvious anymore, and whenever some person says something, you'd have to look them up in all columns (op, voice, halfop, ..., unprivileged) to see what status they have, or right click them -- in my opinion, that's complicating things for no real reason

Even if some may argue that text-based user mode indicators like @, + and % are not very modern, and rightfully so, replacing them with something different like colored dots or icons or whatever would maybe make TL look a tiny bit more modern, but it would confuse everyone who has used IRC before.

xPaw:
> If you remove it from the chat too, user modes wouldn't be obvious anymore, and whenever some person says something, you'd have to look them up in all column

Just to add to that, mode can't be hidden from messages because it can be changed, or the user can leave.

McInkay:
> If the reason behind this PR is to remove stuff that have been around for years, why not remove everything then? Why not completely hide channel modes, user prefixes and masks, and other stuff that have been around for years and people are used to it?

The reason isn't "to remove it for the sake of removing it". I disagree with this PR, but let's not try and mark it as something that it isn't. The reason is to try and make a more user friendly version.

Just like we don't want to remove everything that works, we don't want to keep everything verbatim. 

Let's take your examples

1. Channel modes

These are horrible. `+I` what on earth does this mean to a user? Everytime I have anything to do with a channel that has an invite list, I have to keep going back to my cheatsheets to remember the difference between `+I` and `+i`. These are technical things that power users might want, sure, but normal users don't know what they mean, so we need to find a way to make it more usable and obvious.

2. Host masks

Similarly, why do normal people care? Ops can care because they can use these to ban, but normal users really don't care.

So yeah, actually, why not try and find ways to improve both of these? They are artifacts of an old system that are not intuitive at all.

> I pretty much think that if someone sees <@user> hello, world!, they will associate it with IRC. No other protocol has this, as far as I know. It's part of IRC's identity.

Sure, but that doesn't make it good, necessarily. I'm not saying these are bad and we should get rid of them. But just because something is "part of IRC's identity" doesn't make it good.

Let's take, for example, the fact that originally there was no authentication. Then people wanted authentication and it was implemented by NickServ. That's a really weird way of authenticating. But it's now "part of IRC's identity". Does that mean we keep it? No, plenty of servers are moving towards other ways of authenticating, and hopefully we'll leave NickServ in the past.

> We don't have to adapt to Slack. If you keep comparing to Slack and what it does/has, why not completely abandon the idea of being a IRC client and just become a Slack client instead?

Because Slack isn't an open chat service. It is a closed ecosystem with 1 client. If they gave us an API and allowed us to run our own slack servers, you better bet I'd be writing a client for them. But the things about Slack that I hate...I can't change. The things about the lounge that I dislike? I can change them. 

Is IRC better than Slack? Yes, it's better *because it's open*. But it's worse in many other ways. If IRC was more user intuitive, Slack would probably never have happened. 

No one is saying to have no way to distinguish different user roles. What we are trying to find out is ... is there a better way to do it than a `@` or `&` symbol? The answer...there probably is. But we haven't come to a conclusion yet.

This has clearly turned into a bigger discussion about the direction of the lounge which shouldn't really happen in this PR. But if you want to keep old features of IRC and not improve on them because you want to keep them for the sake of it, then I'm sorry but the lounge isn't the client for you. That's not what we are trying to do. There are plenty of clients that are, and you'll feel more at home there.

If you like the openness of IRC, and you want to create a better web based IRC client that keeps IRC relevant, and can fight against the closed ecosystems of Slack and Discord, this is the place for you. That doesn't mean we'll agree on everything, obviously, but none of us are trying to "hide" irc or "change things for the sake of changing them".

MaxLeiter:
Yeah

PR #2132, opened 2018-02-28
Remove unintentionally included lodash in client build

xPaw:
lodash is *huge*, we don't want to include it for no reason.


astorije:
Supposedly Webpack 4 is much easier and more natural at tree shaking, when we can use it.

PR #2133, opened 2018-02-28
Change ctrl/command up/down to alt/option up/down to align with other apps

McInkay:
In my investigation, Slack, Facebook messenger, and discord all use alt instead of ctrl for moving up and down channels. I think it would be good to align ourselves as it can be confusing when moving between apps.

If we agree to do this, we can either go nuclear and just change it. Or we can keep ctrl as well as alt for a while, and then remove it, or we can keep them both indefinitely.

Thoughts?

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -37,8 +37,8 @@ Mousetrap.bind([
 });
 
 Mousetrap.bind([
-	"mod+up",
-	"mod+down",
+	"alt+up",
+	"alt+down",
 ], function(e, keys) {
 	const channels = sidebar.find(".chan");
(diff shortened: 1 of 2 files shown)


xPaw:
Add Telegram to that list. However, does all this software also use alt on macos?

McInkay:
In mousetrap, alt maps to option on mac.

MaxLeiter:
I think keeping both would be a mistake. It’s not like we’d display something saying “ctrl is now deprecated”

McInkay:
Yup, good catch, I forgot about that. Will get that done this morning.

> but on Macs alt is still considered alt

Ah, that explains it.

McInkay:
Done

PR #2135, opened 2018-02-28
Make code blocks more readable in darker themes

williamboman:


client/themes/zenburn.css
```diff
@@ -221,6 +221,12 @@ body {
 	color: #ccc;
 }
 
+
```
astorije: Yeep, that extra line is causing the linter to fail 😅

client/themes/morning.css
```diff
@@ -195,6 +195,11 @@ body {
 	color: #ccc;
 }
 
+code, .irc-monospace {
```
astorije: Could you add a new line after the commas? Linter is complaining about those.

PR #2143, opened 2018-03-02
Do not crash when user file can not be read or written

### LINKED ISSUE #447: Malformed user configs explode violently
If JSON is incorrect, it seems to be trying to load the file multiple times and printing a stacktrace each time.

### LINKED ISSUE #2128: Crash when TL can't read a user config
When you remove read permissions from a user config file while TL is running, it crashes with
```
fs.js:642
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: EACCES: permission denied, open '/home/Jay2k1/.lounge/users/foobar.json'
    at Error (native)
    at Object.fs.openSync (fs.js:642:18)
    at Object.fs.readFileSync (fs.js:510:33)
    at readUserConfig (/home/Jay2k1/lounge/src/clientManager.js:189:18)
    at ClientManager.loadUser (/home/Jay2k1/lounge/src/clientManager.js:72:21)
    at updatedUsers.forEach (/home/Jay2k1/lounge/src/clientManager.js:57:39)
    at Array.forEach (native)
    at FSWatcher.fs.watch._.debounce (/home/Jay2k1/lounge/src/clientManager.js:57:16)
    at invokeFunc (/home/Jay2k1/lounge/node_modules/lodash/lodash.js:10325:23)
    at trailingEdge (/home/Jay2k1/lounge/node_modules/lodash/lodash.js:10374:18)
``` 
@astorije suggested just issuing a warning instead.

xPaw:
Fixes #2128
Fixes #447

This should handle both JSON and file permission errors.

@Jay2k1 can you test and confirm?

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -132,13 +132,7 @@ ClientManager.prototype.updateUser = function(name, opts, callback) {
 	const user = readUserConfig(name);
 
 	if (!user) {
-		log.error(`Tried to update invalid user ${colors.green(name)}. This is most likely a bug.`);
-
-		if (callback) {
-			callback(true);
-		}
-
-		return false;
+		return callback ? callback(true) : false;
 	}
 
 	const currentUser = JSON.stringify(user, null, "\t");
@@ -185,6 +179,12 @@ function readUserConfig(name) {
 		return false;
 	}
 
-	const data = fs.readFileSync(userPath, "utf-8");
-	return JSON.parse(data);
+	try {
+		const data = fs.readFileSync(userPath, "utf-8");
+		return JSON.parse(data);
+	} catch (e) {
+		log.error(`Failed to read user ${colors.bold(name)}: ${e}`);
+	}
+
+	return false;
 }
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -153,8 +153,6 @@ ClientManager.prototype.updateUser = function(name, opts, callback) {
 		if (callback) {
 			callback(e);
 		}
-
-		throw e;
 	}
 };
 


astorije:
So, I'm having a hard time testing this successfully:

1. The Lounge starts, minding its own business:

```
2018-03-03 04:23:31 [INFO] Available at http://0.0.0.0:9000/ in private mode
2018-03-03 04:23:31 [INFO] User jeremie-local loaded
```

2. In a different window, I run `chmod 400 ~/.thelounge/users/jeremie-local.json`

3. Nothing happens in either the UI or CLI output at first

4. I refresh the UI, The Lounge crashes with:

```
2018-03-03 04:23:37 [ERROR] Failed to update user jeremie-local (Error: EACCES: permission denied, open '~/.thelounge/users/jeremie-local.json')
.../thelounge/src/clientManager.js:163
		throw e;
		^

Error: EACCES: permission denied, open '~/.thelounge/users/jeremie-local.json'
    at Error (native)
    at Object.fs.openSync (fs.js:642:18)
    at Object.fs.writeFileSync (fs.js:1348:33)
    at ClientManager.updateUser (.../thelounge/src/clientManager.js:154:6)
    at Client.updateSession (.../thelounge/src/client.js:311:17)
    at Socket.performAuthentication (.../thelounge/src/server.js:610:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at ..../thelounge/node_modules/socket.io/lib/socket.js:513:12
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
```

---

@Jay2k1, mind sharing the steps you used to test this?

xPaw:
@astorije This PR specifically fixes reading permissions and JSON validness.

The stacktrace you've shown is write permission being denied, which is a slightly different issue.

Jay2k1:
Heh :D 

As for me, I've just set `chmod a-r *.json` and TL instantly crashed before the patch. This didn't happen anymore. I have to admit I didn't do further testing.

xPaw:
@astorije Funnily enough the exception was already handled there, but it was rethrown. I have removed that so it will no longer crash.

astorije:
Alright, so I played with the latest version and made the same permission change that @Jay2k1 mentioned (`chmod a-r *.json`, between the [INFO] and [ERROR] lines below):

screen shot 2018-03-03 at 12 57 55

I think however the user should be removed from the system, no? Similar to removing the file entirely:

screen shot 2018-03-03 at 12 58 17

Similarly when removing write permissions (after re-adding the file):

screen shot 2018-03-03 at 14 32 29

The user gets loaded with write permission error, which means nothing will get properly saved and, unless looking at the CLI output, user/admin will not realize channels/etc. were not saved (also resetting password from the UI doesn't work).

I may be wrong, but I think errors in the user file should just not load / remove the user and treat as in the file was not there (with the error messages still showing up though). That's the difference to me with warnings (deprecations, etc.) which are fine to load the user still.

xPaw:
@astorije I think you're overthinking the problem here. On your first screenshot the user is already loaded (the errors just happen when trying to save it; think no write permission). The errors on your screenshots are just different (read vs update) because they happen on different stages.

If you remove read permission before starting Lounge, the user won't be loaded in the first place. You're seeing the errors because you changed the permission after the user loaded. I don't think this something we should or *need* to handle (realistically, this only complains about not being able to update the file).

astorije:
I actually want to simplify the problem ;)
I think that as soon as a file either loses read or write permission, is not valid JSON, is corrupted / unreadable in any shape or form, we should withdraw from it, i.e. display an error message and remove the user altogether.
I can imagine how holding on to a reference when we can't read the file causes issue, how resetting password from UI without write permissions is bugged, etc. We would be safer to just eject the user.

That said, I don't have a strong opinion, so if you don't want to do that, I'm fine with that.

> (the errors just happen when trying to save it; think no write permission)

No, on the first screenshot, file had read/execute permission, no write. As said I just ran `chmod a-r *.json`, `w` was still there.

xPaw:
> we should withdraw from it

I really don't think we should do it, I can imagine more problems arising if we start adding logic to handle that almost an obscure case.

> I can imagine how holding on to a reference when we can't read the file causes issue

Keep in mind this will only happen if the permissions change *after* the user is loaded, so the admin must be aware of their shenanigans.

> No, on the first screenshot, file had read/execute permission, no write. As said I just ran chmod a-r *.json, w was still there.

Update function first reads the file, so effectively no read permission means no write.

PR #2149, opened 2018-03-04
Cleanup sidebar to use flexbox

xPaw:
First commit removes `div` wrapper added in 0404671 which is not necessary.

The second commit cleans up channels to use flexbox instead of floats. Also fixes up join channel form to use same padding as channels.

These changes should make it easier to further improve the sidebar (e.g. for mobile).

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -249,7 +249,9 @@ kbd {
 #nick button::before,
 #image-viewer .previous-image-btn::before,
 #image-viewer .next-image-btn::before,
-#sidebar .collapse-network::before {
+#sidebar .not-secure-icon::before,
+#sidebar .not-connected-icon::before,
+#sidebar .collapse-network-icon::before {
 	font: normal normal normal 14px/1 FontAwesome;
 	font-size: inherit; /* Can't have font-size inherit on line above, so need to override */
 	-webkit-font-smoothing: antialiased;
@@ -276,14 +278,13 @@ kbd {
 .context-menu-action-kick::before { content: "\f05e"; /* http://fontawesome.io/icon/ban/ */ }
 
 .context-menu-network::before,
-#sidebar .chan.lobby::before,
 #chat .lobby .title::before { content: "\f0a0"; /* http://fontawesome.io/icon/hdd-o/ */ }
 
-#sidebar .not-secure .chan.lobby::before {
+#sidebar .not-secure-icon::before {
 	content: "\f09c"; /* https://fontawesome.com/v4.7.0/icon/unlock/ */
 }
 
-#sidebar .not-connected .chan.lobby::before {
+#sidebar .not-connected-icon::before {
 	content: "\f127"; /* https://fontawesome.com/v4.7.0/icon/chain-broken/ TODO v5: unlink */
 }
 
@@ -555,6 +556,7 @@ kbd {
 
 #sidebar .network,
 #sidebar .network-placeholder {
+	position: relative;
 	margin-bottom: 30px;
 	touch-action: pan-y;
 }
@@ -567,8 +569,8 @@ kbd {
 
 #sidebar .chan,
 #sidebar .chan-placeholder {
-	display: block;
-	padding: 8px 16px 8px 32px;
+	display: flex;
+	padding: 8px 14px;
 	transition: color 0.2s;
 	width: 100%;
 }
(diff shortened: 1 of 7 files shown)


astorije:
Very cool!
I noticed you have some alignment issues:

Before | After
--- | ---
before | after<br><kbd>+</kbd> icon should be closer to the right border to match <kbd>⌄</kbd>'s distance to the left border.
before2 | after2<br>Form should be centered in the sidebar.

dgw:
The join form looks better to me in the after image. Lining up the edges of the form with the edges of the channel list works really well.

astorije:
It feels weird and uncomfortable to me that the form is not visually centered. Unequal margins are just killing me :)

xPaw:
@astorije What I don't like about making the right side match the padding of the arrow is that unread badges look very unbalanced:

xPaw:
Okay I replaced the network icon with the arrow, and moved network status icon separately to the right (and added tooltips).

The only problem I have is that collapsed arrow is changing dimensions due to the transform, and I don't seem to find a solution to that.

astorije:
This is really nice 👏 I love that we can have tooltips on the icons now!

A few things I've noticed:

- Since we're nitpicking on margins, you could remove a bit of the left margin of the arrow, so it's centered between the left side of the window and the network name (removing 5px to both width and padding-left of `#sidebar .network .collapse-network` seems to do the job, so ends up being respectively 39px and 10px). If you also do this on the right-hand side to balance, that's an extra 10px for channel/network names 🤷‍♂️
  If unclear, I am talking of (currently):
  screen shot 2018-03-05 at 23 21 01

- Caret now appears on empty networks.

- Hovering the caret also highlights the server. The opposite works fine, i.e. hovering the server does not highlight the caret.

- Do you want to also get rid of the server icon here: screen shot 2018-03-05 at 23 28 28 (Could use one less problematic icon for Font Awesome v5 transition 😅)

- It looks like channels get double top margin when the join form is opened, not sure if this is intentional or not.

xPaw:
> you could remove a bit of the left margin of the arrow

I centered the arrow on the channel icons below it.

> Caret now appears on empty networks.

Intentionally, otherwise hiding it leaves big empty space.

> Hovering the caret also highlights the server. The opposite works fine, i.e. hovering the server does not highlight the caret.

I'm not sure how I could fix this this CSS.

> Do you want to also get rid of the server icon here:

Nah. It's used in the context menu too.

> It looks like channels get double top margin when the join form is opened

You mean the margin after the join button? That was indeed intentional. I can reduce it.

astorije:
> I centered the arrow on the channel icons below it.

Right, but the channel icons themselves are not centered in the space around them (something like `padding: 8px 14px;`ish instead of `padding: 8px 18px;` L571 would fix that). No big deal though.

> Intentionally, otherwise hiding it leaves big empty space.

I really think this should go. There is no distinction between having no channels or a collapsed network, and the fact that the caret can be clicked and have both states even though there is nothing collapsed/expanded is weird.
I don't mind about the big empty space because it's a temporary/abnormal state that is going to be solved as soon as a channel is joined, but having a button that's effectively not functional is odd.

> Nah. It's used in the context menu too.

I'd be okay if we remove it from there too, but that was more of a thought. We'll rethink icons when we upgrade to FA v5 anyway.

Nice job on the other fixes!

xPaw:
> Right, but the channel icons themselves are not centered in the space around them

Do you mind making a commit to fix this up yourself?

> I don't mind about the big empty space because it's a temporary/abnormal state that is going to be solved as soon as a channel is joined, but having a button that's effectively not functional is odd.

It is indeed an abnormal state, which is why I'm not keen on writing more CSS to cover this case (hide it, and fix empty space, as the arrow isn't absolutely positioned anymore). I think displaying the arrow is completely fine, and is better than the empty space.

astorije:
> Do you mind making a commit to fix this up yourself?

Not at all, done in cb1ad42c, let me know what you think.

> I'm not keen on writing more CSS to cover this case [...] I think displaying the arrow is completely fine, and is better than the empty space.

I very much disagree with this, it simply looks broken at the moment:

broken-collapse

And there isn't much more CSS to write, @McInkay already did it, there was just a bit of adjustment to do.
We should really fix that. I made you 2 versions, you just have to pick one :)

**First version:**

screen shot 2018-03-07 at 18 58 04


```diff
diff --git a/client/css/style.css b/client/css/style.css
index 09e13bf1..fde50569 100644
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -723,6 +723,11 @@ kbd {
 	transform: rotate(45deg) translateZ(0);
 }

+#sidebar .network .lobby:first-child:nth-last-child(2) .collapse-network {
+	/* Hide collapse button if there are no channels/queries */
+	visibility: hidden;
+}
+
 #sidebar .network .collapse-network {
 	width: 40px;
 	opacity: 0.4;
```

**Second version:**

screen shot 2018-03-07 at 18 57 42

```diff
$ git --no-pager diff
diff --git a/client/css/style.css b/client/css/style.css
index 09e13bf1..c22a569f 100644
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -723,6 +723,12 @@ kbd {
 	transform: rotate(45deg) translateZ(0);
 }

+#sidebar .network .lobby:first-child:nth-last-child(2) .collapse-network {
+	/* Hide collapse button if there are no channels/queries */
+	width: 0;
+	overflow: hidden;
+}
+
 #sidebar .network .collapse-network {
 	width: 40px;
 	opacity: 0.4;
```

I'll let you choose, anything is better than a broken button.
You'll notice I even made them in patch format in case you just want to apply yourself (and discovered `git --no-pager` along the way), but I can also make the commit myself, up to you! :)

astorije:
(btw, the force-push was just me rebasing with `master` to test, no changes made to your commits)

McInkay:
Fwiw, I agree with @astorije

And I prefer the empty space one, don't like it expanding to fill the space, looks awkward. And I don't think it massively matters that we have a bit of empty space because, as you both say, it's not like this is a very common thing to have, a server with no channels.

xPaw:
OK, I will hide the button if there are no channels. Using `visibility: hidden` the button is still clickable by the way.

PR #2150, opened 2018-03-04
Cleanup chat/userlist to use flexbox, fix a couple of bugs

### LINKED ISSUE #1548: On iOS, the user list does not seem to be scrollable
When running 2.4.0 and the master build as of this writing, the users list does not seem to be scrollable on iOS for me. It tries to scroll the entire page. The channel list works fine. This is not an issue on Safari for Mac and only affects iOS, so I'm not sure what's going on here and my initial troubleshooting did not turn up an answer. I know it affects at least iOS 10 and 11, and I think it used to work with an older version of The Lounge, but I have not yet had time to check that.

xPaw:
- Chat/sidebar are now positioned using flexbox and not absolute positioning
- Fix toggling userlist not keeping scroll to bottom correctly
- Fix up/down arrows in search bar changing caret position
- Fix error when trying to use arrows while search results are empty
- Changed scrolling code in userlist keybinds to use `scrollIntoView(false)`
- Changed `.sidebar .users` to `.userlist` to reduce confusion
- Userlist toggling is no longer animated, this is an intentional change
- Fixed #1548 - scrolling userlist on iOS devices

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -498,8 +498,8 @@ kbd {
 	float: right;
 }
 
-#viewport.rt #chat .sidebar {
-	right: -180px;
+#viewport.rt #chat .userlist {
+	display: none;
 }
 
 #sidebar {
@@ -954,7 +954,8 @@ button.collapse-network:first-child:nth-last-child(3) {
 }
 
 #chat .chan.active {
-	display: block;
+	display: flex;
+	flex-direction: column;
 }
 
 #chat .condensed {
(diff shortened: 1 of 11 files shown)


astorije:
> Possibly fixed scrolling userlist on iOS devices (#1548), needs testing

Yes! I confirm it's fixed with this PR, switching to "Bug" label.

astorije:
> Userlist toggling is no longer animated, this is an intentional change

Why? I'm not sure I'm in favor of that. Among other things, this means channel list is animated, but user list isn't (channel list animation when clicking the menu button is nice because it mimics the movement the channel list does when swiping it open).

xPaw:
> Why? I'm not sure I'm in favor of that

I really think the animation just gets in the way here, it doesn't add anything. It also looks somewhat disconnected as the chat itself is not animated (we discussed this before, we don't animate it due to reflowing and keeping the scroll position, and importantly performance).

As for the channel list, the current problem I have with it is that it's implemented in a different way (can't even hide it on desktop!).

xPaw:
Okay I went ahead and implemented userlist as an overlay on mobile.

1. On desktop, it pushes chat, there is no animation
2. On mobile, it's an overlay over chat, it's animated

Can you tell me if you prefer this implementation, and if the userlist scrolling still works on iOS.

McInkay:
This looks pretty good, and works quite well.

I think I like the overlay, the chat moving was always a bit odd on mobile. 

One thing I've noticed is that on mobile, if the number of users is low, and it doesn't overflow the screen, then the sidebar just stops at the end of the names. Other than that, seems good in my testing.

xPaw:
@McInkay fixed.

astorije:
> Can you tell me if you prefer this implementation, and if the userlist scrolling still works on iOS.

Yes, this is a great compromise (I also agree that overlay on mobile was a must 👏) and I confirm it still works fine on iOS 🎉

Great stuff, and thanks for being open to alternatives :) I'ma review fully now

PR #2152, opened 2018-03-05
Insert channel/user into channel list at alphabetically sorted point, not just the end

### LINKED ISSUE #53: Sorting of channels / conversations in left sidebar
Currently, newly added channels are just added to the bottom. This is quite strange, especially with private conversations inbetween.

Instead, in the first pass: All channels should be sorted first, and then private conversations. Sorted alphabetically.

In a second pass, we could sort the channels by the amount you use / message in them or something similar. That would make sorting more relevant.
For sure we can sort the private conversations to have the ones with the most recent message first. That’s how any mobile messenger like WhatsApp, Signal etc does it.

McInkay:
Fixes #53

This has 1 issue in that if the user has sorted channels in a differnet way, then it won't work, and it'll be sorted in the first position that it is alphabetically below. But realistically, we don't want to re-sort all the user's chans, because that'll remove the point of having user sorting. And I don't think there's a better way to do this, personally.

### DIFF
--- a/client/js/socket-events/join.js
+++ b/client/js/socket-events/join.js
@@ -10,11 +10,12 @@ const sidebar = $("#sidebar");
 socket.on("join", function(data) {
 	const id = data.network;
 	const network = sidebar.find(`#network-${id}`);
-	network.append(
-		templates.chan({
-			channels: [data.chan],
-		})
-	);
+	const channels = network.children();
+	const position = $(channels[data.index || channels.length - 1]); // Put channel in correct position, or the end if we don't have one
+	const sidebarEntry = templates.chan({
+		channels: [data.chan],
+	});
+	$(sidebarEntry).insertAfter(position);
 	chat.append(
 		templates.chat({
 			channels: [data.chan],
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -128,6 +128,28 @@ Network.prototype.getNetworkStatus = function() {
 	return status;
 };
 
+Network.prototype.addChannel = function(newChan) {
+	let index = this.channels.length; // Default to putting as the last item in the array
+
+	// Don't sort special channels in amongst channels/users.
+	if (newChan.type === Chan.Type.CHANNEL || newChan.type === Chan.Type.QUERY) {
+		// We start at 1 so we don't test against the lobby
+		for (let i = 1; i < this.channels.length; i++) {
+			const compareChan = this.channels[i];
+
+			// Negative if the new chan is alphabetically before the next chan in the list, positive if after
+			if (newChan.name.localeCompare(compareChan.name, {sensitivity: "base"}) <= 0
+				|| (compareChan.type !== Chan.Type.CHANNEL && compareChan.type !== Chan.Type.QUERY)) {
+				index = i;
+				break;
+			}
+		}
+	}
+
+	this.channels.splice(index, 0, newChan);
+	return index;
+};
+
 Network.prototype.export = function() {
 	const network = _.pick(this, [
 		"uuid",
--- a/src/plugins/irc-events/banlist.js
+++ b/src/plugins/irc-events/banlist.js
@@ -37,10 +37,10 @@ module.exports = function(irc, network) {
 				type: Chan.Type.SPECIAL,
 				name: chanName,
 			});
-			network.channels.push(chan);
 			client.emit("join", {
 				network: network.id,
 				chan: chan.getFilteredClone(true),
+				index: network.addChannel(chan),
 			});
 		}
 
--- a/src/plugins/irc-events/join.js
+++ b/src/plugins/irc-events/join.js
@@ -15,12 +15,13 @@ module.exports = function(irc, network) {
 				name: data.channel,
 				state: Chan.State.JOINED,
 			});
-			network.channels.push(chan);
-			client.save();
+
 			client.emit("join", {
 				network: network.id,
 				chan: chan.getFilteredClone(true),
+				index: network.addChannel(chan),
 			});
+			client.save();
 
 			chan.loadMessages(client, network);
 
--- a/src/plugins/irc-events/list.js
+++ b/src/plugins/irc-events/list.js
@@ -46,10 +46,11 @@ module.exports = function(irc, network) {
 				type: Chan.Type.SPECIAL,
 				name: "Channel List",
 			});
-			network.channels.push(chan);
+
 			client.emit("join", {
 				network: network.id,
 				chan: chan.getFilteredClone(true),
+				index: network.addChannel(chan),
 			});
 		}
 
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -67,11 +67,13 @@ module.exports = function(irc, network) {
 						type: Chan.Type.QUERY,
 						name: target,
 					});
-					network.channels.push(chan);
+
 					client.emit("join", {
 						network: network.id,
 						chan: chan.getFilteredClone(true),
+						index: network.addChannel(chan),
 					});
+					client.save();
 					chan.loadMessages(client, network);
 				}
 			}
--- a/test/models/network.js
+++ b/test/models/network.js
@@ -147,4 +147,194 @@ describe("Network", function() {
 			);
 		});
 	});
+
+	describe("#addChannel(newChan)", function() {
+		it("should add channel", function() {
+			const chan = new Chan({name: "#thelounge"});
+
+			const network = new Network({
+				channels: [
+					chan,
+				],
+			});
+			// Lobby and initial channel
+			expect(network.channels.length).to.equal(2);
+
+			const newChan = new Chan({name: "#freenode"});
+			network.addChannel(newChan);
+
+			expect(network.channels.length).to.equal(3);
+		});
+
+		it("should add channel alphabetically", function() {
+			const chan1 = new Chan({name: "#abc"});
+			const chan2 = new Chan({name: "#thelounge"});
+			const chan3 = new Chan({name: "#zero"});
+
+			const network = new Network({
+				channels: [
+					chan1,
+					chan2,
+					chan3,
+				],
+				name: "freenode",
+			});
+
+			const newChan = new Chan({name: "#freenode"});
+			network.addChannel(newChan);
+
+			expect(network.channels[0].name).to.equal("freenode");
+			expect(network.channels[1]).to.equal(chan1);
+			expect(network.channels[2]).to.equal(newChan);
+			expect(network.channels[3]).to.equal(chan2);
+			expect(network.channels[4]).to.equal(chan3);
+		});
+
+		it("should sort case-insensitively", function() {
+			const chan1 = new Chan({name: "#abc"});
+			const chan2 = new Chan({name: "#THELOUNGE"});
+
+			const network = new Network({
+				channels: [
+					chan1,
+					chan2,
+				],
+			});
+
+			const newChan = new Chan({name: "#freenode"});
+			network.addChannel(newChan);
+
+			expect(network.channels[1]).to.equal(chan1);
+			expect(network.channels[2]).to.equal(newChan);
+			expect(network.channels[3]).to.equal(chan2);
+		});
+
+		it("should sort users separately from channels", function() {
+			const chan1 = new Chan({name: "#abc"});
+			const chan2 = new Chan({name: "#THELOUNGE"});
+
+			const network = new Network({
+				channels: [
+					chan1,
+					chan2,
+				],
+			});
+
+			const newUser = new Chan({name: "mcinkay", type: Chan.Type.QUERY});
+			network.addChannel(newUser);
+
+			expect(network.channels[1]).to.equal(chan1);
+			expect(network.channels[2]).to.equal(chan2);
+			expect(network.channels[3]).to.equal(newUser);
+		});
+
+		it("should sort users alphabetically", function() {
+			const chan1 = new Chan({name: "#abc"});
+			const chan2 = new Chan({name: "#THELOUNGE"});
+			const user1 = new Chan({name: "astorije", type: Chan.Type.QUERY});
+			const user2 = new Chan({name: "xpaw", type: Chan.Type.QUERY});
+
+			const network = new Network({
+				channels: [
+					chan1,
+					chan2,
+					user1,
+					user2,
+				],
+			});
+
+			const newUser = new Chan({name: "mcinkay", type: Chan.Type.QUERY});
+			network.addChannel(newUser);
+
+			expect(network.channels[1]).to.equal(chan1);
+			expect(network.channels[2]).to.equal(chan2);
+			expect(network.channels[3]).to.equal(user1);
+			expect(network.channels[4]).to.equal(newUser);
+			expect(network.channels[5]).to.equal(user2);
+		});
+
+		it("should not sort special channels", function() {
+			const chan1 = new Chan({name: "#abc"});
+			const chan2 = new Chan({name: "#THELOUNGE"});
+			const user1 = new Chan({name: "astorije", type: Chan.Type.QUERY});
+			const user2 = new Chan({name: "xpaw", type: Chan.Type.QUERY});
+
+			const network = new Network({
+				channels: [
+					chan1,
+					chan2,
+					user1,
+					user2,
+				],
+			});
+
+			const newBanlist = new Chan({name: "Banlist for #THELOUNGE", type: Chan.Type.SPECIAL});
+			network.addChannel(newBanlist);
+
+			expect(network.channels[1]).to.equal(chan1);
+			expect(network.channels[2]).to.equal(chan2);
+			expect(network.channels[3]).to.equal(user1);
+			expect(network.channels[4]).to.equal(user2);
+			expect(network.channels[5]).to.equal(newBanlist);
+		});
+
+		it("should not compare against special channels", function() {
+			const chan1 = new Chan({name: "#abc"});
+			const chan2 = new Chan({name: "#THELOUNGE"});
+			const user1 = new Chan({name: "astorije", type: Chan.Type.QUERY});
+
+			const network = new Network({
+				channels: [
+					chan1,
+					chan2,
+					user1,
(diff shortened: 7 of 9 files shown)


src/plugins/irc-events/banlist.js
```diff
@@ -37,11 +37,12 @@ module.exports = function(irc, network) {
 				type: Chan.Type.SPECIAL,
 				name: chanName,
 			});
-			network.channels.push(chan);
 			client.emit("join", {
 				network: network.id,
 				chan: chan.getFilteredClone(true),
+				index: network.addChannel(chan),
 			});
+			client.save();
```
xPaw: Is this unintentional?

src/plugins/irc-events/list.js
```diff
@@ -46,11 +46,13 @@ module.exports = function(irc, network) {
...
 				name: "Channel List",
 			});
-			network.channels.push(chan);
+
 			client.emit("join", {
 				network: network.id,
 				chan: chan.getFilteredClone(true),
+				index: network.addChannel(chan),
 			});
+			client.save();
```
xPaw: This too?

McInkay: Yup. Accidental copy paste. Will fix.

astorije:
Very cool! 👍
I haven't looked at the code really well, so apologies if I'm missing something you actually handle:

- What about ignoring non letters in front of channels, so that `#`, `##`, `&`, etc. don't affect sorting? That would mean `#javascript` and `##javascript` would be sorted together, etc.
- A few people have mentioned it would be nice to group channels together, then queries together. I think it works nicely indeed, can your code achieve this?

> This has 1 issue in that if the user has sorted channels in a differnet way, then it won't work, and it'll be sorted in the first position that it is alphabetically below.

Yeah, I would not call this an issue indeed. However, what we could do is: if the list is currently sorted, respect alphabetical order, otherwise just bail out and add at the end of the list.

*The following is outside your PR, just thinking out loud:*
I have seen a lot of chatters about having an alphabetically-sorted list, both on append and as a way to re-sort things. I think at this point I'm very much in favor of getting rid of drag-and-drop reorder, and replacing it with a combination of what you are doing + having a way to pin/start some chats. Add to that quick search of channels, and the channel list would be very nice to use.

McInkay:
> What about ignoring non letters in front of channels, so that #, ##, &, etc. don't affect sorting? That would mean #javascript and ##javascript would be sorted together, etc.

Possibly. How far down this road do we go? So... do we limit it to #, ##, and &, or do we do all non-letters in front of channels? The problem is that `##` doesn't really mean anything, because the `#` is the only bit that defines what the channel is, the second `#` is a server specific thing (freenode) to define something they want. Technically, a channel can only start with `#`, `&`, `+`, and `!` and after that it can basically be anything, so do we ignore all non-letters? I think it's probably easier just to leave it as currently is, because when we start going down that road it gets complicated. Most people will just have `#` channels, with some people having `##` and `&`, and ... I think we don't need to worry about it too much. If we decide to remove user-sorting, then we can come back to it.

> A few people have mentioned it would be nice to group channels together, then queries together. I think it works nicely indeed, can your code achieve this?

The PR already does that. Because it takes the key into consideration :-P So #, etc are always before letters. But if we implement the first, then I'll need to add more logic to 

> Yeah, I would not call this an issue indeed. However, what we could do is: if the list is currently sorted, respect alphabetical order, otherwise just bail out and add at the end of the list.

The problem with this is it essentially doubles the compute time because we need to go through and make sure the list is sorted before figuring out where to put the element. Although, I guess you could do both while going through at the same time so that it doesn't increase worst case complexity (but increases best case complexity from O1 to On. But not a massive issue.

My one problem with this is that if people currently don't have their chanlist sorted, (even just 1 in the wrong place) then we won't put it in the right place, but at the bottom. I'd rather not do this one. Happy to ignore it as being an issue though. I guess, it being put in a random place in the chanlist is no worse than always being put at the bottom.

> I think at this point I'm very much in favor of getting rid of drag-and-drop reorder

Bing bing bing, we have a winner :-P I **knew** you would say that.

I'm trying to think of a good reason not to, and ... well, I can't seem to actually remember the last time I rearranged my channels except when I join a new one, to put it into alphabetical order.

dgw:
> I can't seem to actually remember the last time I rearranged my channels except when I join a new one, to put it into alphabetical order.

Meanwhile I'm trying to remember the last GUI client I used that didn't support drag-and-drop reordering of channels. I can't think of one.

My channels are ordered manually, by how important they are, and losing that sorting ability would take a _lot_ of getting used to. But for what it's worth, sorting alphabetically does seem to be frequently available as an on-demand function via a menu option and/or keyboard shortcut. With the improvements made to The Lounge's context menus so far, "Sort server channels" (name subject to bikeshedding) could become an additional option when right-clicking a network.

McInkay:
> Meanwhile I'm trying to remember the last GUI client I used that didn't support drag-and-drop reordering of channels. I can't think of one.

You don't use slack, discord, facebook messenger, any of these?

> But for what it's worth, sorting alphabetically does seem to be frequently available as an on-demand function via a menu option and/or keyboard shortcut.

I'm not sure I like a button to sort them. It feels like a hacky solution to the problem, tbh.

However, I could be for a setting (an advanced setting) to allow different types of sorting. Manual or alphabetical

xPaw:
> However, I could be for a setting (an advanced setting) to allow different types of sorting. Manual or alphabetical

Overkill in my opinion. The current implementation in this PR should work very well, as it only sorts new channels by one, and not the entire list.

I suggest we go with this for a release, then we can see how well it works in practice for other users.

McInkay:
Yeah, I'm not suggesting that for this PR, that's a bigger question.

dgw:
>> Meanwhile I'm trying to remember the last GUI client I used that didn't support drag-and-drop reordering of channels. I can't think of one.
> 
> You don't use slack, discord, facebook messenger, any of these?

Honestly, no. I don't use Slack at all, and I use Facebook Messenger and Discord almost exclusively through BitlBee (in The Lounge) because I can't stand their official GUIs.

Not that it's relevant what other chat networks I use, right? I've made my thoughts on designing TL as if it's not an IRC client quite clear (in #2130). In the future I will try to remember to always specify "IRC" when I make such statements, to avoid confusion. (I've updated the comment above, as well.)

McInkay:
> Not that it's relevant what other chat networks I use, right? I've made my thoughts on designing TL as if it's not an IRC client quite clear

And I disagree that "it's IRC" has anything to do with it. Manual sorting of channels has nothing to do with IRC, but the UX around a chat client. We can't ignore trends in UX outside of IRC, because that's one of the reasons that Slack and Discord even exist. To ignore that would lead to IRC usage continuing to decline. 

But anyway, the discussion as to whether we keep or remove manual sorting is probably something we should discuss elsewhere. This shouldn't be part of this PR either way. Let's move that discussion to #53

McInkay:
I have removed the 2 unnecessary save() lines but kept the rest the same at this point in time. 

Unless we want to do the alphabetising without the `#` etc (which I don't think we do, but feel free to convince me) then I think we go with this PR as-is and discuss the other stuff elsewhere.

astorije:
> How far down this road do we go?

You mentioned the mandatory first characters (defined in section 1.3 of https://tools.ietf.org/html/rfc2812 if anyone is interested) and because a lot of non-letters can follow that first letter, I would ignore any leading, repeated char. So things like `&`, `#`, `##`, `!!`, etc.
That said, I agree with you that we can keep this simple and get back to this later, or depending on feedback in issues ("why is my channel list not getting ordered as I would expect as a human being?" :D) so I'm fine with holding off on that.

> The PR already does that. Because it takes the key into consideration :-P

I was totally expected this lol. Since we're holding off on "smarter" sorting, that comes for free and that's just fine. 👌 

> The problem with this is it essentially doubles the compute time

Yeah but in the search space of our channel list, that will never be an issue. Even the lousiest smartphone from 10 years ago could sort 1000 strings in a fraction of a second, and it's not something repeated in a given frequency, so I wouldn't be worried about this (plus depending on the implementation, sorting already sorted or almost sorted lists can be virtually free).

But again, while it would be a good way to improve this PR and solve the "issue" you're mentioning, I'm not requesting this, just raising this as an idea if you wanted to "fix" it :)

> I guess, it being put in a random place in the chanlist is no worse than always being put at the bottom.

This is completely speculative so not authoritative, but I can see how it could be confusing that a channel shows up at a random, unexpected place if someone has dozens of channels open. Might be hard to find overall. But that's also something we'd get from feedback, if people open issues saying they can't locate their channels, so we'll see later. I'd say I'd rather choose "not great" over "surprising" (again, not pushing on this, simply justifying where I'm coming from).

> I can't seem to actually remember the last time I rearranged my channels except when I join a new one, to put it into alphabetical order.

This, very much this :D
But yeah, as we're making our client better, it's reasonable to challenge/re-consider solutions we (or Shout) made in the past for lack of a better alternative.

> Meanwhile I'm trying to remember the last GUI IRC* client I used that didn't support drag-and-drop reordering of channels. I can't think of one.

I agree with @McInkay on all the points he mentioned (except the new setting, but you didn't really expect different, did you :D). The Lounge is the only chat application I am using that offers this :)
In fact, it contradicts other paradigms that modern chat apps use. For example, Telegram/Gitter order by most recent activity, Slack can't reorder but offers other paradigms such as starred channels, muted channels, etc.
And yeah, if good UX patterns exist in the field of chat applications, we should at least consider them, whether or not they come from the IRC world (which is not known for its user-friendliness).

But I also totally agree that this PR is not where we will do this, I was just giving thoughts :)

client/js/socket-events/join.js
```diff
@@ -10,11 +10,14 @@ const sidebar = $("#sidebar");
 socket.on("join", function(data) {
 	const id = data.network;
 	const network = sidebar.find(`#network-${id}-chanlist`);
-	network.append(
-		templates.chan({
-			channels: [data.chan],
-		})
-	);
+
+	const channels = network.children();
```
astorije: This seems a bit brittle, for example if we had an extra container or something. Is it worth looking for something a bit more specific, like `.channel` or something? I don't know

src/plugins/irc-events/join.js
```diff
@@ -15,12 +15,12 @@ module.exports = function(irc, network) {
...
 			});
-			network.channels.push(chan);
-			client.save();
+
 			client.emit("join", {
 				network: network.id,
-				chan: chan.getFilteredClone(true),
+				index: network.addChannel(chan),
 			});
+			client.save();
```
astorije: Any consequence in switching `emit`/`save` and, maybe more importantly, not using `chan.getFilteredClone` anymore?
I'm probably missing something but I'm not sure why we don't send `chan` property anymore.

src/plugins/irc-events/message.js
```diff
@@ -67,11 +67,13 @@ module.exports = function(irc, network) {
...
 						name: target,
 					});
-					network.channels.push(chan);
+
 					client.emit("join", {
 						network: network.id,
 						chan: chan.getFilteredClone(true),
+						index: network.addChannel(chan),
 					});
+					client.save();
```
astorije: Why `client.save();` in some files and not in others?

client/js/socket-events/join.js
```diff
@@ -10,11 +10,14 @@ const sidebar = $("#sidebar");
...
+	const position = $(channels[data.index || channels.length - 1]); // Put channel in correct position, or the end if we don't have one
+	const sidebarEntry = templates.chan({
+		channels: [data.chan],
+	});
+	$(sidebarEntry).insertBefore(position);
```
astorije: Sanity check: can you confirm `position` (or `channels.length`) above could never be `0`?

src/plugins/irc-events/message.js
McInkay: We only save when there's something to change. We dont' save in banlist and list because those aren't saved to the user file, so there's no need to save at that point.

src/plugins/irc-events/join.js
McInkay: > Any consequence in switching emit/save

There's only 1 that I can think, and that's that if the server crashes between the 2 commands, then the channel won't be saved in the user's file. I can extract network.addChannel out from the emit itself if you prefer.

> not using chan.getFilteredClone anymore?
> I'm probably missing something but I'm not sure why we don't send chan property anymore.

Not missing anything, refactor error. I didn't test this 1 case after refactoring to move the index calculation into the network. My bad.

client/js/socket-events/join.js
McInkay: Well, the position is not a number itself, that's probably misleading name (but is taken from elsewhere). It's the actual element that we are placing above, if that makes sense. 

So I guess what you actually want is a check that data.index is not 0 or greater than channels.length - 1. On that point... I need to see what happens if we have no channels. This should probably break.

client/js/socket-events/join.js
McInkay: Probably. It's following the same code that we use in the sync_sort though, so we should probably change both of them. Can we do that in another PR?

astorije: Up to you and @xPaw. I'd rather not check in something we want to change directly, but ultimately I don't really mind.

xPaw: @McInkay willing to do a follow up PR?

src/models/network.js
```diff
@@ -123,6 +123,22 @@ Network.prototype.getNetworkStatus = function() {
 	return status;
 };
 
+Network.prototype.addChannel = function(newChan) {
+	let index = this.channels.length - 1;
+
+	// Don't sort special channels in amongst channels/users.
+	if (newChan.type === Chan.Type.CHANNEL || newChan.type === Chan.Type.QUERY) {
+		for (let i = 0; i < this.channels.length; i++) {
```
xPaw: Start `i` at 1 to avoid the lobby (and add a comment about it)

src/models/network.js
```diff
@@ -128,6 +128,22 @@ Network.prototype.getNetworkStatus = function() {
 	return status;
 };
 
+Network.prototype.addChannel = function(newChan) {
```
astorije: Could you add test(s) for this function in [`test/models/network.js`](https://github.com/thelounge/thelounge/blob/yamanickill/sort-channels-on-join/test/models/network.js)?

src/models/network.js
```diff
@@ -128,6 +128,22 @@ Network.prototype.getNetworkStatus = function() {
...
+	let index = this.channels.length - 1;
+
+	// Don't sort special channels in amongst channels/users.
+	if (newChan.type === Chan.Type.CHANNEL || newChan.type === Chan.Type.QUERY) {
+		for (let i = 0; i < this.channels.length; i++) {
+			if (newChan.name < this.channels[i].name) {
```
astorije: Make sure the sort is not case-sensitive (that would also be a nice test case).

McInkay: Oh it will be case sensitive. Good catch.

client/js/socket-events/join.js
McInkay: So I completely forgot that we will always have a channel because the lobby is a channel, so it's fine. So it's not possible for there to be no channels.

This should be fine.

src/models/network.js
McInkay: Done

src/models/network.js
McInkay: Done, I tried to add a test for this, but it's not possible as there is no string that comes before empty string.

src/models/network.js
McInkay: Sorted

McInkay:
I think I have addressed all the comments. Let me know if there are any tests you think I've missed. I manually tested having no channels before adding, but as mentioned that doesn't exist. So this should be good now.

test/models/network.js
```diff
@@ -147,4 +147,130 @@ describe("Network", function() {
...
+
+			expect(network.channels[1]).to.equal(chan1);
+			expect(network.channels[2]).to.equal(chan2);
+			expect(network.channels[3]).to.equal(newUser);
+		});
+
+		it("should sort users alphabetically", function() {
+			const chan1 = new Chan({name: "#abc"});
+			const chan2 = new Chan({name: "#THELOUNGE"});
+			const user1 = new Chan({name: "astorije"});
```
xPaw: You didn't set query type. While it's not important in this test, include it for correctness.

src/models/network.js
xPaw: Shouldn't it be possible to create a network called "xxx" (which in turn creates channel lobby) and add a channel named `aaa` to it? That should keep the order.

test/models/network.js
McInkay: Ah yes, will add that.

src/models/network.js
McInkay: Oh wait, yes, of course, the name of the lobby is coming as '' to me because of course I'm not setting it in the test. Dumb moment.

McInkay:
Alright, I have done all the current comments.

@xPaw mentioned there's an issue with [localecompare in node](https://github.com/nodejs/node/issues/12529) which isn't ideal, however trying to get the workaround working is really horrid and hacky. And using tolowercase wouldn't be any better than we currently have because that sorts the same as localecompare sorts.

I'm not sure what we do with this?

McInkay:
> This'll be an easy fix

test/models/network.js
```diff
@@ -147,4 +147,194 @@ describe("Network", function() {
...
+
+			expect(network.channels[1]).to.equal(chan1);
+			expect(network.channels[2]).to.equal(chan2);
+			expect(network.channels[3]).to.equal(user1);
+			expect(network.channels[4]).to.equal(newUser);
+			expect(network.channels[5]).to.equal(newBanlist);
+		});
+
+		it("should insert before first special channel", function() {
+			const banlist = new Chan({name: "Banlist for #THELOUNGE", type: Chan.Type.SPECIAL});
```
astorije: Until we have a better UI for banlist/channel list, it would be nice if banlists would open right after their corresponding channel, and channel list open at the top of the channel list (well right below server window).

That's an entirely different PR, but just wanted to point that out in case you were thinking about a follow-up PR or something.

PR #2154, opened 2018-03-05
Fix rejectUnauthorized

### LINKED ISSUE #2153: rejectUnauthorized goes to false when modifying user file
Modifying the user file so existing network configs can be reused instead of re-added doesn't work and `rejectUnauthorized` goes to false upon restarting `thelounge`.

xPaw:
Fixes #2153 as the variable names differ.

### DIFF
--- a/client/views/windows/connect.tpl
+++ b/client/views/windows/connect.tpl
@@ -45,8 +45,8 @@
 			</div>
 			<div class="col-sm-9 col-sm-offset-3">
 				<label class="tls">
-					<input type="checkbox" name="allowUnauthorized" {{#unless defaults.rejectUnauthorized}}checked{{/unless}} {{#if lockNetwork}}disabled{{/if}}>
-					Allow untrusted certificates
+					<input type="checkbox" name="rejectUnauthorized" {{#if defaults.rejectUnauthorized}}checked{{/if}} {{#if lockNetwork}}disabled{{/if}}>
+					Only allow trusted certificates
 				</label>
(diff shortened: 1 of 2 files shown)


client/views/windows/connect.tpl
```diff
@@ -45,8 +45,8 @@
 			</div>
 			<div class="col-sm-9 col-sm-offset-3">
 				<label class="tls">
-					<input type="checkbox" name="allowUnauthorized" {{#unless defaults.rejectUnauthorized}}checked{{/unless}} {{#if lockNetwork}}disabled{{/if}}>
-					Allow untrusted certificates
+					<input type="checkbox" name="rejectUnauthorized" {{#if defaults.rejectUnauthorized}}checked{{/if}} {{#if lockNetwork}}disabled{{/if}}>
+					Accept only trusted certificates
```
astorije: I believe this should be `Accept trusted certificates only`, but you'd be a fool to listen to me 🇫🇷 :D
@McInkay, could you confirm?
Apart from that, I'm 👍 on the wording

McInkay: [When your only native English speaker is a Scot ...](https://www.youtube.com/watch?v=sAz_UvnUeuU)

I think `Accept trusted certificates only` would be alright. But we could also do `Only allow trusted certificates`.

xPaw: `Only allow trusted certificates` sounds good to me.

PR #2158, opened 2018-03-06
Do not listen to touch events until client is initialized

### LINKED ISSUE #2148: It should not be possible to swipe the menu open on log in page
This is currently possible on mobile. I could swear we disabled it before, maybe even fixed that very issue before too.

xPaw:
Fixes #2148. I moved the code outside of `libs` to make it align with our other code.

We wouldn't initiate that code twice anyway (e.g. for sliding userlist) as that wouldn't allow us to optimize correctly.


client/js/slideout.js
```diff
@@ -0,0 +1,102 @@
...
+const menu = document.getElementById("sidebar");
+
+let touchStartPos = null;
+let touchCurPos = null;
+let touchStartTime = 0;
+let menuWidth = 0;
+let menuIsOpen = false;
+let menuIsMoving = false;
+
+module.exports = {
```
astorije: Probably silly, but what about:

```js
module.exports = new class SlideoutMenu {
	static enable() { /* ... */ }
	static disable() { /* ... */ }
	status toggle() { /* ... */ }
	status isOpen() { /* ... */ }
};
```

(Would be even nicer if we could create that in a common ancestor file with `new SlideoutMenu()` and pass the ref around, but that's a whooole different arch concern in the current code, hence static functions...)

PR #2159, opened 2018-03-06
Fix default ecdh curve for better compatibility

xPaw:
https://github.com/nodejs/node/issues/16196
https://github.com/nodejs/node/pull/16853

This isn't exactly a problem with The Lounge, but as we support LTS versions, we increase compatibility by including a fix for these Node versions.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -11,6 +11,16 @@ const storage = require("../storage");
 
 process.setMaxListeners(0);
 
+// Fix ECDH curve client compatibility in Node v8/v9
+// This is fixed in Node 10, but The Lounge supports LTS versions
+// https://github.com/nodejs/node/issues/16196
+// https://github.com/nodejs/node/pull/16853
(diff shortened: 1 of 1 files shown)


dgw:
The bug this change fixes manifests as previews failing to load for a small set of websites when The Lounge is run with certain NodeJS versions. TLS cipher negotiation fails because Node's default is not properly set (nodejs/node#16196).

One workaround is to downgrade Node to a version from before the change (v8.5.0). This code change (to auto-negotiate the TLS cipher by default) is the other known workaround, which fixes the issue when running The Lounge under Node v8.6+ and v9.

Node v10 will fix this at the source (nodejs/node#16853), but this will still be needed to support LTS versions until at least December 31, 2019 (when Node v8 support is planned to end).

dgw:
@astorije This affects Node >= 8.6.0, so v8 LTS is included.

#2163 handles restricting the fix to only affected Node versions, so it should be all set until v10 drops.

PR #2166, opened 2018-03-07
Scroll to joined/activated channel

### LINKED ISSUE #2062: Scroll channel list to place joined/switched-to buffer in view
When navigating the channel list with <kbd>Ctrl</kbd>-<kbd>↑</kbd> or <kbd>Ctrl</kbd>-<kbd>↓</kbd>, or joining a new channel (either by command or by the server forcing it), I believe the channel list should scroll up or down until that channel is in view. Not to center it, but just so it's the first/last item visible depending on whether it was hidden above or below the current list viewport.

I don't believe this is a duplicate issue, but apologies if I missed a similar issue while searching.

dgw:
~~WIP because I'm not entirely happy with it, but it does technically function. I need to port this into my production instance for a day or two, because my local TL install isn't behaving right for some reason.~~

I'm using the object option to keep the scrolling down the minimum necessary (`'nearest'`). What @xPaw suggested in #2062, `scrollIntoView(false)`, always aligns the element to the bottom of the viewport, which looks kind of wonky if the target channel is actually above the viewport (it scrolls a long way for no particularly good reason, IMO).

`behavior: 'auto'` as used in this WIP seems to prefer instant scrolling, even on desktop, which doesn't look great. Before this gets merged, I'd like to make that `behavior: 'smooth'` instead, ~~unless there are UX or performance considerations I should know about.~~ but unfortunately it doesn't work consistently and correctly across all browsers. E.g. Chrome 64 appears to just not scroll if given `smooth`.

PR to close #2062 once complete.

Edit: No idea why local `npm test` didn't catch the quotes, but I'll fix that too in the next branch update.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -375,6 +375,7 @@ $(function() {
 				utils.toggleNotificationMarkers(false);
 			}
 
+			utils.scrollIntoViewNicely(self[0]);
 			slideoutMenu.toggle(false);
 		}
 
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -14,6 +14,7 @@ module.exports = {
 	lastMessageId,
 	confirmExit,
 	forceFocus,
+	scrollIntoViewNicely,
 	hasRoleInChannel,
 	move,
 	resetHeight,
@@ -56,6 +57,13 @@ function hasRoleInChannel(channel, roles) {
 // This can only be called from another interactive event (e.g. button click)
 function forceFocus() {
 	input.trigger("click").trigger("focus");
+}
+
+// Reusable scrollIntoView parameters for channel list / user list
+function scrollIntoViewNicely(el) {
+	// Ideally this would use behavior: "smooth", but that does not consistently work in e.g. Chrome
+	// https://github.com/iamdustan/smoothscroll/issues/28#issuecomment-364061459
+	el.scrollIntoView({block: "nearest", inline: "nearest"});
 }
 
 function collapse() {
--- a/client/js/userlist.js
+++ b/client/js/userlist.js
@@ -5,6 +5,7 @@ const fuzzy = require("fuzzy");
 const Mousetrap = require("mousetrap");
 
 const templates = require("../views");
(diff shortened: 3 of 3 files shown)


client/js/lounge.js
```diff
@@ -375,6 +375,7 @@ $(function() {
 				utils.toggleNotificationMarkers(false);
 			}
 
+			self[0].scrollIntoView({behavior: "smooth", block: "nearest", inline: "nearest"});
```
xPaw: Can you make this a helper function so that same parameters can also be used in userlist scrolling without having to duplicate them?

dgw: Done. Still fighting wonk in my development environment (the scroll works if I set a breakpoint and step through the scroll function, but not if I just let the page run uninterrupted) but this should work.

Should I also convert the userlist code?

client/js/utils.js
```diff
@@ -58,6 +59,15 @@ function forceFocus() {
 	input.trigger("click").trigger("focus");
 }
 
+// Reusable scrollIntoView parameters for channel list / user list
```
dgw: This works if I step through it using a debugger breakpoint, but does not appear to actually scroll if I let the page run. But given other little things that don't work quite right in my dev environment, I pushed it anyway to let others with proper dev setups test it.

client/js/utils.js
```diff
@@ -58,6 +59,17 @@ function forceFocus() {
...
+function scrollIntoViewNicely(el) {
+	if (el instanceof $) {
```
xPaw: Doesn't seem worthwhile, just call it on `self[0]`

dgw:
Assuming no further tweaks, I'm ready to squash this once approved. :vulcan_salute:

astorije:
Thanks for this, it's really cool!

And thanks for fixing the userlist scroll that @xPaw broke in #2150! :trollface: 

Before | After
--- | ---
before | after

astorije:
Oops, I merged but meant to approve first.

Virtual post-merge approval! ✅

PR #2172, opened 2018-03-08
Fix background color when hovering users in user list (darker themes)

williamboman:
userlist
userlist1

### DIFF
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -59,6 +59,10 @@ body {
 	color: #fefefe;
 }
 
+#chat .userlist .user.active {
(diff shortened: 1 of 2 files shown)


astorije:
Actually, `master` is even worse than that right now:

screen shot 2018-03-08 at 19 10 35 screen shot 2018-03-08 at 19 10 45

I've added a commit to your PR so I can fix along with the merge :D Not trying to be despotic but it's a very localized and small fix, and I'd like to publish a pre-release tonight!

astorije:
Thanks @williamboman!!

PR #2177, opened 2018-03-09
Always hide horizontal overflow in chat

xPaw:
Resizing the window temporarily shows non-scrollable scrollbar for some reason in Chrome, but then clicking anything recalculates (?) and it disappears.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1041,6 +1041,7 @@ kbd {
 
 #chat .chat {
 	overflow: auto;
+	overflow-x: hidden;
 	display: flex;
 	flex-grow: 1;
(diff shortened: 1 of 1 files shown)


client/css/style.css
```diff
@@ -1041,6 +1041,7 @@ kbd {
 
 #chat .chat {
 	overflow: auto;
+	overflow-x: hidden;
```
MaxLeiter: Is it worth adding a comment here about reasoning?

xPaw: If I'm not wrong, this CSS line already existed before, but got removed when I refactored the flexbox stuff.

PR #2190, opened 2018-03-10
Close sqlite database when user quits

### LINKED ISSUE #2178: SQLite test may fail on CI
https://github.com/thelounge/thelounge/blob/5aa9d7e1dc68ece7edda8742b248e22206107006/test/plugins/sqlite.js#L36-L39

Some kind of race condition is happening and the file is not existing after calling `enable`. Is sqlite doing some async-ness internally that would cause this?

Perhaps using `database.serialize` could be a solution here.

xPaw:
.. or on server shutdown.

Hopefully fixes #2178 by moving file exists check into `after`.

### DIFF
--- a/src/plugins/sqlite.js
+++ b/src/plugins/sqlite.js
@@ -35,7 +35,7 @@ class MessageStorage {
 
 		this.isEnabled = true;
 
-		this.database = new sqlite3.cached.Database(sqlitePath);
+		this.database = new sqlite3.Database(sqlitePath);
 		this.database.serialize(() => {
 			schema.forEach((line) => this.database.run(line));
 
@@ -65,6 +65,22 @@ class MessageStorage {
 
 				this.database.serialize(() => this.database.run("UPDATE options SET value = ? WHERE name = 'schema_version'", currentSchemaVersion));
 			});
+		});
+	}
+
+	close(callback) {
+		if (!this.isEnabled) {
+			return;
+		}
+
+		this.database.close((err) => {
+			if (err) {
+				log.error(`Failed to close sqlite database: ${err}`);
+			}
+
+			if (callback) {
+				callback(err);
+			}
 		});
 	}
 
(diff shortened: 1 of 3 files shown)


src/plugins/sqlite.js
```diff
@@ -35,7 +35,7 @@ class MessageStorage {
 
 		this.isEnabled = true;
 
-		this.database = new sqlite3.cached.Database(sqlitePath);
```
xPaw: Caching doesn't do anything for us. (all it does is keeps internal reference for the filename).

test/plugins/sqlite.js
```diff
@@ -22,25 +22,28 @@ describe("SQLite Message Storage", function() {
 		}
 	});
 
+	after(function(done) {
+		store.close((err) => {
+			expect(err).to.be.null;
+			expect(fs.existsSync(expectedPath)).to.be.true;
```
astorije: Expects in after? This is becoming increasingly bizarre lol. We're not really supposed to do that

xPaw: `after` can be a test just like any other, no? we're just guaranteeing the order here.

astorije: I've noticed in the past that the reporter usually lose it when there is a test in a before/after hook.
Because tests in this file **need** to run in order anyway, and this effectively runs tests, I would move it in a `it` statement at the very bottom of the file.

astorije: Does `expect(fs.existsSync(expectedPath)).to.be.true;` test that the file wasn't destroyed when exiting or something? Do we want/need to test the file is valid, and/or not empty, and/or load, etc.?

xPaw: It's testing that the file is created at the expected path.

astorije: Ah, makes sense. Then in an `it` at the end should do, no?

xPaw: Done.

PR #2192, opened 2018-03-10
Don't access "Notification" if the browser does not support it

### LINKED ISSUE #2191: Synchronization bug on mobile safari: Notification variable not found
The Lounge is running from source (commit 38b8cdb).
ReferenceError: Can't find variable: Notification

After enabling client sync in mobile safari I am no longer able to use the client. I assume this is some issue related to mobile safari not supporting notifications. So the client tries to sync that setting, can't find it, and causes a problem. But that's just my own inference.

creesch:
Fixes #2191

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -150,7 +150,7 @@ function applySetting(name, value) {
 			autocompletion.disable();
 		}
 	} else if (name === "desktopNotifications") {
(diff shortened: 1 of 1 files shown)


zachbr:
Confirming that this fixes the issue.
Thank you!

PR #2194, opened 2018-03-11
Upgrade to Font Awesome 5

### LINKED ISSUE #1823: Upgrade to Font Awesome 5
Font Awesome 5 was released a few days ago, and is really good.
Some icons we use have been redesigned and look much slicker, for example https://fontawesome.com/icons/users?style=solid vs. http://fontawesome.io/icon/users/. Really good stuff.

The transition is not painless though. I [tried it a couple days ago](https://github.com/thelounge/lounge/commit/4118911aa1bd0f6a49e911adb12d336d0012b6a8) and did not complete. If someone wants to give it a shot, you're welcome to do so :)

It would be really cool if we could land this for v3!

xPaw:
Closes #1823

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -283,7 +283,7 @@ kbd {
 #chat .lobby .title::before { content: "\f233"; /* https://fontawesome.com/icons/server?style=solid */ }
 
 #sidebar .not-secure-icon::before {
-	content: "\f09c"; /* https://fontawesome.com/v4.7.0/icon/unlock/ */
+	content: "\f071"; /* https://fontawesome.com/icons/exclamation-triangle?style=solid */
 }
 
 #sidebar .not-connected-icon::before {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1,10 +1,11 @@
 @font-face {
+	/* We use free solid icons - https://fontawesome.com/icons?s=solid&m=free */
 	font-family: "FontAwesome";
 	font-weight: normal;
 	font-style: normal;
 	src:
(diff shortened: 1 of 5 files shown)


xPaw:
@astorije I've replaced `-o` icons with solid ones.

astorije:
Good call. I personally think that almost all `regular` and `light` versions are better than any `solid` version, but the former has way less free icons, and the latter has none, so solid it is.

I have replaced the ugly-looking (personal, of course) unlock icon with warning, what do you think? I think it's reasonable now that we have a tooltip.

Also their globe icon is just horrible, but I looked and there is nothing equivalent to the previous one...

astorije:
> What about https://fontawesome.com/icons/shield-alt?style=solid

I am using this one (and going to use it more) for admin-only stuff, so that wouldn't work, see https://github.com/thelounge/thelounge/pull/1950#issue-161558582.

xPaw:
> I am using this one (and going to use it more) for admin-only stuff, so that wouldn't work

Hm. Exclamation triangle is pretty aggressive, what about just https://fontawesome.com/icons/exclamation?style=solid

astorije:
I personally prefer the triangle one because it's more in line with what you would expect for a warning (it is a warning, really), looks a bit better and self-explanatory than "!" (which could be a notification, etc.), and also slightly easier to hover.

That said, if you feel strongly, I'll go with it 🤷‍♂️ 

`exclamation-triangle` | `exclamation`
--- | ---
screen shot 2018-03-11 at 15 51 04 | screen shot 2018-03-11 at 15 50 30

xPaw:
@astorije That's fair, let's keep it.

PR #2195, opened 2018-03-11
Save collapsed state to localstorage

### LINKED ISSUE #2184: Store server collapsed state across refreshes

McInkay:
Fixes #2184

### DIFF
--- a/client/js/render.js
+++ b/client/js/render.js
@@ -11,6 +11,7 @@ const condensed = require("./condensed");
 const JoinChannel = require("./join-channel");
 const helpers_parse = require("./libs/handlebars/parse");
 const Userlist = require("./userlist");
+const storage = require("./localStorage");
 
 const chat = $("#chat");
 const sidebar = $("#sidebar");
@@ -195,12 +196,18 @@ function renderChannelUsers(data) {
 }
 
 function renderNetworks(data, singleNetwork) {
+	const collapsed = new Set(JSON.parse(storage.get("thelounge.networks.collapsed")));
+
 	sidebar.find(".empty").hide();
 	sidebar.find(".networks").append(
 		templates.network({
 			networks: data.networks,
 		})
 	);
+
+	collapsed.forEach((key) => {
+		collapseNetwork($(`.network[data-uuid="${key}"] button.collapse-network`));
+	});
 
 	// Add keyboard handlers to the "Join a channel…" form inputs/button
 	JoinChannel.handleKeybinds();
@@ -305,8 +312,12 @@ function loadMoreHistory(entries) {
 	});
 }
 
-sidebar.on("click", ".collapse-network", (e) => {
-	const collapseButton = $(e.target);
+sidebar.on("click", "button.collapse-network", (e) => collapseNetwork($(e.target)));
+
+function collapseNetwork(target) {
+	const collapseButton = target.closest(".collapse-network");
+	const networks = new Set(JSON.parse(storage.get("thelounge.networks.collapsed")));
+	const networkuuid = collapseButton.closest(".network").data("uuid");
 
 	if (collapseButton.closest(".network").find(".active").length > 0) {
 		collapseButton.closest(".lobby").click();
@@ -317,10 +328,13 @@ sidebar.on("click", ".collapse-network", (e) => {
 	if (collapseButton.attr("aria-expanded") === "true") {
 		collapseButton.attr("aria-expanded", false);
 		collapseButton.attr("aria-label", "Expand");
+		networks.add(networkuuid);
 	} else {
 		collapseButton.attr("aria-expanded", true);
 		collapseButton.attr("aria-label", "Collapse");
+		networks.delete(networkuuid);
 	}
 
+	storage.set("thelounge.networks.collapsed", JSON.stringify([...networks]));
 	return false;
-});
+}
--- a/client/views/network.tpl
+++ b/client/views/network.tpl
@@ -3,6 +3,7 @@
 	id="network-{{id}}"
 	class="network name-{{slugify name}} {{#if serverOptions.NETWORK}}network-{{slugify serverOptions.NETWORK}}{{/if}} {{#unless status.connected}}not-connected{{/unless}} {{#unless status.secure}}not-secure{{/unless}}"
 	data-id="{{id}}"
+	data-uuid="{{uuid}}"
 	data-nick="{{nick}}"
 	data-options="{{tojson serverOptions}}"
 	role="region"


client/js/render.js
```diff
@@ -195,13 +196,19 @@ function renderChannelUsers(data) {
 }
 
 function renderNetworks(data, singleNetwork) {
+	const networks = JSON.parse(storage.get("networks")) || {};
+	data.networks.forEach((network) => networks[network.uuid] = networks[network.uuid] || {collapsed: false, uuid: network.uuid});
```
xPaw: Change it to store just collapsed UUIDs instead of all of them, then you won't need to do defaults here.

client/js/render.js
```diff
@@ -195,13 +196,19 @@ function renderChannelUsers(data) {
...
+	storage.set("networks", JSON.stringify(networks));
+
 	sidebar.find(".empty").hide();
 	sidebar.find(".networks").append(
 		templates.network({
 			networks: data.networks,
 		})
 	);
 
+	Object.keys(networks).forEach((key) => networks[key] && networks[key].collapsed && collapseNetwork($(`[data-uuid=${key}] button.collapse-network`)));
```
xPaw: Make this multi line please.

Also add `.network` to the selector.

xPaw:
I suspect that networks become uncollapsed on master when websocket reconnects, so this should fix that.

client/js/render.js
McInkay: Fixed

client/js/render.js
McInkay: Fixed

client/js/render.js
```diff
@@ -305,18 +315,24 @@ function loadMoreHistory(entries) {
...
 	collapseButton.closest(".network").toggleClass("collapsed");
 
 	if (collapseButton.attr("aria-expanded") === "true") {
 		collapseButton.attr("aria-expanded", false);
 		collapseButton.attr("aria-label", "Expand");
+		networks[networkuuid] = true;
 	} else {
 		collapseButton.attr("aria-expanded", true);
 		collapseButton.attr("aria-label", "Collapse");
+		networks[networkuuid] = false;
```
xPaw: `delete networks[networkuuid]` instead?

McInkay: I thought about that. But logically, does this make more sense to devs looking?

xPaw: I believe so. You're storing a map of only collapsed networks, not all of their states.

McInkay: No I get that, but my point wsas that `networks[networkuuid] = false` is more obvious what is happening than `delete networks[networkuuid]` I thought, but sure.

client/js/render.js
```diff
@@ -305,18 +315,24 @@ function loadMoreHistory(entries) {
...
 }
 
-sidebar.on("click", ".collapse-network", (e) => {
-	const collapseButton = $(e.target);
+sidebar.on("click", "button.collapse-network", (e) => collapseNetwork($(e.target)));
 
+function collapseNetwork(target) {
+	const collapseButton = target.closest(".collapse-network");
+	const networks = JSON.parse(storage.get("thelounge.networks.collapsed")) || {};
+	const networkuuid = collapseButton.closest(".network").attr("data-uuid");
```
xPaw: `.data("uuid")`?

McInkay: `.data` doesn't refer to the `data-` attributes, does it? I thought it was completely separate to that.

xPaw: It fallbacks to `data-` if `.data()` wasn't changed before (in which case it uses internal cache).

McInkay: Huh, well, I did not know that. Alright, I shall change that up.

client/js/render.js
```diff
@@ -195,13 +196,22 @@ function renderChannelUsers(data) {
...
 	sidebar.find(".networks").append(
 		templates.network({
 			networks: data.networks,
 		})
 	);
 
+	Object.keys(collapsed)
+		.forEach((key) => {
+			if (collapsed[key]) {
+				collapseNetwork($(`.network[data-uuid=${key}] button.collapse-network`));
```
xPaw: Add quotes around `${key}`, missed that before.

client/js/render.js
```diff
@@ -305,18 +315,24 @@ function loadMoreHistory(entries) {
 	});
 }
 
-sidebar.on("click", ".collapse-network", (e) => {
-	const collapseButton = $(e.target);
+sidebar.on("click", "button.collapse-network", (e) => collapseNetwork($(e.target)));
```
astorije: I'd remove `button` from this selector as well.

client/js/render.js
```diff
@@ -305,18 +315,24 @@ function loadMoreHistory(entries) {
...
+function collapseNetwork(target) {
+	const collapseButton = target.closest(".collapse-network");
+	const networks = JSON.parse(storage.get("thelounge.networks.collapsed")) || {};
+	const networkuuid = collapseButton.closest(".network").data("uuid");
 	collapseButton.closest(".network").toggleClass("collapsed");
 
 	if (collapseButton.attr("aria-expanded") === "true") {
 		collapseButton.attr("aria-expanded", false);
 		collapseButton.attr("aria-label", "Expand");
+		networks[networkuuid] = true;
```
astorije: What about using a JS `Set` instead?

- In local storage, it would be stored as an array. When loading from the array, use `new Set(array)`. When writing in local storage, use `JSON.stringify([...set])` or `JSON.stringify(Array.from(set))`
- `if (collapsed[key])` becomes `if (collapsed.has(key))`
- `networks[networkuuid] = true` becomes `networks.add(networkuuid)`
- `Object.keys(collapsed).forEach((key)` becomes `collapsed.forEach((key)`

In terms of performance, users with 10k networks might notice a lag of 15ms 😁

client/js/render.js
McInkay: I feel better having it in because we can never be too specific with these things and it's not like it makes it less efficient, but could stop bugs happening in the future.

client/js/render.js
McInkay: > Add quotes around ${key}

Sure. Weight that it worked, I guess it's just being more accepting of stuff

> And button is probably unnecessary in the selector.

Same to what I said to @astorije ... I'd rather keep it in, tbh.

client/js/render.js
McInkay: Sure

McInkay: Euch, I just realised this'll definitely increase complexity. It goes from O(1) to O(n). I know you don't think that's a big issue, but ... just pointing it out.

astorije: Which part? Isn't JS `Set.has` O(1)?
Also in terms of performance, it's a mixed bag because `collapsed.forEach` is going to go be faster than `Object.keys(collapsed).forEach`, but realistically, in this context, O(n) approximates to O(1) (unlikely someone will have more than... 15 networks, I'd say?)

xPaw: It seems like a lot of unnecessary shuffling to add `Set` here. If these were natively stored in JSON then this would be great, but in this instance just a plain object works fine.

astorije: Plain array would work fine yeah. `Set` is not that bad though, one serialization + one deserialization 🤷‍♂️
But this would be more justified if duplicates was a concern. In the current implementation, it's not.

McInkay: > Which part? Isn't JS Set.has O(1)?

It depends on the implementation, which I can't be bothered finding, but I assume it has to iterate over all the elements in the set to find it if it's in, which would be O(n).

> Also in terms of performance, it's a mixed bag because collapsed.forEach is going to go be faster than Object.keys(collapsed).forEach, 

True, but this happens once every load, whereas checking .has each time we collapse/expand is a lot more.

> but realistically, in this context, O(n) approximates to O(1) (unlikely someone will have more than... 15 networks, I'd say?)

That's not how it works :-P 

You can say that going to O(n) doesn't impact performance too much, sure, but ... it doesn't "approximate to O(1)" by definition, it is much worse :-P 

I don't really care, as you mentioned, it's not like the performance hit is going to be big, and I don't really care.

The code for Set seems a bit more straightforward.

astorije: > That's not how it works :-P

Psssh, [you had to be that guy](https://xkcd.com/386/) :)
I stand by it though. You're saying this statement is wrong, and I'm saying notion of complexity and big-O notation, _in this context_ does not even apply, considering the input size.

Anywayyyy, bikeshedding aside, it looks good! I didn't realize that `Set`s were not always O(1) (TIL: spec apparently requires < O(n), so implementations can be either O(1) or O(log(n))!), and in our input size, plain arrays would do just just fine, up to you.

McInkay:
> Feel free to use plain arrays per @xPaw's suggestion

I'll do it only if @xPaw requires it :-P 

I'm not strongly feeling either way, but I don't care enough to do anything about it. I think using a Set is slightly more logical in terms of how it reads, but don't care either way.

McInkay:
Alright, @astorije said to merge when conflicts sorted. So merging.

PR #2197, opened 2018-03-11
Add support for /ignore, /unignore and /ignorelist commands

RockyTV:

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -48,6 +48,7 @@ const inputs = [
 	"away",
 	"connect",
 	"disconnect",
+	"ignore",
 	"invite",
 	"kick",
 	"mode",
--- a/src/helper.js
+++ b/src/helper.js
@@ -33,6 +33,8 @@ const Helper = {
 	ip2hex,
 	mergeConfig,
 	getDefaultNick,
+	parseHostmask,
+	compareHostmask,
 
 	password: {
 		hash: passwordHash,
@@ -225,4 +227,44 @@ function mergeConfig(oldConfig, newConfig) {
 			return srcValue;
 		}
 	});
+}
+
+function parseHostmask(hostmask) {
+	let nick = "";
+	let ident = "*";
+	let hostname = "*";
+	let parts = [];
+
+	// Parse hostname first, then parse the rest
+	parts = hostmask.split("@");
+
+	if (parts.length >= 2) {
+		hostname = parts[1] || "*";
+		hostmask = parts[0];
+	}
+
+	hostname = hostname.toLowerCase();
+
+	parts = hostmask.split("!");
+
+	if (parts.length >= 2) {
+		ident = parts[1] || "*";
+		hostmask = parts[0];
+	}
+
+	ident = ident.toLowerCase();
+
+	nick = hostmask.toLowerCase() || "*";
+
+	const result = {
+		nick: nick,
+		ident: ident,
+		hostname: hostname,
+	};
+
+	return result;
+}
+
+function compareHostmask(a, b) {
+	return (a.nick.toLowerCase() === b.nick.toLowerCase() || a.nick === "*") && (a.ident.toLowerCase() === b.ident.toLowerCase() || a.ident === "*") && (a.hostname.toLowerCase() === b.hostname.toLowerCase() || a.hostname === "*");
 }
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -18,6 +18,7 @@ const filteredFromClient = {
 	highlightRegex: true,
 	irc: true,
 	password: true,
+	ignoreList: true,
 };
 
 function Network(attr) {
@@ -41,6 +42,7 @@ function Network(attr) {
 			NETWORK: "",
 		},
 		chanCache: [],
+		ignoreList: [],
 	});
 
 	if (!this.uuid) {
@@ -325,6 +327,7 @@ Network.prototype.export = function() {
 		"commands",
 		"ip",
 		"hostname",
+		"ignoreList",
 	]);
 
 	network.channels = this.channels
--- a/src/plugins/inputs/ignore.js
+++ b/src/plugins/inputs/ignore.js
@@ -1,0 +1,120 @@
+"use strict";
+
+const Chan = require("../../models/chan");
+const Msg = require("../../models/msg");
+const Helper = require("../../helper");
+
+exports.commands = [
+	"ignore",
+	"unignore",
+	"ignorelist",
+];
+
+exports.input = function(network, chan, cmd, args) {
+	const client = this;
+	let target;
+	let hostmask;
+
+	if (cmd !== "ignorelist" && (args.length === 0 || args[0].trim().length === 0)) {
+		chan.pushMessage(client, new Msg({
+			type: Msg.Type.ERROR,
+			text: `Usage: /${cmd} <nick>[!ident][@host]`,
+		}));
+
+		return;
+	}
+
+	if (cmd !== "ignorelist") {
+		// Trim to remove any spaces from the hostmask
+		target = args[0].trim();
+		hostmask = Helper.parseHostmask(target);
+	}
+
+	switch (cmd) {
+	case "ignore": {
+		// IRC nicks are case insensitive
+		if (hostmask.nick.toLowerCase() === network.nick.toLowerCase()) {
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
+				text: "You can't ignore yourself",
+			}));
+		} else if (!network.ignoreList.some(function(entry) {
+			return Helper.compareHostmask(entry, hostmask);
+		})) {
+			hostmask.when = Date.now();
+			network.ignoreList.push(hostmask);
+
+			client.save();
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
+				text: `\u0002${hostmask.nick}!${hostmask.ident}@${hostmask.hostname}\u000f added to ignorelist`,
+			}));
+		} else {
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
+				text: "The specified user/hostmask is already ignored",
+			}));
+		}
+
+		break;
+	}
+
+	case "unignore": {
+		const idx = network.ignoreList.findIndex(function(entry) {
+			return Helper.compareHostmask(entry, hostmask);
+		});
+
+		// Check if the entry exists before removing it, otherwise
+		// let the user know.
+		if (idx !== -1) {
+			network.ignoreList.splice(idx, 1);
+			client.save();
+
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
+				text: `Successfully removed \u0002${hostmask.nick}!${hostmask.ident}@${hostmask.hostname}\u000f from ignorelist`,
+			}));
+		} else {
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
+				text: "The specified user/hostmask is not ignored",
+			}));
+		}
+
+		break;
+	}
+
+	case "ignorelist":
+		if (network.ignoreList.length === 0) {
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
+				text: "Ignorelist is empty",
+			}));
+		} else {
+			const chanName = "Ignored users";
+			let newChan = network.getChannel(chanName);
+
+			if (typeof newChan === "undefined") {
+				newChan = client.createChannel({
+					type: Chan.Type.SPECIAL,
+					name: chanName,
+				});
+				client.emit("join", {
+					network: network.uuid,
+					chan: newChan.getFilteredClone(true),
+					index: network.addChannel(newChan),
+				});
+			}
+
+			newChan.pushMessage(client, new Msg({
+				type: Msg.Type.IGNORELIST,
+				ignored: network.ignoreList.map((data) => ({
+					hostmask: `${data.nick}!${data.ident}@${data.hostname}`,
+					when: data.when,
+				})),
+			}), true);
+		}
+
+		break;
+	}
+};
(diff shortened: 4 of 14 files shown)


src/models/network.js
```diff
@@ -162,6 +164,8 @@ Network.prototype.export = function() {
 			return _.pick(chan, keys);
 		});
 
+	network.ignoreList = this.ignoreList;
```
xPaw: This line is unnecessary.

src/models/network.js
```diff
@@ -41,6 +41,7 @@ function Network(attr) {
 			NETWORK: "",
 		},
 		chanCache: [],
+		ignoreList: [],
```
xPaw: Add `ignoreList` to `filteredFromClient`

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,89 @@
...
+		target = args[0].trim();
+		hostmask = Helper.parseHostmask(target);
+	}
+
+	switch (cmd) {
+	case "ignore":
+		if (hostmask.nick !== network.nick && !network.ignoreList.some(function(entry) {
+			return (entry.hostmask.nick === hostmask.nick || entry.hostmask.nick === "*") && (entry.hostmask.ident === hostmask.ident || entry.hostmask.ident === "*") && (entry.hostmask.hostname === hostmask.hostname || entry.hostmask.hostname === "*");
+		})) {
+			Array.prototype.push.apply(network.ignoreList, [{
```
xPaw: `network.ignoreList.push` ?

test/tests/hostmask.js
```diff
@@ -0,0 +1,25 @@
...
+const Helper = require("../../src/helper");
+
+describe("Hostmask", function() {
+	it("should correctly split hostmask components", function() {
+		expect(Helper.parseHostmask("nick!user@host").nick).to.equal("nick");
+		expect(Helper.parseHostmask("nick!user@host").ident).to.equal("user");
+		expect(Helper.parseHostmask("nick!user@host").hostname).to.equal("host");
+	});
+
+	it("fallback to wildcard if components are empty", function() {
```
xPaw: I would like to see more tests for this trying to cover odd cases like empty strings, `!@`, `nick!!!!@thing@@host` and some others.

`compareHostmask` should have some more tests too, where nick is wildcard but hostname is not, etc.

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,89 @@
...
+				when: Date.now(),
+			});
+
+			client.save();
+		}
+
+		break;
+	case "unignore":
+		network.ignoreList.some(function(entry, idx) {
+			if (Helper.compareHostmask(entry.hostmask, hostmask)) {
```
xPaw: I think unignore needs to only remove exact matches, and not wildcards. Also use `forEach` instead of `some` here.

src/helper.js
```diff
@@ -191,3 +193,43 @@ function mergeConfig(oldConfig, newConfig) {
...
+
+	nick = hostmask || "*";
+
+	const result = {
+		nick: nick,
+		ident: ident,
+		hostname: hostname,
+	};
+
+	result.toString = function() {
```
xPaw: This should be removed, as it more than likely will break after being JSON encoded (e.g. server restart). Just format it where it's needed.

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,89 @@
...
+
+	if (cmd !== "ignorelist") {
+		// Trim to remove any spaces from the hostmask
+		target = args[0].trim();
+		hostmask = Helper.parseHostmask(target);
+	}
+
+	switch (cmd) {
+	case "ignore":
+		if (hostmask.nick !== network.nick && !network.ignoreList.some(function(entry) {
```
xPaw: Send an error message when you try to ignore yourself instead of silently doing nothing.

It should also send an error when you try to ignore someone that is already ignored.

src/plugins/inputs/ignore.js
xPaw: Also add messages depending on if unignore worked or not.

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,89 @@
...
+			return Helper.compareHostmask(entry.hostmask, hostmask);
+		})) {
+			network.ignoreList.push({
+				hostmask: hostmask,
```
xPaw: Instead of creating a sub object, I would just add `hostmask.when` so it's flat.

test/tests/hostmask.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+const expect = require("chai").expect;
+const Helper = require("../../src/helper");
+
+describe("Hostmask", function() {
```
astorije: Probably worth colocating the tests in https://github.com/thelounge/thelounge/blob/master/test/src/helperTest.js, and using `describe("#parseHostmask", function() {` (actually should be `".parseHostmask"`).
Thoughts?

xPaw: I really don't mind these function tests being in their own file (a couple of other functions are already separate)

astorije: Fair enough. At least we should split them per function though.

McInkay:
I'd like to block this until #1878 is merged. Hopefully shouldn't be too long as all current comments have been addressed and it's just waiting for a last review from @astorije

src/plugins/irc-events/message.js
```diff
@@ -41,13 +42,23 @@ module.exports = function(irc, network) {
 		let from;
 		let highlight = false;
 		let showInActive = false;
+		let shouldIgnore = false;
 		const self = data.nick === irc.user.nick;
 
+		// Check if the sender is in our ignore list
+		shouldIgnore = network.ignoreList.some(function(entry) {
```
xPaw: You can do `const shouldIgnore = ...` and move it just before `if` where this variable is relevant.

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,117 @@
...
+
+	if (cmd !== "ignorelist") {
+		// Trim to remove any spaces from the hostmask
+		target = args[0].trim();
+		hostmask = Helper.parseHostmask(target);
+	}
+
+	switch (cmd) {
+	case "ignore": {
+		if (hostmask.nick === network.nick) {
```
xPaw: This should be case insensitive check. You can just add `toUpperCase` on both strings.

src/helper.js
```diff
@@ -191,3 +193,40 @@ function mergeConfig(oldConfig, newConfig) {
...
+	parts = hostmask.split("!");
+
+	if (parts.length >= 2) {
+		ident = parts[1] || "*";
+		hostmask = parts[0];
+	}
+
+	nick = hostmask || "*";
+
+	const result = {
```
xPaw: As you asked on IRC, hostmasks indeed also need to be compared case insensitively. I think it's pretty safe to lowercase it in parse function (instead of in compare function all the time).

Don't forget to add tests for lowercasing :)

src/plugins/irc-events/message.js
```diff
@@ -41,13 +42,23 @@ module.exports = function(irc, network) {
...
+		});
+
 		// Server messages go to server window, no questions asked
 		if (data.from_server) {
 			chan = network.channels[0];
 			from = chan.getUser(data.nick);
 		} else {
+			if (shouldIgnore) {
+				return;
+			}
```
astorije: So messages from ignored people are trashed on reception, correct?

I explained that differently [here](https://github.com/thelounge/thelounge/issues/464#issuecomment-373262236), but I think we should not completely delete based ignores on reception. When ignoring someone, I think all their previous messages should be removed too, and when unignoring someone, all messages should re-appear.
Thoughts?

McInkay: I agree. Messages shouldn't disappear into the aether. They should be set as ignored somehow on the message and things that ignore them (highlighting, displaying) should use that feature.

xPaw: That's gonna be much more painful to implement what you are suggesting.

RockyTV: Okay, not wanting to use a fallacy, but every other IRC client just ignores the messages after the mask has been added to the list. And when you remove the mask from the list, previous messages are not shown.

I think there's no easy way to do what you guys are asking for, pretty sure it'd need to make the client aware of these messages and just apply a CSS filter to them or something like that.

dgw: However painful to implement, it's the better UX. I wouldn't say the first iteration of ignore support needs to work that way, since several other IRC clients I've used just irrevocably discard ignored messages. But it's a good goal to have `/ignore` and `/unignore` work retroactively.

McInkay: Yeah, having chatted with @xPaw on IRC, I agree that it would be fine to have this for the first version, but we want to work towawrds it after this PR.

@astorije ?

src/helper.js
```diff
@@ -191,3 +193,44 @@ function mergeConfig(oldConfig, newConfig) {
...
+
+	ident = ident.toLowerCase();
+
+	nick = hostmask.toLowerCase() || "*";
+
+	const result = {
+		nick: nick,
+		ident: ident,
+		hostname: hostname,
+		when: new Date(1970, 0, 1),
```
xPaw: You don't need this line.

src/plugins/irc-events/message.js
astorije: That's fair. Your PR has a lot of useful work around managing ignores, we can improve the actual ignoring later (assuming we will though).

I don't think there needs to be CSS involved though: when a message is received, it is instantiated and added to the channel, but not sent to the client. When a client loads channels, messages are compared to the ignored list and those that should be ignored are not sent to the client.

Of course, it's more complex that just dropping ignored messages, as there will be more edge cases (dealing with messages that usually open a new query window, making this work without reloading the page, etc.) but that can be improved over time :)

Adrian-Turjak:
Can we possibly add slightly more granularity for the ignore functionality when it comes to joins parts, quits? So for channels, and users, having the option to ignore specific interaction types you see, would be good.

E.g. I want to see joins and such for most stuff, but some bot users I don't since they leave/join all the time.

I tried doing this with css (which would work for me), but joins parts and quits do not have data-from for the user they are related to... so I can't.

McInkay:
@Adrian-Turjak Yes, but that'll come later. This is a good start.

McInkay:
This will need to be refactored now that #1878 is merged, and use the new structure for adding context menu items. If you have any questions about how this works, or anything is unclear about what I've changed then give me a shout @RockyTV, either here or on irc.

Adrian-Turjak:
@xPaw 
> You must be using an outdated version, because actions do have data-from set.

We likely are! Will try when we upgrade. We're currently waiting for the next release with the log reloading.

astorije:
Hey @Adrian-Turjak, just curious, when you say "we", do you mean your company? Are you guys using The Lounge for work?

Adrian-Turjak:
@astorije Running it for our team at work on a spare workstation machine with coreos under someone's desk. It's just being used for maybe 6 or so people.

astorije:
@Adrian-Turjak, that's pretty cool! Any chance I could contact you in private (email) to ask you a few follow-up questions on your experience? If you don't want to expose your email on GH, feel free to drop me an email at jeremie(at)astori.fr.

@McInkay, I believe @RockyTV is done updating this PR, so it's open for review again!

RockyTV:
Fixed the merge conflict. Is there anything else I should change or do? If not, then it's ready to be reviewed.

src/client.js
```diff
@@ -46,6 +46,7 @@ const inputs = [
 	"away",
 	"connect",
 	"disconnect",
+	"ignore",
```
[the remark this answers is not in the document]
RockyTV: They're in `ignore.js`, which is a single input file.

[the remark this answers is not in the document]
McInkay: Oh duh, I was thinking this was the list of commands for autocompletion locally, but that's not what this is...sigh. Ignore me, thanks.

astorije:
@xPaw, mind reviewing that? I can also give it a light review after you're okay with it if you want.

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,120 @@
...
+			}));
+		} else if (!network.ignoreList.some(function(entry) {
+			return Helper.compareHostmask(entry, hostmask);
+		})) {
+			hostmask.when = Date.now();
+			network.ignoreList.push(hostmask);
+
+			client.save();
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
```
astorije: This shouldn't be an error. Is there any other type we could use?

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,120 @@
...
+		});
+
+		// Check if the entry exists before removing it, otherwise
+		// let the user know.
+		if (idx !== -1) {
+			network.ignoreList.splice(idx, 1);
+			client.save();
+
+			chan.pushMessage(client, new Msg({
+				type: Msg.Type.ERROR,
```
astorije: Same here.

astorije:
Testing this when rebased on `master`, clicking on "List ignored users" or sending `/ignorelist` doesn't do anything (even after I added someone to the ignore list). Is there something I'm missing?

RockyTV:
@astorije I have double-checked the code and I don't think there's an issue with it. I don't know what happened since last month but lists are now broken, i.e., ban list, ignore list and the channel list aren't being displayed by TL.

RockyTV:
Should be good to go.

astorije:
Alright, testing: `The Lounge source (e81acc4d / v3.0.0-pre.6) (Node.js 8.9.4 on darwin x64)`

`/ignorelist` now works, but right clicking on network and clicking on "List ignored users" does nothing. @RockyTV or @xPaw, could you re-test on your end maybe?

RockyTV:
@astorije Sorry, fixed it now. Should be fine.

src/plugins/inputs/ignore.js
```diff
@@ -0,0 +1,120 @@
...
+				text: "Successfully removed user/hostmask from ignorelist",
```
xPaw: Can we print which hostmask was removed?

src/plugins/irc-events/message.js
zchrykng: It would be nice if it worked similarly to weechat's filters. You can toggle those on and off and all matching messages are hidden/shown depending on the state of the filter.

PR #2198, opened 2018-03-11
Merge pull request #1 from thelounge/master

listairee:
Update Feb 6


astorije:
Hi @listairee, we closed this as it was looking like an error. Let us know if you were trying to do something that didn't turn out the way you expected.

PR #2206, opened 2018-03-13
Add keybinds for cycling through networks/lobbies

### LINKED ISSUE #2203: Cycle through networks with a keyboard shortcut
Could be <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>↑</kbd> (or <kbd>↓</kbd>), as the existing channel-cycling shortcuts are <kbd>Ctrl</kbd>-<kbd>↑</kbd> (or <kbd>↓</kbd>).

<kbd>Alt</kbd> could be used instead of <kbd>Shift</kbd> but for some reason I have a nagging feeling that <kbd>Ctrl</kbd>-<kbd>Alt</kbd>- shortcuts are more likely to be defined at the system level (at least on Windows) and conflict.

Follow-on to #2200.

ESWAT:
Closes #2203. I just used the shortcuts suggested [in this comment](https://github.com/thelounge/thelounge/issues/2203#issuecomment-372371916) for cycling between lobbies.

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -59,6 +59,38 @@ Mousetrap.bind([
 });
 
 Mousetrap.bind([
+	"alt+shift+up",
+	"alt+shift+down",
+], function(e, keys) {
+	const lobbies = sidebar.find(".lobby");
+	const direction = keys.split("+").pop();
+	let index = lobbies.index(lobbies.filter(".active"));
+	let target;
+
+	switch (direction) {
+	case "up":
+		if (index < 0) {
+			target = lobbies.index(sidebar.find(".channel").filter(".active").siblings(".lobby")[0]);
+		} else {
+			target = (lobbies.length + (index - 1 + lobbies.length)) % lobbies.length;
+		}
+
+		break;
+
+	case "down":
+		if (index < 0) {
+			index = lobbies.index(sidebar.find(".channel").filter(".active").siblings(".lobby")[0]);
+		}
+
+		target = (lobbies.length + (index + 1 + lobbies.length)) % lobbies.length;
+
+		break;
+	}
+
+	lobbies.eq(target).click();
+});
+
+Mousetrap.bind([
 	"escape",
 ], function() {
 	contextMenuContainer.hide();
--- a/client/views/windows/help.tpl
+++ b/client/views/windows/help.tpl
@@ -49,6 +49,15 @@
 
 	<div class="help-item">
 		<div class="subject">
+			<kbd class="key-all">Alt</kbd><kbd class="key-apple">⌥</kbd> + <kbd>Shift</kbd> + <kbd>↑</kbd> / <kbd>↓</kbd>
+		</div>
+		<div class="description">
+			<p>Switch to the previous/next lobby in the channel list.</p>
+		</div>
+	</div>
+
+	<div class="help-item">
+		<div class="subject">
 			<kbd class="key-all">Alt</kbd><kbd class="key-apple">⌥</kbd> + <kbd>↑</kbd> / <kbd>↓</kbd>
 		</div>
 		<div class="description">


client/views/windows/help.tpl
```diff
@@ -47,6 +47,15 @@
 
 	<h2>Keyboard Shortcuts</h2>
 
+	<div class="help-item">
+		<div class="subject">
+			<kbd class="key-all">Alt</kbd><kbd class="key-apple">⌥</kbd> + <kbd>shift</kbd> + <kbd>↑</kbd> / <kbd>↓</kbd>
```
xPaw: Can you capitalize `Shift` please.

client/views/windows/help.tpl
```diff
@@ -47,6 +47,15 @@
 
 	<h2>Keyboard Shortcuts</h2>
 
+	<div class="help-item">
+		<div class="subject">
+			<kbd class="key-all">Alt</kbd><kbd class="key-apple">⌥</kbd> + <kbd>Shift</kbd> + <kbd>↑</kbd> / <kbd>↓</kbd>
```
astorije: On the macOS version, it currently looks like this:

screen shot 2018-03-13 at 19 22 35

We tend to use the ⌥/⌘ icons to match how macOS natively shows these things:

screen shot 2018-03-13 at 19 24 09

For Shift, that would be ⇧ (`&#x21E7; – &#8679;`). Would you mind switching to that one?

client/js/keybinds.js
```diff
@@ -58,6 +58,28 @@ Mousetrap.bind([
...
+		target = (lobbies.length + (index - 1 + lobbies.length)) % lobbies.length;
+		break;
+
+	case "down":
+		target = (lobbies.length + (index + 1 + lobbies.length)) % lobbies.length;
+		break;
+	}
+
+	lobbies.eq(target).click();
+});
```
astorije: I've been playing with this and getting a bug when switching from click to keyboard nav, and/or from jumping channels to jumping networks.
I'm not entirely sure of the sequence to repro, but it seems like on a fresh page, if I click on any channel, then go alt+shift+down (which should move me to the first network after the current channel), I jump to the top network.
Would you mind looking into it? 🙏

client/views/windows/help.tpl
ESWAT: So I tried that icon at first but the rendering looked really bad at the small size it appeared… since you’re reformatting things with #2221 maybe you can try it out and see if it works better with the bigger font size.

astorije: Yeah, that's part of what drove me to do #2221. Go with that one, and I'll reformat in that other PR once yours is merged :)

client/js/keybinds.js
ESWAT: OK yeah I was able to reproduce this as well… I’ll check it out.

ESWAT: I’ve fixed this issue; lobby index wasn’t being calculated correctly when the active time is a channel.

However I recommend you play around with it again to see if the fixed behaviour is still desirable.

Example: if you have a channel selected, should `Alt+Shift+Up` select the lobby that channel is in (seems more intuitive UI wise) or should it select the lobby previous to the one the active channel is in (seems more intuitive hierarchy wise and is the current behaviour)

astorije: Alright, trying now!

astorije: Ahah, so I went testing this without reading your last paragraph to avoid bias, and got so confused why it skipped the current network 😅

I think this is a better approach:

> if you have a channel selected, should `Alt+Shift+Up` select the lobby that channel is in (seems more intuitive UI wise)

Otherwise, it feels like it skips the network. Another way to look at it is, if I'm on a channel and want to go to its network, I have to hit `Alt+Shift+Up` then `Alt+Shift+Down`.

What do you think?

client/views/windows/help.tpl
astorije: Actually, feel free to leave as is, I'll switch in my PR, don't worry about this in your PR, it's so close to completion already.

client/js/keybinds.js
ESWAT: Yeah… it felt weird “skipping” to the other network. I’ll get to tweaking that behaviour sometime.

xPaw:
@ESWAT Can you squash the commits please.

astorije:
Congrats on your first merged PR @ESWAT, and welcome to the contributor list!! 🎉

Issue #1869, opened 2017-12-21
Preprocess css files


xPaw:
We need to add compilation to our css files.

Features we absolutely need:
- [x] Minification
- [x] Autoprefixer
- [x] Resolving `@import` (both local files and npm modules)
   - So we can split css into smaller files
   - So we can import modules like `normalize` and `primer-tooltips`
 
I tried this with `postcss` in `webpack`, but it's a pain in the ass to get clean CSS back out (that isn't JS).

astorije:
Also that would solve the fact that `npm run watch` doesn't rebuild on CSS changes.

astorije:
Actually, sometimes it watches, sometimes it does not. Not sure why, I didn't look really hard...

Kinsmir:
@astorije `--watch` depends on the filesystem for tracking changes, in some environments (virtual machines for example) this isn't returned properly.

For that reason you have the option to poll for file changes with `--watch --watch-poll`, this periodically checks (every 1000ms) for changes within your files.

astorije:
@Kinsmir, I don't have a particular setup, just macOS (though some would argue it's a particular setup lol). Also, it's not consistently _not_ working, it usually works fine, except when it doesn't. And only for CSS.
That being said, I didn't look too well, and just restart my `npm run watch` when it does. I'll pay more attention next time.

xPaw:
This was done by #3449

PR #2117, opened 2018-02-25
Upgrade to webpack 4; tracking pr

xPaw:
- [x] Cannot read property 'handlebarsLoader' of undefined
  - Related to https://github.com/pcardune/handlebars-loader/pull/153?
- [x] warning " > babel-loader@7.1.2" has incorrect peer dependency "webpack@2 || 3".
- [x] warning " > istanbul-instrumenter-loader@3.0.0" has incorrect peer dependency "webpack@^2.0.0 || ^3.0.0".
  - https://github.com/webpack-contrib/istanbul-instrumenter-loader/pull/79
  - https://github.com/webpack-contrib/istanbul-instrumenter-loader/pull/84
  - https://github.com/webpack-contrib/istanbul-instrumenter-loader/issues/85
  - https://github.com/thelounge/thelounge/pull/2307
- [x] warning " > mocha-webpack@1.0.1" has incorrect peer dependency "mocha@^2.4.5 || ^3.0.0 || ^4.0.0".
  - https://github.com/zinserjan/mocha-webpack/pull/204
- [x] warning " > mocha-webpack@1.0.1" has incorrect peer dependency "webpack@^2.0.0 || ^3.0.0".
  - https://github.com/zinserjan/mocha-webpack/pull/211
- [x] warning " > mocha-webpack@1.1.0" has incorrect peer dependency "webpack@^2.0.0 || ^3.0.0".
- [x] warning " > webpack-dev-server@2.11.1" has incorrect peer dependency "webpack@^2.2.0 || ^3.0.0".
- [x] warning "webpack-dev-server > webpack-dev-middleware@1.12.2" has incorrect peer dependency "webpack@^1.0.0 || ^2.0.0 || ^3.0.0".
- [x] `yarn test:browser` is empty

PR #2207, opened 2018-03-13
Compile and minify CSS using css-loader

ESWAT:
`style.css` now gets fed through webpack using `css-loader` and `extract-text-webpack-plugin` to solve some low-hanging fruit like local imports and minification. I’ve put the font declarations in a separate file to demonstrate the local import part.

I could have gone ahead and also add `postcss-loader` but figured to go one step at a time; can probably get a lot of mileage out of just being able to have multiple CSS files from source.

Related to #1869 and #1551.

*Note*: For some reason the sourcemap for the CSS gets generated and put in the `public/css` directory even if I tested with setting the `sourceMap: false` declaration for the `css-loader` options… if anyone has any ideas?


xPaw:
Thanks for doing this. Have you look at the changes in webpack 4 in regards to CSS yet? It might be worthwhile to wait for webpack 4 before attempting this.

> For some reason the sourcemap for the CSS gets generated and put in the public/css directory

Is that a problem?

ESWAT:
> Thanks for doing this. Have you look at the changes in webpack 4 in regards to CSS yet? It might be worthwhile to wait for webpack 4 before attempting this.
I have not but can test out what a similar webpack 3 config with these packages would need to upgrade to webpack 4.

> Is that a problem?
I guess not. This actually made me realize I should make a dev config for the CSS that doesn’t minimize the CSS.

astorije:
Unfortunately, some of the Webpack plugins we use seem to stagnate and it might be a while before we can upgrade to Webpack v4 (unless we're okay with forks, if/when they're available).
@xPaw, wouldn't you prefer going forward with this and make progress on our end, and worry about Webpack v4 upgrade when we actually can?

webpack.config.js
```diff
@@ -87,8 +105,8 @@ const config = {
 				from: "./client/themes/*",
 				to: "themes/[name].[ext]",
 			},
-			{ // TODO: Build css with postcss
-				from: "./client/css/*",
+			{
+				from: "./client/css/bootstrap.css",
```
xPaw: Can you just add that as `@import` (same with tooltips)

webpack.config.js
```diff
@@ -20,6 +21,23 @@ const config = {
...
+				include: [
+					path.resolve(__dirname, "client"),
+				],
+				use: ExtractTextPlugin.extract([
+					{
+						loader: "css-loader",
+						options: {
+							allChunks: true,
+							importLoaders: 1,
+							minimize: process.env.NODE_ENV === "production" ? true : false,
```
xPaw: `? true : false` is unnecessary. It already evaluates to a boolean.

xPaw:
@astorije My only worry is `ExtractTextPlugin`, which I believe is no longer necessary in v4, so I'd like to know how that's handled there.

astorije:
Not sure if you've seen this or if it's any relevant, but: https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/701#issuecomment-374551300

xPaw:
Gonna close this as I tackled it in #2117.

PR #2210, opened 2018-03-13
Use undate library to handle modifier keys

xPaw:
This is the same library that is used in autocompletion to handle inserts (it's from the same author).

The main advantage here is that undo (ctrl-z) starts working.

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -2,7 +2,8 @@
 
 const $ = require("jquery");
 const Mousetrap = require("mousetrap");
-const input = $("#input");
+const wrapCursor = require("undate").wrapCursor;
+const input = $("#input").get(0);
 const sidebar = $("#sidebar");
 const windows = $("#windows");
 const contextMenuContainer = $("#context-menu-container");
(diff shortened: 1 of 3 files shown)


client/js/keybinds.js
```diff
@@ -78,21 +79,8 @@ for (const hotkey in colorsHotkeys) {
...
-			// If text is selected, insert formatting character at start and the end
-			newValue += value.substring(cursorPosStart, cursorPosEnd) + colorsHotkeys[e.key] + value.substring(cursorPosEnd, value.length);
-		}
-
-		input
-			.val(newValue)
-			.get(0).setSelectionRange(cursorPosStart + 1, cursorPosEnd + 1);
+		const modifier = colorsHotkeys[e.key];
+
+		wrapCursor(input, modifier, input.selectionStart === input.selectionEnd ? "" : modifier);
```
astorije: Ok so just making sure I read this correctly:

- If there is text selected, add `modifier` characters around that selection.
- If there is nothing selected, effectively "wrap" `modifier` on the left and empty string on the right.

Also if I'm reading this correctly: intuitively it seems like the second case should just be: if no text is selected, simply add character, but it's beneficial to re-use `wrapCursor` with that empty string trick because 1) it avoids tampering the string ourselves to insert the char, 2) whatever undo mechanism the first scenario is offered with the first scenario also applies in the second scenario.

Does that sound correct to you?

xPaw: > If there is nothing selected, effectively "wrap" modifier on the left and empty string on the right.

That's essentially right. I had to add this selection check because that lib *always* wraps around cursor (selection) position.

astorije: :+1: :shipit:

PR #2212, opened 2018-03-13
Allow hiding channel list on desktop

### LINKED ISSUE #17: Fold channel list
All's in the title, this is a feature request.

I can fold user list but not the left hand channel list. If I split my screen in half with something else, half of lounge is about channel names and message time. So I'ld like a button to fold it like user list.

xPaw:
Fixes #17.

Works the same way user list does.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -221,7 +221,6 @@ kbd {
 #viewport .rt::before,
 #chat button.menu::before,
 #sidebar .chan::before,
-#chat .title::before,
 #footer .icon,
 #chat .count::before,
 #settings .extra-experimental,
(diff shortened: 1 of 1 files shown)


McInkay:
> I admit I'd much prefer to have a transition, but I know you won't want that :(

Yeah, that can be added in a separate PR if we feel we want it, but this PR is at least consistent with what the user list currently is, and we don't want to hold up the functionality with a debate about transitions.

FWIW, I'm not against transitions, totally happy to investigate that. I think the issue before was it was janky because only the userlist transitioned, the chat didn't, which was odd.

xPaw:
Transition is expensive because it has to shrink chat container which would cause a lot of updates. If chat container is instantly resized (like it was with user list before) it looks odd to only animate the menu sliding in.

McInkay:
As I say, it can be argued in another PR/issue :-P

PR #2215, opened 2018-03-13
Do not aggressively focus input when clicking anywhere and handle typing out of focus

### LINKED ISSUE #1428: Focus input when typing into "nothing"
Actions like clicking a link in the chat can put the focus outside the input, after which sending another message requires clicking the input box again.

One of the things I love about my desktop client (Textual) is that it will focus the input if I start typing but the focus is not in the input. Would it be practical for The Lounge to do this? Would it conflict at all with accessibility concerns surrounding screenreaders, keyboard shortcuts (existing or future/planned), etc.?

xPaw:
- Ctrl, alt, meta keys do not focus.
  - Ctrl+V focuses to allow pasting.
- Shift only focuses if it's used in combination with other key.
- Tab key does not focus.
- Enter key focuses but eats the input so the message isn't sent

Fixes #1428

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -223,16 +223,6 @@ $(function() {
 				input.trigger("focus");
 			}
 		};
-
-		$(window).on("focus", focus);
-
-		chat.on("click", ".chat", function() {
-			setTimeout(function() {
-				if (window.getSelection().isCollapsed) {
-					focus();
-				}
-			}, 2);
-		});
 	}
 
 	if (navigator.platform.match(/(Mac|iPhone|iPod|iPad)/i)) {
@@ -241,7 +231,7 @@ $(function() {
 
 	$("#form").on("submit", function(e) {
 		e.preventDefault();
-		utils.forceFocus();
+		focus();
 		const target = chat.data("id");
 		const text = input.val();
 
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -107,6 +107,11 @@ const colorsHotkeys = {
 
 for (const hotkey in colorsHotkeys) {
 	Mousetrap.bind("mod+" + hotkey, function(e) {
+		// Do not handle modifier hotkeys if input is not focused
+		if (document.activeElement !== input[0]) {
+			return;
+		}
+
 		e.preventDefault();
 
 		const modifier = colorsHotkeys[e.key];
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -108,7 +108,7 @@ const colorsHotkeys = {
 for (const hotkey in colorsHotkeys) {
 	Mousetrap.bind("mod+" + hotkey, function(e) {
 		// Do not handle modifier hotkeys if input is not focused
-		if (document.activeElement !== input[0]) {
+		if (document.activeElement !== input) {
 			return;
 		}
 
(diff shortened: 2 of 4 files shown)


astorije:
I can confirm, it works great.

Since we have to deal with special cases anyway, would you mind letting pasting (Ctrl/cmd+V) go through? That way one can select on the screen, hit Ctrl/cmd+C, and directly hit Ctrl/cmd+V to paste in the input. I realize I tested that unconsciously as I took it for granted from other chats :)

If I'm not focused, and I hit cmd+I then a word, I remember you mentioning something like this, did you mean to make styling controls focus as well?

When not focused and hitting Enter, it adds a new line. On one end it is weird that it adds a new line seemingly randomly, on the other hand it's a bit scary that hitting Enter could send your message even when not focused.
I played with... no, I'm not gonna name that other chat starting with an S, to see how they were doing it, and if not focused, hitting Enter focuses on the input, so you'd have to hit Enter twice (and not seeing the focus ring, though we don't have one _for now_) to send a message. I find this solution rather clever.

> Tab key does not fixed.

Is that because of nick completion stuff?

xPaw:
> If I'm not focused, and I hit cmd+I then a word, I remember you mentioning something like this, did you mean to make styling controls focus as well?

On master it inserts a styling character but doesn't focus. In this PR it changed to not insert a character.

> Is that because of nick completion stuff?

It doesn't focus to allow tabbing (and shift tabbing) around.

astorije:
(Rebased to fix conflict caused by #2073)

> On master it inserts a styling character but doesn't focus. In this PR it changed to not insert a character.

Ah, gotcha. Don't you think it should add the character though? I don't know, it's a very niche use case and by no means I'm blocking this PR until there is an answer, just wondering.

> It doesn't focus to allow tabbing (and shift tabbing) around.

Right, but I noticed <kbd>Tab</kbd> (not <kbd>Shift</kbd>+<kbd>Tab</kbd>) doesn't work, as in, it should tab to the submit button but it doesn't. I am wondering if that's caused by the nick completion intercepting <kbd>Tab</kbd> or your PR.

client/js/lounge.js
```diff
@@ -224,14 +224,40 @@ $(function() {
...
+				return;
+			}
+
+			// Ignore shift key press if it's not combined
+			if (e.shiftKey && e.which === 16) {
+				return;
+			}
+
+			// Ignore all ctrl keys except for ctrl+v to allow pasting
+			if (e.ctrlKey && e.which !== 86) {
```
astorije: This doesn't work on macOS 😭

My understanding is that `metaKey` is <kbd>cmd/⌘</kbd> on Mac (on Mac copy/pasting is ⌘C ⌘V, not Ctrl+C Ctrl+V, in fact, _most_ things are ⌘ instead of Ctrl).

Alternatively, can you make use of Mousetrap for this code? It seems to handle the Ctrl/Cmd/Alt mess pretty well.

xPaw: Damn macs! I can't really use mousetrap here because the code is very specific and uses `keydown` event.

I'll see what I can do.

xPaw: Should be fixed now.

astorije:
> I think it's a bit of an unexpected behaviour to do that. By not using these keybinds we allow users to use them out of the input (e.g. accessibility tools or custom keybinds).

Yes, testing this again, it's much better indeed. I missed one part before, totally agree with your statement.

> I think this can be look at in a future PR

Yep, just wanted to check with you that it was more likely to be related to nick completion than your code, as nothing was pointing out to your code.

client/js/lounge.js
astorije: Yep, works perfectly.

McInkay:
Will test sometime today.

McInkay:
Few things I've noticed:

* Caps lock focuses. Should it? Not a big deal.
* Modifiers (ctrl+b for bold, etc) don't work at all now, even when the thing is in focus
* Should we remove focusing in the input when we switch channel with this change? There are a few things I think could work (like clicking on a network and using left/right to expand/collapse) but only if we remove this. This feels unnecessary now (and could still cause issues on touch screen devices). I think we should remove that code.

Other than that, looks good I think.

xPaw:
> Caps lock focuses. Should it? Not a big deal.

I'll take a look.

> Modifiers (ctrl+b for bold, etc) don't work at all now, even when the thing is in focus

Got broken by `undate` PR, fixed.

> Should we remove focusing in the input when we switch channel with this change?

It seems pretty consistent with other chat apps that switching to a channel puts focus. I say we keep this behaviour for now.

McInkay:
> It seems pretty consistent with other chat apps that switching to a channel puts focus. I say we keep this behaviour for now.

Yeah, I can see that, I'm just not sure what the purpose is when we will jump to the input when they start typing anyway, you know?

PR #2221, opened 2018-03-14
Improve readability and styling of shortcut keys in the Help section

astorije:
For example, on macOS, shortcuts are displayed such as "⌘C" while on Windows they would be displayed as "Ctrl+C":

screen shot 2018-03-13 at 19 24 09 

Also on Mac these characters in key boxes are reaaally tiny, so I went simpler as I figured simple trumps fancy here (and imo that wasn't really pretty either but that's subjective).

Before | After
--- | ---
screen shot 2018-03-13 at 22 21 12 | screen shot 2018-03-13 at 22 19 18
screen shot 2018-03-13 at 22 20 02 | screen shot 2018-03-13 at 22 19 39

Thoughts?

_I know that this PR is going to conflict with #2206, which should go first, so @ESWAT won't have to deal with conflicts ;)_

### DIFF
--- a/client/views/windows/help.tpl
+++ b/client/views/windows/help.tpl
@@ -49,25 +49,48 @@
 
 	<div class="help-item">
 		<div class="subject">
-			<kbd class="key-all">Alt</kbd><kbd class="key-apple">⌥</kbd> + <kbd>Shift</kbd> + <kbd>↑</kbd> / <kbd>↓</kbd>
+			<span class="key-all"><kbd>Alt</kbd> <kbd>Shift</kbd> <kbd>↓</kbd></span>
+			<span class="key-apple"><kbd>⌥</kbd> <kbd>⇧</kbd> <kbd>↓</kbd></span>
 		</div>
 		<div class="description">
-			<p>Switch to the previous/next lobby in the channel list.</p>
+			<p>Switch to the next lobby in the channel list.</p>
 		</div>
 	</div>
 
 	<div class="help-item">
 		<div class="subject">
-			<kbd class="key-all">Alt</kbd><kbd class="key-apple">⌥</kbd> + <kbd>↑</kbd> / <kbd>↓</kbd>
+			<span class="key-all"><kbd>Alt</kbd> <kbd>Shift</kbd> <kbd>↑</kbd></span>
+			<span class="key-apple"><kbd>⌥</kbd> <kbd>⇧</kbd> <kbd>↑</kbd></span>
 		</div>
 		<div class="description">
-			<p>Switch to the previous/next window in the channel list.</p>
+			<p>Switch to the previous lobby in the channel list.</p>
 		</div>
 	</div>
 
 	<div class="help-item">
 		<div class="subject">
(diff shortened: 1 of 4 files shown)


astorije:
Advantage of splitting this is that I can keep the `<kbd>` for non-Mac, and use text-only for Mac 🤷‍♂️
I have no problem with that, getting rid of them on Mac is what makes sense in terms of readability, they're still readable with `<kbd>` on non-Mac.

I'll let the other reviewer(s) speak and go with whatever they prefer.

astorije:
And thanks for your review btw @dgw!

xPaw:
I agree with @dgw, I think keeping `<kbd>` is better, but need to improve their styling a little.

client/views/windows/help.tpl
```diff
@@ -49,16 +49,28 @@
...
 		<div class="description">
-			<p>Switch to the previous/next window in the channel list.</p>
+			<p>Switch to the next window in the channel list.</p>
 		</div>
 	</div>
 
 	<div class="help-item">
 		<div class="subject">
-			<kbd class="key-all">Ctrl</kbd><kbd class="key-apple">⌘</kbd> + <kbd>K</kbd>
+			<span class="key-all">Alt+Up</span>
```
xPaw: I would keep arrows here too, it's pretty common.

astorije: What does your OS commonly do?

astorije:
Ok, I'll keep kbd for non-Mac and we can improve the styling separately

astorije:
@xPaw and @dgw, what do you think about this, [loosely inspired from this](https://kremalicious.com/using-kbd-for-fun-and-profit/):

Example | Morning | Zenburn
--- | --- | ---
screen shot 2018-03-15 at 02 01 10 | screen shot 2018-03-15 at 02 01 25 | screen shot 2018-03-15 at 02 02 06
screen shot 2018-03-15 at 02 02 49 | screen shot 2018-03-15 at 02 03 08 | screen shot 2018-03-15 at 02 03 19

I'm pretty happy with the result, because the font size is now the same as text, so the readability argument of this original PR is maintained!

xPaw:
@astorije I think the buttons look even more out of place now with the gradients and soft shadows. Perhaps a sharper look will match our UI a bit better?

astorije:
@xPaw, I have removed the outer box shadow, reduced the "3D effect" (not sure how you'd call that), and reduced the size of the key to make it blend in better. Overall sharper (sharp edges since there is no outer shadow), stands out less, but still keeping the targeted font size.
I think it's a good in-between, definitely better than `master` (it also respects dark themes). Plus I really want to move on to other things 😁

astorije:
@xPaw, like we discussed on IRC, very much reduced the 3D effect / shadows / gradients on the dark ones to make them sharper and flatter. They're now very close to the light ones in terms of effect. What do you think?

Before | After
--- | ---
screen shot 2018-03-19 at 01 42 00 | screen shot 2018-03-19 at 01 43 16
screen shot 2018-03-19 at 01 40 55 | screen shot 2018-03-19 at 01 41 20

PR #2225, opened 2018-03-14
Select lobby of network when collapsing network

### LINKED ISSUE #2224: Select lobby when collapsing a network
When you are on a channel, and you collapse a network, it should probably select the lobby of that network.

McInkay:
Fixes #2224 

- [x] Only select lobby if we are on a channel in that network, don't switch network


McInkay:
Updated to fix the issue mentioned on #2224. We only switch if we are on a channel in that network, otherwise we don't.

Issue #584, opened 2016-08-26
Add 'commands' input box to join network 


MaxLeiter:
Sorry if an issue already exists, there should be a 'commands' input box when connecting to a new network

PR #2229, opened 2018-03-14
Allow editing networks via UI

### LINKED ISSUE #179: Renaming networks

### LINKED ISSUE #1161: Edit user commands from interface
- Commands should be editable from connect menu
- Make possible to open connect menu for existing networks to change settings

xPaw:
Some of the network code moved, and the editing code had to be done in such a way to apply same restrictions as adding a new connection. So this needs a lot of testing.

- [x] Fixes #179
- [x] Fixes #1161 (and practically #584, but the UI only exists when editing, not connecting, which is fine by me)
- [x] Update lobby name in UI if changed
- [x] Send `/nick` if nick changes
- [x] Update IRC object with new configuration
- [x] Look if `irc.user` variables need update (e.g. nick while disconnected)
- [x] Default nick to config default instead of hardcoded string

### DIFF
--- a/client/js/contextMenuFactory.js
+++ b/client/js/contextMenuFactory.js
@@ -153,6 +153,21 @@ function addFocusItem() {
 	});
 }
 
+function addEditNetworkItem() {
+	function edit(itemData) {
+		socket.emit("network:get", itemData);
+		$('button[data-target="#connect"]').trigger("click");
+	}
+
+	addContextMenuItem({
+		check: (target) => target.hasClass("lobby"),
+		className: "edit",
+		displayName: "Edit this network…",
+		data: (target) => target.closest(".network").data("uuid"),
+		callback: edit,
+	});
+}
+
 function addChannelListItem() {
 	function list(itemData) {
 		socket.emit("input", {
@@ -207,6 +222,7 @@ function addDefaultItems() {
 	addQueryItem();
 	addKickItem();
 	addFocusItem();
+	addEditNetworkItem();
 	addChannelListItem();
 	addBanListItem();
 	addJoinItem();
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -18,7 +18,6 @@ const utils = require("./utils");
 require("./webpush");
 require("./keybinds");
 require("./clipboard");
-const Changelog = require("./socket-events/changelog");
 const contextMenuFactory = require("./contextMenuFactory");
 const contextMenuContainer = $("#context-menu-container");
 
@@ -320,15 +319,6 @@ $(function() {
 		if (chan.data("needsNamesRefresh") === true) {
 			chan.data("needsNamesRefresh", false);
 			socket.emit("names", {target: self.data("id")});
-		}
-
-		if (target === "#settings") {
-			$("#session-list").html("<p>Loading…</p>");
-			socket.emit("sessions:get");
-		}
-
-		if (target === "#help" || target === "#changelog") {
-			Changelog.requestIfNeeded();
 		}
 
 		// Pushes states to history web API when clicking elements with a data-target attribute.
--- a/src/helper.js
+++ b/src/helper.js
@@ -31,6 +31,7 @@ const Helper = {
 	getGitCommit,
 	ip2hex,
 	mergeConfig,
+	getDefaultNick,
 
 	password: {
 		hash: passwordHash,
@@ -189,6 +190,14 @@ function passwordHash(password) {
 
 function passwordCompare(password, expected) {
 	return bcrypt.compare(password, expected);
+}
+
+function getDefaultNick() {
+	if (!this.config.defaults.nick) {
+		return "thelounge";
+	}
+
+	return this.config.defaults.nick.replace(/%/g, () => Math.floor(Math.random() * 10));
 }
 
 function mergeConfig(oldConfig, newConfig) {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -417,6 +417,7 @@ kbd {
 	margin-right: 9px;
 }
 
+.context-menu-edit::before,
 #set-nick::before {
 	content: "\f303"; /* https://fontawesome.com/icons/pencil-alt?style=solid */
 }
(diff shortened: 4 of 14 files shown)


astorije:
Wow, this is dope!

I haven't looked at the code yet, but to quickly discuss the overall behavior, this is why I would expect:

- If I change host/port/etc., it reconnects
- If I change the network name, it changes the name in the channel list UI (but doesn't reconnect if there is no change in host/port/etc.)
- If I change the nick, it sends a `/nick ...` (unless a full reconnect is necessary). This last bit is cool because we can then completely remove the nick editor UI from the message input!

UI can be made in a way that this behavior specificity is totally free of surprises (change the button to "Reconnect" instead of "Save" when reconnection will appear, extra info message above the save button, etc. several patterns to play with, and I'm happy to help wherever I can)

What are your thoughts on this?

xPaw:
@astorije All the points you point out are indeed achievable. However I want to test out current implementation for unexpected surprises and bugs (because a lot of code was moved around) before adding more code.

astorije:
That's fair. What is the expected behavior when hitting the submit button? i.e. what is it supposed to do and not to do?

xPaw:
@astorije It should update in-memory and on-disk values. Keep in mind that this PR also moves the code around for just connect, so both should work as expected.

I'll have another commit soon that implements `/nick` if it's changed, so only changing lobby name and reconnecting will be left to do.

MaxLeiter:
On mobile so can’t edit, #584 *

xPaw:
@MaxLeiter I think there's hardly any use case where you'd know which commands to enter when creating a network. And it's hidding a bit of clutter as to not confuse users (e.g. on public instances).

xPaw:
I've got all the UI stuff fleshed out. The only point that I didn't implement is auto reconnecting.

client/js/lounge.js
```diff
@@ -141,6 +141,12 @@ $(function() {
 					text: "Join a channel…",
 					data: target.data("id"),
 				});
+				output += templates.contextmenu_item({
+					class: "edit",
+					action: "edit",
+					text: "Edit network",
+					data: target.closest(".network").data("uuid"),
+				});
```
astorije: Mind putting this right after the network name, and calling it "Edit this network..."?

astorije:
Do you think you could add a few tests to this? That would be pretty awesome.

xPaw:
Added some tests.

McInkay:
This will need to be refactored now that #1878 is merged, and use the new structure for adding context menu items.

astorije:
@xPaw, what do you mean by:

> Default nick to config default instead of hardcoded string

xPaw:
@astorije default nick fallback, what we discussed in #2297

src/helper.js
```diff
@@ -176,6 +177,14 @@ function passwordCompare(password, expected) {
 	return bcrypt.compare(password, expected);
 }
 
+function getDefaultNick() {
+	if (!this.config.defaults.nick) {
+		return "thelounge";
+	}
```
astorije: Is this default not supposed to come from our own default copy of the config file?

client/js/contextMenuFactory.js
```diff
@@ -153,6 +153,22 @@ function addFocusItem() {
 	});
 }
 
+function addEditNetworkItem() {
+	function edit(itemData) {
+		socket.emit("network:get", itemData);
+		$("#connect").text("Loading");
```
astorije: Is this loading text legacy from a previous attempt or something? I couldn't manage to trigger it, even when emulating slow network and even when simulating lost connection.

However, I do notice a flash when clicking on "Edit this network", for a very short window (100ms?) the fields are filled with the connect defaults, then change to the actual values. I'm wondering if we can keep them empty until they're filled.

src/helper.js
xPaw: This is a fail-safe if user overrides the default with empty string.

astorije:
@xPaw, sorry about that but I just noticed while playing a bit more with your branch: going to the edit form appends the `#connect` hash to the URL, so reloading the page and browser history don't work, mind looking into that?

Also, I see you amended your commits, did you manage to do something about the "flash" when going to edit form? I just tried and I still see it, can't see the updated `Loading...`.

Also, not asking if that's on purpose, but that PR has evolved since we first talked about it so double-checking: if I change the host/port, I don't get re-connected, this is intended behavior for this PR, correct?

Feel free to assign me to the PR when you're done with changes!

xPaw:
> sorry about that but I just noticed while playing a bit more with your branch: going to the edit form appends the #connect hash to the URL, so reloading the page and browser history don't work, mind looking into that?

> Also, I see you amended your commits, did you manage to do something about the "flash" when going to edit form?

Both the flash and `#connect` are related issues because I'm re-using the connect window. It's a bit painful to create a separate window for editing networks that would work with page reloads and wouldn't involve a bunch of nasty code. So I just chose to keep it this way for now.

> Also, not asking if that's on purpose, but that PR has evolved since we first talked about it so double-checking: if I change the host/port, I don't get re-connected, this is intended behavior for this PR, correct?

Yep intended.

astorije:
> It's a bit painful to create a separate window for editing networks that would work with page reloads and wouldn't involve a bunch of nasty code. So I just chose to keep it this way for now.

I understand and that's fine for now I think 👍 (though maybe create an issue and tag it to v3 to keep track of this?)

Is there any chance you could fix the URL though, at least? It doesn't have to be the most perfect code if in the long-run we're having a separate window or something. That would fix the back/forward buttons, I believe.

MaxLeiter:
xPaw: could you assign a different hash when the window is loaded, and add s temporary check for it on load? 

On second thought, don’t really think it’s worth the trouble

xPaw:
> Is there any chance you could fix the URL though, at least?

That directly relates to window management sadly. Trying to rewrite the url will effectively mean I will have to manage the page reloading/popstate to keep track of it, which @MaxLeiter suggested.

PR #2230, opened 2018-03-15
Wrap join-form with lobby info

### LINKED ISSUE #2218: Join channel UI incorrectly located by moving channels
```
12:53 <+xPaw>  it should be inside of the lobby chan to not cause issues
12:54 <+xPaw>  you'll need to wrap current lobby things into a div, and move join form next to it
12:54 <+xPaw>  and then set a couple of flex css
```
I'll get to this in about 8 hours if no one else does

ESWAT:
Fixes #2218.

However I noticed a weird bug with both Chrome and Firefox where moving the `.join-form` in a container with flexbox set its form font to bold… so I explicitly reset `font-weight: normal` there.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -585,6 +585,13 @@ kbd {
 	font-size: 15px;
 	font-weight: bold;
 	padding-left: 0;
+	flex-direction: column;
+}
+
+#sidebar .lobby-row {
+	display: flex;
+	flex-direction: row;
+	width: 100%;
 }
 
 #sidebar .lobby-wrap {
@@ -724,7 +731,7 @@ kbd {
 	transform: rotate(45deg) translateZ(0);
 }
 
-#sidebar .network .lobby:nth-last-child(2) .collapse-network {
+#sidebar .network .lobby:last-child .collapse-network {
 	/* Hide collapse button if there are no channels/queries */
 	width: 0;
 	overflow: hidden;
@@ -1033,7 +1040,7 @@ kbd {
   */
 #sidebar .join-form {
 	display: none;
-	padding: 0 18px 8px;
+	padding: 8px 0 0 14px;
 }
 
 #sidebar .join-form .input {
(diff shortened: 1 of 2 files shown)


MaxLeiter:
Hm, whenever I try and type my input gets sent to the message box:

https://sr.ht/eEQ3.webm

xPaw:
@MaxLeiter Hm yeah. Clicking inside the form detects the click on `.chan` so it opens the channel, which in turn focuses the input.

McInkay:
That's a beautiful bug. @xPaw will this be fixed with #2215 ?

xPaw:
@McInkay No. The bug here will be annoying because it will switch the channel if your current channel is not a lobby.

McInkay:
Ah right, we need to change the click handler for moving to the lobby then.

xPaw:
I'm wondering why this only affects input elements though. This doesn't happen on `+` or collapse buttons.

McInkay:
They return false, don't they? So they prevent the propagation.

ESWAT:
What’s the next action on this? Should I be looking into why the click handler is still focusing on the input or is this going to be fixed somewhere else?

xPaw:
@ESWAT Probably has to be looked into in this PR. The bug here is that it detects a click event (which in turn focuses).

ESWAT:
FYI won’t have time to look into the remaining issues in this PR anytime soon.

astorije:
Thanks for the update, @ESWAT, and no worries, we get busy schedules :)

@MaxLeiter or @xPaw, do you think you can give a hand on this?

Actually, how do I reproduce the bug? Does it happen when drag-and-dropping networks/channels? I can't see the bug on master and v3.0.0-pre.5 so I'm thinking I'm not testing the correct thing.

MaxLeiter:
@astorije: Drag a channel in a network to a new spot and click the "add channel" button for that same network

astorije:
Ahah, gotcha, woops. I can confirm I see that on pre.5 now, thanks for the precision!

williamboman:
This seems to break `/whois` behaviour.

Issue #2233, opened 2018-03-15
Scroll jumping in browsers that do not support scrollIntoViewOptions


williamboman:
I'm running latest master + 2207, 2206, 2195, 2114

xPaw:
That's an extensive list of PRs. Can you please narrow it down where the problem happens?

williamboman:
Ah, yeah it scrolls to the currently active channel, on page load as well as when joining a channel. I can debug this more later on if you want and find the offending commit

![s](https://user-images.githubusercontent.com/6705160/37452842-0deacf72-2836-11e8-849b-0c6b57522cd1.gif)

xPaw:
What browser are you using?

cc @dgw

williamboman:
I primarily use TL on iOS Safari so I noticed it there first. Screen recording is macOS Safari. Can't repro in Chrome.

dgw:
This report was the last straw. Safari and Edge are just too flaky, and I don't want to do browser-detection hijinks. Revert proposed.

kode54:
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

Neither IE/Edge nor Safari support the entire object method of passing parameters. They only support scroll the object to the top of the view, period. Lovely.

williamboman:
<img src="https://user-images.githubusercontent.com/6705160/37552890-c31273b8-29bd-11e8-8be2-c48cd737b5e6.jpeg" width="250">

@kode54 🤷🏼‍♂️

dgw:
@kode54 Indeed. Apparently Safari, at least, also has awful fallback behavior when a page tries to use arguments that it doesn't understand. One would think ignoring the function call and printing an error to the developer console the logical fallback when passed arguments it doesn't understand. But no, it has to treat the object as `true` and scroll the element to the top of the container. (FWIW `false` will scroll the element to the bottom of the container. But an object is never falsy.) \</rant>

I'm going to test a couple polyfills, and/or try writing a generic scroll function that will be good enough for the user and channel lists (which are the parts of TL that need this). Next week-ish.

kode54:
I love this question and actual useful reply, from 2012:

https://stackoverflow.com/questions/9445842/does-scrollintoview-work-in-all-browsers

Supported, yes. User experience, bad. Has it changed much since then? Nope.

xPaw:
Related issue: #2193

xPaw:
I believe this was fixed.

[...]

PR #2234, opened 2018-03-15
Revert "Scroll to joined/activated channel"

dgw:
Reverts #2166

Since merging, this has caused too many issues due to inconsistent browser support (#2233, for example).

I'm working on a more brute-force approach that doesn't use this flaky API. Until that's done, I'd argue that removing the functionality is better than the current state.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -373,7 +373,6 @@ $(function() {
 				utils.toggleNotificationMarkers(false);
 			}
 
-			utils.scrollIntoViewNicely(self[0]);
 			slideoutMenu.toggle(false);
 		}
 
--- a/client/js/userlist.js
+++ b/client/js/userlist.js
@@ -5,7 +5,6 @@ const fuzzy = require("fuzzy");
 const Mousetrap = require("mousetrap");
 
 const templates = require("../views");
-const utils = require("./utils");
 
 const chat = $("#chat");
 
@@ -94,7 +93,7 @@ exports.handleKeybinds = function(input) {
 		}
 
 		// Adjust scroll when active item is outside of the visible area
-		utils.scrollIntoViewNicely(userlist.find(".user.active")[0]);
+		userlist.find(".user.active")[0].scrollIntoView(false);
 	});
 
 	// When pressing Enter, open the context menu (emit a click) on the active
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -15,7 +15,6 @@ module.exports = {
 	lastMessageId,
 	confirmExit,
 	forceFocus,
-	scrollIntoViewNicely,
 	hasRoleInChannel,
 	move,
 	resetHeight,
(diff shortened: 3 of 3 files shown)


astorije:
That's too bad :(

I don't know if that's any useful to you, but [this is what I originally did for the userlist scrolling-into-view](https://github.com/thelounge/thelounge/pull/1842/commits/ee7272305af0a6f41ccaa86f050896e6d40fb49c#diff-0286bab841c266a97e65f3863daf8488R86).

astorije:
Alternatively we could try one of the scrollIntoView polyfills?

xPaw:
I don't see a reason to revert this, we can just fix the function.

As for a polyfill, something like this? https://github.com/stipsan/scroll-into-view-if-needed

dgw:
@astorije Your code didn't get me very far the last time I looked at it, because the user list is all one big container but channels are nested in multiple parent elements relative to the container we care about scrolling, so it's trickier to do the same thing in a generic way. I might look at it again next week, or I might not, depends on how we get on with polyfills. That was actually my first idea, but I wasn't too keen on adding Yet Another Dependency if it could be done relatively cheaply in a bespoke utility function.

Basically, I see this revert probably isn't going to happen. I will come back to this, but likely not until Monday at the earliest. My weekend is very, very busy in the offline world, so I won't be coding much. :sweat_smile:

McInkay:
We don't need to rush @dgw , this code is not causing a massive issue and is not in a release, so it's not like it needs immediately fixing.

astorije:
> As for a polyfill, something like this? stipsan/scroll-into-view-if-needed

The risk with polyfills that apply if needed is that Safari _kinda_ supports it but not really, so I don't know if polyfills detect that scroll into view is available and just stop there.

https://github.com/iamdustan/smoothscroll seems to have a `scrollIntoView` polyfill, and a way to force it... Could this be a potential candidate?

xPaw:
> The risk with polyfills that apply if needed is that Safari kinda supports it but not really

Polyfill wasn't the right word to use it, as that library has a separate implementation and function export. It doesn't try to use the `Element.scrollIntoView` function. So we should be safe there as far as cross browser support is concerned.

astorije:
Aaah, I see, it's not "if needed" as in "this polyfill applies when native support is not available", it's "if needed" as in "scroll the element if it's not already in view", gotcha.

Yeah, trying with the project you suggested would be nice.

@dgw, I'm going to close this PR because we all agreed we prefer to try an alternative, the issue is bearable since we're not releasing stable right away (and it's already in a pre-release), there are conflicts in this PR now and I'm sure you don't want to deal with conflicts on an automated revert commit/PR :D

dgw:
That's fine, @astorije, you're right of course about the conflicts :sweat_smile:

Still planning to try and address this next week. My marathon weekend is nearly over :tada:

PR #2235, opened 2018-03-15
Store visibility status of sidebar menus

### LINKED ISSUE #51: Store visibility status of sidebar menus in settings object

xPaw:
Fixes #51
Closes #1606

Also fixes sidebar reopening on desktop when switching channels with keyboard binds.

### DIFF
--- a/client/js/socket-events/init.js
+++ b/client/js/socket-events/init.js
@@ -42,11 +42,18 @@ socket.on("init", function(data) {
 
 		webpush.configurePushNotifications(data.pushSubscription, data.applicationServerKey);
 
+		slideoutMenu.enable();
+
+		const viewport = $("#viewport");
+
+		if ($(window).outerWidth() >= utils.mobileViewportPixels) {
+			slideoutMenu.toggle(storage.get("thelounge.state.sidebar") === "true");
+			viewport.toggleClass("rt", storage.get("thelounge.state.userlist") === "true");
+		}
+
 		$(document.body).removeClass("signed-out");
 		$("#loading").remove();
 		$("#sign-in").remove();
-
-		slideoutMenu.enable();
 
 		if (window.g_LoungeErrorHandler) {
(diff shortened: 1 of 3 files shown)


client/js/socket-events/init.js
```diff
@@ -42,12 +42,19 @@ socket.on("init", function(data) {
 
 		webpush.configurePushNotifications(data.pushSubscription, data.applicationServerKey);
 
+		slideoutMenu.enable();
+
+		const viewport = $("#viewport");
+
+		if (viewport.outerWidth() >= 768) {
+			slideoutMenu.toggle(storage.get("thelounge.state.sidebar") === "true");
```
McInkay: Do we need the `=== true`? Surely it'll work fine without?

xPaw: Note the string. false as a string still evaluates to true.

McInkay: Oh stupid.

The purest in me feels like saying use Json.parse then ...

astorije:
Sweet! I noticed a couple bugs though:

- Toggling the user list actually changes the state of the channel list:

  sidebar-bug

- On mobile resolution (I tried both mobile-like in Chrome, and on actual phone), the state is not kept on page refresh. This applies to both channel list and user list, but I only recorded user list:

  sidebar-bug2

xPaw:
> Toggling the user list actually changes the state of the channel list

Brilliant. I'll take a look. This also affects master.

> On mobile resolution (I tried both mobile-like in Chrome, and on actual phone), the state is not kept on page refresh.

This is by design (Anything under 768px is not stored) as I didn't want to store overlays needlessly.

astorije:
> This is by design (Anything under 768px is not stored) as I didn't want to store overlays needlessly.

Aaah, totally missed that. That's fair, and a good starting point.

PR #2236, opened 2018-03-15
Rework how unread marker is moved when status messages are hidden

### LINKED ISSUE #2214: Unread marker shows up when status messages are hidden
Noticed the marker showing up with nothing below it after someone had joined a channel.

@xPaw pointed at https://github.com/thelounge/thelounge/blob/0260bcbf3b55f561834f979046b15a2a561c3f27/client/js/socket-events/msg.js#L76

Happy to look into it soon.

xPaw:
Fixes #2214.

This should be more performant as it only gets last children instead of filtering by `:visible`. We basically no longer need to iterate as status messages are contained within a single `condensed` container.

### DIFF
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -70,12 +70,25 @@ function processReceivedMessage(data) {
 
 	notifyMessage(targetId, channel, data);
 
-	const lastVisible = container.find("div:visible").last();
+	let shouldMoveMarker = data.msg.self;
 
-	if (data.msg.self
-		|| lastVisible.hasClass("unread-marker")
-		|| (lastVisible.hasClass("date-marker")
-		&& lastVisible.prev().hasClass("unread-marker"))) {
+	if (!shouldMoveMarker) {
+		const lastChild = container.children().last();
+
+		// If last element is hidden (e.g. hidden status messages) check the element before it.
+		// If it's unread marker or date marker, then move unread marker to the bottom
+		// so that it isn't displayed as the last element in chat.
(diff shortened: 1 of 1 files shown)


client/js/socket-events/msg.js
```diff
@@ -70,12 +70,25 @@ function processReceivedMessage(data) {
...
-		|| (lastVisible.hasClass("date-marker")
-		&& lastVisible.prev().hasClass("unread-marker"))) {
+	if (!shouldMoveMarker) {
+		const lastChild = container.children().last();
+
+		// If last element is hidden (e.g. hidden status messages) check the element before it.
+		// If it's unread marker or date marker, then move unread marker to the bottom
+		// so that it isn't displayed as the last element in chat.
+		// display properly is checked instead of using `:hidden` selector because it doesn't work in non-active channels.
+		if (lastChild.css("display") === "none") {
```
astorije: I'm testing this and it looks fine, but will this be buggy when changing the settings for condensed messages, until the next app reload?

xPaw: That always was the case. New message or switching channel would fix it.

astorije: I didn't realize that was the case on `master`. Workarounds that involve reloading the app or switching channels are unfortunate, but since it's not a breakage you introduced here, that's fine to ignore for now.

astorije:
Thanks @xPaw for the fix, and thanks @realies for helping review this!

realies:
Noticed this back on several occasions.

PR #2246, opened 2018-03-17
Add a semi-opaque overlay when channel list is open on mobile

astorije:
This also fixes a bug that would let user accidentally open a link when trying to close the channel list, [this](https://github.com/thelounge/thelounge/pull/2185/files#diff-97db1f70168fb5f12457b238ff6052b5L2326).

sidebar-overlay

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -2274,7 +2274,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		position: absolute;
 		left: -220px;
 		z-index: 2;
-		transition: transform 160ms, box-shadow 0.2s;
+		transition: transform 160ms, box-shadow 160ms;
 		transform: translateZ(0);
 	}
 
@@ -2286,8 +2286,8 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		right: 0;
 		background: rgba(0, 0, 0, 0.5);
 		opacity: 0;
-		display: none;
-		transition: opacity 160ms;
+		visibility: hidden;
+		transition: opacity 160ms, visibility 160ms;
 		z-index: 1;
 	}
 
@@ -2302,7 +2302,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 	#viewport.menu-dragging #sidebar-overlay,
 	#viewport.menu-dragging #sidebar {
-		transition: none !important;
+		transition: box-shadow 160ms;
 	}
 
 	#viewport.menu-open #sidebar,
@@ -2312,11 +2312,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 	#viewport.menu-open #sidebar-overlay,
 	#viewport.menu-dragging #sidebar-overlay {
-		display: block;
-	}
-
-	#viewport.menu-open .messages {
-		pointer-events: none;
+		visibility: visible;
 	}
 
 	#sidebar .empty::before {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -2273,9 +2273,26 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		height: 100%;
 		position: absolute;
 		left: -220px;
-		z-index: 1;
-		transition: transform 160ms;
+		z-index: 2;
+		transition: transform 160ms, box-shadow 0.2s;
 		transform: translateZ(0);
+	}
+
+	#sidebar-overlay {
+		position: fixed;
+		top: 0;
+		bottom: 0;
+		left: 0;
+		right: 0;
+		background: rgba(0, 0, 0, 0.5);
+		pointer-events: none;
+		opacity: 0;
+		transition: opacity 160ms;
+		z-index: 1;
+	}
+
+	#viewport.menu-open #sidebar-overlay {
+		opacity: 1;
 	}
 
 	#viewport.menu-open #sidebar {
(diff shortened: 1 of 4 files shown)


client/css/style.css
```diff
@@ -2273,21 +2273,44 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		height: 100%;
 		position: absolute;
 		left: -220px;
-		z-index: 1;
-		transition: transform 160ms;
+		z-index: 2;
+		transition: transform 160ms, box-shadow 0.2s;
```
xPaw: My understanding of transitioning `box-shadow` is that it could affect performance negatively (or look terrible on slow devices) and more common way is animating `opacity`.

xPaw:
I updated this PR to do the following:

- Screen overlay now blocks clicking through to disallow scrolling chat while menu is open
- Slideout can now be trigger anywhere on the screen, not just the edge. Any horizontal movement cancels it.
- `touchmove` event is now also passive.

client/js/slideout.js
```diff
@@ -49,13 +50,18 @@ function onTouchMove(e) {
 	const touch = touchCurPos = e.touches.item(0);
 	let setX = touch.screenX - touchStartPos.screenX;
 
-	if (Math.abs(setX > 30)) {
```
xPaw: We had a bug here, that `> 30` check should have been after `)`.

astorije:
I've noticed a transition issue, I'll look into it later today.
Also, quick test showed that scrolling still goes through the sidebar when it's open, @xPaw can you confirm you didn't forget to push something or whatnot? Just checking before I look into it.

astorije:
It's pretty difficult to open or close the menu right now because the gesture has to be _exactly_ horizontal, which I miss 4 times out of 5 😅
Will look if we can make this a bit less strict easily or not. Mmh, maybe we should have kept this in a different PR :/

client/js/lounge.js
```diff
@@ -40,13 +40,17 @@ $(function() {
...
+		slideoutMenu.toggle(false);
+		storeSidebarVisibility("thelounge.state.sidebar", false);
+	});
 
-		if ((isOpen && $(window).outerWidth() < utils.mobileViewportPixels) || $(e.target).is(".lt")) {
-			slideoutMenu.toggle(!isOpen);
-			storeSidebarVisibility("thelounge.state.sidebar", !isOpen);
-		}
+	$("#windows").on("click", "button.lt", () => {
+		const isOpen = !slideoutMenu.isOpen();
```
astorije: @xPaw, if we can click on this button, doesn't it mean that it's closed anyway?

astorije:
Alright, I've fixed the couple issues with transitions.

I've also changed the behavior a little bit to open the menu: swipes with an angle of 0-45° open the menu; swipes with an angle of 45-90° vertically scroll. I was having such a hard time to open the menu!
Changing that 45° value is trivial btw, so we could reduce/increase to make one or the other easier, I just figured splitting equally would be a good start, and we can refine later if we notice it's not convenient enough.

It works great on Android, however on iOS the chat scroll still goes through the sidebar, the overlay, and even while dragging the menu open (even though the code clearly states that once the menu is being dragged, no chat scroll should occur). I cannot figure that 💩 out. I have spent waaaay too much time searching online, with issues dating back from 2012 / iOS 5, all sorts of solutions that work for some and doesn't work for others. Nothing I could try made a difference.
If anyone wants to give it a shot, I would be thrilled, but I am giving up for now, this is madness. This PR already stones too many birds already (issue is present on `master` as well).

If we forget about that iOS nonsense, this PR should be ready now.

client/css/style.css
```diff
@@ -2286,8 +2286,8 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		right: 0;
 		background: rgba(0, 0, 0, 0.5);
 		opacity: 0;
-		display: none;
-		transition: opacity 160ms;
+		visibility: hidden;
+		transition: opacity 160ms, visibility 160ms;
```
astorije: Have to use `visibility` to correctly animate the overlay, as transition doesn't apply on `display` and that was causing a flash.

client/css/style.css
```diff
@@ -2302,7 +2302,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 	#viewport.menu-dragging #sidebar-overlay,
 	#viewport.menu-dragging #sidebar {
-		transition: none !important;
+		transition: box-shadow 160ms;
```
astorije: It turns out `!important` was not necessary anymore, but `box-shadow` transition must be kept or it makes it blink.

client/css/style.css
```diff
@@ -2312,11 +2312,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 	#viewport.menu-open #sidebar-overlay,
 	#viewport.menu-dragging #sidebar-overlay {
-		display: block;
-	}
-
-	#viewport.menu-open .messages {
-		pointer-events: none;
```
astorije: This is not necessary anymore since overlay eats all clicks. Thankfully even on iOS, only scroll events go through, but clicks are eaten just fine.

client/js/lounge.js
xPaw: You can see both on desktop.

astorije: Duh 😴

McInkay:
Looks pretty good. Found just 1 issue. It might be a bit complicated to explain. Repro steps

1. Open lounge on mobile device
2. Open sidebar
3. Go into a server lobby
4. Open sidebar again
5. Collapse or expand network

What should happen: network is collapsed or expanded and nothing else

What does happen: network is collapsed or expanded and the sidebar is closed.

I think this is probably because it's selecting the lobby when clicking the caret of a lobby that is already selected. Or ... something ...

PR #2247, opened 2018-03-17
Print failed login attempts to console

xPaw:

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -624,6 +624,12 @@ function performAuthentication(data) {
 	const authCallback = (success) => {
 		// Authorization failed
 		if (!success) {
+			if (!client) {
+				log.warn(`Authentication for non existing user attempted from ${colors.bold(getClientIp(socket))}`);
+			} else {
+				log.warn(`Authentication failed for user ${colors.bold(data.user)} from ${colors.bold(getClientIp(socket))}`);
+			}
+
 			socket.emit("auth", {success: false});
 			return;
 		}


McGriddle:
@astorije It's not a solution for my specific problem as I'm running fail2ban on a different host than thelounge.

I believe with this change, one could redirect stdout to a file, and reference that in fail2ban on the local system. The question of log rotation comes up, but I'm getting ahead of msyelf.

dgw:
Running The Lounge as an Upstart service, I get log files generated. They might include only stderr (I can't log in and check, the network I'm on allows **only** ports 80 and 443), but I know I can get stdout logged by, at most, tweaking the service unit file. Upstart rotates logs for me, as well.

Basically, if you run The Lounge as a service using \<insert init system here>, it probably has this facility built in if you add a line or two to the service configuration.

astorije:
@xPaw, should this go to stderr instead?
I don't have experience of how other tools usually do this, but I feel odd having our "regular" logs (server started, user/package loaded, etc.) mixed with login errors and any other thing that fail2ban or such could listen to.

xPaw:
@astorije errors and warnings already go to `console.error`.

astorije:
Yeah, I had forgotten about that.

:shipit:

McGriddle:
@astorije This would work with fail2ban just fine. Following the manual as far as default locations https://www.fail2ban.org/wiki/index.php/MANUAL_0_8 ...

Would want to create a new filter like `/etc/fail2ban/filter.d/thelounge-badauth.conf`
```
[Definition]
  
failregex = Authentication failed for user .* from <HOST>$
            Authentication for non existing user attempted from <HOST>$

ignoreregex =
```

Then include these lines at the bottom of your configuration file like `/etc/fail2ban/jail.local`
```
[thelounge-badauth]
enabled  = true
port     = http,https
filter   = thelounge-badauth
logpath  = /your/log/file/location
maxretry = 10
findtime = 600
bantime = 600
```
So within a timespan of 600 seconds, if the regex matches in the log 10 times, then we temporarily ban the IP for 600 seconds.

fail2ban-regex

astorije:
Hey @McGriddle, this looks like a perfect fit for [our community wiki](https://github.com/thelounge/thelounge/wiki), do you want to add a new page there? :)

lzrdking71:
@McGriddle, @xPaw  @astorije. Where is the above mentioned log file located and what is it named? If I need to create it, how? I am not having much luck trying to figure out the logpath so that I can get fail2ban working with thelounge.

PR #2251, opened 2018-03-19
Render video and audio previews only after canplay event fires

xPaw:
This fixes a couple of things:

- Page jump when video dimensions change (because we set them via `max-width` attribute)
- Incorrect preview when browser fails to preload the metadata or simply can't play the media type

### DIFF
--- a/client/js/renderPreview.js
+++ b/client/js/renderPreview.js
@@ -17,15 +17,17 @@ function renderPreview(preview, msg) {
 	preview.shown = preview.shown && options.shouldOpenMessagePreview(preview.type);
 
 	const template = $(templates.msg_preview({preview}));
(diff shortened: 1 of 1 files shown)


client/js/renderPreview.js
```diff
@@ -17,15 +17,17 @@ function renderPreview(preview, msg) {
...
 
 	const template = $(templates.msg_preview({preview}));
-	const image = template.find("img:first");
+	const image = template.find("img, video, audio").first();
 
 	if (image.length === 0) {
 		return appendPreview(preview, msg, template);
 	}
 
+	const loadEvent = image.prop("tagName") === "IMG" ? "load" : "canplay";
```
astorije: Heh

PR #2254, opened 2018-03-20
Empty local storage on sign out and move the sign out button to the settings

astorije:
This change improves privacy/security by ensuring all local storage (which includes settings, etc.) is destroyed on sign out or when revoking a remote session. Because signing out is now more "risky", the button has been moved to the settings along with other existing sessions.

This PR:

- Removes the power/sign-out icon from the sidebar footer (gives additional room for when the admin panel gets added, WIP #1950)
- Adds a "Sign out" button next to the current session info in the settings session list
- Renames "Disconnect" buttons into "Revoke" to better clarify the intent (I will admit that I borrowed the wording from Telegram)
- Fixes incorrect `localStorage.remove` method
- Uses Sinon.js to mock wrappers for `window.localStorage` and `window.location` (does not mock those themselves, in the "Do not mock what you do not own" fashion, mock our layer instead). I expect we will be able to test more from the UI with this. ~A good next candidate will be the `mockLogger` things.~ EDIT: Nevermind, I did it in there.

screen shot 2018-03-20 at 00 53 18

### DIFF
--- a/client/js/localStorage.js
+++ b/client/js/localStorage.js
@@ -13,7 +13,10 @@ module.exports = {
 	get(key) {
 		return window.localStorage.getItem(key);
 	},
-	remove(key, value) {
-		window.localStorage.removeItem(key, value);
+	remove(key) {
+		window.localStorage.removeItem(key);
+	},
+	clear() {
+		window.localStorage.clear();
 	},
 };
--- a/client/js/location.js
+++ b/client/js/location.js
@@ -1,0 +1,9 @@
+"use strict";
+
+// This is a thin wrapper around `window.location`, in order to contain the
+// side-effects. Do not add logic to it as it cannot be tested, only mocked.
+module.exports = {
+	reload() {
+		window.location.reload();
+	},
+};
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -489,15 +489,6 @@ $(function() {
 	$("#help").on("click", "#view-changelog, #back-to-help", openWindow);
 	$("#changelog").on("click", "#back-to-help", openWindow);
 
-	sidebar.on("click", "#sign-out", function() {
-		socket.emit("sign-out");
-		storage.remove("token");
-
-		if (!socket.connected) {
-			location.reload();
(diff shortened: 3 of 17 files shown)


client/js/localStorage.js
```diff
@@ -13,7 +13,10 @@ module.exports = {
 	get(key) {
 		return window.localStorage.getItem(key);
 	},
-	remove(key, value) {
-		window.localStorage.removeItem(key, value);
+	remove(key) {
+		window.localStorage.removeItem(key);
```
astorije: See https://github.com/thelounge/thelounge/commit/326f1ac47633c71452752f53850a18630411eadb#diff-bb2f08eb9be2feab0f4bac6115eeaa8eR16, `remove|removeItem(key)` became `remove|removeItem(key, value)`.

client/js/lounge.js
```diff
@@ -489,15 +489,6 @@ $(function() {
...
 
-	sidebar.on("click", "#sign-out", function() {
-		socket.emit("sign-out");
-		storage.remove("token");
-
-		if (!socket.connected) {
-			location.reload();
-		}
-	});
-
```
astorije: I did try to take as much as I could from `lounge.js` 🤷‍♂️
As far as I know, this is the only logic about this in there though.

client/js/location.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+// This is a thin wrapper around `window.location`, in order to contain the
+// side-effects. Do not add logic to it as it cannot be tested, only mocked.
+module.exports = {
+	reload() {
+		window.location.reload();
```
xPaw: Do you want to update all `location.reload` references in Lounge to use this wrapper?

I noticed some of them also use `reload(true)` to force a reload.

astorije: Yeah will do, but it turns out there is no other instance that can use this one. I'll add `forceReload` (`reload(true)`) in another PR because I am _trying_ to add some test around it, which is times harder that I expected...

PR #2258, opened 2018-03-20
Unfocus input on touch devices to close the keyboard

### LINKED ISSUE #2257: Mobile keyboard opens on channel switch
1. Have TL open on android and have the keyboard visible, e.g. by sending a message or just tapping into input box
2. Press your phone's back button to close the keyboard
3. switch to another channel

**Expected behavior:** Keyboard stays closed
**Actual behavior:** Keyboard opens

When in step 2, you tap into empty space in the channel instead, that will also hide the keyboard, but in this case, switching channels will not bring up the keyboard, so that's fine. 

I was told #2215 is responsible for this bug.

xPaw:
Fixes #2257.

This also brings back the behaviour of keyboard staying open when sending messages (was removed in #2215).

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -230,28 +230,20 @@ $(function() {
 			chat.find(".chan.active .chat").trigger("keepToBottom"); // fix growing
 		});
 
-	let focus = $.noop;
-
-	if (!("ontouchstart" in window || navigator.maxTouchPoints > 0)) {
-		focus = function() {
-			if (chat.find(".active").hasClass("chan")) {
-				input.trigger("focus");
-			}
-		};
-	}
-
 	if (navigator.platform.match(/(Mac|iPhone|iPod|iPad)/i)) {
 		$(document.body).addClass("is-apple");
 	}
 
-	$("#form").on("submit", function(e) {
-		e.preventDefault();
-		focus();
+	$("#form").on("submit", function() {
+		// Triggering click event opens the virtual keyboard on mobile
+		// This can only be called from another interactive event (e.g. button click)
+		input.trigger("click").trigger("focus");
+
 		const target = chat.data("id");
 		const text = input.val();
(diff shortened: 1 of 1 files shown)


astorije:
While testing this PR, 2 things I've noticed that differ from `pre.3`:

- On desktop (macOS, Chrome, but I'm pretty sure OS/browser agnostic), opening a channel does not focus the input anymore. If possible, I think we should keep this, mainly because typing unfocused and typing focused are not _exactly_ the same thing (I'm thinking of styling control keys for example). As far as I know, there is no downside to being focused (on desktop) when opening a channel, while there is that one to not being focused.

- On mobile (iOS, bloody Safari, pinned to homescreen, might be browser dependent?), when virtual keyboard is open, changing channel closes the keyboard. I actually prefer this when trying it, just making sure this is expected.

xPaw:
> On desktop (macOS, Chrome, but I'm pretty sure OS/browser agnostic), opening a channel does not focus the input anymore.

Odd. That would have happened if your device returned true for `"ontouchstart" in window`, could you test it?

> On mobile (iOS, bloody Safari, pinned to homescreen, might be browser dependent?), when virtual keyboard is open, changing channel closes the keyboard.

That's indeed intentional (I wanted to blur on touch devices and focus on the rest).

astorije:
> That would have happened if your device returned true for "ontouchstart" in window, could you test it?

```js
> "ontouchstart" in window
false
```

In fact, if input is focused and I change channel, input keeps its focused. If input is **not** focused and I change channel, input **does not** get the focus. The former is expected, not the latter.

I can reproduce this behavior both on Chrome and Firefox, but it works as expected on Safari (never thought I'd say that one day). All 3 answer `false` to the code above.

xPaw:
@astorije Can you test again? I think I fixed it.

PR #2259, opened 2018-03-20
Scroll channel list only when using keybinds

xPaw:
It doesn't make much sense to scroll the channel list when clicking on channels.


astorije:
Rebased on `master` to fix conflicts.

PR #2262, opened 2018-03-20
stop propagation, prevent input history call when alt is involved

### LINKED ISSUE #2261: Channel switch shortcut propagates to input history shortcut
Noticed that doing Alt+Up/Down changes the channel focus and the input message text as if Alt was not pressed. Inserting `return false` after this statement does not seem to fix it. Ideas?

https://github.com/thelounge/thelounge/blob/fdd321311ed19d1c39728973f4b07231ad3200f7/client/js/keybinds.js#L57

realies:
fixes #2261

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -55,6 +55,8 @@ Mousetrap.bind([
 	}
 
 	channels.eq(target).click();
+
+	return false;
 });
 
 Mousetrap.bind([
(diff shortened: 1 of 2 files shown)


xPaw:
This was missed when ctrl+up keybind was changed to alt+up.

astorije:
Hey @realies, we have sticker packs for our contributors now!

PR #2270, opened 2018-03-21
Replace pseudo element gradient fade with mask-image

xPaw:
This allows for much simpler theming, as it is no longer necessary to match colors everywhere a fade is used.

Also changed topic overflow from ellipsis to fade for consistency.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -912,7 +912,6 @@ kbd {
 	margin-left: 8px;
 	word-break: break-all;
 	flex-grow: 1;
-	white-space: nowrap;
 	overflow: hidden;
 	font-size: 14px;
 }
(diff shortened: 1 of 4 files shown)


client/css/style.css
```diff
@@ -2216,22 +2211,11 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 @media (min-width: 480px) {
 	/* Fade out for long usernames */
-
 	#chat .from {
 		padding-left: 10px;
+		-webkit-mask-image: linear-gradient(to left, transparent, black 10px);
+		mask-image: linear-gradient(to left, transparent, black 10px);
```
astorije: Mind making this `20px` before merging?

astorije:
Actually @xPaw, I told you on IRC that I'd take care of preview ellipsis because it conflicts with local changes, but it doesn't so feel free to add that to `#chat .toggle-content .head, #chat .toggle-content .body` as well :)

client/css/style.css
xPaw: As you noted in your other PR, increasing the distance here starts fading nicks that are not overflowing. I set it at 10px because that's the padding.

PR #2272, opened 2018-03-22
Improve UI of previews

astorije:
This is something I've had in my stash for a while, just never got to finish and commit it.

- Remove margins around thumbnails and images, increase radius, etc. for a fresher look overall. I give a bit more room to breathe, and that _almost_ does not reduce message density (see screenshots below)
- Replace ellipsis (`…`) when text is too long with a fading text mask, following #2270.
  - ⚠️ As you can see on the screenshot, there is a problem when nothing gets truncated, gradient is still shown. I'm not sure there is a solution to that, which would cause us to stick with `text-overflow: ellipsis` until `text-overflow: fade` is even remotely supported. @xPaw, any idea?
- Increase preview body contrast just enough to pass AA standards
- Add a `More`/`Less` button to expand previews that take more than one line. Button is added only when necessary
  - ⚠️ The way I decide to show or hide the button is questionable, and certainly does not handle resizes well. It does actually work okay as-is, but I'm happy to try out ideas.

Before | After
--- | ---
screen shot 2018-03-22 at 00 34 33 | screen shot 2018-03-22 at 01 19 05
screen shot 2018-03-22 at 01 26 56 | screen shot 2018-03-22 at 01 09 00

Collapsed | Expanded
--- | ---
screen shot 2018-03-22 at 01 02 36 | screen shot 2018-03-22 at 01 02 44


client/js/renderPreview.js
```diff
@@ -97,6 +106,13 @@ $("#chat").on("click", ".text .toggle-button", function() {
 	}
 });
 
+$("#chat").on("click", ".toggle-content .more", function(e) {
+	$(this).closest(".toggle-text-container").toggleClass("opened");
+
+	e.stopPropagation();
```
xPaw: `return false` already does this for you in jQuery.

xPaw:
Errors need tweaking:


---

It shows more button unnecessarily even though it fits:


Seems to only happen if description is shorter than title.

christer88:
I don't like the extra step to see the entire preview... Definently a big NO from me.

xPaw:
> I don't like the extra step to see the entire preview...

??? On master you can only view a single line of the description, this PR adds a feature to see more of it.

christer88:
Then I would please like that reverted, as it's very annoying having to click on a button to see it. Let me se it by default.

astorije:
Just to clarify if someone is concerned: there is nothing to revert once this PR is merged to bring back previous behavior as it is not changing logic or information displayed on the screen.
It displays just as much as `master`, `v3.0.0-pre.3`, and `v2.7.1` do (on IRC we noticed @christer88 must be running some custom CSS/theme), and the "Before" version has effectively looked as such for a very long time.

The "More" button however makes [this custom CSS snippet](https://github.com/thelounge/thelounge/wiki/CSS-Modifications#enlarge-images-in-link-previews) less relevant (since it allows for expanding the preview).

client/js/renderPreview.js
astorije: Yes, it's from the very first version I had, where I had to deal with horrific bubbling, forgot to remove.

astorije:
From #2270:

> As you noted in your other PR, increasing the distance here starts fading nicks that are not overflowing. I set it at 10px because that's the padding.

@xPaw, I have tried reducing the padding here as well (from 20 to 10px, but it's not going too well:

screen shot 2018-03-22 at 20 25 51

Even at 5px, it's clearly noticeable, that's not great.

Can you think of any workaround? I can't actually :/
The more I think about it, the more I feel like we should just keep `text-overflow: ellipsis` everywhere until `text-overflow: fade` is a thing...

xPaw:
@astorije Did you reduce it to `10px` in `mask-image`?

williamboman:
<details>
  <summary><p>
  Here's my suggestion on how to make the error message a bit nicer (expand for diff)
</p>
screenshot 2018-03-23 at 11 30 07


</summary><p>

```diff
commit d3626591b9a300c35d719729bf747ea99d69f845
Author: William Boman <william@redwill.se>
Date:   Fri Mar 23 11:28:40 2018 +0100

    views/msg_preview: change styling of error message

diff --git a/client/css/style.css b/client/css/style.css
index 9e479251..872bcfc5 100644
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1362,6 +1362,7 @@ kbd {
 
 #chat .toggle-type-error {
 	background: transparent;
+	padding: 5px;
 }
 
 #chat .toggle-content img {
diff --git a/client/views/msg_preview.tpl b/client/views/msg_preview.tpl
index 1e1cb916..b8653b55 100644
--- a/client/views/msg_preview.tpl
+++ b/client/views/msg_preview.tpl
@@ -30,20 +30,20 @@
 	{{/equal}}
 	{{#equal type "error"}}
 		{{#equal error "image-too-big"}}
-			<em>
+			<small>
 				This image is larger than {{friendlysize maxSize}} and cannot be
 				previewed.
 				<a href="{{link}}" target="_blank" rel="noopener">Click here</a>
 				to open it in a new window.
-			</em>
+			</small>
 		{{/equal}}
 		{{#equal error "message"}}
-			<em>
+			<small>
 				There was an error loading preview for this link.
 				<a href="{{link}}" target="_blank" rel="noopener">Click here</a>
 				to open it in a new window.
 				<small class="prefetch-error">({{message}})</small>
-			</em>
+			</small>
 		{{/equal}}
 	{{/equal}}
 </div>

```

</p></details>

astorije:
> @astorije Did you reduce it to `10px` in `mask-image`?

Yep, this is 10px in the screenshot above :(

Thanks @williamboman, will take a look.

xPaw:
Perhaps `mask-image` should be set on the container instead of head/body?

astorije:
I'll try that, but I'd be surprised if it works, because in the case of previews, container size is the same as the text size :/

astorije:
I made a few changes for the errors, and one thing leading to another, I improved things I did before:

- Errors are now displayed in a similar fashion, and error details can be viewed with the same pattern as expanding previews:

  Collapsed | Expanded | Mobile
  --- | --- | ---
  screen shot 2018-03-27 at 20 02 37 | screen shot 2018-03-27 at 20 02 50 | screen shot 2018-03-27 at 20 03 28

- Accessibility (`aria-expanded` + `aria-label`)

- I changed how the resize handler compute sizes now, which fixes changing orientation + handler was called twice each time :success:

- I'm having a real hard time with that `mask-image` because, as opposed to the other elements we're mask-imaging, this one has no fixed width, the container takes the width of the content. So we end up with a gradient that always shows (like the after picture in PR description).
  I resorted to the existing `ellipsis` for now. I feel like I have tried everything and there is currently no way to do this with CSS. I can either use JS (bank on that resize handler: we show both the "More" button and the gradient together, or none), or we'll have to hope for `text-overflow: fade;`.

xPaw:
I love the changes, but I can't bring myself to approve it due to the resize handling and triggering 😢 

It's just sadly slow:

astorije:
I think performance would not be an issue on resize because it's a rare thing to do, as I'd say 80% of my goal is orientation change and 20% is desktop window resize (and orientation change event isn't reliable enough, apparently).

That said, I totally understand your reluctance so I'm not at all closed to improve this. I have debounced the resize event with a 200ms threshold, that should solve most of it since resize event is usually a burst and not a continuous stream (lodash example is even about resize handler).
What do you think?

@xPaw, what did you use to profile this? I tried with Chrome and Firefox, but as far as I could find, the report was not nearly as good as yours.

xPaw:
I noticed that switching channels always shows `More` and then hides it, no matter how many times you switch channels.



Even with debouncing it feels slugglish. Sadly I don't think resize is the right way to go here at all, but I can't think of another solution.

astorije:
> I noticed that switching channels always shows More and then hides it, no matter how many times you switch channels.

Ah, good catch! Fixed it, it was caused by handling previews in inactive channels (with width of 0). It also improves performance, by ignoring what's not in the view.

> Even with debouncing it feels slugglish. Sadly I don't think resize is the right way to go here at all, but I can't think of another solution.

I don't see how it can feel more sluggish as there is no DOM update, repaint or whatnot while it's being debounced. Testing on a slower laptop, it feels just as sluggish on `master` on Chrome actually.

Re: resize, one way or another, this needs to be displayed/hidden when the element size changes, so I'm not sure why the handling resize is a bad idea. When Resize Observer API is more available, we can definitely update this code to use it.
But overall, it really is not a big deal! It's not a mouse over event or something that happens all the time. Resizing a window is not something you do every 10 minutes, so a few % decrease of perf on resize (which is mainly orientation change) is not going to be any noticeable _at all_.

client/js/utils.js
```diff
@@ -141,3 +142,8 @@ function requestIdleCallback(callback, timeout) {
 		callback();
 	}
 }
+
+// Force handling preview display
+function togglePreviewMoreButtonsIfNeeded() {
+	window.requestAnimationFrame(() => $(window).trigger("resize"));
```
xPaw: Would it at least work if you triggered `resize` on current channel container instead of window?

xPaw:
@McInkay what do you think?

client/js/utils.js
astorije: I'm not sure, but I'll try.

astorije:
Please don't make me close this...
Almost nothing happens when switching channels, and any potential perf issue would happen only on resize, not on mouseover/scroll or whatnot.

I can add all the checks that you want. I need to make sure it doesn't happen on collapsed previews, but I think it doesn't. Will double-check.

@xPaw, would you be a bit less concerned if I added support for Resize Observer if it's available (and fallback to this if not)?

client/js/utils.js
astorije: I went a different route, I don't trigger `resize` anymore, and I can now target specifically current channel + link previews only + expanded previews only.
Will add more details in a comment in the main thread.

astorije:
Alright, this is now the most optimized thing I have worked on for The Lounge, and the most optimized thing that happens when loading a channel :D

Improvements:

- `.on` instead of `.bind` (no consequence on perf, I just figured I'd mention it since I amended the first commit with that)
- Only runs the handler on expanded previews
- Only runs the handler on active channel
- When opening a side menu and/or a channel, the `resize` event is not fired anymore, instead target the previews that need to be computed specifically
- Only bind on link previews
- Handler now expects to be called on appropriate previews so it does no extra checks when opening channel

Extra considerations I gave:

- Putting aside compatibility issues, ResizeObserver API is actually not a good fit here, because 1) the callback gets called at least twice (as it modifies its own size), 2) it cannot be debounced as it would debounce also when expanding preview / opening side menus.
- I did not find a trivial way to batch debounced call without more hacks. I am confident there is now no performance cost left, so that seemed unnecessary to me anyway.
- I looked if there was a way to handle horizontal resize event only, but there is no such thing. All the suggestions I could find online call things like jQuery's `.width()` which... comes at a cost itself, so whatever.

Alright, assuming I didn't leave any bugs in there, this is as efficient as it should get.

client/js/lounge.js
```diff
@@ -49,6 +49,7 @@ $(function() {
 
 		slideoutMenu.toggle(isOpen);
 		storeSidebarVisibility("thelounge.state.sidebar", isOpen);
+		utils.togglePreviewMoreButtonsIfNeeded();
```
xPaw: Perhaps move both of these calls into `storeSidebarVisibility`? That also removes the call on mobile since it's an overlay there.

xPaw:
Thank you for the changes, and sorry for the trouble :)

> I did not find a trivial way to batch debounced call without more hacks. I am confident there is now no performance cost left, so that seemed unnecessary to me anyway.

My thinking was to bind one resize event globally, and basically call `togglePreviewMoreButtonsIfNeeded` inside of it.

client/js/lounge.js
astorije: Will try, thanks!

astorije:
> My thinking was to bind one resize event globally, and basically call `togglePreviewMoreButtonsIfNeeded` inside of it.

That's a good idea, I'll try and see how that goes, thanks!

astorije:
@xPaw, both your suggestions came together nicely IMO, let me know what you think. I have added those as extra commits so you can easily see what I changed specifically.

This PR has seen a lot of changes over time, do you see anything else that could be improved, or that doesn't make sense (like I forgot to add/remove something along the way, left a poorly named variable, etc.)?

client/js/utils.js
```diff
@@ -141,3 +142,9 @@ function requestIdleCallback(callback, timeout) {
 		callback();
 	}
 }
+
+// Force handling preview display
+function togglePreviewMoreButtonsIfNeeded() {
+	$("#chat .channel.active .toggle-content.toggle-type-link.show")
```
xPaw: This should be `.chan.active`.

PR #2273, opened 2018-03-22
Fix order in oidentd file

### LINKED ISSUE #754: Oidentd not giving different idents
Hello. Been trying to ask for help with this via IRC since didn't find anything worth trying by googling, so decided to open an issue.

My problem is, I'm getting ident, but it's always *rolle@pulina.fi* which is exactly the same ident than my irssi connection has for example. It's always the same for *all* users that use lounge, even if the username is completely different. So every lounge webchat user has the same ident than me - that's obviously a problem.

/etc/oidentd.conf
````
# Configuration for oidentd
# see oidentd.conf(5)
#
default {
	default {
		deny spoof
		deny spoof_all
		deny spoof_privport
		allow random
		allow random_numeric
		allow numeric
		deny hide
	}
}
user root {
     default {
          force reply "UNKNOWN"
     }
}
user rolle {
        default {
                allow spoof
                allow spoof_all
                allow random
                allow hide
       }
}
````

~/.oidentd.conf
````
# Warning: file generated by The Lounge: changes will be overwritten!
to 104.207.136.203 lport undefined from undefined fport 6667 { reply "lounge-user" }
to 104.238.168.67 lport undefined from undefined fport 6667 { reply "Bulibuli" }
to 83.140.172.210 lport undefined from undefined fport 6667 { reply "pulisijasafti" }
to 83.140.172.211 lport undefined from undefined fport 6667 { reply "pulisijasafti" }
to 80.65.57.26 lport undefined from undefined fport 6667 { reply "rollefromlounge" }
to 45.32.162.3 lport undefined from undefined fport 6667 { reply "rolle_test" }
to 66.225.225.66 lport 47541 from 139.59.208.244 fport 6667 { reply "pulisafti" }
to 104.207.136.203 lport undefined from undefined fport 6667 { reply "lounge-user" }
````

Currenly running lounge with debugger on in separate screen session, on user rolle.

**System**: Ubuntu 14.04 LTS server (Digital Ocean VPS)
**lounge**: v2.1.0
**node**: v7.0.0

xPaw:
oident'd parser expects exact order and the man page is written incorrectly.

Confirmed bug in https://github.com/janikrabe/oidentd/pull/13

@ronilaukkarinen If you are still around, can you test if this finally fixes the issues you had? Possibly fixes #754

### DIFF
--- a/src/identification.js
+++ b/src/identification.js
@@ -91,9 +91,9 @@ class Identification {
 
 		this.connections.forEach((connection) => {
 			file += `to ${connection.socket.remoteAddress}`
-				+ ` lport ${connection.socket.localPort}`
-				+ ` from ${connection.socket.localAddress}`
 				+ ` fport ${connection.socket.remotePort}`
+				+ ` from ${connection.socket.localAddress}`
+				+ ` lport ${connection.socket.localPort}`
 				+ ` { reply "${connection.user}" }\n`;
 		});
 


dgw:
That's an interesting build error under Node 9. Suddenly the Windows path parser is broken but only on a certain Node version? Weird.

FardaleM:
I test it and it works for me. And the bug in the man page is correct by this [commit](https://github.com/janikrabe/oidentd/commit/1465ef7015260ee9068f92865bf355dddf47409a) and this [commit](https://github.com/janikrabe/oidentd/commit/e290df88f1435255732db0f0bb03d8ce45dcd9a8) allow to specify the parameter in any order. I think it's good to merge this PR so thelounge will work with old version of oidentd.

ronilaukkarinen:
A little late, but confirmed working. Thanks!

PR #2275, opened 2018-03-22
Autocomplete bracket and quote characters like in a modern IDE

xPaw:
Potential issue: browsers by default replace selected text with the character, but this behaviour can still be achieved by using backspace character.

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -127,6 +127,29 @@ for (const hotkey in colorsHotkeys) {
 	});
 }
 
+// Autocomplete bracket and quote characters like in a modern IDE
+// For example, select `text`, press `[` key, and it becomes `[text]`
+const bracketWraps = {
+	'"': '"',
+	"'": "'",
+	"(": ")",
+	"<": ">",
+	"[": "]",
+	"{": "}",
+	"*": "*",
+	"`": "`",
+	"~": "~",
+	_: "_",
+};
+
+inputTrap.bind(Object.keys(bracketWraps), function(e) {
+	if (e.target.selectionStart !== e.target.selectionEnd) {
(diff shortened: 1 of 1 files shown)


dgw:
How does this affect typing simple emoticons, like `:(`?

xPaw:
@dgw It doesn't, as this logic only applies if you have text selected.

client/js/keybinds.js
```diff
@@ -127,6 +127,28 @@ for (const hotkey in colorsHotkeys) {
...
 }
 
+// Autocomplete bracket and quote characters like in a modern IDE
+// For example, select `text`, press `[` key, and it becomes `[text]`
+const bracketWraps = {
+	'"': '"',
+	"'": "'",
+	"{": "}",
+	"[": "]",
+	"(": ")",
```
astorije: Do you want to add `<>` in the list?

client/js/keybinds.js
```diff
@@ -127,6 +127,28 @@ for (const hotkey in colorsHotkeys) {
...
+	"`": "`",
+	"~": "~",
+	"*": "*",
+	_: "_",
```
MaxLeiter: Should `_` be surrounded by quotes for consistency?

astorije: ESLint will say: `Unnecessarily quoted property '_' found`
(If that becomes a philosophical question about linting lol, let's merge this and improve the linting config only after)

McInkay: Time for some philosophical linting discussion!

I think we should enforce all if 1 needs it, but allow non quoted if only none of them need to be unquoted.

But yeah, I don't care about this being in this PR.

client/js/keybinds.js
xPaw: Added.

PR #2281, opened 2018-03-23
Allow /join to assume channel name as open channel

### LINKED ISSUE #2266: Being kicked isn't very intuitive
Being kicked isn't very intuitive, the input field doesn't disable and the kick message for yourself is collapsed by default. Additionally, issuing `/join` to rejoin doesn't work, since the channel name isn't assumed from the currently open buffer.

I'd suggest kick messages for yourself should display by default and perhaps the input field should be disabled if you are not in the channel

MaxLeiter:
Closes #2266

### DIFF
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -63,12 +63,12 @@ function scrollIntoViewNicely(el) {
 
 function collapse() {
 	$(".chan.active .toggle-button.toggle-preview.opened").click();
-	return true;
+	return "";
 }
 
 function expand() {
 	$(".chan.active .toggle-button.toggle-preview:not(.opened)").click();
-	return true;
+	return "";
 }
 
 function join(args) {
@@ -79,6 +79,12 @@ function join(args) {
 
 		if (chan.length) {
 			chan.trigger("click");
+		}
+	} else {
+		const openChannel = $(".network .chan.active");
+
+		if (openChannel.hasClass("channel")) {
+			return "/join " + openChannel.attr("aria-label");
 		}
 	}
 }
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -240,7 +240,7 @@ $(function() {
 		input.trigger("click").trigger("focus");
 
 		const target = chat.data("id");
-		const text = input.val();
+		let text = input.val();
 
 		if (text.length === 0) {
 			return false;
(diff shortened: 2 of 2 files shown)


xPaw:
I'd rather not handle `/join` on the server. This could easily be added in the client handled join command.

MaxLeiter:
Done, @xPaw

client/js/utils.js
```diff
@@ -63,12 +63,12 @@ function scrollIntoViewNicely(el) {
 
 function collapse() {
 	$(".chan.active .toggle-button.toggle-preview.opened").click();
-	return true;
+	return "";
```
astorije: What is this change for?
Really this function shouldn't return anything, it's weirder that it returns `""`, but it's already weird that it returns anything imo.

xPaw: It returned `true` before to block sending the command to the server. This PR changes it to return a string that should be sent to the server (in this case empty string = don't send anything).

We need returns to selectively block sending it to the server.

MaxLeiter: Should this change or good to go?

astorije: Sorry, I didn't see @xPaw's answer earlier. It's honestly very weird. `collapse()` and `expand()` should really not return anything.

It's bizarre right now, because we have 3 commands including 2 that do not go to the server, and the third one returns a string, so we have them all return a string, then check if output is a string and if the string is empty...

I suggest removing the return values from both `expand` and `collapse` functions. Then your code above could be:

```js
if (typeof utils.inputCommands[cmd] === "function") {
	const output = utils.inputCommands[cmd](args);

	if (output === undefined) {
		return false;
	}

	text = output;
}
```

Would this work? I feel like it wouldn't because of the other branch in `join` (when L75 of this file is `undefined`).

MaxLeiter: This will kind of work - the problem with your code is `/join #chan` doesn't get sent to the server at all (as far as I can tell), so /join is now broken.

MaxLeiter: @astorije poke for when you can get to it

astorije: You are correct that my naive solution is incorrect.
I still think we should refactor a little bit to avoid having a `return ""` to represent "this command is local and should not go to the server". This is just too much error-prone and we _will_ break this soon enough (silently or not).

Maybe @xPaw's suggestion to combine with #2004 could help making this a bit nicer?

MaxLeiter:
I modified #2004 to address the problem discussed above, but did not change it to include this PR as I think its out of the scope of it. 

For this PR to be functional, I will need to add a `isUserInChannel(channel, [nick])` function to utils. This is because the channel may be open, but we may not be in it, so the UI will click the channel, which is empty (as of https://github.com/thelounge/thelounge/commit/bb1e3ee9177a76c54b5ea70511055fb385a4ffe5#diff-e034c06e391c80400520affdba483eb5R31)

xPaw:
Gonna close this as I dealt with this in Vue branch. There's minor but important details to get right with this refactor.

I also addressed @astorije's comment about return values, join command emits socket event on its own now.

PR #2282, opened 2018-03-23
Reduce lightness of low contrast nick colors to reach AA level of WCAG 2.0

astorije:
I know this is going to make some people unhappy, but some of our nick colors were outrageously low-contrasted. Some of them were even very far from [accessible color contrast guidelines](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html).

So rather than doing it by ear (well, eye), which would be subjective and always upset someone, I went with something more objective:

- I entered _each_ color against [this checker](https://leaverou.github.io/contrast-ratio/) (on white background).
- For each color that does not pass AA (i.e. contrast of < 4.5), I picked the HSL equivalent of that color, and reduced the Lightness until it reaches exactly 4.5.

That way, I keep the same Hue and Saturation, in order to keep colors as close to their original ones.
There will be values close to each others, some that people will dislike on their own nick, but I thought this would be the most fair way to get decent values.

Some before/after examples from the user list:

screen shot 2018-03-23 at 00 56 56screen shot 2018-03-23 at 00 57 04&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;screen shot 2018-03-23 at 00 57 44screen shot 2018-03-23 at 00 57 52

screen shot 2018-03-23 at 00 58 32screen shot 2018-03-23 at 00 58 40&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;screen shot 2018-03-23 at 00 59 38screen shot 2018-03-23 at 00 59 46

And if interested, the before/after on colors themselves (remember that effect of color on text is different than background, what seems fairly visible in background might actually be very difficult to see once on text):

screen shot 2018-03-23 at 01 17 35

[If I ever need to generate this again](https://zerobin.net/?63ff7919d0f06474#hUzJvuJxR0+/NxNv6cmsqeY3D+A5PyD1qwVXuonmygA=)


astorije:
@xPaw, updated Zenburn colors as well, overall very little difference with the theme prior to my changes, which is good. Before/After screenshots below.
I've also changed my method, using http://accessible-colors.com/ now, which outputs the same results than the tool mentioned above, except more convenient to use.

Are you still good with this?

I know I need to do that with Morning as well, but I'd like to change Morning to use our official colors first (very little difference with what's on `master`, but still) and cba to do it right now.

screen shot 2018-03-24 at 01 50 22screen shot 2018-03-24 at 01 50 32&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;screen shot 2018-03-24 at 01 50 45screen shot 2018-03-24 at 01 50 56

screen shot 2018-03-24 at 01 49 18

[Updated snippet](https://zerobin.net/?491b64d52eb5e59d#CahbUEal9Ip2oYANNCZIoP+LdCf1hWs3UYBGKjKXtgI=)

PR #2286, opened 2018-03-23
Reimplement input history

xPaw:
I believe the behaviour should be the same as the plugin version.

~~Something I've noticed: it is possible to overwrite history elements (in both versions). Scroll up and start typing, scroll again, and whatever you've typed gets stored at that position.~~

Changes:
- If you scroll up in history and start typing, history position is reset which means it no longer overwrites history entries.
- If you have text selected and press arrows, it no longer scrolls the history.


client/js/keybinds.js
```diff
@@ -103,6 +104,54 @@ Mousetrap.bind([
 });
 
 const inputTrap = Mousetrap(input.get(0));
+
```
McInkay: At the risk of keybinds getting too big, can we stick this in a different file?

I actually think that the keybindings should probably go in next to code similar to it rather than having 1 massive folder defining all the keybindings, but at the very least, can we just have the history code in a different file, and just have the keybinding here?

xPaw: The only real reason I put it here is because I need access to `inputTrap`.

McInkay: We can grab that from elsewhere, can't we? Is there any disadvantage to just instantiating it in another module again? I assume it caches.

xPaw: Mousetrap doesn't cache and it would bind a separate event again. What about moving code that uses `inputTrap` to separate file? Call it `inputHandler` or something.

McInkay: Euch. That feels like just kicking the can down the road. Alright, keep it here for now and we can rethink how we structure these later on.

PR #2287, opened 2018-03-23
Send `Accept` header when fetching links

### LINKED ISSUE #2277: Imgur gallery previews are 404 due to missing "Accept" header
It seems that imgur is requiring `Accept` header to be sent. As adding `accept: text/html` seems to help.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values

xPaw:
Closes #2277. These values should match exactly what Firefox sends.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -52,7 +52,10 @@ module.exports = function(client, chan, msg) {
 	})).slice(0, 5); // Only preview the first 5 URLs in message to avoid abuse
 
 	msg.previews.forEach((preview) => {
-		fetch(normalizeURL(preview.link), {language: client.language}, function(res, err) {
+		fetch(normalizeURL(preview.link), {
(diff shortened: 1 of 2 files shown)


src/plugins/irc-events/link.js
```diff
@@ -52,7 +52,10 @@ module.exports = function(client, chan, msg) {
 	})).slice(0, 5); // Only preview the first 5 URLs in message to avoid abuse
 
 	msg.previews.forEach((preview) => {
-		fetch(normalizeURL(preview.link), {language: client.language}, function(res, err) {
+		fetch(normalizeURL(preview.link), {
+			accept: "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
```
astorije: I don't think so but do you need to do something similar in L108?

xPaw: Correct, not needed there. It will default to `*/*`.

PR #2289, opened 2018-03-24
Add auto-prepend behaviour for unprefixed channel names

lol768:
This change adds behaviour to automatically prefix channel names passed in via the "?join=x,y,z" query string/search parameter which do not appear to include an appropriate channel symbol. For example, `channel` becomes `#channel` but `#channel` and `~channel` are left alone.

**Known limitation:** Unicode characters at the start of the channel name will not be considered to match `/^\w/`. These will not be auto-prefixed.

If this is not the correct place to implement this functionality, please let me know where it would be better suited.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -585,7 +585,19 @@ $(function() {
 			// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Iterating_over_own_properties_only
 			for (let key in params) {
 				if (params.hasOwnProperty(key)) {
-					const value = params[key];
+					let value = params[key];
+
+					if (key === "join") {
+						const channels = value.split(",");
+						value = channels.map((c) => {
+							if (c.match(/^\w/)) {
+								return "#" + c;
+							}
+
+							return c;
+						}).join(",");
+					}
+
 					// \W searches for non-word characters
 					key = key.replace(/\W/g, "");
 


client/js/lounge.js
```diff
@@ -585,7 +586,20 @@ $(function() {
 			// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Iterating_over_own_properties_only
 			for (let key in params) {
 				if (params.hasOwnProperty(key)) {
-					const value = params[key];
+					let value = params[key];
+
+					if (key === "join") {
+						let channels = value.split(",");
+						channels = _.map(channels, (c) => {
```
xPaw: Use a native method instead of bringing lodash here (we don't use it on the client).

lol768: I've addressed this and squashed the commit.

client/js/lounge.js
```diff
@@ -585,7 +585,20 @@ $(function() {
...
+					if (key === "join") {
+						let channels = value.split(",");
+						channels = channels.map((c) => {
+							if (c.match(/^\w/)) {
+								return "#" + c;
+							}
+
+							return c;
+						});
+						value = channels.join(",");
```
xPaw: You can chain `value = split.map.join`.

lol768: Good shout, changed

xPaw: Did you push the change? Doesn't appear to be changed.

lol768: Would seem I didn't `git add` before amending the commit. Fixed now in 25dee77

MaxLeiter:
👍 from me

xPaw:
@MaxLeiter You can review the PR from files changed tab.

astorije:
> **Known limitation:** Unicode characters at the start of the channel name will not be considered to match `/^\w/`. These will not be auto-prefixed.

Would it work if, instead of adding `#` when the string starts with `\w`, this add `#` if the string does not start with `#`, `&`, or any [other chan types](https://www.alien.net.au/irc/chantypes.html) (at the moment [we only parse `&` and `#` in messages](https://github.com/thelounge/thelounge/blob/master/client/js/libs/handlebars/parse.js#L81), but I wouldn't be surprised if we had similar logic in more than one place).

lol768:
> Would it work if, instead of adding # when the string starts with \w, this add # if the string does not start with #, &, or any other chan types (at the moment we only parse & and # in messages, but I wouldn't be surprised if we had similar logic in more than one place).

We could certainly make the decision to limit it to those characters (and have some centralised logic somewhere for fetching/checking for channel type characters).

astorije:
What do @xPaw and/or @McInkay prefer?

McInkay:
Ideally we should use chantypes from the server, no?

xPaw:
We don't know them before connecting.

McInkay:
Hmmm, but this is just for joining channels, isn't it? So we should already be on the network.

McInkay:
In that case, I'd prefer the current solution than just using `#` and `&`.

astorije:
Alright then. I still think we should check for chan prefixes (and we can probably do it statically by including chan types that might not be on this specific server), but I guess this is the most conservative approach.

astorije:
(I'm merging as-is, the linting issue reported by Travis is not caused by this PR and was fixed on `master`)

astorije:
Thanks a lot @lol768 and congrats on your first merged PR!! :) 🎉

astorije:
Hey @lol768, we have sticker packs for our contributors now!

lol768:
Awesome stuff @astorije :smile:  - filled in the form!

PR #2290, opened 2018-03-24
Add ?channels=x,y,z alias for ?join=x,y,z

lol768:
For backwards compat. with Iris and other systems.

`?join=%23foo,%23bar` and `?channels=%23foo,%23bar` are treated as equivalent with this patch.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -581,6 +581,10 @@ $(function() {
 		$("#connect").one("show", function() {
 			const params = URI(document.location.search).search(true);
 
+			if ("channels" in params) {
+				params.join = params.channels;
+			}
+
 			// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
 			// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Iterating_over_own_properties_only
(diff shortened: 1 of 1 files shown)


xPaw:
@astorije We'd be breaking existing links that use `join`. Would also need to change client code to convert channels into an array:

https://github.com/thelounge/thelounge/blob/d8eed4d15e1506b3032d0404df21af445eb51a45/src/client.js#L166-L168

xPaw:
If we want to go with removing `join`, then client needs to split channels string into an array before sending. And there needs to be more checks added on the server (e.g. is channels actually an array, before we try to iterate it).

I'm not opposed to this change.

astorije:
No I meant with this PR

xPaw:
Oh right. Either solution is fine by me. If you guys want this, then go ahead.

lol768:
Personally not a fan of breaking BC here.

Happy to make the change regardless, but would prefer if merged in current state as not to break existing `join=` links.

astorije:
Yeah that's fine like this. I'll let @xPaw give a first review.

lol768:
:smile: Thanks for reveiwing, both.

PR #2291, opened 2018-03-24
Add lockUsername config option

lol768:
Allows network administrators to force specific usernames (intended for use in public mode).

Feature parity: iris.conf ident_string

May be preferable to hide the username field entirely?

### DIFF
--- a/client/views/windows/connect.tpl
+++ b/client/views/windows/connect.tpl
@@ -66,7 +66,11 @@
 			<label for="connect:username">Username</label>
 		</div>
 		<div class="col-sm-9">
-			<input class="input username" id="connect:username" name="username" value="{{defaults.username}}">
+			{{#if lockUsername}}
+				<input class="input username" name="username" disabled value="{{lockUsername}}">
+        	{{else}}
+        		<input class="input username" id="connect:username" name="username" value="{{defaults.username}}">
+        	{{/if}}
 		</div>
 		{{/unless}}
 		<div class="col-sm-3">
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -119,6 +119,17 @@ module.exports = {
 	lockNetwork: false,
 
 	//
+	// Lock username
+	//
+	// If enabled, users will be forced to connect with
+	// the specified username.
+	//
+	// @type     string
+	// @default  null
+	//
+	lockUsername: null,
+
+	//
 	// Hex IP
 	//
 	// If enabled, clients' username will be set to their IP encoded has hex.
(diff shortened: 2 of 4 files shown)


xPaw:
I don't want yet another way for usernames to be restricted (it's already enough of a hassle with hexip).

Is there any reason you can't use hexip if you need to lock usernames?

lol768:
I do not see the point of `hexIp`. Not only does it not work for v6, each client is surely going to have a different username which does not help with our goal of locking the username to something static.

Our use-case is not breaking existing bans targeting `*!webchat@*` or similar.

xPaw:
> Our use-case is not breaking existing bans targeting *!webchat@* or similar.

That's a minor thing for us to merge this, all things considered. webirc configured is going to allow you spoof hostnames which will allow you ban webchat if you really want to (yes, it will break existing bans).

lol768:
> That's a minor thing for us to merge this, all things considered.

Fair enough. I'm sure you can appreciate why BC is a priority for me though :smile: 

> webirc configured is going to allow you spoof hostnames which will allow you ban webchat if you really want to (yes, it will break existing bans).

Yes, we've experimented with prepending `webchat/` to hostnames and this functionality works as described. Yes, it could be used for bans going forward.

----------

No worries if you decide not to merge this. Everything I can get upstreamed though is one less patch though so I thought I'd try anyway in the absence of client packages.

lol768:
>webirc configured is going to allow you spoof hostnames which will allow you ban webchat if you really want to

The other issue with this approach is that bans for a hostname don't carry over which again breaks user expectations. Consider:

* User A joins channel and spams. Chanop bans `*!*@username.isp.com`
* User A is annoyed and attempts to use webchat to evade. Is assigned `webchat/` prefixed hostname which is then not banned and allows them to effortlessly bypass the ban.

xPaw:
I believe this can be solved on the ircd? AFAIK on freenode if you ban an ip, it also bans webchat masks.

xPaw:
We've discussed internally, and have decided not go with this. It increases complexity (e.g. in your PR you haven't covered all the cases, and it will become even more troublesome when network editing is in.). Sorry for the trouble.

lol768:
No worries @xPaw - appreciate the response :)

lol768:
> in your PR you haven't covered all the cases, and it will become even more troublesome when network editing is in

For our benefit, would you mind briefly explaining the cases that aren't covered by this diff?

xPaw:
Basically the same cases where `hexip` is handled.

PR #2292, opened 2018-03-24
Fix lint issue on master; increase timeout globally on CI

xPaw:


test/server.js
```diff
@@ -5,10 +5,11 @@ const expect = require("chai").expect;
 const request = require("request");
 const io = require("socket.io-client");
 
-describe("Server", function() {
-	// Travis is having issues with slow workers and thus tests timeout
-	this.timeout(process.env.CI ? 25000 : 5000);
+if (process.env.CI) {
+	mocha.setup({timeout: 15000});
```
astorije: `mocha` doesn't exist here 🤷‍♂️

xPaw: Welp that only works in the browser. There's no good way to configure mocha conditionally.

PR #2294, opened 2018-03-25
Mark advanced settings as advanced in settings

MaxLeiter:
screen shot 2018-03-25 at 11 09 52 am

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1592,6 +1592,11 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	display: none;
 }
 
+#settings .advanced {
+	opacity: 0.6;
+	float: right;
+}
+
 #settings .opt {
 	display: block;
 	padding: 5px 0 5px 1px;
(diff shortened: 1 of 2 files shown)


client/css/style.css
```diff
@@ -1598,6 +1598,12 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	display: none;
 }
 
+#settings [data-advanced] h2::after {
+	content: "advanced";
```
xPaw: That's gonna be painful to translate and you know it :)

MaxLeiter: translations be damned?

MaxLeiter: ill find another way 😞

xPaw:
@MaxLeiter Why not just put it in the options template?

MaxLeiter:
@xPaw: feels like that's duplicating the fact an option is `advanced`. Also, plugins should be able to add advanced options long-term and this will just complicate the template. I'm open to changing it.

PR #2297, opened 2018-03-26
Fix/Improve some nick fallbacks

astorije:
- Rename a forgotten `lounge-user`
- Generate nick fallbacks when already in use by appending 0-9 instead of 10-98 (?!).
- Generate nick fallbacks when invalid similarly to our config default instead of random string. This is to make it less confusing when fallback gets used.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -140,7 +140,7 @@ Client.prototype.find = function(channelId) {
 
 Client.prototype.connect = function(args) {
 	const client = this;
-	const nick = args.nick || "lounge-user";
+	const nick = args.nick || "thelounge";
 	let webirc = null;
 	let channels = [];
 
--- a/src/plugins/irc-events/error.js
+++ b/src/plugins/irc-events/error.js
@@ -39,7 +39,7 @@ module.exports = function(irc, network) {
 		lobby.pushMessage(client, msg, true);
 
 		if (irc.connection.registered === false) {
-			const random = (data.nick || irc.user.nick) + Math.floor(10 + (Math.random() * 89));
+			const random = (data.nick || irc.user.nick) + Math.floor(Math.random() * 10);
 			irc.changeNick(random);
 		}
 
@@ -59,8 +59,7 @@ module.exports = function(irc, network) {
 		lobby.pushMessage(client, msg, true);
 
 		if (irc.connection.registered === false) {
-			const random = "i" + Math.random().toString(36).substr(2, 10); // 'i' so it never begins with a number
-			irc.changeNick(random);
+			irc.changeNick("thelounge" + Math.floor(Math.random() * 100));
 		}
 
 		client.emit("nick", {


src/client.js
```diff
@@ -140,7 +140,7 @@ Client.prototype.find = function(channelId) {
 
 Client.prototype.connect = function(args) {
 	const client = this;
-	const nick = args.nick || "lounge-user";
+	const nick = args.nick || "thelounge";
```
astorije: This was changed in https://github.com/thelounge/thelounge/commit/84e524c6352e3a7f6c093585d408c0582568760d. However, I'm wondering if we actually still need this fallback.

src/plugins/irc-events/error.js
```diff
@@ -39,7 +39,7 @@ module.exports = function(irc, network) {
 		lobby.pushMessage(client, msg, true);
 
 		if (irc.connection.registered === false) {
-			const random = (data.nick || irc.user.nick) + Math.floor(10 + (Math.random() * 89));
+			const random = (data.nick || irc.user.nick) + Math.floor(Math.random() * 10);
```
astorije: This is really, [reeeeeally old](https://github.com/thelounge/thelounge/commit/4ef13d6a18551a27d473cb38583bb5300e0cb72c#diff-4f2bea0c158bc3b83f37d4b5c02bb65bR16). It turns out it was generating numbers between 10 (no idea why) and 98 (that's most likely an overlook).

I reduced to 0-9 because this will keep appending numbers anyway, and I find it confusing that requesting `thelounge42` could lead to `thelounge4265`. This change would just make it `thelounge420` to `thelounge429`. If this one already exists, a new integer will be appended, and so on.

src/plugins/irc-events/error.js
```diff
@@ -59,8 +59,7 @@ module.exports = function(irc, network) {
 		lobby.pushMessage(client, msg, true);
 
 		if (irc.connection.registered === false) {
-			const random = "i" + Math.random().toString(36).substr(2, 10); // 'i' so it never begins with a number
-			irc.changeNick(random);
+			irc.changeNick("thelounge" + Math.floor(Math.random() * 100));
```
astorije: This reduces confusion by going to something that looks like our default fallback.

I debated bringing this from the default config file itself, but that would have meant moving code around to put [this](https://github.com/thelounge/thelounge/blob/5738642d44c1f4c2dd44b1f67dca35bc8c84c96a/src/server.js#L572) into a shared location, but I figured it was too much of a hassle at 3am for such a simple change.
Happy to continue the improvement, but that'd be in a further PR. Where would it be a reasonable shared location?

src/plugins/irc-events/error.js
xPaw: It generated from `10` so that it always ended up with 2 numbers (if it generated a `5` then it wouldn't have a zero prefix).

src/client.js
xPaw: We do need a fallback, I've changed it in #2229 by the way.

astorije: Mind reminding me what this fallback is for, vs the one in config defaults?

src/plugins/irc-events/error.js
astorije: Oh wow, ok. Well, I definitely prefer the one-int version then!

src/client.js
xPaw: Actually good question as to why we don't use the one from `defaults`. But the fallback is to disallow connecting with an empty nick.

I'll fix the default in my PR.

astorije: Cool! Feel free to remove from my pr if you're happy with the rest and ok to merge

Issue #2301, opened 2018-03-27
installing with yarn does not install yarn locally... and `thelounge install` does not handle stderr from the yarn process


benharri:
* *Node version:* 6.13.1
* *Browser version:* firefox 59
* *Device, operating system:* ubuntu 16.04
* *The Lounge version:*  v3pre3

---

I installed thelounge with yarn: `yarn global add thelounge@next`. Somewhere along the line, the `yarn` executable (used for `thelounge install`) does not get installed locally. 

Additionally, the stderr from `thelounge install` is not returned, leaving just `Failed to install` and an exit code of 1.

xPaw:
We probably should fallback to executing global `yarn` if the locally installed one doesn't exist.

benharri:
Yeah that's a good idea. 

AFAICT, yarn installs the dependencies at one level, skipping duplicates for the node_modules of each installed package and installing those at the top level. This means we can't depend on the yarn executable to be inside thelounge's node_modules.

xPaw:
You are right about that. However `thelounge/node_modules/.bin/yarn` does exist, which is symlinked to the javascript file.

benharri:
I'm not seeing any symlinks to yarn on linux (ubuntu 16.04) in `thelounge/node_modules`

Just: `ansi-styles/  chalk/  supports-color/`

astorije:
Fixed by #2303.

xPaw:
It isn't fixed.

astorije:
Woops, my bad!

PR #2303, opened 2018-03-27
catch stderr from yarn subprocess, addresses #2301

benharri:


src/command-line/install.js
```diff
@@ -90,6 +90,10 @@ program
 				});
 			});
 
+			add.stderr.on("data", (data) => {
+				log.error(data);
```
xPaw: Does it work without `toString`?

benharri: Good call. I added `.toString()`

xPaw:
Can you rebase your PR correctly?

benharri:
Fixed the rebase. Should be good to go.

benharri:
Squashed!

PR #2308, opened 2018-03-29
Add client plugin support

McInkay:


McInkay:
Just to be clear, this is purely client plugins in general, with no API. It's probably simplest to agree on how we do client plugins separately from anything else.

This does have a use to merge in on it's own as well because people can create plugins that touch the dom and do some messing around that way, which will give people somewhere to start, and the APIs for our Javascript can come later.

client/index.html.tpl
```diff
@@ -7,6 +7,11 @@
 
 	<link rel="preload" as="script" href="js/bundle.vendor.js">
 	<link rel="preload" as="script" href="js/bundle.js">
+	<!--we import the package manager first, so that it's cached for plugins and plugins can't override the object -->
```
xPaw: Why are they included in head and not after lounges code?

client/js/packages/packageManager.js
```diff
@@ -0,0 +1 @@
+export const contextMenu = window.TheLounge.contextMenu;
```
McInkay: This shouldn't have been left in btw.

client/index.html.tpl
McInkay: It is after lounges code, it's after the bundle.

xPaw: This is preload. This should be at the bottom of the file

McInkay: Oh I missed that. Alrighty. I need to re-test this and make sure things actually work anyway, but yeah good point.

client/js/packages/packageApi.js
```diff
@@ -0,0 +1,24 @@
+// Our public API
+window.TheLounge = {
+};
```
McInkay: I think I should probably add at least 1 tiny thing in here so I can test this stuff works, something like ... maybe ... the version of the lounge that we are running against maybe?

lol768: >maybe ... the version of the lounge that we are running against maybe

Would be useful, IMO

McInkay:
I think I'll close this just now, but keep the code. I think I should probably focus on building up the server-side api before getting this in. We'll get there.

astorije:
@McInkay, I'll delete the branch to remove clutter, but that doesn't mean the code goes away: closed PRs are like "archived" so code stays forever, and you can re-open the branch whenever you're ready to pick this up. I'm going to do the exact same thing in #1950 :)

PR #2315, opened 2018-03-30
Revamp default theme a bit

astorije:
Before going into details, my objectives with this:

- Make sure text colors are accessible (WCAG level AA, contrast of 4.5 minimum)
- Update our stock theme, renovate the styling for v3, visually split from Shout
- Use our official colors
- Display the logo in channel list
- Emphasize opened channel
- Make sure none of the changes enforce styling onto themes
- Do not unnecessarily reduce available screen estate simply for styling sake

## Random thoughts

In terms of accessibility and colors, I based everything on the official blue/gray, used https://color.adobe.com/ for decent complimentary colors/shades, and tweaked light ones with http://accessible-colors.com/ to ensure accessible contrasts.
Once this is settled, I will write down rules / helpers for colors so we don't have to reinvent the wheel / re-define new colors by accident each time.

In terms of screen real estate, the outcome is virtually equivalent: on the chat window, -1px in one direction, +1px in the other, so pretty much nothing. No changes on mobile.
The logo obviously eats a bit on the channel list, but this is mitigated by smaller margins between networks and a slightly less high footer, and overall the consequence is minimal. The logo can be easily hidden via custom CSS if that's a big deal, which results in actually bigger channel list real estate than currently.

At the moment, themes outside of the default one are completely untouched and probably look broken. I did not worry about them for now, but obviously whatever the conclusion, themes will be fixed.
Similarly, the code is not fully ready yet, it needs to be ironed out and cleaned up, but I wanted to show some results first.

I think this integrates pretty well with the changes I'm proposing for the previews (https://github.com/thelounge/thelounge/pull/2272).

This was not a goal, but as a side-effect, this effectively solves https://github.com/thelounge/thelounge/issues/2239.

You will notice that the nick editor is absent from the screenshots. The reason is that it is on its way out. Once https://github.com/thelounge/thelounge/pull/2229 makes it into `master`, it will be possible to change nick from there. For quick nick update, `/nick` will still work. This gives back some precious real estate, especially on mobile, when sending a message.

## Before we get started...

Alright so I know this will most likely get a lot of pushback (UI changes usually do) and strong opinions. Keep in mind that I am doing nothing to break people's workflow on purpose, make them unhappy, force my preferences onto others, etc. I really want what's best for the project and the people using it (or none of us would spend hours of their free time on it). Yes, there are personal preferences mixed in with this, obviously, but I spent several hours experimenting, reading and using tools to make it as objective as possible.
I'd love to have a healthy discussion, please bear with me and keep an open mind :) Thanks a lot in advance!! 🙏

## Results

I suggest opening these screenshots in full screen to get a better sense of the results.

**Active state on channel list items:**

screen shot 2018-03-30 at 00 53 17

**Footer icons now have an active state:**

screen shot 2018-03-30 at 00 52 55

**How it looks with previews from #2272:**

screen shot 2018-03-30 at 00 53 31

**Channel list and user list closed:**

screen shot 2018-03-30 at 00 53 52

**Public mode landing page:**

screen shot 2018-03-30 at 00 58 20

**Mobile views:**

screen shot 2018-03-30 at 00 59 57 screen shot 2018-03-30 at 01 00 08


dgw:
Holy shit.

You're not getting any pushback from me. This is :cherry_blossom:**beautiful**:cherry_blossom: work! Honestly, the only thing I don't like is the white, because I prefer dark backgrounds, but 1) I know there will be a dark version eventually once this concept is finalized and 2) I can't argue that it looks super clean with a white background on the chat.

I'm guessing this will mean putting the unread indicator from mobile onto desktop when the channel list is closed, which is :+1: as far as I'm concerned. (And if not, also :+1:, I don't really care because we still have a favicon indicator to work with too.)

About the only thing I can see to nitpick (without staring at the screenshots for twenty minutes) is that empty network names look strange consuming the space on the left that would be occupied by a caret if they contained any channels/queries/windows. A quick search didn't help me find an issue where that was discussed, but I'm pretty sure it was (and it's probably not this PR's responsibility to change if wanted).

Unlike in that other thread (2239), I _am_ going to hit you this time, @astorije — but it's a proverbial slap on the back for doing something amazing. :heart_eyes_cat:

astorije:
Hot damn, now I can go to sleep happy! Thanks for the kind words, @dgw :)

I am indeed planning to replicate whatever we have in the dark version (clearly Morning is the dark equivalent of Example) because it's kind of industry standard to provide both a bright and dark version, especially when audience has developers in it :)

> I'm guessing this will mean putting the unread indicator from mobile onto desktop when the channel list is closed, which is 👍 as far as I'm concerned.

Are you talking about screen shot 2018-03-30 at 02 43 51 (after) / screen shot 2018-03-30 at 02 44 05 (before)?
If so then I can't take the credit, that comes from https://github.com/thelounge/thelounge/pull/2212 :)

> empty network names look strange consuming the space on the left that would be occupied by a caret

Yep, this was discussed in https://github.com/thelounge/thelounge/pull/2149. The conclusion was either take the whole width, or leave a blank (but do not leave a caret that hides/shows nothing), and we went for this 🤷‍♂️
Empty network is a temporary or broken thing, so I'm not worried too much, but if there are preferences for the other way, I really don't mind, as long as we don't display a caret that looks broken. However, if we do anything about it, I'd prefer if this is done outside of this PR, it's not strictly related/caused by the proposal here, if you don't mind.

Thanks again for your input!

client/css/style.css
```diff
@@ -1840,6 +1846,10 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	flex: 0 0 auto;
 	border: 1px solid transparent;
 	transition: border-color 0.2s;
+
+	/* TODO: Get rid of this once networks can be edited
+					 See https://github.com/thelounge/thelounge/pull/2229 */
+	display: none;
```
xPaw: Can we exclude this from this PR please.

client/index.html.tpl
```diff
@@ -32,6 +32,7 @@
 	<body class="signed-out<%- public ? " public" : "" %>" data-transports="<%- JSON.stringify(transports) %>">
 		<div id="viewport" role="tablist">
 			<aside id="sidebar">
+				<img src="img/logo-horizontal-transparent-bg-inverted.svg" class="logo" alt="The Lounge" height="45">
```
creesch: This logo isn't included in this commit.

creesch:
In general it looks good, I do have some things that I am not sure are ideal. 

In a practical sense, you say that the screen estate impact is minimal which is true but it somehow feels less light and a bit more cramped. I think it has to do with the shadow, when I disable `box-shadow` on `#windows .window` it suddenly feels much cleaner to me personal. 


I don't think I like the logo being included, I mean it is a pretty logo and I understand you wanting to show it but it doesn't add much value to me personally as I know I am using the lounge ;) So it is just visual clutter for me so I'll probably hide it with custom css.

client/index.html.tpl
xPaw: Also wrap the logo in a div.

client/css/style.css
astorije: Yes of course, I should have written WIP or REMOVE and not TODO, I meant to leave it untouched, but while I didn't want to have it get in the way or appear on the screenshot while I was iterating on this. I am not planning to work on its style since it's meant to go, but yes I will remove it from this PR :)

client/index.html.tpl
astorije: 🤦‍♂️
Of course I forgot to include it.

Alright, I have pushed it, but it's quickly as I have 5 minutes on the laptop. It's not properly minified, no non-inverted version not wrapped in a div, etc.
I'll look and address other comments later :)

MaxLeiter:
Would it be out of this PRs scope to fix the absurdly large margin between the top of the page and the page title? (In this case, `Help`)

screen shot 2018-03-30 at 10 35 57 pm

astorije:
Good point, I'll look into it. I'd rather do it outside of this PR, but yeah I'll try something and open a PR for feedback!

client/index.html.tpl
astorije: > Also wrap the logo in a div.

@xPaw, you sure? As far as I can tell, we're not doing this for other logos (loading and sign in pages).

astorije:
> I don't think I like the logo being included, I mean it is a pretty logo and I understand you wanting to show it but it doesn't add much value to me personally as I know I am using the lounge ;)

Hey hey, I know you're not criticizing the logo, don't worry I didn't take it like that ;)

I think I like it that way, but I am not married to having the logo here. I remember a few weeks (days? iono, my memory is 💩) ago we mentioned adding the logo here with the team, but I don't think we had a strong conclusion. Happy to hear thoughts from @thelounge/maintainers too.

An alternative would be this:

screen shot 2018-04-01 at 19 32 56

@creesch, what do you think of this? I understand your point that you know you're using The Lounge so repeating the name is not the most important thing.
A few additional thoughts:

- At the moment the scrolling is only on the channels, so logo and footer always stay on screen. I could change this to include the logo in the scrollable area so it's not always on screen when scrolling.
- On public mode, you don't necessarily know you're using The Lounge, so if conclusion is to not have the logo at all there, I would still vote add it when connected to no networks, i.e. at the top of:
  screen shot 2018-04-01 at 19 53 50

Let me know what you think. Thanks for your input!

astorije:
> the only thing I don't like is the white, because I prefer dark backgrounds, but 1) I know there will be a dark version eventually once this concept is finalized

@dgw, I finally managed to give some love to Morning.
My goal is eventually that the default theme and Morning are identical, with the only difference that the latter is a dark version of the former.
This is 1) to simplify development and maintenance on the main repo, and 2) it gives an easy choice for whenever a new user arrives on the project and wants a dark version.
I also improved contrasts (accessibility) on Morning a bit.
Anyway, let me know what you think! 😊

screen shot 2018-04-03 at 01 12 14 screen shot 2018-04-03 at 01 12 25

dgw:
@astorije :cherry_blossom::cherry_blossom::cherry_blossom: Why is there no `:grabby_hands:` emoji?! :grinning:

Oh (sidetracking a bit), I read through #2149 and saw support for leaving the blank space around empty network items, but not for taking the whole sidebar width. But it is what it is, and networks shouldn't be empty often enough to matter that much. :)

MaxLeiter:
@creesch:
Regarding it feeling more cramped, I agree.
However, instead of removing the shadow, I think shrinking the padding is a better option.

<table>
<th> Before </th> <th> After </th>
<tr>
<td> screen shot 2018-04-02 at 11 05 06 pm</td>
<td> screen shot 2018-04-02 at 11 04 49 pm
</td>



</table>

client/index.html.tpl
xPaw: Wrapping is gonna allow for better styling (e.g. a background). It's more reasonble on the sidebar because there's not much you can do with it on the loading/sign in screens.

astorije:
Sounds good, will look into that.

@thelounge/maintainers, what do you prefer re: logo? [Logo+logotype](https://github.com/thelounge/thelounge.github.io/blob/master/assets/logos/logo/TL_White%26Yellow_Horizontal_logotype_Grey_Bg/TL_White%26Yellow_Horizontal_logotype_Grey_Bg.svg) or [logo-only](https://github.com/thelounge/thelounge.github.io/blob/master/assets/logos/logo/TL_White%26Yellow_Logo_Grey_Bg/TL_White%26Yellow_Logo_Grey_Bg.svg)?
I am fine with either. Logo-only for private mode and logo+logotype for public mode sounds good to me (though if it's easier/makes more sense to do and maintain, I could also discriminate on empty/non-empty channel list rather than mode, it doesn't matter).

xPaw:
> Logo-only for private mode and logo+logotype for public mode sounds good to me 

That's fine by me.

williamboman:
This is so awesome 👏! A few minor remarks from me:

- I think the logo in the sidebar is a bit too much, I've hidden it completely with custom CSS.
- I think the background-color transition effect on channel hover is too sluggish, perhaps speed it up a bit?

astorije:
Thanks @williamboman, I'll look into that!

astorije:
> I think the background-color transition effect on channel hover is too sluggish, perhaps speed it up a bit?

@williamboman, I have disabled the transition when hovering on channels, so it's only enabled when hovering off. Could you play with it and tell me what you think? I don't know myself if I prefer this or remove the transition entirely, so happy to hear feedback from you.

astorije:
@williamboman, agreed. I have removed the transition.

astorije:
Alright, I've made improvements here and there, tried to integrate people's feedback (to the extent of what I was trying to achieve initially, sorry 😅).

I tested on several devices, monitors, laptops, etc. I actually expected much worse results on shitty old, cheap TN monitors but it looked pretty good (aaand that also showed me that some of our contrasts are actually horrible on cheaper monitors, but that's a fight for another day)!
I also showed the result to a few people (not nearly enough to call it a UX panel, but more than just me and my own opinion) to get feedback and it was ok overall.

To make things smoother, I also created a [new theme called "Classic"](https://github.com/thelounge/thelounge-theme-classic) that they can simply use, or take this as an example to build custom CSS snippets. The theme is not published yet because it depends on this PR, but I will publish it as soon as this lands.

This will not be perfect from the start (plus there will always be some subjectivity involved), and I'm obviously open to making it evolve (before v3 or after v3 or both), but I like to think this is a decent starting point to split from the Shout-legacy default theme. I'd love if we could go with this and improve over time as feedback/bug reports arise :)

xPaw:
Tried last update here's few thoughts:

- The fading on scrollable elements doesn't seem to fit to well. I don't think it's needed.
- Transitions on hover elements (channel list mainly) isn't necessary. I have been against it before and still am. I can sometimes feel it lag behind my mouse.
- What do we actually achieve by making the chat window "float"?
 - This looks somewhat acceptable on the white default theme but on Morning it adds nothing especially with the shadow that just blends it in.

astorije:
> The fading on scrollable elements doesn't seem to fit to well. I don't think it's needed.

I had moved the commit to #2527, and replied to your comment there. We can keep the discussion there, I'm not planning to re-add the commit here regardless of what we agree on :)

> Transitions on hover elements (channel list mainly) isn't necessary. I have been against it before and still am. I can sometimes feel it lag behind my mouse.

I agree that it felt laggy on channel list, like @williamboman reported, and I had removed the background transition. Wondering if you ran the latest version.

> What do we actually achieve by making the chat window "float"?

... looking good :)
I think the border-radius is a really nice touch, and that wouldn't be possible without the padding. Besides, by doing this in the default theme, I explicitly do not rip off Slack's UI. For once I can't be accused of making TL look like Slack :D

That said, I understand where you're coming from, and I have divided the spacing by 2 (10px -> 5px) like @MaxLeiter suggested.

Pushed my latest changes and rebased on master. Good enough for a first version? 🙏

astorije:
Good catch from @williamboman, I have fixed this: https://github.com/thelounge/thelounge/pull/2527#issuecomment-397443666 (and rebased on `master`)

PR #2316, opened 2018-03-30
Sync on both load and reconnect

### LINKED ISSUE #2278: Custom CSS only syncs across clients when they're open
* *Node version: 6.11.3* 
* *Browser version: many* 
* *Device, operating system: many* 
* *The Lounge version: 453eab3* 

---

Have one client open and change your custom CSS, then open another (like when you come home from work), your changes won't be there. 

Edit custom CSS while another client is open though and it gets the changes.

You'd expect it to work even when changes were made while other clients were closed.

creesch:
Fixes #2278

Turns out that I forgot to change `settings:get` to `setting:get` so settings never got synced on startup. Also made it so that the client will now sync settings when reconnecting.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -303,5 +303,5 @@ function initialize() {
 	// Local init is done, let's sync
 	// We always ask for synced settings even if it is disabled.
 	// Settings can be mandatory to sync and it is used to determine sync base state.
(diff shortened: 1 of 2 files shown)


creesch:
Closing this in favor of #2317

Issue #2279, opened 2018-03-22
Enable settings sync by default if localStorage is empty


xPaw:
Would be a good transition, it's also wanted now that logging out clears out localStorage.

McInkay:
In reply to @astorije 

> I don't think I have mentioned this in the past (though I do have poor memory lol). 

Maybe I imagined it? I thought you had said it.

> One thing I'm wondering though: if we enable it by default, does it stay an "advanced setting"? I think so, but again, just clarifying the consequences/details.

It needs to be an advanced setting even more so when it's on by default.

> That I disagree with, however. We've discussed in the past that some settings should always be synced, because that's how they work on the server, such as custom highlights. Not having them synced simply means something is broken.

Yeah, but the problem is some settings always being synced is confusing. How do we make it clear to users that whether this setting is on or not, this one will be synced? My thinking of turning this on by default, we can say "Odd things might happen" if you disable it, around features like that.

> I actually don't think it will (or should) make it simpler to enable settings-sync by default than not. I think overall it is just 2 different options, for convenience/preference, but I don't foresee one or the other being easier to support.

Possibly, but I think it makes more sense, logically, to be enabled by default.

---

In reply to @creesch 

> This might be a bit more complicated than it seems on the surface. The first client that enables sync becomes the "master" and all settings for that client will be synced back to the server as the base set. Any client after that when sync is enabled will receive these settings.

Maybe we can solve this by allowing the user to set any future client overriding the server, rather than just taking the server settings by default.

> So as far as transition goes you might not want to do this in 3.0.0 where sync is introduced for the first time 

Actually, I think now is the best time to do it, because this is a short term issue. 

> A third option would be to implement a "force sync" sort of button that will sync all settings from that specific client as the new master going forward.

Oh wait, just saw you edited this into your comment. I'm good with this option. Combine that with enabling sync for all new clients, and soon everyone will default syncing.

astorije:
> Yeah, but the problem is some settings always being synced is confusing. How do we make it clear to users that whether this setting is on or not, this one will be synced? My thinking of turning this on by default, we can say "Odd things might happen" if you disable it, around features like that.

I can think of a couple ways to improve the UI to make it clear that some parameter is always synced across devices, so I'm not concerned about that.
In a way, we already have some settings in the page that don't act like the other ones: resetting password and active sessions. I know it's different from what we are talking about, just pointing out that these exist, and we can live with it, and also build our UI to make this easy to work with.

[...]

PR #2317, opened 2018-03-30
Enable sync on empty localstorage, force sync, sync on both load and reconnect.

### LINKED ISSUE #2278: Custom CSS only syncs across clients when they're open

creesch:
This fixes #2278

Turns out that I forgot to change `settings:get` to `setting:get` so settings never got synced on startup. Also made it so that the client will now sync settings when reconnecting. 

This also implements #2279 with the force sync button as discussed in that issue.

### DIFF
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -49,11 +49,16 @@ const noSync = ["syncSettings"];
 const alwaysSync = [];
 
 // Process usersettings from localstorage.
-let userSettings = JSON.parse(storage.get("settings")) || {};
+let userSettings = JSON.parse(storage.get("settings")) || false;
 
-for (const key in settings) {
-	if (userSettings[key] !== undefined) {
-		settings[key] = userSettings[key];
+if (!userSettings) {
+	// Enable sync by default if there are no user defined settings.
+	settings.syncSettings = true;
+} else {
+	for (const key in settings) {
+		if (userSettings[key] !== undefined) {
+			settings[key] = userSettings[key];
+		}
 	}
 }
 
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -319,5 +319,5 @@ function initialize() {
 	// Local init is done, let's sync
 	// We always ask for synced settings even if it is disabled.
 	// Settings can be mandatory to sync and it is used to determine sync base state.
-	socket.emit("settings:get");
+	socket.emit("setting:get");
 }
--- a/client/js/options.js
+++ b/client/js/options.js
@@ -312,7 +312,7 @@ function initialize() {
 		}
 	});
 
-	$settings.on("click", "#forceSync", () => {
+	$settings.find("#forceSync").on("click", () => {
 		syncAllSettings(true);
 	});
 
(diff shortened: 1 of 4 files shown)


client/js/options.js
```diff
@@ -300,8 +312,12 @@ function initialize() {
 		}
 	});
 
+	$settings.on("click", "#forceSync", () => {
```
xPaw: Should this enable sync?

creesch: I don't think so, now it has a single clear purpose. I personally wouldn't expect such a button to also enable sync.

And there might be use cases where people just want to force sync the settings of one client without enabling sync.

I also don't really expect buttons in an interface with check boxes to enable anything (Push notifications being a button seems odd to me for example). 

I thought about doing two checkboxes. One for enabling sync and one for force syncing before enabling sync but that seems confusing and easier to overlook than a big green button saying you can force sync. 

In short, this is the cleanest clearest solution in my opinion.

xPaw: Okay this is fine by me, I'll let other decide on this.

However can you change this line to read `$settings.find("#forceSync").on("click", () => {`

creesch: Done

client/views/windows/settings.tpl
```diff
@@ -28,6 +28,10 @@
 			</label>
 			<p class="sync-warning-override"><strong>Warning</strong> Checking this box will override the settings of this client with those stored on the server.</p>
 			<p class="sync-warning-base"><strong>Warning</strong> No settings have been synced before. Enabling this will sync all settings of this client as the base for other clients.</p>
+			<label class="opt force-sync-button">
+				<button type="button" class="btn" id="forceSync">Force sync settings</button><br>
```
astorije: I feel like this is more commonly called "Set/Mark as default", but it's not perfect either. I just find "Force sync settings" not ideal.

I'm overall very confused by the wording of the feature, but I'm not sure how to make it clearer just now. Worth waiting for the feature to stabilize and see if we can improve.

screen shot 2018-04-02 at 02 22 24

Good thing it's an advanced feature :D

astorije:
LGTM, but I know @McInkay has been pretty involved in this feature so it would be fair for him to give his opinion/approval before this makes it in :)

Thanks a lot for this, @creesch!

client/views/windows/settings.tpl
creesch: I agree the wording isn't ideal but as you pointed out neither is other wording. And yeah this is a proper advanced setting only used for those people that already have settings they are particularly attached to.

McInkay:
@astorije Do you want to review this anymore or are you happy for it to go in without you?

astorije:
Nah, I trust both of you on this, and my UI/UX-hat-wearing concern I raised is shared by both you and @creesch so I know it's a "we're not sure how to do better for now" and not a "this is actually what we want", so we can iterate on that when we have better ideas.

astorije:
Thanks a lot @creesch!

PR #2325, opened 2018-04-01
Fix server side rendering for href of initial theme

Raqbit:
The index.html template renders the initial theme's href incorrectly, making the browser try to load `/themes/themes/example.css.css`. 
This is because the `theme` variable already includes the `/themes/` prefix and `.css` suffix, which the template then adds again.


Raqbit:
Thought I'd make a quick fix but it seems like something else is happening here, closing.

astorije:
@Raqbit, are you sure your config is correct, i.e. says `theme: "mytheme"` and not `theme: "mytheme.css"`?

PR #2326, opened 2018-04-01
Replace "." character with random digit in nick param

lol768:
Feature parity: iris (https://github.com/atheme-legacy/iris/commit/c747578046dcf9925a4353c7cf9740c51367a7f0)

Use-case: Support channels, ensuring minimal chance of nick collision.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -414,6 +414,10 @@ $(function() {
 				params.join = params.channels;
 			}
 
+			if ("nick" in params) {
+				params.nick = params.nick.replace(/\./g, () => "" + Math.floor(Math.random() * Math.floor(10)));
+			}
+
 			// Possible parameters:  name, host, port, password, tls, nick, username, realname, join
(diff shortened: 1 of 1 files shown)


astorije:
@xPaw recently did something related to this for https://github.com/thelounge/thelounge/pull/2090. I don't know if we want it to work on URL params as well (I don't have an opinion), but if we do, we probably want to match the same format (`%`) and ideally reuse the same code.

@xPaw, what's your take on this?

astorije:
> it's unexpected it only works server-side

I'm not sure I agree with that. The randomization is indeed very much expected on the server-side because it's documented as such, but on the client, it's not obvious that any character entered in the URL will be randomized.
Thinking about it, I would probably be surprised to see something different in the nick input from what I typed in the URL.

I don't have a string opinion here, so I'll defer to @xPaw here. I'll be fine with whatever he prefers.

xPaw:
I agree with @astorije.

xPaw:
We have decided to keep the current behaviour. Sorry for the trouble.

PR #2328, opened 2018-04-02
Astorije/local themes

astorije:
So this was a long time coming, but... fixes https://github.com/thelounge/thelounge/issues/846.

This PR does a few things and has a few requirements:

## Zenburn and Crypto

This PR removes Zenburn and Crypto from this repository.
My goal is to only keep 2 themes in the core codebase: the default theme (currently Example) and a dark version of it (Morning).

We have a history of focusing on testing/applying changes to the default theme, and breaking other ones. By keeping only those 2 in the core, making them as close-looking as possible (besides the light/dark background), it will be easier to keep them fully working across PRs.

This doesn't mean Zenburn and Crypto are simply deleted however. They will live in their own repositories / npm packages. A huge benefit for us is that we can have different, dedicated maintainers for them!
If no one claims Crypto, it will become abandonware, but having it in a separate repo means a new contributor/maintainer could give it a reboot, removing all those things that make it undesired.

It will overall make for more stable default themes, and with proper maintenance, Zenburn will have a better life as well.

**TODO:** Create a repository / npm package for Crypto and Zenburn

## Renaming Example

We have outlived the keyword "Example", way outlived. I have renamed it to "Default" for now because it's more descriptive than "Example", but I am open to find it a better name to match the fanciness of "Morning". Suggestions welcome.
We could also go for boring "Default (light)" and "Default (dark)", but that would require a bit more changes in the code.

This seems like a silly change, but bumping major version is a decent moment to do so (remember, we didn't have good theme management until recently), and "Example" does not reflect what the default theme of The Lounge is anymore ([it is legacy, in fact](https://github.com/erming/shout/blob/master/client/themes/example.css)).

## Switching client to default theme when selected theme does not exist

This is something missing from `master` that we would have needed anyway, but the changes in this PR make it mandatory. At the moment, if an admin removes a theme, it will half-break the UI, by picking up no theme and getting 404s for requesting an invalid theme.

This fixes it by switching back to the default theme for that server when it happens.


client/js/options.js
```diff
@@ -280,9 +280,9 @@ function initialize() {
 		$warningUnsupported.show();
 	}
 
-	$settings.on("change", "input, select, textarea", function(e) {
-		// We only want to trigger on human triggerd changes.
-		if (e.originalEvent) {
+	$settings.on("change", "input, select, textarea", function(e, {force} = {}) {
+		// We only want to trigger on human triggered changes.
+		if (e.originalEvent || force) {
```
xPaw: I don't think check is needed anymore as options no longer trigger a fake `change`. Right @creesch?

creesch: Not entirely, synced settings will still be applied to their respective input fields which might trigger a fake change. So it probably needs to stay just for that.

client/js/socket-events/configuration.js
```diff
@@ -25,6 +25,12 @@ socket.on("configuration", function(data) {
 	options.initialize();
 	webpush.initialize();
 
+	// If localStorage contains a theme that does not exist on this server, switch
+	// back to its default theme.
+	if (!data.themes.map((t) => t.name).includes(options.settings.theme)) {
+		$("#theme-select").val(data.defaultTheme).trigger("change", {force: true});
```
creesch: This does not belong here but in options.js I think

client/js/options.js
creesch: However, force is not needed at all as far as I can tell. @astorije what are you attempting to do here? This seems like moving away again from the settings refactoring I did, you should use `processSetting()` if you want to update a setting.

client/js/socket-events/configuration.js
astorije: That's awesome, thanks for the feedback. Definitely better than my stupid `force` stuff if I can make it work, I simply missed that capability 😅

client/js/options.js
astorije: Done!

client/js/socket-events/configuration.js
astorije: That worked well, thanks!

@creesch, totally outside of this PR, but just a thought: what would you think about changing `updateSetting`'s signature from `updateSetting(name, value, sync)` to updateSetting(name, value, {sync})`?
That would allow:
```js
options.processSetting("theme", data.defaultTheme, {sync: true});
```

It's a silly change so no big deal, it's just that boolean arguments can be confusing and sometimes error-prone. Definitely more of a style thing, I was just curious about what your opinion on this was.

creesch: Uh yeah that would be fine I guess. 
Though I am not sure how that would apply to `processSetting()` as the boolean there is `save` ;) 

Looking at it `processSetting()` should probably be restructured a little bit with an options parameter instead so it can be called like this. 

```javascript
processSetting(name, value, {
    save: false,
    sync: false
})
```

Making the declaration something like this 

```javascript
processSetting(name, value, {save, sync}) { 
    // code
}
```

Because the way it is currently written it will not sync the theme value.  

I'll work that all in after this gets merged. Also I'll leave a bit more comments to make it clear to people how options function.

astorije: Sounds good, thanks for your efforts @creesch!

astorije:
@creesch, could you confirm this PR is in better shape regarding your recent rewrite? Or is there anything else that looks suspicious to you?

creesch:
Looks good to me.

PR #2336, opened 2018-04-06
Split some CSS into separate files

xPaw:


astorije:
Mmmh, you probably should have waited for #2315 to be merged before doing that, rebasing is going to be quite expensive :/

xPaw:
Is it? I don't think I moved any CSS that you change in that PR.

astorije:
You're right, it looks like you've been careful :D
Cool, I'll make sure to avoid touching parts this PR changes until it's merged if I make other UI changes then.

PR #2337, opened 2018-04-06
Remove nick editor

xPaw:
Depends on #2229

### DIFF
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -19,8 +19,6 @@ module.exports = {
 	hasRoleInChannel,
 	move,
 	resetHeight,
-	setNick,
-	toggleNickEditor,
 	toggleNotificationMarkers,
 	requestIdleCallback,
 	togglePreviewMoreButtonsIfNeeded,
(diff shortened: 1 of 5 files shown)


astorije:
Lets remove the nick entirely here, like shown [here](https://user-images.githubusercontent.com/113730/38126670-0acb853a-33c0-11e8-94f7-cc2657acc59c.png) or [your Solarized theme](https://github.com/thelounge/thelounge-theme-solarized/blob/master/screenshot.png). Removing it was part of my design update and I still think we should do that.

xPaw:
I didn't remove it entirely because it's probably beneficial in some cases (e.g. public mode). We could hide it by default, and make it visible in public or those that want to show it with custom css.

astorije:
Sure, let's do that.

client/js/utils.js
```diff
@@ -83,17 +82,8 @@ function join(args) {
...
-	$("#nick-value").prop("contenteditable", toggle);
-}
-
 function setNick(nick) {
-	// Closes the nick editor when canceling, changing channel, or when a nick
-	// is set in a different tab / browser / device.
-	toggleNickEditor(false);
-
-	$("#nick-value").text(nick);
+	$("#nick").text(nick);
```
astorije: I believe this 1-line function is now called in 2 places left, and I don't think we'll have any further case. Move its content to where it's called instead?

xPaw:
Yes.

PR #2340, opened 2018-04-08
Prevent "undefined" OS version in session list

dgw:
Noticed this in my session list after spinning up a new Ubuntu machine & logging in to The Lounge from it. Here's a quick little PR to make the ugly `undefined` go away.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -316,7 +316,11 @@ Client.prototype.updateSession = function(token, ip, request) {
 	}
 
 	if (agent.os.name) {
-		friendlyAgent += ` on ${agent.os.name} ${agent.os.version}`;
+		friendlyAgent += ` on ${agent.os.name}`;
+
+		if (agent.os.version) {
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -316,7 +316,11 @@ Client.prototype.updateSession = function(token, ip, request) {
 	}
 
 	if (agent.os.name) {
-		friendlyAgent += ` on ${agent.os.name} ${agent.os.version}`;
+		friendlyAgent += ` on ${agent.os.name}`;
+	}
+
+	if (agent.os.version) {
```
xPaw: Should this be inside of the previous `if`?

dgw: Oh, probably.

Coding at 25 hours awake problems... I'll go to sleep right after fixing that. :sleeping:

astorije:
Thanks, good catch @dgw! 🙏

astorije:
Hey @dgw, we have sticker packs for our contributors now!

PR #2349, opened 2018-04-12
Fix previews on Morning (and others)

astorije:
Follow-up of #2272.

Well, it _had_ to happen, didn't it...

Before | After
--- | ---
screen shot 2018-04-12 at 00 43 58 | screen shot 2018-04-12 at 00 51 16

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -2,6 +2,11 @@
 @import "bootstrap.css";
 @import "../../node_modules/primer-tooltips/build/build.css";
 
+:root {
+	/* Links and link-looking buttons */
+	--link-color: #50a656;
+}
+
 html,
 body {
 	height: 100%;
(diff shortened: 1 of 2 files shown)


client/css/style.css
```diff
@@ -2,6 +2,11 @@
 @import "bootstrap.css";
 @import "../../node_modules/primer-tooltips/build/build.css";
 
+:root {
```
xPaw: put this in a separate constants file and import it maybe?

astorije: Good idea, I'll do that as part of #2315 though.

I don't know about other editors, but vars integration in Atom is _really_ cool! I'm wondering how good it stays when putting vars in a different files (I know this isn't TL's problem at all if it doesn't).

PR #2355, opened 2018-04-13
Add 'op' and 'voice' user context menu items

### LINKED ISSUE #1081: OP/Voice options in user context menu
Request the ability to choose +/- Operator and +/- Voice statuses in user context menu for current channel assuming user has sufficient privileges.

Issue is closely related to Issue #347

Jay2k1:
As a first step to adding more commands to the context menu, this adds the probably most common modes +o (op) and +v (voice) when you have +o (similar to the kick entry).

image

If the target is op, it shows "Revoke op (-o)", same for voice. All these only appear when you're op yourself.

I discussed with @McInkay and @xPaw and we concluded that in the future, 

> what we ideally want is entries on the context menu for each of the ones in PREFIX (with the name from the handlebars plugin) up to your own priviledge, except voice gets none

and also

> 17:18:44 <+McInkay> I'd be ok with the first version being "show op and voice if you are an op" and nothing else

So here goes. 

Also fixes #1081.

### DIFF
--- a/client/js/contextMenuFactory.js
+++ b/client/js/contextMenuFactory.js
@@ -125,6 +125,82 @@ function addKickItem() {
 	});
 }
 
+function addOpItem() {
+	function op(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/op " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) =>
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"]) &&
+			!utils.hasRoleInChannel(target.closest(".chan"), ["op"], target.data("name")),
+		className: "action-op",
+		displayName: "Give operator (+o)",
+		data: (target) => target.data("name"),
+		callback: op,
+	});
+}
+
+function addDeopItem() {
+	function deop(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/deop " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) =>
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"]) &&
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"], target.data("name")),
+		className: "action-op",
+		displayName: "Revoke operator (-o)",
+		data: (target) => target.data("name"),
+		callback: deop,
+	});
+}
+
+function addVoiceItem() {
+	function voice(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/voice " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) =>
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"]) &&
+			!utils.hasRoleInChannel(target.closest(".chan"), ["voice"], target.data("name")),
+		className: "action-voice",
+		displayName: "Give voice (+v)",
+		data: (target) => target.data("name"),
+		callback: voice,
+	});
+}
+
+function addDevoiceItem() {
+	function devoice(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/devoice " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) =>
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"]) &&
+			utils.hasRoleInChannel(target.closest(".chan"), ["voice"], target.data("name")),
+		className: "action-voice",
+		displayName: "Revoke voice (-v)",
+		data: (target) => target.data("name"),
+		callback: devoice,
+	});
+}
+
 function addFocusItem() {
 	function focusChan(itemData) {
 		$(`.networks .chan[data-target="${itemData}"]`).click();
@@ -206,6 +282,10 @@ function addDefaultItems() {
 	addWhoisItem();
 	addQueryItem();
 	addKickItem();
+	addOpItem();
+	addDeopItem();
+	addVoiceItem();
+	addDevoiceItem();
 	addFocusItem();
 	addChannelListItem();
 	addBanListItem();
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -42,16 +42,16 @@ function resetHeight(element) {
 	element.style.height = element.style.minHeight;
 }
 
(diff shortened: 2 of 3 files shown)


client/js/utils.js
```diff
@@ -42,16 +42,16 @@ function resetHeight(element) {
...
+function hasRoleInChannel(channel, roles, nick) {
 	if (!channel || !roles) {
 		return false;
 	}
 
 	const channelID = channel.data("id");
 	const network = $("#sidebar .network").has(`.chan[data-id="${channelID}"]`);
-	const ownNick = network.data("nick");
-	const user = channel.find(`.names-original .user[data-name="${escape(ownNick)}"]`).first();
+	const target = (nick && nick || network.data("nick"));
```
xPaw: This can be simplified to `nick || network.data("nick")`

client/js/contextMenuFactory.js
```diff
@@ -125,6 +125,74 @@ function addKickItem() {
...
+			target: $("#chat").data("id"),
+			text: "/op " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => utils.hasRoleInChannel(target.closest(".chan"), ["op"]) && target.closest(".chan").data("type") === "channel",
+		className: "action-op",
+		displayName: "Give op (+o)",
+		data: (target) => target.data("name"),
```
xPaw: Change `op` to `operator`

client/js/contextMenuFactory.js
```diff
@@ -125,6 +125,74 @@ function addKickItem() {
...
+function addOpItem() {
+	function op(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/op " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => utils.hasRoleInChannel(target.closest(".chan"), ["op"]) && target.closest(".chan").data("type") === "channel",
```
xPaw: Remove all these channel checks, they're unnecessary.

client/js/contextMenuFactory.js
```diff
@@ -125,6 +125,74 @@ function addKickItem() {
...
+function addDeopItem() {
+	function deop(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/deop " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => utils.hasRoleInChannel(target.closest(".chan"), ["op"]) && utils.hasRoleInChannel(target.closest(".chan"), ["op"], target.data("name")) && target.closest(".chan").data("type") === "channel",
```
xPaw: Move `target.closest(".chan")` to a variable so you're not calling it twice. Same in devoice.

client/js/contextMenuFactory.js
xPaw: Also did you forget to add a check if user doesn't have op already?

client/js/contextMenuFactory.js
```diff
@@ -125,6 +125,74 @@ function addKickItem() {
...
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/op " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => utils.hasRoleInChannel(target.closest(".chan"), ["op"]) && utils.hasRoleInChannel(target.closest(".chan"), ["op"], target.data("name")) === false,
+		className: "action-op",
+		displayName: "Give op (+o)",
```
astorije: I'm really not a fan of "op". We use "Operators" in the user list menu and I find it more appropriate.
You'll need to change style.css L1896 from 160px to 180px.

client/js/contextMenuFactory.js
```diff
@@ -125,6 +125,74 @@ function addKickItem() {
...
+function addOpItem() {
+	function op(itemData) {
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/op " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) => utils.hasRoleInChannel(target.closest(".chan"), ["op"]) && utils.hasRoleInChannel(target.closest(".chan"), ["op"], target.data("name")) === false,
```
astorije: Can you break this line and others below?

I'll post a diff, probably simpler for you.

client/js/contextMenuFactory.js
```diff
@@ -125,6 +125,82 @@ function addKickItem() {
...
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/op " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) =>
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"]) &&
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"], target.data("name")) === false,
```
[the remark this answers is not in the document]
Jay2k1: Apparently it does. I had tried that unsuccessfully before, but I guess that was during a phase of trial-and-error where something else was buggy. I'll change it

astorije:
Thanks a lot for this and congrats on getting your first PR in, @Jay2k1! 🎉

client/js/contextMenuFactory.js
```diff
@@ -125,6 +125,82 @@ function addKickItem() {
...
+		socket.emit("input", {
+			target: $("#chat").data("id"),
+			text: "/voice " + itemData,
+		});
+	}
+
+	addContextMenuItem({
+		check: (target) =>
+			utils.hasRoleInChannel(target.closest(".chan"), ["op"]) &&
+			!utils.hasRoleInChannel(target.closest(".chan"), ["voice"], target.data("name")),
```
McInkay: I think this will show "Give voice" even if the user is already op. I know that technically users can have both of the user modes, that's how it works, but it seems a bit odd to have it on the UI. The use cases for having both are very small and quite niche.

Thoughts? I don't think it's big enough to delay the PR, but I just thought I'd point it out.

xPaw: The way it shows is correct. Server is aware of all the modes (due to multi mode cap) but they are not propagated to the client.

astorije:
Merging since @xPaw answered you, @McInkay. We can revisit before v3 if you feel strongly, in which case make sure to open an issue (or a PR!) so we don't forget.

@Jay2k1, thanks so much for your help and patience. Congrats again on your first PR!

lzrdking71:
@ Jay2k1 Would it be possible to have this menu also available for half ops, admins, and owners?

PR #2361, opened 2018-04-15
Add project version in outputs that mention source SHA

astorije:
Example:

```sh
$ thelounge --version
source (1d6a120d / v3.0.0-pre.5)
```

Before | After
--- | ---
screen shot 2018-04-15 at 23 52 57 | screen shot 2018-04-15 at 23 53 06
screen shot 2018-04-15 at 23 56 04 | screen shot 2018-04-15 at 23 56 14

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -62,6 +62,11 @@ function getGitCommit() {
 		return _gitCommit;
 	}
 
+	if (!fs.existsSync(path.resolve(__dirname, "..", ".git", "HEAD"))) {
+		_gitCommit = null;
+		return null;
+	}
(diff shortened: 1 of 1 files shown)


astorije:
So it turns out AppVeyor doesn't get the commit version (meaning [this](https://github.com/thelounge/thelounge/blob/astorije/ui/src/helper.js#L64-L67) most likely throws). @xPaw, do you have an idea of what that could be?

screen shot 2018-04-16 at 15 33 06

xPaw:
`git` is probably not available in PATH.

PR #2366, opened 2018-04-17
Refactor userLog to be the same as sqlite logger

### LINKED ISSUE #1392: Duplicate log entries when using two nicks

xPaw:
- [x] Fixes #1392 - Networks now use GUID as folder name
- [x] Fixes log writing failing due to invalid stuff in network or channel names (mostly on Windows).
- [x] Depends on first message store to be sqlite, need to fix that

This doesn't introduce an API for registering storage and whatnot, I intentionally left this out of the PR.


xPaw:
@astorije Hopefully.

astorije:
Alright, I'll assign you to the PR until you do it, feel free to assign to us when you're done.

xPaw:
Got it done.

src/plugins/messageStorage/sqlite.js
```diff
@@ -139,6 +139,10 @@ class MessageStorage {
 			));
 		});
 	}
+
+	canProvideMessages() {
```
astorije: What does `canProvideMessages` mean?

xPaw: It means this storage can load stored messages back into memory.

xPaw:
GUIDs were specifically created to solve an issue where you connected to the same network twice it still logged separately. So deleting a network and creating a new one will generate a new guid, which is by design.

I don't think folder names are such a big issue as your channels inside are more than likely to be unique, so it if you're searching your logs it should be easy.

src/plugins/messageStorage/sqlite.js
astorije: Oh I see. Kind of a bikeshed but maybe we need better wording for this. `supportsReload`? `supportsRead`?
Or instead, maybe only if `getMessages` is available and `isEnabled` is true or something?

astorije:
> GUIDs were specifically created to solve an issue where you connected to the same network twice it still logged separately. So deleting a network and creating a new one will generate a new guid, which is by design.

That makes sense, but I think there will be a use case for reusing an existing network data. Just like closing/re-opening a channel reloads the messages, removing/re-adding a network could do something analogous. Again, not as part of your PR obviously, but something that would be nice.

> I don't think folder names are such a big issue as your channels inside are more than likely to be unique, so it if you're searching your logs it should be easy.

I strongly disagree. `logs` folder now contains folders with random names similar to `$THELOUNGE_HOME/storage`, `.git/objects`, or some of `/tmp`. These logs are meant for human consumption (as opposed to, say, `.git/objects` which is just internal storage for git) so they should be easily browsable by humans.
I can try to do what I suggested if you don't want to spend the time 😅

McInkay:
Regarding folder name, how about "networkname-guid"

That way it's clear to you if you have unique network names which is the correct one, but we are guaranteed the folders will be unique

xPaw:
What about the case where you edit your network and change the name or address? Your UUID will stay the same, but folder name will change if you include these.

McInkay:
Would it be too silly to change the name of the folder that already exists if we change the name? Meaning that there's the 1 folder, rather than 2, and the name is the new one. I feel like maybe it is silly.

xPaw:
Yes, I'd rather not touch that.

dgw:
Are the user-set network names not enforced to be unique? I feel like they should be. One shouldn't be able to confuse oneself by creating two different networks with the same name, IMO. And the user should never have to be confronted by even part of a UUID. The human-readable network name should be all they need to know about.

That said, I'm well aware that this is getting into bikeshedding territory. Perhaps I should just make a new issue proposing that network names be made unique instead?

McInkay:
> Are the user-set network names not enforced to be unique? I feel like they should be. One shouldn't be able to confuse oneself by creating two different networks with the same name, IMO. 

They are not. We have discussed this and, in principle, we all agree that it would be good. However, the problem is trying to fix it when there may well be people who already have non-unique names.

This also doesn't solve the issue of the name changing and therefore the folder changing.

astorije:
> Are the user-set network names not enforced to be unique?

I think we all agree on that yeah, but indeed that's a different issue. If anything because the folder name is the URL and not the network name. Maybe we could also enforce uniqueness of URL, like if you already have a irc.freenode.net then you cannot connect to a second one, but that becomes very restrictive and I'm sure most people will be against that.

> Regarding folder name, how about `networkname-guid`

Yep that's what I suggested, except I meant networkUrl-shortUUID where we only pick the first 8 chars of the UUID for human-friendliness (à la git short SHA). That's 200bn+ combinations, we won't collide, ever :D

> What about the case where you edit your network and change the name or address? Your UUID will stay the same, but folder name will change if you include these.

In the several years I have used IRC, I have had to change an address only once (outside of TL testing, I mean), and that was a very specific scenario, so really I would not worry about this. However I very often look for things in the text logs.
I'm not worried about having a new folder created when changing the address.

@xPaw, like I said, happy to make the change myself if you can't be bothered :)

xPaw:
Ok I pushed network name into the folder name. I'm basically cutting length of the network name from guid so that all folder names remain 36 characters total.

astorije:
> Nope, bad idea. One example would be ZNC and connecting to other networks through it.

Yeah, I figured there would be tons of edge cases like this. I was not convinced myself when typing out lol, thanks for clarifying.

> Ok I pushed network name into the folder name.

Any reason why we're switching from host to name? Is it for the same reason as the ZNC case? I was in favor of host at first, but mostly out of habit, I don't have much of a preference anymore.

> I'm basically cutting length of the network name from guid so that all folder names remain 36 characters total.

Naively, I would have done simpler, such as name + N chars from UUID. Out of curiosity, did you see edge cases that made this way desirable instead?

xPaw:
> Any reason why we're switching from host to name? Is it for the same reason as the ZNC case? 

ZNC is one, and you've talked about making names unique so that also aligns with this.

> Out of curiosity, did you see edge cases that made this way desirable instead?

Sounded like a great idea in my head last night.

src/plugins/messageStorage/text.js
```diff
@@ -0,0 +1,99 @@
...
+}
+
+module.exports = TextFileMessageStorage;
+
+function cleanFilename(name) {
+	name = filenamify(name, {replacement: "_"});
+	name = name.toLowerCase();
+
+	return `${name}.log`;
+}
```
astorije: Any chance you could write tests for this file? (It seems side-effect heavy so it's likely you can't 🤷‍♂️)

PR #2377, opened 2018-04-22
Make context menus accessible with keyboard

xPaw:
- [x] Each context menu item is focusable
- [x] Each context menu can be opened with enter
- [x] Can navigate menu with up/down arrows
- [ ] Context menu events must be bound only once (currently they are bound every time context menu opens, bug on master too)
- [ ] When escaping context menu opening in user list, return focus to the search field


xPaw:
@astorije I left them as is precisely because of #1873.

astorije:
@xPaw, they're really ugly here though. I'd rather remove the outline for now, since there is a background color already. 🙏

PR #2378, opened 2018-04-22
Add prefix to channels from connect window

### LINKED ISSUE #2327: Nicks entered in the "join" field of Connect form open channel-looking windows
* *The Lounge version:* Confirmed on v3.0.0-pre.5

---

When entering a nick in the join field of connect form:

screen shot 2018-04-01 at 19 09 07

or when using the URL params such as https://demo.thelounge.chat/?join=astorije (please don't spam me 😅), a window correctly open, but it has the layout of a channel:

screen shot 2018-04-01 at 18 56 14

It all works fine, and it's actually useful for something like support, or contact tool embedded in personal websites, but we should be able to display a query layout instead.

### LINKED ISSUE #2424: Query params no longer working for prefilling fields in public mode
* *Node version:* 8.11.1
* *Browser version:* Firefox 61.0a1
* *Device, operating system:*  Linux, Ubuntu
* *The Lounge version:* 4a9c3236

---

Query params no longer seem to work on this revision, the fields aren't populated with the data from the search string.

To reproduce: Visit http://localhost:9000/?nick=foo and observe it still uses the defaults from the config file.

Is this a known issue?

xPaw:
Fixes #2327.
Fixes #2424.

I changed the regex to known chan types because stuff like `[test]` wouldn't match otherwise.

Also fixed 2 other bugs:
- Fix connect window opening twice in public mode
- Override server provided defaults with parameters passed in the URL if they match the data type (this is currently broken on master)

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -361,14 +361,13 @@ $(function() {
 					let value = params[key];
 
 					if (key === "join") {
-						const channels = value.split(",");
-						value = channels.map((c) => {
-							if (c.match(/^\w/)) {
-								return "#" + c;
+						value = value.split(",").map((chan) => {
+							if (!chan.match(/^[#&!+]/)) {
+								return `#${chan}`;
 							}
 
-							return c;
-						}).join(",");
+							return chan;
+						}).join(", ");
 					}
 
 					// \W searches for non-word characters
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -4,7 +4,6 @@
 require("jquery-ui/ui/widgets/sortable");
 const $ = require("jquery");
 const moment = require("moment");
-const URI = require("urijs");
 
 // our libraries
 require("./libs/jquery/stickyscroll");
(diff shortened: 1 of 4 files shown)


client/js/lounge.js
```diff
@@ -361,14 +361,13 @@ $(function() {
 					let value = params[key];
 
 					if (key === "join") {
-						const channels = value.split(",");
-						value = channels.map((c) => {
-							if (c.match(/^\w/)) {
-								return "#" + c;
+						value = value.split(",").map((chan) => {
+							if (!chan.match(/^[#&!+]/)) {
+								return `#${chan}`;
```
astorije: I actually like the idea that we could open a given query window from the `join` URL param.

I have seen a few personal websites where people set up a web chat that opens a query window with them (the website author) but that's just one use case. I can see how it could be useful to have default channels _and_ queries open by default.

Any chance we could keep that behavior? That was actually what #2327 was about rather than convert everything to a channel.

xPaw: That would break #2289.

astorije: I guess it does, I didn't think of that at the time of that PR, which I now think we shouldn't have done.

I really don't want to get rid of the capability to auto-open queries though. I can understand how it's a pain to add # in the URL (still would prefer to open a query though), but in the join field of the connect window, I really think we shouldn't assume that `foo` actually means `#foo`. If no `#` is specified, it should open a query IMO.

xPaw: I think better over all to have a consistent behaviour where client and server agree, in this case to prefix the channel.

Honestly, I don't see much of a use to support opening queries there, the field is called `Channels`.

client/js/socket-events/configuration.js
```diff
@@ -84,4 +84,36 @@ socket.on("configuration", function(data) {
 				$(this).data("lastvalue", nick);
 			});
 	});
+
+	if ($(document.body).hasClass("public")) {
+		const params = new URLSearchParams(document.location.search);
```
astorije: According to caniuse, this is not available in Edge 16, are we okay with this?

client/js/lounge.js
astorije: Alright then. I do see a use for it (I even gave an actual use case for it, that I have actually seen out there), but I'm not going to push for it. We'll see if people create tickets for it :D

client/js/socket-events/configuration.js
xPaw: Hm. Edge 17 is released with the next Windows update which should be this month.

astorije: That feels a bit too bleeding-edge, no? As in, `master` will be broken on Edge until the next release, we're not at "last 2 versions" at all anymore lol.
Is there another mechanism we can use for now?

xPaw: OK, I switched back to URIjs.

client/js/socket-events/init.js
```diff
@@ -23,10 +23,6 @@ socket.on("init", function(data) {
 
 	if (data.networks.length === 0) {
 		sidebar.find(".empty").show();
-
-		$("#footer").find(".connect").trigger("click", {
-			pushState: false,
-		});
```
astorije: Mind explaining why you could remove this? I'm not seeing it...

xPaw: Same code gets called by the code that opens correct window.

PR #2380, opened 2018-04-22
Add `upgrade` CLI command for themes and packages

jake-walker:
> I apologise if I was supposed to create an issue but I'd already made it by the time I had seen that in the contributing guidelines.

Earlier on today I needed to update some of my themes and I thought it was quite a long winded way to do `thelounge install thelounge-theme-mytheme` for each theme I wanted to update, so I made a `thelounge update` command which updates all of the themes installed.

I've tested it when themes need to be updated, when themes are already up to date and when there are no themes (and no packages folder) and it seems to work fine.

### DIFF
--- a/src/command-line/upgrade.js
+++ b/src/command-line/upgrade.js
@@ -1,0 +1,66 @@
+"use strict";
+
+const colors = require("chalk");
+const program = require("commander");
+const Helper = require("../helper");
+const Utils = require("./utils");
+
+program
+	.command("upgrade [packages...]")
+	.description("Upgrade installed themes and packages to their latest versions.")
+	.on("--help", Utils.extraHelp)
+	.action(function(packages) {
+		const fs = require("fs");
+		const path = require("path");
+
+		// Get paths to the location of packages directory
+		const packagesPath = Helper.getPackagesPath();
+		const packagesConfig = path.join(packagesPath, "package.json");
+		const packagesList = JSON.parse(fs.readFileSync(packagesConfig)).dependencies;
+		const argsList = [
+			"upgrade",
+			"--latest",
+			"--json",
+			"--production",
+			"--ignore-scripts",
+			"--non-interactive",
+			"--cwd",
+			packagesPath,
+		];
+
+		let count = 0;
+
+		// Check if the configuration file exists
+		if (!fs.existsSync(packagesConfig)) {
+			log.warn("There are no packages installed.");
+			return;
+		}
+
+		// If a package names are supplied, check they exist
+		if (packages.length) {
+			log.info("Upgrading the following packages:");
+			packages.forEach((p) => {
+				log.info(`- ${colors.green(p)}`);
+
+				if (packagesList.hasOwnProperty(p)) {
+					argsList.splice(1, 0, p);
+					count++;
+				} else {
+					log.error(`${colors.green(p)} is not installed.`);
+				}
+			});
+		} else {
+			log.info("Upgrading all packages...");
+		}
+
+		if (count === 0 && packages.length) {
+			log.warn("There are not any packages to upgrade.");
+			return;
+		}
+
+		return Utils.executeYarnCommand(...argsList).then(() => {
+			log.info("Package(s) have been successfully upgraded.");
+		}).catch((code) => {
+			throw `Failed to upgrade package(s). Exit code ${code}`;
+		});
+	});
--- a/src/command-line/index.js
+++ b/src/command-line/index.js
@@ -48,6 +48,7 @@ if (!Helper.config.public && !Helper.config.ldap.enable) {
 
 require("./install");
 require("./uninstall");
+require("./upgrade");
 
 // `parse` expects to be passed `process.argv`, but we need to remove to give it
 // a version of `argv` that does not contain options already parsed by


Jay2k1:
Sounds like a great feature especially for people with many themes. My only concern is that the wording might suggest that this command actually updates The Lounge, when in fact it only updates its themes. Maybe thelounge update-themes or something like that would be more adequate? Let's see what the maintainers think about that.

dgw:
> Maybe thelounge update-themes or something like that

`thelounge update-packages`? `thelounge package-update`? Long, but it does update all packages, not just themes (even though AFAIK only themes exist so far, that will change).

McInkay:
I think `update` is fine. If you think about, say, npm, `npm update` you would never expect to update npm itself, but the packages you have installed. You are never going to use `thelounge update` without first using `install` and you don't expect `thelounge install` will install thelounge.

Also, in the future we'll put a front-end to this for admins (once #1950 is implemented) and most people won't need to use the CLI anyway.

jake-walker:
An alternative is to maybe have `install`, `uninstall` and `update` under a parent `package` command so it would be `thelounge package install`, `thelounge package update` etc..., then it would completely remove any confusion.

McInkay:
That would certainly be an acceptable compromise. If we were doing that, I would want to symlink install and uninstall to keep the backwards compatibility. But yeah. Will see what the other maintainers think.

I'm hoping to be able to test this soon, code looks generally good but I need to make sure it works on my end :-P

astorije:
Hey @jake-walker, thanks a lot for your first PR!! ♥️

I would suggest using `upgrade` instead of `update`. At least in my (recent?) experience, the different package managers I use (`apt`, `brew`, `yarn`, and others) tend to use `upgrade` to bump packages and `update` to refresh their cache. Just my 2c.

Regarding confusion between upgrading packages and upgrading TL itself, I have the same feeling about `install`/`uninstall` and `add`/`remove`. This will become more and more an issue as we add commands, for example `thelounge list` currently lists users but it could also be thought of as listing packages.
I agree that namespacing would be nice, or maybe have a `tlpm` CLI dedicated to managing packages :D

It would be great if `thelounge upgrade` could accept an optional argument to pass a package name to upgrade specifically, but it doesn't have to be in this PR if you don't feel up for it.

benharri:
Once #2385 is merged, make sure you use the yarn util function. I second @astorije's suggestion of `upgrade` over `update`. 

Looks awesome! I know I'll use it!

benharri:
@jake-walker can you squash and rebase?

jake-walker:
I think that's it?

benharri:
Nice, thanks! LGTM.

src/command-line/upgrade.js
```diff
@@ -0,0 +1,57 @@
...
+
+		// Check if the configuration file exists
+		if (!fs.existsSync(packagesConfig)) {
+			log.error("There are no packages installed.");
+			return;
+		}
+
+		// If a package name is supplied, check it exists
+		if (cmd.package !== undefined) {
+			if (Object.keys(JSON.parse(fs.readFileSync(packagesConfig)).dependencies).indexOf(cmd.package) < 0) {
```
xPaw: Instead of `Object.keys.indexOf` you can use `hasOwnProperty`

src/command-line/upgrade.js
```diff
@@ -0,0 +1,57 @@
...
+				log.error(`${colors.green(cmd.package)} isn't installed!`);
```
xPaw: `is not installed.`

jake-walker:
Ok, @xPaw I've changed the things you said

src/command-line/upgrade.js
```diff
@@ -0,0 +1,57 @@
...
+program
+	.command("upgrade")
+	.option("-p, --package [name]", "Update only the specified package")
+	.description("Update installed themes and packages to their latest versions. Specify one theme or package to update with the --package argument.")
+	.on("--help", Utils.extraHelp)
+	.action(function(cmd) {
+		const fs = require("fs");
+		const path = require("path");
+
+		log.info(`Updating ${(cmd.package !== undefined ? colors.green(cmd.package) : "packages")}...`);
```
astorije: `Upgrading`?

src/command-line/upgrade.js
```diff
@@ -0,0 +1,57 @@
+"use strict";
+
+const colors = require("chalk");
+const program = require("commander");
+const Helper = require("../helper");
+const Utils = require("./utils");
+
+program
+	.command("upgrade")
+	.option("-p, --package [name]", "Update only the specified package")
```
astorije: `Upgrade`

Also, could it be possible to make the command `thelounge upgrade [package]` (or ``thelounge upgrade [...packages]`)? If `package` is specified, only upgrades this package(s), and if not upgrades them all.
I don't know to which extent commander would let you do that.

src/command-line/upgrade.js
```diff
@@ -0,0 +1,57 @@
...
+				log.error(`${colors.green(cmd.package)} is not installed.`);
+				return;
+			}
+
+			argsList.splice(1, 0, cmd.package);
+		}
+
+		return Utils.executeYarnCommand.apply(
+			this,
+			argsList
```
astorije: I believe `Utils.executeYarnCommand(...argsList)` does exactly this and is a bit nicer than the `.apply(this, ...)` pattern.

(This is kinda why I preferred the argument of `executeYarnCommand` to be one array instead of a dynamic number of arguments)

src/command-line/upgrade.js
jake-walker: I tried but there was no way to make it an optional argument so it would not allow the command to be executed without the package name.

jake-walker:
Ok @astorije, I've changed `update` to `upgrade` in the help and messages and also replaced `executeYarnCommand.apply(this, argsList)` with `executeYarnCommand(...argsList)`.

astorije:
Hey @jake-walker! Looking good, looking good, almost there!

After looking at [`commander.js`'s "Variadic arguments" section](https://github.com/tj/commander.js#variadic-arguments), it _seems_ to me that we can do that actually. I played around the `thelounge list` command so it doesn't make sense, but that's just for the example.
Here is the diff:

<details>
<summary>Diff</summary>

```diff
───────────────────────────────────────────
modified: src/command-line/users/list.js
───────────────────────────────────────────
@ list.js:10 @ const Helper = require("../../helper");
 const Utils = require("../utils");

 program
-   .command("list")
+   .command("list [packages...]")
    .description("List all users")
    .on("--help", Utils.extraHelp)
-   .action(function() {
+   .action(function(packages) {
+       if (packages.length) {
+           log.info("Upgrading the following packages:");
+           packages.forEach(p => log.info(`- ${p}`));
+       } else {
+           log.info("Upgrading all packages");
+       }
+
+       return;
+
```
</details>
<br>

And then use it like this:

- No arguments:
  ```
  $ thelounge list
  2018-04-28 05:53:57 [INFO] Upgrading all packages
  ```
- 1..* arguments:
  ```$ thelounge list foo bar
  2018-04-28 05:53:56 [INFO] Upgrading the following packages:
  2018-04-28 05:53:56 [INFO] - foo
  2018-04-28 05:53:56 [INFO] - bar
  ```

The usage help output looks like this:

```
$ thelounge list --help

  Usage: list [options] [packages...]
```

Does this help?

jake-walker:
Ah OK, I'll have a play around and see what I can do

jake-walker:
@astorije, that worked perfectly. Now someone can run:
* `thelounge upgrade` for upgrading all packages
* `thelounge upgrade thelounge-theme-foo` for a particular package
* `thelounge upgrade thelounge-theme-foo thelounge-theme-bar` for 2 particular packages and so on...

jake-walker:
OK @astorije, how does that look?

astorije:
Very nice! There is just one minor comment left :)

jake-walker:
Oh forgot about that one. That should do it now

PR #2383, opened 2018-04-24
Make condensed joins and parts shorter

xPaw:
saying channel is redundant in my opinion.

### DIFF
--- a/client/js/condensed.js
+++ b/client/js/condensed.js
@@ -40,10 +40,10 @@ function updateText(condensed, addedTypes) {
 				strings.push(obj[type] + (obj[type] > 1 ? " users have changed hostname" : " user has changed hostname"));
 				break;
 			case "join":
-				strings.push(obj[type] + (obj[type] > 1 ? " users have joined the channel" : " user has joined the channel"));
+				strings.push(obj[type] + (obj[type] > 1 ? " users have joined" : " user has joined"));
 				break;
 			case "part":
-				strings.push(obj[type] + (obj[type] > 1 ? " users have left the channel" : " user has left the channel"));
+				strings.push(obj[type] + (obj[type] > 1 ? " users have left" : " user has left"));
 				break;
 			case "quit":
 				strings.push(obj[type] + (obj[type] > 1 ? " users have quit" : " user has quit"));


client/js/condensed.js
```diff
@@ -40,10 +40,10 @@ function updateText(condensed, addedTypes) {
 				strings.push(obj[type] + (obj[type] > 1 ? " users have changed hostname" : " user has changed hostname"));
 				break;
 			case "join":
-				strings.push(obj[type] + (obj[type] > 1 ? " users have joined the channel" : " user has joined the channel"));
+				strings.push(obj[type] + (obj[type] > 1 ? " users have joined" : " user has joined"));
 				break;
 			case "part":
-				strings.push(obj[type] + (obj[type] > 1 ? " users have left the channel" : " user has left the channel"));
+				strings.push(obj[type] + (obj[type] > 1 ? " users have left" : " user has left"));
```
astorije: Would you mind leaving this one? I'm fine with removing it for the "joined" event, but the distinction between leaving (a channel) and quitting (a network) is frankly obscure if you're not used to IRC already. I find `2 users have left and 3 users have quit` confusing.

McInkay: I'm with @xPaw on this one.

I think if there's confusion for new people with the 2 messages, that we should combine the 2 in condensed.

So instead of `2 users have left and 3 users have quit` you would have `5 users have left`. Because, let's be honest, to most people (for most cases) it doesn't matter why someone has left the channel, they aren't in the channel now. Does that make sense?

xPaw: I was just thinking about that too, and combining part+quit into one isn't such a bad idea.

McInkay: We'd still keep the non-condensed lines the same so if you care you can look there. And, let's be honest, if you care why someone left you are probably going to want to look at the part/quit message anyway, so I don't think it removes any real useful information from the condensed while saving some space and making it less confusing.

astorije: With that extra change, that's totally reasonable! It would be weird to have `2 users have left and 3 users have quit`, but `5 users have left` with separation in detailed status makes complete sense. Nice!

PR #2385, opened 2018-04-24
Make a separate function to execute yarn commands; fallback to global yarn

### LINKED ISSUE #2301: installing with yarn does not install yarn locally... and `thelounge install` does not handle stderr from the yarn process

### LINKED ISSUE #2348: Warning when installing themes
* *Node version:* 8.9.4
* *Device, operating system:* macOS 10.13.3
* *The Lounge version:* source (0010d496)

---

It's actually marked as an error, but the theme gets correctly installed, and it doesn't kill the process, so it looks more like a warning than anything else:

```
$ thelounge install thelounge-theme-solarized
2018-04-12 03:59:35 [INFO] Retrieving information about the package...
2018-04-12 03:59:35 [INFO] Installing thelounge-theme-solarized v1.0.21...
2018-04-12 03:59:36 [ERROR] {"type":"warning","data":"Ignored scripts due to flag."}

2018-04-12 03:59:36 [INFO] thelounge-theme-solarized v1.0.21 has been successfully installed.
```

I have not noticed any actual error yet besides this output.
Note that `thelounge` is obtained from running `yarn link`.

xPaw:
Fixes #2301
Fixes #2348

### DIFF
--- a/src/command-line/install.js
+++ b/src/command-line/install.js
@@ -13,7 +13,6 @@ program
 		const fs = require("fs");
 		const fsextra = require("fs-extra");
 		const path = require("path");
-		const child = require("child_process");
 		const packageJson = require("package-json");
 
 		if (!fs.existsSync(Helper.getConfigPath())) {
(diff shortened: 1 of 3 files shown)


astorije:
> fallback to global yarn

When global yarn is available, is `yarn` not even installed in `node_modules`?! I can see how that could become a world of pain for us...

Also, in which "global" case the local version is not installed? Is that when installing `npm install -g yarn`? I appear to have both on my machine, but I believe the global one comes from Homebrew, not npm.

xPaw:
Global fallback is just a failsafe. The flat path (along side lounge) should work in most if not all cases when lounge is installed globally.

benharri:
Yay! Thanks @xPaw

src/command-line/install.js
```diff
@@ -53,59 +52,20 @@ program
...
+				"add",
+				"--json",
+				"--exact",
+				"--production",
+				"--ignore-scripts",
+				"--non-interactive",
+				"--cwd",
+				packagesPath,
+				`${json.name}@${json.version}`
+			).then(() => {
```
astorije: Why did you go for rest params here? I'd find it clearer to have an array: `Utils.executeYarnCommand(["add", "--json", /* ... */ ])`

xPaw: I need to prepend the executable.

astorije: What you're doing would still work: [yarn, ...parameters]

xPaw: How does having an array change anything then?

astorije: It just seems clearer to pass a single array argument than an arbitrary number of parameters imo.
As far as I can tell, I mostly see this in console.log and array.push (very rarely).
No big deal though

PR #2386, opened 2018-04-25
Count quits as parts in condensed messages to reduce information density

xPaw:
See https://github.com/thelounge/thelounge/pull/2383#discussion_r183915328 for discussion


astorije:
(You left a trailing comma :D)

PR #2388, opened 2018-04-26
Fix Alt+Up/Down cycling through channels of collapsed networks

### LINKED ISSUE #2200: Alt+Up/Down should not cycle through channels of a collapsed network

Raqbit:
This Fixes #2200.
It now filters channels inside a collapsed network when cycling through channels.

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -41,7 +41,7 @@ Mousetrap.bind([
 	"alt+up",
 	"alt+down",
 ], function(e, keys) {
-	const channels = sidebar.find(".chan");
+	const channels = sidebar.find(".chan").not(".network.collapsed :not(.lobby)");
 	const index = channels.index(channels.filter(".active"));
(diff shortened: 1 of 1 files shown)


client/js/keybinds.js
```diff
@@ -41,7 +41,7 @@ Mousetrap.bind([
 	"alt+up",
 	"alt+down",
 ], function(e, keys) {
-	const channels = sidebar.find(".chan");
+	const channels = sidebar.find(".chan").not(".network.collapsed .channel");
```
xPaw: This still allows queries and other windows to be targeted. This needs to target `:not(.lobby)` or something like that.

MaxLeiter:
@astorije, based on xPaw's last comment, I think queries will still be targeted.

xPaw:
Yeah I just tested it, and I can still iterate through hidden channels and queries. So it's not fixed.

PR #2391, opened 2018-04-26
Gracefully fail if sqlite3 module fails to load

xPaw:
This should make it easier to run on servers that don't have precompiled sqlite binaries.

### DIFF
--- a/src/plugins/sqlite.js
+++ b/src/plugins/sqlite.js
@@ -2,9 +2,18 @@
 
 const path = require("path");
 const fsextra = require("fs-extra");
-const sqlite3 = require("sqlite3");
 const Helper = require("../helper");
 const Msg = require("../models/msg");
+
+let sqlite3;
+
+try {
+	sqlite3 = require("sqlite3");
+} catch (e) {
+	Helper.config.messageStorage = Helper.config.messageStorage.filter((item) => item !== "sqlite");
+
+	log.error("Unable to load sqlite3 module. You might need to install it manually.");
+}
 
 const currentSchemaVersion = 1520239200;
 


astorije:
@xPaw, I've looked a bit into the SQLite issue, and I don't think we will ever be at peace with `node-sqlite3`, with the way we ship our software.

It seems to me that https://github.com/WebReflection/dblite is very appropriate for us as it solves most issues I can think of when it comes to distribution.
Distro packages like `thelounge-deb` can add a dependency on `sqlite3` (such as [here](https://github.com/thelounge/thelounge-deb/blob/master/debian/control#L6)) and git/npm installs would require one more step to install SQLite, with a similar graceful fallback to the one you made if not available.

The mental model behind it makes sense to me:
- distro packages are 1-step methods that take care of dependencies on Node and SQLite (currently the Debian package depends on the Node one), and set up the OS init service, all in one go, hassle-free.
- npm install is more hands-on because you have to install Node yourself (already the case) and, optionally, SQLite. You also have to take care of the auto-start/restart mechanism with your tool of choice (pm2, forever, screen, supervisor, systemd, etc.)
- git install is even more hands-on because you have to do all of the above + install devDependencies + `yarn build`.

Without this, I don't see how the distro packages can work, as we currently see on `thelounge-deb` that is currently completely broken.

I don't know how much change is required to switch from `node-sqlite3` to `dblite`, but I don't expect a lot.

What do you think?

xPaw:
> It seems to me that https://github.com/WebReflection/dblite is very appropriate for us as it solves most issues I can think of when it comes to distribution.

It doesn't solve the problem, it just moves it to a different domain, and will require users to manually install sqlite separately globally. That's worse.

astorije:
It _does_ solve the problem, how do you solve https://github.com/thelounge/thelounge-deb/issues/26 otherwise?!
It will require installing sqlite separately when installing through npm and git just like it does with Node/npm/yarn (or git itself). If you're using a distro package, it's transparent.

xPaw:
> thelounge/thelounge-deb#26 

We haven't tried adding sqlite as a dependency there, as you suggest. It might work.

astorije:
Correct, we didn't. I'm not sure how that could even work though, but worth a shot.

McInkay:
Good suggestions but ... they are secondary to this, I think. I think this is a good thing to add and we can see if there are other ways to sort it out.

astorije:
I'm not convinced this is secondary, actually. I'm not sure we should have such fallback: it seems to me that any potential errors caused by `sqlite3` currently happen at install time.
I might totally be missing a specific case so let me know if that's the case, but I don't believe this is a good fix for our issue(s).

Such fallback would be necessary however if we go for a solution that requires an extra step like I mentioned above. I would leave this PR open until we solve the bigger issue. Again, let me know if I'm totally wrong 😅

McInkay:
> I don't believe this is a good fix for our issue(s).

It's not a fix.

> it seems to me that any potential errors caused by sqlite3 currently happen at install time.

Correct, but at install time it's really hard to see this stuff

This shouldn't be merged in to solve the problem, but the problem exists in master and people are encountering it as we speak. We don't have another fix ready yet, and having this in would give people clear information about this until we do have a better fix. That's all.

xPaw:
@McInkay is right, this isn't a fix, but to allow running lounge if you don't care about sqlite. Use case: I wanted to test some things on node 10 on windows, and since there is no prebuilt package yet, runtime failed.

PR #2396, opened 2018-04-27
Use per-client channel and message ids

### LINKED ISSUE #2341: Remove global incremental ids for networks, channels, users and messages
We're unnecessarily "leaking" how many users use a particular instance by keeping global counts of all the objects. For example, connect to our demo and you'll see `#chan-1300` in the URL.

- [x] Networks: I think we can just straight up remove ids and use guids that were recently added. #2390
- [x] Clients: I haven't looked very closely, but I think ids here are only used for `socket.join` so we can just as easily use guids here if necessary. #2344
- [x] Channels: keep a local id per client #2396
- [x] Messages: Keep a local counter per client as there's some logic that relies on incremental ids (like loading more messages). #2396

xPaw:
Fixes #2341

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -70,6 +70,8 @@ function Client(manager, name, config = {}) {
 		attachedClients: {},
 		config: config,
 		id: uuidv4(),
+		idChan: 1,
+		idMsg: 1,
 		name: name,
 		networks: [],
 		sockets: manager.sockets,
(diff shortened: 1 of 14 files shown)


src/client.js
```diff
@@ -70,6 +70,8 @@ function Client(manager, name, config = {}) {
 		attachedClients: {},
 		config: config,
 		id: uuidv4(),
+		idChan: 1,
+		idMsg: 1,
```
astorije: Those are just for cosmetic purpose?

xPaw: Well message ids are used for loading more history and re connection logic.

Issue #1302, opened 2017-07-06
URL parser does not stop on apostrophe, valid in certain parts of link?


kode54:
The following link:

```
https://www.google.com's
```

Parses to include the 's in the link, which of course won't resolve.

xPaw:
This is a URI.js issue, can you create it there? https://github.com/medialize/URI.js

astorije:
@xPaw, do we want to add a test case (not enabled or merged for now) even though this is an issue on their side? It would still be nice to "monitor" this.

kode54:
Hope this is enough.

medialize/URI.js#339

kode54:
See? It's not an issue upstream, it's an issue with how you're using the library.

xPaw:
We are using the library precisely how he suggests, but he wants us to do filtering on our side.

astorije:
I think his answers were fair though, no?
If my (probably shitty) suggestion of checking TLDs is not possible, do we want to use `URI.findUri.trim` like he suggests, at least for apostrophe (and maybe others?)?

xPaw:
Fixed on master.

PR #2397, opened 2018-04-27
Switch to `linkify-it` for url detection in messages

xPaw:
This finds links with valid TLD but without protocol and www. prefix. Also finds protocol-aware urls. See updated test cases for examples. Also see https://markdown-it.github.io/linkify-it/ for more examples.

This also fixes a couple of issues we've had (but not all of them). For example #1302.  
This disables fetching previews for urls that contain username or password. (http://a:b@example.com)


astorije:
The only thing that worries me is that, despite its drawbacks URI.js is _fairly_ maintained (last commit was 3 months ago, 5k stars) while `linkify-it` hasn't seen any commits in 1.5 years.
Is this a concern? I see the owner did reply to you a few days ago, so at least he's around :)

xPaw:
@astorije To be fair, URI.js has some pain points in `withinString` that the author was unwilling to fix (for example it finds links that later on URIjs itself considers invalid).

astorije:
Fine by me. What's @McInkay's take?

client/js/libs/handlebars/ircmessageparser/findLinks.js
```diff
@@ -1,68 +1,42 @@
...
-			const matchedScheme = commonSchemes.find((scheme) => parsedScheme.endsWith(scheme));
-
-			// A known scheme was found, extract the unknown part from the URL
-			if (matchedScheme) {
-				const prefix = parsedScheme.length - matchedScheme.length;
-				start += prefix;
-				url = url.slice(prefix);
-			}
+for (const schema of commonSchemes) {
+	linkify.add(schema + ":", "http:");
```
astorije: So I'm reading https://github.com/markdown-it/linkify-it#addkey-value and it says that `value` is an alias to another schema, what does that mean?
Playing with your PR, the schema detection does work fine, so I'm not sure what alias here means.

test/client/js/libs/handlebars/ircmessageparser/findLinks.js
```diff
@@ -104,16 +104,150 @@ describe("findLinks", () => {
 		expect(actual).to.deep.equal(expected);
 	});
 
+	it("should find domains without www. but valid tld", () => {
+		const input = "google.com google.lv google.museum";
+		const expected = [{
+			link: "http://google.com",
```
astorije: Let's be forward and prefix with `https://` instead 🙏 🙏 🙏
I am reading in a few places that today about 75% of the web is HTTPS-ready (and it's only going to go up), so most links posted without a scheme should be fine. And if not, then post the http scheme explicitly.

src/plugins/irc-events/link.js
```diff
@@ -33,43 +31,45 @@ module.exports = function(client, chan, msg) {
...
-	if (links.length === 0) {
-		return;
-	}
+	// Only preview the first 5 URLs in message to avoid abuse
+	let allowedLinks = 5;
+
+	msg.previews = findLinks(cleanText).reduce((cleanLinks, link) => {
+		const url = normalizeURL(link.link);
+
+		if (url !== null && allowedLinks-- > 0 && !cleanLinks.some((l) => l.link === link.link)) {
```
astorije: Mind adding a comment for this check? It's one hell of a statement :D
I understand it checks that we haven't fetched 5 links yet + no duplicates, correct?

I think it would be even nicer if you were to split it a bit, like for example `allowedLinks-- > 0` is a bit weird to find in an `if`. No big deal though.

Also, I'm surprised there is no test for duplicate URLs (test that we don't fetch twice), do you think you could add a test for it, if that's something we can test?

client/js/libs/handlebars/ircmessageparser/findLinks.js
xPaw: Alias means it uses same rules for parsing and validation, so that we don't have to write our own rules for each protocol.

test/client/js/libs/handlebars/ircmessageparser/findLinks.js
xPaw: It's not our job to assume https. It is the web servers and browsers (if HSTS is used) to do that. I can imagine this unnecessarily breaking some sites, including local/internal ones.

src/plugins/irc-events/link.js
xPaw: Yeah I'll split the if and add a test.

xPaw: Done.

test/client/js/libs/handlebars/ircmessageparser/findLinks.js
lol768: I agree with @astorije on this, http should be the exception. They can specify the scheme explicitly if there's a good reason it *has* to be HTTP. Local/internal is not a good reason - these should be signed by an appropriate development/intranet CA.

xPaw: Everything everywhere has always defaulted to http. HTTPS isn't configured by default.

*All* websites will be available over http (even to just redirect to https) while the opposite is not true. I strongly feel it's not a good reason to break legitimate links that don't have https configured.

No other client (or browser) defaults to https as far as I know, and there's no rfc suggesting that I can find.

lol768: >Everything everywhere has always defaulted to http

Argumentum ad antiquitatem?

> All websites will be available over http (even to just redirect to https)

This is not the case. All internal services where I work refuse to accept connections on port 80 for instance. It's rare, but completely possible.

> I strongly feel it's not a good reason to break legitimate links that don't have https configured.

This will not impact full, existing legitimate links which are copied. The number of sites it will impact is small (and dropping everyday) and users impacted can easily 

> No other client (or browser) defaults to https as far as I know

Every single browser running HTTPS Everywhere will default to HTTPS where an appropriate ruleset is in place. Non-secure HTTP [is deprecated](https://blog.mozilla.org/security/2015/04/30/deprecating-non-secure-http/) and there are a not insignificant number of features you can't even use without serving your pages securely.

----

I get that backwards-compat. is important, but your argument seems to boil down to "nobody else does it so we shouldn't". In 1981, everyone used plaintext SMTP connections to deliver mail and there was [no RFC suggesting to the contrary](https://tools.ietf.org/html/rfc788#page-16).

astorije: > It's not our job to assume https. It is the web servers and browsers (if HSTS is used) to do that.

Yet we do assume _a_ scheme, HTTP. To this day and age, assuming HTTPS is sane. Because we do assume a scheme, it's totally our job to choose a sensitive one.

Like @lol768 said, if HTTP is expected, it should be specified. If nothing is specified, it's to our discretion.

> Everything everywhere has always defaulted to http.

I don't agree that this argument justifies going for HTTP. Browsers starting to treat HTTPS as a default and HTTP as insecure goes in this direction.

> _All_ websites will be available over http (even to just redirect to https) while the opposite is not true.

While this is generally the case, nothing technically enforces this. And like @lol768 there are legitimate reasons where this does not apply.

> I strongly feel it's not a good reason to break legitimate links that don't have https configured.

We are not breaking anything by doing this. Links without a scheme are already not valid URLs, so again, it is up to our interpretation. Explicitly specifying `http://` when it is important is the direction the web is heading towards by making things secure by default. My proposal is simply going towards the same direction.

> We're not a major player here, so it would be dumb to do something different.

Major player or not doesn't prevent us to do the _right_ thing. My opinion is still that the right thing to do is to interpret scheme-less links as HTTPS for better security, with extremely limited impact when this (rarely) fails.

> Either way, this entire discussion doesn't belong in this PR. Either make a separate one changing this behaviour or an issue to discuss further.

If I'm not mistaken, this PR introduces the fact that `example.com` is treated as a link, therefore converted to `http://example.com`. My proposal to go for `https://example.com` does belong in this PR, because I do not think we should introduce it as is.

dgw: > If I'm not mistaken, this PR introduces the fact that `example.com` is treated as a link

I've seen auto-linking bare domains like this go awry, FWIW. For example, clients that happily link every reference to Python or Perl filenames in a programming channel. Is that actually desirable behavior? Especially considering that The Lounge performs link prefetching, and would then try to go and get the page title for a (quite likely) nonexistent domain.

xPaw: I used a known TLD list, so the only question is whether `.py` is filtered against it's second-level domains, as they don't have open registrations (https://en.wikipedia.org/wiki/.py)

If so, that effectively avoids the problem you are describing.

@astorije We also prefix `http` for `www.` links, so this behaviour is not new. And we're still derailing this PR from improving the link matching to one specific small (mis)feature we have. I'd really like it if you created a separate issue to discuss all the details.

astorije: Good point, @dgw. Like @xPaw said, filtering against known TLDs should help. I'm still expecting some weird 💩 once in a while though, but I'm happy to give it a shot and see if the ratio usefulness/noise ratio is reasonable 😅

> We also prefix `http` for `www.` links, so this behaviour is not new. And we're still derailing this PR

We're not talking about the `www.` previews, so we're not derailing. I'm fine if after this PR we have something not fully consistent as we are introducing a new way to prefetch (but if you want to switch `www.` prefetches to HTTPS as well, be my guest, I would highly support that!)
Current proposal is a one-char change in code (as far as I can tell) + a few chars in tests, not a huge request that derails the PR.

> I'd really like it if you created a separate issue to discuss all the details.

My main concern is that you are against it in the first place. Not doing it in this PR + requesting a new issue effectively looks like a straight no, that will not be reconsidered. I do not want this new feature without HTTPS, and approving this change while hoping of a future issue/PR to do this would be foolish of me.
Heck, it would take me less time to make the change in this PR myself than opening an issue about it (that at this point I have no doubt you would refuse).

🤷‍♂️

xPaw: Would using protocol relative urls (and http for preview fetching as previews don't really care about it) be a good middle ground?

PR #2413, opened 2018-05-02
Use overscroll-behavior

xPaw:
Ref: https://developers.google.com/web/updates/2017/11/overscroll-behavior

This should prevent pull to refresh and any other navigation swipes in supporting browsers.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -19,6 +19,7 @@
 html,
 body {
 	height: 100%;
+	overscroll-behavior: none; /* prevent overscroll navigation actions */
 }
 
 body {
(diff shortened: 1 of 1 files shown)


astorije:
Do we need to do anything with https://github.com/thelounge/thelounge/blob/e1b6ba9f0dee869be6ae566a6aef91cd0bfdeeb8/client/css/style.css#L37-L41 ?

astorije:
Tested on iOS and unfortunately nothing changes there. Too bad, that's really where the experience is lacking. It does remove the overflow at boundary on macOS though, which I have no preference about (you just get used to it since it's consistent across the entire OS).

xPaw:
@astorije Safari doesn't support it is why nothing changes.

That overflow you linked should probably be removed in a future version, due to the browser support.

PR #2414, opened 2018-05-02
Fix join channel form not working

### LINKED ISSUE #2329: Join button when joining channel on existing server doesn't work
* *Node version:* 8.11.1 LTS
* *Browser version:* Safari 11.1
* *Device, operating system:* Mac OS 11.13.4
* *The Lounge version:* 3.0.0-pre.5

---
When clicking the + on the sidebar on an existing server, entering a channel name, and clicking Join, nothing happens and the channel is not joined. This seems to be the case regardless of browser. astorije on IRC said they were able to reproduce this on master.

xPaw:
Fixes #2329 and did some cleanup. I believe 6f47a78afd92f79f16106a52dc2e7dbd9de58c73 to be the offending commit.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -164,7 +164,7 @@ $(function() {
 		const target = self.attr("data-target");
 
 		if (!target) {
-			return false;
+			return;
 		}
 
 		// This is a rather gross hack to account for sources that are in the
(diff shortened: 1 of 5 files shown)


client/js/lounge.js
```diff
@@ -164,7 +164,7 @@ $(function() {
 		const target = self.attr("data-target");
 
 		if (!target) {
-			return false;
+			return;
```
xPaw: This is the actual bugfix. 

As it turns out `button` is `clicked` even when using enter keypress, unlike `input type="button"` so this click handler prevented it from being submitted.

Issue #2360, opened 2018-04-15
Allow showing passwords in password fields


xPaw:
Mostly relevant on edit network page.

Trivial demo: http://demo.itsolutionstuff.com/demo/demo-bootstrap-show-hidetoggle-password-input-field-using-bootstrap-show-passwordjs-pluginexample.html

c-ciobanu:
Hey, i want to contribute with this project, can i take this task?

xPaw:
Yes.

PR #2429, opened 2018-05-08
Add password visibility toggle feature

c-ciobanu:
My solution for the issue #2360.
What do you think?
Something to improve?

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1666,7 +1666,20 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	color: #2ecc40;
 }
 
-#settings #change-password .see-pw {
+#settings .error {
+	color: #e74c3c;
+	margin-top: 0.2em;
+}
+
+.password-container {
+	position: relative;
+}
+
+#sign-in .password-container {
+	width: 100%;
+}
+
+.password-container .see-pw {
 	font: normal normal normal 14px/1 FontAwesome;
 	font-size: 16px;
 	color: #cdd3da;
@@ -1675,17 +1688,17 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	right: 25px;
 }
 
-#settings #change-password .see-pw::before {
-	content: "\f06e";
+#sign-in .password-container .see-pw {
+	top: 42px;
+	right: 10px;
 }
 
-#settings #change-password .see-pw.visible::before {
-	content: "\f070";
+.password-container .see-pw::before {
+	content: "\f06e"; /* https://fontawesome.com/icons/eye?style=solid */
 }
 
-#settings .error {
-	color: #e74c3c;
-	margin-top: 0.2em;
+.password-container .see-pw.visible::before {
+	content: "\f070"; /* https://fontawesome.com/icons/eye-slash?style=solid */
 }
 
 #help .help-item {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1694,23 +1694,27 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	right: 15px;
 }
 
-.password-container .reveal-password button {
+.password-container .reveal-password span {
 	font: normal normal normal 14px/1 FontAwesome;
 	font-size: 16px;
 	color: #cdd3da;
-	padding: 10px;
+	width: 37px;
+	height: 37px;
+	display: flex;
+	justify-content: center;
+	align-items: center;
 }
 
-.password-container .reveal-password button:hover {
+.password-container .reveal-password span:hover {
 	color: #79838c;
 }
 
-.password-container .reveal-password button::before {
+.password-container .reveal-password span::before {
 	content: "\f06e"; /* https://fontawesome.com/icons/eye?style=solid */
 	transition: color 0.2s;
 }
 
-.password-container .reveal-password.visible button::before {
+.password-container .reveal-password.visible span::before {
 	content: "\f070"; /* https://fontawesome.com/icons/eye-slash?style=solid */
 	color: #ff4136;
 }
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1675,30 +1675,49 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	position: relative;
 }
 
+.password-container input {
+	padding-right: 37px;
+}
+
 #sign-in .password-container {
 	width: 100%;
 }
 
-.password-container .see-pw {
+#sign-in .password-container .reveal-password {
+	top: 31px;
+	right: 0;
+}
+
+.password-container .reveal-password {
+	height: 37px;
+	width: 37px;
+	position: absolute;
+	top: 2px;
+	right: 15px;
+	cursor: pointer;
+}
+
+.password-container .reveal-password i {
 	font: normal normal normal 14px/1 FontAwesome;
 	font-size: 16px;
 	color: #cdd3da;
-	position: absolute;
-	top: 13px;
-	right: 25px;
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	height: 100%;
 }
 
-#sign-in .password-container .see-pw {
-	top: 42px;
-	right: 10px;
+.password-container .reveal-password i:hover {
+	color: #79838c;
 }
 
-.password-container .see-pw::before {
+.password-container .reveal-password i::before {
 	content: "\f06e"; /* https://fontawesome.com/icons/eye?style=solid */
 }
 
-.password-container .see-pw.visible::before {
+.password-container .reveal-password.visible i::before {
 	content: "\f070"; /* https://fontawesome.com/icons/eye-slash?style=solid */
+	color: #ff4136;
 }
 
 #help .help-item {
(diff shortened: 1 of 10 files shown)


xPaw:
There's also a password field on login and connect (edit) windows.

c-ciobanu:
Ok, later will add the 2 other fields.

client/css/style.css
```diff
@@ -1666,6 +1666,23 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+	font: normal normal normal 14px/1 FontAwesome;
+	font-size: 16px;
+	color: #cdd3da;
+	position: absolute;
+	top: 13px;
+	right: 25px;
+}
+
+#settings #change-password .see-pw::before {
+	content: "\f06e";
```
MaxLeiter: If you could add `/** **/` comments here with links to the font-awesome website for consistency with other CSS, that'd be great

c-ciobanu: Will do

client/css/style.css
```diff
@@ -1671,6 +1671,36 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
 
+.password-container {
+	position: relative;
+}
+
+#sign-in .password-container {
+	width: 100%;
+}
+
+.password-container .see-pw {
```
xPaw: Add `cursor: pointer` and `:hover` styles to change color to something more visible.

client/css/style.css
```diff
@@ -1671,6 +1671,36 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+#sign-in .password-container .see-pw {
+	top: 42px;
+	right: 10px;
+}
+
+.password-container .see-pw::before {
+	content: "\f06e"; /* https://fontawesome.com/icons/eye?style=solid */
+}
+
+.password-container .see-pw.visible::before {
```
xPaw: Maybe also change color for visible icon to red or something.

xPaw:
This is conflicting with Edge's own reveal button, which seems to be an easy fix: https://github.com/vuetifyjs/vuetify/issues/537

client/views/windows/settings.tpl
```diff
@@ -187,17 +187,20 @@
...
+				<div class="col-sm-12 password-container">
 					<label for="new_password_input" class="sr-only">Enter desired new password</label>
 					<input type="password" id="new_password_input" name="new_password" class="input" placeholder="Enter desired new password">
+					<i class="see-pw" title="Show Password"></i>
 				</div>
-				<div class="col-sm-12">
+				<div class="col-sm-12 password-container">
 					<label for="verify_password_input" class="sr-only">Repeat new password</label>
 					<input type="password" id="verify_password_input" name="verify_password" class="input" placeholder="Repeat new password">
+					<i class="see-pw" title="Show Password"></i>
```
xPaw: - Use `aria-label` (you can also add tooltipped class to show a proper tooltip).
- Put `Hide password in an alt-attribute something like that: https://github.com/thelounge/thelounge/blob/49c5e6a502bf3261302386143b89d49262a23fe0/client/js/socket-events/more.js#L118
- Since `see-pw` element is a little elaborate, you could make a separate template file for it.
- Change the classname to something more verbose like `reveal-password`.

c-ciobanu:
Fixed all the things following your suggestions.
Coudn't create a new template, can't understand how the templates interact with each other.
If you can explain it to me would be great.

client/css/style.css
```diff
@@ -1671,6 +1671,55 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+	align-items: center;
+	height: 100%;
+}
+
+.password-container .reveal-password i:hover {
+	color: #79838c;
+}
+
+.password-container .reveal-password i::before {
+	content: "\f06e"; /* https://fontawesome.com/icons/eye?style=solid */
```
astorije: Could you add `transition: color 0.2s;` here? 🙏

client/css/style.css
```diff
@@ -1671,6 +1671,55 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+	right: 0;
+}
+
+.password-container .reveal-password {
+	height: 37px;
+	width: 37px;
+	position: absolute;
+	top: 2px;
+	right: 15px;
+	cursor: pointer;
```
astorije: I would love if we could make the `.reveal-password` element a `button` instead of a `span`. It comes with benefits such as no need to set the cursor property, but also improves accessibility with keyboard navigation and such. Thoughts?

client/views/windows/connect.tpl
```diff
@@ -78,8 +78,11 @@
 		<div class="col-sm-3">
 			<label for="connect:password">Password</label>
 		</div>
-		<div class="col-sm-9">
+		<div class="col-sm-9 password-container">
 			<input class="input" id="connect:password" type="password" name="password" value="{{defaults.password}}">
+			<span class="reveal-password tooltipped tooltipped-n tooltipped-no-delay" aria-label="Show Password" data-alt-label="Hide Password">
```
astorije: This is rather nitpicky, but in order to be consistent with similar elements like this, could you go for `Show password` and `Hide password` (i.e. drop the uppercase on `Password`)? Thanks!

client/views/windows/sign_in.tpl
```diff
@@ -2,15 +2,16 @@
...
+	<input class="input" name="user" autofocus>
 
-	<label>
-		Password
+	<div class="password-container">
+		<label>Password</label>
 		<input class="input" type="password" name="password">
-	</label>
+		<span class="reveal-password tooltipped tooltipped-n tooltipped-no-delay" aria-label="Show Password" data-alt-label="Hide Password">
+			<i></i>
```
astorije: I don't think we need (should?) have an `i` elements for those. If you look at other icon buttons, you'll see we don't wrap those in `i` elements. That will probably mean you'll need a container for the tooltip. That's rather unfortunate, but that's consistent with the rest so that when we improve/change things, we can do it all over the place more easily.

astorije:
That is great stuff @c-ciobanu, thanks a lot for helping with that!! ❤️

I have noticed (on the default theme) that the icons are not correctly vertically centered:

- Connect window: screen shot 2018-05-10 at 01 03 56
- Login page: screen shot 2018-05-10 at 01 04 53

It's more visible on the second one.

client/css/style.css
xPaw: I'm not sure it should be keyboard accessible (tabbing into it would be counter intuitive?).

Edge doesn't allow tabbing into their reveal button.

client/js/utils.js
```diff
@@ -97,6 +98,24 @@ function toggleNotificationMarkers(newState) {
 	viewport.toggleClass("notified", newState);
 }
 
+function togglePasswordField(elem) {
+	$(elem).on("click", function() {
```
xPaw: This needs to pass in a more specific selector or direct jquery element, otherwise you will be binding multiple events to the same buttons if you open connect window 5 times (it'll bind 5 times on the button in settings).

xPaw:
@c-ciobanu for a template, you just create a file in views folder (like `reveal-password.tpl`) and then include it like `{{> reveal-password}}` in html where you need it.

c-ciobanu:
@astorije the icons are not aligned because the css resets are not working properly, on chrome they look fine but on firex not. A separate file just for resets should be created for clarity, i sow some rules repeated on the css files.
For the span and i i would do something like this:
```
<span class="tooltipped tooltipped-n tooltipped-no-delay" aria-label="Show password">
<button type="button" class="reveal-password" aria-label="Show password" data-alt-label="Hide password"></button>
</span>
```
This can go?
@xPaw can i associate a js file with the logic to the template like in react or vue?

xPaw:
> @xPaw can i associate a js file with the logic to the template like in react or vue?

We don't have anything like that setup, sadly.

client/js/utils.js
```diff
@@ -97,6 +98,24 @@ function toggleNotificationMarkers(newState) {
...
+		const $this = $(this);
+		const input = $this.closest("div").find("input");
+
+		input.attr("type") === "password" ? input.attr("type", "text") : input.attr("type", "password");
```
xPaw: `input.attr("type", input.attr("type") === "password" ? "text" : "password");` is shorter and reads better.

client/js/utils.js
astorije: @xPaw, is this still relevant?

client/js/utils.js
astorije: @c-ciobanu, mind doing this?

astorije:
> For the span and i i would do something like this: [...]
> This can go?

Yep, I think it's nicer than `<i>`, if @xPaw is okay with that (that enables keyboard tabbing so might prefer `span`, I don't know, up to you guys).

client/js/utils.js
c-ciobanu: Sure, will commit this along with the changes on the `<i>` when @xPaw will express his opinion on what to use.

xPaw:
Use `span` for consistency with other icons we already have.

client/views/reveal-password.tpl
```diff
@@ -0,0 +1,3 @@
+<span class="reveal-password tooltipped tooltipped-n tooltipped-no-delay" aria-label="Show password" data-alt-label="Hide password">
+    <button type="button" aria-label="Show password" data-alt-label="Hide password"></button>
```
xPaw: You changed it to `button` instead of `span` as discussed.

xPaw: Yes. In this instance we don't want to use `button`.

c-ciobanu:
Thank you, I'll take a look into it and will submit a pr soon.

PR #2439, opened 2018-05-15
Specify parseInt base in respondToIdent

xPaw:
Just a minor issue to fix identd replying to malformed requests (when ports parse to NaN).


xPaw:
Either one will work here. `Number` doesn't have radix parameter though.

astorije:
Yeah that's why I've been preferring it lately lol. Some weird cases with empty string though

PR #2443, opened 2018-05-16
apply antialiased font-smoothing on body

williamboman:
Difference in Safari (make sure you view in 1:1 scale):

screenshot 2018-05-14 at 17 54 21
screenshot 2018-05-14 at 17 54 40

I have just tested this in macOS Safari, opened PR to set the ball rolling

How it looks like in `master` now: https://streamable.com/omf67

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -31,6 +31,8 @@ body {
 	-moz-user-select: none;
 	-ms-user-select: none;
 	user-select: none;
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
 	cursor: default;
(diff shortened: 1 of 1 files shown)


williamboman:
Actually, the `mask-image` property on container elements is what's causing this, setting global `font-smoothing` is just a dirty fix

astorije:
@williamboman, are you still looking into this? If so, do you want to do it in this PR or a new one?

williamboman:
@astorije Sorry for abandoning this one, was planning on updating it with a proper solution. I’ll close this one and create an issue. I’ll work on it when I get some spare time unless someone else manages to do so before me.

Issue #2008, opened 2018-01-28
False nick detection when it's a part of a word with `


christer88:
![thelounge](https://user-images.githubusercontent.com/11909212/35480786-5df28c92-0416-11e8-92ed-0c067e22226c.png)

User S was linked and coloured when the word **_it's_** was written in channel. User S shouldn't have been detected since it was a part of a bigger word.

christer88:
It's a tricky edge case since we would want user **_test_** to be detected in _**test's**_, but at the same time not user **_S_**. A fix for this would be to not detect users after a **_'_**?

MaxLeiter:
@christer but what if I ping you as ‘christer’?

realies:
Wrong tag, but what if there's a check which validates `'s` as the end of a word and skips highlighting?

MaxLeiter:
No, meant it at him

christer:
Me? don't think so.

Jay2k1:
Apparently this issue is not limited to apostrophes, umlauts also do it.

![image](https://user-images.githubusercontent.com/2070503/36103047-8177892a-100e-11e8-92ee-b89631c9ccbf.png)

ivan:
It does not make any sense to detect nicks in the middle of messages: I'm in a channel with a bot named `the`. Every `the` is highlighted. This is nonsense. The best fix here is to get rid of the feature.

xPaw:
@ivan That's for your constructive feedback.

Daniel15:
It would be good to have a user-editable blacklist for this, so users could add common words to it to avoid highlighting them. `##javascript` on FreeNode has the same issue - There's a user called `t`, so every mention of `t` after an apostrophe (eg. `can't`, `won't`), highlights the `t`.

Daniel15:
What I mean is that some people may like the short nicks being highlighted, while other people don't. Even then, some people may like only particular short nicks to be highlighted (eg. names like "dan" are fine to highlight, whereas common words like "so", "the", "and" are not). What I'd suggest is to have a user-editable blacklist of words to not highlight, and maybe seed it with blacklist of common short English words ("the", "but", "and", "so", etc). I wasn't sure how opinionated the developers of this project would want to be, as having a default blacklist is quite opinionated.

stevenengler:
I agree that the client should be smarter about this by default, but for those of us who want this feature, you can use the following CSS to blacklist common nicks:

```
.message .content .user[data-name="s"],
[...]

PR #2444, opened 2018-05-17
naive fix for false nick highlighting

benharri:
addresses #2008 to fix highlighting users after `'`. does not account for unicode characters since the `\w` escape in current js regex only looks for ascii letters. 

after a small bit of preliminary research, it looks like solving this would require choosing a unicode range, adding common diacritic characters to the regex, or using an external lib for unicode regex (such as [xregexp](http://xregexp.com)).

let's discuss how to solve the unicode issue.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -110,8 +110,9 @@ module.exports = function(irc, network) {
 		}
 
 		let match;
+		const boundaryRegExp = /\b(.*)\b/g;
 
-		while ((match = nickRegExp.exec(data.message))) {
+		while ((match = boundaryRegExp.exec(data.message))) {
 			if (chan.findUser(match[1])) {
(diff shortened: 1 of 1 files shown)


astorije:
I noticed @xPaw commented about this PR on the channel, but not sure what the conclusion was, so I'll leave him the first review. Thanks for your help @benharri! :)

xPaw:
@astorije This won't fix it because changing regex this way just makes the assumption that `'` is allowed in a nick. It doesn't fix detecting nicks in other languges. We should look for word boundaries instead.

xPaw:
I think we could get away by adding adding optional space before and after, in the original nick regex.

`\b` won't work because it only works on a-z0-9, so that doesn't really fix the issue of non English languages.

xPaw:
I'm gonna close this for now. If you come up with a better solution, reopen it.

PR #2445, opened 2018-05-18
Pointer cursor hovering nicks in userlist

MaxLeiter:


astorije:
I would rather turn these into `button`s (which would make the cursor come for free). Do you want to look into this instead, @MaxLeiter?
Also, @xPaw, is there any reason that would make us prefer the current `span role="button"` instead? It seems to me these should be actual buttons.

xPaw:
I believe there were problems with actual buttons. For one, I think it's not possible to select button text.

astorije:
That's fine, we made the userlist un-selectable anyway.

xPaw:
Nicks in chat though.

astorije:
This change is for `#chat .names .user`, and `.names` is only in userlist I believe.

xPaw:
We use one template everywhere nick is rendered, so yeah.

astorije:
Ah, that makes sense. Thanks for clarifying.

McInkay:
How about using links? They make more sense semantically than a button, and they get the cursor for free, but still allow selecting of the text.

astorije:
> How about using links?

@MaxLeiter, @xPaw?

xPaw:
I don't think using `<a>` will solve anything we haven't already (besides showing a useless href preview in browser when hovering over nicks).

astorije:
That makes sense.

PR #2466, opened 2018-05-24
Update Travis and Renovate configs

xPaw:
As the lockfile no longer updates on travis (we had to do it with greenkeeper), we can make use of `--frozen-lockfile` (we already use it on appveyor).

Also enabled `rebaseStalePrs` in renovate to keep rebasing open dependency PRs onto master.


xPaw:
@astorije That would be a breaking change. For example greenkeeper-lockfile usage is one (as they have update the lockfile to make a commit).

PR #2471, opened 2018-05-28
Remove URIjs

### LINKED ISSUE #2469: Remove URIjs
Now that linkify-js was merged, URIjs is only used to parse `location.search` here:

https://github.com/thelounge/thelounge/blob/f5a8f23d2bf19774c38f671f4f22d58a6fb82ea5/client/js/socket-events/configuration.js#L90

We could use `URLSearchParam`: https://caniuse.com/#feat=urlsearchparams

xPaw:
Fixes #2469

### DIFF
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -1,7 +1,6 @@
 "use strict";
 
 const $ = require("jquery");
-const URI = require("urijs");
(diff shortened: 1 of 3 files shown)


astorije:
So no polyfill like I mentioned at https://github.com/thelounge/thelounge/issues/2469#issuecomment-392389845?

PR #2488, opened 2018-05-31
Add asterisk in front of nick in action lines

### LINKED ISSUE #2486: Copy pasting actions (/me) should add an asterisk-like character
* *The Lounge version: source (commit 2de0222)* 

---

Some time ago, a feature was introduced that would remove unnecessary whitespace and add `<>` around nicks when copy pasting PRIVMSGs, so they look like output from classic IRC clients like mIRC or irssi:
image
becomes 
`13:51 <@Jay2k1> i am testing` 
when copy pasted.

Actions/descriptions don't get anything added:
image
becomes
`13:52 @Jay2k1 is testing`

I think they should get a `*` to be in line with this, so it looks like this:
`13:52 * @Jay2k1 is testing`

A bullet point &bull; instead of an asterisk would be an alternative, just, you know, *something*.

Jay2k1:
fixes #2486

This adds a `*` in front of the sender nickname on copy pasting actions.

Tests:
image

### DIFF
--- a/client/js/clipboard.js
+++ b/client/js/clipboard.js
@@ -22,6 +22,10 @@ function copyMessages() {
 			el.text(`<${el.text()}>`);
 		});
 
+	$(documentFragment)
+		.find(".content > .user")
+		.prepend("* ");
+
 	div.id = "js-copy-hack";
 	div.appendChild(documentFragment);
 	chat.appendChild(div);


xPaw:
This also affects actions like joins and quits.

For example:
```
16:02  * card.freenode.net sets mode +n
16:02  * card.freenode.net sets mode +s
```

Ultimately I don't think it matters, but something to keep in mind.

client/js/clipboard.js
```diff
@@ -22,6 +22,13 @@ function copyMessages() {
 			el.text(`<${el.text()}>`);
 		});
 
+	$(documentFragment)
+		.find(".content > .user")
+		.each((_, el) => {
+			el = $(el);
+			el.text(`* ${el.text()}`);
```
xPaw: I think this can be just `$(el).prepend("* ");`

Issue #2489, opened 2018-05-31
Add indicator to page title on new query/highlight


Jay2k1:
This has heavily been discussed in IRC, so let's have some ordered, readable discussion here.

I had the idea of adding something to the title as (another) indicator for new query messages and highlights - basically everything that would trigger the "alerted" favicon. The main intention is to make Chrome *(edit: and Firefox, thx @MrLemur)* show a dot on a pinned TL tab like so:
![image](https://user-images.githubusercontent.com/2070503/40783231-1a147ed6-64e2-11e8-9d79-0cdeb219b74e.png)
![image](https://user-images.githubusercontent.com/2070503/40784219-4e92103a-64e5-11e8-81ab-780cac90fe7b.png)

They do this when the title changes and the tab is pinned.

Because other browsers do not do this afaik, and because you don't always have TL pinned, it makes sense to think about *how* to change the title so it benefits regular tabs in all browsers too. I have two ideas:

1. prepend the title with a &bull; bullet point (Chrome on top, Vivaldi below): 
![image](https://user-images.githubusercontent.com/2070503/40783812-e51169cc-64e3-11e8-806b-0a221bd0076b.png)
![image](https://user-images.githubusercontent.com/2070503/40783379-80bd68be-64e2-11e8-8f50-a07849598b36.png)

2. prepend the title with a count of events: 
![image](https://user-images.githubusercontent.com/2070503/40783782-cef9bff4-64e3-11e8-88ab-d19c51a5c55b.png)
![image](https://user-images.githubusercontent.com/2070503/40783437-ad33fcb4-64e2-11e8-8d93-3c3d3a3b355d.png)

As you can see, the Vivaldi browser adds its own "count badge" to the favicon when the title starts with (number).
It is also Twitter's way of doing it.

I prefer 2. for these reasons.

Something that came up is what the count should include. New query messages? Highlights? Notices? Maybe even disconnects? Server notices?

MrLemur:
Just a note: Firefox shows a blue icon under a pinned tab if the title changes.

xPaw:
To futher add a reason to change title:

On windows, pinned websites use favicon image that doesn't change, so the alerted icon is not visible in the taskbar.

Putting number of highlights in the title should work just fine. The only problem that will need fixing is propagating the actual number to the client, AFAIK the client doesn't know how many messages were highlights.

McInkay:
[...]

PR #2496, opened 2018-06-01
[WIP] Add alert count to title 

Jay2k1:
So, as discussed in #2489, this is my first version of

> showing the number of `.badge.highlight` (or whatever the classname) elements, which is how alerted favicon is applied in the first place IIRC.

In the future, the server should communicate the number of highlights (including queries) to the client and that number should be shown instead, because it can be bigger than just the number of highlight-type badges in the channel list.


client/js/utils.js
```diff
@@ -98,6 +99,22 @@ function toggleNotificationMarkers(newState) {
...
+	let title = $(document.body).data("app-name");
+	const chanTitle = $("#sidebar").find(".chan.active").attr("aria-label");
+
+	if (chanTitle.length > 0) {
+		title = `${chanTitle} — ${title}`;
+	}
+
+	// change title to display event count
+	// for now, we only show the number of "highlight" class badges
+	const alertEventCount = document.getElementsByClassName("badge highlight").length;
```
xPaw: Can you update this to count up `data-highlight`? :)

Also use jQuery selectors.

Jay2k1:
The future is now!

Now showing the number of highlight events in the title.

client/js/lounge.js
```diff
@@ -225,16 +225,9 @@ $(function() {
 			.trigger("show");
 
 		utils.togglePreviewMoreButtonsIfNeeded();
+		utils.updateTitle();
 
-		let title = $(document.body).data("app-name");
 		const chanTitle = chan.attr("aria-label");
```
xPaw: Can you move this variable where it's used now (inside of an if)

client/js/utils.js
```diff
@@ -99,6 +100,24 @@ function toggleNotificationMarkers(newState) {
...
+	if (chanTitle.length > 0) {
+		title = `${chanTitle} — ${title}`;
+	}
+
+	// add highlight count to title
+	let alertEventCount = 0;
+	$(".badge.highlight").each(function() {
+		alertEventCount += parseInt($(this).attr("data-highlight"));
+	});
+	const titleAlert = (alertEventCount === 0 ? "" : "(" + alertEventCount + ") ");
```
xPaw: Can you change this to something like this:
```js
if (alertEventCount > 0) {
	title = `(${alertEventCount}) ${title}`;
}
```

PR #2498, opened 2018-06-02
Minify handlebar templates

xPaw:
This saves about 7kb on production build, ~~and probably reduces memory usage at runtime by a little bit (since there's less whitespace in DOM)~~

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -58,7 +58,7 @@ const config = {
 				include: [
 					path.resolve(__dirname, "client/views"),
 				],
-				use: {
+				use: [{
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -68,7 +68,16 @@ const config = {
...
 					},
-				},
+				}, {
+					loader: "html-minifier-loader",
+					options: {
+						ignoreCustomFragments: [
+							/{{[\s\S]*?}}/,
+						],
+						sortAttributes: true,
+						sortClassName: true,
```
astorije: Why are those necessary?

PR #2501, opened 2018-06-02
Improve formatting of text logs

### LINKED ISSUE #2499: Better formatted logging.
I am going to come right out about this, the logging format is really crappy. Logs like this would be easier to read, and just look better. http://echelog.com/logs/browse/supybot/1311285600 Hope something like this can happen in the future.

Jay2k1:
fixes #2499 

This adds individual formatting for different IRC events to logging to text.

I am using the same log format that znc uses.

### DIFF
--- a/src/plugins/messageStorage/text.js
+++ b/src/plugins/messageStorage/text.js
@@ -5,6 +5,7 @@ const fsextra = require("fs-extra");
 const path = require("path");
 const filenamify = require("filenamify");
 const Helper = require("../../helper");
+const Msg = require("../../models/msg");
 
 class TextFileMessageStorage {
 	constructor(client) {
@@ -41,26 +42,57 @@ class TextFileMessageStorage {
 
 		let line = `[${Helper.getHumanDate()}] `;
 
-		if (msg.type === "message") {
-			// Format:
-			// [2014-01-01 00:00:00] <Arnold> Put that cookie down.. Now!!
-			line += `<${msg.from.nick}> ${msg.text}`;
-		} else {
-			// Format:
-			// [2014-01-01 00:00:00] * Arnold quit
-			line += `* ${msg.from.nick} `;
+		// message types from src/models/msg.js
+		switch (msg.type) {
+		case Msg.Type.ACTION:
+			// [2014-01-01 00:00:00] * @Arnold is eating cookies
+			line += `* ${msg.from.mode}${msg.from.nick} ${msg.text}`;
+			break;
+		// case Msg.Type.INVITE: // neither incoming nor outgoing invites are logged
+		case Msg.Type.JOIN:
+			// [2014-01-01 00:00:00] *** Arnold (~arnold@foo.bar) joined
+			line += `*** ${msg.from.nick} (${msg.hostmask}) joined`;
+			break;
+		case Msg.Type.KICK:
+			// [2014-01-01 00:00:00] *** Arnold was kicked by Bernie (Don't steal my cookies!)
+			line += `*** ${msg.target.nick} was kicked by ${msg.from.nick} (${msg.text})`;
+			break;
+		case Msg.Type.MESSAGE:
+			// [2014-01-01 00:00:00] <@Arnold> Put that cookie down.. Now!!
+			line += `<${msg.from.mode}${msg.from.nick}> ${msg.text}`;
+			break;
+		case Msg.Type.MODE:
+			// [2014-01-01 00:00:00] *** Arnold set mode +o Bernie
+			line += `*** ${msg.from.nick} set mode ${msg.text}`;
+			break;
+		case Msg.Type.NICK:
+			// [2014-01-01 00:00:00] *** Arnold changed nick to Bernie
+			line += `*** ${msg.from.nick} changed nick to ${msg.new_nick}`;
+			break;
+		case Msg.Type.NOTICE:
+			// [2014-01-01 00:00:00] -Arnold- pssst, I have cookies!
+			line += `-${msg.from.nick}- ${msg.text}`;
+			break;
+		case Msg.Type.PART:
+			// [2014-01-01 00:00:00] *** Arnold (~arnold@foo.bar) left (Bye all!)
+			line += `*** ${msg.from.nick} (${msg.hostmask}) left (${msg.text})`;
+			break;
+		case Msg.Type.QUIT:
+			// [2014-01-01 00:00:00] *** Arnold (~arnold@foo.bar) quit (Connection reset by peer)
+			line += `*** ${msg.from.nick} (${msg.hostmask}) quit (${msg.text})`;
+			break;
+		case Msg.Type.CHGHOST:
+			// [2014-01-01 00:00:00] *** Arnold changed host to: new@fancy.host
+			line += `*** ${msg.from.nick} changed host to '${msg.new_ident}@${msg.new_host}'`;
+			break;
+		case Msg.Type.TOPIC:
+			// [2014-01-01 00:00:00] *** Arnold changed topic to: welcome everyone!
+			line += `*** ${msg.from.nick} changed topic to '${msg.text}'`;
+			break;
 
-			if (msg.hostmask) {
-				line += `(${msg.hostmask}) `;
-			}
-
-			line += msg.type;
-
-			if (msg.new_nick) { // `/nick <new_nick>`
-				line += ` ${msg.new_nick}`;
-			} else if (msg.text) {
-				line += ` ${msg.text}`;
-			}
+		default:
+			// unhandled events will not be logged
+			return;
 		}
 
 		line += "\n";


src/plugins/messageStorage/text.js
```diff
@@ -47,23 +47,56 @@ class TextFileMessageStorage {
 		let line = `[${time}] `;
 
 		if (msg.type === "message") {
```
xPaw: Probably better off as `switch`.

src/plugins/messageStorage/text.js
```diff
@@ -46,24 +46,67 @@ class TextFileMessageStorage {
...
-			// [2014-01-01 00:00:00] <Arnold> Put that cookie down.. Now!!
-			line += `<${msg.from.nick}> ${msg.text}`;
-		} else {
-			// Format:
-			// [2014-01-01 00:00:00] * Arnold quit
-			line += `* ${msg.from.nick} `;
+		switch (msg.type) {
+		case "message":
+			// [2014-01-01 00:00:00] <@Arnold> Put that cookie down.. Now!!
+			line += `<${msg.from.mode}${msg.from.nick}> ${msg.text}`;
```
xPaw: I believe ZNC doesn't log with modes. Does PISG for example support modes in energymech log parser?

src/plugins/messageStorage/text.js
```diff
@@ -46,24 +46,67 @@ class TextFileMessageStorage {
...
+			line += `*** ${msg.from.nick} sets mode: ${msg.text}`;
+			break;
+		case "topic":
+			// ignore topic lines you get on join
+			if (typeof msg.from.nick === "undefined") {
+				return;
+			}
+
+			// [2014-01-01 00:00:00] *** Arnold changes topic to 'welcome everyone!'
+			line += `*** ${msg.from.nick} changes topic to '${msg.text}'`;
```
Zarthus: I recommend "changes topic to:" instead to avoid

(1) trailing or leading whitespace being painfully obvious
(2) quotes in the text itself not looking misleading

src/plugins/messageStorage/text.js
```diff
@@ -46,24 +46,67 @@ class TextFileMessageStorage {
...
+			break;
+		case "topic_set_by":
```
Zarthus: Could perhaps exit early here (L29 instead).

xPaw: It's fine here, this event happers rarely.

src/plugins/messageStorage/text.js
xPaw: This is following ZNC's format, as we're trying to follow it, and not invent our own.

src/plugins/messageStorage/text.js
Jay2k1: You are right. I looked at PISG's individual log parsers, and I found one (IRCap) where the user mode is being removed from the nick. They don't do that for energymech, so I guess that's a no.

So you want to be compatible with something like PISG I take it?

christer88: Yes please, PISG support would be highly appreciated!

src/plugins/messageStorage/text.js
```diff
@@ -46,24 +46,67 @@ class TextFileMessageStorage {
...
+			// [2014-01-01 00:00:00] *** Joins: Arnold (~arnold@foo.bar)
+			line += `*** Joins: ${msg.from.nick} (${msg.hostmask})`;
+			break;
+		case "part":
+			// [2014-01-01 00:00:00] *** Parts: Arnold (~arnold@foo.bar) (Bye all!)
+			line += `*** Parts: ${msg.from.nick} (${msg.hostmask}) (${msg.text})`;
+			break;
+		case "quit":
+			// [2014-01-01 00:00:00] *** Quits: Arnold (~arnold@foo.bar) (Connection reset by peer)
+			line += `*** Quits: ${msg.from.nick} (${msg.hostmask}) (${msg.text})`;
```
astorije: Shouldn't joins, parts, and quits be something like "foo has quit" instead of "Quits: foo" to match nick, kick, mode, etc.?

Jay2k1: @xPaw's goal here was to match an existing, common log format, and that's how znc is doing it.

astorije: I get that, but I think consistency would make more sense than following ZNC's format blindly. We're not bound to ZNC in any way. Thoughts?

xPaw: The idea was to follow an established log format (not ZNC specifically, as theirs is based upon energymech).

https://pisg.github.io/ has a nice list of different log formats, if there's a better looking one, we could switch to that.

astorije:
@Jay2k1 @xPaw I know you guys are currently discussing this on the channel (can't keep up with the current pace of 1k message every 2 days :D), feel free to assign me when you're done!

Jay2k1:
@astorije well, we've been discussing the log format and the "conclusion" was "screw it, let's make our own". So we discussed a bit and this is the result:
```
[2014-01-01 00:00:00] * @Arnold is eating cookies
[2014-01-01 00:00:00] *** Arnold (~arnold@foo.bar) joined
[2014-01-01 00:00:00] *** Arnold was kicked by Bernie (Don't steal my cookies!)
[2014-01-01 00:00:00] <@Arnold> Put that cookie down.. Now!!
[2014-01-01 00:00:00] *** Arnold set mode +o Bernie
[2014-01-01 00:00:00] *** Arnold changed nick to Bernie
[2014-01-01 00:00:00] -Arnold- pssst... I have cookies!
[2014-01-01 00:00:00] *** Arnold (~arnold@foo.bar) left (Bye all!)
[2014-01-01 00:00:00] *** Arnold (~arnold@foo.bar) quit (Connection reset by peer)
[2014-01-01 00:00:00] ?Arnold? TIME
[2014-01-01 00:00:00] !Arnold! TIME Mon Jun 04 00:15:46 2018
[2014-01-01 00:00:00] *** Arnold changed host to 'new@fancy.host'
[2014-01-01 00:00:00] *** Arnold changed topic to 'welcome everyone!'
```
Notes:
- CTCP is my idea, hasn't been talked about
- user modes in msg/action is my idea too, but I'm not pinned down on that one
- we agreed that it would be nice to have the tenses consistent (not "is now known as" -> present and "has quit" -> past)
- we also agreed we do not need "has" and thus opted for "X joined" instead of "X has joined"

Opinions?

I figured before further coding happens, we should first agree on a format.

astorije:
A couple comments:

- I find the `-`, `?`, `!` to be too mysterious. I'd prefer to spell it out, something like "Arnold sent a notice: ...", "Arnold sent a CTCP request", etc. Of course, not at all married to the wording so feel free to improve, but I think you get the idea re: symbols.
- I'd go with double quotes around topic/host, or actually I'd even prefer @Zarthus's suggestion above, use a colon and no quotes.

Apart from that it looks pretty good! 👏

Jay2k1:
> use a colon and no quotes.

I can live with that.

> something like "Arnold sent a notice: ...", "Arnold sent a CTCP request", etc.

Heh, I knew you'd say that. As usual, my response is that this is too verbose and wordy for my taste. Using "Arnold sent a notice" instead of `-Arnold-` is like using "Arnold sent a channel message" instead of `<Arnold>` to me.

Side note: currently, both incoming and outgoing invites, CTCP requests and notices are not sent to the text logger and I'm not sure if this is going to change, i.e. if we want TL to log these at all (I think it should log notices, but probably not CTCPs).

xPaw:
@astorije `-` is used to avoid mistaking it for other message types.

I don't think we need to include CTCP in logs.

Zarthus:
Just a few cents from someone who has worked on a public logging service in the past;

- I'm on the boat of not including CTCP in logs
  - I already agree with @astorije about the special characters for ctcp in case you do want to include them, but `!` is a mode (protect) on _some_ smaller networks and might be confusing for them.
- make sure you don't make op/vnotice (or op/vmsg) as a regular notice to avoid confusion, or clearly write a ticket to improve this in the future.[1]
- It might be worth debating including the channel in things (it offers no significant benefit if the logfile is already named by the channel, but I personally like the extra verbosity) - `x has kicked y from #c`
- If we subscribe to `extended-join`, can we include the realname in the join message?
----

[1] To explain more in depth, there are some more arbitrary status (e.g. STATUSMSG (`/msg @#channel what up`, `/notice @#channel what up`)) which makes sense to test whether it's logged/not logged, and if it is, if we can somehow include the "channel mode" in the logs. However it might not be exposed in the TL api. -> ticket for the future at risk of making them look like normal notices IMO

Jay2k1:
- yeah, let's not include CTCPs
- about omsg/vmsg: I tested if the logger receives them, this is what I found:
  - mIRC's `/onotice #chan foobar` (send notice to all chan ops) creates a regular type NOTICE object, just that the text field looks like `@ foobar`
  - mIRC's `/omsg #chan foobar` (send message to all chan ops) creates a regular query message
Looking at rawlog, these come in as
```
<- @time=2018-06-04T21:44:20.968Z :PDA!PDA@host PRIVMSG Jay2k1 :foobar
<- @time=2018-06-04T21:44:33.193Z :PDA!PDA@host NOTICE @#pda :@ foobar
```
So I guess /omsg is really just regular PRIVMSG, just with selected targets, and /onotice does have a proper target which is probably just not parsed by either irc-fw or TL. 

* include channel: I agree that `foo (bar@baz) joined #foobar` does look nice even if each chan gets its own log.

As for other things: Looking at how broken the text logger currently is, I'd be happy if we could agree on a format and then push out a working logger that covers the most common message types and doesn't make obvious mistakes. Known issues should get a ticket then, as you suggested.

If there's enough demand (or a volunteer ;) ) it can still be extended to cover edge cases later, but I'm afraid this PR might go into bikeshedding and waiting until something is being fixed in irc-fw, and it's *just the text logger*.

Not sure how the rest of you thinks about that.

astorije:
Just curious, why not include CTCP requests/responses? It's a log, after all. Plus we already store everything (everything?) else such as MOTDs and such, so I'm not sure why silencing the CTCPs.
I actually see a value in logging CTCP requests made against a user, for example to build stats or monitoring (oddly I have had a lot of CTCP requests recently, so keeping track of those is interesting).

Zarthus:
(1) they aren't all that interesting
(2) there is no guarantee of a response, the logs could look "broken" or
confusing.
(3) they always reply the same thing; the output is predictable. (for
logging, all of it but Version is useless; time is useless, finger is
useless, client info and I can't even remember the third most common one.)

The only cases I can think where this is interesting is when someone else
sends out a mass ctcp (against a channel) and you want to review it before
banning them. (but if you have to look in the logs, chances are some other
chanop/Oper beat you to it already.)

In the 10 years I've used irc, I've never went back to my logs one time to
look up and old ctcp I sent or received. It is the more dynamic content
that mattered (the joins or the messages).

From what I've heard, if you desperately need ctcp RQ/RS (for stat
generation, this makes more sense anyway) , you can query against sqlite. I
would then much prefer that in favour of cleaner and less confusing log
output.

astorije:
CTCP logs would be in the network files, so it's not like it would be noisy for the channel logs. I'm still in favor of including them.

> you can query against sqlite. I
> would then much prefer that in favour of cleaner and less confusing log
> output.

I disagree with the argument that one type of log should have more or less than another one. Storage solution and granularity are 2 different things IMO, and log content shouldn't depend on where you're storing them (remember that in the future, log storage would be abstracted to a package interface so that one could send logs wherever).
Clean logs (especially the text files, but I've _just_ said I think we shouldn't discriminate on format) is the least of my worries. I would prefer comprehensive logs over handpicked ones. And again, this is for network logs, so I would shove as much as possible in there without worrying.

astorije:
And yeah, I would definitely want to log CTCP requests sent against me, at the very least (though still would want all of them logged).

xPaw:
>  Plus we already store everything (everything?) else such as MOTDs and such

MOTD and some other types are filtered out.

astorije:
Well... I'd like to have those in there too (maybe more CTCP / raw than MOTD though), but I feel like I'm losing the argument. I've never seen situations where a software simply logs too much that we have to hand pick what we want to exclude, but 🤷‍♂️

Fine with whatever, and if the need to log more stuff arise (I'm thinking some sort of fail2ban-like workflow), then we can re-evaluate.

@xPaw, I can review when you're 👍 on this, as I'm not sure if @Jay2k1 finished making the changes you guys talked about on the channel.

I still think spelling out `NOTICE` rather than `-nick-` is much much better, especially since `-` itself is allowed in nicks (though not at first, I know, but `[-foo-]` or `foo-` like I'm seeing right now would make this super confusing). Ideas: `NOTICE nick: lorem ipsum`, `NOTICE <nick> lorem ipsum`, `nick NOTICE lorem ipsum`, etc.
Would that solve the extra verbosity of the "sent a notice" I initially suggested?

Jay2k1:
> Fine with whatever, and if the need to log more stuff arise (I'm thinking some sort of fail2ban-like workflow), then we can re-evaluate.

I don't really care either, it's just that if we decide to not log something, we don't need to discuss its log format ;). And re: fail2ban, what does that have to do with logs? If anything, CTCP flood protection should be implemented directly in TL server, or do I misunderstand something?

Regarding notices:

I've looked into how notices are being logged in other (what I think are) popular IRC clients:
mIRC: `2018-06-06 13:31:12 -Jay2k1- test notice`
irssi: `13:53:20-Jay2k1(J@y2k1.it)- hi there`
Textual: `[2018-06-06T13:30:00+0200] -Jay2k1- test notice`
hexchat: `Jun 06 14:08:30 -Jay2k1-	hi there`

This is why it makes sense to me to do it like this. I didn't come up with this format, I've just made it the same way I know it from my previous IRC experience.

astorije:
> it's just that if we decide to not log something, we don't need to discuss its log format ;)

That's fair.

> And re: fail2ban, what does that have to do with logs? If anything, CTCP flood protection should be implemented directly in TL server, or do I misunderstand something?

Tools built using the same principles as fail2ban parse logs to take actions. Could be sending an email, shutting down a process, adding an IP to a firewall (in the case of SSH for example), etc.
I'm just thinking potential use cases that don't exist yet so don't mind me. Like I said, we can go without for now and see if feedback shows that anyone needs it.

> I didn't come up with this format, I've just made it the same way I know it from my previous IRC experience.

Thanks for pointing out examples from other tools, that's pretty useful! I understand that, but if we're making our own format to have something "better" or at least more consistent, we can propose something better here too. Imagine if your nick was `[-Jay2k1-]` or `Jay2k1-` (I can see examples of both along with dozen others in my current channels right now), that would look like this:

```
[2014-01-01 00:00:00] -[-Jay2k1-]- test notice
[2014-01-01 00:00:00] -Jay2k1-- test notice
```

That doesn't look too practical to me. Maybe a few alternatives:

```
[2014-01-01 00:00:00] NOTICE Jay2k1: test notice
[2014-01-01 00:00:00] Jay2k1 NOTICE: test notice
[2014-01-01 00:00:00] Jay2k1 sent a notice: test notice
```

I don't like the second one too much. Last one is most verbose but most consistent with topic changes for example.
I'm not strongly in favor of any of those, so if you or @xPaw have a better idea, I can be easily convinced! :)

xPaw:
> That doesn't look too practical to me. 

The format works fine, otherwise so many different implementations wouldn't agree on it. Don't forget there's spaces around the dashes, so if you're regexing or parsing it you're guaranteed to parse it just fine.

Jay2k1:
> Last one is most verbose but most consistent with topic changes for example.

Just that a topic change is an event and a notice is a message. It should be consistent with messages, not with events, so in my opinion that's the least favorable way. 

I also agree with xPaw.

The point you are making is readability when nicknames contain or end with a dash, which is understandable. 
I guess it was never a big enough problem for the "big players" on the IRC client market to change it (found out znc also uses that format), and I'm not a big fan of doing things differently if there's no strong need to, especially when they're common and have been done their way for many years.

Additionally, I don't know about others but I hardly ever get a notice by real people. In my personal experience, most notices are messages or responses from bots and make <1% of messages, so I don't think readability would be an issue.

Then again, the clients I listed above (minus znc because it's not a client) also _show_ notices like that, they don't only log it in this format. So if you talk about consistency, it would probably be consistent to use the first of your three suggestions as it comes close to how TL shows notices in the client (which, if it was for me, would also be `-nick- text`, but I understand that's unlikely to happen).

astorije:
> I'm not a big fan of doing things differently if there's no strong need to

As part of this PR, we explicitly agreed to make up our own format, so we _are_ doing things differently.

I still think we are not headed in the right direction here (whitespaces around dashes don't make up for the confusion) but whatever, I don't care enough to fight this and you've done enough research to bring objective arguments, so you guys do what you want and I'll go care about the UI instead ;)

I'll approve whatever syntax @xPaw approves, and I can review the code after @xPaw gives his 👍

xPaw:
@Jay2k1 can you push your latest changes?

src/models/chan.js
```diff
@@ -185,7 +185,8 @@ Chan.prototype.writeUserLog = function(client, msg) {
 	}
 
 	// Is this particular message or channel loggable
-	if (!msg.isLoggable() || !this.isLoggable()) {
+	// exception for notices so they get logged even when sent to the lobby
+	if (!msg.isLoggable() || (!this.isLoggable() && msg.type !== "notice")) {
```
xPaw: I think if we change this to `if (!msg.isLoggable() && !this.isLoggable()) {`

It should work just as fine.

src/models/msg.js
```diff
@@ -47,6 +47,8 @@ class Msg {
 			this.type !== Msg.Type.BANLIST &&
 			this.type !== Msg.Type.IGNORELIST &&
 			this.type !== Msg.Type.TOPIC_SET_BY &&
+			this.type !== Msg.Type.CTCP &&
```
xPaw: Instead of filtering them here, just remove the `default` case in the text logger, then it won't get written, but will remain in sqlite.

src/plugins/messageStorage/text.js
```diff
@@ -41,31 +42,74 @@ class TextFileMessageStorage {
...
-		if (msg.type === "message") {
-			// Format:
-			// [2014-01-01 00:00:00] <Arnold> Put that cookie down.. Now!!
-			line += `<${msg.from.nick}> ${msg.text}`;
-		} else {
-			// Format:
-			// [2014-01-01 00:00:00] * Arnold quit
-			line += `* ${msg.from.nick} `;
+		// name the file after the channel/query, or in case of NOTICE, the sender
+		const logfilename = msg.type === Msg.Type.NOTICE ? msg.from.nick : channel.name;
```
xPaw: This isn't entirely correct as there can be notices sent to a channel.

src/plugins/messageStorage/text.js
```diff
@@ -41,26 +42,57 @@ class TextFileMessageStorage {
...
-			} else if (msg.text) {
-				line += ` ${msg.text}`;
-			}
+		// message types from src/models/msg.js
+		switch (msg.type) {
+		case Msg.Type.ACTION:
+			// [2014-01-01 00:00:00] * @Arnold is eating cookies
+			line += `* ${msg.from.mode}${msg.from.nick} ${msg.text}`;
+			break;
+		// case Msg.Type.INVITE: // neither incoming nor outgoing invites are logged
```
xPaw: Remove this line

src/plugins/messageStorage/text.js
```diff
@@ -41,26 +42,57 @@ class TextFileMessageStorage {
...
+		case Msg.Type.QUIT:
+			// [2014-01-01 00:00:00] *** Arnold (~arnold@foo.bar) quit (Connection reset by peer)
+			line += `*** ${msg.from.nick} (${msg.hostmask}) quit (${msg.text})`;
+			break;
+		case Msg.Type.CHGHOST:
+			// [2014-01-01 00:00:00] *** Arnold changed host to: new@fancy.host
+			line += `*** ${msg.from.nick} changed host to '${msg.new_ident}@${msg.new_host}'`;
+			break;
+		case Msg.Type.TOPIC:
+			// [2014-01-01 00:00:00] *** Arnold changed topic to: welcome everyone!
```
Zarthus: Comment/text is wrong.

Issue #1486, opened 2017-09-01
All servers default to UTC for logging


McInkay:
For some reason, [we default all logging on all servers to UTC](https://github.com/thelounge/lounge/blob/master/src/log.js#L10). The only way to get around this is to set a timezone in the config, which obviously won't change with DST.

Options:

1. Remove the option altogether
2. Remove it as a default config, but use the config if it is there

1 has the disadvantage of if anyone is actually using it, they won't be able to now. 2 has the disadvantage of requiring people to delete the config from their file to fix the issue.

The disadvantage of 2 does lead us to another option:

3. Create a new line in the config that has no default, but users can set if they want to.

That would mean everyone would default to their server's time, but people can set another timezone if they want. It has the disadvantage of leaving an unused config in people's files, and if anyone currently uses the setting, it'll reset it for them.

Thoughts? I'm currently thinking of going for 3, despite the horribleness of it.

YaManicKill:
Alright, this gets worse.

We have 2 different types of logging.

1. Server logging, just stdout, for errors and stuff
2. IRC logging, where we log the messages for a user

This much makes sense. However.... Server logging uses the timezone and format settings from IRC logging. There are no settings for server logging. Irc logging is per user, but the only logging setting per user is whether or not to do it. The format is the same for all users.

So... we should probably

1. Add a new config entry for server logging, that defaults the format to a sensible default (with timezone in it, unlike just now), and defaults the timezone to blank, which means we'll just use server time.
2. Make the irc logging config entirely per-user, and either default to server time or utc. Again, the time format should probably include the timezone as well by default either way, to make it clear to the users what timezone it is in if they haven't set it. They can delete the timezone from the format if they want.

Jay2k1:
This is exactly how I would do it.

Users who have (ab)used the current "logs:timezone" setting to set a custom timezone for the console logging (which differs from their server time) will be affected by this though. But I can't imagine that there's many users who did this...

YaManicKill:
There is 1 question left with this that I'm unsure about.

If a user doesn't have their user log settings set up, do we default to sensible format and server time, or what the global log setting currently is? If someone has set the global userlogging settings, we don't really want that to be lost, do we?

xPaw:
How about we just remove `timezone` option and it's code? It's going to default to server's (node.js) timezone.

astorije:
Agreed. `timezone` and `format` do not make sense **for logs** IMO. We should just log everything in ISO 8601 format.

PR #2506, opened 2018-06-03
Remove ability to change date format and timezone

xPaw:
Timezones were discussed in #1486. Removing it from config means the timezone of the server will be used.

Being able to change date format in logs makes little sence when you can't format anything else in the log.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -8,6 +8,7 @@ const fs = require("fs");
 const net = require("net");
 const bcrypt = require("bcryptjs");
 const colors = require("chalk");
+const moment = require("moment");
 
 let homePath;
 let configPath;
@@ -30,6 +31,7 @@ const Helper = {
 	setHome,
 	getVersion,
 	getGitCommit,
+	getHumanDate,
 	ip2hex,
 	mergeConfig,
 	getDefaultNick,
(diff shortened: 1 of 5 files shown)


src/helper.js
```diff
@@ -205,6 +207,10 @@ function passwordCompare(password, expected) {
 	return bcrypt.compare(password, expected);
 }
 
+function getHumanDate() {
+	return moment().format("YYYY-MM-DD HH:mm:ss");
```
astorije: Any reason why our own format instead of [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)?
I realize it breaks from current default format, but I think it would be nicer/more standard for logs.

xPaw: Our format is effectively 8601 minus the `T` separator, which is also agreed upon, for example in RFC 3339: https://tools.ietf.org/html/rfc3339#section-5.6

      NOTE: ISO 8601 defines date and time separated by "T".
      Applications using this syntax may choose, for the sake of
      readability, to specify a full-date and full-time separated by
      (say) a space character.

astorije: Fair enough.

astorije: What about returning `return moment().format()` and letting Moment's default (ISO 8601) do the job?

xPaw: I think having a space is more readable.

astorije:
Add a test before merging? 🙏

PR #2511, opened 2018-06-04
Ensure HTML minifying does not incorrectly remove some whitespaces

astorije:
#2498 was a bit too strict in what it removed. I noticed at least one instance where it removed a whitespace too strictly:

Before #2498 | After #2498
--- | ---
screen shot 2018-06-04 at 01 06 01 | screen shot 2018-06-04 at 01 06 12

My guess is that these sneaky "whitespaces as margins" can be found in multiple places in the app. I'd rather not break those silently for now.

This PR fixes that by ensuring the last whitespace to be kept.
Note that `preserveLineBreaks: true` achieves the same fix, not sure if it is actually better but I assumed not (see https://github.com/kangax/html-minifier#options-quick-reference).

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -71,6 +71,7 @@ const config = {
 				}, {
 					loader: "html-minifier-loader",
 					options: {
+						conservativeCollapse: true,
 						ignoreCustomFragments: [
 							/{{[\s\S]*?}}/,
 						],


xPaw:
Hm. Looking at how that button is rendered, this is one of things where I wanted to collapse the whitespace, and we would be better off to add the space in the `More` string. What do you think?

astorije:
Superseded by #2539.

PR #2523, opened 2018-06-07
Display visual feedback on self away/back

### LINKED ISSUE #1997: Away reply raws 305/306 not shown in the client
Currently, TL doesn't show 'you are now marked as being away' and 'you are no longer marked as being away' messages, so when you set yourself away/back, you do not get any feedback about that. This used to be different. @xPaw pointed out that this is "maybe because irc-fw handles them now, and we handle it via away event".

I've looked, and irc-fw emits 'away' and 'back' upon receiving these raws.

I've had a look into `src/plugins/irc-events/away.js` and found out that it doesn't send a message `if (!user || user.away === away)`. There is no user though, because user is defined as 
 `'chan.findUser(data.nick)`, however data.nick in this event is now the server, i.e. `tolkien.freenode.net`.

Also, where there's an on away event, there probably should be an on back event too.

Jay2k1:
Fixes #1997.

I hope I did this correctly -- ~~I introduced two new message types and two new (and identical) templates.~~

### DIFF
--- a/src/plugins/irc-events/away.js
+++ b/src/plugins/irc-events/away.js
@@ -4,8 +4,25 @@ const Msg = require("../../models/msg");
 
 module.exports = function(irc, network) {
 	const client = this;
-	irc.on("away", (data) => {
+
+	irc.on("away", (data) => handleAway(Msg.Type.AWAY, data));
+	irc.on("back", (data) => handleAway(Msg.Type.BACK, data));
+
+	function handleAway(type, data) {
 		const away = data.message;
+
+		if (data.self) {
+			const msg = new Msg({
+				self: true,
+				type: type,
+				text: away,
+				time: data.time,
+			});
+
+			network.channels[0].pushMessage(client, msg, true);
+
+			return;
+		}
 
 		network.channels.forEach((chan) => {
 			const user = chan.findUser(data.nick);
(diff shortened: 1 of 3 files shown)


Jay2k1:
image

Worth discussing/thinking about:
Should away/back messages (which go to the status window) increase the window's unread counter?
- **Pro:** otherwise you'll still get no (obvious) indication that your away/back command did something
- **Con:** If you've set TL to auto-away you upon detaching all clients, refreshing the page with only one client opened (or just connecting) will cause two of these events
  - then again, on connect you get a lot of unread counter increasing lines in status anyway

McInkay:
> Should away/back messages (which go to the status window) increase the window's unread counter?

I'd say the pro outweights the con, so yes I would do that.

However...

> which go to the status window

I feel like they should only go to the status window if it's automatic back/away. I think they should show in the active window if the user has typed them. I know that's currently difficult because we don't support message tagging but ... it's less than ideal, imo.

Jay2k1:
Fully agreed.

src/plugins/irc-events/away.js
```diff
@@ -25,4 +46,9 @@ module.exports = function(irc, network) {
 			user.away = away;
 		});
 	});
+
+	irc.on("back", (data) => {
```
xPaw: Does this actually work with away-notify?

It seems that the irc-fw code changed but we never updated it here.

Jay2k1: Yes.
image

Jay2k1:
**Status:** The code currently works although it shouldn't. There was a bug in irc-fw which has already been fixed in master (kiwiirc/irc-framework/pull/151), but we should wait for a new irc-fw release so we don't have to change the code of this PR again.

Issue #2448, opened 2018-05-22
mask-image causes aliased text


williamboman:
The mask-image property, introduced in 23c969a, on container elements is what's causing this. Not all browsers seems to have this issue (only noticed on macOS Safari). Only happens on monitors with low-medium DPI.

Difference in Safari (make sure you view in 1:1 scale):

![screenshot 2018-05-14 at 17 54 21](https://user-images.githubusercontent.com/6705160/40138771-f84aef64-594d-11e8-9b1e-7524c89b4572.png)
![screenshot 2018-05-14 at 17 54 40](https://user-images.githubusercontent.com/6705160/40138772-f8660bf0-594d-11e8-80f7-5a8d7ff36cde.png)

xPaw:
Is this still happening?

williamboman:
Nope

PR #2527, opened 2018-06-08
Add fading scroll overflows in channel list, user list, and main chat, and make user list sections sticky

astorije:
I initially did this as part of #2315 but there are small stupid issues on Firefox (and I'm expecting others in Edge), and I might be doing this entirely wrong anyway.

It actually looks pretty cool and I think it's worth that I spend more time on it, but it doesn't have to block #2315, however I'm opening this earlier to get feedback or suggestions for improvements.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -522,6 +522,14 @@ kbd {
 	touch-action: pan-y;
 	overscroll-behavior: contain;
 	-webkit-overflow-scrolling: touch;
+	-webkit-mask-image:
+		linear-gradient(to bottom, transparent, black 6px, black 60%, transparent 60%),
+		linear-gradient(to top, transparent, black 20px, black 60%, transparent 60%),
+		linear-gradient(to left, black, black 8px, transparent 8px, transparent);
+	mask-image:
+		linear-gradient(to bottom, transparent, black 6px, black 60%, transparent 60%),
+		linear-gradient(to top, transparent, black 20px, black 60%, transparent 60%),
+		linear-gradient(to left, black, black 8px, transparent 8px, transparent);
 }
 
 #sidebar .logo-container {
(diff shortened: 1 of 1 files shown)


xPaw:
I looked at this in the revamped theme PR, and I didn't like that much. Do we really need these gradients?

A sharper look is better, imo.

astorije:
Do you really find anything less sharp though? This PR shouldn't affect that. Like, anything blurry or other bugs? (I noticed there is a bug in user list with the last nick per category at the moment)

The gradients are a nice cue that there is more than what's currently visible. It's especially helpful on devices that have no scrollbars. Reasoning is similar to gradients in lieu of ellipsis.
It also helps with accessibility, especially around cognitive disorders.

That said, I am having second thoughts about the main window, especially because it scrolls from bottom, and it's the main item on the screen, focusing the attention already.
I do see a value in having this in channel list and user list though. Thoughts?

For comparison, we are using goddamn Jira at work and every week I can see someone confused about the sidebar and its scrolling content, which was truly made by the devil. A simple cue like a gradient, an icon, whatever setting up expectations and reducing surprises would make it _so much_ better.

astorije:
Yikes, thanks for noticing that @williamboman. I'll fix #2315!

astorije:
@xPaw, I extracted the sticky position into #2643 because I figured it wasn't controversial and buggy like the fading.

What's your take on this, are you opposed to this altogether, or you think it just needs to be improved? For example, I could get rid of it in main content but keep it in channel/user lists which is where I find it actually useful.
If you have other suggestions/ideas, I'm all ears. If you just think it's garbage at all, then I'll ditch it I guess 🤷‍♂️

xPaw:
@astorije This might cause more problems due to #2448.

astorije:
Fair enough. I'll close for now, can play with this again in a few <enter unit of time> when #2448 is resolved and/or browsers get their 💩 together. It would probably take me a few cycles to get the scrollbar thingy right, and I don't want to waste time if it's causing other issues I can't fix.

PR #2532, opened 2018-06-09
use http for protocol-less URLs

### LINKED ISSUE #2525: Auto-detected URLs shouldn't assume https
* *Node version:* v9.11.1
* *Browser version:* chrome latest
* *Device, operating system:* windows 10
* *The Lounge version:* source@de1e609

URLs without a scheme are now detected by linkify-it.js since #2397 was merged.

It's the server's responsibility to redirect from http->https, not the client's. This breaks many sites that do not have SSL certs where I would expect a link in thelounge would work, but instead results in a nasty `ERR_CERT_AUTHORITY_INVALID` error in the browser.

benharri:
closes #2525


dgw:
Need to update the tests too.

PR #2536, opened 2018-06-10
Focus into input if trying to type into any non-input element

### LINKED ISSUE #2473: Typing does not focus input box when a link is focussed
* *The Lounge version: source (commit 0de90da)* 
* Tested in Chrome, Firefox and Safari on macOS

---

When you click a URL, it opens in a new tab. Close the tab, you're back in TL (depending on the browser and its settings) but the input box doesn't have focus and typing doesn't focus it either. You have to click somewhere (empty space in chat/userlist/channel list or input box) to make it focus. 

*This ^ applies to Chrome and Firefox on macOS, in Safari it works as expected. This is because URLs in channels are not tab-selectable in Safari, only channel names are (another bug?).*

To reproduce, you can also shift-tab in a channel to move the focus from the input box to focusable element in the chat and then start typing.

*This ^ fails as described in all three browsers I tested, just that Safari always focuses something starting with `#` and the others select the most recent URL or channel name etc.*

I'm not sure how intentional this is, but especially in the "URL clicking" scenario it feels a bit... clunky.

xPaw:
Fixes #2473

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -247,16 +247,22 @@ const ignoredKeys = {
 	224: true, // Meta
 };
 
-$(document.body).on("keydown", (e) => {
-	// Ignore if target isn't body (e.g. focused into input)
+$(document).on("keydown", (e) => {
 	// Ignore any key that uses alt modifier
 	// Ignore keys defined above
-	if (e.target !== document.body || e.altKey || ignoredKeys[e.which]) {
+	if (e.altKey || ignoredKeys[e.which]) {
 		return;
 	}
 
 	// Ignore all ctrl keys except for ctrl+v to allow pasting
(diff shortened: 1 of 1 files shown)


client/js/keybinds.js
```diff
@@ -247,11 +247,17 @@ const ignoredKeys = {
...
+$(document).on("keydown", (e) => {
 	// Ignore any key that uses alt modifier
 	// Ignore keys defined above
-	if (e.target !== document.body || e.altKey || ignoredKeys[e.which]) {
+	if (e.altKey || ignoredKeys[e.which]) {
+		return;
+	}
+
+	const tagName = e.target.tagName;
+
```
dgw: Easy fix for the tests: Remove spaces.

dgw:
Does this also close #1428?

xPaw:
That one is closed already.

PR #2538, opened 2018-06-10
De-duplicate link fetching to one request at once

### LINKED ISSUE #1614: Await existing link expander promises

xPaw:
Converts `fetch` into a Promise, and returns existing Promise if there's one already.

Hashed by `uri` and `headers` passed in so that language is considered.

Supercedes #1952.
Fixes #1614.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -8,7 +8,7 @@ const Helper = require("../../helper");
 const cleanIrcMessage = require("../../../client/js/libs/handlebars/ircmessageparser/cleanIrcMessage");
 const findLinks = require("../../../client/js/libs/handlebars/ircmessageparser/findLinks");
 const storage = require("../storage");
-
+const currentFetchPromises = new Map();
 const mediaTypeRegex = /^(audio|video)\/.+/;
 
 // Fix ECDH curve client compatibility in Node v8/v9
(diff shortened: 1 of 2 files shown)


src/plugins/irc-events/link.js
```diff
@@ -313,71 +310,89 @@ function getRequestHeaders(headers) {
 	return formattedHeaders;
 }
 
-function fetch(uri, headers, cb) {
-	let req;
+function fetch(uri, headers) {
+	// Stringify the object otherwise the objects won't compute to the same value
+	const cacheKey = JSON.stringify({uri, headers});
```
xPaw: Probably should be an array instead of an object to reduce length of the string.

astorije: That would go from:
`'{"uri":"https://example.com","headers":{"foo":"bar"}}'`
to:
`'["https://example.com",{"foo":"bar"}]'`
Correct?

If so, that saves us 16 characters per string, I wouldn't worry too much.
I'm obviously not opposed if you prefer to do that though.

src/plugins/irc-events/link.js
```diff
@@ -313,71 +310,89 @@ function getRequestHeaders(headers) {
...
-	} catch (e) {
-		return cb(null, e);
+	if (promise) {
+		return promise;
 	}
 
-	const buffers = [];
-	let length = 0;
-	let limit = Helper.config.prefetchMaxImageSize * 1024;
+	const removeCache = () => currentFetchPromises.delete(cacheKey);
```
astorije: I'm probably missing something, but any reason why you're creating this here and not just above where it's used, L391?

test/plugins/link.js
```diff
@@ -422,4 +422,40 @@ describe("Link plugin", function() {
 
 		expect(message.previews).to.be.empty;
 	});
+
+	it("should fetch same link only once at the same time", function(done) {
```
astorije: 👏

src/plugins/irc-events/link.js
xPaw: I tried chained promises, forgot to move it back down.

src/plugins/irc-events/link.js
xPaw: Worth it!

PR #2540, opened 2018-06-10
Make non-highlighted favicon look more different than highlighted one

astorije:
This follows a discussion started in https://github.com/thelounge/thelounge/issues/2470 and hopefully solves it.

Only the "normal" version changed, but posting both to see the result:

normal
highlighted

That way:

- Keeps the existing highlighted favicon, which looks pretty dope IMO
- Respects branding guidelines along with suggested foreground/background contrasts
- Difference between normal and highlighted is more eye-catchy

I think that's worth a shot, see in a pre-release if it feels better, and if not, open to reconsider. Thoughts?


MaxLeiter:
My one thought regarding this is id love for themes to be able to choose between favicon designs. For example, the dark favicon fits morning well and the bright one fits default.

astorije:
Something like logo in-page, makes sense.
I'm not sure if we could do this currently though (i.e. without any addition to theme support) as themes are only CSS. But either way, I think that would be out of scope for this PR.

christer88:
I disagree. Favicon and logo needs to be done one way to make TheLounge stand out as one software. You don't see different versions of Discord logo, for very good reasons.

xPaw:
Let's try this for the next prerelease.

PR #2543, opened 2018-06-11
Fix multiple <title> tags being concatenated

xPaw:


test/plugins/link.js
```diff
@@ -72,7 +72,25 @@ describe("Link plugin", function() {
 		});
 
 		this.irc.once("msg:preview", function(data) {
-			expect(data.preview.head, "opengraph test");
+			expect(data.preview.head).to.equal("opengraph test");
```
xPaw: this was a broken test

PR #2544, opened 2018-06-11
Remove label on force sync button

xPaw:
The entire label was clickable, which is no good.


client/views/windows/settings.tpl
```diff
@@ -28,10 +28,10 @@
 			</label>
 			<p class="sync-warning-override"><strong>Warning</strong> Checking this box will override the settings of this client with those stored on the server.</p>
 			<p class="sync-warning-base"><strong>Warning</strong> No settings have been synced before. Enabling this will sync all settings of this client as the base for other clients.</p>
-			<label class="opt force-sync-button">
+			<div class="opt force-sync-button">
 				<button type="button" class="btn" id="forceSync">Force sync settings</button><br>
 				This will override any settings already synced to the server.
```
astorije: Can you wrap this in `<p></p>` actually? Otherwise the text has a bigger font (we'll fix the kinda messed up styling for settings later on, no need to tweak this in your PR as we would probably break something else).

PR #2547, opened 2018-06-11
Synchronise number of highlighted messages to client

### LINKED ISSUE #1791: Polish self-message handling (ref #1615)
I've noticed some possible issues with my own feature (marking buffers that receive self-messages as read) and am documenting them here for when I get a chance to look into it further. (Would someone please assign this issue to me? :smile_cat:)

In particular, I carried on a query conversation (note to self: it was BitlBee through ZNC, not a "real" IRC network) via mobile client that I then transferred to desktop & The Lounge. While the PM buffer was not marked as highlighted (good), it did still have a badge for the number of unseen messages (hmm, could still be OK), and the number was much higher than it should have been if TL was only counting messages received since my last sent line (nope, not OK).

Probably there is a code path I missed, or something. Nothing obvious, or this would be a PR already. :grin:

xPaw:
Also fixes #1791

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -48,8 +48,23 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 	// If this channel is open in any of the clients, do not increase unread counter
 	const isOpen = _.find(client.attachedClients, {openChannel: chan}) !== undefined;
 
-	if ((increasesUnread || msg.highlight) && !isOpen) {
-		obj.unread = ++this.unread;
+	if (msg.self) {
+		// reset counters/markers when receiving self-/echo-message
+		this.unread = 0;
+		this.firstUnread = 0;
+		this.highlight = 0;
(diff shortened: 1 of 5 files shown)


src/models/chan.js
```diff
@@ -48,8 +48,23 @@ Chan.prototype.pushMessage = function(client, msg, increasesUnread) {
 	// If this channel is open in any of the clients, do not increase unread counter
 	const isOpen = _.find(client.attachedClients, {openChannel: chan}) !== undefined;
 
-	if ((increasesUnread || msg.highlight) && !isOpen) {
-		obj.unread = ++this.unread;
+	if (msg.self) {
+		// reset counters/markers when receiving self-/echo-message
+		this.unread = 0;
```
xPaw: @dgw this line was missing

dgw: Ooh, nice catch. Thanks!

PR #2548, opened 2018-06-11
Fix text not having enough space in <select> with some fonts

xPaw:


client/css/style.css
```diff
@@ -837,12 +837,15 @@ kbd {
 	border-color: #79838c;
 }
 
-#windows select.input {
+#windows .input {
```
astorije: I wouldn't remove the specificity, as this is messing up a few things as far as I can see:

- The main message input is now off by 1px (that can probably easily fixed)
- The custom CSS box now has crazy lines:

Before | After
--- | ---
screen shot 2018-06-11 at 23 42 20 | screen shot 2018-06-11 at 23 42 30


Maybe wait after #2315 to remove the specificity (so I don't have to fix input in there too :D)

astorije: Also even with https://github.com/thelounge/thelounge/wiki/CSS-Modifications#bigger-custom-css-field now, custom CSS box is always 150px because this 35px get higher priority than the height in this snippet (the only one I actually use from the wiki 😂) and therefore the min 150px height applies.

PR #2550, opened 2018-06-12
Log notices as correct sender when it will be shown in active window

xPaw:


src/models/chan.js
```diff
@@ -185,9 +186,19 @@ Chan.prototype.writeUserLog = function(client, msg) {
 		return;
 	}
 
+	let targetChannel = this;
+
 	// Is this particular message or channel loggable
 	if (!msg.isLoggable() || !this.isLoggable()) {
-		return;
+		// Because notices are nasty and can be shown in activate channel on the client
```
astorije: "active", right?

dgw: Yes.

astorije: Thanks @dgw!

Issue #1679, opened 2017-11-01
Include image/file upload functionality


creesch:
This is just a feature I was thinking about that would be nice to have as things like discord, slack and irccloud have this build in. 

There are several ways I see how this can work depending on how complete we want to make this. 

1. Make lounge its own file host where the upload functionality uploads it internally and also serves. 
2. Use third part content hosters like imgur for images, github gist for code snippets, pastebin, etc. 
3. Allow people to define custom uploaders, similary to how the [screenhot application ShareX does this](https://getsharex.com/docs/custom-uploader) 

Option 1. would be pretty cool but expands the scope of the lounge and comes with various issues regarding privacy (people might run the lounge on their home network) and hosting capacity (image urls might be reshared and get really popular taking down the user's lounge machine). 

Option 2. would probably be the most trivial one to implement if it is simply anonymous uploads. This leaves people without the option to delete whatever they uploaded which would complicate the implementation somewhat and will add additional settings. 

Option 3. is the most flexible of the three however will make things complicated for the user. 

As far as the UI/UX goes in the most simple form it would simply be an upload icon where the destination is determined on the filetype and that on success returns the url in the input textarea. 

A nice UX touch could be to offer to upload user input to gist/pastebin when someone writes a very long piece of text. This is something that irccloud does and allows for very easy code sharing.

YaManicKill:
So, we do serve images from the lounge, as we cache preview images now, so we can use that.

I would suggest:

1. Built-in support only works with uploading to our internal storage and serving the same way. This will work for most people, and gives "normal" people the simplicity of it
2. Add plugin support for custom uploaders. Someone could make a plugin for imgur if they wanted, someone could make a plugin for allowing people to upload to their own servers. This gives people the ability to do all the custom stuff they want.

Having the 2 of these would be the best of both worlds. We should start off with just 1 and eventually do 2. 

Similarly, we should probably start off with only supporting uploading images as that's simplest and takes least security thoughts.

The gist idea is nice (I miss that from irccloud) but also brings in a few issues with making sure that we don't accidentally create XSS. 

> As far as the UI/UX goes in the most simple form it would simply be an upload icon where the destination is determined on the filetype and that on success returns the url in the input textarea.

Also, drag/drop onto chat would be nice.

P4k3:
I loved how it worked in irccloud. Image in clipboard paste in textinput on chatclient and it posts a link to the image.

creesch:
[...]

PR #2552, opened 2018-06-12
Add file uploading support

MaxLeiter:
Seems to be pretty ready to go.

UI can be improved in another PR.


defaults/config.js
```diff
@@ -86,6 +86,28 @@ module.exports = {
...
+	//
+	// Maximum file size for uploaded files
+	//
+	// If upload is enabled, The Lounge will only allow uploading files with a size < the value below
+	// Specified value is in kilobytes. Default value is 1e7 bytes (10 MB).
+	//
+	// @type     int
+	// @default  1e7
+	//
+	uploadMaxFileSize: 1e7,
```
xPaw: 10mb would be 10240.

1e7 is 10 gigabytes.

MaxLeiter: its in bytes not KB

src/server.js
```diff
@@ -281,6 +284,11 @@ function initializeClient(socket, client, token, lastMessage) {
 
 	client.clientAttach(socket.id, token);
 
+	if (Helper.config.upload) {
+		const upload = require("./plugins/upload");
```
xPaw: require it once at the top.

defaults/config.js
xPaw: Your comment says kilobytes.

The other max size we have in the config is also KB, so do the math in the code.

src/plugins/upload.js
```diff
@@ -0,0 +1,35 @@
...
+		client.emit("upload:success", url);
+	});
+
+	siofu.dir = Helper.getFileUploadPath();
+	siofu.maxFileSize = Helper.config.uploadMaxFileSize;
+	siofu.listen(socket);
+}
+
+/** https://stackoverflow.com/a/12900504 **/
+function getExtension(path) {
```
xPaw: Use `path.extname`.

client/js/upload.js
```diff
@@ -0,0 +1,53 @@
...
+	$("#loading-page-message, #connection-error").text(event.message);
+});
+
+/** https://css-tricks.com/drag-and-drop-file-uploading/ **/
+function isDragDropSupported() {
+	const div = document.createElement("div");
+	return (("draggable" in div) || ("ondragstart" in div && "ondrag" in div)) && "FormData" in window && "FileReader" in window;
+}
+
+if (isDragDropSupported()) {
```
xPaw: This check isn't really necessary because you're only binding events in the `if`, and they simply won't do anything if there's no support.

Also the fact that all the evergreen browsers have support for it.

client/js/upload.js
```diff
@@ -0,0 +1,53 @@
...
+	const $form = $("#chat-container");
+	const $overlay = $("#chat-container-overlay");
+
+	$form.on("drag dragstart dragend dragover dragenter dragleave", function(e) {
+		e.preventDefault();
+		e.stopPropagation();
```
xPaw: These 2 functions are shorthanded by `return false` in jQuery.

client/js/upload.js
```diff
@@ -0,0 +1,53 @@
...
+instance.addEventListener("progress", function(event) {
+	const percent = `${((event.bytesLoaded / event.file.size) * 100)}%`;
+	$("#upload-progressbar").width(percent);
+});
+
+instance.addEventListener("error", function(event) {
+	// Reset progressbar
+	$("#upload-progressbar").width(0);
+	$("#loading-page-message, #connection-error").addClass("shown");
+	$("#loading-page-message, #connection-error").text(event.message);
```
xPaw: Chain `text` after `addClass`. You also don't need to target `#loading-page-message`.

MaxLeiter: I know I don't need to, but its targeted elsewhere so did this for consistency

xPaw: I know what you mean, but in other places it makes sense because the error can be shown before the app fully loads.

In this case uploads only work after the app loaded, so there is no loading page.

src/server.js
```diff
@@ -46,6 +47,9 @@ module.exports = function() {
 		.use("/storage/", express.static(Helper.getStoragePath(), {
 			redirect: false,
 			maxAge: 86400 * 1000,
+		})).use("/uploads/", express.static(Helper.getFileUploadPath(), {
```
xPaw: We'll want to figure something to make most files have `plain/text` content type, because we don't want to make js/css/etc files arrive with their respective content types.

client/index.html.tpl
```diff
@@ -76,12 +76,18 @@
...
 				<div id="chat-container" class="window">
+					<div id="chat-container-overlay"></div>
 					<div id="chat"></div>
 					<div id="connection-error"></div>
+					<span id="upload-progressbar"></span>
 					<form id="form" method="post" action="">
 						<div class="input">
 							<span id="nick"></span>
 							<textarea id="input" class="mousetrap"></textarea>
+							<span id="submit-tooltip" class="tooltipped tooltipped-w tooltipped-no-touch" aria-label="Upload file">
```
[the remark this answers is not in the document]
MaxLeiter: Done

christer88:
https://gfycat.com/ZanyUnsteadyAndalusianhorse was linked earlier on IRC. Shows the lovely uploading feature!

dgw:
This excites me so much, even though I probably wouldn't use the feature as-is because of the dependency on The Lounge and its instance's address for serving the images.

The future may bring writing to a custom uploads directory (on a network mount, for example) and returning a custom URL based on that, so uploads can go somewhere besides The Lounge's server if wanted/needed. Maybe it will even support a few third-party hosts (Imgur, etc.) and/or standalone self-hosted image/file servers (Pomf, ownCloud, etc.) via plugins, who knows.

My point is, much of the truly heavy lifting is already in this PR, and I'm positively tingling with excitement for the future of this function. 🎉👏 @MaxLeiter! (Thanks to @christer88 for passing along the video clip of this in action, too!)

Man, it's a great time to be an IRC user. My former IRC friends who've jumped on the Discord bandwagon don't know the Renaissance they're missing! (Even if it is services like Discord that are pushing IRC clients to step up… 😉)

src/server.js
dgw: Any thoughts on sending `Content-Disposition: inline` as well to encourage browsers to display instead of prompting to save? (If The Lounge saves the original filename anywhere, that could be passed along in this header too, I think.)

defaults/config.js
```diff
@@ -86,6 +86,28 @@ module.exports = {
...
 	prefetchStorage: false,
 
+	//
+	// Allow uploading files to the server hosting TheLounge.
+	// If enabled, The Lounge will store files in the `${THELOUNGE_HOME}/storage` folder.
+	//
+	// Images will not expire or be removed by TheLounge, so be wary of memory and your hardware.
+	//
+	// @type     boolean
+	// @default  false
```
Zarthus: wrong default.

client/js/socket-events/configuration.js
```diff
@@ -30,6 +30,10 @@ socket.on("configuration", function(data) {
 		pop.play();
 	});
 
+	if (!data.upload) {
+		$("#upload").hide();
```
Zarthus: What about making it hidden by default and only showing it if it's enabled? (in event of js error it doesn't look strange then)

xPaw: The UI is not visible at this stage.

MaxLeiter: It's in the index.html, so it will be rendered by default regardless. The alternatives would be to use an underscore template, but that exposes the config option of whether or not it's enabled or not.

defaults/config.js
MaxLeiter: No -- wrong value imo. I think the default should be false, as this is unwise on public instances at the moment.

Zarthus: Agreed :)

src/server.js
```diff
@@ -48,6 +50,50 @@ module.exports = function() {
...
+		case "video/ogg":
+		case "video/mp4":
+			break;
+		default:
+			res.set("content-type", "text/plain");
+		}
+
+		if (!uploadedFile) {
+			return res.status(404).send("Not found");
+		}
```
MaxLeiter: Should move this check above setting header so

SinZ163:
What happens when a filename conflict happens?

MaxLeiter:
The library automatically renames the file with a number (-0, -1, etc)

src/plugins/upload.js
```diff
@@ -0,0 +1,44 @@
...
+			if (err) {
+				callback(err);
+			} else {
+				siofu.dir = folder;
+				callback(true);
+			}
+		});
+	};
+
+	siofu.on("complete", (data) => {
```
xPaw: Handle `error` event (and possibly others? need to look at the source code) so we don't crash.

src/server.js
```diff
@@ -48,6 +50,50 @@ module.exports = function() {
...
+		case "image/gif":
+		case "image/jpg":
+		case "image/jpeg":
+		case "image/webp":
+		case "video/webm":
+		case "video/ogg":
+		case "video/mp4":
+			break;
+		default:
+			res.set("content-type", "text/plain");
```
xPaw: If i upload an `exe`, will it get printed in the browser?

src/server.js
```diff
@@ -48,6 +50,50 @@ module.exports = function() {
 			maxAge: 86400 * 1000,
 		}));
 
+	app.get("/uploads/:file", (req, res) => {
```
xPaw: Perhaps we should move this handler to `uploads.js`.

src/server.js
```diff
@@ -48,6 +50,50 @@ module.exports = function() {
...
+		const fileName = req.params.file;
+		const uploadPath = Helper.getFileUploadPath();
+		const contentType = mime.contentType(path.extname(fileName));
+		const filePath = path.join(uploadPath, fileName);
+		const uploadedFile = upload.ensureFileExists(filePath);
```
xPaw: Just use `fs.existsSync` directly

src/plugins/upload.js
```diff
@@ -0,0 +1,44 @@
...
+		const base = data.file.base;
+		const extension = path.extname(data.file.pathName);
+		const url = `/uploads/${base}${extension}`;
```
xPaw: Remove leading slash.

src/plugins/upload.js
```diff
@@ -0,0 +1,44 @@
+"use strict";
+
+const SocketIOFileUploadServer = require("socketio-file-upload/server");
+const Helper = require("../helper");
+const path = require("path");
+const fsextra = require("fs-extra");
+const fs = require("fs");
+
+function init(socket, client) {
```
xPaw: Convert this file to be a `class`

src/plugins/upload.js
```diff
@@ -0,0 +1,44 @@
...
+		const url = `/uploads/${base}${extension}`;
+		client.emit("upload:success", url);
+	});
+
+	// maxFileSize is in bytes, but config option is passed in as KB
+	siofu.maxFileSize = Helper.config.uploadMaxFileSize * 1024;
+	siofu.listen(socket);
+}
+
+function ensureFileExists(filePath) {
```
xPaw: `return fs.existsSync(filePath);`

src/server.js
MaxLeiter: package and theme serving also lives here, fwiw.

xPaw: I know, but we should move it out at some point.

defaults/config.js
MaxLeiter: fixed

src/server.js
```diff
@@ -48,6 +49,25 @@ module.exports = function() {
 			maxAge: 86400 * 1000,
 		}));
 
+	app.get("/uploads/:name/:slug*?", (req, res) => {
```
MaxLeiter: https://stackoverflow.com/questions/6161567/express-js-wildcard-routing-to-cover-everything-under-and-including-a-path

client/js/socket-events/uploads.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+const $ = require("jquery");
+const socket = require("../socket");
+const wrapCursor = require("undate").wrapCursor;
+
+socket.on("upload:success", (url) => {
+	const fullURL = new URL(url, location);
```
xPaw: Let's future proof this and check if `url` starts with `http`. If it does then it's a full url.

src/plugins/uploader.js
```diff
@@ -0,0 +1,89 @@
...
+
+		try {
+			buffer = readChunk.sync(filePath, 0, 4100);
+		} catch (e) {
+			if (e.code === "ENOENT") { // doesn't exist
+				return;
+			}
+
+			log.warning(`Failed to read ${filePath}`);
+			return;
```
xPaw: Change these 2 returns to `null` to match `fileType`.

src/plugins/uploader.js
```diff
@@ -0,0 +1,89 @@
...
+		});
+
+		// maxFileSize is in bytes, but config option is passed in as KB
+		siofu.maxFileSize = Helper.config.uploadMaxFileSize * 1024;
+		siofu.listen(socket);
+
+		// Returns new file name
+		function handleSaving(data) {
+			const tempPath = path.join(Helper.getFileUploadPath(), ".tmp", data.file.name);
+			const randomName = crypto.randomBytes(16).toString("hex");
```
xPaw: I think we can get away with shorter random strings. 8 bytes perhaps?

src/plugins/uploader.js
```diff
@@ -0,0 +1,89 @@
...
+			const destPath = path.join(Helper.getFileUploadPath(), randomName);
```
xPaw: Take first 2 bytes of the string and make it a folder so that it's split into subfolders like preview storage does it.

However keep just the hash in URL (no need to show subfolder to the user).

src/server.js
```diff
@@ -48,6 +49,25 @@ module.exports = function() {
...
+		const name = req.params.name;
```
xPaw: Add a length (and perhaps a-z0-9?) check to match the hash format we generate. This will avoid trying to request `.tmp` folder as well.

MaxLeiter: Im a little hesitant to enforce too strict rules regarding this because plugins may want to do something differently and there’s no easy way for them to interface with express. But that’s a ways off, so I’m fine with it for now

xPaw: That's one of the reasons I wanted you to move this route to the upload plugin, because I expect other plugins to provide their own route if they need it. I expect other plugins to provide custom URLs though (imgur, etc).

src/plugins/uploader.js
```diff
@@ -0,0 +1,89 @@
...
+
+			// Try again
+			if (fs.exists(destPath)) {
+				return handleSaving(data);
```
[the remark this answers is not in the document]
MaxLeiter: should be `while (!fs.exists(destPath))`, right?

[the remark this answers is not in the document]
xPaw: No, you want it to keep trying to find a filename if one it generated exists already.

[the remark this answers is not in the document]
MaxLeiter:
```
path.js:57
function normalizeString(path, allowAboveRoot, separator, isPathSeparator) {
                        ^

RangeError: Maximum call stack size exceeded
    at normalizeString (path.js:57:25)
    at Object.normalize (path.js:1133:12)
    at Object.join (path.js:1168:18)
    at handleSaving (/Users/maxleiter/Documents/lounge/src/plugins/uploader.js:51:26)
```

src/server.js
MaxLeiter: Is `[0-9a-fA-F]+` fine with you?

Need to figure out best way to access express in `uploader.js` in order to handle routing there

src/plugins/uploader.js
MaxLeiter: Done

src/plugins/uploader.js
MaxLeiter: Done

src/plugins/uploader.js
MaxLeiter: Done

src/plugins/uploader.js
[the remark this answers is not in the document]
MaxLeiter: Fixed

src/server.js
MaxLeiter: Fixed with what we discussed on IRC

src/plugins/upload.js
MaxLeiter: Obsolete now

src/plugins/upload.js
MaxLeiter: Done

src/plugins/upload.js
MaxLeiter: Done

src/server.js
MaxLeiter: Now will be served as an attachment

src/plugins/upload.js
MaxLeiter: Done

src/server.js
```diff
@@ -48,9 +49,41 @@ module.exports = function() {
...
+		let contentDisposition = "attachment";
+
+		// doesn't exist
+		if (fileType === undefined) {
+			return res.status(404).send("Not found");
+		}
+
+		// If whitelisted
+		if (fileType && fileType.mime && Uploader.isValidType(fileType.mime)) {
+			mimeType = fileType.mime;
```
xPaw: I think `mimeType` should be set regardless. And `Uploader.isValidType` will control `contentDisposition`.

Are we going to look into using `is-utf8` as we discussed?

src/plugins/uploader.js
xPaw: You haven't done this.

src/server.js
xPaw: For the route, you can do `Uploader.router(app);`

MaxLeiter:
Do you mean move most of this code to getFileType?

src/plugins/uploader.js
```diff
@@ -0,0 +1,121 @@
...
+					siofu.dir = folder;
+					callback(true);
+				}
+			});
+		};
+
+		siofu.on("complete", (data) => {
+			const randomFileName = handleSaving(data);
+
+			if (!randomFileName) {
```
xPaw: This `if` doesn't appear to be necessary.

src/plugins/uploader.js
xPaw: Nevermind, you actually have. My bad!

Jay2k1:
Referencing #1679 which is a feature request for this

client/js/socket-events/init.js
```diff
@@ -29,6 +29,7 @@ socket.on("init", function(data) {
 
 	$("#connection-error").removeClass("shown");
 	$(".show-more button, #input").prop("disabled", false);
+	$("#upload").show();
```
xPaw: Is this gonna show itself up upon reconnect if uploads are disabled?

client/js/socket-events/uploads.js
```diff
@@ -0,0 +1,15 @@
...
+	const textbox = document.getElementById("input");
+	wrapCursor(textbox, fullURL, "");
+});
+
+socket.on("upload:failure", () => {
```
xPaw: This is never emitted.

MaxLeiter:
Plugins may choose to emit it?

client/js/socket-events/uploads.js
MaxLeiter: Leave it in for plugins?

src/plugins/uploader.js
```diff
@@ -0,0 +1,156 @@
...
+			const tempPath = path.join(Helper.getFileUploadPath(), ".tmp", data.file.name);
+			let randomName, destPath;
+
+			// If file conflicts
+			do {
+				randomName = crypto.randomBytes(8).toString("hex");
+				destPath = path.join(Helper.getFileUploadPath(), randomName.substring(0, 2), randomName);
+			} while (fs.exists(destPath));
+
+			fsextra.moveSync(tempPath, destPath, (err) => {
```
xPaw: Let's use the async version. It should be really simple as adding `callback` and calling it with `randomName` in this move callback.

client/js/socket-events/uploads.js
xPaw: Let's worry about that later, remove it for now.

client/index.html.tpl
```diff
@@ -76,12 +76,18 @@
 					<script async src="js/loading-error-handlers.js"></script>
 				</div>
 				<div id="chat-container" class="window">
+					<div id="chat-container-overlay"></div>
```
xPaw: Move this overlay to body, next to the image viewer. Keeping it here means you can't start dragging over the channel list.

And give this a better id like `upload-overlay` or something.

src/plugins/uploader.js
```diff
@@ -0,0 +1,156 @@
...
+					log.error(`Could not move ${tempPath} to ${destPath}.`);
+				}
+			});
+
+			return randomName;
+		}
+	}
+
+	static isValidType(mimeType) {
+		const whitelist = [
```
xPaw: Move this whitelist to the top (after requres)

src/plugins/uploader.js
```diff
@@ -0,0 +1,156 @@
...
+		let type;
+
+		try {
+			buffer = readChunk.sync(filePath, 0, 4100);
+		} catch (e) {
+			if (e.code === "ENOENT") { // doesn't exist
+				return;
+			}
+
+			log.warning(`Failed to read ${filePath}`);
```
xPaw: It's log.warn not log.warning.

client/index.html.tpl
MaxLeiter: I think you can start dragging over channel lost, as long as you drag on to chat container after, right? The onDrop listener is only on the chat container, not the channel list as well

xPaw: Yeah  change the onDrop to be on the entire document. It's rather weird UX to drag a file into Lounge and have it opened in that tab instead of being uploaded.

client/js/socket-events/uploads.js
MaxLeiter: done

client/index.html.tpl
MaxLeiter: done

src/plugins/uploader.js
MaxLeiter: Fixed

src/plugins/uploader.js
MaxLeiter: Fixed

xPaw:
Some things to test/do before shipping this feature:

- How is the upload handled when you try to drag multiple files onto the window
 - What about selecting multiple files in the file select?
- What happens when you try to drag a file (link) from another browser tab
- Should we have a confirmation dialog before doing the actual upload? We can also render a small thumbnail using browser technology.

client/js/socket-events/init.js
```diff
@@ -31,6 +31,10 @@ socket.on("init", function(data) {
 	$(".show-more button, #input").prop("disabled", false);
 	$("#submit").show();
 
+	if (data.uploads) {
+		$("#upload").show();
```
xPaw: As discussed in IRC, let's keep hide/show in configuration event.

client/js/socket.js
```diff
@@ -11,6 +11,12 @@ const socket = io({
 	reconnection: !$(document.body).hasClass("public"),
 });
 
+$("#loading-page-message, #connection-error").on("click", function() {
+	if ($(this).hasClass("shown")) {
```
xPaw: Unnecessary check.

client/js/upload.js
```diff
@@ -0,0 +1,42 @@
...
+	$("#upload-progressbar").width(percent);
+});
+
+instance.addEventListener("error", (event) => {
+	// Reset progressbar
+	$("#upload-progressbar").width(0);
+	$("#connection-error").addClass("shown").text(event.message);
+});
+
+const $form = $("#chat-container");
```
xPaw: I think this target should be `$(document)`.

defaults/config.js
```diff
@@ -86,6 +86,28 @@ module.exports = {
...
+	//
+	upload: false,
+
+	//
+	// Maximum file size for uploaded files
+	//
+	// If upload is enabled, The Lounge will only allow uploading files with a size < the value below
+	// Specified value is in kilobytes. Default value is 10240 bytes (10 MB).
```
xPaw: Default value is *kilo*bytes.

Also probably should be 10000 since it gets multiplied by 1024? (I know I said 10240 myself before).

src/plugins/uploader.js
```diff
@@ -0,0 +1,154 @@
...
+
+class Uploader {
+	constructor(client, socket) {
+		const siofu = new SocketIOFileUploadServer();
+
+		siofu.uploadValidator = function(event, callback) {
+			const folder = path.join(Helper.getFileUploadPath(), ".tmp");
+			fsextra.ensureDir(folder, (err) => {
+				if (err) {
+					callback(err);
```
xPaw: Any reason this folder creator is in `uploadValidator` and not simply in the constructor?

And looking at the library code, err shouldn't be passed through to correctly abort the error: 
https://github.com/sffc/socketio-file-upload/blob/c1946fe3e43bd0692100c350c93771ec01ac0823/server.js#L314-L316

src/plugins/uploader.js
```diff
@@ -0,0 +1,154 @@
...
+					siofu.dir = folder;
+					callback(true);
+				}
+			});
+		};
+
+		siofu.on("complete", (data) => {
+			handleSaving(data).then((randomName) => {
+				const randomFileName = randomName;
+				const slug = data.file.name;
```
xPaw: I think `file.base` should be the cleaned up and safer name, so we probably should use that.

src/plugins/uploader.js
```diff
@@ -0,0 +1,154 @@
...
+
+			return fsextra.move(tempPath, destPath)
+				.then(() => randomName).catch(() => {
+					log.warn(`Unable to move file ${tempPath} to ${destPath}`);
+				});
+		}
+	}
+
+	static isValidType(mimeType) {
+		return whitelist.indexOf(mimeType) > -1;
```
xPaw: We can do `return whitelist.includes(mimeType);` here.

src/plugins/uploader.js
```diff
@@ -0,0 +1,154 @@
...
+
+	static router(express) {
+		express.get("/uploads/:name/:slug*?", (req, res) => {
+			const name = req.params.name;
+			const folder = req.params.name.substring(0, 2);
+			const uploadPath = Helper.getFileUploadPath();
+			const filePath = path.join(uploadPath, folder, name);
+			const nameRegex = /^[0-9a-f]{16}$/;
+
+			if (!nameRegex.test(name)) {
```
xPaw: Move this `if` higher, right after defining `name` const.

src/plugins/uploader.js
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	static isValidType(mimeType) {
+		return whitelist.indexOf(mimeType) > -1;
+	}
+
+	static router(express) {
+		express.get("/uploads/:name/:slug*?", (req, res) => {
+			const name = req.params.name;
+			const folder = req.params.name.substring(0, 2);
```
xPaw: `req.params` is unnecessary here as you have it in `name` variable.

src/plugins/uploader.js
MaxLeiter: No, except it exists in the library for basically this purpose. I can do either way

xPaw: I think we will be fine just creating the folder in the constructor.

Instead of error callback, print a warning to the console.

MaxLeiter: if you want a warning in console on error, do you mean I should leave the `ensureDir` in the uploadValidator? If we're creating the directory in the constructor, then no need to override uploadValidator.

MaxLeiter: Nevermind, got it

defaults/config.js
MaxLeiter: As other config values are specific (prefetch is 2048), i'll instead multiply by 1000 for consistency.

src/plugins/uploader.js
MaxLeiter: We don't want file extensions?

client/js/socket-events/init.js
MaxLeiter: Done

client/js/upload.js
MaxLeiter: Done

defaults/config.js
MaxLeiter: Done (my comment)

src/plugins/uploader.js
MaxLeiter: done

client/js/upload.js
```diff
@@ -0,0 +1,42 @@
...
+const $overlay = $("#upload-overlay");
+
+$form.on("drag dragstart dragend dragover dragenter dragleave", () => false);
```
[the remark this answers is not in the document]
MaxLeiter: Done

MaxLeiter:
xPaw: Followed your comments, along the way the overlay broke, and I cant for the life of me figure it out. Will look into more once other comments addressed.

defaults/config.js
astorije: I think it's fine to start with until it gets stable, but I think in the long run we should set this to `true` by default:

- Config sets instance in private mode by default
- For discoverability, I think we should have features enabled by default with a way to disable if necessary, but not the other way around (I have the same opinion about prefetch)

astorije:
> Should we have a confirmation dialog before doing the actual upload? We can also render a small thumbnail using browser technology.

Not sure what you mean by browser technology.
I'm not sure a dialog is necessary, but I think it would be nice to have (oh god how am I going to explain this clearly) thumbnail(s) of what just got uploaded above the message input, each with a &times; button in the corner. When clicked, the thumbnail disappears, and on the backend we destroy the upload.

Either way, I think this is waaay beyond your PR :) I'd be absolutely fine with what you currently have: no confirmation, no UI, etc. Just the bare feature, and then we build up on it!

MaxLeiter:
I think xPaw meant we can show the file before its uploaded in the dialog client-side (so we can display the preview without it being uploaded).

I'm all for no confirmation UI this PR :)

xPaw:
> I think, for the initial version, we may want to display an error (using disconnect GUI) saying multiple file uploads is not yet supported?

Fine by me.

client/js/upload.js
```diff
@@ -0,0 +1,42 @@
...
+instance.addEventListener("error", (event) => {
+	// Reset progressbar
+	$("#upload-progressbar").width(0);
+	$("#connection-error").addClass("shown").text(event.message);
+});
+
+const $form = $(document);
+const $overlay = $("#upload-overlay");
+
+$form.on("drag dragstart dragend dragleave", () => false);
```
xPaw: Remove `dragleave` here.

And add `return false` in the next 2 events, I guess that's why you had them multiplied.

client/js/upload.js
```diff
@@ -0,0 +1,42 @@
...
+
+$form.on("dragover dragenter", () => {
```
xPaw: Do we actually need `dragover` and `dragleave` events?

Looking at MDN tells me we kind of don't.

https://developer.mozilla.org/en-US/docs/Web/Events/dragenter
https://developer.mozilla.org/en-US/docs/Web/Events/dragover
https://developer.mozilla.org/en-US/docs/Web/Events/dragend
https://developer.mozilla.org/en-US/docs/Web/Events/dragleave

MaxLeiter: Nope

client/js/upload.js
MaxLeiter: Done

astorije:
> The idea is to review the file you have dragged in before being automatically uploaded. I have tried at least 3 different chats, and they all asked.
> 
> But I guess the fact we put an URL in input instead of sending it also helps here.

Both statements make sense to me, and I agree. I think we're all on the same page, and also that it should go in a future PR, so I'm not going to spend too much time thinking about UI/UX for this just now. But thanks for clarifying!

client/css/style.css
```diff
@@ -296,6 +297,7 @@ kbd {
 #footer .settings::before { content: "\f013"; /* http://fontawesome.io/icon/cog/ */ }
 #footer .help::before { content: "\f059"; /* http://fontawesome.io/icon/question/ */ }
 
+#form #upload::before { content: "\f093"; /* http://fontawesome.io/icon/upload/ */ }
```
astorije: Super minor: URL is https://fontawesome.com/icons/upload?style=solid

However, I'm not a huge fan of this icon, not very pretty. I went through the ones FA provide and I think the ones below look better (more or less ordered by preference):

- https://fontawesome.com/icons/paperclip?style=solid
- https://fontawesome.com/icons/cloud-upload-alt?style=solid
- https://fontawesome.com/icons/file-image?style=solid
- https://fontawesome.com/icons/share-square?style=solid
- https://fontawesome.com/icons/desktop?style=solid or https://fontawesome.com/icons/laptop?style=solid (maybe far fetched)

What do you think?

astorije:
@MaxLeiter, I'm getting the following when I test your PR with `./scripts/run_pr.sh 2552`:

```
.../thelounge/src/server.js:617
	config.uploadMaxFileSize = uploader.getMaxFileSize();
	                                    ^

TypeError: Cannot read property 'getMaxFileSize' of null
    at getClientConfiguration (.../thelounge/src/server.js:617:38)
    at initClient (.../thelounge/src/server.js:638:32)
    at authCallback (.../thelounge/src/server.js:688:3)
    at Socket.performAuthentication (.../thelounge/src/server.js:702:11)
```

src/server.js
```diff
@@ -603,6 +614,8 @@ function getClientConfiguration(network) {
 		config.defaults.nick = Helper.getDefaultNick();
 	}
 
+	config.uploadMaxFileSize = uploader.getMaxFileSize();
```
astorije: If I read this correctly, when `Helper.config.upload` is `false` (or not set in the config, i.e. coming from defaults which are `false`), `uploader` hasn't been set so it's `undefined` 💥

MaxLeiter:
@astorije: fixed

@astorije, @xPaw:
Moved upload code to an `init()` function called in the configuration socket event. This way the file upload overlay won't show and SIOFU wont be instantiated if its not enabled on the server.

client/js/socket-events/configuration.js
```diff
@@ -30,6 +31,14 @@ socket.on("configuration", function(data) {
 		pop.play();
 	});
 
+	if (data.upload) {
+		$("#upload").show();
```
xPaw: Just realized that this show/hide needs to be moved before `if (options.initialized)`.

Keep the init here though.

client/js/socket.js
```diff
@@ -11,6 +11,10 @@ const socket = io({
 	reconnection: !$(document.body).hasClass("public"),
 });
 
+$("#loading-page-message, #connection-error").on("click", function() {
```
xPaw: Remove `#loading-page-message` here, we don't want it clickable.

client/css/style.css
xPaw: `paperclip` makes sense

client/js/socket-events/configuration.js
MaxLeiter: done

client/js/upload.js
```diff
@@ -0,0 +1,54 @@
+"use strict";
+
+const $ = require("jquery");
+const socket = require("./socket");
+const SocketIOFileUpload = require("socketio-file-upload/client");
+const instance = new SocketIOFileUpload(socket);
+
+function init() {
```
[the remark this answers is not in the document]
MaxLeiter: done

client/js/socket.js
MaxLeiter: done

src/plugins/uploader.js
```diff
@@ -0,0 +1,157 @@
...
+	"audio/x-mpeg",
+	"audio/x-mpeg-3",
+	"audio/wave",
+	"audio/wav",
+	"audio/x-wav",
+	"audio/x-pn-wav",
+	"audio/webm",
+	"video/webm",
+	"audio/ogg",
+	"video/ogg ",
```
xPaw: Hey there's a space here

MaxLeiter: I did this on purpose to differentiate. I can remove if youd prefer

xPaw: Differentiate what? Having a space in the string means this type won't be found.

MaxLeiter:
Relies on #2592

client/css/style.css
MaxLeiter: Done

src/plugins/uploader.js
```diff
@@ -0,0 +1,146 @@
...
+const path = require("path");
+const fsextra = require("fs-extra");
+const fs = require("fs");
+const fileType = require("file-type");
+const readChunk = require("read-chunk");
+const crypto = require("crypto");
+const isUtf8 = require("is-utf8");
+const log = require("../log");
+
+const whitelist = [
```
McInkay: The plan is to open up this whitelist later on, or allow it to be turned off, yeah?

defaults/config.js
McInkay: I mean, discoverability will be better when we get #1950 in and we can add the server settings to the frontend for admins.

defaults/config.js
```diff
@@ -86,6 +86,28 @@ module.exports = {
...
+	//
+	// Maximum file size for uploaded files
+	//
+	// If upload is enabled, The Lounge will only allow uploading files with a size < the value below
+	// Specified value is in kilobytes. Default value is 10240 kilobytes (10 MB).
+	//
+	// @type     int
+	// @default  10240
+	//
+	uploadMaxFileSize: 10240,
```
McInkay: Can we allow this to be set to `-1` to be disabled?

package.json
```diff
@@ -44,19 +44,23 @@
     "cheerio": "0.22.0",
     "commander": "2.16.0",
     "express": "4.16.3",
+    "file-type": "8.0.0",
```
McInkay: Just before you merge, it's probably worth checking these are up to date, as I'm pretty sure we only get PRs for updates after it's merged in.

src/plugins/uploader.js
```diff
@@ -0,0 +1,146 @@
...
+	"image/webp",
+	"text/plain",
+	"video/mp4",
+	"video/ogg",
+	"video/webm",
+];
+
+class Uploader {
+	constructor(client, socket) {
+		const siofu = new SocketIOFileUploadServer();
```
McInkay: Can we rename this? It's pretty confusing.

defaults/config.js
MaxLeiter: Previews dont have a max size, but sure.

src/plugins/uploader.js
MaxLeiter: Up for discussion. It's tied to the plugin, so future plugins can use their own.

src/plugins/uploader.js
MaxLeiter: I disagree, it's pretty self-explanatory. It's what the SIOFU docs use, so I think thats a fine reason for keeping as-is. I don't care too much if you feel strongly.

package.json
MaxLeiter: Yep, updated some. Good call.

src/plugins/uploader.js
McInkay: It's self explanatory after you go up and read the name and go "oh, I guess that's what it means" but it's really quick to forget what that means. But if you read it down you have 0 idea, no clue whatsoever what it is.

defaults/config.js
McInkay: What do you mean?

src/plugins/uploader.js
McInkay: Yeah, I wasn't meaning for right now, but it can be a discussion later. Cool beans.

defaults/config.js
MaxLeiter: Previews have a max fetch size, but we have no way of setting unlimited. So this just isnt very consistent

src/plugins/uploader.js
MaxLeiter: My argument is I think anyone editing the file should have an idea of what it is, and this way we stay consistent with the docs

defaults/config.js
McInkay: Ah right, well we can leave it for just now if we don't do it anywhere else.

src/plugins/uploader.js
McInkay: I disagree.

I think that the "this should be slightly confusing so that you know what you are doing" is a bad trope that exists in software development that I think we should be getting rid of. It's an elitist view, imo, which is really not great for software development as a whole. We should be making things easier for people, not more difficult.

And there's no need to stay consistent with the docs, the docs have named something in a confusing way, let's get rid of that.

MaxLeiter: Fixed

defaults/config.js
```diff
@@ -139,6 +139,29 @@ module.exports = {
 	// This value is set to `2048` kilobytes by default.
 	prefetchMaxImageSize: 2048,
 
+	// ### `upload`
```
MaxLeiter: @astorije: if you could review this file that'd be great

package.json
```diff
@@ -44,19 +44,23 @@
     "cheerio": "0.22.0",
     "commander": "2.17.1",
     "express": "4.16.3",
+    "file-type": "8.1.0",
```
xPaw: Update to last version

package.json
```diff
@@ -44,19 +44,23 @@
...
     "mime-types": "2.1.19",
     "moment": "2.22.2",
     "package-json": "5.0.0",
     "primer-tooltips": "1.5.7",
     "read": "1.0.7",
+    "read-chunk": "2.1.0",
     "request": "2.88.0",
     "semver": "5.5.0",
     "socket.io": "2.1.1",
+    "socketio-file-upload": "0.6.1",
```
xPaw: Update to last version

MaxLeiter: done

defaults/config.js
```diff
@@ -139,6 +139,29 @@ module.exports = {
...
+	//
+	// Allow uploading files to the server hosting TheLounge.
+	// If enabled, The Lounge will store files in the `${THELOUNGE_HOME}/uploads` folder.
+	//
+	// Images will not expire or be removed by TheLounge, so be wary of memory and your hardware.
```
astorije: Typo: `TheLounge` (and at L144).

Instead of `so be wary of memory and your hardware`, I would go with `which can cause issues depending on your hardware, for example in terms of disk usage.` No big deal though.

defaults/config.js
astorije: Nitpick: Do we want to call this `fileUpload`? `upload` alone could be confusing, especially as we add new features in the future. 🤷‍♂️

defaults/config.js
```diff
@@ -139,6 +139,29 @@ module.exports = {
...
+	//
+	// This value is set to `false` by default.
+	upload: false,
+
+	// ### `uploadMaxFileSize`
+	//
+	// Maximum file size for uploaded files
+	//
+	// If upload is enabled, The Lounge will only allow uploading files with a size < the value below
```
astorije: This seems redundant with the line above. Instead, I would phrase it to explain what happens when a file above this limit is sent. Something like:
`When file upload is enabled, users sending files above this limit will be prompted an error message in their browser.` (I have no clue if this is remotely accurate as I have not reviewed the rest of the PR yet, it's just to illustrate).

defaults/config.js
```diff
@@ -139,6 +139,29 @@ module.exports = {
...
+	// Specified value is in kilobytes. Default value is 10240 kilobytes (10 MB).
+	//
+	// `-1` will disable the file size limit and will allow files of any size. Use at your own risk.
+	//
+	// This value is set to `10240` by default.
```
astorije: You are specifying this twice.

defaults/config.js
```diff
@@ -139,6 +139,29 @@ module.exports = {
...
+	//
```
astorije: Remove this empty line

defaults/config.js
astorije: So, an alternative to:

```
fileUpload: true,
fileUploadMaxSize: 10240,
```

is to nest it like we do in a few other places:

```
fileUpload: {
  enabled: true,
  maxFileSize: 10240,
}
```

Thoughts?

xPaw:
@astorije I fixed your comments.

client/js/socket.js
```diff
@@ -11,6 +11,10 @@ const socket = io({
 	reconnection: !$(document.body).hasClass("public"),
 });
 
+$("#connection-error").on("click", function() {
+	$(this).removeClass("shown");
+});
```
astorije: I'm not sure to understand what this is for. My understanding is that the banner is now clickable and clicking it removes it. What's the reason behind it?

xPaw: Upload errors re-use that red banner.

defaults/config.js
```diff
@@ -139,6 +139,27 @@ module.exports = {
...
+	// - `enable`: When `true`, the identd daemon runs on server start.
+	// - `maxFileSize`: When file upload is enabled, users sending files above this
+	//    limit will be prompted an error message in their browser.
+	//    Specified value is in kilobytes. Default value is 10000 kilobytes (10 MiB).
+	//
+	// `-1` will disable the file size limit and will allow files of any size. Use at your own risk.
+	fileUpload: {
+		enable: true,
+		maxFileSize: 10000,
```
astorije: @xPaw, why did you change this from 10240 to 10000? Other fields (well, at least `prefetchMaxImageSize`) are a power of 2, and it's also more predictable for users as most OS show sizes in MB. I'd rather remember to check that a file is < 10MB and not 9.77MB :)

xPaw: I changed the multiplier to 1024 to match the other max size field we have in the config already.

src/plugins/uploader.js
```diff
@@ -0,0 +1,152 @@
...
+	"audio/midi",
+	"audio/mpeg",
+	"audio/ogg",
+	"audio/x-wav",
+	"image/bmp",
+	"image/gif",
+	"image/jpeg",
+	"image/png",
+	"image/webp",
+	"text/plain",
```
astorije: Just curious, things like log files (often in `.log`) and Markdown files (`.md`), they would be  treated as plain text, right? (I haven't tested yet, doing code review first then I'll play with the feature)

src/plugins/uploader.js
```diff
@@ -0,0 +1,152 @@
...
+			return res.sendFile(filePath);
+		});
+	}
+
+	static getMaxFileSize() {
+		const configOption = Helper.config.fileUpload.maxFileSize;
+
+		if (configOption === -1) { // no file size limit
+			return null;
+		}
```
astorije: Do we have a protection against negative values (or zero) somewhere? Not sure how the uploader library reacts on 0 or -42, maybe we should blow up the process ourselves?

src/plugins/uploader.js
xPaw: The extension doesn't matter, it detects the type from the file contents, so yes logs or markdown will work fine.

src/plugins/uploader.js
xPaw: Looked at the source code, it will simply act as a limit (say file exceeded limit)

`if (self.maxFileSize !== null && file.size > self.maxFileSize) {`

defaults/config.js
```diff
@@ -139,6 +139,27 @@ module.exports = {
...
+	//
+	// Allow uploading files to the server hosting The Lounge.
+	// If enabled, The Lounge will store files in the `${THELOUNGE_HOME}/uploads` folder.
+	//
+	// Files will not expire or be removed by The Lounge, which can cause issues
+	// depending on your hardware, for example in terms of disk usage.
+	//
+	// The available keys for the `fileUpload` object are:
+	//
+	// - `enable`: When `true`, the identd daemon runs on server start.
```
astorije: @xPaw, I made the following fixes/changes here:

<details>

```diff
@ config.js:145 @ module.exports = {
    // ### `fileUpload`
    //
    // Allow uploading files to the server hosting The Lounge.
-   // If enabled, The Lounge will store files in the `${THELOUNGE_HOME}/uploads` folder.
    //
-   // Files will not expire or be removed by The Lounge, which can cause issues
-   // depending on your hardware, for example in terms of disk usage.
+   // Files are stored in the `${THELOUNGE_HOME}/uploads` folder, do not expire,
+   // and are not removed by The Lounge. This may cause issues depending on your
+   // hardware, for example in terms of disk usage.
    //
    // The available keys for the `fileUpload` object are:
    //
-   // - `enable`: When `true`, the identd daemon runs on server start.
-   // - `maxFileSize`: When file upload is enabled, users sending files above this
-   //    limit will be prompted an error message in their browser.
-   //    Specified value is in kilobytes. Default value is 10000 kilobytes (10 MiB).
-   //
-   // `-1` will disable the file size limit and will allow files of any size. Use at your own risk.
+   // - `enable`: When set to `true`, files can be uploaded on the client with a
+   //   drag-and-drop or using the upload dialog.
+   // - `maxFileSize`: When file upload is enabled, users sending files above
+   //   this limit will be prompted an error message in their browser. A value of
+   //   `-1` disables the file size limit and allows files of any size. **Use at
+   //   your own risk.** This value is set to `10000` kilobytes by default.
    fileUpload: {
        enable: true,
        maxFileSize: 10000,
```
</details>

This is how it looks in the docs page:

screen shot 2018-09-03 at 00 15 46

Let me know what you think :)

defaults/config.js
astorije: Right, but you made the change so that it multiplies 10000kb into 10000*1024 bytes, this value should still be 10240 so it's 10mb (10*1024*1024 = 10,485,760 bytes)... no?
At the very least, `10240` here would be consistent with the `2048` above.

src/plugins/uploader.js
astorije: Awesome, that's what I thought, thanks for confirming!

defaults/config.js
xPaw: You're right, fixed.

MaxLeiter:
🎉

mostlyvirtual:
Thank you for implementing this. What was the reasoning of having this disabled by default? Also, would it be possible to include the enable/disable toggle in the UI settings?

PR #2554, opened 2018-06-13
WHOWAS support

### LINKED ISSUE #2050: Not showing /whowas response
The server response for 

**/whowas _nick_** 

seems not to be handled bei lounge. There is simply no output anywhere.

Jay2k1:
fixes #2050 

This is basically just a copy of the WHOIS event, adapted to fit WHOWAS.

Things to note:
- it uses the `whois` css class, not sure if that's okay or if it should get its own
- the template says `WHOWAS info for <nick>`, the whois template just says `<nick>`
- I only added support for the event (i.e. server response), not for the command itself, not sure if that is necessary

Suggestions welcome.

image

### DIFF
--- a/client/views/actions/whois.tpl
+++ b/client/views/actions/whois.tpl
@@ -1,4 +1,8 @@
-<p>{{> ../user_name nick=whois.nick}}</p>
+<p>
+	{{> ../user_name nick=whois.nick}}
+	{{#if whois.whowas}} is offline, last information:{{/if}}
+</p>
+
 <dl class="whois">
 {{#if whois.account}}
 	<dt>Logged in as:</dt>
--- a/src/plugins/irc-events/whois.js
+++ b/src/plugins/irc-events/whois.js
@@ -5,23 +5,37 @@ const Msg = require("../../models/msg");
 
 module.exports = function(irc, network) {
 	const client = this;
-	irc.on("whois", function(data) {
+
+	irc.on("whois", handleWhois);
+
+	irc.on("whowas", (data) => {
+		data.whowas = true;
+
+		handleWhois(data);
+	});
+
+	function handleWhois(data) {
 		let chan = network.getChannel(data.nick);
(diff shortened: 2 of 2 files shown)


client/views/actions/whowas.tpl
```diff
@@ -0,0 +1,10 @@
+<p>WHOWAS info for {{> ../user_name nick=whowas.nick}}</p>
```
MaxLeiter: It may be better to say `Available info for disconnected {{nick}}` or something. Not a fan of showing `WHOWAS` here.

Zarthus:
Somewhat relevant: https://github.com/kiwiirc/irc-framework/pull/157

client/views/actions/whowas.tpl
```diff
@@ -0,0 +1,10 @@
...
+<dl class="whois">
+
+	<dt>Host mask:</dt>
+	<dd class="hostmask">{{whowas.ident}}@{{whowas.host}}</dd>
+
+	<dt>Real name:</dt>
+	<dd>{{{parse whowas.real_name}}}</dd>
```
Zarthus: Dynamically looping over all whowas.* fields may bring forward compatibility with changes to kiwiirc/irc-framework's event emission, as long as we are OK with showing everything that a normal WHOWAS would also show and don't want to exclude something for convenience.

client/views/actions/whowas.tpl
xPaw: How about we straight up use `whois` template/message type? We can change this leading line by adding extra flag in whois response like `whowas: true`.

Jay2k1: It's pretty big already, but since we should get more or less the same fields that also exist in whois, why not.

xPaw: That's the idea, since the fields are shared, there's little reason to use a separate template.

PR #2564, opened 2018-06-16
Move .nyc_output folder out of root

xPaw:
`nyc` already uses `node_modules/.cache/nyc` folder, so I don't know why they don't also put this temp output folder there too.


xPaw:
@astorije I left a comment here https://github.com/istanbuljs/nyc/issues/605#issuecomment-397799952

PR #2565, opened 2018-06-16
Target more modern browsers in babel

xPaw:
Reason being is that `last 2 versions` is a really poor descriptor and will forever target things like IE 10 and 11, even though Lounge doesn't work there even after transpiling (due to functions string.endsWith being used).

Use modern targets means we can actually transpile less code and end up with smaller final bundle size. This means arrow functions and rest parameters are left as-is.

The target I changed to seems rather reasonable, but might require some tweaking: http://browserl.ist/?q=last+1+year%2C+firefox+esr

References:
https://jamie.build/last-2-versions
http://browserl.ist/?q=last+2+versions

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -46,7 +46,10 @@ const config = {
 						presets: [
 							["env", {
 								targets: {
-									browsers: "last 2 versions",
+									browsers: [
+										"last 1 year",
+										"firefox esr",
+									],
 								},
 							}],
 						],


astorije:
This is very interesting, thanks for sharing those links. I totally support making this change as we head into v3, it's the right time to ask those questions 👏

What do you think about the recommendation given in your first reference: http://browserl.ist/?q=%3E0.25%25%2C+not+ie+11%2C+not+op_mini+all
Compared to your PR:
- On desktop, it adds Safari 10, Opera 51, older Chromes. On mobile, it adds Samsung Internet 4, iOS Safari 9 and 10, the Android browser, UC browser
- It removes Opera < 51, Firefox 55, 56, 57

This seems a bit better to me, but you have done more research than me so I am curious what your thoughts were.

xPaw:
@astorije With that query, there's minimal change because it includes browsers like Android 4.4 and UC Browser, and because of these targets it keeps transpiling all of the things.

astorije:
Alright then!

PR #2567, opened 2018-06-17
Do not remove date marker when loading history if date changes

### LINKED ISSUE #2566: Loading more history doesn't create a date marker after last message
Running b4693fb7. I think the issue is entirely client-side related to the `more` event. I could reproduce it with sqlite history.

Basically goes like this:
1. Join a channel
2. Leave a channel
3. Join the channel again after the date changes

I believe the issue is that it doesn't try to create a date marker after the last message in `more` buffer because it doesn't know that there's more messages in DOM already.

xPaw:
Fixes #2566, see that issue for more details

### DIFF
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -21,17 +21,24 @@ socket.on("more", function(data) {
 		return;
 	}
 
-	// Remove the date-change marker we put at the top, because it may
-	// not actually be a date change now
+	// Remove the date marker at the top if date does not change
 	const children = $(chan).children();
 
-	if (children.eq(0).hasClass("date-marker-container")) { // Check top most child
-		children.eq(0).remove();
-	} else if (children.eq(1).hasClass("date-marker-container")) {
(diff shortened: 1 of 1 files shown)


client/js/socket-events/more.js
```diff
@@ -21,17 +21,24 @@ socket.on("more", function(data) {
...
-	// not actually be a date change now
+	// Remove the date marker at the top if date does not change
 	const children = $(chan).children();
 
-	if (children.eq(0).hasClass("date-marker-container")) { // Check top most child
-		children.eq(0).remove();
-	} else if (children.eq(1).hasClass("date-marker-container")) {
-		// The unread-marker could be at index 0, which will cause the date-marker to become "stuck"
-		children.eq(1).remove();
-	} else if (children.eq(0).hasClass("condensed") && children.eq(0).children(".date-marker-container").eq(0).hasClass("date-marker-container")) {
```
xPaw: This code doesn't appear to be necessary because condensed messages are split if date changes.

PR #2570, opened 2018-06-19
Warn if config owner doesn't match process uid

xPaw:


astorije:
Is that really a problem? As long as the process can read the file, it should be fine, no? I.e. if process uid is in group assigned to the file, or if the file has +r on "others", shouldn't everything work as expected?

(As a follow-up, it might be nice to warn if the process is able to write in the config file, if that's something that can be checked)

xPaw:
This is to fix the issue I raised in debian package.

> Is that really a problem? As long as the process can read the file, it should be fine, no?

Lounge also writes and creates files, so if you run it under root you can cause issues later on.

> (As a follow-up, it might be nice to warn if the process is able to write in the config file, if that's something that can be checked)

We can check mode in the same stat call.

astorije:
> Lounge also writes and creates files, so if you run it under root you can cause issues later on.

I get that, but I'm more thinking about use cases where config file has either 1) user `nobody` and group `thelounge` (assuming TL runs under user `thelounge`), 2) does not share user/group with the process, but file has `o+r` permission.
Those are pretty common situations and are perfectly safe/legal, no?

EDIT: Ooooh this is for the home path, for some reason I read config file! I guess my use cases above still apply, but it's a bit trickier.

Is there anything else you want to add/change to this PR or do you think it should go as is?

astorije:
Ok then. Good to review now?

xPaw:
Sure.

PR #2571, opened 2018-06-19
Implement irc:// protocol handler

xPaw:


client/views/windows/settings.tpl
```diff
@@ -14,6 +14,10 @@
 	</div>
 
 	<div class="row">
+		<div class="col-sm-12">
+			<button type="button" class="btn" id="make-default-client">Open irc:// URLs in The Lounge</button>
```
astorije: `Opent <code>irc://</code> and <code>ircs://</code> in...`? (Also, `with` instead of `in`? Don't trust me on that one)

xPaw: Would be pretty long to include `ircs://` in the button?

astorije: Oooh I'm dumb, I didn't even pay attention that it wasn't a checkbox!
Yeah it absolutely would. In fact, remove `<code>` as well actually, it's pretty gross on buttons. We can re-think that when we improve buttons (v3.1? 😁).

client/views/windows/settings.tpl
```diff
@@ -14,6 +14,10 @@
 	</div>
 
 	<div class="row">
+		<div class="col-sm-12">
+			<button type="button" class="btn" id="make-default-client">Open <code>irc://</code> URLs with The Lounge</button>
```
astorije: Since GitHub hides the comment: https://github.com/thelounge/thelounge/pull/2571#discussion_r197602539

xPaw:
@MaxLeiter want to play around with this too?

PR #2578, opened 2018-06-20
Expand topic on hover (or tap on mobile)

### LINKED ISSUE #276: Topic is cut off

starquake:
When the topic is too big this expands the header on hover.

Took the fix by @Jay2k1 and made some minor adjustments:
- Used classes that already existed.
- Kept the height at 45px like the original header.
- Header now has a background so it's not transparent anymore
- whitespace: pre-wrap should stay as it is
- the smaller line-height now has padding to make sure the text does not move up on expand

Fixes #276

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -928,12 +928,20 @@ background on hover (unless active) */
 }
 
 #windows .header {
+	position: absolute;
+	width: 100%;
+	background: var(--window-bg-color);
 	line-height: 45px;
 	height: 45px;
 	padding: 0 6px;
 	display: flex;
 	flex-shrink: 0;
 	overflow: hidden;
+}
+
+#windows .chat-content,
+#windows .container {
+	margin-top: 45px;
 }
 
 #windows #chat .header {
@@ -953,6 +961,27 @@ background on hover (unless active) */
 	flex-grow: 1;
 	overflow: hidden;
 	font-size: 14px;
+}
+
+#windows .header:hover {
+	height: auto;
+	z-index: 1;
+}
+
+#windows .header:hover .topic {
+	-webkit-mask-image: none;
+	mask-image: none;
+	line-height: 25px;
+	padding: 10px 0;
+	word-break: normal;
+	text-overflow: ellipsis;
+}
+
+@media (max-width: 768px) {
+	#windows .header:hover .topic {
+		padding: 12px 0;
+		line-height: 21px;
+	}
 }
 
 #windows #chat-container.active {


xPaw:
I can imagine `:hover` being annoying on desktop. Should we handle `click` in JS?

starquake:
Hover feels logical to me. It's not uncommon for cut off text to show a tooltip when it doesn't fit for example. I thought click was reserved for editing the topic in the future.

astorije:
I'm with @xPaw on this: hover is a delicate state as it won't show up on mobile, and accesibility-wise it's usually a nightmare (not just screen readers/blind users but all sorts of impairments) as it presents no affordances.
We already have a `title` on the topic that was rather a temporary solution, so doing it correctly should involve clicking somewhere on the topic IMO.

Then the question is how to correctly display that this is clickable/extendable (on top of the gradient). There are a few things I can think of, none are ideal:
- Use a caret (on the right-end side of the topic line) similarly to other carets usage we have, but that's a bit cryptic.
- Display a "Click to reveal/expand" message at the end of the topic. To make it less intrusive, it could only show up... when hovering. Kinda back to square 1, but not as much.
- A dedicated button next to the context menu button, but that's not very good either.

I'm favoring the second option for now, which can be added later. In the meantime, simply toggling expand/collapse when clicking on the topic would be enough for this PR.

Re: editing, we can see when we get there, for example it could be an "Edit" button in expanded mode or something.

starquake:
I agree that a caret that allows you to expand the topic is the best solution.

> hover is a delicate state as it won't show up on mobile
Funny thing is: apparently a element with :hover works on focus after click on mobile. I tested it and it automagically works. But I see your point.

I'll see if I can familiarize myself with the JS part of thelounge and see if I can make a clickable topic.

astorije:
> Funny thing is: apparently a element with :hover works on focus after click on mobile. I tested it and it automagically works. But I see your point.

Wow, I did not know that. Maybe that's how Primer tooltips work and I never paid attention? No idea. Thanks for bringing that up!

> I'll also make some mockups of ideas I have so we can pick one.

That's a great way to approach this, thanks!

RockyTV:
Why not click to expand, double-click to edit? When double-clicking, it could copy the original topic to the input bar or just have `/topic` in the bar when you click to edit.

astorije:
That doesn't sound like a very intuitive UX, and I definitely think we can do better. In particular, having the topic edits in the message input is not great, it should stay close to the place you are editing if possible.

Jay2k1:
I like hover-to-expand, which, as has already been said, automagically works on mobile with touch-to-expand and touch-somewhere-else to collapse. The only UX problem I found with this (using it for months now) is that due to the title tag, when you hover, not only does the topic expand but you also get the tooltip which gets in the way.

As for the idea of single-click-to-expand, double-click-to-edit: Why not just add a pencil icon to the topic that you can click to edit? And editing should be inside the topic container, not copy to the input bar, I agree.
The pencil could be added contitionally, e.g. only when you are allowed to set the topic (i.e. if chanmode +t is set, only when you're op). Or, if it's always there, maybe disable it conditionally or something.

williamboman:
There’s an issue [when toggling the user list sidebar](
https://streamable.com/uww3h) on mobile devices.

xPaw:
Do you want to give this another go, perhaps with some javascript now that the chat window has been ported to Vue?

starquake:
I'll give it another go this or next weekend. Not really familiar with Vue so will have to look into that.

xPaw:
@starquake Did you get a chance to play with this yet?

starquake:
Yeah sorry about that. Not really able to pick it up soon. Should we close it?

xPaw:
Closing this, if you're willing to fix this up, feel free the reopen!

PR #2593, opened 2018-06-26
Switch close button text on channels to Leave and kept as Close for other types.

MaxLeiter:
Matches other UI elements, like context menus

### DIFF
--- a/client/views/chan.tpl
+++ b/client/views/chan.tpl
@@ -25,14 +25,19 @@
 		<span class="add-channel-tooltip tooltipped tooltipped-w tooltipped-no-touch" aria-label="Join a channel…" data-alt-label="Cancel">
 			<button class="add-channel" aria-label="Join a channel…" aria-controls="join-channel-{{id}}"></button>
 		</span>
-	{{/equal}}
-	{{#notEqual type "lobby"}}
+	{{else}}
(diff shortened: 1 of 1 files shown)


client/views/chan.tpl
```diff
@@ -29,9 +29,16 @@
...
 		<span class="badge{{#if highlight}} highlight{{/if}}" data-highlight="{{highlight}}">{{#if unread}}{{roundBadgeNumber unread}}{{/if}}</span>
-		<span class="close-tooltip tooltipped tooltipped-w" aria-label="Leave">
-			<button class="close" aria-label="Leave"></button>
-		</span>
+		{{#equal type "channel"}}
+			<span class="close-tooltip tooltipped tooltipped-w" aria-label="Leave">
+				<button class="close" aria-label="Leave"></button>
+			</span>
+		{{/equal}}
+		{{#notEqual type "channel"}}
```
astorije: Mind replacing these 2 lines with `{{else}}` like we do in https://github.com/thelounge/thelounge/blob/master/client/views/image_viewer.tpl#L18 for example?
Feel free to do the same in L29/42 if you want (but don't have to).

MaxLeiter: done

PR #2595, opened 2018-06-26
Clarify /topic command without newtopic

joandrsn:
Just found it that I could get the current topic with `/topic` and since it wasn't mentioned in the help, I would like to add it.

### DIFF
--- a/client/views/windows/help.tpl
+++ b/client/views/windows/help.tpl
@@ -545,7 +545,7 @@
 		<div class="description">
 			<p>
 				Get the topic in the current channel. 
-				If <code>newtopic</code> is specified the command sets the 
+				If <code>newtopic</code> is specified, sets the 
 				topic in the current channel.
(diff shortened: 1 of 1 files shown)


client/views/windows/help.tpl
```diff
@@ -540,10 +540,14 @@
...
 		<div class="subject">
-			<code>/topic newtopic</code>
+			<code>/topic [newtopic]</code>
 		</div>
 		<div class="description">
-			<p>Set the topic in the current channel.</p>
+			<p>
+				Get the topic in the current channel. 
+				If <code>newtopic</code> is specified the command sets the 
+				topic in the current channel.
```
astorije: Could you remove "the command" and add a comma so it ends up as follows:

```html
If <code>newtopic</code> is specified, set the topic in the current channel.
```

Thanks!

PR #2599, opened 2018-06-27
Make userlist state consistent on mobile and desktop

xPaw:
I got myself a new phone, which has 18:9 screen and in landscape mode it ends up being higher width (839px) than Lounge's mobile breakpoint, and it toggles userlist into opposite state.

This PR does the following things:
1. Rename `.rt` state into `.userlist-open`
2. If `.userlist-open` is set, user list is open regardless of viewport width
3. If The Lounge is opened on a device with width of less than 1024px, and there is no stored state preference, the user list is closed by default.

### DIFF
--- a/client/js/lounge.js
+++ b/client/js/lounge.js
@@ -28,10 +28,6 @@ $(function() {
 	const viewport = $("#viewport");
 
 	function storeSidebarVisibility(name, state) {
-		if ($(window).outerWidth() < utils.mobileViewportPixels) {
-			return;
-		}
-
 		storage.set(name, state);
 
 		utils.togglePreviewMoreButtonsIfNeeded();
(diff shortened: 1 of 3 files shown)


client/js/lounge.js
```diff
@@ -40,20 +36,26 @@ $(function() {
...
 
 		slideoutMenu.toggle(isOpen);
-		storeSidebarVisibility("thelounge.state.sidebar", isOpen);
+
+		if ($(window).outerWidth() >= utils.mobileViewportPixels) {
+			storeSidebarVisibility("thelounge.state.sidebar", isOpen);
+		}
 	});
 
 	viewport.on("click", ".rt", function() {
```
SinZ163: Doesn't this PR rename .rt to .userlist-closed, what is being clicked?

xPaw: Nope the button stays `.rt` (same with `.lt`).

PR #2604, opened 2018-06-28
Introduce issue templates

Zarthus:
Live demo: https://github.com/Zarthus/thelounge/issues/new/choose

~~(branches off of the previous PR, ID 2603) no longer relevant cause its merged~~

If desired, please take the time to discuss in depth what should be there, what else you want to add, whether it really brings any value at the moment, etc. #yakshaving

### DIFF
--- a/.github/ISSUE_TEMPLATE/Feature_Request.md
+++ b/.github/ISSUE_TEMPLATE/Feature_Request.md
@@ -1,0 +1,12 @@
+---
+name: Feature Request
+about: Request a new feature
+
+---
+
+<!-- Have a question? Join #thelounge on freenode -->
+
+* *The Lounge version:* 
+
+---
+
--- a/.github/ISSUE_TEMPLATE/Feature_Request.md
+++ b/.github/ISSUE_TEMPLATE/Feature_Request.md
@@ -4,9 +4,7 @@ about: Request a new feature
 
 ---
 
-<!-- Have a question? Join #thelounge on freenode -->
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/Feature_Request.md
```diff
@@ -0,0 +1,12 @@
+---
```
Zarthus: I don't think this template is really necessary, but without it it would look like this, which might not be as user friendly:

img

MiniDigger:
(the title should be issue templates rather than commit templates)

.github/ISSUE_TEMPLATE/Feature_Request.md
```diff
@@ -0,0 +1,12 @@
...
+name: Feature Request
+about: Request a new feature
+
+---
+
+<!-- Have a question? Join #thelounge on freenode -->
+
+* *The Lounge version:*
```
astorije: I'm not 100% sure the version really matters here, but maybe things such as: "Expected behavior", "Description", etc., everything needed to make a feature request comprehensive.

Also I think it would be useful to have a block comment telling people to check open and closed issues to see if the feature has been requested before, etc.

Zarthus: I added a generic "Feature Description" block, I looked at the existing TL issues and I think most feature requests have been going well off without prompting for too much data.

PR #2617, opened 2018-07-05
Disable username autocapitalize on input field when signing on

christer88:
Fixes https://github.com/thelounge/thelounge/issues/2616


christer88:
This is one annoying feature.... when you try to log in on a phone or something else where it autocapitalize. You start to type the username and soon finds out it has been autocapitalize, even though usernames 99 % of the times have small letters. Let's stop this. 👍

PR #2622, opened 2018-07-06
Fix incorrect left padding for highlights on mobile

astorije:
Fixing something @xPaw mentioned to me a couple days ago. Hope I got it right.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -2463,6 +2463,10 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		padding: 0;
 	}
 
+	#chat .channel .message.highlight {
+		padding-left: 5px;
+	}
+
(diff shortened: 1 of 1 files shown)


xPaw:
Actually this isn't correct fix either:

client/css/style.css
```diff
@@ -2463,7 +2463,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 		padding: 0;
 	}
 
-	#chat .channel .message.highlight .time {
+	#chat .channel .message.highlight {
 		padding-left: 0;
```
xPaw: This should be `5px` (see screenshot I posted)

astorije:
@xPaw, should be good to go now, thanks for noticing this!

PR #2626, opened 2018-07-06
Fix incorrect target in (dis)connect context menu

xPaw:


client/js/contextMenuFactory.js
```diff
@@ -127,57 +127,35 @@ function addCloseItem() {
...
-	let clickedNetwork;
-
-	function isDisconnected(target) {
-		return target.parent().hasClass("not-connected");
-	}
-
-	function connect() {
+	function connect(itemData) {
 		socket.emit("input", {
-			target: $("#chat").data("id"),
```
xPaw: This just disconnected whatever active network you were in.

Issue #109, opened 2016-02-25
Message search


xPaw:
Would be a killer feature to have a log search (could be channel only, network only, or global).

astorije:
This needs *a lot* of discussion and design... :-)
But definitely a great long-term goal!

Dr-Horv:
Sounds like a great feature. Most IRC clients I have used have had implementation on channel-basis and have some grep-like interface. Might not be suitable for a more GUI-focused interface like thelounge though.

dgw:
Would it be bad to override Ctrl/Cmd+F? My habitual desktop client, Textual, searches in the current channel that way, though it is not as-you-type like the browser's built-in search would be.

bnvk:
This would be an ideal feature component to have, and if implemented correctly would solve performance / stability issues like #110

IMHO, this should be done with an "[database abstraction layer](https://en.wikipedia.org/wiki/Database_abstraction_layer)" as peoples views, familiarity, needs, and use cases are always a bit different and having a good DBAL would be a really attractive feature of The Lounge. Of course, this adds a bit of complexity to engineering... so prolly just pick "one" to implement and do minimal abstracting, but at least think about it.

Some nice datastore options to consider:

- [LevelDB](http://leveldb.org)
- [SQLite](https://www.sqlite.org)
- [ElasticSearch](https://www.elastic.co)

Some nice features for this would be allowing a user to *choose* what types of messages get saved in the DB... some ideas:

- All messages and notifications
- Only messages
- Only messages by and mentioning a given "user" (ideally with some leading / trailing buffer)
- Only certain channels and/or networks
- Only when a user is "active" within some timeframe (20 mins, etc...)
- User action choosing to "save" one or more selected messages

xPaw:
@bnvk We talked about it briefly on IRC, and indeed a lightweight option is required (by default, something like sqlite). Filtering would be left for plugins though, not base implementation.

richrd:
Hi and many thanks to all the work you guys put in thelounge, especially 2.0. Loving it! :heart: 

I would love a find option, and currently I just bash the "Show older messages " button and hit Ctrl+F. That's fine for me but it doesn't work on mobile when the app is added to the desktop (no chrome menu to access).

I'm against a mandatory database (as I mentioned here https://github.com/thelounge/lounge/pull/506#issuecomment-250759696). 
[...]

PR #2627, opened 2018-07-06
MVP Message Search (WIP)

richrd:
This is not ready, but I'm opening this PR for discussion (issue #109). This is my first contribution so there's probably a lot to do/change/fix etc. This is how it looks like on mobile: https://www.youtube.com/watch?v=jid1v9-jN7k (video is a bit out of date at this point)

What this does:
- Adds a search field to the channel and network header bars (on mobile a search icon is shown, clicking it opens the search field as a dropdown)
- Pressing enter in the field runs the search and navigates the user to the search results page
- Searching from the lobby searches all channels, searching from a channel searches only in that channel
- The search result page allows the search to be refined and shows how many results were found
- The search keyword is highlighted in each search result message
- The backend db search is just a simple `LIKE %keyword%` and is limited to 100 results

Let me know what you think.


richrd:
Some thoughts:
- We could ditch the search icon on desktop and just show the search field in the header
- We need to do something with the UI if search (logs) are not available
- The dropdown animation could be ditched,  its just unneccessary eyecandy
- It would be really nice to highlight the search term in the message results
- Should the search also apply to nicks?

christer88:
Suggestion: I think it would be nice if it is possible to click on the text line to get you to where it is in the channel, or some other way to show you more than just the single line that had the word you searched for. Sometimes you kind of need to see what happen before and after the text line to understand the context.

richrd:
I agree, thats more or less a must have feature. Its not an easy thing though so it would be good to add it later. I dont know whether this will make it into 3.0 or 3.1 but hopefully sooner than later :)

richrd:
Any comments on this?

richrd:
I'd actually like to close this and reimplement it after vue is released. It would be a lot cleaner to do it after that. Thoughts?

McInkay:
@richrd That would probably make most sense. Thanks for the initial work, and sorry it didn't get in this time.

PR #2630, opened 2018-07-08
Fix disabling autocomplete and fix completing special channels

xPaw:


client/js/autocompletion.js
```diff
@@ -318,7 +318,7 @@ function completeChans(word) {
 		.each(function() {
 			const self = $(this);
 
-			if (!self.hasClass("lobby")) {
+			if (self.hasClass("channel")) {
```
astorije: This disables from queries as well, right? Is that intentional?

astorije: So this should be `if (self.hasClass("channel") || self.hasClass("query")) {` I think.

xPaw: But what reason do you have to complete queries after typing `#` somewhere?

astorije: Oh wait, I think I'm reading this incorrectly.

astorije: Yes for some reason I read this as the location it's autocompleted from (which window) and not what is being populated, my bad.

PR #2631, opened 2018-07-08
Use CSS for toggling seconds in message timestamps.

richrd:
Just leaving this here. Probably not very relevant after Vue takes over :)

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1184,9 +1184,14 @@ background on hover (unless active) */
 	padding-left: 10px;
 	width: 55px;
 }
-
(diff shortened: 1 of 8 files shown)


xPaw:
👎 on this. This also won't really do much when we add support for 12H clock (there was a PR for it).

McInkay:
Agreed, I think it'll be nicer once we've got vue in and the state can be read by other things, and that'll remove some of the logic we have.

I think we don't get much advantage of doing this either, as we still use JS to set whether it is showing or not.

PR #2634, opened 2018-07-09
Use webpack-dev-server in yarn watch/dev

xPaw:
This is from Vue branch, enables hot reloading.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:browser": "webpack-dev-server --config=webpack.config-browser.js",
     "test:client": "nyc --nycrc-path=test/.nycrc-mocha-webpack mocha-webpack --colors --opts=test/mocha-webpack.opts",
     "test:server": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
-    "watch": "webpack --watch"
+    "watch": "webpack-dev-server --hot --inline"
   },
   "keywords": [
     "lounge",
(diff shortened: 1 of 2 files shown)


astorije:
To make sure I got this right: the only thing it provides (at least for now) is that page auto-reloads on change as opposed to previous `yarn watch` where you had to hit refresh by hand, correct?

package.json
```diff
@@ -22,7 +22,7 @@
     "test:browser": "webpack-dev-server --config=webpack.config-browser.js",
     "test:client": "nyc --nycrc-path=test/.nycrc-mocha-webpack mocha-webpack --colors --opts=test/mocha-webpack.opts",
     "test:server": "nyc --nycrc-path=test/.nycrc-mocha mocha --colors",
-    "watch": "webpack --watch"
+    "watch": "webpack-dev-server --hot --inline"
```
astorije: Do you actually need to specify `--inline`?

```
  --inline      Inline mode (set to false to disable including client scripts
                like livereload)                       [boolean] [default: true]
```

astorije: Does `--hot` bring anything at the moment? (I couldn't tell the difference with/without)

astorije:
@xPaw do you need to do any further changes on this?

williamboman:
Worth noting is that `webpack-dev-server` has sort of been discontinued in favor for https://github.com/webpack-contrib/webpack-serve.

xPaw:
Damn it, webpack

PR #2637, opened 2018-07-10
Do not write /list messages to logs

xPaw:
Noticed this while working on Vue, this is a temporary fix (hence why I'm not adding new types for it) as this isn't a problem in Vue branch (it doesn't use Msg at all).

### DIFF
--- a/src/models/msg.js
+++ b/src/models/msg.js
@@ -46,6 +46,9 @@ class Msg {
 			this.type !== Msg.Type.ERROR &&
 			this.type !== Msg.Type.BANLIST &&
 			this.type !== Msg.Type.IGNORELIST &&
+			this.type !== "channel_list" &&
+			this.type !== "channel_list_loading" &&
+			this.type !== "channel_list_truncated" &&
 			this.type !== Msg.Type.TOPIC_SET_BY &&
 			this.type !== Msg.Type.WHOIS;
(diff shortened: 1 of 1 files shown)


src/models/msg.js
```diff
@@ -46,6 +46,9 @@ class Msg {
 			this.type !== Msg.Type.ERROR &&
 			this.type !== Msg.Type.BANLIST &&
 			this.type !== Msg.Type.IGNORELIST &&
+			this.type !== "channel_list" &&
```
McInkay: Why are these not in the enum? They probably should be.

src/models/msg.js
```diff
@@ -46,6 +46,9 @@ class Msg {
 			this.type !== Msg.Type.ERROR &&
```
McInkay: It would probably be nice to create an array of types that we don't log and switch the code to

`return this.TYPES_NOT_TO_LOG.includes(this.type)` 

Might be nicer overall?

src/models/msg.js
xPaw: That's what I meant, these types are all gone in Vue which is why I didn't bother adding them to the enum.

McInkay: Oh I missed that in your description. Ignore me. Alrighty.

PR #2643, opened 2018-07-13
Make userlist section titles sticky on scroll, improve contrasts for accessibility, use CSS variables

### LINKED ISSUE #909: User list puts more space between header and section body than between sections
A visual annoyance that's been bothering me for a while now. Occurs on all four default themes shipped with 2.2.0 (Example, Crypto, Morning, and Zenburn).

The user list looks like this, roughly:

````
Owners
____

user1
user2

Administrators
____

user3

Operators
____

user4
user5
user6
<etc.>
````

This does not visually group the section headers with the content of each section correctly, IMO. The subtle line separating the heading from the rest of the section puts too much space in, and the headings visually group with the section before them instead of the section they describe.

I'd say putting the line between sections (instead of between the headers and their section bodies) would fix the apparent grouping. @astorije, I know you like to weigh in on UI stuff. Do you see what I'm talking about, or should I prepare a screenshot or two to illustrate? :)

### LINKED ISSUE #2635: html-minifier-loader deletes a closing tag and nests divs that should be siblings
* *Node version:* 8.11.3
* *The Lounge version:* master, v3.0.0-pre.7

---

For some reason I can't figure out, `html-minifier-loader` gets rid of [this closing tag](https://github.com/thelounge/thelounge/blob/master/client/views/user.tpl#L5).

With `html-minifier-loader` | Without `html-minifier-loader`
--- | ---
screen shot 2018-07-10 at 01 46 14 | screen shot 2018-07-10 at 01 45 51

Probably some misconfiguration of `html-minifier-loader` but I couldn't figure it out. @xPaw, I know you're all in with Vue right now, but mind taking a quick look at this? 🙏

### PR DESCRIPTION
**Pull request by astorije**
Stickiness extracted from https://github.com/thelounge/thelounge/pull/2527. Also slightly updated styling, and improved accessibility of muted texts on default theme (via the new CSS var) and was able to get rid of some overrides in Morning due to that CSS var.

Browsers that do not support `position: sticky` essentially do nothing with it (i.e. treat it as an element positioned by default) so that's good enough graceful degradation. It's essentially only Safari at this point so I'm not planning on a polyfill (Autoprefixer would take care of this anyway).

Before | After
--- | ---
userlist_before | userlist_after

~(This is currently affected by #2635. It still works but doesn't look as good, so test without `html-minifier-loader` if you want to get the expected look)~
This PR now fixes #2635.

This happens to fix #909.

### DIFF
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -107,10 +107,6 @@
 	color: #b7c5d1;
 }
 
-#windows .header .topic {
-	color: #b7c5d1;
-}
-
 #viewport .lt,
 #viewport .rt,
 #chat button.menu,
@@ -123,35 +119,12 @@
 	color: #f3f3f3;
 }
 
-#chat .self .text {
-	color: #b7c5d1;
-}
-
 #chat .error,
 #chat .error .from {
 	color: #f92772;
 }
 
 #chat .unhandled .from {
-	color: #b7c5d1;
-}
-
-#chat .msg.topic {
-	color: #b7c5d1;
-}
-
-#chat .time,
-#chat .condensed .content,
-#chat .away .content,
-#chat .back .content,
-#chat .join .content,
-#chat .kick .content,
-#chat .mode .content,
-#chat .nick .content,
-#chat .part .content,
-#chat .quit .content,
-#chat .topic .content,
-#chat .topic_set_by .content {
 	color: #b7c5d1;
 }
 
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -2,6 +2,7 @@
 
 :root {
 	--body-color: #f3f3f3;
+	--body-color-muted: #b7c5d1;
 	--link-color: #77abd9;
 	--window-bg-color: #303e4a;
 	--date-marker-color: #97ea70;
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -124,14 +124,6 @@
 	color: #f92772;
 }
 
-#chat .unhandled .from {
-	color: #b7c5d1;
-}
-
-#chat table.channel-list td {
-	color: #b7c5d1;
-}
-
 #chat .msg.motd .text,
 code,
 .irc-monospace {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -5,6 +5,9 @@
 	/* Main text color */
 	--body-color: #222;
 
+	/* Secondary text color, dimmed. Make sure to keep contrast WCAG 2.0 AA compliant on var(--window-bg-color) */
+	--body-color-muted: #767676;
+
 	/* Background color of the whole page */
 	--body-bg-color: #415364;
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by xPaw** [comment]
Looks fine. Should we perhaps put a border under the search bar?

**Comment by richrd** [comment]
Would it be enough if the input had a bottom border? Also would a border color CSS var make any sense? It would be nice to have IMO, but not in this PR.

**Comment by astorije** [comment]
> Looks fine. Should we perhaps put a border under the search bar?

@xPaw, it doesn't look like it because 💩 GIF, but the search has a different background (gray). It appears white on the GIF.

Otherwise I'm not touching the search field in this PR, it will be for a later PR, there are other things that need to be done on the search field anyway.

**Comment by xPaw** [comment]
Does this PR fix #909 then?

**Comment by xPaw** [comment]
So the sticky menu hides active user when scrolling with keyboard arrows (scroll down a little, then arrow up).

**Comment by astorije** [comment]
> Does this PR fix #909 then?

Yes! Thanks for noticing.

**Comment by astorije** [comment]
> So the sticky menu hides active user when scrolling with keyboard arrows (scroll down a little, then arrow up).

Good catch. It looks like there is not solution to have `position: sticky` and `scrollIntoView()` to play well together.

However, this is resolved by changing to `block: "center"` [here](https://github.com/thelounge/thelounge/blob/astorije/sticky-userlist/client/js/utils.js#L63). What do you think about this?
It has minimal impact but, more importantly, as I was playing with this, I actually think it is nicer to use. When you press up/down, you know what's coming before/after rather than blindly going for the next user/channel. I do think it's a UX improvement actually.
Would you be okay with this change?

**Comment by xPaw** [comment]
Does that option work in Firefox esr? We had problems with it.

**Comment by astorije** [comment]
Just tested and it works in Firefox 60. I believe the issues were with the "smooth" behavior, not this. Note that I don't have ESR but I believe ESR is 60, no?

On Safari, it works exactly the same with `center` or `end`, which means... it doesn't really work. Both act like `start`.

So yeah, I think we should use `center`. 👍 / 👎 ?

File: client/themes/morning.css [hunk 1]
```diff
@@ -135,25 +128,6 @@
 	color: #b7c5d1;
```
xPaw: This and `#chat table.channel-list td` also use this color, are we not able to use `var` here?

astorije: I've done a couple things in 8c09be9d, let me know what you think:

- For unhandled messages, I moved the override to main styling and used muted color (I also applied it to the icons for MOTD and CTCP). I removed color for `#chat .from` which should always be set for specific things: colored or uncolored nicknames, icons, etc.
- Removed color overrides for the channel list. There was no real reason to override the main color so now channel list table looks the same as ignored/ban user tables. I also removed the override on channel names in this table because I didn't think it brings anything to do that, and it's now more consistent with inline channels in chat content, and would keep it consistent if at some point we decided to style inline channels.

I didn't touch `#windows .window h2` because that's something I'd want to do in the main style without resorting to this specific variable (so have a new variable for headings for example) and that's way outside of what I was trying to achieve here, I'll deal with this later if that works for you.

dgw:
> This happens to fix #909.

Works for me!

PR #2647, opened 2018-07-15
Port sidebar, chat, user list, state to Vue

### LINKED ISSUE #443: && (and most likely others like that) should not be parsed as channel
At the moment, typing `&&` (such as messaging shell commands) will parse `&&` as a channel name, and therefore will be clickable and tab-able. It should not.

If someone takes this, extra credits given for a test suite that we can augment with these cases, and run as part of `npm test` :-)

### LINKED ISSUE #706: Have separate input box for channels and PM's
For example, if I type `Yea sure thing` in #channela, but before I finish, I get a pm and decide to reply to that first, when I switch to another channel (or in this case pm), the input field is reset. When I go back to #channela what I typed is put back into the input box.

### LINKED ISSUE #813: Hide day markers if there's no messages (accounting for hidden stuff, too)
image

It believe it isn't hiding them because there are hidden join/quit messages which are turned off in my settings.

### LINKED ISSUE #932: Replace jquery-ui with a different (lightweight) drag and drop solution
We include jquery ui to implement channel sorting. There are couple of problems with it:
1. It's heavy
2. It does not work on mobile (no touch support)

### LINKED ISSUE #1270: Lazily load autolink images/previews
Intersection Observer API can be used for this. https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

### LINKED ISSUE #1311: Notify user of new messages via document.title
IRCCloud does this well - they prepend a `(+)` to the `document.title` when there are unread messages in the open buffer when the tab isn't focused.

https://github.com/erming/shout/pull/348/ tried to accomplish this in Shout.

### LINKED ISSUE #2223: Show that there are unreads/highlights in collapsed network
Currently, if you have a network collapsed, you can't see if there are unread messages and notifications in channels under that network. This can especially be confusing if you have a number of collapsed networks, and there's an unread highlight somewhere, meaning you have to uncollapse each network until you find where it is, if you want to clear the notification.

### LINKED ISSUE #2245: Show channel name in list when joined successfully
I hope this issue has not been discussed before, but would be great to display channel name in list only   if you join successfully.

### LINKED ISSUE #2249: Channel list empty on refresh
Should probably re-fetch `/list`

### LINKED ISSUE #2266: Being kicked isn't very intuitive

### LINKED ISSUE #2746: Unread markers are not updated when closing a window from the context menu
* *Node version:* v8.11.3
* *Browser version:* Chrome 68.0.3440.106
* *Device, operating system:* macOS 10.13.6
* *The Lounge version:* 3.0.0-pre.7 (I know, haven't updated to pre.8 yet...)

---

When closing a query/channel window with unread messages from the context menu (without opening the window), the notification icons (favicon + left sidebar icon dot) do not clear, and the unread counter in window title is not updated.
These are correctly reset when clicking any other window.

### LINKED ISSUE #2791: per-window history
The message history is currently shared between all the windows (channels and queries). It would be nice to have it separated, so when ArrowUp is pressed the last message/command sent in that window is repeated.

### LINKED ISSUE #2800: Website links that error out always show preview cannot be loaded box
Docker: 3.0.0-rc.1

The below boxes show on links that error out, regardless of the "auto-expand websites". Surely if I've unchecked the box, I do not care of they cannot be expanded. A few sites tend to error, due to to user-agent string too, which makes this box show up often.

image

### LINKED ISSUE #2801: Copy text includes <> when username is not selected
Docker: 3.0.0-rc.1
Firefox: v59

Doesn't copy the full username when you dont select the full line

image

Video: https://s.rgp.io/20180924l10u7.mp4

xPaw:
~~- [ ] Dragging a channel onto chat fades it as it expects a file drop~~
~~- [ ] Pressing up/down arrows in multi line input moves the chat up (when there's no history yet)~~
- [x] `Auto expand media` is off, but preview still opens when sending a link
- [x] Add `preserveWhitespace: false`
- [x] `*` is most likely prepended to any users in messages when copying messages
- [x] Next/prev buttons in media viewers are broken
- [x] Copying action lines lost `*`
- [x] Scroll button needs to be moved somewhere where parent is not `display: flex`, as its introducing bugs the button is animating
- [x] Check auto expand media options
- [x] Fix `keep-scroll-position="function () { [native code] }"` on condensed messages
- [x] Unread counter in sidebar may remain non-zero for active channel very rarely
- [x] `lodash` is not being tree shaked it seems
- [x] Scroll to bottom button (and most likely bottom sticky check itself) are buggy on iOS
- [x] Fix `istanbul-instrumenter-loader` failing
- [x] Hidden status messages will most likely break unread marker (because unread message id will not be in `condensedMessages`)
- [x] Set input history limit as its per-channel and we don't want to indefinitely grow memory
- [x] `updateDateMarkers` in `lounge.js` needs to be fixed for Vue
- [x] Selecting with arrows in autocomplete should skip history
- [x] `parse` test needs to be fixed to work with Vue
- [x] Input history is not implemented
- [x] Implement file uploading, which was added in master
- [x] Toggling user list and sidebar doesn't keep scroll at bottom
- [x] Move `togglePreviewMoreButtonsIfNeeded` to Vue
- [x] "More" button is not re-checked on window resize
- [x] Channels can't be sorted (we tried vuedraggable already, but its really poor on mobile)
- [x] Client reconnects need to be handled correctly (right now it just overwrites in `init`)
- [x] Unread marker doesn't display when it should be displayed after condensed (??)
- [x] Server disagrees with client on what the first unread message is
- [x] Resize scroll to bottom is not instant on keyboard open
- [x] Input resize is wacky
- [x] Pending message is reset on reconnection
- [x] Pressing tab again after completing a nick inserts `undefined`
- [x] Switching to another channel doesn't click `show more history`
- [x] Toggling previews and collapsed messages keeps scroll to bottom
- [x] Autocompleting emoji and pressing enter in virtual keyboard (gboard, android) doesn't actually update the model
- [x] Pressing send button with autocomplete menu doesn't close it
- [x] Handle `#connection-error` with Vue state instead of setting text with jQuery
- [x] Context menu on channels with unread counter display the unread text in it like `#channel (1 unread)`
- [x] Custom highlights do not work on page reload (only work when receiving a message) - Will probably leave this option open because custom highlights need to be moved to the server already

Changes in Vue:
- Only active channel is rendered in DOM instead of them all. Fixes #1270
- Input is no longer global (each channel gets its own input field). Fixes #706
- Input history is also per-channel now. Fixes #2791
- Collapsing a network stays in active channel (and the active channel is still visible in the list)
- Collapsed networks still show channels with highlights. Fixes #2223
- Client commands are refactored and `/join` can assume current channel. Fixes #2004 Fixes #2281 Fixes #2266 
- Channel list persists on page reloads. Fixes #2249 
- Message parser now returns Vue elements (via createElement) instead of raw html content (no more manual escaping!)
- Hidden status messages are no longer rendered in DOM. Fixes #813
- Parser now actually accepts CHANTYPES/PREFIX from network. Fixes #443
- Channel state is synchronised to the client (and is displayed). Fixes #2245
- Chat input is no longer disabled when disconnected (still can't submit though)
- All channels user lists are cleared when disconnected from the network
- New utility function to update window title, notified state, favicon consistently. Which fixes titles or favicon sometimes not updating, e.g. on channel part. Fixes #2746. Fixes #1311
- `<>` and `*` around nicks have been moved to DOM so copying is consistent. Fixes #2801
- Added `***` before actions when copying (e.g. topic set, joins, etc), and dashes around nick for notices.
- Removed `jquery-ui` in favor of another sorting library. Fixes #932
- Image too big error is now handled by `auto expand media` setting, while other errors are handled by `auto expand websites` option. Fixes #2800
- Custom highlights are now synced to the server and highlight code works on the server, this means that push notifications work for custom highlights, and they get synced across clients correctly.
- PageDown/PageUp now scroll natively without manually calculating and animating the offset.

Overall this should make our code base much cleaner, implements proper state (which is much needed for plugin support) and performance improvements (many component actions are async, and no DOM actions wasted on updating hidden channels).

[Vue devtools](https://chrome.google.com/webstore/detail/vuejs-devtools/nhdogjmejiglipccpnnnanhbledajbpd?hl=en) is recommended.

Some things may still be rough in the code because not the entire app has been ported to Vue yet. The idea is to get this PR stable and merged, and then work on moving remaining windows and code parts to Vue, which should remove remaining quirks in the code.

I am already running this branch on my main client (and some other people do too) to catch any remaining issues in this port.

Thanks to @timmw and @richrd for helping with the code!


xPaw:
This PR has an insane amount of commits because they were not cleaned up. I think I could rewrite the commit history to be somewhat cleaner later on after code review?

richrd:
Are any of the above unchecked features up for grabs?

client/js/upload.js
```diff
@@ -26,7 +23,7 @@ function initialize() {
 	instance.addEventListener("error", (event) => {
 		// Reset progressbar
 		$("#upload-progressbar").width(0);
-		$("#connection-error").addClass("shown").text(event.message);
+		vueApp.connectionError = event.message;
```
MaxLeiter: Depending how temporary using the UI like this is, we may want to rename `connectionError`

client/css/style.css
```diff
@@ -1103,8 +1103,8 @@ background on hover (unless active) */
 }
 
 .scroll-down {
-	position: absolute;
-	bottom: 16px;
+	position: fixed;
```
richrd: Nice pun here, fixed fixed it. :smirk:

client/components/App.vue
```diff
@@ -92,7 +92,7 @@
 </template>
 
 <script>
-import {throttle} from "lodash";
+const throttle = require("lodash/throttle");
```
astorije: Doesn't `import { throttle } from 'lodash/throttle'` work?

xPaw: Right, that exact wording would work.

client/components/ChatInput.vue
```diff
@@ -0,0 +1,219 @@
...
+</template>
+
+<script>
+const commands = require("../js/commands/index");
+const socket = require("../js/socket");
+const upload = require("../js/upload");
+const Mousetrap = require("mousetrap");
+const {wrapCursor} = require("undate");
+
+const colorsHotkeys = {
```
MaxLeiter: These aren't really all colors hotkeys (in fact, only one is)

client/components/App.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+	<div
+		id="viewport"
+		role="tablist">
+		<aside id="sidebar">
+			<div class="scrollable-area">
+				<div class="logo-container">
+					<img
+						:src="'img/logo-' + ( isPublic() ? 'horizontal-' : '' ) + 'transparent-bg.svg'"
```
astorije: Nit: I believe our JS files are set up to have no spaces inside parentheses, not sure. I think that would make sense.

Also thoughts on:

```suggestion
						:src="`img/logo-${isPublic() ? 'horizontal-' : ''}transparent-bg.svg`"
```

(None of this is holding off the PR of course, just thinking out loud)

client/components/App.vue
```diff
@@ -0,0 +1,142 @@
...
+			<div
+				id="changelog"
+				class="window"
+				aria-label="Changelog" />
+		</article>
+	</div>
+</template>
+
+<script>
+const throttle = require("lodash/throttle");
```
astorije:
```suggestion
import { throttle } from "lodash/throttle";
```

client/components/App.vue
```diff
@@ -0,0 +1,142 @@
...
+
+import NetworkList from "./NetworkList.vue";
+import Chat from "./Chat.vue";
+
+export default {
```
astorije: Sidenote, but at work we found ourselves almost exclusively using named imports. Reasons are multiple, but:

- TS historically used to be a pain on default imports (doesn't apply here)
- That kinda enforces naming being used across app/app (`import { App } from '../App' and, yes, I know you can rename, but it has to be explicit to override as opposed to typo). This makes searching across codebase very nice.
- You can't `export default` an ES6 class
- It makes for nice index files where we import/export everything in our libraries: `export { ThisThing } from './src/ThisThing'` (as opposed to `export default as ThisThing from '...'` or whatever the syntax is).

Not asking for any change here, just meant to share experience :)

client/components/App.vue
xPaw: Using import on lodash breaks tree shaking.

xPaw:
The (parted) is shown like that because I was debugging the channel states, making sure we tracked them correctly. I will leave making it pretty for a separate PR.

client/components/App.vue
xPaw: Hm, I'll look around maybe there's a way to fix eslint to apply inside of vue strings like that.

client/components/App.vue
xPaw: Looks like what you are suggesting is currently experimental: https://github.com/vuejs/vue-class-component

client/components/App.vue
astorije: Even when specifying the sub-package like this? That's weird, I would expect the opposite, that tree shaking works better on import than require.

astorije:
> The (parted) is shown like that because I was debugging the channel states, making sure we tracked them correctly. I will leave making it pretty for a separate PR.

Makes sense re: styling in a future PR. Can we remove it from this PR for now then? It looks really sub-par as is.

client/components/App.vue
xPaw: Hm. Possibly, but I do specifically remember having issues with lodash, so I don't want to try again right now.

xPaw:
@astorije Changed it to a class, so we can simply apply styles later on.

astorije:
> @astorije Changed it to a class, so we can simply apply styles later on.

Perfect!

astorije:
I just noticed the loading page does not have the frame like the log in page has. Is that on purpose, or lost in translation?

xPaw:
@astorije the loading screen is outside of the viewport element now (because it's taken over by Vue).

williamboman:
image

Just noticed that scroll to bottom button is missing styling in the Morning theme

richrd:
Same in most themes I would expect. I reckon at least the official themes will be updated before Vue is released. I will be updating zenburn.

astorije:
Just noticed this: when a server window is active, the window title is incorrect:

screen shot 2019-02-15 at 00 35 24

In a channel, it's correct, but it's a little funny:

screen shot 2019-02-15 at 00 35 56

The first one should be `Freenode — The Lounge`.
The second one with the 2 `—`s is weird. I get why the network name is in there, but I was fine not having it, I'm on multiple networks and that has never bothered me. If we do want to keep it, something like any of the following seem better to me: `#thelounge (Freenode) — The Lounge`, `#thelounge on Freenode — The Lounge`, `#thelounge | Freenode — The Lounge`, etc.

astorije:
> @astorije the loading screen is outside of the viewport element now (because it's taken over by Vue).

Got it. I fixed it and took the liberty to add the fix to your branch. It's not the best CSS code ever, but hopefully when re-doing window management with Vue it will get a lot better.

xPaw:
@astorije Yeah lets do the styling, and the title you mentioned in separate PRs.

package.json
```diff
@@ -73,23 +73,25 @@
...
     "fuzzy": "0.1.3",
     "graphql-request": "1.8.2",
     "handlebars": "4.1.0",
     "handlebars-loader": "1.7.1",
     "html-minifier": "3.5.21",
     "html-minifier-loader": "1.4.1",
     "intersection-observer": "0.5.1",
-    "istanbul-instrumenter-loader": "3.0.1",
     "jquery": "3.3.1",
-    "jquery-ui": "1.12.1",
```
astorije: 🎉

astorije: The new behavior is less good than the old one though, but I'm hoping we can fix that later.
Namely the drag also follows the x-axis, which is annoying, but not a big deal; also the content of the initial position gets duplicated. Not a big deal for channels but makes it ugly on whole networks. Not blocking merging this of course.

PR #2649, opened 2018-07-15
Use translate3d consistently

xPaw:
Also restores transition animation on user list on mobile


astorije:
Didn't we remove some translate3d some months ago? I forget why but I think it had an impact on sharpness on iOS or made userlist non functional. Could be wrong obviously. @williamboman does that ring a bell to you?

astorije:
It might have been this thread too https://github.com/thelounge/thelounge/pull/1120

xPaw:
Looks like you are right. I'll close this.

PR #2661, opened 2018-07-18
Add opacity on user names hovering in nick list

christer88:
Add opacity on user names hovering in nick list. User names have same opacity on hovering in channel, and by adding this to user names in nick list we have more consistency, in my opinion.


christer88:
Making new pull requests due to commit issues.

PR #2664, opened 2018-07-18
Recommend using Yarn instead of NPM in changelog and readme

christer88:
TheLounge has started using Yarn, but does also support NPM. Yarn seems to in some cases solve issues people experience when using NPM, e.g. Sqlite3 installation.

I think it's good and about time to recommend using Yarn instead of NPM.

### DIFF
--- a/scripts/changelog.js
+++ b/scripts/changelog.js
@@ -106,7 +106,7 @@ Please refer to the commit list given above for a complete list of changes, or w
 As with all pre-releases, this version requires explicit use of the \`next\` tag to be installed:
 
 \`\`\`sh
-npm install -g thelounge@next
+[sudo] yarn global add thelounge@next
 \`\`\`
 `;
 }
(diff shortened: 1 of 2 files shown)


jmcbee:
Are people experiencing problems installing thelounge via npm? If not, we're not really solving a problem. Besides, npm has already improved.

MiniDigger:
Yarn is already recommended in the readme, this just fixed the changelogs for consistency.

scripts/changelog.js
```diff
@@ -106,7 +106,7 @@ Please refer to the commit list given above for a complete list of changes, or w
 As with all pre-releases, this version requires explicit use of the \`next\` tag to be installed:
 
 \`\`\`sh
-npm install -g thelounge@next
+[sudo] yarn global add thelounge@next
```
McInkay: Remove sudo, you don't need root for yarn global, and we shouldn't recommend that.

MiniDigger: should be changed in readme here too then https://github.com/thelounge/thelounge/blob/master/README.md#running-stable-releases-using-yarn-recommended

PR #2665, opened 2018-07-18
Add opacity on user names hovering in userlist

christer88:
Add opacity on user names hovering in userlist. User names have same opacity on hovering in channel, and by adding this to user names in userlist we have more consistency, in my opinion.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1564,7 +1564,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	white-space: nowrap;
 }
 
-+#chat .names .user:hover {
+#chat .names .user:hover {
 	opacity: 0.6;
 }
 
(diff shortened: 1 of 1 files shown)


richrd:
Don't we already have a background-color hover effect? I'm not sure if two effects are necessary.

astorije:
I'm with @richrd. In fact, the lower opacity + gray background (on default theme) now makes highlighted nicks rather hard to read.

I actually think more and more that we should get rid of the lower opacity on link hover. It's a more difficult change because there are more implications (e.g. lower opacity on icon buttons feels fine to me, channel sidebar is its own set of visual conventions, etc.) but I do think that lower opacity is not great on text when hovering.

I would close this until we deal with links/buttons altogether in a more sensible way.

PR #2669, opened 2018-07-21
refactor getHumanDate() to accept a timestamp; avoids possible test failure; more accurate log times

MaxLeiter:


test/src/helperTest.js
```diff
@@ -56,6 +56,7 @@ describe("Helper", function() {
 	});
 
 	describe("#getHumanDate()", function() {
-		expect(log.getHumanDate()).to.equal(moment().format("YYYY-MM-DD HH:mm:ss"));
+		const ts = 1455090480; // https://github.com/thelounge/thelounge/pull/2
+		expect(log.getHumanDate(ts)).to.equal(moment(ts).format("YYYY-MM-DD HH:mm:ss"));
```
xPaw: Just do `equal("2016-02-10 09:48:00")`.

test/src/helperTest.js
```diff
@@ -56,6 +56,6 @@ describe("Helper", function() {
 	});
 
 	describe("#getHumanDate()", function() {
-		expect(log.getHumanDate()).to.equal(moment().format("YYYY-MM-DD HH:mm:ss"));
+		expect(log.getHumanDate(moment(1455090480000).utc())).to.equal("2016-02-10 07:48:00")
```
astorije: You're missing a semi-colon here, hence the failures.

PR #2682, opened 2018-07-29
Remove nick highlighting feature to avoid highlighting non-nick words

### LINKED ISSUE #2008: False nick detection when it's a part of a word with `

ivan:
thelounge highlights all nicks in a message and turns them into user menu buttons.  But people sometimes use nicks that are frequently-used words like "the" or "that".  thelounge then highlights any usage of "the" or "that" in that channel, even when the word does not refer to a user with that nick.  In fact, many more nicknames are regular words, and thelounge can't possibly know if a word was used in nick or non-nick context.

Because highlighting random words is confusing and distracting, remove the nick highlighting feature entirely. 

Fixes #2008


xPaw:
This feature is not going to be removed. If you don't want nick highlights, you can easily reset the color via CSS so it appears just as normal text.

ivan:
Thanks for taking a look, anyway.  Not that you need to justify this, but given that it does weird things in some channels, why not?  I prefer not having to customize software to get rid of weirdness.  I don't think getting rid of this makes it much harder to look up a user: you can still get to them pretty quickly by clicking a nick to the left of a message, or by searching for them in the nick column.

xPaw:
Removing the feature outright because it has false positives on common words like `the` is just not a good solution. A better solution to this could be figured out (e.g. keep the nick detection, but only start applying colors if the nick is 4 characters or longer).

dgw:
> but only start applying colors if the nick is 4 characters or longer

Or if it's not in a dictionary of common short words. (My nick is 3 characters, so I have some interest in this topic.)

The obvious problem with that approach is, common in what language? The Lounge cannot assume all channels are English, after all.

But whether any of these suggestions are implemented, removing the feature would certainly irritate far more users than a few false positives.

williamboman:
This could be a fun NLP problem for you linguistics people out there :)

ivan:
HexChat might strike the right balance here: it does scan for nicks and #channels in messages, but because the words are not guaranteed to be nicks or channels, it keeps the appearance of the parsed items identical to regular text.  On mouseover, it shows an underline and changes the cursor to a pointer.

Anyway, I won't press the issue further.

realies:
Would having an on/off toggle for the feature be a good solution?

McInkay:
> Would having an on/off toggle for the feature be a good solution?

No

Issue #2699, opened 2018-08-10
Failed to read the 'localStorage' property from 'Window'


adamus1red:
* *Node version:* v10.8.0
* *Browser version:* 67.0.3396.99
* *Device, operating system:* Windows 10
* *The Lounge version:* v3.0.0-pre.7

---

When accessing thelounge from an iFrame embedded in a page on a different domain the UI breaks because it cannot read localStorage.

*Console log*
```js
localStorage.js:14 Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
    at Object.get (https://webchat.domain1/beta/js/bundle.js:1:16005)
    at Object.<anonymous> (https://webchat.domain1/beta/js/bundle.js:1:8023)
    at s (https://webchat.domain1/beta/js/bundle.js:1:518)
    at Object.<anonymous> (https://webchat.domain1/beta/js/bundle.js:1:11618)
    at s (https://webchat.domain1/beta/js/bundle.js:1:518)
    at Object.<anonymous> (https://webchat.domain1/beta/js/bundle.js:1:256645)
    at s (https://webchat.domain1/beta/js/bundle.js:1:518)
    at a (https://webchat.domain1/beta/js/bundle.js:1:387)
    at https://webchat.domain1/beta/js/bundle.js:1:1458
    at https://webchat.domain1/beta/js/bundle.js:1:1462
```

PR #2700, opened 2018-08-11
Add try catch to all localStorage methods.

adamus1red:
Return a warning to console when we need to catch.
Return null when attempting to get an item from localStorage

REF: #2699

### DIFF
--- a/client/js/localStorage.js
+++ b/client/js/localStorage.js
@@ -14,7 +14,12 @@ module.exports = {
 		try {
 			return window.localStorage.getItem(key);
 		} catch (e) {
-			console.warn("Cannot get item from localStorage");
+			// Do nothing. You can end up here because localStorage isn't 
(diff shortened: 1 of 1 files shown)


client/js/localStorage.js
```diff
@@ -11,12 +11,25 @@ module.exports = {
...
 	clear() {
-		window.localStorage.clear();
+		try {
+			window.localStorage.clear();
+		} catch (e) {
+			console.warn("Cannot clear localStorage");
+		}
 	},
-};
+};
```
xPaw: Keep the new line

xPaw:
I don't think `console.warn` are really necessary.

But do add a comment saying why we wrap it in try/catch (can move the existing comment out and add security restrictions may also apply when using iframes).

xPaw:
The comments could be combined into one big comment on top of the file instead of duplicating the info.

PR #2702, opened 2018-08-11
Fix bugs with URL overrides in Connect window

astorije:
Follow-up of https://github.com/thelounge/thelounge/commit/94f1d8dde053570bade85096b67cc8a49ee8b809 and probably another commit.

💥 _3 bug fixes for the price of 1!_ 💥

### 1. Disallow URL override of network-related fields on the client when the network is locked

Before | After
--- | ---
screen shot 2018-08-11 at 18 16 38 | screen shot 2018-08-11 at 18 16 47

### 2. Pass network name along to the client connect window when network is locked and not displayed

Before | After
--- | ---
screen shot 2018-08-11 at 18 08 00 | screen shot 2018-08-11 at 18 09 40

### 3. Make sure the network name cannot be changed through URL override when the network info is not displayed

Before | After
--- | ---
screen shot 2018-08-11 at 18 10 37 | screen shot 2018-08-11 at 18 11 20

### DIFF
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -176,6 +176,12 @@ function parseOverrideParams(params, data) {
 			continue;
 		}
 
+		// When the network is locked, URL overrides should not affect disabled fields
+		if (data.lockNetwork &&
+				["host", "port", "tls", "rejectUnauthorized"].includes(key)) {
+			continue;
+		}
+
 		if (key === "join") {
 			value = value.split(",").map((chan) => {
 				if (!chan.match(/^[#&!+]/)) {
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -182,6 +182,11 @@ function parseOverrideParams(params, data) {
 			continue;
 		}
 
+		// When the network is not displayed, its name in the UI is not customizable
+		if (!data.displayNetwork && key === "name") {
+			continue;
+		}
+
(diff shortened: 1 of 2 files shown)


client/js/socket-events/configuration.js
```diff
@@ -176,6 +176,17 @@ function parseOverrideParams(params, data) {
 			continue;
 		}
 
+		// When the network is locked, URL overrides should not affect disabled fields
+		if (data.lockNetwork &&
+				["host", "port", "tls", "rejectUnauthorized"].includes(key)) {
+			continue;
+		}
+
+		// When the network is not displayed, its name in the UI is not customizable
```
xPaw: I believe this is not enforced on the server.

astorije: Mmmh, indeed. Anything (simple if possible) I need to add to this PR? Or for later?
It's plain weird that we can change the name on the title from the URL when not displayed (and a regression) but I see what you're saying.

xPaw: Hm. I think this `if` you added isn't even needed because you shouldn't have added `name` on the server in `getClientConfiguration`.

Server defaults to defaults.name if none provided here and `displayNetwork ` is false: https://github.com/thelounge/thelounge/blob/master/src/client.js#L204

Technically it's still possible to directly send a name by modifying the form in devtools, but overriding the URL shouldn't work?

PR #2705, opened 2018-08-18
Hide `Native App` settings section when no items are available

MaxLeiter:
on Safari, there would be a "Native App" header with no options underneath it.

Alternatively, instead of hiding the options, I could set them to `disabled` and add a tooltip. Thoughts?


xPaw:
This certainly might not work as expected, as `webapp-install-button` can become visible in an event later on.

PR #2717, opened 2018-08-25
Renamed username forms name to match standards

### LINKED ISSUE #1740: Firefox won't offer to save password
Currently when loggin in with username and password there is no popup asking me if I want to remember the login data in my browser password directory. It would be great if lounge could setup the elements accordlingy so this works :).

MaxLeiter:
ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

I believe this fixes #1740

### DIFF
--- a/client/js/socket-events/auth.js
+++ b/client/js/socket-events/auth.js
@@ -35,7 +35,7 @@ socket.on("auth", function(data) {
 				values[obj.name] = obj.value;
 			});
 
-			storage.set("user", values.user);
+			storage.set("user", values.username);
 
 			socket.emit("auth", values);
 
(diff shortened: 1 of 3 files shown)


richrd:
I dont see any standard for the `name` attribute. Looks like you meant to use the `autocomplete` attribute instead?

xPaw:
@dgw Can you check if this actually fixes that issue?

dgw:
@xPaw I'll test if I get a chance, but we should also ping @gsantner as the original reporter.

xPaw:
@MaxLeiter I was worried something would be missed in the rename, and it is, e.g. in `auth()` function call.

Could you revert the rename and just keep the `autocomplete` option? It should fix the password saving without potentially breaking anything.

xPaw:
Playing around with this, this doesn't fix saving passwords in Firefox.

xPaw:
Superceded by #2741

PR #2718, opened 2018-08-25
Remember disconnected networks between server restarts

xPaw:


src/plugins/inputs/connect.js
```diff
@@ -5,21 +5,26 @@ const Msg = require("../../models/msg");
...
-		if (irc.connection.connected) {
+		if (irc.connection && irc.connection.connected) {
 			chan.pushMessage(this, new Msg({
 				type: Msg.Type.ERROR,
 				text: "You are already connected.",
 			}));
 			return;
 		}
 
-		irc.connection.connect();
```
xPaw: For some reason this incorrectly called `connect` on the connection, and not the client instance itself.

PR #2738, opened 2018-09-01
Add google analytics support

adamus1red:
- Modify CSP policy to allow google analytics in script and img policies when requried
- Inject the google analytics tracking and google tag manager as required for tracking.
  - Auto fill tracking ID based on Google Analytics ID value
- Add to default configuration.


xPaw:
This does not belong in the core, sorry.

adamus1red:
Well is there somewhere else to put it because I don't see anywhere else but core to hook stuff like this in.

xPaw:
This would be suited as a plugin, which we sadly do not have yet.

But because we don't support plugins doesn't mean we should be putting such features in the core.

PR #2744, opened 2018-09-03
Remove incorrect closing tag

xPaw:


client/index.html.tpl
```diff
@@ -89,7 +89,7 @@
 						<span id="nick"></span>
 						<textarea id="input" class="mousetrap"></textarea>
 						<span id="upload-tooltip" class="tooltipped tooltipped-w tooltipped-no-touch" aria-label="Upload File">
-							<input id="upload-input" type="file" multiple></button>
+							<input id="upload-input" type="file" multiple>
```
McInkay: This should be self-closing then, if not closed by a closing tag.

xPaw: Inputs don't have to.

SinZ163: In good ol' XHTML it would have needed to so it can be valid XML, but it doesn't need to be strict XHTML to be valid html5

McInkay: Yeah I saw that on w3c after posting it. I mostly learnt with xhtml and so it's a hangover from that :-P

PR #2749, opened 2018-09-04
Move correct file after upload, put file extension in slug

xPaw:


src/plugins/uploader.js
```diff
@@ -42,39 +42,37 @@ class Uploader {
...
+			log.error(`File upload failed: ${data.error}`);
 		});
 
 		// maxFileSize is in bytes, but config option is passed in as KB
 		uploader.maxFileSize = Uploader.getMaxFileSize();
 		uploader.listen(socket);
-
-		// Returns Promise
-		function handleSaving(data) {
-			const tempPath = path.join(Helper.getFileUploadPath(), ".tmp", data.file.name);
```
xPaw: `name` is the original name, while `pathName` is the path where uploader plugin actually stored the file (e.g. when deduplicating).

richrd:
Does this change the filename into a random hash? I'm not strictly against that but I'd really like the original filename to be used with the original extension. The filename adds context to the upload and if/when the file is downloaded/saved it will get a more meaningful name.

xPaw:
Effectively this or does the same thing as master, but also adds an extension.

astorije:
I think I'm with @richrd, but I agree this is unrelated to this PR.
@richrd, mind opening an issue so we can discuss about it? There might be details I'm not considering right now.

src/plugins/uploader.js
```diff
@@ -42,39 +42,37 @@ class Uploader {
...
+
+			fsextra.move(data.file.pathName, destPath).then(() => {
+				const slug = path.basename(data.file.pathName);
+				const url = `uploads/${randomName}/${slug}`;
 				client.emit("upload:success", url);
+			}).catch(() => {
+				log.warn(`Unable to move uploaded file "${data.file.pathName}"`);
+
+				// Emit failed upload to the client if file move fails
+				socket.emit("siofu_error", {
```
astorije: What do we do with this error? I'm not seeing any handling of `siofu_error` on the client, is it specific to the file upload library or something?
If it's for own consumption, maybe `file_upload_error` instead?

xPaw: Client library emits error event which our code hsndles.

astorije: Yeah, but this is our code emitting the error to our client, no?

xPaw: Yeah, but I didn't want to write our own error handler here because this error should basically never happen.

PR #2009, opened 2018-01-28
Favicons, favicons everywhere

astorije:
Quite a few things that will feel odd when going through this PR, but keep in mind that I have spent **hours** on this, reading half of the web, testing on multiple devices, going back and forth between decisions, etc.
So none of this is final of course, there are a few TODOs, but remember that (almost) everything is there for a reason before thinking something is stupid :D

## Details

- Browser tab icons have been bundled to ICO files, so they can contain all of the dimensions we have: 16px, 32px, and 64px. According to [this](https://github.com/audreyr/favicon-cheat-sheet#ico-file), we might want to add 48px, though the use case seems to be achieved by the 64px one. ❓
- I have added a watermark for the size for now, this is helpful for testing, to make sure the correct icon is being picked up. For example, this proved me that even though iPhone 8 uses 120px icons, having only 180px and 64px doesn't work, there _needs_ to be a specific 120px version. Same for all other iOS devices :man_facepalming:
- From [this list](https://stackoverflow.com/a/19933647), I decided to leave out `iPhone(first generation or 2G), iPhone 3G, iPhone 3GS` and `iPad and iPad mini @1x` because too old and probably not compatible with The Lounge. I also did not include the Android devices, handled in `manifest.json`.
- Speaking of Android, I did not include the 128px version... Should we❓ (I mean, we have 120px 😕)
- All SVGs have been optimized with `svgo --multipass`, gaining between 32 and 45% depending on the file.
- `logo-dark.svg` was unused.
- ~~I have not done the notification icons part yet.~~
- `manifest.json` contains a list of every PNG we have, and a SVG as the biggest size + 1px, so that all potential OS that needs > 192px (I couldn't test any) would use the SVG version as a fallback.
- Then there is the case of the stupid Safari mask icon. First of, to make it work (for future reference), I had to replace `style="enable-background:new 0 0 64 64;"` with `enable-background="new 0 0 64 64"`. Apart from that, it's [this icon](https://github.com/thelounge/thelounge.github.io/blob/master/assets/logos/icons/TL_Square_Favicon_Transparent_Bg/64x64/TL_Square_Favicon_Transparent_Bg_64x64px.svg). According to Apple docs, we should change all colors to black, but it seems to work that way. Thoughts❓
- I hardcoded the official color in Safari mask icon, because this is the color of the icon, which should always be official-ish, not depending on theme. I hesitated with the orange, but went with the gray/blue for better contrast.
- Safari started to support the Web App Manifest, but it's too early to just rely on it, I think.

[...]

PR #2755, opened 2018-09-05
Replace normal icon with the transparent version

### LINKED ISSUE #2470: [UX] Alert type favicon no longer eye catchy
* *The Lounge version: v3.0.0-pre.6* 

---
My better half complained that she often doesn't notice that I queried her ever since I upgraded her instance to pre6 because "the icon no longer changes to red". 

Ironically, it took me about 50 minutes to realize she told me this for the very same reason 😂 

Before the introduction of the new favicons in #2009, the alert icon had a red background and was clearly distinguishable from the regular icon, but now, the difference is rather slight and as it seems not eye catchy enough. Check out these screenshots of my tab:

image vs 
image

Of course, if you see them side-by-side like this, they are clearly different. But it's not flashy enough to catch your attention when you're not directly looking at it in my opinion (and hers).

A first suggestion would be to exchange the grey and orange like so:
image
Or in Chrome:
image

This way, it would still stick to the [Brand guidelines](https://github.com/thelounge/thelounge/issues/282#issuecomment-360368920) colors but still be way more eye catchy and alerting.

Bigger mockup:
image
vs
image
edit: or to increase contrast of the "eyes":
image

Thoughts?

McInkay:
Fixes #2470


astorije:
I'm an idiot. This is different from #2009 because "normal" background is now transparent, and favicon was generated by... me, I did it. Why am I asking then.
😩💤

PR #2756, opened 2018-09-06
Add 3 commands that were missing on client autocomplete

McInkay:


client/js/constants.js
```diff
@@ -32,11 +32,13 @@ const commands = [
 	"/cs",
 	"/ctcp",
 	"/cycle",
+	"/dehop",
 	"/deop",
 	"/devoice",
 	"/disconnect",
 	"/expand",
 	"/ho",
+	"/hop",
```
astorije: Is `/ho` intentional? Never paid attention to it, and it's not in the help page.

client/js/constants.js
```diff
@@ -56,6 +58,7 @@ const commands = [
 	"/os",
 	"/part",
 	"/query",
+	"/quote",
```
astorije: As far as I remember, `/quote` being a simple alias of `/raw`, we didn't include it here on purpose. Shouldn't we remove `/quote` and `/send` here, and keep `/raw` only instead?

client/js/constants.js
McInkay: I'm not sure actually. There are a few commands in here that are pass-through commands that we don't do anything with (`as`, `bs`, even `join`) so it could be a weird specific one like that?

client/js/constants.js
McInkay: Well, we have all the other aliases in this list, so it would be odd to not have 1 of the aliases of this particular one.

PR #2757, opened 2018-09-06
Add support for command plugins

McInkay:
A few things I added to this, we have the command to add a command, which uses the same structure as the internal plugins currently. Secondly, if you return a string, it will assume it is a new message to parse, and pass it through the same function again (so we don't have to do the same logic for sending messages, or actually map to proper IRC commands as we do that elsewhere).

The other thing I wasn't 100% sure about was...currently there's no way to change the target. I assume this is the best way to do things as we probably don't want plugins to be able to change the target, do we? But should we require plugins to return an object rather than a string, just incase there's something else we want to add there at some point? Thoughts?

Example plugin using this api:

```
const shrug = {
  input: function(_, _, _, args, runCommand) {
    runCommand(args.join(" ").replace(/^\//, "") + " ¯\\_(ツ)_/¯");
  }
};

module.exports = {
  onServerStart: thelounge => {
    thelounge.Commands.add("shrug", shrug);
  }
};
```


src/server.js
```diff
@@ -567,6 +568,7 @@ function initializeClient(socket, client, token, lastMessage) {
 			active: client.lastActiveChannel,
 			networks: client.networks.map((network) => network.getFilteredClone(client.lastActiveChannel, lastMessage)),
 			token: tokenToSend,
+			commands: constants.getCommands()
```
xPaw: Let's not overload initial events with too much information. This could easily be emitted as a separate event.

fnutt:
https://github.com/thelounge/thelounge/issues/1025

src/server.js
xPaw: Is that a problem to initialize command autocomplete later on?

McInkay: Well, it means that there would be a time period between the user being able to type commands, and the list of comamnds changing. Meaning they would initially not see all possible commands in the autocomplete. Could be confusing.

xPaw: You removed the command list from the client, so it simply won't complete any commands on initial load.

However, if you send commands in a separate event right after `init`, I don't see it being a problem (lazy load is fine by me, you don't usually type commands instantly after init).

I can also see this potentially being re-used if new commands are added while the server is running (plugin auto load, or whatever we end up doing).

src/constants.js
```diff
@@ -0,0 +1,51 @@
...
+	"nick",
+	"notice",
+	"query",
+	"quit",
+	"raw",
+	"topic",
+	"list",
+	"whois",
+].reduce(function(plugins, name) {
+	const plugin = require(`./plugins/inputs/${name}`);
```
xPaw: Can we think of a better solution of loading core plugins? I don't like this being in a file called constants

src/constants.js
```diff
@@ -0,0 +1,51 @@
...
+    plugin.commands.forEach((command) => plugins[command] = plugin);
+    return plugins;
+}, {});
+
+const getCommands = () => Object.keys(inputs).map((command) => `/${command}`).concat(clientSideCommands).concat(passThroughCommands);
+
+module.exports = {
+    getCommands, inputs
+}
```
xPaw: new line

src/server.js
McInkay: All good points.

> I can also see this potentially being re-used

Yeah, this alone is probably big enough of an advantage to warrant the small time period of having no autocomplete.

src/constants.js
McInkay: We can just have this file list the core commands and we can import the files in another folder. I guess the thing is, it makes perfect sense to have the list of command names in here, but somewhere else (I'll have a think later) can import them specifically?

McInkay:
I think I have addressed your points, @xPaw

src/constants.js
McInkay: So, realisticall,y this isn't really a constants file. So I've renamed it and moved it into the `inputs` folder, as that's all it's referring to.

src/server.js
McInkay: I have moved this into it's own event. It's pretty small, but does the job. Nice and simple code :-P

src/client.js
```diff
@@ -362,12 +336,16 @@ Client.prototype.inputLine = function(data) {
...
+	if (Object.prototype.hasOwnProperty.call(inputs.userInputs, cmd) && typeof inputs.userInputs[cmd].input === "function") {
+		const plugin = inputs.userInputs[cmd];
 
 		if (connected || plugin.allowDisconnected) {
 			connected = true;
-			plugin.input.apply(client, [target.network, target.chan, cmd, args]);
+			const response = plugin.input.apply(client, [target.network, target.chan, cmd, args]);
+
+			if (response && typeof response === "string") {
+				this.inputLine({target: data.target, text: response});
```
xPaw: Can you explain this to me? Plugins can return a string that would execute as a new command...?

xPaw:
Got this:
```
Uncaught TypeError: Cannot read property 'slice' of undefined
    at completeCommands (autocompletion.js:313)
```

src/plugins/inputs/index.js
```diff
@@ -0,0 +1,53 @@
...
+	const plugin = require(`./${name}`);
+	plugin.commands.forEach((command) => (plugins[command] = plugin));
+	return plugins;
+}, {});
+
+const getCommands = () =>
+	Object.keys(userInputs)
+		.map((command) => `/${command}`)
+		.concat(clientSideCommands)
+		.concat(passThroughCommands);
```
xPaw: This probably should be sorted.

src/client.js
McInkay: Yeah, so basically it's the easiest way to alias to another command if that's what we want to do.

So, if someone wants to alias `/j` to `/join`, we'd have to pass all the commands to them and they'd then have to pick out the right one and call it with the right params.

This way, we just take what they return and throw it through the parser again, re-using all the existing logic.

src/plugins/inputs/index.js
McInkay: What you referring to?

src/client.js
xPaw: I'm not a fan of putting this specific functionality here, can't the plugin call `client.inputLine` or have a better framework in place for aliasing.

Strings being interpreted as new input feels like magic to me.

src/plugins/inputs/index.js
xPaw: The final constructed array should be sorted if we want to match the previously hardcoded array command we had.

McInkay: Ahhh sorted, gotcha. Yup, good catch.

src/client.js
McInkay: Sure, I'm happy with providing the plugin with `client.inputLine`. I'll call it something different for them, obviously. 

It's not just aliasing it's useful for. For example, if you look at the shrug command I put in the title, it just sends the message it is passed with shrug as well. So there are many other useful things this gives us, so I wouldn't want to replace this with a specific aliasing thing.

But I think that explicitly passing the `inputLine` command (with a better name) to the plugin would be a good way to solve it. I guess the issue is that we are interpreting the return value, which is very transparent magic?

McInkay:
Alright, I have fixed the error with commands, and I have defaulted it to `[]` in the file, so that if we don't get them before we try to autocomplete, we don't get that error.

I've also sorted the commands in the function.

I still need to pass the `inputLine` function, will do that later.

McInkay:
I've put in a second piece to the public API that will allow us to run commands directly from plugins (for example, we might want to listen for another event, and then run a command for something), but also keeping things simple for aliases, etc.

The shrug example from above is now:

```
const shrug = {
  input: function(_, _, _, args, runCommand) {
    runCommand(args.join(" ").replace(/^\//, "") + " ¯\\_(ツ)_/¯");
  }
};

module.exports = {
  onServerStart: thelounge => {
    thelounge.Commands.add("shrug", shrug);
  }
};
```

This means that plugins now get a `runCommand` function, meaning that they can just forward any command to the parser again. In this example, it just sends the command back (minus `/shrug`) with the shrug on the end. 

However, we also have the public API, allowing us to do something like:

```
thelounge.Commands.run(command, previouslyStoredUser, target);
```

In this case, obviously the plugin will need to keep track of the lounge users that it is needing to do something with, and that's totally fine. But in the above example, they don't need to care, they are just changing the command that is sent. 

We had quite a bit of a chat in irc about what the best thing to do is, and I've come to the conclusion personally that only doing the public API makes it overly complicated. An exmaple of what they'd need to do with this:

```
const shrug = api => ({
  input: function(_, _, _, args, targetId) {
    api.Commands.run(
      args.join(" ").replace(/^\//, "") + " ¯\\_(ツ)_/¯",
      this.name,
      target
    );
  }
});

module.exports = {
  onServerStart: thelounge => {
    thelounge.Commands.add("shrug", shrug(thelounge));
  }
};
```

So in this one, the plugin has to pass through the name of the lounge user, and has to be passed the target id and pass it through to the API as well, because the API is essentially stateless, it has no information about what user is running things. 

There may be a nicer way to do all of this, but I feel like this feature is growing arms and legs nad this feels like a reasonable middle ground.

PR #2760, opened 2018-09-07
Add check for outdated packages, and show on the help screen.

McInkay:


src/packageManager.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
```
xPaw: We have `src/plugins/packages/index.js`, which is probably better suited to be there

McInkay: Yeah, good shout. I'll do that when I fix the conflicts for this.

src/packageManager.js
```diff
@@ -0,0 +1,35 @@
...
+
+const fs = require("fs");
+const path = require("path");
+const Helper = require("./helper");
+const log = require("./log");
+const Utils = require("./command-line/utils");
+
+const outdated = () => {
```
McInkay: Add caching to this, similar to `changelog.fetch`

McInkay:
Ok, I have moved the code into the `packages/index.js` and I have added a cache for it.

I do want to add a way for `thelounge upgrade` to clear that cache, but we currently don't have a way to talk between lounge processes, and that's probably bigger than this PR.

src/plugins/packages/index.js
```diff
@@ -86,3 +95,44 @@ function loadPackages(clientManager) {
...
+	// Check if the configuration file exists
+	if (!fs.existsSync(packagesConfig)) {
+		log.warn("There are no packages installed.");
+		return;
+	}
+
+	// If we get an error from calling outdated and the code isn't 0, then there are no outdated packages
+	await Utils.executeYarnCommand(...argsList)
+		.then(() => updateOutdated(false))
+		.catch((code) => updateOutdated(code === 0 ? false : true));
```
xPaw:
```suggestion
		.catch((code) => updateOutdated(code !== 0));
```

src/plugins/packages/index.js
```diff
@@ -86,3 +95,44 @@ function loadPackages(clientManager) {
...
+		"--ignore-scripts",
+		"--non-interactive",
+		"--cwd",
+		packagesPath,
+	];
+
+	// Check if the configuration file exists
+	if (!fs.existsSync(packagesConfig)) {
+		log.warn("There are no packages installed.");
+		return;
```
xPaw: Probably should be

```suggestion
		return false;
```

src/plugins/packages/index.js
```diff
@@ -86,3 +95,44 @@ function loadPackages(clientManager) {
 		log.info(`Package ${colors.bold(packageName)} loaded`);
 	});
 }
+
+async function outdated() {
+	if (cache.outdated !== undefined) {
```
xPaw:
```suggestion
	if (typeof cache.outdated !== "undefined") {
```

src/command-line/outdated.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
+
+const program = require("commander");
+const Utils = require("./utils");
+const packageManager = require("../packageManager");
```
xPaw: Error: Cannot find module '../packageManager'

xPaw:
I think the `outdated` command should print which packages are out of date.

src/plugins/packages/index.js
McInkay: Oh duh, rookie error. I was clearly half asleep.

src/plugins/packages/index.js
McInkay: I'm not sure this is unnecessary, because we specifically specify it is undefined and then delete it, so when would this be different?

src/command-line/outdated.js
McInkay: Yeah, fixed this.

McInkay:
> I think the outdated command should print which packages are out of date.

Can we add that in another PR? It's much bigger than what I'm doing here, and is much more complicated. This is useful in and of itself, and we can add that later.

McInkay:
I think this is ready? Unless there's something I missed.

PR #2761, opened 2018-09-07
Add server config api

McInkay:
- [x] Add documentation to site
- [x] Rebase

### DIFF
--- a/src/plugins/packages/index.js
+++ b/src/plugins/packages/index.js
@@ -20,6 +20,9 @@ const packageApis = function(packageName) {
 		Stylesheets: {
 			addFile: addStylesheet.bind(this, packageName),
 		},
+		Config: {
+			getConfig: () => Helper.config,
+		},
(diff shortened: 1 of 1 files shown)


McInkay:
This os a real simple one, but could be really useful. Allows plugins to view and change config. 

I guess the question issss...is there anything in this we want to make sure we black out. Don't want plugins reading passwords for example.

xPaw:
> Don't want plugins reading passwords for example.

They will be able to just use `fs` and read the file directly?

McInkay:
> They will be able to just use fs and read the file directly?

Good point, but do we want to give them an official way? Or do we not care?

xPaw:
@McInkay document this in the site repo.

McInkay:
Will do that ...  probably Monday.

McInkay:
Done. thelounge/thelounge.github.io#212

PR #2787, opened 2018-09-20
Use Date.toISOString for logging

xPaw:
@astorije thinks we would avoid more troubles by using full ISO format.

### DIFF
--- a/src/log.js
+++ b/src/log.js
@@ -2,10 +2,11 @@
 
 const colors = require("chalk");
 const read = require("read");
-const moment = require("moment");
 
 function timestamp() {
-	return colors.dim(module.exports.getHumanDate());
+	const datetime = (new Date()).toISOString().split(".")[0].replace("T", " ");
+
+	return colors.dim(datetime);
 }
 
 module.exports = {
@@ -30,9 +31,5 @@ module.exports = {
 	prompt(options, callback) {
 		options.prompt = [timestamp(), colors.cyan("[PROMPT]"), options.text].join(" ");
 		read(options, callback);
-	},
-
-	getHumanDate(ts) {
-		return moment(ts).format("YYYY-MM-DD HH:mm:ss");
 	},
 };
--- a/src/plugins/messageStorage/text.js
+++ b/src/plugins/messageStorage/text.js
@@ -40,7 +40,7 @@ class TextFileMessageStorage {
 			return;
 		}
 
-		let line = `[${log.getHumanDate(msg.time)}] `;
+		let line = `[${msg.time.toISOString()}] `;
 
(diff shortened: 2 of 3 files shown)


src/log.js
```diff
@@ -32,7 +31,11 @@ module.exports = {
...
 	},
 
-	getHumanDate(ts) {
-		return moment(ts).format("YYYY-MM-DD HH:mm:ss");
+	getHumanDate(date) {
+		if (!date) {
+			date = new Date();
+		}
+
+		return date.toISOString().split(".")[0] + "Z";
```
xPaw: This drops `.000` milliseconds from the string. Unless we want to keep it?

astorije: I would vote to keep them. Added precision can't hurt, esp. if it makes the code simpler.
Is it always `.000` or some/all servers send precision to milliseconds? I don't know, even if it's always `.000`, it might not hurt to keep them, future-wise.

src/log.js
```diff
@@ -32,7 +31,11 @@ module.exports = {
 		read(options, callback);
 	},
 
-	getHumanDate(ts) {
-		return moment(ts).format("YYYY-MM-DD HH:mm:ss");
+	getHumanDate(date) {
+		if (!date) {
+			date = new Date();
+		}
```
astorije: Alternatively, we could do `return new Date(date).toISOString()`, as `new Date()`ing a date still returns a similar date (extra object, but whatever IMO).

astorije:
> @astorije thinks we would avoid more troubles by using full ISO format.

Well, it also helps your plan of getting rid of moment :D

src/log.js
astorije: Note to self, you mentioned on IRC that time is already UTC and not server time. In https://github.com/thelounge/thelounge/pull/2506 we mentioned server time, so I need to edit the upgrade guide accordingly (I also prefer storing logs in UTC, personally, so that's good).

For the record:

```
$ date
Thu Sep 20 21:18:12 EDT 2018
$ node
> new Date().toISOString()
'2018-09-21T01:18:18.274Z'
```

🤷‍♂️

xPaw: @McInkay What do you think?

McInkay: I guess the only real reason to remove the milliseconds is if we assume there are some people using scripts to read their logs. Which I assume there probably aren't many, if any. And we are already changing the format, aren't we?

xPaw:
Changed text logs to call `toISOString` on `time` directly. And changed console log code to preserve the exact same format we currently have (`YYYY-MM-DD HH:mm:ss`), but merely dropped `moment` dependency.

PR #2814, opened 2018-09-29
Specify minimum Yarn version and upgrade yarn.lock with intengrity

xPaw:
Yarn 1.10 and up now automatically add integrity field to yarn.lock.

I specified it in `engines` because lower version will *remove* the field because they're unaware of it.

Renovate bot will also update integrity fields if it detects `integrity` in yarn.lock.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
   ],
   "license": "MIT",
   "engines": {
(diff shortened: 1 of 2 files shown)


xPaw:
Forcing Yarn version so soon is not a good idea.

PR #2817, opened 2018-10-01
Remove underline from textcomplete dropdown

### LINKED ISSUE #2816: underline on links in emoji autocomplete shouldn't underline emoji
maybe this is intentional, but I think it would look a lot better if the emoji was not always underlined

screen shot 2018-10-01 at 4 06 32 pm

xPaw:
The `<a>` tag doesn't cover the entire line, and we change the background color. This matches context menus.

Fixes #2816


MaxLeiter:
closes #2816

PR #2818, opened 2018-10-01
Add setting to hide version from CTCP query

RockyTV:
I don't know if admins should be able to configure custom VERSION replies. 
This just adds a setting that allows you to disable it, returning `thelounge -- https://thelounge.chat` instead of `thelounge v3.0.0-rc.1 -- https://thelounge.chat`.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -176,6 +176,14 @@ module.exports = {
 	// default.
 	leaveMessage: "The Lounge - https://thelounge.chat",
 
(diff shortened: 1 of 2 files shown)


RockyTV:
I had no idea that you reached the agreement to keep showing the client version :yum:

Would `defaultMessage` be a user defined reply to `CTCP VERSION` or any query?

xPaw:
Closing as per @astorije's comment.

PR #2821, opened 2018-10-02
Add file size to link preview

### LINKED ISSUE #2754: File size on links in chat
We currently have config for not showing previews on images larger than X mb. But, what if the image is above this limit and you wonder if you should open it or not?  What about pdfs and other files we do not show previews of?

I can see the usefulness of knowing the file size when you're not at home and have limited data traffic. Considering we have implemented file uploading, there will (my guess) probably be more files and images shared in the future since it's far easier now - making it even more useful to know the file sizes.

I suggest: https://www.test.com/test.pdf (X MB)

skjermbilde 2018-09-05 kl 13 19 47

Raqbit:
This adds the file size of the content next to the preview toggle button.

A couple of things that might want to get added as well:

* Some styling? Currently it looks okay but it might blend too much with the message.

How it looks:

filesizes

Would really like some input from @fnutt :).

Fixes #2754

### DIFF
--- a/client/components/LinkPreviewFileSize.vue
+++ b/client/components/LinkPreviewFileSize.vue
@@ -1,0 +1,34 @@
+<template>
+	<span class="preview-size">({{ previewSize }})</span>
+</template>
+
+<script>
+const constants = require("../js/constants");
+
+export default {
+	name: "LinkPreviewFileSize",
+	props: {
+		size: Number,
+	},
+	computed: {
+		previewSize() {
+			let size = this.size;
+
+			// Threshold for SI units
+			const thresh = 1024;
+
+			let u = 0;
+
+			do {
+				size /= thresh;
+				++u;
+			} while (size >= thresh && u < constants.sizeUnits.length - 1);
+
+			const displaySize = size.toFixed(1);
+			const unit = constants.sizeUnits[u];
+
+			return `${displaySize} ${unit}`;
+		},
+	},
+};
+</script>
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1475,6 +1475,11 @@ background on hover (unless active) */
 	border-left: 1px solid var(--highlight-bg-color);
 }
 
+#chat .preview-size {
+	margin-left: 5px;
+	user-select: none;
+}
+
 #chat .toggle-content.opened .more-caret, /* Expand/Collapse link previews */
 #chat .toggle-button.opened, /* Thumbnail toggle */
 #chat .msg.condensed:not(.closed) .toggle-button { /* Expanded status message toggle */
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -27,10 +27,13 @@ const timeFormats = {
 	msgWithSeconds: "HH:mm:ss",
 };
 
+const sizeUnits = ["B", "KiB", "MiB", "GiB", "TiB"];
+
 module.exports = {
 	colorCodeMap,
 	commands: [],
 	condensedTypes,
 	condensedTypesQuery,
 	timeFormats,
+	sizeUnits,
 };
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -9,6 +9,7 @@ const merge = require("./ircmessageparser/merge");
 const colorClass = require("./colorClass");
 const emojiMap = require("../fullnamemap.json");
 const LinkPreviewToggle = require("../../../components/LinkPreviewToggle.vue").default;
+const LinkPreviewFileSize = require("../../../components/LinkPreviewFileSize.vue").default;
 const emojiModifiersRegex = /[\u{1f3fb}-\u{1f3ff}]/gu;
 
 // Create an HTML `span` with styling information for a given fragment
@@ -96,12 +97,16 @@ module.exports = function parse(createElement, text, message = undefined, networ
 
 		// Wrap these potentially styled fragments with links and channel buttons
 		if (textPart.link) {
-			const preview = message && message.previews.find((p) => p.link === textPart.link);
+			const preview =
+				message &&
+				message.previews &&
+				message.previews.find((p) => p.link === textPart.link);
 			const link = createElement(
 				"a",
 				{
 					attrs: {
 						href: textPart.link,
+						dir: preview ? null : "auto",
 						target: "_blank",
 						rel: "noopener",
 					},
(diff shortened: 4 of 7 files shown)


MaxLeiter:
I think showing the `+` if it exceeds 2MB is necessary. Otherwise it's misleading.

astorije:
As the error message says `This image is larger than 2 MB and cannot be previewed`, I would say that the link doesn't need the size in parentheses at all.

fnutt:
> As the error message says `This image is larger than 2 MB and cannot be previewed`, I would say that the link doesn't need the size in parentheses at all.

Yes, you do see the image is larger than 2 MB, but what is the size? Wether or not you open the image even though it's larger than 2 MB probably highly depends on the size of the image, especially if you're on mobile connections with limited data.

If we don't have the size of the image/file behind urls, can we atleast have this info in the "This image is larger than...." message? Also, what happens when it's pdf? or zip? etc.

dgw:
It's definitely possible to know the total size if the remote server sends an accurate `Content-Length` header. That code aborts without even fetching anything if the content-length is provided and is too big.

williamboman:
I feel like this is a bit odd for regular links (HTML documents, pretty much). I feel like it makes more sense to only display file size for URLs that match `/\.((?!html)\w+)$/i`, or if the URI doesn't return a `Content-Type: text/html`, I don't know.

client/views/msg_preview_filesize.tpl
```diff
@@ -0,0 +1,5 @@
+{{#preview}}
+    {{#if showSize}}
+        <span class="preview-size">({{friendlysize size}})</span>
```
williamboman: Add a space before opening parenthesis? Looks kinda weird when there's no whitespace between the link and the file size.

McInkay:
For this one, I'm totally for showing a size if it's above the size limit, and it's an image (not a website), and we know the size. If we don't know the size, but it's still above the size limit, we should say that we don't know the size.

McInkay:
@Raqbit Will you have any time to rewrite this in Vue at all? We are trying to clear up old PRs so we don't have loads hanging around that have had nothing done in ages. I think we are in agreement that it show the size if it's above the size limit, and it's specifically an image, so we'd be happy to get that in, but it needs rewritten.

Raqbit:
@McInkay Yes, I'll have time to work on this soon-ish.

Raqbit:
@McInkay I think the branch reflects the old version again, but in Vue this time.

McInkay:
Great, thanks for that. Will review soon.

client/js/constants.js
```diff
@@ -27,10 +27,13 @@ const timeFormats = {
 	msgWithSeconds: "HH:mm:ss",
 };
 
+const sizeUnits = ["B", "kB", "MB", "GB", "TB"];
```
xPaw:
```suggestion
const sizeUnits = ["B", "KiB", "MiB", "GiB", "TiB"];
```

Since it's being divided by 1024, let's use actual mebibyte units to prevent any potential confusion.

src/plugins/irc-events/link.js
```diff
@@ -55,6 +55,8 @@ module.exports = function(client, chan, msg) {
 			head: "",
 			body: "",
 			thumb: "",
+			size: -1,
+			showSize: true,
```
xPaw: Can we perhaps combine this logic into just `size`? As far as I can see it's only being set to false for HTML content and errors. Keeping size at `-1` in these cases should be fine, and then only display size if its greater than zero.

client/components/LinkPreviewFileSize.vue
```diff
@@ -0,0 +1,33 @@
...
+const constants = require("../js/constants");
+
+export default {
+	name: "LinkPreviewFileSize",
+	props: {
+		size: Number,
+	},
+	computed: {},
+	methods: {
+		formatPreviewSize(size) {
```
xPaw: Perhaps this is better suited as `computed`?

client/components/LinkPreviewFileSize.vue
```diff
@@ -0,0 +1,35 @@
...
+				++u;
+			} while (size >= thresh && u < constants.sizeUnits.length - 1);
+
+			const displaySize = size.toFixed(1);
+			const unit = constants.sizeUnits[u];
+
+			return `${displaySize} ${unit}`;
+		},
+	},
+	methods: {},
```
xPaw:
```suggestion
```

Not necessary.

xPaw:
Copying a link doesn't put a space before the size: `unknown.png(209.2 KiB)`. Perhaps it shouldn't be selectable/copyable at all?

fnutt:
> Copying a link doesn't put a space before the size: `unknown.png(209.2 KiB)`. Perhaps it shouldn't be selectable/copyable at all?

I agree. It's not information that should be copied.

client/js/libs/handlebars/parse.js
```diff
@@ -97,35 +98,53 @@ module.exports = function parse(createElement, text, message = undefined, networ
...
+			if (preview && preview.size > 0) {
+				linkEls.push(
+					createElement(
+						LinkPreviewFileSize,
+						{
+							props: {
+								size: preview.size,
+							},
+						},
+						fragments
```
xPaw: This `fragments` is not needed.

client/js/libs/handlebars/parse.js
```diff
@@ -97,35 +98,53 @@ module.exports = function parse(createElement, text, message = undefined, networ
...
+				linkEls.push(
+					createElement(
+						LinkPreviewToggle,
+						{
+							class: ["toggle-button", "toggle-preview"],
+							props: {
+								link: preview,
+							},
+						},
+						fragments
```
xPaw: Here too.

xPaw:
@Raqbit I fixed up the code for you to how it should be with the RTL fixes.
```js
		if (textPart.link) {
			const preview = message && message.previews.find((p) => p.link === textPart.link);
			const link = createElement(
				"a",
				{
					attrs: {
						href: textPart.link,
						dir: preview ? null : "auto",
						target: "_blank",
						rel: "noopener",
					},
				},
				fragments
			);

			if (!preview) {
				return link;
			}

			const linkEls = [link];

			if (preview.size > 0) {
				linkEls.push(
					createElement(LinkPreviewFileSize, {
						props: {
							size: preview.size,
						},
					})
				);
			}

			linkEls.push(
				createElement(LinkPreviewToggle, {
					props: {
						link: preview,
					},
				})
			);

			// We wrap the link, size, and the toggle button into <span dir="auto">
			// to correctly keep the left-to-right order of these elements
			return createElement(
				"span",
				{
					attrs: {
						dir: "auto",
					},
				},
				linkEls
			);
		}
```

Some tests will need to have `dir="auto"` added to links because that was missed in the RTL pr, and I'm just adding it here for the sake of reducing noise.

client/css/style.css
```diff
@@ -1482,6 +1482,14 @@ background on hover (unless active) */
 	border-left: 1px solid var(--highlight-bg-color);
 }
 
+#chat .preview-size {
+	margin-left: 5px;
+	-webkit-user-select: none;
+	-moz-user-select: none;
+	-ms-user-select: none;
```
xPaw: Remove the verdor prefixes, if you rebase to master again, we have autoprefixer now.

richrd:
Awesome, thanks a lot @Raqbit ! Nice to get this in finally :) :tada:

PR #2823, opened 2018-10-03
Fix for font-src Content Security Policy directive

slaykovsky:
Now I get lots of "Refused to load the font '<data URL>' because it
violates the following Content Security Policy directive:
"font-src 'self' https:".

This patch fixes adds 'data:' directive and solves the problem.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -257,7 +257,7 @@ function index(req, res, next) {
 		"worker-src 'self'", // service worker
 		"child-src 'self'", // deprecated fall back for workers, Firefox <58, see #1902
 		"manifest-src 'self'", // manifest.json
-		"font-src 'self' https:", // allow loading fonts from secure sites (e.g. google fonts)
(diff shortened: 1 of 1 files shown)


xPaw:
Can you share your CSS that triggers the warning for `data:`?

slaykovsky:
Hi @xPaw, well, it's default CSS, I just started thelounge and on login page there are tens of them.

xPaw:
That's odd because default CSS uses separate font file.

What browser/OS?

slaykovsky:
Chrome 69.0.3497.100, macOS Mojave

xPaw:
Yeah that definitely shouldn't happen. Are you sure it's not an extension or something? Try to find the actual CSS that triggers it.

slaykovsky:
I did remove Grammarly and everything is okay now. So I guess it's Grammarly problem then?

xPaw:
Yep, they're injecting content into the page.

PR #2824, opened 2018-10-03
Fix stored image previews not being dereferenced

xPaw:
Broken in 64ebe0f4372b789e79033b7307d0259360d96ed1 (#1303) when we added support for multiple previews in a single message


capi:
I have tested your changes, it fixes the issue for me. I tested both, images being references from only one channel and images being referenced from multiple channels. In both cases, files were now deleted as soon as the last channel which referenced the image had been parted.

PR #2828, opened 2018-10-05
Put unread and highlight counts in aria-label in the channel list

xPaw:
Putting it in 3.0 because this worked on 2.x and was broken in commit e448dc711c164f26ee2d3bd16e1958cb9677320c as it added `aria-label` attribute.


astorije:
Nice!

PR #2849, opened 2018-10-13
Use hasOwnProperty when checking for webirc

xPaw:


src/models/network.js
```diff
@@ -153,7 +153,7 @@ Network.prototype.createWebIrc = function(client) {
 		this.hostname = this.ip;
 	}
 
-	if (Helper.config.webirc[this.host] instanceof Function) {
```
xPaw: `typeof` is preferred when checking for simple types

PR #2850, opened 2018-10-13
Enable Windows and macOS builds on Travis, remove AppVeyor

xPaw:
This should make it a little neater, especially after switching to `travis-ci.com`, which uses github checks.


.travis.yml
```diff
@@ -6,10 +6,20 @@ node_js:
...
-    - node_js: 8 # Version used to deploy to npm registry
+    - node_js: 8
+      os: windows
+      env: YARN_GPG=no # starts gpg-agent that never exits
+    - node_js: 8
+      os: osx
+    - name: "Production build"
+      node_js: 8 # Version used to deploy to npm registry
+      os: linux
       env: BUILD_ENV=production
```
astorije: Do we need to update `deploy.on` a bit to target linux only?

xPaw: Production only runs on linux as can be seen from https://travis-ci.com/thelounge/thelounge/builds/88431681

And even then, `deploy.on` has no `os` filter, I already tried that.

astorije:
@xPaw, AppVeyor build is failing, can you fix it? 😂

astorije:
@xPaw, is there anything else you want to do on this PR, or is it good to go?

xPaw:
It's good to go.

.travis.yml
astorije: Sorry, I misread the config, thanks!

.travis.yml
```diff
@@ -6,10 +6,20 @@ node_js:
...
+os:
+  - linux
+
 matrix:
   fast_finish: true
   include:
-    - node_js: 8 # Version used to deploy to npm registry
+    - node_js: 8
+      os: windows
+      env: YARN_GPG=no # starts gpg-agent that never exits
```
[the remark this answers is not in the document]
xPaw: `YARN_GPG=no` is needed because travis does not kill all processes on build finish (installing yarn starts gpg-agent). I dont know if they will fix it, but we will be able to remove this env if they do.

.travis.yml
```diff
@@ -6,10 +6,20 @@ node_js:
 - 8 # EOL: December 2019
 - 6 # EOL: April 2019
 
+os:
+  - linux
```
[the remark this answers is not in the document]
xPaw: I didn't include osx/windows for all node versions because these are indeed slower. osx is practically linux, so no reason to build it multiple times, and windows we didn't build multi versions on appveyor either.

PR #2877, opened 2018-10-28
Add Node.js v11 to the Travis CI matrix

astorije:


xPaw:
Do we want to test odd node versions? Their scheme is that even versions become lts and odd ones quickly eol.

astorije:
I believe we've always had stable versions in our matrix, we had v9 before.
I think we should always have the stable version listed at https://nodejs.org/en/ + whatever is in LTS cycle. People do install stable version and we can't really say TL doesn't support stable Node :/
Plus it's rather free to just have it in the CI to make sure basic tested stuff works 🤷‍♂️

PR #2907, opened 2018-11-27
Fix missing colors dependency, simplify version warning

xPaw:


index.js
```diff
@@ -5,32 +5,16 @@
...
+// Doing this check as soon as possible allows us to
+// avoid ES6 parser errors or other issues
 const pkg = require("./package.json");
 
 if (!require("semver").satisfies(process.version, pkg.engines.node)) {
-	let colors;
-	let log;
-
-	try {
-		colors = require("colors/safe");
```
xPaw: We switched `colors` to `chalk`. We can't really use it here though because `chalk` targets Node 6+, and that makes it useless here.

index.js
```diff
@@ -5,32 +5,16 @@
...
-	}
-
-	try {
-		log = require("./src/log");
-	} catch (e) {
-		log = {};
-		log.error = (msg) => console.error(`[ERROR] ${msg}`); // eslint-disable-line no-console
-	}
-
-	log.error(`The Lounge requires Node.js ${colors.green(pkg.engines.node)} (current version: ${colors.red(process.version)})`);
```
xPaw: The comment above talks about ES6 parsing errors, but we used template literals here.

scripts/generate-config-doc.js
```diff
@@ -9,7 +9,7 @@
 // ```
 
 const {readFileSync, writeFileSync} = require("fs");
-const colors = require("colors/safe");
```
xPaw: It worked here because it's still hoisted from `css-loader` dev dependency.

PR #2934, opened 2018-12-16
Don't require $HOME unless shortened paths are used

Mrmaxmeier:
While [packaging TheLounge for NixOS](https://github.com/NixOS/nixpkgs/pull/51947) I've run into an exception in `os.homedir()`.
This happened because the service is run as a systemd DynamicUser and the user doesn't have a home directory assigned.
The easy workaround setting `$HOME` to something. It'd be nice though to be able to remove the workaround when updating the package to v3.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -205,6 +205,12 @@ function expandHome(shortenedPath) {
 		return "";
 	}
 
+	// no need to lookup the home dir if path doesn't contain '~'.
+	// os.homedir might throw and nothing will be expanded anyways.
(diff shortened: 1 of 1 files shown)


xPaw:
I see no reason why you can't assign `HOME` to lounge's config folder in your package.

This PR doesn't actually fix this problem and it will still happen if `~` is used.

Mrmaxmeier:
Sure. I don't feel like the workaround is particularly annoying so I'll close this.

PR #2955, opened 2019-01-04
Catch LDAP bind error and search warning log messages in ldap tests

### LINKED ISSUE #2950: LDAP tests produce (expected) console messages
After I added extra logging in LDAP, it now logs messages into CLI when running tests:

```
2019-01-02 12:07:41 [ERROR] LDAP bind failed: InsufficientAccessRightsError: InsufficientAccessRightsError
2019-01-02 12:07:41 [WARN] LDAP Search did not find anything for: eve (0)
```

We should catch these messages so they are not printed (could even add asserts for them).

moundahiwale:
Fixes #2950

### DIFF
--- a/test/plugins/auth/ldap.js
+++ b/test/plugins/auth/ldap.js
@@ -5,6 +5,8 @@ const ldapAuth = require("../../../src/plugins/auth/ldap");
 const Helper = require("../../../src/helper");
 const ldap = require("thelounge-ldapjs-non-maintained-fork");
 const expect = require("chai").expect;
+const stub = require("sinon").stub;
+const TestUtil = require("../../util");
 
 const user = "johndoe";
 const wrongUser = "eve";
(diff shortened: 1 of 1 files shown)


xPaw:
Thanks for the PR!

We should however use stubs like it is done in other tests already, example file: https://github.com/thelounge/thelounge/blob/061fa615cea3297d97daf7d3054933d741c2b302/test/src/command-line/utilsTest.js

That way we will be able to also have a `expect()` call on the log message to test that it actually logs what we want.

test/plugins/auth/ldap.js
```diff
@@ -95,17 +95,27 @@ function testLdapAuth() {
...
 	it("should fail to authenticate with incorrect password", function(done) {
+		const originalLog = log.info;
+
+		log.error = () => {};
+
 		ldapAuth.auth(manager, client, user, wrongPassword, function(valid) {
 			expect(valid).to.equal(false);
 			done();
 		});
+		log.error = originalLog;
```
xPaw: You would need to do the restore inside of the callback

moundahiwale:
Tests updated to assert the actual log messages

test/plugins/auth/ldap.js
```diff
@@ -95,15 +97,25 @@ function testLdapAuth() {
...
 		});
 	});
 
 	it("should fail to authenticate with incorrect username", function(done) {
+		let warning = "";
+		stub(log, "warn").callsFake(TestUtil.sanitizeLog((str) => warning += str));
+
 		ldapAuth.auth(manager, client, wrongUser, correctPassword, function(valid) {
 			expect(valid).to.equal(false);
+			expect(warning).to.equal("LDAP Search did not find anything for: eve (0)\n");
```
xPaw: @astorije Any particular reason why `sanitizeLog` appends a new line?

PR #2959, opened 2019-01-06
Relax identd checks to make it more compatible

xPaw:
I looked at other identd implementations, and none of them checked the IP.

Quassel for example only checks the local port and disregards the fport, I decided to keep it as it comes as the query.

This change is required to make [oidentd forwarding](https://deploy-preview-192--thelounge.netlify.com/docs/guides/identdfile#using-oidentd-to-forward-requests-to-built-in-server) work, as the remote ip will be oidentd's ip, and not the IRC server's ip.


xPaw:
I suspect this also will fix issues in Docker: https://github.com/thelounge/thelounge-docker/issues/51

TheMythPT:
Tested by me in Freenode (EliteGod aka Exterminador) and all seems to work properly 😃

PR #2961, opened 2019-01-07
Only generate entries for commits/PRs that do not already exist in the CHANGELOG upon re-generation

astorije:
This is helpful when commits are added after starting to work on the changelog entry.

For example, I used this to generate the following using `node scripts/changelog.js 3.0.0`:

<details>
<summary>Diff</summary>

```diff
──────────────────────────
modified: CHANGELOG.md
──────────────────────────
@ CHANGELOG.md:7 @ All notable changes to this project will be documented in this file.

 <!-- New entries go after this line -->

+## v3.0.0 - 2019-01-06
+
+For more details, [see the full changelog](https://github.com/thelounge/thelounge/compare/v2.7.1...v3.0.0) and [milestone](https://github.com/thelounge/thelounge/milestone/28?closed=1).
+
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+@@ DESCRIPTION, ANNOUNCEMENT, ETC. @@
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+### Added
+
+### Changed
+
+
+
+### Deprecated
+
+
+
+### Removed
+
+### Fixed
+
+### Security
+
+
+
+### Documentation
+
+
+
+On the [website repository](https://github.com/thelounge/thelounge.github.io):
+
+- Replace `logs.format` in the example command as that no longer exists ([`3ff6d09`](https://github.com/thelounge/thelounge.github.io/commit/3ff6d091e082457ac5e5f9816d415e4dc346da61) by [@xPaw](https://github.com/xPaw))
+- browserl.ist supports https now ([`52df65c`](https://github.com/thelounge/thelounge.github.io/commit/52df65c85adba44b502cdaac9dfee14ef2f7acc6) by [@xPaw](https://github.com/xPaw))
+- Disallow robots in non-production ([#175](https://github.com/thelounge/thelounge.github.io/pull/175) by [@xPaw](https://github.com/xPaw))
+- Change reverse proxy configs to point to root ([#186](https://github.com/thelounge/thelounge.github.io/pull/186) by [@xPaw](https://github.com/xPaw))
+- Add extra aria attributes for the search field ([#185](https://github.com/thelounge/thelounge.github.io/pull/185) by [@xPaw](https://github.com/xPaw))
+- Add very minimal google analytics script ([#187](https://github.com/thelounge/thelounge.github.io/pull/187) by [@xPaw](https://github.com/xPaw))
+- Generate clientId as it is required ([#188](https://github.com/thelounge/thelounge.github.io/pull/188) by [@xPaw](https://github.com/xPaw))
+- SET the item ([`2f83b05`](https://github.com/thelounge/thelounge.github.io/commit/2f83b052746f7981359aa431e5d95b7bd1e84c92) by [@xPaw](https://github.com/xPaw))
+- Don't reference binding in reverse proxy guide ([#189](https://github.com/thelounge/thelounge.github.io/pull/189) by [@xPaw](https://github.com/xPaw))
+- Fix info message [Install and upgrade] ([#191](https://github.com/thelounge/thelounge.github.io/pull/191) by [@GewoonYorick](https://github.com/GewoonYorick))
+- Don't add canonical tag for 404 pages ([#190](https://github.com/thelounge/thelounge.github.io/pull/190) by [@xPaw](https://github.com/xPaw))
+
+### Internals
+
+
+- Update development dependencies to their latest versions:
+  - `webpack-dev-server` ([#2945](https://github.com/thelounge/thelounge/pull/2945))
+
+@@@@@@@@@@@@@@@@@@@
+@@ UNCATEGORIZED @@
+@@@@@@@@@@@@@@@@@@@
+
```

</details>

I can't say that I'm proud of this code but, hey, it just works.

### DIFF
--- a/scripts/changelog.js
+++ b/scripts/changelog.js
@@ -59,6 +59,8 @@ const token = process.env.CHANGELOG_TOKEN;
 const readFile = util.promisify(fs.readFile);
 const writeFile = util.promisify(fs.writeFile);
 
+const changelogPath = "./CHANGELOG.md";
+
 // CLI argument validations
 
 if (token === undefined) {
@@ -115,6 +117,13 @@ yarn global add thelounge@next
 `;
 }
 
+// Check if the object is empty, or if all array values within this object are
+// empty
+function isEmpty(list) {
+	const values = Object.values(list);
+	return values.length === 0 || values.every((entries) => entries.length === 0);
+}
+
 function stableTemplate(items) {
 	return `
 ## v${items.version} - ${items.date}
@@ -129,7 +138,7 @@ For more details, [see the full changelog](${items.fullChangelogUrl}) and [miles
 
 ### Changed
 
-${_.isEmpty(items.dependencies) ? "" :
+${isEmpty(items.dependencies) ? "" :
 		`- Update production dependencies to their latest versions:
 ${printDependencyList(items.dependencies)}`
 }
@@ -148,13 +157,13 @@ ${printList(items.security)}
 
 ### Documentation
 
-${_.isEmpty(items.documentation) ? "" :
+${items.documentation.length === 0 ? "" :
 		`In the main repository:
 
 ${printList(items.documentation)}`
 }
 
-${_.isEmpty(items.websiteDocumentation) ? "" :
+${items.websiteDocumentation.length === 0 ? "" :
 		`On the [website repository](https://github.com/thelounge/thelounge.github.io):
 
 ${printList(items.websiteDocumentation)}`
@@ -162,8 +171,7 @@ ${printList(items.websiteDocumentation)}`
 
 ### Internals
 
-${printList(items.internals)}${
-	_.isEmpty(items.devDependencies) ? "" : `
+${printList(items.internals)}${isEmpty(items.devDependencies) ? "" : `
 - Update development dependencies to their latest versions:
 ${printDependencyList(items.devDependencies)}`}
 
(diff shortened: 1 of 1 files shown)


scripts/changelog.js
```diff
@@ -148,22 +157,21 @@ ${printList(items.security)}
 
 ### Documentation
 
-${_.isEmpty(items.documentation) ? "" :
+${items.documentation.length === 0 ? "" :
```
xPaw: Why some checks are `length` now and others are `isEmpty`?

scripts/changelog.js
```diff
@@ -651,6 +669,27 @@ function parse(entries) {
...
 
+function dedupeEntries(changelog, items) {
+	const isNewEntry = (entry) => !changelog.includes(printEntryLink(entry));
+
+	items.deprecations = items.deprecations.filter(isNewEntry);
+	items.documentation = items.documentation.filter(isNewEntry);
+	items.websiteDocumentation = items.websiteDocumentation.filter(isNewEntry);
+	items.internals = items.documentation.filter(isNewEntry);
+	items.security = items.documentation.filter(isNewEntry);
+	items.uncategorized.feature = items.uncategorized.feature.filter(isNewEntry);
```
xPaw: This could be iterated like dependencies below.

scripts/changelog.js
astorije: When the thing being checked is an array, I switched from `_.isEmpty` to `.length === 0`, and when the thing being checked is an object, I switched from `_.isEmpty` to my own `_.isEmpty`.

I needed to check that things like `dependencies: { foo: [] }` (empty because all entries would have been filtered out) would still be considered empty, which I couldn't do with `_.isEmpty`. Then I figured it wasn't worth keeping it for arrays, since it's a simple `length` check.

Does that make sense? It is also entirely possible I missed something obvious when switching...

astorije:
@xPaw, I cleaned up that function you commented on (hey, I did say this was not good programming :D). Do you find this better?

xPaw:
It's alright for a script you run once in a blue moon 😅

PR #2974, opened 2019-01-15
Set scrollbar properties for Firefox 64+

xPaw:
This should make the scrollbar thin on platforms that allow for it, and changes the scrollbar color in Morning theme so it doesn't stand out like a sore thumb.

Refs:
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color
https://drafts.csswg.org/css-scrollbars-1/


xPaw:
Turns out `scrollbar-width` is not inhereted and actually works in Firefox, so I applied it where its relevant.

Issue #2442, opened 2018-05-16
Minimize shortcut in Firefox on macOS doesn't work


sometoby:
On macOS, whichever shortcut The Lounge binds to ⌘M obscures Firefox' native minimize shortcut, making it impossible to minimize Firefox using ⌘M when a tab with The Lounge is active.

Jay2k1:
> whichever shortcut The Lounge binds to ⌘M

![image](https://user-images.githubusercontent.com/2070503/40117026-f42f953e-5915-11e8-9a87-0c009545c985.png)

It's not only Firefox - Chrome, Vivaldi and Safari are affected too, since ⌘M is the default shortcut for "minimize to Dock" on MacOS.

sometoby:
I'm unable to reproduce the problem in Safari.

Jay2k1:
Oh, indeed. I didn't actually test the behaviour, I just checked each browsers' "Window" menu.

![image](https://user-images.githubusercontent.com/2070503/40117307-e03dae5c-5916-11e8-8031-7118c53f7e63.png)

MaxLeiter:
This only occurs when the message input box isn't focused. Not too sure we can do anything here except change the keybind. As of #2215

MaxLeiter:
No, I mean I expected keybinds like bold and monospace to focus on the input box. If that’s not the actual behavior then I can’t reproduce this bug

xPaw:
Lounge's keybinds don't do anything (not even set focus) when input is focused. This means any OS/browser keybiding should work just fine if input isn't focus.

xPaw:
Is there anything we can/should do here?

As of now, if you don't have the input focused, TL's keybinds will not work, and the browser's one should.

Jay2k1:
OP's issue is that they want to minimize Firefox with cmd+M and TL inserts the monospace control char instead. This can be avoided by not having the input focused. I'd call this a minor annoyance.

What's worse is that in Safari, cmd+M *always* minimizes the window, even with input focused, so you cannot make monospace formatted text in Safari as of now. This should be fixed if possible, I just doubt that there's anything we can do except change keybinds, which should probably be avoided...

xPaw:
>This can be avoided by not having the input focused. I'd call this a minor annoyance.

Agreed.

> What's worse is that in Safari, cmd+M always minimizes the window, even with input focused

Can we add a keybind just for safari, like cmd+shift+M or something?

Jay2k1:
[...]

PR #2976, opened 2019-01-15
use ctrl for color hotkeys

williamboman:
This is to avoid overriding native keybindings in macOS.

I keep failing to format things cause I keep using <kbd>Ctrl</kbd>. I feel it's very unnatural (and dare I say uncommon) to use the <kbd>Cmd</kbd> key to control web content.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1813,7 +1813,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	margin-bottom: 0;
 }
 
-.is-apple #help .key-all,
+.is-apple #help .key-apple + .key-all,
 #help .key-apple {
 	display: none;
 }
--- a/client/views/windows/help.tpl
+++ b/client/views/windows/help.tpl
@@ -49,8 +49,8 @@
 
 	<div class="help-item">
 		<div class="subject">
-			<span class="key-all"><kbd>Alt</kbd> <kbd>Shift</kbd> <kbd>↓</kbd></span>
 			<span class="key-apple"><kbd>⌥</kbd> <kbd>⇧</kbd> <kbd>↓</kbd></span>
+			<span class="key-all"><kbd>Alt</kbd> <kbd>Shift</kbd> <kbd>↓</kbd></span>
 		</div>
 		<div class="description">
 			<p>Switch to the next lobby in the channel list.</p>
@@ -59,8 +59,8 @@
 
 	<div class="help-item">
 		<div class="subject">
-			<span class="key-all"><kbd>Alt</kbd> <kbd>Shift</kbd> <kbd>↑</kbd></span>
 			<span class="key-apple"><kbd>⌥</kbd> <kbd>⇧</kbd> <kbd>↑</kbd></span>
+			<span class="key-all"><kbd>Alt</kbd> <kbd>Shift</kbd> <kbd>↑</kbd></span>
 		</div>
 		<div class="description">
 			<p>Switch to the previous lobby in the channel list.</p>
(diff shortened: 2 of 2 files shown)


williamboman:
This is also how it used to work before (pre https://github.com/thelounge/thelounge/pull/1844) due to "overbinding" keybindings.

xPaw:
You will need to change the help page too.

client/css/style.css
```diff
@@ -1813,7 +1813,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	margin-bottom: 0;
 }
 
-.is-apple #help .key-all,
+.is-apple #help .key-apple + .key-all,
```
williamboman: Maybe this is too hacky?

xPaw:
I guess this fixes #2442 as well?

williamboman:
> I guess this fixes #2442 as well?

Yep

client/js/keybinds.js
```diff
@@ -166,7 +166,7 @@ const colorsHotkeys = {
 };
 
 for (const hotkey in colorsHotkeys) {
```
astorije: I know this is not your PR, but I don't think this should be named `colorsHotkeys` anymore...

astorije:
> I keep failing to format things cause I keep using Ctrl. I feel it's very unnatural (and dare I say uncommon) to use the Cmd key to control web content.

I'm not sure this is something we should do. I tested a few web apps that came to my mind (3 Google ones - Gmail, Hangouts, Google Docs -, Slack in the browser, the textarea where I'm typing this comment on GitHub, Telegram on the Web, even Open Collective's editor lol) and all of them used <kbd>cmd</kbd> on macOS, couldn't find a single one using <kbd>ctrl</kbd>. And as far as I know, **all** native apps use <kbd>cmd</kbd>.

I think it's more common the way it currently is (did you have any specific example in mind?). The unnatural part, I could see that if you are often switching back-and-forth between Mac and Windows, but maybe I'm missing something?

McInkay:
Either way, probably shouldn't be 3.0.1? If we are doing it, should probably be 3.1.0

dgw:
The only Mac app I can think of that uses Ctrl for formatting is Textual. (Used? It's been a while—I'm almost exclusively on Linux now.)

When I was a regular Mac user, shortcuts that _didn't_ use Cmd felt unnatural, because Cmd is the de facto Ctrl key on OS X.

I'll defer to @astorije's thorough testing for hard data, but anecdotally… I don't think this would be a good idea.

xPaw:
Sounds like using Cmd is better over all, so I suggest closing this and fix Cmd+M by adding a modifier to it (what's the common modifier on macs?) like Cmd+Alt+M or whatever.

xPaw:
> However, before we do so, I would like to ask: is it a real issue that we override the shortcut to minimize the window (especially since it's only when the message input is focused)?

See linked issue, apparently it doesn't work in Safari?

astorije:
Oh I didn't catch that one, right. Indeed, it doesn't, just tested to confirm. Then in that case I agree with your conclusion to close this and add Shift modifier to this shortcut.

dgw:
Can we make both with and without Shift work?

astorije:
Would that be confusing, you think? It seems a little unusual to have multiple shortcuts for a single action, I am fine only having <kbd>⌘</kbd><kbd>⇧</kbd><kbd>M</kbd> (this is what Confluence does, btw).
If we do have 2 shortcuts, in the Help page, do we document both, or just the one with the Shift modifier, and let the other one be _our little secret_?

dgw:
Re: documentation, depends how sneaky you want to get with sniffing browsers :smirk:

PR #2990, opened 2019-01-21
extend custom highlight regex

Jay2k1:
As discussed and tested on irc, this extends matching for custom highlights. Main reason was that a word defined in custom highlights followed by e.g. a colon, a comma or an exclamation mark, such as `Jay: hi!` or `Hi Jay!`, wouldn't be caught by the regex. 

For some tests, see https://regex101.com/r/DDGVfN/11


xPaw:
Note: This PR is against the vue branch.

Jay2k1:
i will put this in a new PR as I can't seem to get this rebased properly

PR #3004, opened 2019-01-28
Bump minimum node version to 8 LTS

xPaw:
8.9.0 is the first Carbon LTS, as seen on https://nodejs.org/en/download/releases/

6 LTS expires in April, as per https://github.com/nodejs/Release

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -2,6 +2,9 @@
 
 root: true
 
+parserOptions:
+  ecmaVersion: 2017
+
 env:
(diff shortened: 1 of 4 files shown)


astorije:
Instead of v3.1, don't we want to target the first minor release that comes after April, to be thorough?

xPaw:
Not to mention both 6.x and 8.x are in maintenance LTS right now.

astorije:
Good points, thanks for the input.

PR #3010, opened 2019-01-30
Fix incorrect 404 error for files with unknown file type

xPaw:
For example a binary file like `libsomething.so`

### DIFF
--- a/src/plugins/uploader.js
+++ b/src/plugins/uploader.js
@@ -92,18 +92,19 @@ class Uploader {
 			const folder = name.substring(0, 2);
 			const uploadPath = Helper.getFileUploadPath();
 			const filePath = path.join(uploadPath, folder, name);
-			const type = Uploader.getFileType(filePath);
-			const mimeType = type || "application/octet-stream";
-			const contentDisposition = Uploader.isValidType(type) ? "inline" : "attachment";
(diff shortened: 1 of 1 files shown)


src/plugins/uploader.js
```diff
@@ -119,31 +120,32 @@ class Uploader {
...
-		} else if (isUtf8(buffer)) {
-			type = "text/plain";
+			if (isUtf8(buffer)) {
+				return "text/plain";
+			}
+
+			return null;
+		} catch (e) {
+			if (e.code !== "ENOENT") {
+				log.warn(`Failed to read ${filePath}`);
```
Zarthus: Could log the exception msg here

src/plugins/uploader.js
```diff
@@ -119,31 +120,32 @@ class Uploader {
 		return configOption * 1024;
 	}
 
+	// Returns false if an error occurred (file not found)
```
astorije: `string | false | null`, that's kind of a tricky return. What about returning an `Error` when an error occurred? Could also have a custom error instead of `null`, but that's fine.

xPaw: I changed it to `?string`.

Returns null on error, and the string can be empty if no type was detected.

PR #3011, opened 2019-01-30
Fix up link insertion after uploads to be saner

xPaw:
- If there's text before cursor, it will add a space before the url
- Will always put the cursor after the link + space
- Will no longer override selected text, instead will insert it where the selection starts

### DIFF
--- a/client/js/socket-events/uploads.js
+++ b/client/js/socket-events/uploads.js
@@ -1,10 +1,24 @@
 "use strict";
 
 const socket = require("../socket");
-const wrapCursor = require("undate").wrapCursor;
+const updateCursor = require("undate").update;
 
 socket.on("upload:success", (url) => {
-	const fullURL = new URL(url, location);
+	const fullURL = (new URL(url, location)).toString();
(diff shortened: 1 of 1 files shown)


client/js/socket-events/uploads.js
```diff
@@ -1,10 +1,22 @@
...
-	const fullURL = new URL(url, location);
+	const fullURL = (new URL(url, location)).toString();
 	const textbox = document.getElementById("input");
-	wrapCursor(textbox, fullURL, " ");
+	const initStart = textbox.selectionStart;
+
+	// Get the text before the cursor, and append a space if it's not in the beginning
+	const headToCursor = textbox.value.substr(0, textbox.selectionStart) + (textbox.selectionStart > 0 ? " " : "");
+
+	// Get the remaining text after the cursor and append a space
```
MaxLeiter: prepend?

xPaw: it appends in relation the text before cursor, not the link

PR #3019, opened 2019-02-02
plugin/ctcp: Let the user know a CTCP request was sent

Zarthus:
Because responding to a CTCP request is completely optional,
sometimes thelounge will just do absolutely nothing. (the request
was received, but the client did not respond to it)

This alleviates the problem by always notifying the user that
*something* was sent.

----

I tested some CLIENTINFOs on myself and the response was `* Zarthus CLIENTINFO VERSION` - which was a bit confusing to me.

----

img

### DIFF
--- a/src/plugins/inputs/ctcp.js
+++ b/src/plugins/inputs/ctcp.js
@@ -1,9 +1,23 @@
 "use strict";
 
+const Msg = require("../../models/msg");
+
 exports.commands = ["ctcp"];
 
 exports.input = function({irc}, chan, cmd, args) {
-	if (args.length > 1) {
-		irc.ctcpRequest(...args);
+	if (args.length < 2) {
+		chan.pushMessage(this, new Msg({
+			type: Msg.Type.ERROR,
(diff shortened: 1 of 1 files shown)


Zarthus:
Updated: I also changed the nickname to be from "You" to your own nickname.

updated

Zarthus:
You pretty much nailed it here:
>I'm also expecting the answer to be "yes but kind of a pain", so if you'd rather not do that and move on with your life, I'd get it and would still be okay with this PR :)

I think CTCPs aren't used often enough and the eventually new plugin system would/should/might improve error handling to make these kinds of checks redundant.

I originally tried to use "You", but my inexperience with vue, debugging hell, and some more inconveniences along the way just made it easier to keep it like this. I would wholly agree if this was some important component, but I don't think the average user sends or receives more than a single CTCP per week.

xPaw:
@astorije You alright with this going in as is? I am.

xPaw:
Merging due to 3 approvals, it can be improved upon later on.

PR #3031, opened 2019-02-05
Invites should not automatically highlight you.

creesch:
What it says in the title. Because this 
image

Is just silly as that was not me who got invited.


xPaw:
To clarify, this only happens with `invite-notify` cap.

PR #3037, opened 2019-02-07
Re-implement file uploading with old school multipart forms

### LINKED ISSUE #3032: Rewrite file uploading to use fetch()/FormData
Right now we use a library that sends files by manually chunking them over socket.io, but this can be done much simper by just using the old school file uploads where the browser itself will do all the necessary chunking when sending the file. Luckily all the modern browsers support `FormData` which simplifies this.

Server code would also be simplified a lot, as we won't need to need to rely on a third party library, which to be frank, has a lot of questionable decisions.

Here's how I think it should work:
- Create an express route to handle file uploads
- Over the existing socket.io connection, client asks to allow for a new file upload
- Server generates a one time use token (which can expire) and sends it to the client
- Client uses this token to upload a file to the new route

On server, we won't need to sanitize file names, as we always generate a random name for the file.

Implementing the uploading ourselves should also allow us to do stream transformation directly during the upload process (e.g. to detect jpeg files and strip EXIF data from it) without saving the file to disk and then copying it.

Potential problem: I think `fetch` is unable to provide current upload progress, but we can work this around by sending the progress over the socket.io connection.

xPaw:
In collaboration with @richrd. Fixes #3032

- This avoids sending data over the existing socket.io connection
- We no longer rely on a third party library that does all the file chunking, encoding and sending. This library has some questionable design decisions in the server implementation.
- Implement a proper file queue on the client, meaning all files will upload correctly one after another, even if you try to upload another file while the previous one hasn't finished yet
- Fixed dragging overlay from appearing when trying to drag text or link
- Makes it easier to implement uploading files with pasting

We did have to write our own code for it, but overall the implementation turned out to be much cleaner. On the server, we directly stream into the final file path without having to sanitize file names or deal with temporary files.

The existing socket client connection is only used to authenticate an upload request (one token per file), the entire upload process goes via `/uploads/new/{token}` route.


src/plugins/uploader.js
```diff
@@ -28,95 +29,194 @@ const whitelist = [
...
+		// generate a random output filename for the file
+		// we use do/while loop to prevent the rare case of generating a file name
+		// that already exists on disk
+		do {
+			randomName = crypto.randomBytes(8).toString("hex");
+			destDir = path.join(Helper.getFileUploadPath(), randomName.substring(0, 2));
+			destPath = path.join(destDir, randomName);
+		} while (fs.existsSync(destPath));
+
+		// we split the filename into subdirectires (by taking 2 letters from the begging)
```
MaxLeiter: beginning*

MaxLeiter: and subdirectories*

src/plugins/uploader.js
```diff
@@ -129,14 +229,17 @@ class Uploader {
...
 			if (file) {
 				return file.mime;
 			}
 
+			// if the buffer is a valid UTF-8 buffer, use text/plain
 			if (isUtf8(buffer)) {
 				return "text/plain";
 			}
 
+			// otherwise assume its random binary data
```
MaxLeiter: it's

src/plugins/uploader.js
xPaw: use github suggestions damn it!

astorije:
Mind fixing conflicts?
Is the PR ready to review? Is there anything you recommend being on the lookout for when testing/reviewing this?

richrd:
Yep, conflicts are resolved and this is ready for review.

The only thing I can think of that should be noted is, that if you switch to the settings or help page before an upload has completed it will throw `TypeError: textbox is null` since there is no text input rendered in the dom when the upload completes. Not a big deal and it's actually caused by the vue PR instead of this one but still something to keep in mind.

xPaw:
This is still ready for review/merge.

The issue @richrd described can and will be handled later on when we routing done in Vue because we need full state for this.

astorije:
I'll review this ASAP.
@xPaw, @richrd, for that bug, can we cancel any current upload when switching anywhere? Not as a complete solution, but temporary fix until the underlying issue is properly handled.

xPaw:
@astorije Okay I added code to abort current file upload if ChatInput component gets destroyed to avoid this problem.

We will need to re-visit this later when we do a final Vue porting pass.

astorije:
I gave a shot at cheap styling to get away from that ungracious `blue` lol (plagiarized from GitHub's progress bar).

I noticed an issue when uploading, getting to the Settings while the file was uploading, then back to the channel:

Screen Shot 2019-03-17 at 20 42 10

Screen Shot 2019-03-17 at 20 42 23

Anything I'm missing here?

astorije:
Actually, I am getting this error even without uploading. As soon as I click on Settings, the message is here when I get back to the channel.

xPaw:
@astorije Looks like I forgot to export it, fixed.

PR #3044, opened 2019-02-12
Fix copying text in Firefox

### LINKED ISSUE #3035: user-select:none breaks copying in Firefox
Try to copy multiple lines in chat, and if there's an expand button (links or condensed messages) it will only copy text before that.

I think this is caused by the fact that we only do `getRange(0)` here: 

https://github.com/thelounge/thelounge/blob/0f4e29d60dd923e25b9c2a47fe9dc9b129f15613/client/js/clipboard.js#L14

I haven't tested it, but if that's the case, we will need to read all of the ranges.

xPaw:
Fixes #3035.

1. Firefox makes `<button>` unselectable by default
2. Firefox creates multiple ranges which go around the unselectable button element

I have a Firefox bug created, but it's easier to just fix it in Lounge for now.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -103,6 +103,8 @@ button {
 	margin: 0;
 	outline: none;
 	padding: 0;
+	-moz-user-select: inherit; /* Firefox makes buttons unselectable by default which breaks copying */
+	user-select: inherit;
 }
 
 code,


xPaw:
This fix broke pointer cursor on buttons, so need to fix that.

PR #3045, opened 2019-02-12
Change build priority order in Travis

xPaw:
Windows build with cache disabled is the first build to be triggered, so in theory, the over all build time should be lower as Linux builds can pass faster one after another.

I moved versions around, so the current LTS/windows/macos/publishment version can be updated once without touching the matrix.

I also removed node-lts-linux-development build, because we already test lts version on windows/macos and on linux-production so this removes one build from the matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,32 @@
 language: node_js
 
 # https://github.com/nodejs/Release
+# Specify current LTS version here, which is used for publishing to npm
 node_js:
-  - 11 # EOL: June 2019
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,27 +1,33 @@
 language: node_js
 
 # https://github.com/nodejs/Release
+# Specify current LTS version here, which is used for publishment
```
astorije: Publishing :)

Issue #2446, opened 2018-05-21
Chat view moves out of place after loading old messages with inline media in Firefox


iiiGerardoiii:
* *Node version:* 
9.10.1
* *Browser version:* 
Firefox 61.0b5 / Chrome 66.0.3359.181
* *Device, operating system:* 
Windows 10
* *The Lounge version:* 
2.7.1
---

![8766](https://user-images.githubusercontent.com/6264194/40286198-9fd52500-5c6a-11e8-9d75-eaa3a191f368.gif)
Note: I didn't scroll after reaching the top. Sometimes it goes up, sometimes it goes up and then down. Quite confusing.

If you disable inline media, the chat view stays in the correct place.

MaxLeiter:
Can you try this on master? See the README for instructions on setup

iiiGerardoiii:
@MaxLeiter yup, still happens on 3.0.0-pre.6 with Firefox. Haven't tried on chrome though.

xPaw:
Okay yeah I can somewhat reproduce this in Firefox.

I have no idea how we could fix it though.

xPaw:
@iiiGerardoiii Do you still see this bug on latest Firefox?

iiiGerardoiii:
yup, still present

xPaw:
@iiiGerardoiii Can you test v4.0 pre release and see if it still happens?

iiiGerardoiii:
Apologies for late response.
I'm on v4.1.0 and it doesn't happen anymore 👍

iiiGerardoiii:
It's present on v4.4.3.

PR #3046, opened 2019-02-12
Remove margin-top: auto

xPaw:
This CSS is used to push messages to the bottom of the chat if there aren't enough messages to fill the container, however this property is causing a lot of mysterious scroll related issues in Chrome both on master and Vue branches.

In it's current state, it's causing way more problems with scroll than it solves.

This property is also the real cause behind the #2975 bug (while that was incorrect css on our part, it still shouldn't have scrolled). And this is causing problems in @MiniDigger's theme where he changes opacity in user list.

Do we have a better way of pushing messages to the bottom if we want to keep this behavior?

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1106,10 +1106,6 @@ background on hover (unless active) */
 	touch-action: pan-y;
 }
 
-#chat .chan:not(.special) .messages {
-	margin-top: auto;
-}
-
 #chat .msg {
 	word-wrap: break-word;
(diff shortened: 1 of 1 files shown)


xPaw:
@iiiGerardoiii can you test this PR in firefox and see if it fixes your problem with #2446?

xPaw:
This is a Chrome regression, so we don't need to merge this.

https://bugs.chromium.org/p/chromium/issues/detail?id=931274#c3

Issue #3009, opened 2019-01-29
Query Message Not Being Sent


infosisio:
IRCCloud and all web clients, when you do /query user "yourmessage" it will send the message after switching to that tab, thelounge doesn't, it just switches you to the tab and then you have to re-enter that message.

xPaw:
This is an easy contribution to do for someone that wants to.

Essentially just need to merge this file https://github.com/thelounge/thelounge/blob/master/src/plugins/inputs/query.js

into this https://github.com/thelounge/thelounge/blob/master/src/plugins/inputs/msg.js

Effectively just wrapping the channel creation into an `if(cmd === "query")` just before sending a message.

infosisio:
@xPaw If it's easy, it would be a good idea for the issue to be sorted out. Even though it may seem insignificant, it's a UX flaw for those who are trying to ditch things like IRCCloud and move to thelounge. I would do it myself if I could.

xPaw:
I gave pointers on how to effectively solve it.

infosisio:
@xPaw I know, I wasn't asking anything from you or anyone specifically, just added that for the record.

infosisio:
If anyone is interested in solving this with a bounty, I am willing to deposit it on bountyissues or issuehunt. I think this is worth getting solved.

infosisio:
@xPaw Are you one of the main project authors? 

What's wrong with doing the merge and improving the functionality of the solution, it seems to be something trivial for you?

infosisio:
@xPaw Would you consider solving it, considering how trivial it is to you and how it would improve usability of the solution?

xPaw:
@infosisio Pinging maintainers every few days is a quick way to get yourself blocked.

I have no interest in fixing this, another user tried to make a PR but closed it without finishing it. This is an open source project and people work on it in their free time.

infosisio:
@xPaw Where did you see me pinging you every few days?

Dominent:
Went through the previous pull request, updated and fixed the issue with running query without msg #3138

Dominent:
#3138

PR #3049, opened 2019-02-14
Merged the query input into msg

itwbaer:
The msg input function now executes the query code before sending a
message if that command is given. This opens the channel and then sends
the message, addressing issue #3009.


src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,63 @@
...
+				type: Msg.Type.ERROR,
+				text: "You cannot open a query window without an argument.",
+			}));
+			return;
+		}
+
+		const query = _.find(network.channels, {name: target});
+
+		if (typeof query !== "undefined") {
+			return;
```
xPaw: This return means if there's a query open already, the message will not be sent.

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,63 @@
...
+const Msg = require("../../models/msg");
+
+exports.commands = ["msg", "say", "query"];
 
 exports.input = function(network, chan, cmd, args) {
-	const target = cmd === "msg" ? args.shift() : chan.name;
+	const target = cmd === "say" ? chan.name : args.shift();
+
+	if (cmd === "query") {
+		if (args.length === 0 || target.length === 0) {
```
xPaw: I think due to the args.shift above, this will break opening query windows without any message. `/query user`.

itwbaer: You were correct, this broke direct message.

itwbaer:
I tried to squash my merges and screwed my branch up (sorry I am new to this). Just going to close it out.

infosisio:
@itwbaer Is there a particular barrier pending?

PR #3058, opened 2019-02-16
Downgrade webpack and regenerate lockfile

xPaw:
1. Webpack@4.29.4 pulls a dependency directly from git (future version should fix that)
2. Add integrity field to all packages in the lock file

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "vue-server-renderer": "2.6.6",
     "vue-template-compiler": "2.6.6",
     "vuedraggable": "2.17.0",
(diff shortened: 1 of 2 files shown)


astorije:
> Webpack@4.29.4 pulls a dependency directly from git (future version should fix that)

We are not ok with that?

xPaw:
> We are not ok with that?

It doesn't work for me, at least on Windows. Which is probably a Yarn bug, but that's beside the point if we can avoid it.

astorije:
Got it

PR #3059, opened 2019-02-16
Set up Prettier for JS and Vue files

astorije:
### This PR differs from #3060 in the way that it goes one step further. That other PR is a lot more straightforward and would more easily get in than this one.

This PR sets up [Prettier](https://prettier.io).
First commit is set up, all other commits have been generated with no manual intervention, so any rebase would be rather easy and straightforward. I figured right after Vue-branch merge would be the least disruptive moment to do that.

**I recommend looking at this (and individual commits) with [`?w=1`](https://github.com/thelounge/thelounge/pull/3059/files?w=1).**

Some changes are less than optimal, but over time it will get better. We switched our entire codebase at work a few months ago using a similar setup, and despite some of us being very skeptical (e.g. me), it has been highly beneficial to us on the long-run.
Favorite benefit: auto-format on save, no need to worry about writing things perfectly anymore. Example:

prettier-example

(Ignore the whitespace indentation, this is on my end, this PR does keep tabs correctly, I don't mean to start a war lol)


.prettierrc.yml
```diff
@@ -0,0 +1,3 @@
+arrowParens: always
+bracketSpacing: false
```
astorije: While we're here, I would _love_ to remove those 2 overrides. I know the first it is going to be a tough sell, but I'm still trying :D

McInkay: What's your reasoning for not always having arrow parens?

astorije: For me it's mainly a readability / unencessity thing, but by no means am I trying to start a debate or anything. I was simply wondering if that was something we were open to change or not, totally okay if not. To be fair, Prettier makes this a lot less worthy of debating anyway (at least at write time), because one can type without arrow parens, and they will be added on save.

What do you think about `bracketSpacing` however? It seems to me that having spacing is a lot more standard, and also I find `{foo: "bar"}` generally harder to read than `{ foo: "bar" }` (of course not in a 6-char example, but when browsing through a bigger codebase).

xPaw:
I dislike a lot of the stylistic choices prettier uses, but if other maintainers prefer it, I won't block and will just learn to deal with it.

.prettierrc.yml
McInkay: So, I'm totally not bothered either way snot bracket spacing. I personally would not write code like that but I don't object to it and I understand why people like the extra spaces. 

However, for the same reason as you like the spacing, I like always having the brackets in arrow functions. 

1. It makes it more visibly obvious where the parameter is
2. It means there's 1 style for all types of arrow functions (rather than 1 parameter ones looking different)
3. It means if we add an extra parameter to the function, we don't have to add in the parens at that point

It just looks nicer imo. 

But yeah, I don't mind the bracket spacing. But this is aside from the conversation about whether we should use prettier. I'm not awake enough for that discussion today. I'll chat about it tomorrow.

richrd:
I'm not sure about the stylistic default choices but in general I have liked prettier. Have been using it on some work projects and what I like most is that I don't need to worry so much about the code style myself. I didn't know that eslint can do the same thing though. I don't really have a strong opinion on this; as long as there is some code style that is enforced all is good.

xPaw:
We do already have code style enforced with eslint. However prettier uses an approach where it rewrites code until it matches max length per line (multi line arrays can turn into single line), and code can be split into multiple lines in very odd ways (especially with strings).

.prettierrc.yml
astorije: > I like always having the brackets in arrow functions.

Another point in favor of parens that I forget is when using TypeScript and explicitly typing the argument (which, granted, is fairly rare):

```ts
[1, 2, 3].map((x: number) => x * 2);
```

> But this is aside from the conversation about whether we should use prettier.

You're right, I shouldn't have brought this up so soon, that was premature, my bad.
Thanks for bringing up good points :)

astorije:
> We do already have code style enforced with eslint.

Looking at our codebase, we have inconsistent styling between JS code contained in our JS files and our Vue files, so that's something that Prettier achieves (or intends to, minus the couple minor bugs on Vue files you noticed).

One of the things I like the most about Prettier is how polyglot it is. Not only does it work the same way for all our file formats, it also goes deeper than that: JS code within Markdown will be styled (and syntax checked!) the same way as in our main codebase for example. <sub>This is also true and very effective for GraphQL code in JS, but that's not for us.</sub>

McInkay:
> Have been using it on some work projects and what I like most is that I don't need to worry so much about the code style myself.

This does seem to be the biggest thing that people like about prettier. However ... the problem with that is it's great if you don't care too much about code styles in general. Whereas, if you are really picky about these things, it can be quite frustrating.

I'm ok with different coding styles than I would specifically choose, but only if there's a reason and we come to a decision as a group. The problem with prettier is it just seems to be "here are the styles" and we just put up with them, really. You know?

xPaw:
Was thinking about it again. How about we just bite the bullet and just move to using prettier, which should be more helpful for new contributors.

However I'd remove the manual changes astorije did here after reformatting and rather get them as separate PRs (to lessen the chance shipping unintentional changes amongst thousands of style changes).

A question though, will our CI complain if prettier wasn't run on code (style differs), e.g. in pull requests or when using the github web editor?

McInkay:
> Was thinking about it again. How about we just bite the bullet and just move to using prettier, which should be more helpful for new contributors.

Yeah, I think I'm fine with it. I don't think that I have an issue with prettier in general. I don't think that I worry that this community will shut down any debate about code style, so yeah. 

> A question though, will our CI complain if prettier wasn't run on code (style differs), e.g. in pull requests or when using the github web editor?

You can tie prettier into eslint, so that it runs prettier after linting, and complains if it's wrong. Pretty sure you can run prettier in a mode that lets you do that as well, rather than just formatting.

PR #3060, opened 2019-02-17
Set up Prettier on HTML, JSON, Markdown, and YAML files

astorije:
This is a the counterpart to #3059 that sets up Prettier on all our files except JS and Vue files, at least for now.

Similar description than the other PR:

> This PR sets up [Prettier](https://prettier.io).
> First commit is set up, all other commits have been generated with no manual intervention, so any rebase would be rather easy and straightforward. I figured right after Vue-branch merge would be the least disruptive moment to do that.


astorije:
#3059 is a lot more polarizing and probably requires work to be good enough (if it ever makes it), but this PR is a lot more straightforward and adds style/syntax checking to files that currently do not have any (except CSS).

I believe it is now ready, can we start by getting this in first? 🙏

.prettierrc.yml
```diff
@@ -0,0 +1,7 @@
+arrowParens: always
```
xPaw: Can you add `printWidth: 100` from the JS pr

astorije: Done!

client/thelounge.webmanifest
```diff
@@ -6,8 +6,7 @@
 	"display": "standalone",
```
astorije: Should this file use spaces as well like our other JSON files, or do we want to keep tabs here? (My guess is the latter, I don't mind either way)

.eslintrc.yml
```diff
@@ -28,26 +27,38 @@ rules:
   eqeqeq: error
   handle-callback-err: error
   indent: [error, tab]
-  key-spacing: [error, {beforeColon: false, afterColon: true}]
-  keyword-spacing: [error, {before: true, after: true}]
+  key-spacing:
+    - error
+    - beforeColon: false
+      afterColon: true
```
McInkay: This looks a little odd not having the `-` before each line

.github/ISSUE_TEMPLATE/Bug_Report.md
```diff
@@ -1,15 +1,13 @@
...
-
 ---
 
 <!-- Have a question? Join #thelounge on freenode -->
 
-* *Node version:* 
-* *Browser version:* 
-* *Device, operating system:* 
-* *The Lounge version:* 
+- _Node version:_
```
McInkay: Hmmmmm.

README.md
```diff
@@ -43,11 +43,11 @@
 
 ## Overview
 
-* **Modern features brought to IRC.** Push notifications, link previews, new message markers, and more bring IRC to the 21st century.
-* **Always connected.** Remains connected to IRC servers while you are offline.
-* **Cross platform.** It doesn't matter what OS you use, it just works wherever Node.js runs.
-* **Responsive interface.** The client works smoothly on every desktop, smartphone and tablet.
-* **Synchronized experience.** Always resume where you left off no matter what device.
+- **Modern features brought to IRC.** Push notifications, link previews, new message markers, and more bring IRC to the 21st century.
```
McInkay: Why is it replacing our bullets with `-`?

client/themes/morning.css
```diff
@@ -51,43 +51,113 @@
...
-#chat.colored-nicks .user.color-24 { color: #fcbbba; }
-#chat.colored-nicks .user.color-25 { color: #eef975; }
-#chat.colored-nicks .user.color-26 { color: #c7ff93; }
-#chat.colored-nicks .user.color-27 { color: #ffade1; }
-#chat.colored-nicks .user.color-28 { color: #98ecf2; }
-#chat.colored-nicks .user.color-29 { color: #91a2f5; }
-#chat.colored-nicks .user.color-30 { color: #b19aea; }
-#chat.colored-nicks .user.color-31 { color: #f2a4eb; }
-#chat.colored-nicks .user.color-32 { color: #85f27d; }
+#chat.colored-nicks .user.color-1 {
```
McInkay: I think this is where I'm not a huge fan of this. I think we can go too far into the "just use sensible defaults" and end up in a situation where something becomes less readable.

In general, we want newlines in this sort of situation, but in this exact situation, it is definitely better in the before. 

I feel like it's not simple enough to just go "Here are the rules and they are implmeneted", you know?

renovate.json
```diff
@@ -1,13 +1,7 @@
...
-  "extends": [
-    "config:base"
-  ],
-  "ignorePaths": [
-    "test/"
-  ],
-  "labels": [
-    "Type: Dependencies"
-  ],
+  "extends": ["config:base"],
```
McInkay: I don't understand the logic that prettier is using here. This is the complete opposite of the css issue.

.editorconfig
```diff
@@ -16,6 +16,6 @@ insert_final_newline = true
 [*.md]
 trim_trailing_whitespace = false
 
-[*.{json,yml}]
+[*.{json,md,yml}]
 indent_style = space
```
McInkay: Wait, why are we indenting these with spaces? Sidenote, I know.

However, adding md into this means that markdown and html are inconsistent which annoys me.

xPaw:
@McInkay I've somewhat came to the same conclusion in the JS pr https://github.com/thelounge/thelounge/pull/3059#issuecomment-464426358 https://github.com/thelounge/thelounge/pull/3059#issuecomment-464453373

.eslintrc.yml
astorije: It's not stylistic, it's how (multiline) lists of dictionaries are done in YAML like [here](https://en.wikipedia.org/wiki/YAML#Example). Adding a `-` would be incorrect and ESLint would choke on it.
We already do this with `space-before-function-paren` for instance:

```yaml
  space-before-function-paren:
    - error
    - anonymous: never
      named: never
      asyncArrow: always # Otherwise requires `async()`
```

.github/ISSUE_TEMPLATE/Bug_Report.md
astorije: Yeah, that's definitely an opinion, but even though `*` can be used for italics (resp. `__` for bold), it's somewhat common to prefer `_` instead (resp. `**` for bold). Example, hitting <kbd>Ctrl</kbd>+<kbd>B</kbd> on a GitHub comment uses `**` and hitting <kbd>Ctrl</kbd>+<kbd>I</kbd> uses `_`.

We're using that format _almost_ everywhere anyway (website included), so that's nice to stick to a single syntax.

README.md
astorije: Because that's also fairly accepted to use `-` for bullets rather than `*` in Markdown (note that the outcome in generated HTML is the same, HTML `li`s). One reason is that `*` is also used for bold (and sometimes italics), and that becomes confusing to read: `* **` is a bullet followed by bold (like it was here), `***` is italics+bold`, etc.

I just noticed that GitHub also uses `-` when using the "Add a bulleted list" button in the comment input.

Also this was the only place in Markdown files in this repo where we're using `*` (and we don't on the website), so that's not so disruptive.

.editorconfig
astorije: > Wait, why are we indenting these with spaces? Sidenote, I know.

I don't know, JSON/YAML files always have here. I'm not opposed to changing that (already established that I don't care about tabs vs. spaces as long as some tool helps me not caring about that), but we probably don't want to do it here.

> However, adding md into this means that markdown and html are inconsistent which annoys me.

I agree with you, but I'm not adding `.md` here to change anything, rather to reflect reality. All our Markdown files have spaces, and not doing this change causes lists in Markdown to be indented with 4 spaces, as such:

```md
-   This is the first item of the list
    Second line of the first item
-   This is the second item of the list
```

Yep, `-<3 spaces>Text`... 😞

client/themes/morning.css
astorije: I agree with you here, it's not ideal.
Some of the things it formatted around FontAwesome icon definitions is kind of a mess before, but I agree that for those long lists of colors like these ones, we're better off without those changes.

There is no way to ignore portions of files (and even then, it's not rule-based so we'd ignore all of it), and overall I find it not super useful on our CSS files which are pretty decent already. I'll remove CSS formatting from the PR.

renovate.json
astorije: Formatting tries to be consistent across languages when possible, but some languages are very different and can't have 1:1 matching, like CSS vs. <anything>.

In general, they try to fit on one line if it's not too long (for arrays, because for objects that doesn't apply if you've intentionally broken it up into multiple lines), but with CSS, it's always 1 statement per line.

.editorconfig
astorije: > Wait, why are we indenting these with spaces? Sidenote, I know.

I just remembered one of the key reasons is that npm rewrites `package.json` on dependency install/update, and they do so with 2 spaces. However, it seems like `yarn` respects tabs correctly as opposed to `npm`. Good to know.

Also I just learned that tabs are forbidden in YAML......

So we could theoretically use tabs for JSON, but not for YAML (e.g. Prettier still uses spaces even without this override). However, we'd have to remove `indent_size = 4` from the file for Markdown, but as far as I know tabs are not a thing in Markdown (e.g. Prettier didn't rewrite everything with tabs).

I'm personally happy to remove as many overrides as we can, but it would make the initial diff (and git blame) a little annoying, which is not such a big deal on JSON files IMO. Do you want me to add a commit without this `space` override to see how that looks to you? 🙂

.eslintrc.yml
McInkay: Oh I see, it's not a list of 3 items, it's a list of 2 items, one item is then an object containing 2 items. That's much less obvious in the new syntax, whereas the old syntax made that quite clear. Can we not wrap the object with the curly braces like we do in the before?

astorije: We can, but despite the `-`, I find:

```yaml
key-spacing:
    - error
    - beforeColon: false
      afterColon: true
```

less confusing than:

```yaml
key-spacing:
  - error
  - {beforeColon: false, afterColon: true}
```

Also, I notice that editors have a harder time coloring the second line, e.g. GitHub's color syntax (Atom as well, I don't know about VS Code).
Up to you though, I don't think this is really important so I'm happy to go either way 😅

McInkay:
Also you need to rebase this and sort out the conflicts. I suspect that'll be easy as it's all automatic stuff.

xPaw:
Actually before merge, as I mentioned in the JS/Vue PR, some script needs to be setup on CI so builds fail on files that did not have prettier run on.

https://prettier.io/docs/en/cli.html#list-different ?

McInkay:
As @astorije is tied up just now, I'll pick this up. It'll be good to get this done.

astorije:
> Removed the specifying of which files _to_ format, and added all the ones we want to ignore into the `.prettierignore` file instead.

Sorry to show up after the party. The reason I had done that was to "prepare" for the future, see https://github.com/prettier/prettier/issues/3503:

> - https://github.com/prettier/prettier/issues/2691 - CLI defaults to "**/*.{js,ts,...}".
>   People should just be able to do prettier --write without passing messy globs #0CJS

So when Prettier v2 lands (which I thought would have been much sooner), there won't be a need for either your glob or mine, nor an endless list in `prettierignore`! In the meantime, both solutions work, and I don't think there is an obvious winning one (in a work setup where we have a single Prettier config shared across many projects, the "big glob" solution makes it more reusable, but that's not a concern here, at least not right now).

Thanks a ton for this! 😊

xPaw:
We can review it when v2 lands. We settled on prettierignore because other tooling that calls into prettier will use it.

PR #3062, opened 2019-02-18
Remove network name from window title

astorije:
Follow-up of https://github.com/thelounge/thelounge/pull/2647.

It can be redundant on network tab.

This follows a comment I made in the initial Vue PR: https://github.com/thelounge/thelounge/pull/2647#issuecomment-463914534
I went with no network at all because I don't think it brings any particular value, see comment for more details. I am not married to the idea of entirely removing network name if it's something we think is really needed though, so let me know.

### DIFF
--- a/client/js/utils.js
+++ b/client/js/utils.js
@@ -85,7 +85,7 @@ function updateTitle() {
 	let title = vueApp.appName;
 
 	if (vueApp.activeChannel) {
-		title = `${vueApp.activeChannel.channel.name} — ${vueApp.activeChannel.network.name} — ${title}`;
+		title = `${vueApp.activeChannel.channel.name} — ${title}`;
 	}
 
 	// add highlight count to title


client/js/utils.js
```diff
@@ -85,7 +85,7 @@ function updateTitle() {
 	let title = vueApp.appName;
 
 	if (vueApp.activeChannel) {
-		title = `${vueApp.activeChannel.channel.name} — ${vueApp.activeChannel.network.name} — ${title}`;
+		title = `${vueApp.activeChannel.channel.name} — ${title}`;
```
astorije: @xPaw, do you know if we have access to the current window (not channel), at this point in the code?
I wanted to see if I could add special windows here, like `Settings - The Lounge` but I don't believe we have it yet, in which case it would have to wait for better window management.

xPaw: Windows are not ported yet, so no.

PR #3063, opened 2019-02-18
Improve scroll-down button UI

astorije:
Follow-up of https://github.com/thelounge/thelounge/pull/2647.

Addresses a comment made by @williamboman at https://github.com/thelounge/thelounge/pull/2647#issuecomment-462155721.

Theme | Before | After
--- | --- | ---
Default (normal) | screen shot 2019-02-18 at 01 03 06 | screen shot 2019-02-18 at 13 52 51
Default (hover) | screen shot 2019-02-18 at 01 03 20 | screen shot 2019-02-18 at 13 52 56
Morning (normal) | screen shot 2019-02-18 at 01 10 21 | screen shot 2019-02-18 at 01 10 48
Morning (hover) | screen shot 2019-02-18 at 01 10 31 | screen shot 2019-02-18 at 13 26 54

**With the tooltip on hover:**

screen shot 2019-02-18 at 01 26 26

### DIFF
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -58,7 +58,8 @@
 					v-else
 					class="chat-content">
 					<div
-						:class="['scroll-down', {'scroll-down-shown': !channel.scrolledToBottom}]"
+						:class="['scroll-down tooltipped tooltipped-w', {'scroll-down-shown': !channel.scrolledToBottom}]"
+						aria-label="Jump to recent messages"
 						@click="$refs.messageList.jumpToBottom()">
 						<div class="scroll-down-arrow" />
 					</div>
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1097,7 +1097,7 @@ background on hover (unless active) */
 	pointer-events: none;
 	opacity: 0;
 	transform: translateY(16px);
-	transition: all 0.2s ease-in-out;
+	transition: transform 0.2s, opacity 0.2s;
 	cursor: pointer;
 }
 
@@ -1108,18 +1108,23 @@ background on hover (unless active) */
 }
 
 .scroll-down-arrow {
-	width: 48px;
-	height: 48px;
-	line-height: 48px;
+	width: 36px;
+	height: 36px;
+	line-height: 34px;
 	border-radius: 50%;
-	background: #fff;
-	color: #333;
-	border: 1px solid #84ce88;
+	background: var(--window-bg-color);
+	color: #84ce88;
+	border: 2px solid #84ce88;
 	text-align: center;
+	transition: background 0.2s, color 0.2s;
+	box-shadow:
+		0 3px 5px -1px rgba(0, 0, 0, 0.2),
+		0 6px 10px 0 rgba(0, 0, 0, 0.14);
 }
 
 .scroll-down:hover .scroll-down-arrow {
 	background: #84ce88;
+	color: var(--window-bg-color);
 }
 
 .scroll-down-arrow::after {
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1121,9 +1121,7 @@ background on hover (unless active) */
 	border: 2px solid var(--button-color);
 	text-align: center;
 	transition: background 0.2s, color 0.2s;
-	box-shadow:
-		0 3px 5px -1px rgba(0, 0, 0, 0.2),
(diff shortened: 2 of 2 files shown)


client/components/Chat.vue
```diff
@@ -58,8 +58,10 @@
 					v-else
 					class="chat-content">
 					<div
-						:class="['scroll-down', {'scroll-down-shown': !channel.scrolledToBottom}]"
-						@click="$refs.messageList.jumpToBottom()">
+						:class="['scroll-down tooltipped tooltipped-w', {'scroll-down-shown': !channel.scrolledToBottom}]"
+						aria-label="Jump to recent messages"
```
astorije: I started to "Scroll to bottom" which I didn't like (too cold/formal, also not a "scroll" per se) then shamelessly stole the wording from Slack, which I found more accurate and friendly.

client/css/style.css
```diff
@@ -1097,7 +1097,7 @@ background on hover (unless active) */
 	pointer-events: none;
 	opacity: 0;
 	transform: translateY(16px);
-	transition: all 0.2s ease-in-out;
+	transition: transform 0.2s, opacity 0.2s;
```
astorije: `all` causes a transition on `right` when toggling the user list, so I went explicit here.
Also went with the same transition as we use everywhere else 🤷‍♂️

client/css/style.css
```diff
@@ -1108,18 +1108,23 @@ background on hover (unless active) */
...
-	height: 48px;
-	line-height: 48px;
+	width: 36px;
+	height: 36px;
+	line-height: 34px;
 	border-radius: 50%;
-	background: #fff;
-	color: #333;
-	border: 1px solid #84ce88;
+	background: var(--window-bg-color);
```
astorije: This and the same change at R1127 is the bit of magic that makes it work for Morning, and likely for other themes as well 👌

client/css/style.css
```diff
@@ -1108,18 +1108,23 @@ background on hover (unless active) */
...
+	color: #84ce88;
+	border: 2px solid #84ce88;
 	text-align: center;
+	transition: background 0.2s, color 0.2s;
+	box-shadow:
+		0 3px 5px -1px rgba(0, 0, 0, 0.2),
+		0 6px 10px 0 rgba(0, 0, 0, 0.14);
```
astorije: Stole this from Material-UI for floating buttons. They have 3 box-shadows, I removed the last one which made the button shadow too thick and present compared to the rest of our UI, but re-used the other 2 shadows as-is. I tried to remove one then the other, it doesn't look as nice imo.

client/css/style.css
```diff
@@ -1108,18 +1108,23 @@ background on hover (unless active) */
...
-	line-height: 48px;
+	width: 36px;
+	height: 36px;
+	line-height: 34px;
 	border-radius: 50%;
-	background: #fff;
-	color: #333;
-	border: 1px solid #84ce88;
+	background: var(--window-bg-color);
+	color: #84ce88;
```
xPaw: Do we want to make this color a css var?

astorije: I can do that.

xPaw: I'm just thinking if we should make it a generic one (we use it a lot elsewhere it seems) or specifically for this button.

astorije: I don't think a variable for a given color makes it any useful. Let me come up with something and you tell me if you like it or not.

xPaw:
The increased border with shadow kinda makes the button stand out like a sore thumb on the white theme.  Perhaps reducing the shadow a little might help?

client/css/style.css
astorije: Done. Went a simple route, and if we ever need more granularity (different variables for border, text color, etc.), we can add them, but I think for now that's enough and provides some customization level already.

astorije:
> The increased border with shadow kinda makes the button stand out like a sore thumb on the white theme. Perhaps reducing the shadow a little might help?

I don't mind that it makes it stand out, especially since I reduce the size a little, and because having it stand out is kind of the point, I think. Also I use the white theme and that wasn't shocking to me.
But I get what you're saying, so I removed one of the 2 shadows to make it stand out less and I think it looks nice like that:

screen shot 2019-02-18 at 13 58 12

EDIT: Ah, you already approved :)

PR #3073, opened 2019-02-19
extend custom highlight regex

Jay2k1:
As discussed and tested on irc, this extends matching for custom highlights. Main reason was that a word defined in custom highlights followed by e.g. a colon, a comma or an exclamation mark, such as `Jay: hi!` or `Hi Jay!`, wouldn't be caught by the regex.

Also, it adds a test for custom highlights.

### DIFF
--- a/test/tests/customhighlights.js
+++ b/test/tests/customhighlights.js
@@ -1,0 +1,77 @@
+"use strict";
+
+const expect = require("chai").expect;
+
+const Client = require("../../src/client");
+
+const client = new Client({}, "test", {clientSettings: {highlights: "foo, @all,   sp ace   , 고"}});
+
+describe("Custom highlights", function() {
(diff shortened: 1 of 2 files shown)


test/tests/customhighlights.js
```diff
@@ -0,0 +1,77 @@
+"use strict";
+
+const expect = require("chai").expect;
+
+const Client = require("../../src/client");
+
+const client = new Client({}, "test", {clientSettings: {highlights: "foo, @all,   sp ace   , 고"}});
```
Zarthus: i'd put it on a newline so it's very clear what you're testing against (maybe even `[..., ..., ...].join(',')`)

fnutt:
For reference, https://github.com/thelounge/thelounge/pull/2990

PR #3079, opened 2019-02-22
Put channel errors and unhandled numerics to relevant channel if it exists

xPaw:


xPaw:
@Zarthus You'll probably want to test this one out.

xPaw:
@astorije Correct, if it detects a channel to put the message in, it disables 'show in active'.

PR #3082, opened 2019-02-25
Display a broken link icon and red text color when not joined on a channel

### LINKED ISSUE #3080: Add styling for parted channels
They have a class, just need to add actual styling for it.

Change icon? Make them faded or italic?

astorije:
Closes #3080.

screen shot 2019-02-25 at 00 21 59

### DIFF
--- a/client/components/Channel.vue
+++ b/client/components/Channel.vue
@@ -10,6 +10,7 @@
 			class="badge">{{ channel.unread | roundBadgeNumber }}</span>
 		<template v-if="channel.type === 'channel'">
 			<span
+				v-if="channel.state === 0"
 				class="parted-channel-tooltip tooltipped tooltipped-w"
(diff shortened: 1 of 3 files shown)


client/components/Channel.vue
```diff
@@ -9,6 +9,11 @@
 			:class="{ highlight: channel.highlight }"
 			class="badge">{{ channel.unread | roundBadgeNumber }}</span>
 		<template v-if="channel.type === 'channel'">
+			<span
```
xPaw: Hide this with v-if

astorije: I did a thing, can you check if that's what you meant? 🙏

xPaw: Yep. We'll want to do this with other icons too (the ones you moved styles for).

PR #3083, opened 2019-02-25
Move closing brackets on a new line in Vue files

astorije:
### Advantages

- This is the default behavior of the rule. That means one less override (minus the fact that the plugin reports all non-base/essential rules as violations), which is always neat.
- Arguably easier to read. As I was working on #3082, I was browsing through existing code, and several times I missed a nested `span` because of that. Put in an example, in a mid-size file with a several elements and indentations, I find that it's easier to get confused by:
  ```vue
  <span
    foo="foo"
    bar="bar">
    <span baz="baz" />
  </span>
  ```
  than by:
  ```html
  <span
    foo="foo"
    bar="bar"
  >
    <span baz="baz" />
  </span>
  ```
- An argument was made against Prettier that it makes diffs harder because it reorganizes existing lines. Having `>` on the same line as an attribute has the same effect: adding an attribute at the end of a list, or re-organizing attributes (that include the last one) adds to the diffs.
- This puts last attribute and content on 2 different lines, which also improves diffing (and makes things more readable overall). Before, we would have:
  ```vue
  <span
    someAttribute="bar">{{ someContent }}</span>
  ```
  and we now get:
  ```vue
  <span
    someAttribute="bar"
  >{{ someContent }}</span>
  ```

### Drawbacks

- It makes files longer, as we can tell by the fact that this PR adds more lines than it removes.
- There is one less level of indentation on nested elements that are not on their own line.
  Before that PR:
  ```vue
  <span
    foo="foo"><button
      type="button" /></span>
  ```
  After that PR:
  ```vue
  <span
    foo="foo"
  ><button
    type="button"
  /></span>
  ```
  Note that elements that live on their own line do not have this (we have a mix of both for, say, `span`s and `button`s in our codebase):
  ```vue
  <span
    foo="foo"
  >
    <button
      type="button"
    />
  </span>
  ```

---

Despite those drawbacks, I find the code a lot more readable to read through with this change. It's also reasonable to do this now as this is very new code.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -92,7 +92,7 @@ rules:
   vue/require-default-prop: off
   vue/no-v-html: off
   vue/no-use-v-if-with-v-for: off
-  vue/html-closing-bracket-newline: [error, {singleline: never, multiline: never}]
+  vue/html-closing-bracket-newline: error
   vue/multiline-html-element-content-newline: off
   vue/singleline-html-element-content-newline: off
 
--- a/client/components/MessageCondensed.vue
+++ b/client/components/MessageCondensed.vue
@@ -5,15 +5,18 @@
 			<span class="from" />
 			<span
 				class="content"
-				@click="onCollapseClick">{{ condensedText }}<button
-					class="toggle-button"
-					aria-label="Toggle status messages" /></span>
+				@click="onCollapseClick"
+			>{{ condensedText }}<button
+				class="toggle-button"
+				aria-label="Toggle status messages"
+			/></span>
 		</div>
 		<Message
 			v-for="message in messages"
 			:key="message.id"
 			:network="network"
-			:message="message" />
+			:message="message"
+		/>
 	</div>
 </template>
 
--- a/client/components/App.vue
+++ b/client/components/App.vue
@@ -1,60 +1,72 @@
 <template>
 	<div
 		id="viewport"
-		role="tablist">
+		role="tablist"
+	>
 		<aside id="sidebar">
 			<div class="scrollable-area">
 				<div class="logo-container">
 					<img
 						:src="`img/logo-${isPublic() ? 'horizontal-' : ''}transparent-bg.svg`"
 						class="logo"
-						alt="The Lounge">
+						alt="The Lounge"
+					>
 					<img
 						:src="`img/logo-${isPublic() ? 'horizontal-' : ''}transparent-bg-inverted.svg`"
 						class="logo-inverted"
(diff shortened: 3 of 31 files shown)


.eslintrc.yml
```diff
@@ -92,7 +92,7 @@ rules:
   vue/require-default-prop: off
   vue/no-v-html: off
   vue/no-use-v-if-with-v-for: off
-  vue/html-closing-bracket-newline: [error, {singleline: never, multiline: never}]
+  vue/html-closing-bracket-newline: error
```
astorije: Without this, a violation is reported as a warning, which wouldn't fail our CI. The issue is other non-essential rules from this plugin are reported as warnings.

What about removing this instead, and passing a `--max-warnings=0` flag to ESLint calls, to make sure we never get bit by this?
I'm ready to bet we explicitly enabled some other rules in this file (despite using the recommended preset) for this reason too, so we could probably clean it up a little.

client/components/MessageCondensed.vue
```diff
@@ -5,15 +5,18 @@
...
 			<span
 				class="content"
-				@click="onCollapseClick">{{ condensedText }}<button
-					class="toggle-button"
-					aria-label="Toggle status messages" /></span>
+				@click="onCollapseClick"
+			>{{ condensedText }}<button
+				class="toggle-button"
+				aria-label="Toggle status messages"
+			/></span>
```
Zarthus: I personally think you get in indentation hell if you allow structures like this. (which is evidenced by the alignment of the buttons' attributes being misindented.)

Maybe the span should end on a newline?

current:
```
			<span
				class="content"
				@click="onCollapseClick"
			>{{ condensedText }}<button
				class="toggle-button"
				aria-label="Toggle status messages"
			/></span>
```

expected:
```
			<span
				class="content"
				@click="onCollapseClick"
			>{{ condensedText }}<button
					class="toggle-button"
					aria-label="Toggle status messages"
				/>
			</span>
```

astorije: Makes sense, but keep in mind that I didn't make any manual changes, just ran ESLint with `--fix`.
Technically, `<span><button>` and `<span>[newline]</button>` are not the same thing, which is why I believe ESLint went that way. In fact, if you look at our current codebase, we do have both, such as [here](https://github.com/thelounge/thelounge/pull/3083/files#diff-7d16de2e6b8e5d9dd5451e842b5d9034R18), which I think is more in line with what you are suggesting, if I'm not mistaken.

PR #3107, opened 2019-03-07
Send SETNAME command if user edits realname field

xPaw:
Implements command part of https://github.com/ircv3/ircv3-specifications/pull/361

The capability is not relevant as we don't track realnames of other users yet.


jwheare:
I mean, you could use the cap to determine whether you can use that command without errors.

xPaw:
That's not a terrible idea, will do!

PR #3111, opened 2019-03-07
Use async/await in service worker, do not wait for cache for successful requests

xPaw:
Use async instead of promises, browsers that support service workers also support async already.

I also changed the cache logic to avoid blocking the request response. This means fetch() and its response are returned immediately if request succeeds and refreshing the content in cache is done in a separate async call.

Before:
1. Open cache
2. Fetch request
3. Put into cache
4. Return response

After:
1. Fetch request
2. Return response
3. Open cache
4. Put into cache


xPaw:
I also implemented cache busting.

PR #3114, opened 2019-03-12
Change styling when dragging channels.

### LINKED ISSUE #3097: Channel sorting/dragging needs polishing
With the switch to a new dragging library, it needs to be polished up.

xPaw:
- Fixes #3097.
- Replace deprecated `:options` with attributes
- ~Change drop placeholder with a green (button color) line of where the channel will end up, which I think reflects it better due to changed ghost placeholder behaviour.~

Upload overlay appearing when dragging is fixed in #3037.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -582,6 +582,9 @@ kbd {
 }
 
 #sidebar .chan {
+	display: flex;
+	padding: 8px 14px;
+	position: relative;
 	cursor: pointer;
 }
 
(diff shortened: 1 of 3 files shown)


astorije:
Re: UI, I think the v3.0.1 behavior was a lot better compared to what we have now (both `master` and your PR):

v3.0.1 | `master` | #3114
--- | --- | ---
 | master | pr

Of the 3 solutions, the first one is my preferred by far. The green line is confusing me, I much prefer a placeholder area, that is a lot more expected and predictable IMO.
Can we make the content of the target transparent and fix the margins so we get the initial look-and-feel back? Happy to help if you're open to it.

xPaw:
You can try if you want.

richrd:
Now that I see the 3 behaviors side by side, I too prefer the first one. If that can be made happen then great :)

xPaw:
Okay actually, I think I managed to it. I show an overlay with dashed border in `::after` so it doesn't change the real dimensions.

I also fixed up stuff around removing background colors on active channels, I think it should be good now.

client/css/style.css
```diff
@@ -606,19 +609,30 @@ background on hover (unless active) */
...
 /* Remove background on hovered/active channel when sorting/drag-and-dropping */
-#sidebar .ui-sortable-helper .chan.active, /* Networks */
-#sidebar .ui-sortable-helper .chan:hover,
-#sidebar .chan.ui-sortable-helper.active, /* Channels */
-#sidebar .chan.ui-sortable-helper:hover {
-	background-color: transparent;
+.ui-sortable-dragged,
+.ui-sortable-dragged .chan,
+#sidebar .ui-sortable-active .chan {
+	background: transparent !important;
```
astorije: Whoa, the big guns :)
Is there a way to make the selector more specific to avoid `!important` at all?

xPaw:
It better be more smooth as it's using native drag and drop 🙃

client/css/style.css
xPaw: Fixed up.

PR #3137, opened 2019-03-21
Force no-cache on service-worker and sourcemap files

xPaw:
These files are not cache busted, so we should force proxies (like cloudflare) to not cache them.


McInkay:
> as long as a reverse proxy such as nginx does not overwrite the headers.

Yeah, we can't really deal with that, because any reverse proxy could theoretically remove any of our headers and mess up everything.

PR #3138, opened 2019-03-22
Move query to msg, fix #3049

### LINKED ISSUE #3009: Query Message Not Being Sent

Dominent:
See: https://github.com/thelounge/thelounge/pull/3049

Fixes #3009


src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,73 @@
...
+				type: Msg.Type.ERROR,
+				text: "You cannot open a query window without an argument.",
+			}));
+			return;
+		}
+
+		const query = _.find(network.channels, {name: target});
+
+		if (typeof query !== "undefined") {
+			return;
```
xPaw: I left the same comment in previous PR:

> This return means if there's a query open already, the message will not be sent.

Dominent: oh sorry about that I didn't get it. The correct behavior will be to send the message right  ?

xPaw: Yes. It will be very odd otherwise where first /query will open and send a message, but the next one will do nothing.

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,73 @@
...
+	if (cmd === "query") {
+		if (target.length === 0) {
+			chan.pushMessage(this, new Msg({
+				type: Msg.Type.ERROR,
+				text: "You cannot open a query window without an argument.",
+			}));
+			return;
+		}
+
+		const query = _.find(network.channels, {name: target});
```
xPaw:
```suggestion
		const query = network.getChannel(target);
```

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,71 @@
...
+				type: Msg.Type.ERROR,
+				text: "You cannot open a query window without an argument.",
+			}));
+			return;
+		}
+
+		const query = network.getChannel(target);
+		const isQueryOpened = (typeof query) !== "undefined";
+
+		if (!isQueryOpened) {
```
xPaw:
```suggestion
		if (typeof query === "undefined") {
```

We never do it the other way in our codebase.

Dominent: The Idea was to make it more readable

xPaw: It ends up being less readable though. It's being negated twice.

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,71 @@
...
+		return chan.name;
+	}
+}
 
 exports.input = function(network, chan, cmd, args) {
-	const target = cmd === "msg" ? args.shift() : chan.name;
+	const target = parseInput(cmd, args, chan);
+
+	if (cmd === "query") {
+		if (target.length === 0) {
```
xPaw: Executing `/query` crashes:
> Cannot read property 'length' of undefined

Dominent: Forgot about that use case, sorry.

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,71 @@
...
+		return chan.name;
+	}
+}
 
 exports.input = function(network, chan, cmd, args) {
-	const target = cmd === "msg" ? args.shift() : chan.name;
+	const target = parseInput(cmd, args, chan);
+
+	if (cmd === "query") {
+		if (typeof target === "undefined" || target.length === 0) {
```
xPaw:
```suggestion
		if (!target) {
```

This would cover both of these cases, wouldn't it? It's done this way below for the `/msg`

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,71 @@
...
+			chan.pushMessage(this, new Msg({
+				type: Msg.Type.ERROR,
+				text: "You cannot open a query window without an argument.",
+			}));
+			return;
+		}
+
+		const query = network.getChannel(target);
+		const isQueryOpened = (typeof query) !== "undefined";
```
xPaw: this is unused var now

src/plugins/inputs/msg.js
Dominent: Correct but would have the issue with isQueryOpened that we discussed previously ?

xPaw: What issue?

Dominent: misunderstood, convention you are correct and it way cleaner. good catch

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,70 @@
...
+	case "query":
+		return args.shift();
+	default:
+		return chan.name;
+	}
+}
 
 exports.input = function(network, chan, cmd, args) {
-	const target = cmd === "msg" ? args.shift() : chan.name;
+	const target = parseInput(cmd, args, chan);
```
McInkay: Pretty picky point here, but maybe this should be `getTarget` rather than `parseInput`, just because it's not parsing all the input, it's just getting the target.

src/plugins/inputs/msg.js
```diff
@@ -1,9 +1,70 @@
...
+	if (cmd === "query") {
+		if (!target) {
+			chan.pushMessage(this, new Msg({
+				type: Msg.Type.ERROR,
+				text: "You cannot open a query window without an argument.",
+			}));
+			return;
+		}
+
+		const query = network.getChannel(target);
```
McInkay: Hmmm, maybe we should have `targetName` instead of target and `target` instead of query?

xPaw: Since `query` is not used anywhere besides the `if` check, you could name it `existingChannel` to avoid even more confusion.

Or just move the `getChannel` call directly into the `if`.

McInkay: Oh yeah of course, because it's not actually the target we are trying to get. Yeah, I think moving the call into the if would be more clear. `if(network.getChannel(targetName))` is pretty obvious what it's trying to check.

McInkay:
@Dominent would you have time sometime soon just to fix the few small wording issues with the PR? If not, let us know and we can sort that out ourselves. Would be great to get this in.

xPaw:
@Dominent or @McInkay can you fix this up so it can be merged?

McInkay:
Squashed the commits and fixed the couple of small naming issues. 

@xPaw feel free to merge if you are happy.

@Dominent I hope you are fine with me just doing this myself, you'll still get credit (you are co-author on the commit), as you hadn't replied for nearly 2 months,

PR #3145, opened 2019-03-25
Change condensed summary hover to an underline

xPaw:
This makes it more consistent with other elements like links and the more button in link previews.

Also avoids a dumb bug in Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=781566


xPaw:
Yeah underline might not be the best here, but I'd argue it works better than the opacity change.

PR #3176, opened 2019-04-14
Implement invite list

xPaw:
I changed banlist into a generic list handler, which will hopefully also work fine for other lists (quiet, exceptions).

However these other lists need to be implemented in irc-framework first.

View with `/mode +I`


MaxLeiter:
Do we want to rename banlist.js to something more ambigious, like speciallists.js? (I cant think of a better name right now, but one probably exists)

McInkay:
How about just `lists.js`? :-P

xPaw:
We already have `list.js` so that will be confusing. `modelists`?

src/plugins/irc-events/banlist.js
```diff
@@ -6,19 +6,36 @@ const Msg = require("../../models/msg");
...
+			hostmask: invite.invited,
+			invited_by: invite.invited_by,
+			invited_at: invite.invited_at * 1000,
+		}));
 
-		if (!bans || bans.length === 0) {
+		handleList(Chan.SpecialType.INVITELIST, "Invite list", list.channel, data);
+	});
+
+	function handleList(type, name, channel, data) {
```
McInkay: Do we need type *and* name? We could remove name, I think, and just reformat the type for the UI. They are a consistent format, after all.

xPaw: Ideally it would be entirely in the UI yeah, but we don't have a translation framework in place yet.

There's also the error message saying X list is empty.

McInkay: Yeah but we can just take off the end of the type and use that for the error message.

McInkay:
Oh yeah.

I guess we could have `list.js` renamed to `chanlist.js` then this one could be `userlist.js`? The key point is, they are both lists, but one is asking for a list of channels, and this is all lists of users, just in different ways.

I guess technically they are all lists of modes, but that feels kinda technical. But if we can't think of anything better, then sure.

xPaw:
> I guess technically they are all lists of modes, but that feels kinda technical.

Well it is a technical implementation :)

McInkay:
Sure, but I guess what I mean is it feels *too* technical. Like, at this level we don't care about the underlying "this is what means that people are on the list", we care about "what people are on X list", right?

src/plugins/irc-events/banlist.js
xPaw: What do you mean?

xPaw:
Merging, we can always improve this in further PRs.

I went with `modelist` for the file name as `userlist` can be easily confused for multiple other things. I'd rather have the file name be closely related to what it does so it's easier to find.

PR #3179, opened 2019-04-17
Replace `request` with `got`

xPaw:
`request` is basically deprecated by its maintainers: https://github.com/request/request/issues/3142

`got` is basically a lighter version of `request` so hopefully nothing broke by switching the libs.


astorije:
Any particular reasons why you went with `got`? They maintain a list of alternatives and I was wondering if you got `got` (ah!) from here: https://github.com/request/request/issues/3143
I personally like `node-fetch` because it's coming straight from the spec.

xPaw:
No particular reason, got is well maintained and had a migration guide which kinda helped.

xPaw:
@richrd mind giving my link previewer stream/buffering a review?

I tested that code around with console logs, it seems to abort correctly when going over the size limit and whatnot.

PR #3226, opened 2019-05-18
Add paste file functionality

### LINKED ISSUE #3224: Add paste support for clipboard images
Currently TheLounge doesn't support it to paste images from the clipboard. Only drag-and-drop is supported. It would be nice if support could be added for it. It seems that all major browsers support pasting images.

SwayUser:
Closes #3224

### DIFF
--- a/client/js/upload.js
+++ b/client/js/upload.js
@@ -16,6 +16,7 @@ class Uploader {
 		document.addEventListener("dragover", (e) => this.dragOver(e));
 		document.addEventListener("dragleave", (e) => this.dragLeave(e));
 		document.addEventListener("drop", (e) => this.drop(e));
+		document.addEventListener("paste", (e) => this.paste(e));
 	}
 
 	dragOver(event) {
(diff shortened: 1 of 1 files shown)


xPaw:
For reference, here is @richrd's code: https://github.com/thelounge/thelounge/commit/e8d449984875be032ecff02fc10014fa86a52626

It appears to handle any file, not just images.

SwayUser:
@xPaw I changed it to handle any type of file.

xPaw:
I believe `event.preventDefault()` needs to be called if there are any files that will be passed into `triggerUpload`

Ref https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event

SwayUser:
I changed it again; although I think there's no real need to prevent the default action. I think the intention of the sentence in the documentation was more like along the lines of "to prevent the default action from occuring" instead of "to add custom behaviour for pasting events, the default action needs to be stopped." (Or did you mean something different why I'd need to call it?)

xPaw:
We "modify" the paste in the sense that we take the pasted files and upload them, which is why we want to prevent the default action.

SwayUser:
Ah, it might cause issues (if the default action isn't prevented) if there's both images and text in the clipboard. Should there be any additional handling when there's also text?

xPaw:
That indeed sounds like it could potentially bring in some issues.

If you can figure out some extra handling for this case, that would be good.

SwayUser:
OK, after thinking about it, I don't think it's a good idea to paste text also in any way if there's an image. Probably programs who put images and text in the clipboard simultaneously do it to provide an "alternative" to the image, in case the application doesn't support showing an image. If it's intended to both paste an image and text then they would use something like HTML as a target.

xPaw:
Can you fix the build please, there are eslint errors.

SwayUser:
Should be good now.

richrd:
Looks good to me! It doesn't have the `if (!event.clipboardData) {` check that I had in place but after a quick test it doesn't seem to be needed. Approving this :+1:

PR #3276, opened 2019-07-03
Make sure unhandled command has params

xPaw:
We can crash on `network.getChannel(undefined)` call.

### DIFF
--- a/src/plugins/irc-events/unhandled.js
+++ b/src/plugins/irc-events/unhandled.js
@@ -8,16 +8,18 @@ module.exports = function(irc, network) {
 	irc.on("unknown command", function(command) {
 		let target = network.channels[0];
 
-		// Do not display users own name
-		if (command.params[0] === network.irc.user.nick) {
-			command.params.shift();
-		} else {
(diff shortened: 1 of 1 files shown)


src/plugins/irc-events/unhandled.js
```diff
@@ -8,16 +8,18 @@ module.exports = function(irc, network) {
...
-			// put this message in that channel
-			const channel = network.getChannel(command.params[0]);
+		if (command.params.length > 0) {
+			// Do not display users own name
+			if (command.params[0] === network.irc.user.nick) {
+				command.params.shift();
+			} else {
+				// If this numeric starts with a channel name that exists
+				// put this message in that channel
+				const channel = network.getChannel(command.params[0]);
```
McInkay: If we only have 1 element, and it's the user's nick, then we can still come across the same issue, as we remove that from the params list. We should probably check for that as well.

xPaw: We won't, it's in the `else`.

McInkay: Oh, duh.

PR #3281, opened 2019-07-05
Refactor command API for plugins

McInkay:
So, previously we were treating plugin commands as exactly the same as internal ones. That was a bit confusing. We are changing them to have:

1. A new publicClient wrapper around the client, so plugins don't have access to all the internal functions that we want
2. The publicClient as an argument, rather than bound to `this`
3. The whole target object, rather than just the network and channel as separate objects. This one, after having done the work, might not be necessary as I think the target is just an object containing the 2 of them ... but it is nice not having to have the 2 different args. I dunno.

Here's the code for my test plugin up to date with the new API.

```
const shrug = {
  input: function(client, target, command, args) {
    const messageWithShrug = args.join(" ").replace(/^\//, "") + " ¯\\_(ツ)_/¯";
    client.runAsUser(messageWithShrug, target.chan.id);
  }
};

module.exports = {
  onServerStart: thelounge => {
    thelounge.Commands.add("shrug", shrug);
  }
};
```

Client is the publicClient, as can be seen in `publicClient.js`.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -12,6 +12,7 @@ const UAParser = require("ua-parser-js");
 const uuidv4 = require("uuid/v4");
 const escapeRegExp = require("lodash/escapeRegExp");
 const inputs = require("./plugins/inputs");
+const PublicClient = require("./plugins/packages/publicClient");
 
 const MessageStorage = require("./plugins/messageStorage/sqlite");
 const TextFileMessageStorage = require("./plugins/messageStorage/text");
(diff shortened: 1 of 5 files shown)


src/client.js
```diff
@@ -336,20 +337,19 @@ Client.prototype.inputLine = function(data) {
 	const irc = target.network.irc;
 	let connected = irc && irc.connection && irc.connection.connected;
 
-	if (Object.prototype.hasOwnProperty.call(inputs.userInputs, cmd) && typeof inputs.userInputs[cmd].input === "function") {
-		const plugin = inputs.userInputs[cmd];
+	if (inputs.userInputs.get(cmd) && typeof inputs.userInputs.get(cmd).input === "function") {
```
xPaw: You're calling `get` three times (the first one should have been a `has`.

Ideally you can cache `get` call into a variable, and check if it's not undefined.

McInkay: Yeah, I was thinking about that.

McInkay: Ok, I've pulled that out, but I've done it inside the if and used `has` for the if statement. Are you happy with that, or do you want it pulled out above as well? That would feel unnecessary for me, as we'd need to either do both before we know whether we have 1 or the other, or we'd have to split up the if which probably wouldn't look nice.

There's probably a nicer way to remove some of the duplication of this actually, but because they take different arguments, it would be a bit overly complicated, probably.

McInkay:
I have also re-added the `runAsUser` into the public api for packages, so that they can pass throug hthe same arguments, plus the client, if they want to send a command from elsewhere in the application.

PR #3294, opened 2019-07-11
Verify reverse DNS when looking up hostnames for webirc

### LINKED ISSUE #3293: Check forward/backward DNS per IRCv3 spec when passing hostname along with WEBIRC command
As discussed on IRC, consider IP 183.87.215.110. 

```
$ dig 110.215.87.183.in-addr.arpa ptr +short
undefined.hostname.localhost.
```

We should ideally check the A record on the returned PTR to make sure it matches up. This was clarified in the IRCv3 version of the WEBIRC spec, which The Lounge's implementation predates

xPaw:
Fixes #3293. See https://ircv3.net/specs/extensions/webirc.html#format

cc @lol768 for testing

Quick test:
```js
reverseDnsLookup('183.87.215.110', (a) => console.log(a));
reverseDnsLookup('2a00:1450:400f:80b::200e', (a) => console.log(a));
reverseDnsLookup('8.8.8.8', (a) => console.log(a)); // dns.google
reverseDnsLookup('8.8.4.4', (a) => console.log(a)); // dns.google
reverseDnsLookup('2001:4860:4860::8888', (a) => console.log(a)); // dns.google
reverseDnsLookup('2001:4860:4860::8844', (a) => console.log(a)); // dns.google
```

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -744,11 +744,23 @@ function performAuthentication(data) {
 }
 
 function reverseDnsLookup(ip, callback) {
-	dns.reverse(ip, (err, hostnames) => {
-		if (!err && hostnames.length) {
-			return callback(hostnames[0]);
+	dns.reverse(ip, (reverseErr, hostnames) => {
+		if (reverseErr || hostnames.length < 1) {
+			return callback(ip);
 		}
 
-		callback(ip);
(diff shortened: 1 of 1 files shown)


lol768:
Thanks both!

PR #3312, opened 2019-07-17
Set up prettier for JS/Vue files

McInkay:
Well, I mucked up #3059 and github won't let me re-open despite having re-created the branch it was pointing to, so I needed to create a new one.

I've fixed up this PR with rebasing and it should work fine with the JS/Vue files now.


xPaw:
Should we add some explanation of prettier (what/why/how) in contributing.md with some pointers to editor extensions or what command to run if you dont want/have it

McInkay:
Well, by default it runs prettier before committing, so if people don't want to run an extension, they don't need to run any comamnd. Happy to add something though, yeah.

McInkay:
Done, and fixed the conflicts.

Should be good?

Issue #3322, opened 2019-07-21
IDN-URL gets pasted encrypted by file upload


Lucki:
- _Node version:_ 8.10.0
- _Browser version:_ FF 68.0.1
- _Device, operating system:_ CentOS 6
- _The Lounge version:_ 3.1.1

---

While using TheLounge with a [Internationalized domain name URL](https://en.wikipedia.org/wiki/Internationalized_domain_name) and uploading a file the resulting pre-pasted url into the input field contains the url in the encrypted version instead of the human readable version.

Using the example of wikipedia:
- _TheLounge_url:_ `thelounge.bücher.example`
- _Pre-pasted_url:_ `thelounge.xn--bcher-kva.example/uploads/…`
- _expected_url:_ `thelounge.bücher.example/uploads/…`

xPaw:
We use browser APIs to handle the URLs and they deal with punycode. I don't think we can easily (or safely) convert it back to 'human' form. This also helps with compatibility with other IRC clients which may not deal with internationalized domains correctly. As such I will close this issue, as I don't think we'd change this behaviour.

For reference:
```js
new URL("https://thelounge.bücher.example").href
"https://thelounge.xn--bcher-kva.example/"
```

Lucki:
Sad, IMHO the user should always see the readable format. But I can see that it's not easily possible without an [external librarie](https://github.com/bestiejs/punycode.js/). Would be nice to have an internal function like [`decodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI) to make access to them easy.

PR #3325, opened 2019-07-21
Decode punycoded URLs before inserting

Lucki:
It's possible the `url` parameter is punycoded when a file gets uploaded.
This uses `punycode` to decode the `fullURL` into human readable format before it gets inserted into the `textbox`.

Fix bug #3322

### DIFF
--- a/client/js/upload.js
+++ b/client/js/upload.js
@@ -2,6 +2,7 @@
 
 const socket = require("./socket");
 const updateCursor = require("undate").update;
+const punycode = require('punycode');
 
 class Uploader {
 	init() {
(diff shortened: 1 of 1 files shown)


xPaw:
So just to provide some context, we use `URL` API specifically so we don't have to include third party code, and browser APIs can handle the urls correctly for us. Adding a punycode converter here goes against that.

Using a punycode converter on top of that will yield its own issues like this: https://github.com/bestiejs/punycode.js/issues/46

And if you want to support UTS46 correctly, you will have to include ridicilous maps like this: https://github.com/jcranmer/idna-uts46/blob/master/idna-map.js

As such, I still think taking the hit and simply relying on the browser API is the best way to go here. @McInkay suggested that plugins would be able to override upload urls (this was discussed some time ago), for things like CDNs, and I think this issue would fall within that.

Lucki:
You're right, already thought this will get rejected. If it's possible later on through a plugin that would also be fine.
As I already said in the linked Issue there should ideally a way to specify the URL format we get from the browser so it's not necessary to rely on external libraries.

PR #3345, opened 2019-08-03
improve RTL text support

### LINKED ISSUE #3344: better RTL support
Please improve RTL support as follows:
All user's messages should be aligned to the right
When writing text in the input box it also should be aligned to the right

Jay2k1:
This adds the following for hebrew (or any other RTL) text:
- chat messages will be right-to-left (RTL), but still be aligned to the left (because otherwise on wider screens, there will be a large gap between sender nickname and short messages): 
image
- action messages will be RTL with the sender nickname on the right: 
image
- RTL typing and right-aligned text in the input box: 
image

LTR languages are not affected by this.

Fixes #3344.

Thanks for testing, @NeatNit

### DIFF
--- a/client/components/Message.vue
+++ b/client/components/Message.vue
@@ -19,8 +19,8 @@
 		</template>
 		<template v-else-if="message.type === 'action'">
 			<span class="from"><span class="only-copy">* </span></span>
-			<span class="content">
-				<Username :user="message.from" />&#32;<ParsedMessage
+			<span class="content" dir="auto">
+				<Username :user="message.from" dir="auto" />&#32;<ParsedMessage
 					:network="network"
 					:message="message"
 				/>
@@ -47,7 +47,7 @@
 					<span class="only-copy">- </span>
 				</template>
 			</span>
-			<span class="content">
+			<span class="content" dir="auto">
 				<ParsedMessage :network="network" :message="message" />
 				<LinkPreview
 					v-for="preview in message.previews"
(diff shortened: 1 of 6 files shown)


client/components/Message.vue
```diff
@@ -40,7 +40,7 @@
 					<span class="only-copy">&gt; </span>
 				</template>
 			</span>
-			<span v-else class="from">
+			<span v-else class="from" dir="auto">
```
xPaw: Shouldn't this be on content?

Jay2k1: Yup, my mistake.

client/css/style.css
```diff
@@ -1506,6 +1506,10 @@ background on hover (unless active) */
 	color: #717171;
 }
 
+#chat .toggle-text {
```
xPaw: theres already toggle-text selector on line 1489 where you can put this

NeatNit:
Still needs done:
1. Links should be surrounded by <bdi> to make the arrow appear to the right of the URL:
image

2. Nicks and channel names should have `dir="auto"` on them.

xPaw:
Code change wise this looks fine, I'll need to play with it a little.

Presumably @NeatNit, who can read hebrew, is a-ok with the changes?

NeatNit:
Yup, this behaves very well. It's worth mentioning that I don't actually use IRC in Hebrew in day-to-day use, I merely threw everything I could think of at this and gave feedback for what needed to be changed. All good now.

PR #3349, opened 2019-08-04
Add functionality to edit channel topic from the user interface

### LINKED ISSUE #277: Click on the channel's topic to edit it
You can view, but not directly edit, a channel's topic from The Lounge. It would be nice if you could click on a topic and get an edit prompt so you can change it.

ollipa:
Closes #277.

This PR adds the functionality to edit channel topic from the user interface. The user can either double-click the topic to toggle editing or select "Edit topic" from the channel context menu. The new topic is saved when the user presses enter. Editing is canceled and the changes are discarded if the input field loses focus or if the user presses esc key.

image

image

I've been following this project for quite a while and always thought that I will make a contribution some day. I hope this is helpful :)

### DIFF
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -20,15 +20,18 @@
 				<div class="header">
 					<button class="lt" aria-label="Toggle channel list" />
 					<span class="title">{{ channel.name }}</span>
-					<input
-						v-if="channel.editTopic === true"
-						:value="channel.topic"
-						class="topic-edit"
-						placeholder="Set channel topic"
-						@keyup.enter="saveTopic"
-						@keyup.esc="channel.editTopic = false"
-						@blur="channel.editTopic = false"
-					/>
+					<div v-if="channel.editTopic === true" class="topic-container">
+						<input
+							:value="channel.topic"
+							class="topic-input"
+							placeholder="Set channel topic"
+							@keyup.enter="saveTopic"
+							@keyup.esc="channel.editTopic = false"
+						/>
+						<span aria-label="Save topic" class="save-topic" @click="saveTopic">
+							<span type="button" aria-label="Save topic"></span>
+						</span>
+					</div>
 					<span v-else :title="channel.topic" class="topic" @dblclick="editTopic"
 						><ParsedMessage
 							v-if="channel.topic"
@@ -133,13 +136,13 @@ export default {
 				this.channel.editTopic = true;
 
 				this.$nextTick(() => {
-					document.querySelector(`#chan-${this.channel.id} .topic-edit`).focus();
+					document.querySelector(`#chan-${this.channel.id} .topic-input`).focus();
 				});
 			}
 		},
-		saveTopic(event) {
+		saveTopic() {
 			this.channel.editTopic = false;
-			const newTopic = event.target.value;
+			const newTopic = document.querySelector(`#chan-${this.channel.id} .topic-input`).value;
 
 			if (this.channel.topic !== newTopic) {
 				const target = this.channel.id;
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -20,7 +20,16 @@
 				<div class="header">
 					<button class="lt" aria-label="Toggle channel list" />
 					<span class="title">{{ channel.name }}</span>
-					<span :title="channel.topic" class="topic"
+					<input
+						v-if="channel.editTopic === true"
+						:value="channel.topic"
+						class="topic-edit"
+						placeholder="Set channel topic"
+						@keyup.enter="saveTopic"
+						@keyup.esc="channel.editTopic = false"
+						@blur="channel.editTopic = false"
+					/>
+					<span v-else :title="channel.topic" class="topic" @dblclick="editTopic"
 						><ParsedMessage
 							v-if="channel.topic"
 							:network="network"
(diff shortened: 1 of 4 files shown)


client/components/Chat.vue
```diff
@@ -118,6 +129,24 @@ export default {
 		hideUserVisibleError() {
 			this.$root.currentUserVisibleError = null;
 		},
+		editTopic() {
+			this.channel.editTopic = true;
+
+			this.$nextTick(() => {
+				$(`#chan-${this.channel.id} .topic-edit`).focus();
```
xPaw: We want to eventually remove jquery, would be nice to do with vanilla js

xPaw:
Thanks for the PR, appears to work fine as you'd expect it to. I think there should be a save button (checkmark on the right for example).

client/components/Chat.vue
```diff
@@ -118,6 +129,24 @@ export default {
...
+			});
+		},
+		saveTopic(event) {
+			this.channel.editTopic = false;
+			const newTopic = event.target.value;
+
+			if (newTopic && this.channel.topic !== newTopic) {
```
xPaw: This check prevents you from removing a topic.

client/components/Chat.vue
```diff
@@ -118,6 +129,24 @@ export default {
 		hideUserVisibleError() {
 			this.$root.currentUserVisibleError = null;
 		},
+		editTopic() {
```
xPaw: This should have a check for channel type, right now you can double click in lobby (and other types) and the input field appears.

client/css/style.css
```diff
@@ -978,6 +978,17 @@ background on hover (unless active) */
 	font-size: 14px;
 }
 
+#windows .header .topic-edit {
+	color: #222;
```
xPaw: Doing the following styles makes it work nicely on other themes (e.g. with dark background):

```css
    color: inherit;
    background: inherit;
```

client/components/Chat.vue
ollipa: I actually added that check after I noticed that sending only `/topic` just prints out the current topic in the channel. I could not find a standard way of clearing topic in IRC. However, I could set the topic equal to one white-space character if the user sends an empty topic. What do you think?

xPaw: That's a good point. Leave it as is then.

xPaw: Edit, apparently it's possible by doing `> TOPIC #ircv3 :`, but that also requires changing the server command to work with this.

ollipa:
> Thanks for the PR, appears to work fine as you'd expect it to. I think there should be a save button (checkmark on the right for example).
> 
> EDIT: Any ideas how this will work on mobile due to very limited width?

It actually works quite well at least on Android & Chrome. The input field automatically scrolls if the topic gets too long. See the screenshots below.

image

image

Obviously it is not the best user experience because of the limited space. I was thinking I could later try work on issue #276 and come up with a solution that would improve the topic editing as well on mobile.

I can add the save button, but I think it would be better to hide it in the mobile view because of limited space.

client/components/Chat.vue
xPaw: I created #3351 to handle this

xPaw: Can you change this to send `/raw TOPIC ${channelname} :${newTopic}` command?

I just don't know if `/topic ` is a good idea to clear the topic. #3352 fixes the raw command so the colon works as expected.

ollipa: I changed this to use raw command and tested that it works as expected with your change.

ollipa:
@xPaw, I have now addressed all your comments. I'll check how to add the save button next.

xPaw:
That `/raw` PR was merged, you can rebase your PR so it can be tested correctly.

ollipa:
I rebased and added a save button to the edit topic input field:
image

Previously the editing was canceled if the edit topic input field lost focus (clicking anywhere outside of the input box). I could not figure out a simple way to make the save button work together with that functionality so I removed it. Editing can still be canceled by hitting esc key or just saving the topic without making any changes.

xPaw:
Thanks for the PR! Anything you want to do before merge?

> I could not figure out a simple way to make the save button work together with that functionality so I removed it.

What trouble was it causing?

client/css/style.css
```diff
@@ -978,6 +984,42 @@ background on hover (unless active) */
...
+}
+
+.topic-container .save-topic {
+	position: absolute;
+	top: 6px;
+	right: 0;
+}
+
+.topic-container .save-topic span {
+	font: normal normal normal 14px/1 FontAwesome;
```
xPaw: Look on line 249, we define the fontawesome font in the big selector there.

ollipa:
> Thanks for the PR! Anything you want to do before merge?
> 

I think this is ready to merge. I've tested the changes myself with Chrome, Firefox & Chrome on Android and everything seems to work as intended.

> > I could not figure out a simple way to make the save button work together with that functionality so I removed it.
> 
> What trouble was it causing?

Earlier I had a Vue event handler on the input field, which canceled the editing when the input field lost focus:
```
@blur="channel.editTopic = false"
```
After adding the save button, this event handler was triggered also when the save button was clicked, causing the changes to be discarded instead of being saved.

xPaw:
Perhaps nextTick on blur could help?

client/css/style.css
ollipa: This is fixed now.

ollipa:
> Perhaps nextTick on blur could help?

I tried that but it didn't work.

xPaw:
Oh well, problem for another day.

PR #3351, opened 2019-08-05
Allow clearing topic by sending `/topic ` command

xPaw:


xPaw:
I just don't know if it's actually a good idea, because you might not expect this to clear the topic.

irccloud does it this way:

> 20:01 <@jwheare> xPaw: click "edit topic" button
> 20:07 <@jwheare> we don't have a command other than raw

dgw:
I'd be against if doing `/top<tab>` completed to `/topic ` with the space, but since command autocomplete does not append a space it's probably unlikely that people will trigger this accidentally.

That said, I do agree that the behavior this adds is probably unexpected. I've never seen a way like this to clear the topic of a channel without sending a raw `TOPIC #channel :` command.

xPaw:
Yeah I'm gonna close this, with the UI to edit topic, this will be less of a problem.

PR #3359, opened 2019-08-08
Add keybind for cycling to the next unread window

plett:
This PR replicates Irssi and WeeChat's `Alt+A` keyboard shortcuts which move to the next window which has unread activity in it.

I am not a front-end developer, feel free to suggest improvements!

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -3,6 +3,7 @@
 const $ = require("jquery");
 const Mousetrap = require("mousetrap");
 const utils = require("./utils");
+const {vueApp} = require("./vue");
 
 Mousetrap.bind(["alt+up", "alt+down"], function(e, keys) {
 	const sidebar = $("#sidebar");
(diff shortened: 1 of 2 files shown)


client/js/keybinds.js
```diff
@@ -70,6 +70,21 @@ Mousetrap.bind(["alt+shift+up", "alt+shift+down"], function(e, keys) {
 	return false;
 });
 
+Mousetrap.bind(["alt+a"], function() {
+	const sidebar = $("#sidebar");
+	const channels = sidebar
```
xPaw: Would be great if you could switch this filtering to use Vue state. We want to move away from jquery fully, and keybinds file has not been touched yet, but this looks like a great place to start.

Here's example code on how to loop through channels: https://github.com/thelounge/thelounge/blob/b4eb538903c7f2b1b865302829165591c7812cf9/client/js/utils.js#L58

plett: Thanks for the suggestion. I've re-written it using Vue, see the latest force-push.

I couldn't work out how to switch to the channel using Vue, so that is still using jquery in the same way that `client/js/commands/join.js` does it.

I've never touched Vue before today, so feel free to tell me more things that could be improved!

xPaw: > so that is still using jquery in the same way that client/js/commands/join.js does it.

Yeah channel/window switching hasn't been ported just yet.

client/js/keybinds.js
```diff
@@ -70,6 +71,33 @@ Mousetrap.bind(["alt+shift+up", "alt+shift+down"], function(e, keys) {
...
+	let targetchan;
+
+	outer_loop: for (const network of vueApp.networks) {
+		for (const chan of network.channels) {
+			if (chan.highlight) {
+				targetchan = chan;
+				break outer_loop;
+			}
+
+			if (chan.unread) {
```
xPaw:
```suggestion
			if (chan.unread && !targetchan) {
```

xPaw:
Thanks for your PR! This works fine as described.

xPaw:
My bad on the github suggestion, that inner `if (!targetchan)` shouldn't be there.

Fix that up and squash?

plett:
> Fix that up and squash?

Done. Thanks for your help!

PR #3362, opened 2019-08-12
Adds data-name tag to input form per channel

gunnvaldr:
Allows for custom CSS filtering per channel for the input form


richrd:
If a data attribute is added for this it could made more general purpose if it was added to e.g. #chat-container here: https://github.com/thelounge/thelounge/blob/master/client/components/Chat.vue#L2

That way any of the elements in the chat view could be styled based on which channel it is.
Similar to the aria-label here https://github.com/thelounge/thelounge/blob/master/client/components/Chat.vue#L17

xPaw:
Agree with @richrd, and the attribute could be named like `data-current-channel` to make it a little more verbose.

PR #3366, opened 2019-08-13
Adds data-current-channel to #chat-container

gunnvaldr:
Allows for broader customization of CSS per channel(s).


richrd:
As discussed in #3362 I think this adds some valuable custom CSS possibilities so approving :+1:

PR #3404, opened 2019-09-15
Implement keep nick when client gets "nick in use" on connection

### LINKED ISSUE #3380: Let users provide a nickname AND alternative-nickname to prevent continuous nickname lengthening on reconnects.
Having this discussed earlier I want to open an issue for this:

Currently, when TheLounge loses internet connection (the 'connection reset by peer'-case) and it starts to reconnect to irc-networks, the currently used nickname is already taken (by the old connection). It takes several minutes until the old connection drops out and the nickname is free again.
TheLounge currently handles this case by adding a random number to the current nickname - and does this on every further reconnect!
So if you are away for some days and when in the meantime several disconnects have happened, you might end up having a nickname that has been built like this:
`Mynick -> Mynick4 -> Mynick42 -> Mynick429 -> Mynick4291 -> Mynick42917 -> etc.`
This is a bad behaviour for several reasons but most important it isn't even necessary. 
There is a reason why as good as all IRC-Clients let users provide a nickname + an alternative-nickname. Because if one nickname is already taken, it will use the other one. It always alternates between the two chosen nicknames and doesn't add several random numbers/text to the nick. In most cases the two nicknames are sufficient.
Some argue it isn't necessary on freenode because there's nickserv and you can free your nick. Well, that's true for irc-networks with nickserv and the like. But there are dozens of irc-networks that don't provide that feature and here you have to rely on alternative nicknames.

To fix this, add a second field under the nickname setting to provide an alternative nickname and let TheLounge alternate between those nicknames on reconnect if one is already taken.

xPaw:
Clients usually get nick in use on connect when reconnecting to a network after a network failure (like ping timeout), and as a result of that, TL will append a random number to the nick.

keepNick will try to set the original nick name back if it sees a QUIT for that nick. This should work fine in most cases, as you'll rejoin the same channels you were in. Changing nick or editing network will cancel this.

This essentially implements a subset of ZNC's altnick+keepnick features while being hidden from the user.

Fixes #3380

### DIFF
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -19,6 +19,7 @@ const filteredFromClient = {
 	irc: true,
 	password: true,
 	ignoreList: true,
+	keepNick: true,
 };
 
 function Network(attr) {
@@ -43,6 +44,7 @@ function Network(attr) {
 		},
 		chanCache: [],
 		ignoreList: [],
+		keepNick: null,
 	});
 
 	if (!this.uuid) {
@@ -188,6 +190,7 @@ Network.prototype.edit = function(client, args) {
 	const oldNick = this.nick;
 	const oldRealname = this.realname;
 
+	this.keepNick = null;
 	this.nick = args.nick;
 	this.host = String(args.host || "");
 	this.name = String(args.name || "") || this.host;
@@ -217,7 +220,7 @@ Network.prototype.edit = function(client, args) {
 		if (this.nick !== oldNick) {
 			if (connected) {
 				// Send new nick straight away
-				this.irc.raw("NICK", this.nick);
+				this.irc.changeNick(this.nick);
 			} else {
 				this.irc.options.nick = this.irc.user.nick = this.nick;
 
@@ -269,6 +272,10 @@ Network.prototype.setNick = function(nick) {
 		// Case insensitive search
 		"i"
 	);
+
+	if (this.keepNick === nick) {
+		this.keepNick = null;
+	}
 };
 
 /**
--- a/src/plugins/irc-events/error.js
+++ b/src/plugins/irc-events/error.js
@@ -44,10 +44,22 @@ module.exports = function(irc, network) {
 	});
 
 	irc.on("nick in use", function(data) {
+		let message = data.nick + ": " + (data.reason || "Nickname is already in use.");
+
+		if (irc.connection.registered === false && !Helper.config.public) {
+			message += " An attempt to use it will be made when this nick quits.";
+
+			// Clients usually get nick in use on connect when reconnecting to a network
+			// after a network failure (like ping timeout), and as a result of that,
+			// TL will append a random number to the nick.
+			// keepNick will try to set the original nick name back if it sees a QUIT for that nick.
+			network.keepNick = irc.user.nick;
+		}
+
 		const lobby = network.channels[0];
 		const msg = new Msg({
 			type: Msg.Type.ERROR,
-			text: data.nick + ": " + (data.reason || "Nickname is already in use."),
+			text: message,
 			showInActive: true,
 		});
 		lobby.pushMessage(client, msg, true);
--- a/src/plugins/inputs/nick.js
+++ b/src/plugins/inputs/nick.js
@@ -41,16 +41,27 @@ exports.input = function(network, chan, cmd, args) {
 		return;
 	}
 
+	// If we were trying to keep a nick and user changes nick, stop trying to keep the old one
+	network.keepNick = null;
+
 	// If connected to IRC, send to server and wait for ACK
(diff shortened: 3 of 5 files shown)


src/plugins/irc-events/error.js
```diff
@@ -44,15 +44,27 @@ module.exports = function(irc, network) {
 	});
 
 	irc.on("nick in use", function(data) {
+		let message = data.nick + ": " + (data.reason || "Nickname is already in use.");
+
+		if (irc.connection.registered === false) {
+			message += " We will try to set this nick back if we see old client quit.";
```
xPaw: Probably needs better wording (who is we?)

Jay2k1: Instead of "we", I would use "I". Some suggestions:
* `Nickname is already in use. Will retry nickchange when this nick is seen quitting.`
* `Nickname is already in use. I will retry nickchange when this nick is seen quitting.`
* `Nickname is already in use. Will retry when this nick is seen quitting.`
* `Nickname is already in use. I will retry when this nick is seen quitting.`
* `Nickname is already in use. Attempting to get it back...`

McInkay: I think "I" is no better, because the issue is we are there going "Who is I?". Whereas if we actually did "The Lounge will", or "TL will" if we insist that is too long, then it's very clear what is doing it. We don't have to figure out if it's the lounge, the server, or a bot that we haven't noticed.

xPaw: We already use wording like "and will not reconnect.", and I don't think we ever reference TL. With that in mind, perhaps this?

> Nickname is already in use. An attempt to use it will be made when this nick is seen quitting.

McInkay: Yeah I guess that could work. Certainly sounds less bad than "we" or "I"

src/plugins/irc-events/error.js
```diff
@@ -44,10 +44,22 @@ module.exports = function(irc, network) {
 	});
 
 	irc.on("nick in use", function(data) {
+		let message = data.nick + ": " + (data.reason || "Nickname is already in use.");
+
+		if (irc.connection.registered === false && !Helper.config.public) {
+			message += " An attempt to use it will be made when this nick quits.";
```
MaxLeiter: `quits` -> `becomes available`?

xPaw: `quits` at least reflects when TL will actually attempt to use that nick, if you say `becomes available`, people might expect a little more magic behind that phrasing.

src/models/network.js
```diff
@@ -269,6 +272,10 @@ Network.prototype.setNick = function(nick) {
 		// Case insensitive search
 		"i"
 	);
+
+	if (this.keepNick === nick) {
+		this.keepNick = null;
```
McInkay: I've not looked, but I'm presuming this is messing up your build. Yay prettier.

src/models/network.js
```diff
@@ -43,6 +44,7 @@ function Network(attr) {
 		},
 		chanCache: [],
 		ignoreList: [],
+		keepNick: null,
```
McInkay: Make sense to default it to false rather than null?

src/plugins/irc-events/error.js
McInkay: Yeah, `quit` is probably too jargon.

Maybe...

```suggestion
			message += " An attempt to use it will be made when this nick is no longer in use.";
```

src/models/network.js
McInkay: Oh, hmmm. Your build is passing. I wonder why this is messed up then, I wonder if it's just github UI getting confused with something.

McInkay: Oh. It's refined github messing it up. Ignore me.

src/models/network.js
xPaw: Using null means its `?string` instead of `string|bool`, which I think makes more sense here.

src/plugins/irc-events/error.js
xPaw: `quit` is an action you can see in channels, so I don't think it's too jargon. I'd rather have this sentence attached to an actual action (which hopefully prevents people from questioning on what use/available means).

If we update keep nick logic to also work in other scenarios like nick changes, then we can certainly change this wording ☺

src/models/network.js
McInkay: Oh sorry, I had thought this was a config to let them decide to have keepNick or not, ignore me. Good point.

src/plugins/irc-events/error.js
McInkay: I'm not convinced tbh, but I won't fight it. It's not something that'll come up much anwyay.

PR #3406, opened 2019-09-17
Print error and stacktrace when package fails to load

xPaw:
Stacktraces are needed for plugin authors to fix their code.

I also moved `!packageInfo.thelounge` check higher so we don't load the package at all if it's not meant for lounge.

### DIFF
--- a/src/plugins/packages/index.js
+++ b/src/plugins/packages/index.js
@@ -65,23 +65,22 @@ function loadPackages() {
 	}
 
 	packages.forEach((packageName) => {
-		const errorMsg = `Package ${colors.bold(packageName)} could not be loaded`;
 		let packageInfo;
 		let packageFile;
 
 		try {
-			packageInfo = require(path.join(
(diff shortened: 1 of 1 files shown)


AlMcKinlay (commented):
Debug is enabled on a flag, right?

xPaw:
@mcinkay, nope we have no debug flag.

McInkay:
Oh, our log is just a wrapper around colouring console logs.

I'm wondering whether it's worth always printing the stack trace, or whether that would be confusing for users. Should we put in a debug flag to hide this behind?

xPaw:
We can probably introduce a verbosity/debug flag to hide `log.debug` messages behind, not really a job for this PR though.

I do think always printing stacktrace is a good way of doing it, users have to hunt less to get extra details.

McInkay:
Alright, let's argue about that outside this PR.

AlMcKinlay (approved):

PR #3416, opened 2019-09-24
Set correct file owner for created user files and warn about it

xPaw:
This is targeted at users that run `thelounge add` as root when they have TL installed a service.

This will fix the file owner for them, but also complain about it and link to the documentation.


xPaw:
Also fixed up the config check on start, to fix this nonsense with async console printing:

TL

PR #3430, opened 2019-10-01
Update emoji maps

stevenengler:
Problem: The Lounge does not recognize certain emojis such as the unmodified "Cloud With Rain" emoji (🌨). You can test this by pasting it into a chat and seeing that the size is incorrect.

The current emoji maps used by The Lounge have the "FE0F" variation selector, but the latest version of the Github emojis has [removed this codepoint](https://github.com/github/gemoji/issues/163) from the sequence for many emojis.

The current list used by The Lounge causes problems with emojis such as the "Cloud With Rain" emoji, which has the codepoint "U+1F327". The Lounge does not currently recognize this emoji since it is in the emoji map as "U+1F327-U+FE0F", and therefore does not get surrounded by the `<span class="emoji">` tag.

The updated maps were generated by simply re-running the "generate-emoji.js" script.

Note: I am not able to test this change, so it needs to be tested by someone who is set up to do so. I'm also not 100% sure this will fix the problem, but I believe it will.


xPaw:
There's two tests failing now because they have emojis with FE0F, can you update these please?

client/js/libs/handlebars/parse.js
```diff
@@ -10,6 +10,7 @@ const colorClass = require("./colorClass");
 const emojiMap = require("../fullnamemap.json");
 const LinkPreviewToggle = require("../../../components/LinkPreviewToggle.vue").default;
 const emojiModifiersRegex = /[\u{1f3fb}-\u{1f3ff}]/gu;
+const emojiVariationsRegex = /[\u{fe0e}-\u{fe0f}]/gu;
```
xPaw: I think it's better to combine it with previous regex for perf reasons.

Could you also add some tests for this specifically? So we know emojis with fe0e and without match correctly.

stevenengler: Okay sounds good. And I'll try adding some tests, but I don't use Node and don't have it set up, although I think it should be easy to copy and modify one of the existing tests though.

xPaw:
Actually looking at `simplemap.json` diff, it seems to break emojis that had variant selector before.

02-1428-CaringRainbowfish

And besides, these emojis without the variant selector still won't get bigger in Lounge because the variant selector identifies them as emojis (which are found by `emoji-regex`, not from these maps).

gemoji has this code that has hardcoded list of emojis which need the modifier: 
https://github.com/github/gemoji/blob/59b3fde91efa85cc576c8946cd7e69f6b2bd6a49/lib/emoji.rb#L58-L61

So this PR currently doesn't solve the initial problem you described, and breaks these existing emojis by making them non-emoji-variant.

stevenengler:
Can you confirm that pasting this emoji into a message in The Lounge correctly identifies it as an emoji?

https://pad.riseup.net/p/zxmYJpnnHIg8fqPtbXrW

It does not on my version of The Lounge (the latest release 3.2.0). It is not identified as an emoji and therefore not surrounded by a span tag.

And regardless of merging this PR or not, if you ever want to update your emoji list from gemoji in the future, you will need to update The Lounge to properly parse emojis in the new fixed format that they're using.

xPaw:
Well yeah, the update script would need to be updated to construct emojis with the variant. I believe it is correct not to detect these symbols without the emoji variant (again, it's the emoji-regex doing it, not this emoji map).

stevenengler:
Okay if you only want to detect emojis with the variant explicitly specified and not all emojis, then this commit is not what you want. My change matches emojis with any variant selector since some fonts (for example [Google's Noto Color Emoji](https://www.google.com/get/noto/help/emoji/)) render emojis without a variant selector the same as having an emoji variant selector. The Lounge should probably set fonts that use text-based emojis by default if no variant is specified (at least on stock Ubuntu).

PR #3433, opened 2019-10-02
New feature: Added onJoin commands for channels

lacrioque:
Hey, I always had problems with running the op command on one of my regular channels, since it always ran before the connection to the channel was made.

I have just added a new parameter for the channel to have a channel separate command section, that is triggered after the connection to the channel was successfull.
I have also updated the tests that are connected.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -164,6 +164,7 @@ Client.prototype.connect = function(args) {
 					name: chan.name,
 					key: chan.key || "",
 					type: chan.type,
+					commands: chan.commands || [],
 				})
 			);
 		});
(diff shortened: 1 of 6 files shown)


McInkay:
@lacrioque Do you have any interest in changing how this works, and making a join event for plugins instead of a hard coded feature for commands?

xPaw:
I agree that this isn't really a good feature to add in core, this indeed would be better suited for plugins.

That being said, if you specify commands when editing a network, they run *after* all join commands are sent.

lacrioque:
Hey sorry for the late response. 
I will check to make this a plugin. 
Sounds interesting.

PR #3435, opened 2019-10-03
Key condensed messages by first message in array

### LINKED ISSUE #3139: Condensed messages container closes when updated
*The Lounge version:* master after vue merge

1. It has to be at the bottom
2. Open a condensed container
3. Receive another event (join/leave/etc)
4. Condensed container closes itself

What probably happens is that the `MessageCondensed` component gets re-rendered and the collapsed state gets reset:
https://github.com/thelounge/thelounge/blob/892737d1d8ade7b80fe22356a191c6715a0c08e6/client/components/MessageCondensed.vue#L39

FryDay:
By changing `MessageCondensed` key to the first message id in the array (instead of last), Vue will no longer redraw the component and it will stay open.

Fixes #3139

### DIFF
--- a/client/components/MessageList.vue
+++ b/client/components/MessageList.vue
@@ -34,7 +34,7 @@
 
 				<MessageCondensed
 					v-if="message.type === 'condensed'"
-					:key="message.id"
+					:key="message.messages[0].id"
 					:network="network"
 					:keep-scroll-position="keepScrollPosition"
 					:messages="message.messages"


xPaw:
That id is there for a reason, I remember specifically having issues when writing that code.

Are you sure it doesn't break unread marker? (Before condensed, after, between)

FryDay:
@xPaw: Correct, v-for forces a re-render of a component if the id changes. I thought about setting the id to the first message in the condensed container but it looks like that's how it was prior to the current code. https://github.com/thelounge/thelounge/commit/a21e0e34cd8b53d1853f1a63e3755cd7734a2c5a

xPaw:
> I thought about setting the id to the first message in the condensed container

That's I suggested using first id just for keying (add a new prop), instead of changing the existing id.

FryDay:
Oh I see, I misread your comment. I'll take a look at trying that. 👍

FryDay:
I don't think there is a need for another prop. A condensed message component won't be created unless there are messages in the array, so using the first by index should be okay.

xPaw:
Yep you're right, I just tested it and it seems to work fine.

Could you squash the commits, then I can merge this.

Thanks!

PR #3437, opened 2019-10-04
Allow tab completion in middle of input

### LINKED ISSUE #1801: [tab] nick completion does not work when cursor is not at the end
`ast[tab] some text` will not work. This only affects the non-textcomplete tab completion.

FryDay:
Fixes #1801


xPaw:
Thanks!

This appears to work as expected, however the cursor jumps to the end of input after <tab>, can we set it before remainder text? Just like @-autocompletion works.

xPaw:
Could you cache `input.get(0)` into a variable in that callback please.

PR #3438, opened 2019-10-04
Prefix channel before join

### LINKED ISSUE #1154: Being able to join a channel without a prefix
I've found that in KiwiIRC you're able to join a channel by simply typing its name without the need of a channel prefix.

After mentioning about it on #thelounge, @YaManicKill suggested that if this is taken into consideration, there has to be a structured way in which the prefix is assumed and provided a relevant link to the [specs](https://tools.ietf.org/html/rfc1459#section-1.3). @prawnsalad pointed out that `#` and `&` should not be taken as the default prefix and suggested to check the [network isupport](http://www.irc.org/tech_docs/005.html) which contains the supported channel prefixes. In addition, he mentioned those are already available in `client.network.options`. 

Intially this seemed feasible with a client side check on whether the first character of a channel is not `#` or `&` after which the code would prepend a `#`, making `/join thelounge` -> `/join #thelounge` before it goes to the server.

After some awareness on ISUPPORT it seems the default prepend character has to depend on the CHANTYPES of the current server. If this information is indeed already processed, it should be fairly straight-forward to implement.

But then, is the first character of the CHANTYPES value the default channel prefix that the server uses? Perhaps some further docs look-ups or server config investigations might answer it.

FryDay:
Fixes #1154

### DIFF
--- a/client/components/JoinChannel.vue
+++ b/client/components/JoinChannel.vue
@@ -68,8 +68,15 @@ export default {
 				const $ = require("jquery");
 				$(`#sidebar .chan[data-id="${existingChannel.id}"]`).trigger("click");
 			} else {
+				const chanTypes = this.network.serverOptions.CHANTYPES;
+				let channel = this.inputChannel;
+
+				if (chanTypes && chanTypes.length > 0 && !chanTypes.includes(channel[0])) {
+					channel = chanTypes[0] + channel;
+				}
+
 				socket.emit("input", {
-					text: `/join ${this.inputChannel} ${this.inputPassword}`,
+					text: `/join ${channel} ${this.inputPassword}`,
 					target: this.channel.id,
 				});
 			}
--- a/client/js/commands/join.js
+++ b/client/js/commands/join.js
@@ -8,13 +8,33 @@ exports.input = function(args) {
 	const {vueApp} = require("../vue");
 
 	if (args.length > 0) {
-		const channel = args[0];
+		let channels = args[0];
 
-		if (channel.length > 0) {
-			const chan = utils.findCurrentNetworkChan(channel);
+		if (channels.length > 0) {
+			const chanTypes = vueApp.activeChannel.network.serverOptions.CHANTYPES;
+			const channelList = args[0].split(",");
+
+			if (chanTypes && chanTypes.length > 0) {
+				for (let c = 0; c < channelList.length; c++) {
+					if (!chanTypes.includes(channelList[c][0])) {
+						channelList[c] = chanTypes[0] + channelList[c];
+					}
+				}
+			}
+
+			channels = channelList.join(",");
+
+			const chan = utils.findCurrentNetworkChan(channels);
 
 			if (chan) {
 				$(`#sidebar .chan[data-id="${chan.id}"]`).trigger("click");
+			} else {
+				socket.emit("input", {
+					text: `/join ${channels} ${args.length > 1 ? args[1] : ""}`,
+					target: vueApp.activeChannel.channel.id,
+				});
+
+				return true;
 			}
 		}
 	} else if (vueApp.activeChannel.channel.type === "channel") {


xPaw:
This changes the join channel UI, but not the `/join` command.

https://github.com/thelounge/thelounge/blob/531ca494a451ef4ffb71ac8473f99af6258c8e66/client/js/commands/join.js

client/components/ChatInput.vue
```diff
@@ -210,6 +210,23 @@ export default {
 				const args = text.substr(1).split(" ");
 				const cmd = args.shift().toLowerCase();
 
+				if (cmd === "join") {
```
xPaw: Can you move this to `/js/commands/join.js` file that specifically handles the join command?

FryDay: See if you like that better. It feels a little wonky to me, but it does work.

client/js/commands/join.js
```diff
@@ -8,14 +8,37 @@ exports.input = function(args) {
...
+			let hasPrefix = false;
+
+			for (const prefix of chanTypes) {
+				if (channel.startsWith(prefix)) {
+					hasPrefix = true;
+					break;
+				}
+			}
+
+			if (!hasPrefix && chanTypes && chanTypes.length > 0) {
```
xPaw: You check `chanTypes` here but already tried to loop above. both of them need to be wrapped.

client/js/commands/join.js
```diff
@@ -8,14 +8,37 @@ exports.input = function(args) {
...
+				channel = chanTypes[0] + channel;
+			}
+
 			const chan = utils.findCurrentNetworkChan(channel);
 
 			if (chan) {
 				$(`#sidebar .chan[data-id="${chan.id}"]`).trigger("click");
 			}
+
+			if (!hasPrefix) {
```
xPaw: `else if`?

client/components/JoinChannel.vue
```diff
@@ -68,6 +68,20 @@ export default {
...
+
+				for (const prefix of chanTypes) {
+					if (this.inputChannel.startsWith(prefix)) {
+						hasPrefix = true;
+						break;
+					}
+				}
+
+				if (!hasPrefix && chanTypes && chanTypes.length > 0) {
+					this.inputChannel = chanTypes[0] + this.inputChannel;
```
xPaw: Probably shouldn't override `this.inputChannel`, make a local var.

client/js/commands/join.js
```diff
@@ -8,14 +8,37 @@ exports.input = function(args) {
...
+				socket.emit("input", {
+					text: `/join ${channel} ${args.length > 1 ? args[1] : ""}`,
```
xPaw: I do wonder if need to care about comma separated channels because you can join multiple channels at once.

/join #chan1,#chan2
/join chan1,#chan2
/join #chan1,chan2

etc. How should we handle prefixes here?

https://tools.ietf.org/html/rfc2812#section-3.2.1

FryDay: Good question. Joining a list of channels does not appear to work without these changes either.

xPaw: It does work for me. (without this pr at least)

FryDay: You're right, it does. I had a space after the comma in the list. I suppose I could split the list up and loop through.

FryDay: There, that should work for all scenarios you posted.

client/js/commands/join.js
```diff
@@ -8,13 +8,42 @@ exports.input = function(args) {
...
+		if (channels.length > 0) {
+			const chanTypes = vueApp.activeChannel.network.serverOptions.CHANTYPES;
+			const channelList = args[0].split(",");
+
+			if (chanTypes && chanTypes.length > 0) {
+				for (let c = 0; c < channelList.length; c++) {
+					let hasPrefix = false;
+
+					for (const prefix of chanTypes) {
+						if (channelList[c].startsWith(prefix)) {
```
xPaw: What about `chanTypes.includes(channelList[c][0])`? That would remove the loop.

We do the same here: https://github.com/thelounge/thelounge/blob/b8948856f3db0c0a693d9acdfa13c8ce3fd1160b/src/plugins/inputs/msg.js#L35-L41

PR #3439, opened 2019-10-05
Add Webpack context replacement plugin to remove vue server renderer errors

JessWallin:
Webpack was throwing the following critical warning while compiling `vue-server-renderer`:

```
Warning in ./node_modules/vue-server-renderer/build.dev.js
Critical dependency: the request of a dependency is an expression
```

This was due to dynamic imports in the `vue-server-renderer` module (line 9301 and related function). Since I can't change the offending function, I added a webpack plugin for the `vue-server-renderer` module to provide context for these dynamic imports and clear the error.

Hope it helps!

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -147,6 +147,10 @@ const config = {
 			/debug/,
 			path.resolve(__dirname, "scripts/noop.js")
 		),
+		new webpack.ContextReplacementPlugin(
+			/vue-server-renderer/,
+			path.resolve(__dirname, "./src/vue-server-renderer")
+		),
 	],
 };
 


webpack.config.js
```diff
@@ -147,6 +147,10 @@ const config = {
 			/debug/,
 			path.resolve(__dirname, "scripts/noop.js")
 		),
+		new webpack.ContextReplacementPlugin(
+			/vue-server-renderer/,
+			path.resolve(__dirname, "./src/vue-server-renderer")
```
xPaw: This doesn't resolve to anything useful. Just changing it to a random string also gets rid of the warning for some reason.

Any idea why that happens?

JessWallin: I'm guessing it's just not specific enough to do what I think it's doing, so it's fixing the file path error by providing a static context for dynamic imports, but not in a way that it's limited to the specific module that's causing the error. I haven't worked with this type of webpack plugin before so it probably needs more finesse. 

I'm getting it to throw some new errors locally though, so I'll see if I can get it to work as intended.

xPaw: Did you get a chance to look at this?

xPaw:
I put this as `new webpack.ContextReplacementPlugin(/vue-server-renderer$/)` in the `vue` branch, good enough fix for now.

Thanks.

PR #3446, opened 2019-10-12
updateded css-loader to 3.2.0

ericblancas23:
The previous CSS-loader has been updated to 3.2.0 and compatibility has been reassured to work with the newer versions of css-loader.


MiniDigger:
don't commit IDE metadata please

xPaw:
Just updating the package is not enough. Since their v1 version they removed a lot of options like minifying and autoprefixer.

PR #3449, opened 2019-10-14
Use postcss to optimize css

### LINKED ISSUE #3154: Update or replace css-loader and minifier
We currently have old version of css-loader because the major update removed a lot of stuff related to minification so we never updated.

We should keep our css deps updated to make sure it doesn't break in the future.

Some requirements:
- Emit clean css file so it can be loaded directly with a link tag.
- Minification
- Combine `@import` into one file

There's modules like postcss-loader and cssnano that can be looked at.

xPaw:
Fixes #3154

- [x] Added `.browserslistrc` file so any plugin that uses browserslist will have correct target
- [x] Updated css-loader
- [x] Full css minification (including imported css)
- [x] Autoprefixer
- [x] cssnano optimizations
- [x] correct sourcemap after all processing
- [ ] Still can't import `primer-tooltips.css` properly, I don't know how to fix it.


xPaw:
It's not a new issue, we already have it in master. Basically we're unable to `@import` the tooltips css (because its in node_modules) for whatever reason.

McInkay:
Ah, of course, because `@import` is css and dealt with on the browser, rather than at build time.

I guess there are 2 options there, that I can think of:

1. We use a pre-processor to pull that in at build time
2. We create a route on the backend that forwards to css files in `node_modules`

xPaw:
It is dealt at build time, we already have that in the build process. That's where it breaks.

PR #3453, opened 2019-10-15
Synchronize user list correctly on reconnection

xPaw:


client/js/socket-events/init.js
```diff
@@ -163,7 +163,16 @@ function mergeChannelData(oldChannels, newChannels) {
 			// Server sends an empty users array, client requests it whenever needed
 			if (key === "users") {
 				if (channel.type === "channel") {
-					channel.usersOutdated = true;
```
xPaw: This should have been `currentChannel` instead of `channel`.

This line effectively did not because the `usersOutdated` prop was not copied into the existing object.

PR #3454, opened 2019-10-15
Automatically apply GitHub labels on issue creation

Zarthus:
Automatically apply the bug label on bugs and the feature label on features if using the github commit template. As seen here: https://github.com/Zarthus/thelounge/issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_Report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.md
@@ -1,6 +1,7 @@
 ---
 name: Bug Report
 about: Create a bug report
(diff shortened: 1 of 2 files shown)


xPaw:
We kind of discussed this before and came to the conclusion that auto tagging isn't always desirable.

But perhaps @thelounge/maintainers have another opinion now?

Zarthus:
When I went over the issues it looked to me like it was an overall net-benefit and couldn't really find any issue where it would have been 'wrong'. It's especially helpful that there's no distinguishment in the labels between enhancements and features. 

Maybe there is value in letting unconfirmed bugs be unlabeled though.

PR #3471, opened 2019-10-22
Add message type for plugins

MiniDigger:
This adds a new public api to send messages, with their own message type so that its clearly visible to a user where the message came from.
(it also changes some internals to make the package info available in the api)

Usage (in some command for example):
```js
const red = '04';
client.sendMessage(red + "You did something wrong", target.chan);
client.sendMessage("Try this instead", target.chan);
```

Ideally publicClient would automatically know who just called the method so that plugins don't need to pass their identifier every time, but I wasn't sure on how to accomplish that nicely. 
We could have a "shortName" in the package info for that I guess.

### DIFF
--- a/client/components/Message.vue
+++ b/client/components/Message.vue
@@ -43,7 +43,7 @@
 			<span v-else-if="message.type === 'plugin'" class="from">
 				<template v-if="message.from && message.from.nick">
 					<span class="only-copy">[</span>
-					*{{ message.from.nick }}
+					{{ message.from.nick }}
 					<span class="only-copy">] </span>
 				</template>
 			</span>
--- a/client/components/Message.vue
+++ b/client/components/Message.vue
@@ -40,6 +40,13 @@
 					<span class="only-copy">&gt; </span>
 				</template>
 			</span>
+			<span v-else-if="message.type === 'plugin'" class="from">
+				<template v-if="message.from && message.from.nick">
+					<span class="only-copy">[</span>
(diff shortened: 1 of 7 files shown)


client/components/Message.vue
```diff
@@ -40,6 +40,13 @@
 					<span class="only-copy">&gt; </span>
 				</template>
 			</span>
+			<span v-else-if="message.type === 'plugin'" class="from">
+				<template v-if="message.from && message.from.nick">
+					<span class="only-copy">[</span>
+					*{{ message.from.nick }}
```
xPaw: We can style this nicer, give it an icon with FA for example.

src/plugins/packages/publicClient.js
```diff
@@ -37,4 +39,24 @@ module.exports = class PublicClient {
...
+	 * @param {String} identifier the identifier/name of the packages that is sending this message
+	 */
+	sendMessage(text, chan, identifier) {
+		chan.pushMessage(
+			this.client,
+			new Msg({
+				type: Msg.Type.PLUGIN,
+				text: text,
+				from: {
+					nick: identifier,
```
xPaw: Perhaps we can use (or at least default to) package's name?

xPaw:
Can you make a corresponding PR to the site repo documenting this? Thanks.

MiniDigger:
sure, https://github.com/thelounge/thelounge.github.io/pull/220

PR #3481, opened 2019-10-29
Remove punctuation mark in settings

muesli:
I assume this was just an oversight. Keeps it consistent with all other check-boxes on the settings page.


xPaw:
Thanks, I fixed this in the vue branch and I will close this PR so we don't unnecessarily have to rebase.

https://github.com/thelounge/thelounge/blob/f434cc546e922cdacca70406e3720b8623f1d728/client/components/Windows/Settings.vue#L60

PR #3482, opened 2019-10-31
Load existing users on startup when LDAP is enabled

### LINKED ISSUE #3219: Load existing LDAP users on startup
Currently LDAP users are not loaded on startup.  

This means that when thelounge is restarted, users are not connected to their IRC networks in between the service restarting and when they first authenticate.

```
<xPaw> oh i see, it might be that ldap users aren't loaded at all on start
<xPaw> https://github.com/thelounge/thelounge/blob/061fa615cea3297d97daf7d3054933d741c2b302/src/clientManager.js#L23-L25 which would be this
<xPaw> cli log would tell you when the user is loaded
<thinkl33t> Yeah, i get <nothing> <auth> <error: user already exists> <user loaded>
<xPaw> the fix would to split the code in autoloadUsers before "fs.watch" into a separate func, and call it regardless of the ldap option
```

xPaw:
Fixes #3219

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -20,8 +20,14 @@ ClientManager.prototype.init = function(identHandler, sockets) {
 	this.identHandler = identHandler;
 	this.webPush = new WebPush();
 
-	if (!Helper.config.public && !Helper.config.ldap.enable) {
-		this.autoloadUsers();
+	if (!Helper.config.public) {
+		this.loadUsers();
+
+		// LDAP does not have user commands, and users are dynamically
+		// created upon logon, so we don't need to watch for new files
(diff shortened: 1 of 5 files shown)


xPaw:
@McInkay I have not, but there shouldn't be any issues as the overall logic is the same. The only difference was that having LDAP enabled loaded the user file upon logon (because it goes through auth then checks if user file exists).

McInkay:
Presumably there is a check there to make sure we don't load the user if the user is already loaded. Because, obviously, that'll already exist as otherwise it would keep loading the user everytime they logged on.

xPaw:
Yeah of course, as you can auth multiple times in LDAP. This doesn't change that.

Issue #872, opened 2017-01-06
Display the away status in the user list sidebar


MiniDigger:
if someone is set to away, his name should be displayed in gray or something like that to mark that he is not available right now. 
hexchat does it like this:
![http://i.minidigger.me/2017/01/chrome_07_00-51-32.png](http://i.minidigger.me/2017/01/chrome_07_00-51-32.png)

MaxLeiter:
@xpaw had the idea to dim names in the user list (or we could add another user list section like Modes currently have)

astorije:
I like this idea, but I'm not in favor of dimming names because we are starting to consider it for everything: when hovering on a nick, in #815, now this, etc.

Instead, I'd rather stick something next to the nick, be it http://fontawesome.io/icon/clock-o/ or something like @MiniDigger's screenshot, either same color than nick or green/empty similar to Slack:

<img width="39" alt="screen shot 2017-01-07 at 23 42 01" src="https://cloud.githubusercontent.com/assets/113730/21747392/485b1632-d533-11e6-98fc-8235eebd8ef0.png">

ElrondMcBong:
This is reallly a missing feature in my current implementation. I like the idea of @MaxLeiter to move away-users to another user list section.

MiniDigger:
(fixed the original image) 

I like the icon (either clock or just a colored circle) @astorije suggested, maybe we could even change the sorting to display active users over users who are marked as away.

dgw:
The icon would be plenty on its own. No need to make finding a user more difficult by further subdividing the list. (Anecdotally, I've never used an IRC client that used a separate section for away users.)

FingerlessGlov3s:
Any news on this?

McInkay:
> Any news on this?

Nope. If someone ends up doing it, they'll post here to say so. Feel free to pick it up yourself :-)

FingerlessGlov3s:
> Nope. If someone ends up doing it, they'll post here to say so. Feel free to pick it up yourself :-)

I did start looking at the code, need to make sense of it all first :-)

Is all the server code including the calls to the websocket done code in the src folder?
I then presume all the clientside code in client folder?

MiniDigger:
if you plan on working on this and need help with the code, I suggest joining IRC for discussion, way easier to explain stuff there :)

FingerlessGlov3s:
> if you plan on working on this and need help with the code, I suggest joining IRC for discussion, way easier to explain stuff there :)

[...]

PR #3494, opened 2019-11-02
Remove away messages from channels

### LINKED ISSUE #3026: Away status will appear in all channels you share with a user when opening a query
* *Node version:* 10.15.1
* *Browser version:* Firefox 64
* *Device, operating system:* macOS 10.14.2
* *The Lounge version:* 3.0.0

---

When sending a private message to a user and they are set as away, thelounge will output their away status across all chatrooms the users share on a given network. I think the intended behaviour for this feature is that the notice will only show up in the corresponding query window.

xPaw:
Fixes #3026

It's unreasonably spammy with away-notify. We should implement #872 instead

### DIFF
--- a/client/components/MessageCondensed.vue
+++ b/client/components/MessageCondensed.vue
@@ -55,22 +55,6 @@ export default {
 			constants.condensedTypes.forEach((type) => {
 				if (obj[type]) {
 					switch (type) {
-						case "away":
-							strings.push(
-								obj[type] +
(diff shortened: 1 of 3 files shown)


fuzzy76:
If I understand correctly, releasing this before #872 would leave users without any easy way to figure out who's away and not?

xPaw:
In channels? Yes, but that's not different to how it was before, as this only removes away-notify messages.

PR #3498, opened 2019-11-06
Print package versions on startup

MiniDigger:
I opted for storing the version in the packageInfo, since thats passed to public client and might be of use later somewhere else.

### DIFF
--- a/src/plugins/packages/index.js
+++ b/src/plugins/packages/index.js
@@ -97,8 +97,10 @@ function loadPackages() {
 			return;
 		}
 
+		const version = packageInfo.version;
 		packageInfo = packageInfo.thelounge;
 		packageInfo.packageName = packageName;
+		packageInfo.version = version;
 
 		packageMap.set(packageName, packageFile);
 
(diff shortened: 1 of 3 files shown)


src/plugins/packages/index.js
```diff
@@ -97,8 +97,10 @@ function loadPackages() {
 			return;
 		}
 
+		const version = packageInfo.version;
 		packageInfo = packageInfo.thelounge;
 		packageInfo.packageName = packageName;
+		packageInfo.version = version;
```
xPaw: You read version from `packageInfo.version` already, I don't think this or line 100 are needed?

MiniDigger: as noted in the PR description, I saved the version into the packageInfo since then others get access to it via the public client.
Not something that is 100% needed now, but I liked the idea of having the version next to the name in that packageInfo object (what you define under "thelounge" in your package.json)

xPaw: Ah whoops, I missed the `packageInfo = packageInfo.thelounge;`.

src/plugins/packages/index.js
```diff
@@ -116,7 +118,9 @@ function loadPackages() {
 			packageFile.onServerStart(packageApis(packageInfo));
 		}
 
-		log.info(`Package ${colors.bold(packageName)} loaded`);
+		log.info(
+			`Package ${colors.bold(packageName)} ${colors.green("v" + packageInfo.version)} loaded`
```
xPaw: Use local `version` as you set it above anyway?

(also squash pls)

PR #3499, opened 2019-11-07
Proposed fix to thelounge#3495

Samuel-Chris:
Set unread counter for all channels if tab is visible

-The modification includes a check if the tab is visible using the document.hidden API, if not visible unread counters will be set whether or not the channel is active.

This commit resolves thelounge#3495

thelounge#3495

### DIFF
--- a/client/js/socket-events/msg.js
+++ b/client/js/socket-events/msg.js
@@ -40,7 +40,8 @@ socket.on("msg", function(data) {
 		channel = vueApp.activeChannel.channel;
 
 		data.chan = channel.id;
-	} else if (!isActiveChannel) {
+		//Set unread counter for all channels if tab is inactive
+	} else if (!isActiveChannel || document.hidden) {
 		// Do not set unread counter for channel if it is currently active on this client
 		// It may increase on the server before it processes channel open event from this client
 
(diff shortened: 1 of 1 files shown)


xPaw:
That's not gonna work this way because open channels do not increase counters on server.

Samuel-Chris:
Hello @xPaw, what module/file handles the counter on the server?

xPaw:
We also check `hasFocus` here at the moment to trigger a notification (mind you it checks `msg.highlight` which is a boolean, not a counter).

https://github.com/thelounge/thelounge/blob/9ef5c6c67e463c1f401e33b21dfb5641636e5ed1/client/js/socket-events/msg.js#L95-L96

Vue branch currently rewrites a lot of code, so unfortunately changing client logic around highlight counts is not desired at the moment.

That being said, I do believe that this fix needs to be discussed in the issue first. How we should handle open clients but don't have the tab open (including mobiles).

This would also lead into the discussion of push notifications, whether we should push to other devices if you actually have a window open and front on any client.

xPaw:
Gonna close this PR, please leave your thoughts on how this should work in the issue (see comment above).

PR #3521, opened 2019-11-18
Create release github action workflow

xPaw:
npm tags inspired by https://github.com/roddolf/jsonld-sorter/blob/86816cfd1f572207576cbc1ef08c0190158d8610/.github/workflows/publish.yml#L22

This should keep our existing travis behaviour of automatically publishing/deleting `next` tag.

This deletes travis config, as at this point we only kept it around for publishing.


xPaw:
I tested the same workflow in a theme repo: https://github.com/thelounge/thelounge-theme-solarized/actions?query=workflow%3ARelease

I think we require a npm token without 2fa enabled on the account for this to work.

PR #3535, opened 2019-11-27
Fix keybinds when chat input is focused.

richrd:


xPaw:
For context, broken by fcf7488e1e7c04c34e6e49e94646aabfb4d8c252

Mousetrap requires a class to work as intended.

PR #3553, opened 2019-12-04
Refactor some CSS styling

richrd:
This PR makes the following changes:
- Replace `.chan` class in sidebar with `.channel-list-item`. This class can be used globally, and it isn't recommended to prefix it with `#sidebar` or any other selector, since these elements will be used outside the sidebar in future features.
- Remove the type class (`lobby`, `channel`, `query` and `special`) from `.channel-list-item` and `.chan` and instead use a data attribute e.g. `data-type="lobby"`. Use the following selector in themes: e.g. `.channel-list-item[data-type="lobby"]`.
- For the child element of the `#chat` container, remove classes `chan` and `active` and add class `chat-view`. Before: `.chan.channel`, now: `.chat-view[data-type="channel"]`
- Remove `#windows`
- `#upload-progressbar` is now part of `#form` so the content doesn't shift
- Added a CSS variable `--overlay-bg-color` for sidebar overlay, upload dragover overlay, and image viewer background color

### DIFF
--- a/client/components/ChannelWrapper.vue
+++ b/client/components/ChannelWrapper.vue
@@ -4,8 +4,7 @@
 		v-if="isChannelVisible"
 		ref="element"
 		:class="[
-			'chan',
-			channel.type,
+			'channel-list-item',
 			{active: activeChannel && channel === activeChannel.channel},
 			{'parted-channel': channel.type === 'channel' && channel.state === 0},
 			{'has-draft': channel.pendingMessage},
@@ -13,6 +12,7 @@
 		:aria-label="getAriaLabel()"
 		:title="getAriaLabel()"
 		:data-name="channel.name"
+		:data-type="channel.type"
 		:aria-controls="'#chan-' + channel.id"
 		:aria-selected="activeChannel && channel === activeChannel.channel"
 		:style="channel.closed ? {transition: 'none', opacity: 0.4} : null"
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -10,7 +10,8 @@
 		>
 			<div
 				:id="'chan-' + channel.id"
-				:class="[channel.type, 'chan', 'active']"
+				class="chat-view"
+				:data-type="channel.type"
 				:aria-label="channel.name"
 				role="tabpanel"
 			>
@@ -88,7 +89,6 @@
 		>
 			{{ this.$store.state.currentUserVisibleError }}
 		</div>
-		<span id="upload-progressbar" />
 		<ChatInput :network="network" :channel="channel" />
 	</div>
 </template>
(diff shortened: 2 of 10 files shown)


client/components/ChannelWrapper.vue
```diff
@@ -4,7 +4,7 @@
 		v-if="isChannelVisible"
 		ref="element"
 		:class="[
-			'chan',
+			'channel-list-item',
 			channel.type,
```
xPaw: Make this a data attribute as we discussed?

client/components/Chat.vue
```diff
@@ -1,16 +1,18 @@
...
+		id="chat-container"
+		:class="{
+			window: true,
+			'hide-motd': !$store.state.settings.motd,
+			'colored-nicks': $store.state.settings.coloredNicks,
+			'show-seconds': $store.state.settings.showSeconds,
+		}"
+		:data-current-channel="channel.name"
+	>
+		<div id="chat">
```
xPaw: Remove this div?

client/components/Chat.vue
```diff
@@ -1,16 +1,18 @@
...
 			<div
 				:id="'chan-' + channel.id"
-				:class="[channel.type, 'chan', 'active']"
+				:class="[channel.type, 'chan', 'chat-view']"
```
xPaw: make `channel.type` a data attribute?

richrd: Yeah, will do. `.chan` should no longer be needed after I'm done with this.

client/components/Chat.vue
richrd: I guess we have to keep it as getting rid of it proved tricky.

xPaw:
@MiniDigger can you play around with this and see if you spot anything odd.

client/components/Chat.vue
```diff
@@ -1,16 +1,18 @@
...
+			'colored-nicks': $store.state.settings.coloredNicks,
+			'show-seconds': $store.state.settings.showSeconds,
+		}"
+		:data-current-channel="channel.name"
+	>
+		<div id="chat">
 			<div
 				:id="'chan-' + channel.id"
-				:class="[channel.type, 'chan', 'active']"
+				:class="[channel.type, 'chat-view']"
```
xPaw: Still have to make the type a data attribute.

xPaw:
@richrd Can you split scrollbar vars into a separate PR please. I don't think we should change `scrollbar-color` in default theme.

richrd:
> @richrd Can you split scrollbar vars into a separate PR please. I don't think we should change `scrollbar-color` in default theme.
I'm not opposed to it, but the changes I made still use the same colors as before. I just moved them to the CSS variables and also set them on firefox for consistency. Why don't we want to do that?

xPaw:
> set them on firefox for consistency.

Will need to test that.

Anyway, it will just make it this PR easier to merge without any actual user facing changes.

PR #3594, opened 2019-12-16
Jump to channel switcher

### LINKED ISSUE #1454: "Jump to" channel switcher
It'd be nice to have a pop box that allows users to quickly switch to a DM or channel with a couple of characters, a la the "Jump to" modal in Slack. 

(In Slack, cmd + k on a Mac will pop up the modal, you can start typing the name of the channel and it will use fuzzy logic to match to a channel.)

richrd:
This adds a search box to the sidebar, that can be used to filter and navigate channels. It can be toggled with <kbd>alt</kbd>+<kbd>j</kbd>.

This is how it looks on the default theme:
image

Fixes #1454

### DIFF
--- a/client/components/Channel.vue
+++ b/client/components/Channel.vue
@@ -1,5 +1,5 @@
 <template>
-	<ChannelWrapper ref="wrapper" :network="network" :channel="channel">
+	<ChannelWrapper ref="wrapper" v-bind="$props">
 		<span class="name">{{ channel.name }}</span>
 		<span v-if="channel.unread" :class="{highlight: channel.highlight}" class="badge">{{
 			unreadCount
@@ -36,6 +36,8 @@ export default {
 	props: {
 		network: Object,
 		channel: Object,
+		active: Boolean,
+		isFiltering: Boolean,
 	},
 	computed: {
 		unreadCount() {
--- a/client/components/NetworkList.vue
+++ b/client/components/NetworkList.vue
@@ -117,6 +117,66 @@
 	</div>
 </template>
 
+<style>
+.jump-to-input {
+	margin: 8px;
+	position: relative;
+}
+
+.jump-to-input .input {
+	margin: 0;
+	width: 100%;
+	border: 0;
+	color: inherit;
+	background-color: rgba(255, 255, 255, 0.1);
+	padding-right: 35px;
+}
+
+.jump-to-input .input::placeholder {
+	color: rgba(255, 255, 255, 0.35);
+}
+
+.jump-to-input::before {
+	content: "\f002"; /* http://fontawesome.io/icon/search/ */
+	color: rgba(255, 255, 255, 0.35);
+	position: absolute;
+	right: 8px;
+	top: 0;
+	bottom: 0;
+	pointer-events: none;
+	line-height: 35px !important;
+}
+
+.jump-to-results {
+	margin: 0;
+	padding: 0;
+	list-style: none;
+	overflow: auto;
+}
+
+.jump-to-results .no-results {
+	margin: 14px 8px;
+	text-align: center;
+}
+
+.jump-to-results .channel-list-item.active {
+	cursor: pointer;
+}
+
+.jump-to-results .channel-list-item .add-channel,
+.jump-to-results .channel-list-item .close-tooltip {
+	display: none;
+}
+
+.jump-to-results .channel-list-item[data-type="lobby"] {
+	padding: 8px 14px;
+}
+
+.jump-to-results .channel-list-item[data-type="lobby"]::before {
+	content: "\f233";
+}
+</style>
+
 <script>
 import Mousetrap from "mousetrap";
 import Draggable from "vuedraggable";
@@ -181,18 +241,14 @@ export default {
 		},
 	},
 	mounted() {
-		Mousetrap.bind("alt+j", this.toggleSearch);
-	},
-	destroyed() {
-		Mousetrap.unbind("alt+j", this.toggleSearch);
-	},
-	mounted() {
 		Mousetrap.bind("alt+shift+right", this.expandNetwork);
 		Mousetrap.bind("alt+shift+left", this.collapseNetwork);
+		Mousetrap.bind("alt+j", this.toggleSearch);
 	},
 	beforeDestroy() {
 		Mousetrap.unbind("alt+shift+right", this.expandNetwork);
 		Mousetrap.unbind("alt+shift+left", this.collapseNetwork);
+		Mousetrap.unbind("alt+j", this.toggleSearch);
 	},
 	methods: {
 		expandNetwork() {
--- a/client/components/NetworkList.vue
+++ b/client/components/NetworkList.vue
@@ -1,5 +1,5 @@
 <template>
-	<div v-if="networks.length === 0" class="empty">
+	<div v-if="$store.state.networks.length === 0" class="empty">
 		You are not connected to any networks yet.
 	</div>
 	<div v-else ref="networklist">
@@ -52,7 +52,7 @@
 		</div>
 		<Draggable
 			v-else
-			:list="networks"
+			:list="$store.state.networks"
 			:filter="isCurrentlyInTouch"
 			:prevent-on-filter="false"
 			handle=".channel-list-item[data-type='lobby']"
@@ -66,7 +66,7 @@
 			@end="onDragEnd"
 		>
 			<div
-				v-for="network in networks"
+				v-for="network in $store.state.networks"
 				:id="'network-' + network.uuid"
 				:key="network.uuid"
 				:class="{
@@ -80,7 +80,10 @@
 				<NetworkLobby
 					:network="network"
 					:is-join-channel-shown="network.isJoinChannelShown"
-					:active="activeChannel && network.channels[0] === activeChannel.channel"
+					:active="
+						$store.state.activeChannel &&
+							network.channels[0] === $store.state.activeChannel.channel
+					"
 					@toggleJoinChannel="network.isJoinChannelShown = !network.isJoinChannelShown"
 				/>
 				<JoinChannel
@@ -109,7 +112,10 @@
 						:key="channel.id"
 						:channel="channel"
 						:network="network"
-						:active="activeChannel && channel === activeChannel.channel"
+						:active="
+							$store.state.activeChannel &&
+								channel === $store.state.activeChannel.channel
+						"
 					/>
 				</Draggable>
 			</div>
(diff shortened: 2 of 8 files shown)


client/components/NetworkList.vue
```diff
@@ -82,10 +135,61 @@ export default {
...
+			for (const network of this.$store.state.networks) {
+				for (const channel of network.channels) {
+					if (
+						this.$store.state.activeChannel &&
+						channel === this.$store.state.activeChannel.channel
+					) {
+						continue;
+					}
+
+					items.push([network, channel]);
```
xPaw: Use an object

client/components/NetworkList.vue
```diff
@@ -82,10 +135,61 @@ export default {
...
+				}
+			}
+
+			// Sort by most unreads
+			items.sort((a, b) => {
+				return b[1].unread - a[1].unread;
```
xPaw: Stabilize the sort by name when unreads equal

richrd: As discussed in IRC I removed sorting so that we have the same predictable order as in the normal sidebar list.

client/components/NetworkList.vue
richrd: Done

richrd:
I would say this pretty much ready as it is, except in an ideal world we'd use <kbd>ctrl</kbd>+<kbd>k</kbd> for the shoctcut. That's what slack, discord, mattermost and others use for this feature.
Problem is that <kbd>ctrl</kbd>+<kbd>k</kbd> is used for color selector, do we want to change that or just go with <kbd>alt</kbd><kbd>j</kbd> or some other less standard shortcut?

Mikaela:
Do you have any idea how much TheLounge users use colours or formatting? I think IRC users mostly don't use colours and even people using them may be open to changing to another shortcut as they would likely be used to `ctrl`+`k` of the other apps.

dgw:
Any rebinding of Ctrl-K away from colors should be _very_ carefully considered. It's the de facto standard color shortcut.

Mikaela:
Where is it the de-facto standard color shortcut? I cannot remember coming across it in other context than channel switching.

dgw:
Granted it's an unscientific sampling, but every client I've ever installed used Ctrl/Cmd+K for colors (mIRC, KVIrc, Textual, TL). Search for "irc color ctrl-k" and see for yourself how much newbie documentation references it, e.g. https://en.wikichip.org/wiki/irc/colors and (currently not loading for me) http://www.ircbeginner.com/ircinfo/colors.html

client/components/NetworkList.vue
```diff
@@ -123,6 +223,110 @@ export default {
...
+			}
+
+			this.activateSearch();
+		},
+		activateSearch() {
+			if (this.$refs.searchInput === document.activeElement) {
+				return;
+			}
+
+			this.sidebarWasOpen = this.$store.state.sidebarOpen;
```
xPaw: This is only set when using the keybind, not when simply focusing in already open sidebar.

Perhaps instead set this to true to mean "sidebar was closed, and we opened it just now", that way it's simple to track correctly.

client/components/NetworkList.vue
```diff
@@ -123,6 +223,110 @@ export default {
...
+			this.$store.commit("sidebarOpen", true);
+			this.$nextTick(() => {
+				this.$refs.searchInput.focus();
```
xPaw: Add `scrollToActive` so it scrolls after clicking an entry in search.

client/css/style.css
```diff
@@ -2792,3 +2790,61 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 #chat table.channel-list .topic {
 	white-space: pre-wrap;
 }
+
+/* Jump to feature WIP styles */
```
xPaw: The whitespace hack was deliberately the last style in css.

We should use `@import` already :)

client/components/NetworkList.vue
richrd: shouldn't that be done in `selectResult`?

client/css/style.css
richrd: Will move the styles to the component itself :+1:

client/components/NetworkList.vue
richrd: So only set it to true if `this.$store.state.sidebarOpen` is false?

xPaw: Should be.

client/components/NetworkList.vue
```diff
@@ -2,71 +2,185 @@
...
+	margin: 8px;
+	position: relative;
+}
+
+.jump-to-input .input {
+	margin: 0;
+	width: 100%;
+	border: 0;
+	color: inherit;
+	background-color: rgba(255, 255, 255, 0.1);
```
xPaw: Add `color: #fff;` so the typed text contrast is good.

xPaw:
Looks like our component style compiling puts `jump-to` styles above our `style.css` which is probably not a good idea. Can we tweak the order somehow?

Besides that, this PR seems good to go for me.

client/components/NetworkList.vue
```diff
@@ -143,6 +298,111 @@ export default {
 				order: channel.network.channels.map((c) => c.id),
 			});
 		},
+		toggleSearch(event) {
+			event.preventDefault();
```
xPaw: change this to `return false`, mousetrap can do it.

and add a `textarea` check like in https://github.com/thelounge/thelounge/pull/3720

client/components/NetworkList.vue
richrd: This should be fixed now.

client/components/NetworkList.vue
richrd: This should be fixed now.

client/components/NetworkList.vue
richrd: This should be fixed now.

client/components/NetworkList.vue
```diff
@@ -2,71 +2,185 @@
...
+		<div class="jump-to-input">
+			<input
+				ref="searchInput"
+				:value="searchText"
+				placeholder="Jump to..."
+				type="search"
+				class="search input mousetrap"
+				aria-label="Search among the channel list"
+				tabindex="-1"
+				@input="setSearchText"
```
xPaw: Why are you using this instead of `v-model`?

client/components/NetworkList.vue
```diff
@@ -143,6 +298,116 @@ export default {
...
+			}
+
+			if (!channel) {
+				channel = this.results[0].channel;
+			}
+
+			this.activeSearchItem = channel;
+		},
+		selectResult() {
+			if (!this.searchText || !this.results.length) {
```
xPaw: So there's a race condition with selecting a channel, and `activeChannel` changes before `selectResult` is called, which results in the search input not being cleared.

`items()` is recomputed, so the channel you clicked on is excluded due to being active and this `if` returns early.

client/components/NetworkList.vue
```diff
@@ -2,71 +2,185 @@
...
-				@change="onChannelSort"
-				@start="onDragStart"
-				@end="onDragEnd"
+		</div>
+		<div v-if="searchText" class="jump-to-results">
+			<div v-if="results.length">
+				<div
+					v-for="item in results"
+					:key="item.channel.id"
+					v-on="{mouseover: () => setActiveSearchItem(item.channel)}"
```
xPaw:
```suggestion
					@mouseenter="setActiveSearchItem(item.channel)"
```

client/components/Channel.vue
```diff
@@ -36,6 +36,8 @@ export default {
 	props: {
 		network: Object,
 		channel: Object,
+		active: Boolean,
+		isFiltering: Boolean,
```
xPaw: Is `isFiltering` never used?

client/components/NetworkList.vue
```diff
@@ -2,71 +2,185 @@
...
+						:active="activeChannel && channel === activeChannel.channel"
+					/>
+				</Draggable>
+			</div>
+		</Draggable>
+	</div>
 </template>
 
+<style>
+.jump-to-input {
```
xPaw: If we tweak the background to be non transparent on the input, we can make this sticky. What do you reckon?

```css
	position: sticky;
	top: 8px;
	z-index: 1;
```

client/components/NetworkList.vue
```diff
@@ -82,18 +196,59 @@ export default {
...
 		Draggable,
 	},
+	data() {
+		return {
+			searchText: "",
+			activeSearchItem: null,
+		};
+	},
 	computed: {
+		activeChannel() {
```
xPaw: Remove this.

client/components/NetworkList.vue
```diff
@@ -82,18 +196,59 @@ export default {
...
+			return this.$store.state.activeChannel;
+		},
 		networks() {
```
xPaw: I'd remove this too, so we access the state directly.

client/components/NetworkList.vue
richrd: I think we needed it to get the event on each keypress due to the vue issue with composition https://github.com/vuejs/vue/issues/8723.

xPaw: Oh oops. I changed it in https://github.com/thelounge/thelounge/pull/3730 and it seemed to work. Will need to test more closely.

client/components/NetworkList.vue
```diff
@@ -143,6 +298,116 @@ export default {
...
+			}
+
+			this.activeSearchItem = channel;
+		},
+		selectResult() {
+			if (!this.searchText || !this.results.length) {
+				return;
+			}
+
+			this.$root.switchToChannel(this.activeSearchItem);
```
xPaw: The switch is handled by click in `ChannelWrapper`, this shouldn't be needed I think.

This relates to the race condition I mentioned above.

client/components/Channel.vue
richrd: Yeah, passed down from `NetworkList.vue` with `:is-filtering="true"` so a bit hard to spot with find or grep etc.

client/components/NetworkList.vue
richrd: This seems to work without calling `switchToChannel` only when actually clicking the result. Not when hitting enter in the search box. Wonder what the best approach is here..

client/components/NetworkList.vue
richrd: Not sure, this seems to cause issues with the magnifying glass icon positioning. Maybe we can test this out later and do it if it seems nice?

xPaw: Okay.

PR #3595, opened 2019-12-16
Adjust sizeUnits names to correspond with LinkPreviews

fnutt:
Adjusting sizeUnits names to correspond with client/components/LinkPreview.vue#L103, as they are not consistent with each others. Also would argue KB, MB, GB and TB are more familiar for most people.

KB, MB, GB and TB, instead of KiB, MiB, GiB and TiB.

Example:
image

### DIFF
--- a/client/js/constants.js
+++ b/client/js/constants.js
@@ -27,7 +27,7 @@ const timeFormats = {
 	msgWithSeconds: "HH:mm:ss",
 };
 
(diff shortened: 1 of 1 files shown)


xPaw:
Kibibytes is the correct suffix to use here, as we divide by 1024, not 1000.

I however didn't realize we put a separate size calculator in the link preview component, we should use the existing helper for that.

I'll make a PR for it.

xPaw:
I've made #3596 and added you as co-author there.

PR #3596, opened 2019-12-16
Use `friendlysize` helper consistently

xPaw:
Pointed out by @fnutt in #3595

### DIFF
--- a/test/client/js/helpers/friendlysizeTest.js
+++ b/test/client/js/helpers/friendlysizeTest.js
@@ -4,15 +4,26 @@ const expect = require("chai").expect;
 const friendlysize = require("../../../../client/js/helpers/friendlysize").default;
 
 describe("friendlysize helper", function() {
-	it("should render big values in human-readable version", function() {
-		expect(friendlysize(51200)).to.equal("50 KB");
+	it("should render human-readable version", function() {
+		expect(friendlysize(51200)).to.equal("50 KiB");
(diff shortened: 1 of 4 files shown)


test/client/js/helpers/friendlysizeTest.js
```diff
@@ -5,14 +5,23 @@ const friendlysize = require("../../../../client/js/helpers/friendlysize").defau
...
 
 	it("should round with 1 digit", function() {
-		expect(friendlysize(1234567)).to.equal("1.2 MB");
+		expect(friendlysize(1234567)).to.equal("1.2 MiB");
 	});
 
 	it("should render special case 0 as 0 Bytes", function() {
-		expect(friendlysize(0)).to.equal("0 Bytes");
+		expect(friendlysize(0)).to.equal("0 B");
+	});
```
NeatNit: From the title "it should render **special case** 0 as 0 Bytes" it sounds like it was very deliberate and it shouldn't change.

xPaw: It's a special case because Math.log(0) doesn't work, there's a `size>0` in the code for that.

NeatNit: Oh.

As a user though, I think "Bytes" is probably clearer than "B", so perhaps that shouldn't change.

xPaw: That is a good point, I will change that.

PR #3606, opened 2019-12-19
Add preview for text/plain urls

xPaw:
If url is `text/plain`, it will use the content as the description. This also adds truncation to head/body for html urls.

19-1143-GenerousBlueandgoldmackaw


richrd:
How much of the content will be displayed when More is clicked?

xPaw:
I limited it to 300 characters.

PR #3608, opened 2019-12-19
Relocate not-secure and not-connected classes and make sure channel exists in jumpToChannel

richrd:


client/css/style.css
```diff
@@ -729,22 +729,22 @@ background on hover (unless active) */
 	margin: 0 8px;
 }
 
-.network.not-secure .channel-list-item[data-type="lobby"] {
+.channel-list-item.not-secure[data-type="lobby"] {
```
xPaw: `[data-type="lobby"]` is no longer necessary.

PR #3658, opened 2020-01-02
Display icon when update is available, check on server start

### LINKED ISSUE #2758: Add "update" indicator on help icon
### Feature Description

We currently show if there are lounge updates that the user hasn't installed on the help screen, however you have to go to the help screen to do that. It would be nice if we were to add an indicator to the icon so that you get an indication when you are just going about normally.

xPaw:
02-1052-TediousArachnid
02-1052-IncomparableGemsbok

Fixes #2758

### DIFF
--- a/client/components/Sidebar.vue
+++ b/client/components/Sidebar.vue
@@ -52,12 +52,22 @@
 					aria-controls="settings"
 					:aria-selected="$route.name === 'Settings'"
 			/></span>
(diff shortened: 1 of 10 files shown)


xPaw:
Probably need to add a daily timer to check updates, because right now the icon will show up if there's an update on server start, or someone visits the help page.

xPaw:
I added a timer to check every day for updates, and if one becomes available it will also notify all currently connected clients.

PR #3661, opened 2020-01-03
Pipe uploaded files through exif-be-gone

### LINKED ISSUE #3025: File uploader should remove EXIF data from images (especially location data)
The file uploader is working nicely. On mobile it even gives the option to capture a picture and upload it directly. Unfortunately, unlike with image hosts like imgur, the EXIF location data is not removed from uploaded images, which can lead to unsuspecting and unknowing users accidentally revealing their location.

It would be nice if the automatic removal of EXIF data from images when uploading would be an option (which should also be the default behavior in my opinion).

xPaw:
Fixes #3025

I'm not entirely sure how safe the `exif-be-gone` is. https://github.com/joshbuddy/exif-be-gone/blob/master/index.js

This will most likely also remove wanted data like rotation and color profiles.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -154,6 +154,8 @@ module.exports = {
 	//   this limit will be prompted with an error message in their browser. A value of
 	//   `-1` disables the file size limit and allows files of any size. **Use at
 	//   your own risk.** This value is set to `10240` kilobytes by default.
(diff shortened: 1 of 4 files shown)


xPaw:
Another approach would be stripping it on the client, for example by rendering it into a canvas first.

PR #3664, opened 2020-01-03
Message search

### LINKED ISSUE #109: Message search

richrd:
This is mainly similar to https://github.com/thelounge/thelounge/pull/2627 except it's all on vue.

What this does:
- Adds a search field to the channel and query header bars (on mobile a search icon is shown, clicking it opens the search field as a dropdown)
- Pressing enter in the field runs the search and navigates the user to the search results page (the url contains the search terms)
- The search result page allows the search to be changed and run again
- The results show max 100 messages at once and older messages can be loaded with the "Show older messages" button
- The backend db search is just a simple `LIKE %keyword%`. This might require some improvement to get more relevant results

**TODO:**
- [ ] Channels with upper case letters in the name cannot be searched

Things to improve after this MVP:
- Highlight the search keyword in each search result message
- Append new results when loading more messages instead of replacing the old results
- Ability to jump back in history to a specific message by clicking on a search result


client/components/Windows/SearchResults.vue
```diff
@@ -0,0 +1,192 @@
...
+							aria-relevant="additions"
+						>
+							<template v-for="(message, id) in messages">
+								<DateMarker
+									v-if="shouldDisplayDateMarker(message, id)"
+									:key="message.time + '-date'"
+									:message="message"
+								/>
+								<Message
+									:key="message.time"
```
xPaw: I had this happen:
> [Vue warn]: Duplicate keys detected: '2019-12-27T18:15:45.232Z'. This may cause an update error.

xPaw: Also, context menus need fixing as they rely on `activeChannel` and that breaks context menus in search:

```
vue.js:105 TypeError: Cannot read property 'network' of null
    at VueComponent.openUserContextMenu 
```

client/components/Windows/SearchResults.vue
```diff
@@ -0,0 +1,192 @@
...
+						</div>
+						<div
+							v-else
+							class="messages"
+							role="log"
+							aria-live="polite"
+							aria-relevant="additions"
+						>
+							<template v-for="(message, id) in messages">
+								<DateMarker
```
xPaw: This might be too noisy, I tried searching something and almost half of the results were date markers.

client/components/Windows/SearchResults.vue
```diff
@@ -0,0 +1,192 @@
...
+	},
+	data() {
+		return {
+			offset: 0,
+			moreResultsAvailable: false,
+		};
+	},
+	computed: {
+		search() {
+			return this.$store.state.messageSearchResults;
```
xPaw: Are we able to store these inside of this component instead of the global state?

client/components/Windows/SearchResults.vue
```diff
@@ -0,0 +1,192 @@
...
+		},
+		messages() {
+			if (!this.search) {
+				return [];
+			}
+
+			return this.search.results.slice().reverse();
```
xPaw: Is this sorting so that recent messages are at the bottom?

I think search should operate in the way that recent messages are at the top, and going down they're older. So the load button should be moved to the bottom, and new results are just appended.

src/plugins/messageStorage/sqlite.js
```diff
@@ -188,9 +188,52 @@ class MessageStorage {
...
+			'SELECT msg, type, time, channel FROM messages WHERE type = "message" AND network = ? AND json_extract(msg, "$.text") LIKE ?';
+		const params = [query.networkUuid, `%${query.searchTerm}%`];
+
+		if (query.channelName) {
+			select += " AND channel = ? ";
+			params.push(query.channelName);
+		}
+
+		select += " ORDER BY time DESC LIMIT 100 OFFSET ? ";
+		params.push(query.offset || 0);
```
xPaw: Add `query.offset = parseInt(query.offset, 10);`

src/server.js
```diff
@@ -567,6 +567,12 @@ function initializeClient(socket, client, token, lastMessage, openChannel) {
 
 	socket.on("sessions:get", sendSessionList);
 
+	socket.on("search", (query) => {
```
xPaw: Move this to the `public` if check below.

client/components/MessageSearchForm.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+	<form :class="['message-search', {opened: searchOpened}]" @submit.prevent="searchMessages">
+		<div class="input-wrapper">
+			<input
+				ref="searchInputField"
+				type="text"
+				name="search"
+				class="input"
+				placeholder="Search messages..."
```
xPaw:
```suggestion
				placeholder="Search messages…"
```

client/components/Windows/SearchResults.vue
```diff
@@ -0,0 +1,192 @@
...
+								class="btn"
+								@click="onShowMoreClick"
+							>
+								<span v-if="$store.state.messageSearchInProgress">Loading…</span>
+								<span v-else>Show older messages</span>
+							</button>
+						</div>
+
+						<div v-if="$store.state.messageSearchInProgress" class="search-status">
+							Searching...
```
xPaw:
```suggestion
							Searching…
```

src/client.js
```diff
@@ -500,6 +500,16 @@ Client.prototype.more = function(data) {
 	};
 };
 
+Client.prototype.search = function(query) {
+	if (this.messageStorage) {
+		for (const storage of this.messageStorage) {
+			if (storage.database) {
```
xPaw: Storage is supposed to be opaque to TL, for example load message has a `canProvideMessages` helper.

https://github.com/thelounge/thelounge/blob/6f7444dfe38f71310bd22147f431e1273182e865/src/models/chan.js#L243

xPaw (changes requested):
We do need some extra UI to allow searching all channels on network, or even all networks.

Thinking about it in the scope of this PR because it may require a different archicture before we move further.

src/plugins/messageStorage/sqlite.js
richrd: What is that for? offset should already be a number.

xPaw: It comes from socket, not guaranted to be.

client/components/Windows/SearchResults.vue
richrd: I don't have a strong opinion on this. I did the "newest at the bottom" order since that's what we do in everywhere else. The opposite order will probably require some tweaking of the date markers and such (if we decide to keep them),

src/client.js
```diff
@@ -501,6 +501,11 @@ Client.prototype.more = function(data) {
 	};
 };
 
+Client.prototype.search = function(query) {
+	const messageStorage = this.messageStorage.find((s) => s.canProvideMessages());
+	return messageStorage.search(query);
```
xPaw: Need to handle the case when sqlite is disabled.

src/plugins/messageStorage/sqlite.js
```diff
@@ -188,9 +188,65 @@ class MessageStorage {
...
+		if (!this.isEnabled) {
+			return Promise.resolve([]);
+		}
+
+		let select =
+			'SELECT msg, type, time, channel FROM messages WHERE type = "message" AND (json_extract(msg, "$.text") LIKE ?';
+		const params = [`%${query.searchTerm}%`];
+
+		if (query.searchNicks) {
+			select += ' OR json_extract(msg, "$.from.nick") LIKE ?)';
```
xPaw: Shouldn't this be `AND` if you want to search that a user has said?

xPaw:
1. Clicking on a username or channel throws `Cannot destructure property 'network' of 'this.$store.state.activeChannel' as it is null.`
2. I think searching for something else that returns same message as currently visible throws this warning: `uplicate keys detected: '2020-03-18T01:16:03.495Z'. This may cause an update error.`
3. Is 'show more messages' supposed to replace current messages instead of prepending to the top?
4. The search box on default theme stands out and doesn't match any other existing fields color wise.
5. Pagination is off by one and includes 1 message from previous page.

client/components/Windows/SearchResults.vue
richrd: I'm getting duplicate key errors just by running a single search. There's no guarantee that the dates are unique and I indeed have two messages with the exact same date in my history. I don't know what to do about that since we don't have unique message IDs :(

richrd:
>     1. Clicking on a username or channel throws `Cannot destructure property 'network' of 'this.$store.state.activeChannel' as it is null.`
> 
>     2. I think searching for something else that returns same message as currently visible throws this warning: `Duplicate keys detected: '2020-03-18T01:16:03.495Z'. This may cause an update error.`
> 
>     3. Is 'show more messages' supposed to replace current messages instead of prepending to the top?
> 
>     4. The search box on default theme stands out and doesn't match any other existing fields color wise.
> 
>     5. Pagination is off by one and includes 1 message from previous page.

Points 1, 2, 3 and 5 should be fixed now. Clicking show more adds messages to the top and keeps scroll position. Point 2 still needs testing but seems to work ok for me.
I'm not yet sure of the best way to style the search box on default theme.

xPaw:
1. Do we want the 'show more messages' to automatically load when scrolling up? It doesn't at the moment.
2. For the sake of styling, should we also use the search icon on desktop? That way the search input won't stand out, and when focused we can make it stand out with styles.
3. Clicking a channel in message produces this:
```
vue.js:108 TypeError: Cannot read property 'network' of null
    at Object.<anonymous> (store.js:161)
    at VueComponent.onClick (InlineChannel.vue:17)
```

I think because there is no 'activeChannel'.

client/components/Windows/SearchResults.vue
xPaw: Have you looked whether we can keep the search state contained in this component without global state?

client/js/autocompletion-custom.js
```diff
@@ -0,0 +1,340 @@
+/*
```
xPaw, line 1: remove this file

client/js/helpers/contextMenu.js
```diff
@@ -170,10 +170,12 @@ export function generateChannelContextMenu($root, channel, network) {
 }
 
 export function generateUserContextMenu($root, channel, network, user) {
-	const currentChannelUser = channel.users.find((u) => u.nick === network.nick) || {};
+	const currentChannelUser = channel
```
xPaw, line 173: `(channel && channel.users.find((u) => u.nick === network.nick)) || {}` is slightly cleaner

src/client.js
xPaw: We still need some kind of handling (and probably hiding the search UI) if sqlite is disabled OR if user has logs disabled.

fnutt:
> 1. Do we want the 'show more messages' to automatically load when scrolling up? It doesn't at the moment.
> 2. For the sake of styling, should we also use the search icon on desktop? That way the search input won't stand out, and when focused we can make it stand out with styles.

Yes to both 1. and 2 👍. It would be more consistent with the rest of the client. Please also add this to the context menus so it is easy to find (e.g. I assume it's normal behaviour to right click a channel to search in it).

fnutt:
First of all I would like to say this is still working very nicely, and even better than last time I tested it. However there's a few things I would like to mention:

(1) Can the search result make a "new channel", similar to ban list? For me it takes a bit of time to search for a word, and if I have to step out of the search result for a second I have to re-search to see the result again. Can be a bit time consuming.

(2) Would also very much like for the search area at upper right corner to be a search button similar to mentions, as well as search being implemented into context menus for left sidebar when right clicking a channel.

(3) #thelounge @ Freenode 23.07.2020: "<cottongin> it'd also be cool if you could /search <search text>"

cottongin:
Some suggestions/comments:

1. Can you add a way to search directly from the input form? e.g. `/search query`
2. I also like the suggestion to make the desktop client-view utilize the search icon and overflow into a separate window like `@` already does for mentions. Though I have no strong feelings about it.
3. I feel like search results should open into their own window/view like other similar features (`/banlist`, `/ignorelist`, etc). As it stands now it is not immediately obvious how to "get back" to the channel you searched from.
4. It seems channel activity messages (joins/parts/quits/topic/modes) are ignored when searching, can an option be added to include these? This would be useful for searching for who set a ban in a channel, for example.
5. Similar to 4, a context option would be nice (like `grep -C 5`, for example, to also include additional context surrounding a search query.

vvug:
Thanks for this, it's great! Will it land in time for 4.3.0?

Mstrodl:
@richrd Is there anything specific needed (help-wise, I mean)? From what I saw from the comments above, it seems to be working well for other people.

fnutt:
> @richrd Is there anything specific needed (help-wise, I mean)? From what I saw from the comments above, it seems to be working well for other people.

At least a rebase has to be done. "Channels with upper case letters in the name cannot be searched" seems to be on the TODO-list.

There was also some questions about how to filter the search, e.g. search just messages from user X, older than X etc. If this is added the PR may have to be redone regarding the implementation.

Personally I would love to see a rebase and get the channel upper case letters bug fixed, and then merged to master. If we want to add search filter and other stuff later, someone can redo the implementation if needed.

PR #3667, opened 2020-01-04
Allow dismissing unread marker

richrd:
Hovering the marker shows the tooltip, and clicking the marker resets the unread marker (hides it). Effectively this is a "mark as read" feature. See https://github.com/thelounge/thelounge/issues/3666
image

### DIFF
--- a/client/components/MessageList.vue
+++ b/client/components/MessageList.vue
@@ -29,7 +29,9 @@
 					:key="message.id + '-unread'"
 					class="unread-marker"
 				>
-					<span class="unread-marker-text" />
+					<span class="tooltipped tooltipped-n" aria-label="Dismiss">
+						<span class="unread-marker-text" @click="clearUnreadMarker" />
+					</span>
 				</div>
 
 				<MessageCondensed
@@ -332,6 +334,9 @@ export default {
 
 			const el = this.$refs.chat;
 			el.scrollTop = el.scrollHeight;
+		},
+		clearUnreadMarker() {
+			this.channel.firstUnread = this.channel.messages[this.channel.messages.length - 1].id;
 		},
 	},
 };
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1244,6 +1244,10 @@ textarea.input {
 	border-top: 1px solid var(--unread-marker-color);
 }
 
+#chat .unread-marker-text {
+	cursor: pointer;
+}
+
 #chat .unread-marker-text::before {
 	content: "New messages";
 	background-color: var(--window-bg-color);


fnutt:
This is a feature that I would very much like to not see in The Lounge.

(1) The feature is not consistent with the rest of the client. We do not hide any other messages when clicking on them (we use /ignore to hide chat messages).
(2) It is not very intuitive to click on the unread marker to remove it. As far as I know not many other client and chats have click-to-remove.
(3) Unread marker goes away when switching channel.

If we want more ways to remove unread marker, what about doing it other ways which doesn't require non-intuitive manual clicking? E.g. removing unread marker when the client window is not in focus anymore? Or when channel is in focus after x-minutes?

Also would like to mention that the maintainers earlier very clearly stated that The Lounge was not supposed to have a lot of small features to suit everyone. I find this feature to be a small feature to suit just a few people.

richrd:
(1) This is not a message. This is a marker that shows which messages are new. This doesn't hide messages, just hides the unread marker.
(2) Currently there is no intuitive way of clearing the marker. One needs to switch to another channel and back. This has the tooltip so you know what happens if you decide to click the marker.
(3) Yes, but can't be dismissed within channel.

I'm open to the other features you suggested but I don't see why we shouldn't allow dismissing the marker manually. This isn't something that gets in anyones way.

Mikaela:
Is it possible that clicking the message sends the read marker onto the bottom so if there are new messages before the user scrolls down, those will have the readmarker above them?

richrd:
It might be possible but might be hard to do or possibly a bit unreliable due to having to somehow detect which message is the last visible one.

xPaw:
> Is it possible that clicking the message sends the read marker onto the bottom so if there are new messages before the user scrolls down, those will have the readmarker above them?

This is exactly how it will work because our unread marker works by setting "last read message id", which this PR also does.

You can test that now by sending a message into a channel (it resets the unread), and receiving another message.

maxpoulin64:
I think it would be fine to set it to the last message, but I would like to see it synchronized with the server so the unread marker doesn't come back on other clients.

That would be in line with what Discord, Slack and others do when you click the "Mark as read" thing that marks the whole channel as read, without attempting to do any of the smarts of checking if you've scrolled past every single message.

I think that's plenty enough and achieve what most people want: clear the unread marker on command in a more intuitive way than switching channels. Nothing more, nothing less. New messages will bring the unread bar again as it would anyway, which is sufficient for the user to have another point of reference from the last time they clicked the unread bar.


> It might be possible but might be hard to do or possibly a bit unreliable due to having to somehow detect which message is the last visible one.

That can be done calling `getBoundingClientRect()` on the message DOM Element and comparing it with the bounding rectangle of the nearest `.chat`. If the message is within the rectangle of the scrollable box, then it is visible:

```javascript
function isMsgInView(msgNode) {
	const scrollRect = msgNode.parentElement.parentElement.getBoundingClientRect()
	const msgRect =  msgNode.getBoundingClientRect()
	// Checks that the bottom of the message is higher than the bottom-most pixel of the scrollable
	return msgRect.top + msgRect.height < scrollRect.top + scrollRect.height
}
```

But I don't think that's necessary here.

richrd:
Yeah, I agree with you. The use case of  "Mark entire channel as read" is what I was going for here, and that's what it does. So do you think this feature is worth it if I add the syncing across clients?

xPaw:
> So do you think this feature is worth it if I add the syncing across clients?

Sure, let's do that.

fnutt:
As I wrote in my comment earlier in this PR, I would much rather want to make find ways that it is removed automagically. OR, even better, remove the real issue - when the "New messages" UI element is shown.

<NeatNit> I would say the New Messages marker should only be added if:
<NeatNit> 1. channel is not the active channel (obviously)
<NeatNit> 2. The Lounge is not in focus
<NeatNit> 3. The Lounge is in focus and the user has been idle for >x minutes

I assume with the methods above regarding when the UI element is shown will solve most of the issues some people have with it.

PR #3670, opened 2020-01-05
Show an icon for "show in active" messages

xPaw:
05-1626-BurlywoodCavy

Hopefully this removes some of the hidden magic, so the user understands the behaviour of notices and errors.

### DIFF
--- a/client/components/Message.vue
+++ b/client/components/Message.vue
@@ -58,6 +58,12 @@
 				</template>
 			</span>
 			<span class="content" dir="auto">
+				<span
+					v-if="message.showInActive"
+					aria-label="This message was shown in your active channel"
+					class="msg-shown-in-active tooltipped tooltipped-e"
+					><span></span
+				></span>
 				<ParsedMessage :network="network" :message="message" />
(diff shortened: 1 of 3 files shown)


richrd:
Nice one. The icon feels a bit peculiar though. I took a look at what else is available and these might be some possible alternatives and reasoning on their symbolism:

- **asterisk**: generic icon for here's some info you might want to know about
    https://fontawesome.com/icons/asterisk?style=solid
- **hashtag**: refers to channels, in this case "this message was originally in a channel"
    https://fontawesome.com/icons/hashtag?style=solid
- **thumbtack**: these messages were somewhere else, but now they've been stuck here
    https://fontawesome.com/icons/thumbtack?style=solid

image

xPaw:
Can also add https://fontawesome.com/icons/eye?style=solid to the list.

fnutt:
I vote for thumbtack.

GewoonYorick:
I like the eye what @xPaw  suggested. 

It's about the visibility of the message, so a eye icon makes sense

PR #3709, opened 2020-01-18
Prefetch timeout and EXIF Orientation

SimplyLinn:
JPEG images can have Exif data deciding orientation, and the actual pixel-order of the image itself is not how you want to display it.

Browsers does not honor Exif metadata by themselves, and I got increasingly frustrated with the fact that most of the selfies I saw people posting in my IRC channels (mine included) got turned 90°

So I took the matter into my own hands and wrote my own functionality that ensures that Exif orientation metadata is honored during render.

Also made the prefetch timeout configurable, since my test images tended to be large photos straight from a camera and sometimes took longer than 5000ms to fetch


xPaw:
Thanks for the PR.

I want avoid parsing images on the server at all. This also ends up being quite a lot of code on the client too (and that's not even covering all the orientations afaik), and it will end up being even more when we implement exif stripping for uploaded images.

I would rather look into using `image-orientation: from-image` CSS property (or the attribute on img tag that should also be coming - https://github.com/whatwg/html/issues/4495). It's supported in Firefox, and looking at Chrome's bug tracker there's work happening around it.

SimplyLinn:
https://developer.mozilla.org/en-US/docs/Web/CSS/image-orientation

That one is likely to be deprecated, and there's little to no info on what will replace it, if anything.

xPaw:
`from-image` should be staying from the look of things. If it gets removed, it would be replaced by the `autorotate` attribute (see my link).

I just looked at chrome canary, and it actually started supporting exif orientations without even setting the `image-orientation` property.

SimplyLinn:
Also unsure how you'er gonna pull off this combination of things, will the client do the exif stripping before sending it to the server?
> I want avoid parsing images on the server at all. [...] when we implement exif stripping for uploaded images.

xPaw:
Looking at the CSSWG issues, it's certainly not going anywhere, and they updated the spec to use `from-image` as default not long ago - https://github.com/w3c/csswg-drafts/commit/bacb9b7b91d0dd6cc5fdc1e469570972dc8805aa

So we can simply set `image-orientation: from-image` to support Firefox now, and Chrome will work in a future update.

> Also unsure how you'er gonna pull off this combination of things, will the client do the exif stripping before sending it to the server?

Yeah, I wanted to strip it by drawing it into a canvas and taking the bitmap out. It works but it breaks the orientation, which would also be fixed in future browser updates - https://bugs.chromium.org/p/chromium/issues/detail?id=1042141

SimplyLinn:
🤷‍♀  I mainly made it for myself since I roll my own fork of thelounge. This turned out to be less of a hack that required manual edits in configuration files to work and I made a PR of it.

On another note, feel free to cherrypick [ab1f03b](https://github.com/thelounge/thelounge/pull/3709/commits/ab1f03b0bdc2c1a6f580a06910a5e4d66dc8bea1) if that's desierable whereas the rest isn't

xPaw:
Can you make a PR setting `image-orientation: from-image` so we at least support Firefox right now?

Chrome 81 should release in March, and then it will also respect the exif orientation. Any browser using Chromium (basically all of them, lol) would also follow.

SimplyLinn:
I'll get around to it once I've managed to get my deployed branch up to speed with this feature

PR #3731, opened 2020-01-29
Allow to provide list of ignored channels for logging

CirnoT:
Allows user to specify list of channels to be ignored by logging mechanism, per network.


xPaw:
We want to have per-channel settings eventually, and specifying whether you get notification, or log said channel fits very well there.

Unfortunately I'm not willing to add this feature in the way this PR adds it.

CirnoT:
Sure, I'll keep this in my fork that actually keeps usable features that work right now as opposed to imaginary functions that will happen never.

PR #3741, opened 2020-02-04
Write prettier error messages for certain errors

### LINKED ISSUE #251: Display nice error messages
As @astorije pointed out in #167, some errors are only printed as special strings, instead of something descriptive.

Here's how Kiwi handles these: https://github.com/prawnsalad/KiwiIRC/blob/6f90124e0f15d4cc373496110d1150ffd75072fc/client/src/models/network.js#L856-L917

### LINKED ISSUE #797: Add colons to unhandled messages first param

RockyTV:
Closes #251.
Closes #797.

[Inspired by KiwiIRC's implementation](https://github.com/prawnsalad/KiwiIRC/blob/master/client/src/models/network.js#L849), though there are some messages that Kiwi didn't include, and some that I skipped because they aren't used by Freenode.

I'd like to ask, is the current format good? `#[channel]: The channel is full`, or would it be better to display it as `#[channel] is full`, `You have been banned from #[channel]`? I.e., removing the line that appends the channel to the error message.

### DIFF
--- a/client/components/MessageTypes/error.vue
+++ b/client/components/MessageTypes/error.vue
@@ -1,0 +1,58 @@
+<template>
+	<span class="content">
+		<ParsedMessage :network="network" :message="message" :text="errorMessage" />
+	</span>
+</template>
+
+<script>
+import ParsedMessage from "../ParsedMessage.vue";
+
+export default {
+	name: "MessageTypeError",
+	components: {
+		ParsedMessage,
+	},
+	props: {
+		network: Object,
+		message: Object,
+	},
+	computed: {
+		errorMessage() {
+			switch (this.message.error) {
+				case "bad_channel_key":
+					return `Cannot join ${this.message.channel} - Bad channel key.`;
+				case "banned_from_channel":
+					return `Cannot join ${this.message.channel} - You have been banned from the channel.`;
+				case "cannot_send_to_channel":
+					return `Cannot send to channel ${this.message.channel}`;
+				case "channel_is_full":
+					return `Cannot join ${this.message.channel} - Channel is full.`;
+				case "chanop_privs_needed":
+					return "Cannot perform action: You're not a channel operator.";
+				case "invite_only_channel":
+					return `Cannot join ${this.message.channel} - Channel is invite only.`;
+				case "no_such_nick":
+					return `User ${this.message.nick} hasn't logged in or does not exist.`;
+				case "not_on_channel":
+					return "Cannot perform action: You're not on the channel.";
+				case "password_mismatch":
+					return "Password mismatch.";
+				case "too_many_channels":
+					return `Cannot join ${this.message.channel} - You've already reached the maximum number of channels allowed.`;
+				case "unknown_command":
+					return `Unknown command: ${this.message.command}`;
+				case "user_not_in_channel":
+					return `User ${this.message.nick} is not on the channel.`;
+				case "user_on_channel":
+					return `User ${this.message.nick} is already on the channel.`;
+				default:
+					if (this.message.reason) {
+						return `${this.message.reason} (${this.message.error})`;
+					}
+
+					return this.message.error;
+			}
+		},
+	},
+};
+</script>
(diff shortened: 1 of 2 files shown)


xPaw:
This should be all of them that need to be handled: https://github.com/kiwiirc/irc-framework/blob/master/src/commands/handlers/generics.js

RockyTV:
Added the missing errors to the list.

src/plugins/irc-events/error.js
```diff
@@ -13,12 +13,51 @@ module.exports = function(irc, network) {
...
+				if (data.reason) {
+					text += `${data.reason} (${data.error})`;
+				} else {
+					text += data.error;
+				}
+
+				break;
 		}
 
 		const msg = new Msg({
```
xPaw: As I mentioned on IRC, the text formatting should be moved to the client.

Make an `error` messagetype component so the text can be formatted there. Remove `text` from the `Msg` and put `data.channel`, `data.reason`, `data.error` there instead, so the entire string formatting moves to the client.

RockyTV:
I've moved the error description handling to the client. Took me a bit to figure out what `ParsedMessage` does, but I managed to understand it after looking at its source.

Right now the only issue I've found is that the parser treats `#channel:` as a channel (and it's valid), so I think we should use dashes to separate the error summary and its description, e.g. `Cannot join #channel - You have been banned.`

xPaw:
> so I think we should use dashes to separate the error summary and its description

Sure, should be fine to do that.

xPaw:
@RockyTV Since you're handling errors, do you want to also look into turning `ERR_UNKNOWNCOMMAND` into an `error` msg like the other errors are so we can format it into a nicer message?

See #797

RockyTV:
Sure thing, I can add a nicer message for unknown command.

RockyTV:
I don't know if it will be better to handle the unknown command error in TL, but it was already merged in irc-fw. I see two options: either wait for the irc-fw update and merge this, or just add a custom handler and remove it as soon as irc-fw is updated.

xPaw:
Just add the case for unknown command. When irc-fw is updated, then it will just work.

client/components/MessageTypes/error.vue
```diff
@@ -0,0 +1,63 @@
...
+				case "no_such_nick":
+					return `User ${nick} hasn't logged in or does not exist.`;
+				case "not_on_channel":
+					return "Cannot perform action: You're not on the channel.";
+				case "password_mismatch":
+					return "Password mismatch.";
+				case "too_many_channels":
+					return `Cannot join ${channel} - You've already reached the maximum number of channels allowed.`;
+				case "unknown_command":
+					return `Unknown command: ${command}`;
```
xPaw:
```suggestion
					return `Unknown command: ${this.message.command}`;
```

client/components/MessageTypes/error.vue
```diff
@@ -0,0 +1,63 @@
...
+	props: {
+		network: Object,
+		message: Object,
+	},
+	computed: {
+		errorMessage() {
+			const reason = this.message.reason;
+			const errorType = this.message.error;
+			const nick = this.message.nick;
+			const channel = this.message.channel;
```
xPaw: Can you remove these consts and just reference them directly, as each error message uses different properties.

src/plugins/irc-events/error.js
```diff
@@ -7,24 +7,13 @@ module.exports = function(irc, network) {
...
-		}
-
 		const msg = new Msg({
 			type: Msg.Type.ERROR,
-			text: text,
+			error: data.error,
 			showInActive: true,
+			nick: data.nick,
+			channel: data.channel,
+			reason: data.reason,
```
xPaw: `command` needs to be added here. Or perhaps we should just copy all properties from `data` dynamically?

RockyTV:
I've pushed a new commit that removes the consts and references the properties directly in each case. I believe all commits can be squashed into a single one when ready to merge, either from my side or from GitHub's interface.

xPaw:
We don't use squash on github because it doesn't create a merge commit, so if you could rebase locally that would be great.

src/plugins/irc-events/error.js
xPaw: You marked this as resolved but didn't add `command: data.command`.

PR #3745, opened 2020-02-07
Add shrug to codebase

Nek-:
Problem: god, the command shrug was absent
Solution: adding it 👍 ✨ ✌️

Please notice that awesome refactoring ✨ for the echo message function that avoids code duplication. (this is the stuff that makes it appear in local after sending it)

### DIFF
--- a/src/plugins/inputs/action.js
+++ b/src/plugins/inputs/action.js
@@ -3,7 +3,7 @@
 const Chan = require("../../models/chan");
 const Msg = require("../../models/msg");
 
-exports.commands = ["slap", "me"];
+exports.commands = ["slap", "me", "shrug"];
 
 exports.input = function({irc}, chan, cmd, args) {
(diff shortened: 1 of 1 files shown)


xPaw:
There's a shortcuts plugin you can use to do this.

Nek-:
Excuse me but what do you mean?

xPaw:
https://github.com/MiniDigger/thelounge-plugin-shortcuts

MiniDigger:
Basically, we shouldnt fill the Core codebase with stuff like this, thats why the plugin api was created.
If we have shrug, ppl will want a tableflip too. And then another emoji.
Instead, you can just add whatever you want via plguins, my shortcut plugin makes the really easy.

Nek-:
Except shrug is by default on slack and discord as well.

MiniDigger:
That command was "inherited" from shout, so it wasnt added by the current maintainer team.
https://github.com/thelounge/thelounge/commit/dbcf26d1bd07c8512fec25fbe1a05bac6228a31c

PR #3753, opened 2020-02-11
Wrap stdout parsing from yarn into try/catch

xPaw:
cc @Mikaela for test

Since we only parse it for `success:true`, it should arrive in a separate `data` event, andwe should be fine.


Mikaela:
What is the easiest way to test this while following the Debian packages?

xPaw:
Find `src/command-line/utils.js` on your disk and apply the patch.

PR #3764, opened 2020-02-17
Render images in canvas before upload to remove exif

### LINKED ISSUE #3025: File uploader should remove EXIF data from images (especially location data)

xPaw:
Fixes #3025

Added an option to allow users to disable this if desired.

### DIFF
--- a/client/components/Windows/Settings.vue
+++ b/client/components/Windows/Settings.vue
@@ -229,6 +229,29 @@
 				</div>
 			</template>
 
+			<div
+				v-if="$store.state.settings.advanced && $store.state.serverConfiguration.fileUpload"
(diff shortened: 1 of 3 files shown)


xPaw:
Let's merge after Mar 17 when Chrome 81 hits stable.

PR #3778, opened 2020-02-25
Clear channel history (and a new confirmation dialog)

xPaw:
Clear history:
- Clears out in-memory, resets unread/highlight data
- Deletes the channel in all message storages (~so the log file, and~ in sqlite)
  - EDIT: Only deletes from sqlite for now, someone brought up a good point that log files may be expected to be append-only.

TODO:
- [x] Send event to other open clients that a channel was cleared

Dialog:
image

It's also used for the network removal confirmation.


client/css/style.css
```diff
@@ -2708,6 +2709,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 /* Image viewer and drag-and-drop overlay */
 
+#confirm-dialog-overlay,
```
richrd: I'm beginning to feel we might want to have a generic overlay class for all of these overlays to use. Maybe that can be done in another PR if we think it's worth it.

xPaw: Yeah probably should figure something out, but there's slight differences between all of them, so that doesn't help.

PR #3798, opened 2020-03-09
Implement navigation in image viewer

### LINKED ISSUE #3522: Previous/next image buttons in the image viewer
Vue branch currently does not implement these, and I'm making this separate issue for it because I'm more than certain it won't be implemented in the `vue` branch.

It was initially implemented in TL in #1365

I want to know whether people use these buttons, and if they do, does it work as expected? In over 2 years of these existing I have not used them once.

Furthermore, the imageviewer in vue implements a lot more interactions with the image (zooming, moving around, etc), and adding arrow buttons might affect that behaviour.

richrd:
Fixes #3522

### DIFF
--- a/client/components/ImageViewer.vue
+++ b/client/components/ImageViewer.vue
@@ -76,26 +76,25 @@ export default {
 		},
 	},
 	watch: {
-		link() {
+		link(newLink, oldLink) {
 			// TODO: history.pushState
-			if (this.link === null) {
+			if (newLink === null) {
+				this.$root.$off("escapekey", this.closeViewer);
+				this.$root.$off("resize", this.correctPosition);
+				Mousetrap.unbind("left", this.previous);
+				Mousetrap.unbind("right", this.next);
 				return;
 			}
 
 			this.setPrevNextImages();
 
-			this.$root.$on("resize", this.correctPosition);
+			if (!oldLink) {
+				this.$root.$on("escapekey", this.closeViewer);
+				this.$root.$on("resize", this.correctPosition);
+				Mousetrap.bind("left", this.previous);
+				Mousetrap.bind("right", this.next);
+			}
 		},
-	},
-	mounted() {
-		this.$root.$on("escapekey", this.closeViewer);
-		Mousetrap.bind("left", this.previous);
-		Mousetrap.bind("right", this.next);
-	},
-	destroyed() {
-		this.$root.$off("escapekey", this.closeViewer);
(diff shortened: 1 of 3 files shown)


client/components/ImageViewer.vue
```diff
@@ -78,8 +105,34 @@ export default {
 			}
 
 			this.$root.$off("resize", this.correctPosition);
+			this.channel = null;
+			this.previousImage = null;
+			this.nextImage = null;
 			this.link = null;
 		},
+		getNewImage(delta) {
```
xPaw: Better off assigning next/prev within this once, so it doesn't have to filter twice.

xPaw:
cc @dgw for testing

xPaw:
10-1325-MistyrosePuma

The arrow buttons don't overlay particularly well over the image. I believe the old image viewer had padding to account for them.

Perhaps they should be repositioned?

richrd:
Yeah I think we want to avoid showing them on top of the image at least by default, unless the user zooms it. Alternatively we could only show them for a short while on hover and hide them after a timeout, similarly as the controls of video players usually do.

xPaw:
We could also reposition them to the bottom for example.

maxpoulin64:
Just checked with the old one for comparison, and the arrows were just always on top of the image in the same position: 

image

I think this is fine to accept as-is at least for feature parity with the older one, and it can be improved separately.

Code looks fine to me, but I'm not confident enough with the new Vue thing to give it a full review/approval.

richrd:
If that's how it was in the old version, I agree, we can improve later if we feel like it.

client/components/ImageViewer.vue
```diff
@@ -56,30 +76,62 @@ export default {
...
+
+			if (!oldLink) {
+				this.$root.$on("escapekey", this.closeViewer);
+				this.$root.$on("resize", this.correctPosition);
+				Mousetrap.bind("left", this.previous);
+				Mousetrap.bind("right", this.next);
+			}
 		},
 	},
-	mounted() {
```
xPaw, line 68: Failed rebase?

richrd: I think this is covered by the watch, or is it?

xPaw: Oh this PR includes a commit which is already on master. Can you rebase it cleanly?

xPaw: I didn't realize my commit wasn't on master, so this is fine.

PR #3828, opened 2020-03-23
Fix body overscroll and overflow on iOS Safari

stevenengler:
In order to prevent scrolling past the edges of the body (overscroll) in Safari on iOS, the overflow must be hidden (not only overflow-y). The new behavior fixes the top bar and the text input bar to the top and bottom of the window respectively.

Standard view: https://ibin.co/5Gda9ZtZROs7.jpg
View while scrolling down (latest version): https://ibin.co/5Gdb7DBJALRM.jpg
View while scrolling down (this PR): https://ibin.co/5GdbSRUbiaqS.jpg

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -104,7 +104,7 @@ body {
 	  * Disable pull-to-refresh on mobile that conflicts with scrolling the message list.
 	  * See http://stackoverflow.com/a/29313685/1935861
 	  */
-	overflow-y: hidden;
+	overflow: hidden; /* iOS Safari requires overflow rather than overflow-y */
 }
 
 a,


xPaw:
Does this still disable the pull-to-refresh on Android with this change?

stevenengler:
I'm not sure unfortunately as I don't have a way to test this. Before merging someone should add this to the custom stylesheet in settings and double check that it doesn't break the pull-to-refresh on Android. I suspect that it doesn't as the functionality of `overflow-y` should be a subset of `overflow`, but it would be good to double check.

xPaw:
Just played around on my Android, and `overflow: hidden` does disable pull to refresh with `overflow-y: unset`.

PR #3830, opened 2020-03-24
Improved web app support on iOS

stevenengler:
This improves the style when running as a web app on iOS.

Examples:
Current version: https://ibin.co/5Gkr15GfgtTp.jpg
New version: https://ibin.co/5Gkr5ldAE0hR.jpg
New version: https://ibin.co/5Gkr9pHK9QyE.jpg

It does this by adding `viewport-fit=cover` to the HTML and adding custom css to support this viewport setting. I tested this to make sure nothing broke on desktop (Firefox) and iPhone/iPad (Safari, Firefox, and as a web app). If it is to be merged, it should also be tested on other platforms/browsers.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -105,6 +105,13 @@ body {
 	  * See http://stackoverflow.com/a/29313685/1935861
 	  */
 	overflow: hidden; /* iOS Safari requires overflow rather than overflow-y */
+
+	/* To support "viewport-fit=cover" */
+	padding:
+		env(safe-area-inset-top)
+		env(safe-area-inset-right)
+		env(safe-area-inset-bottom)
+		env(safe-area-inset-left);
 }
 
 a,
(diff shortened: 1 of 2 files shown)


MaxLeiter:
related: https://github.com/thelounge/thelounge/pull/3444

stevenengler:
Thanks for taking a look at this and for pointing me at the other issues/PRs. Closing this is fair due to bad `viewport-fit=cover` support in browsers. The advantage of this pull request compared to #3444 is that it doesn't require the absolute position. Also I wrapped most of the css in `display-mode: standalone` media queries so I think it should have limited impact on browsers.

richrd:
I like how this looks but what's the browser support like?

amaranth:
The new CSS features are sort of standard (not finalized yet for some of it) and seem to be supported everywhere except IE.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/display-mode
https://developer.mozilla.org/en-US/docs/Web/CSS/env

On the other hand the actual `viewport-fit` rule, is, according to MDN and caniuse, completely unsupported. I guess maybe only iOS web apps support it?

https://developer.mozilla.org/en-US/docs/Web/CSS/@viewport/viewport-fit

xPaw:
Thanks for researching this, but I'm gonna close this for now due to `viewport-fit` not being supported anywhere according to caniuse. We can reopen and review this when browsers start to ship this.

kode54:
FYI, apparently, Apple doesn't even support this any more, either. Leave it to Stack Overflow to have an alleged solution:

https://stackoverflow.com/questions/60154236/is-viewport-fit-cover-no-longer-working-on-the-ios-safari

PR #3844, opened 2020-04-09
CertFP support; separate SASL configuration

### LINKED ISSUE #3197: Incorrectly using "Nick" instead of "Username" when logging in to SASL server
* *Node version:* 
```
# node --version
v10.15.1  
```

* *Browser version:* 
```
Chrome Version 74.0.3729.131 (Official Build) (64-bit)
```

* *Device, operating system:* 
```
Windows 10 Pro 1809
```

* *The Lounge version:* 
```
# thelounge --version
v3.0.1 
```

---

I have a private server that requires SASL to log in. If I set my username and password correctly, login fails. If I set my Nick to my user, login succeeds. The workaround right now is to set my Nick to my username, log in, and then change my nick after.

- [ ] SASL should use the Username field for login purposes.

### LINKED ISSUE #3836: Password field serves no purpose for services-enabled public instances
For instances with services and with public Lounge there is no point in having the password field visible - it is confusing for non-technical and novice users, and serves no purpose as it seems entering password in the field does not identify against nickserv anyway. It is also misleading to new users who might be looking for registration form.

xPaw:
- Automatically generates TLS certificates
- Allows using SASL EXTERNAL
- Added separate UI to control SASL username/password, or whether you prefer using EXTERNAL
- Combined `displayNetwork` into `lockNetwork`

This needs testing, and the connect/edit network UI still needs some work (especially in public mode).

Closes #3836
Fixes #3197


richrd:
I don't know much about certificate generation so can't comment much on that part but otherwise looks good to me.

PR #3845, opened 2020-04-09
Implement native app badges for highlights (Chrome 81+)

### LINKED ISSUE #2933: Implement 'badging' when it starts shipping in browsers
Reference: https://developers.google.com/web/updates/2018/12/badging-api

Basically it allows controlling badges on the app icon:
badges

The Lounge already syncs unread counts per channel, and we already support push notifications with service workers. So we will just need to add a sync count for _all_ unread channels when pushing notifications and add a new sync message when the unread counter decreases (might be a problem as browsers require showing a notification whenever something arrives in a SW).

xPaw:
This needs some testing to see whether the app badge clears out when expected and stuff like that.

~~I'm unsure if app badge over service worker works correctly, will probably remove it later (especially because we have no way to clear it at the moment).~~

Fixes #2933

### DIFF
--- a/client/js/vue.js
+++ b/client/js/vue.js
@@ -100,6 +100,14 @@ store.watch(
 	(_, getters) => getters.highlightCount,
 	(highlightCount) => {
 		favicon.setAttribute("href", highlightCount > 0 ? faviconAlerted : faviconNormal);
+
+		if (navigator.setAppBadge) {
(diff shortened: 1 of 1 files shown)


xPaw:
I removed service worker code from this PR, so it only sets the app badge in the same code where the title is updated.

PR #3846, opened 2020-04-09
Try to find channel from brackets in notice message

xPaw:
Do we actually want to support this feature? Some clients handle it this way, some clients don't.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -72,8 +72,14 @@ module.exports = function(irc, network) {
 			if (typeof chan === "undefined") {
 				// Send notices that are not targeted at us into the server window
 				if (data.type === Msg.Type.NOTICE) {
-					showInActive = true;
-					chan = network.channels[0];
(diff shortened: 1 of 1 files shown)


maxpoulin64:
That sounds like a pretty bad can of worms to open there.

This effectively allows any user on the network to send messages into other channels, without those channel operators being able to control it in any way. This could easily be used to trick users into thinking these are official channel notices when they are in fact not, especially with a convincing enough nick:

image

We could start checking if the user is in fact in those channels, but I have examples of channels where ChanServ sends those messages while ChanServ itself is not in the channel.

We could check if the user is a service bot, but I don't know how reliable that can realisticly be. Some networks allows for custom ChanServ names that will further complicate this.

Also, although fairly easy to fix, this can also be used to inject content into people's PMs:
image

xPaw:
Ah that's a good catch, let's not do this then. The client where I got inspired by this code didn't do any checks as far as I could see.

PR #3867, opened 2020-04-19
Fix off-by-one color error in webmanifest

maxpoulin64:
I don't know why that came to be and why it haven't been noticed before, but the color in the manifest is one shade of blue off from the icon, leaving a visible square around the logo while the app is loading:

Screenshot_20200419-175024_Chrome

I've recently started noticing it, and other users of my instance did too. Checking the colors in an image editor shows that the values are way off. Maybe Chrome added something that causes images to get a bit oversaturated to make them more vivid?

Anyway, that fixes it to the right color, and the square is now gone.


richrd:
Nice one! :+1: I think i noticed this once on my phone but then forgot about it.

PR #3868, opened 2020-04-20
Some fixes in mentions window

xPaw:
- Fixes right clicking in mentions window
- Fixes localetime filter
- Adds hover color to close button
- Adds tooltip to close button
- Add flex styles so the close button always stays aligned to the right


fnutt:
Thanks for fixing the issue with placement of close button. 👍

PR #3870, opened 2020-04-22
Safeguard nick randomizer up to allowed length

xPaw:
It looks like oragano sends "nick in use" error for unrelated checks that happen for their multi-client features, and it causes lounge to keep trying to change the nick.


xPaw:
Corresponding issue in oragano: https://github.com/oragono/oragono/issues/955

PR #3871, opened 2020-04-22
Filter user loading at startup for "advanced" LDAP

### LINKED ISSUE #3220: When an LDAP user is loaded, confirm their account is enabled.
In #3219, we are going to be loading LDAP users automatically on startup.

It would be useful if thelounge would check these users against the existing LDAP user filter (`config.ldap.searchDN.filter`) to make sure their LDAP account is enabled, and not load or remove the user if their account doesn't match the filter.

This would be useful so that when a user is no longer a part of the organisation the LDAP belongs to, their account will not automatically reconnect for ever after.

ebardie:
Fixes #3220 

This PR is a two-for-one offer.

1. Modify the auth plugin API mechanism to hide implementation details from module users. Initially this just tidies up calls to auth, but the motivation is to save us from having to duplicate the auth plugin handling code from server.js in clientManager.js in (2). 

2. Filter user loading at startup for "advanced" LDAP

@xPaw : is this approach okay? If not, what would you like changing?

It occurs to me that we could extend config.js to allow the admin to choose between logging a message on extinct users (what happens with this PR), silencing the message, and deleting the user config file. 

This work has been performed on behalf of my employer [Codethink Ltd.](https://www.codethink.co.uk)

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -6,6 +6,7 @@ const colors = require("chalk");
 const crypto = require("crypto");
 const fs = require("fs");
 const path = require("path");
+const Auth = require("./plugins/auth");
 const Client = require("./client");
 const Helper = require("./helper");
 const WebPush = require("./plugins/webpush");
@@ -45,7 +46,15 @@ ClientManager.prototype.loadUsers = function () {
 		);
 	}
 
-	users.forEach((name) => this.loadUser(name));
+	// This callback is used by Auth plugins to load users they deem acceptable
+	const callbackLoadUser = (user) => {
+		this.loadUser(user);
+	};
+
+	if (!Auth.loadUsers(users, callbackLoadUser)) {
+		// Fallback to loading all users
+		users.forEach((name) => this.loadUser(name));
+	}
 };
 
 ClientManager.prototype.autoloadUsers = function () {
--- a/src/plugins/auth.js
+++ b/src/plugins/auth.js
@@ -21,6 +21,12 @@ module.exports = {
 
 	// Must override: implements authentication mechanism
 	auth: () => unimplemented("auth"),
+
+	// Optional to override: implements filter for loading users at start up
+	// This allows an auth plugin to check if a user is still acceptable, if the plugin
+	// can do so without access to the user's unhashed password.
+	// Returning 'false' triggers fallback to default behaviour of loading all users
+	loadUsers: () => false,
 };
 
 // local auth should always be enabled, but check here to verify
--- a/src/plugins/auth.js
+++ b/src/plugins/auth.js
@@ -1,0 +1,44 @@
+"use strict";
+
+const log = require("../log");
+const colors = require("chalk");
+
+// The order defines priority: the first available plugin is used.
+// Always keep 'local' auth plugin at the end of the list; it should always be enabled.
+const plugins = [require("./auth/ldap"), require("./auth/local")];
+
+function unimplemented(funcName) {
+	log.debug(
+		`Auth module ${colors.bold(
+			module.exports.moduleName
+		)} doesn't implement function ${colors.bold(funcName)}`
+	);
+}
+
+// Default API implementations
+module.exports = {
+	moduleName: "<module with no name>",
+
+	// Must override: implements authentication mechanism
+	auth: () => unimplemented("auth"),
+};
+
+// local auth should always be enabled, but check here to verify
+let somethingEnabled = false;
+
+// Override default API stubs with exports from first enabled plugin found
+for (const plugin of plugins) {
+	if (plugin.isEnabled()) {
+		somethingEnabled = true;
+
+		for (const name in plugin) {
+			module.exports[name] = plugin[name];
+		}
+
+		break;
+	}
+}
+
+if (!somethingEnabled) {
+	log.error("None of the auth plugins is enabled");
+}
--- a/src/plugins/auth/ldap.js
+++ b/src/plugins/auth/ldap.js
@@ -3,6 +3,7 @@
 const log = require("../../log");
 const Helper = require("../../helper");
 const ldap = require("ldapjs");
+const colors = require("chalk");
 
 function ldapAuthCommon(user, bindDN, password, callback) {
 	const config = Helper.config;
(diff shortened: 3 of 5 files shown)


src/clientManager.js
```diff
@@ -45,7 +46,15 @@ ClientManager.prototype.loadUsers = function () {
...
 	}
 
-	users.forEach((name) => this.loadUser(name));
+	// This callback is used by Auth plugins to load users they deem acceptable
+	const callbackLoadUser = (user) => {
+		this.loadUser(user);
+	};
+
+	if (!Auth.loadUsers(users, callbackLoadUser)) {
+		// Fallback to loading all users
```
xPaw, line 55: I don't think this fallback is needed, just make the local auth return true (call the callback).

src/plugins/auth/ldap.js
```diff
@@ -139,6 +140,89 @@ function ldapAuth(manager, client, user, password, callback) {
...
+	const base = config.ldap.searchDN.base;
+
+	ldapclient.on("error", function (err) {
+		log.error(`Unable to connect to LDAP server: ${err}`);
+	});
+
+	ldapclient.bind(config.ldap.searchDN.rootDN, config.ldap.searchDN.rootPassword, function (err) {
+		if (err) {
+			log.error("Invalid LDAP root credentials");
+		} else {
```
xPaw, line 165: Do return here for less identation

src/plugins/auth/ldap.js
```diff
@@ -139,6 +140,89 @@ function ldapAuth(manager, client, user, password, callback) {
...
+				scope: config.ldap.searchDN.scope,
+				filter: `${config.ldap.searchDN.filter}`,
+				attributes: [config.ldap.primaryKey],
+				paged: true,
+			};
+
+			ldapclient.search(base, searchOptions, function (err2, res) {
+				if (err2) {
+					log.error(`LDAP search error: ${err2}`);
+				} else {
```
xPaw, line 180: Do return here for less identation

src/plugins/auth.js
```diff
@@ -0,0 +1,47 @@
...
+	moduleName: "<module with no name>",
+
+	// Must override: implements authentication mechanism
+	auth: () => unimplemented("auth"),
+
+	// Optional to override: implements filter for loading users at start up
+	// This allows an auth plugin to check if a user is still acceptable, if the plugin
+	// can do so without access to the user's unhashed password.
+	// Returning 'false' triggers fallback to default behaviour of loading all users
+	loadUsers: () => false,
```
xPaw, line 29: `unimplemented("loadUsers")` for consistency

src/clientManager.js
ebardie: A simpleLdapLoadUsers would also need to duplicate the same (albeit simple) code, and if other auth mechanisms were to be introduced in the future, they would also need to duplicate code. 

Having the fallback mechanism keeps the code in just the one place, close to where the function it calls is defined.

xPaw: Okay that's fair.

src/plugins/auth.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+// Override default API stubs with exports from first enabled plugin found
+for (let i = 0; i < plugins.length; ++i) {
```
xPaw, line 33: `for of` could work here

src/plugins/auth.js
```diff
@@ -0,0 +1,47 @@
...
+	if (plugins[i].isEnabled()) {
+		let name;
+
+		for (name in plugins[i]) {
```
xPaw, line 37: `const name`?

src/plugins/auth/ldap.js
```diff
@@ -139,11 +140,96 @@ function ldapAuth(manager, client, user, password, callback) {
...
+					res.on("searchEntry", function (entry) {
+						const user = entry.attributes[0]._vals[0].toString();
```
xPaw, line 182: This seems error prone, why doesn't `advancedLdapAuth` do this?

src/plugins/auth/ldap.js
```diff
@@ -139,11 +140,96 @@ function ldapAuth(manager, client, user, password, callback) {
...
+
+	ldapclient.on("error", function (err) {
+		log.error(`Unable to connect to LDAP server: ${err}`);
+	});
+
+	ldapclient.bind(config.ldap.searchDN.rootDN, config.ldap.searchDN.rootPassword, function (err) {
+		if (err) {
+			log.error("Invalid LDAP root credentials");
+		} else {
+			const userMap = {};
```
xPaw, line 166: `Set` will be cleaner here, you can remove from it whenever a user is confirmed to be loaded, then in `end` you can iterate over remaining entries.

src/plugins/auth.js
```diff
@@ -0,0 +1,47 @@
...
+			module.exports[name] = plugins[i][name];
+		}
+
+		break;
+	}
+}
+
+if (module.exports.length === 0) {
```
xPaw, line 45: This won't happen as you're exporting some defaults.

src/plugins/auth/ldap.js
ebardie: Because ldapjs returns DNs (which advancedLdapAuth uses) as strings in objectName, but attributes (such as 'uid') as buffers:

2020-04-23 13:18:36 [DEBUG] entry:  SearchEntry {
  messageID: 2,
  protocolOp: 100,
  controls: [],
  log: { child: [Function (anonymous)] },
  objectName: 'cn=Genevieve,ou=accounts,dc=localhost',
  attributes: [ Attribute { type: 'uid', _vals: [Array] } ],

or :

2020-04-23 13:21:10 [DEBUG] entry.attributes:  [ Attribute { type: 'uid', _vals: [ <Buffer 78 61 76 69 65 72> ] } ]

So the best we could do here is to check that the attribute type is the same as the filter requested, but that seems redundant?

src/plugins/auth.js
ebardie: Oops. Thanks.

src/plugins/auth/ldap.js
xPaw: I guess `attributes[0]` will match `attributes: [config.ldap.primaryKey]` ?

src/plugins/auth/ldap.js
xPaw: Sure.

src/plugins/auth/ldap.js
ebardie: yes. I can put a check in here if you think that it's worth it, but as I say it seems redundant to me,

xPaw:
Code seems ok (outside of ldap, can't really verify that), I'll give this a quick test later.

PR #3889, opened 2020-04-28
Changes required for vue 3

timmw:


client/components/MessageList.vue
```diff
@@ -63,6 +63,8 @@ import Message from "./Message.vue";
 import MessageCondensed from "./MessageCondensed.vue";
 import DateMarker from "./DateMarker.vue";
 
+let unreadMarkerShown = false;
```
xPaw, line 66: For context, this variable wasn't defined as `data`, and if you do that it will become reactive and cause updates in an infinite loop.

PR #3907, opened 2020-05-12
Mention improvements 

MaxLeiter:
Added clear all button, added colored nicks support, spacing and styling improvements

img

### DIFF
--- a/client/components/Mentions.vue
+++ b/client/components/Mentions.vue
@@ -5,20 +5,37 @@
 		@click="containerClick"
 		@contextmenu="containerClick"
 	>
-		<div class="mentions-popup">
+		<div
+			:class="{
+				'mentions-popup': true,
+				'colored-nicks': $store.state.settings.coloredNicks,
+			}"
+		>
 			<div class="mentions-popup-title">
-				Recent mentions
+				<span>
+					Recent mentions
+					<span v-if="resolvedMessages.length">({{ resolvedMessages.length }})</span>
+				</span>
+				<button
+					v-if="resolvedMessages.length"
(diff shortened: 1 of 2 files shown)


client/components/Mentions.vue
```diff
@@ -171,6 +200,11 @@ export default {
 		messageTime(time) {
 			return dayjs(time).fromNow();
 		},
+		removeMentions() {
```
xPaw, line 203: This certainly would be better to fire off as a single event to the server. Something like `mentions:clear_all` would be simple, then you can clear `state.mentions` once too.

client/components/Mentions.vue
```diff
@@ -5,20 +5,37 @@
 		@click="containerClick"
 		@contextmenu="containerClick"
 	>
-		<div class="mentions-popup">
+		<div
+			:class="{
+				'mentions-popup': true,
+				'colored-nicks': $store.state.settings.coloredNicks,
```
xPaw, line 11: Maybe it's better to move this class to `App`?

xPaw:
I think the clear all button is too big.

client/components/Mentions.vue
```diff
@@ -61,15 +78,17 @@
 	width: 400px;
 	right: 80px;
 	top: 55px;
-	max-height: 400px;
+	max-height: 800px;
```
xPaw, line 81: Maybe vh or % would work better.

xPaw:
I noticed the colored nick styles is not enough change. I'm gonna open my own pr to add hide-all and a few other fixes.

PR #3918, opened 2020-05-20
Make usernames case-insensitive when logging in

### LINKED ISSUE #2943: Lounge usernames should be case-insensitive
I'm running thelounge 2.7.1 in the provided Docker container, with Edge on Windows 10.

I've configured the software to authenticate using LDAP and initially signed in on my PC. I then opened a web browser on my phone and signed in … but didn't realise that the phone's keyboard had auto-capitalised the first letter and so, although authentication succeeded, it created essentially a new user with their own configuration.

I can't imagine many scenarios where the Lounge login username would need to be case-sensitive, so would it be safer to explicitly lower-case the names or, for backwards compatibility perhaps, do a case-insensitive check against existing names or config files?

ashwinikammar:
Fixes #2943

**What is the fix:**
For a new user, a profile will be created in thelounge and every time the same profile will be loaded irrespective of the username case.
For an existing user, one of his/her profile will be loaded irrespective of the username case. Further, duplicate profiles of the user will be removed.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -34,17 +34,39 @@ ClientManager.prototype.init = function (identHandler, sockets) {
 };
 
 ClientManager.prototype.findClient = function (name) {
-	return this.clients.find((u) => u.name === name);
+	name = name.toLowerCase();
+	return this.clients.find((u) => u.name.toLowerCase() === name);
 };
 
 ClientManager.prototype.loadUsers = function () {
-	const users = this.getUsers();
+	let users = this.getUsers();
 
 	if (users.length === 0) {
 		log.info(
 			`There are currently no users. Create one with ${colors.bold("thelounge add <name>")}.`
 		);
+
+		return;
 	}
+
+	const alreadySeenUsers = new Set();
+	users = users.filter((user) => {
+		user = user.toLowerCase();
+
+		if (alreadySeenUsers.has(user)) {
+			log.error(
+				`There is more than one user named "${colors.bold(
+					user
+				)}". Usernames are now case insensitive, duplicate users will not load.`
+			);
+
+			return false;
+		}
+
+		alreadySeenUsers.add(user);
+
+		return true;
+	});
 
 	// This callback is used by Auth plugins to load users they deem acceptable
 	const callbackLoadUser = (user) => {
--- a/src/server.js
+++ b/src/server.js
@@ -803,6 +803,10 @@ function performAuthentication(data) {
 		return;
 	}
 
+	if (typeof data.user !== "string") {
+		return;
+	}
+
 	const authCallback = (success) => {
 		// Authorization failed
 		if (!success) {


client/components/Windows/SignIn.vue
```diff
@@ -59,6 +59,7 @@
 import storage from "../../js/localStorage";
 import socket from "../../js/socket";
 import RevealPassword from "../RevealPassword.vue";
+import config from "../../../defaults/config";
```
xPaw, line 62: If you want to force lowercase usernames, this code needs to be in the server, and use `Helper.config`, not import defaults.

xPaw:
Is LDAP case-insensitive? Why only fix it for LDAP, and not in The Lounge overall?

src/clientManager.js
```diff
@@ -34,7 +34,11 @@ ClientManager.prototype.init = function (identHandler, sockets) {
 };
 
 ClientManager.prototype.findClient = function (name) {
-	return this.clients.find((u) => u.name === name);
+	// return this.clients.find((u) => u.name === name);
```
xPaw, line 37: Remove this comment.

It also would be nice to move `name = name.toLowerCase()` here, as there's no reason to lowercase the input more than once.

xPaw:
Can you undo the signin/config change and squash please?

xPaw:
@McInkay what do you reckon about this? It allows the files on disk to stay in whatever capitalization, but login will be case-insensitive.

Unsure how we should (if we even need to) handle same usernames with different casing on case-sensitive file systems.

src/clientManager.js
ashwinikammar: Can you please have a look once again, I have modified the code after your comment

xPaw: My comment still applies.

ashwinikammar:
Hi @xPaw, Can you please have a look at the latest changes

xPaw:
Yeah that looks fine, just need to give it some testing (see my comment above).

Did you verify that this works on LDAP?

ashwinikammar:
Yes I have tested with LDAP as well, it works fine

McInkay:
> what do you reckon about this? It allows the files on disk to stay in whatever capitalization, but login will be case-insensitive.

Sounds generally sensible to me.

> Unsure how we should (if we even need to) handle same usernames with different casing on case-sensitive file systems.

Maybe we need to add a check to startup to see if there are any duplicated names just with case sensitivity differences. 

We can deprecate this until the next major version, just skipping this code for now if we find one, and just refusing to start up from the next major version? I suspect this won't affect many people, but we probably need to do it properly.

xPaw:
We can do a loop after this code to find any duplicates that are case insensitive here:
https://github.com/thelounge/thelounge/blob/3194777b989571c29179b1f6599e9dc765cc8028/src/clientManager.js#L41-L43

> We can deprecate this until the next major version, just skipping this code for now if we find one

I say we just print an error for user that has duplicates and not load it (remove the duplicate entry from users array).

ashwinikammar:
Hi @xPaw , 
Do you want me to proceed with the approach you mentioned?

xPaw:
Should be fine.

src/clientManager.js
```diff
@@ -34,11 +34,33 @@ ClientManager.prototype.init = function (identHandler, sockets) {
...
 
 ClientManager.prototype.loadUsers = function () {
-	const users = this.getUsers();
+	const usersList = this.getUsers();
+	const updatedUsersList = [];
+
+	for (let i = 0; i < usersList.length; i++) {
+		const index = [];
+
+		for (let j = 0; j < usersList.length; j++) {
```
xPaw, line 51: Double loop may get more expensive than it needs to be.

You can use a `Set` to avoid the inner loop entirely. Like so:

Lowercase the name and check whether its in that set, if it is, then remove it from the user list and print a message, otherwise add it to set and continue.

If you loop from the end, then you can splice out the user from the array while iterating.

ashwinikammar: hi @xPaw ,
I can rewrite it using a single loop and the end-user array list will be in lowercase. Is that fine?

ashwinikammar:
hi @xPaw 
I have avoided using for loop in my latest change. Please have a look
Also, it would be helpful if you can suggest the changes to solve the build error for node 12.x on windows latest.

xPaw:
The updated code still essentially loops multiple times, just in a functional manner.

Here's a diff of what I had in mind, could you apply that to your PR (and squash it so its a single commit).
```diff
diff --git a/src/clientManager.js b/src/clientManager.js
index ca1f55b6..942d31ff 100644
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -34,18 +34,39 @@ ClientManager.prototype.init = function (identHandler, sockets) {
 };
 
 ClientManager.prototype.findClient = function (name) {
-	return this.clients.find((u) => u.name === name);
+	name = name.toLowerCase();
+	return this.clients.find((u) => u.name.toLowerCase() === name);
 };
 
 ClientManager.prototype.loadUsers = function () {
-	const users = this.getUsers();
+	let users = this.getUsers();
 
 	if (users.length === 0) {
 		log.info(
 			`There are currently no users. Create one with ${colors.bold("thelounge add <name>")}.`
 		);
+		return;
 	}
 
+	const alreadySeenUsers = new Set();
+	users = users.filter((user) => {
+		user = user.toLowerCase();
+
+		if (alreadySeenUsers.has(user)) {
+			log.error(
+				`There is more than one user named "${colors.bold(
+					user
+				)}". Usernames are now case insensitive, duplicate users will not load.`
+			);
+
+			return false;
+		}
+
+		alreadySeenUsers.add(user);
+
+		return true;
+	});
+
 	// This callback is used by Auth plugins to load users they deem acceptable
 	const callbackLoadUser = (user) => {
 		this.loadUser(user);
```

> Also, it would be helpful if you can suggest the changes to solve the build error for node 12.x on windows latest.

Looks like a temporary issue.

xPaw:
Thanks for applying the change.

I also thought that we need to make sure `user` is a string before calling `findClient`, can you apply this too:
```diff
diff --git a/src/server.js b/src/server.js
index ba1ab417..699cabb8 100644
--- a/src/server.js
+++ b/src/server.js
@@ -803,6 +803,10 @@ function performAuthentication(data) {
 		return;
 	}
 
+	if (typeof data.user !== "string") {
+		return;
+	}
+
 	const authCallback = (success) => {
 		// Authorization failed
 		if (!success) {
```

Could you please squash your commits into one after doing so? https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

Sorry for the trouble.

ashwinikammar:
Sure, i'll add that check and squash all commits into one.

ashwinikammar:
After adding below changes, users with the integers as username can also login because the numbers sent through input fields are considered as strings. Any thoughts on this?

> Thanks for applying the change.
> 
> I also thought that we need to make sure `user` is a string before calling `findClient`, can you apply this too:
> 
> ```diff
> diff --git a/src/server.js b/src/server.js
> index ba1ab417..699cabb8 100644
> --- a/src/server.js
> +++ b/src/server.js
> @@ -803,6 +803,10 @@ function performAuthentication(data) {
>  		return;
>  	}
>  
> +	if (typeof data.user !== "string") {
> +		return;
> +	}
> +
>  	const authCallback = (success) => {
>  		// Authorization failed
>  		if (!success) {
> ```
> 
> Could you please squash your commits into one after doing so? https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git
> 
> Sorry for the trouble.

xPaw:
> After adding below changes, users with the integers as username can also login because the numbers sent through input fields are considered as strings.

Sure, numeric usernames are valid. And the input field being sent through websocket would remain a string.

ashwinikammar:
> > After adding below changes, users with the integers as username can also login because the numbers sent through input fields are considered as strings.
> 
> Sure, numeric usernames are valid. And the input field being sent through websocket would remain a string. That check is to verify we can call `toLowerCase` on it.

okay got it! thank you

PR #3983, opened 2020-07-12
Add option to disable media preview.

dalcde:
This disables image previews iff prefetchStorage is disabled. This
stops the client from making any requests to third-party sites.

The change is backwards comaptible. Existing config files will have
disableMediaPreview set to `undefined`, which is functionally equivalent
to `false`.

Help wih setting up tests would be appreciated. I couldn't figure out how to
test that a preview is *not* displayed.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -84,11 +84,6 @@ function parseHtml(preview, res, client) {
 					$('meta[property="og:description"]').attr("content") ||
 					$('meta[name="description"]').attr("content") ||
 					"";
-				let thumb =
-					$('meta[property="og:image"]').attr("content") ||
-					$('meta[name="twitter:image:src"]').attr("content") ||
-					$('link[rel="image_src"]').attr("href") ||
-					"";
 
 				if (preview.head.length) {
 					preview.head = preview.head.substr(0, 100);
@@ -97,6 +92,17 @@ function parseHtml(preview, res, client) {
 				if (preview.body.length) {
 					preview.body = preview.body.substr(0, 300);
 				}
+
+				if (!Helper.config.prefetchStorage && Helper.config.disableMediaPreview) {
+					resolve(res);
+					return;
+				}
+
+				let thumb =
+					$('meta[property="og:image"]').attr("content") ||
+					$('meta[name="twitter:image:src"]').attr("content") ||
+					$('link[rel="image_src"]').attr("href") ||
+					"";
 
 				// Make sure thumbnail is a valid and absolute url
 				if (thumb.length) {
@@ -127,6 +133,10 @@ function parseHtml(preview, res, client) {
 
 function parseHtmlMedia($, preview, client) {
 	return new Promise((resolve, reject) => {
+		if (Helper.config.disableMediaPreview) {
+			reject();
+		}
+
 		let foundMedia = false;
 
 		["video", "audio"].forEach((type) => {
@@ -203,6 +213,10 @@ function parse(msg, chan, preview, res, client) {
 		case "image/jpg":
 		case "image/jpeg":
 		case "image/webp":
+			if (!Helper.config.prefetchStorage && Helper.config.disableMediaPreview) {
+				return removePreview(msg, preview);
+			}
+
 			if (res.size > Helper.config.prefetchMaxImageSize * 1024) {
 				preview.type = "error";
 				preview.error = "image-too-big";
@@ -228,6 +242,10 @@ function parse(msg, chan, preview, res, client) {
 				break;
 			}
 
+			if (Helper.config.disableMediaPreview) {
+				return removePreview(msg, preview);
+			}
+
 			preview.type = "audio";
 			preview.media = preview.link;
 			preview.mediaType = res.type;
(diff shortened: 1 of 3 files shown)


xPaw:
I think `parseHtmlMedia` should always resolve to false with this option so that HTML pages will still show a description, instead of no preview at all.

> Existing config files will have
disableMediaPreview set to undefined, which is functionally equivalent
to false.

TL falls back to default config for missing keys, FYI.

dalcde:
Updated. Thanks!

xPaw:
Can you add tests for this?

dalcde:
I'm not sure how to do this. A "loading" preview exists by default and then is removed asynchronously. Checking if the preview still exists after a timeout seems prone to errors.

xPaw:
You can still test things like not picking videos from html.

dalcde:
I've added a test, but I'm not a huge fan of this. It modifies the 
global config object, which has to be restored at the end of the test.  
This seems pretty prone to errors in the future.

xPaw:
You can group that test with `describe` and then use `before` and `after` to correctly reset it: https://stackoverflow.com/a/37897703

dalcde:
Done!

xPaw:
I don't see anything changed.

dalcde:
Sorry forgot to push!

test/plugins/link.js
```diff
@@ -222,6 +222,40 @@ Vivamus bibendum vulputate tincidunt. Sed vitae ligula felis.`;
 		});
 	});
 
+	describe("test disableMediaPreview", function () {
+		beforeEach(function (done) {
```
xPaw, line 226: You don't need `done` as there is nothing async. Same in `afterEach`.

xPaw:
"I also would like to see more tests like a direct link to image where it should not produce an image thumbnail, but still return the image size."

Can you add more tests please?

dalcde:
On Sun, Jul 19, 2020 at 08:16:22AM -0700, Pavel Djundik wrote:
>"I also would like to see more tests like a direct link to image where it should not produce an image thumbnail, but still return the image size."

At the moment this does not happen. It appears that the image size comes 
as part of the preview. If I don't return a preview, there will be no 
accompanied size. How would I achieve this?

xPaw:
Ah yeah you're right that we can't do that. In that case we can avoid the image thumbnail request here:

https://github.com/thelounge/thelounge/blob/b770c33a714f68627746c56bd8bf5ec08cf82c99/src/plugins/irc-events/link.js#L106-L107

As it will be dropped later on anyway. Again, I would like to see tests for it (e.g. to see that msg:preview triggered, but thumbnail url was NOT visited).

dalcde:
I added the test, and added a similar test for videos.

xPaw:
Thanks, that looks better now. Did you test that `throw` actually works there? I was thinking of tracking a `bool` there instead. But if throw works, that's fine too.

I would like to see tests for direct image/video links too (without html), for that you would need get routes that return correct content-type though.

That way we can also verify that direct links do not generate previews. Not sure how we would track `removePreview` calls though, since it won't emit a preview event.

Sorry if I'm being catious/annoying, I just want the link prefetch code to be test covered to avoid any problems in the future.

src/plugins/irc-events/link.js
```diff
@@ -98,26 +93,36 @@ function parseHtml(preview, res, client) {
 					preview.body = preview.body.substr(0, 300);
 				}
 
-				// Make sure thumbnail is a valid and absolute url
-				if (thumb.length) {
-					thumb = normalizeURL(thumb, preview.link) || "";
-				}
+				if (Helper.config.prefetchStorage || !Helper.config.disableMediaPreview) {
```
xPaw, line 96: This could be an early return instead of putting all the thumb code inside the `if` for less nesting.

dalcde:
On Mon, Jul 20, 2020 at 09:18:28AM -0700, Pavel Djundik wrote:
>Thanks, that looks better now. Did you test that `throw` actually works there? I was thinking of tracking a `bool` there instead. But if throw works, that's fine too.

Throw actually works. After all, our expect calls just throw when the 
unexpected happens

(we end up getting a Timeout instead of an assertion failure, both with 
expect and the throw)

>I would like to see tests for direct image/video links too (without html), for that you would need get routes that return correct content-type though.
>
>That way we can also verify that direct links do not generate previews. Not sure how we would track `removePreview` calls though, since it won't emit a preview event.

Yeah I couldn't figure out how to do that

>Sorry if I'm being catious/annoying, I just want the link prefetch code to be test covered to avoid any problems in the future.

No worries. Tests are important!

xPaw:
There's a `/real-test-image.png` route available for all tests that you can use.

For videos and images, you can set the type in express route, something like this: https://stackoverflow.com/questions/51661744/how-to-set-content-type-when-doing-res-send/51661772

dalcde:
On Tue, Jul 21, 2020 at 12:49:19AM -0700, Pavel Djundik wrote:
>There's a `/real-test-image.png` route available for all tests that you can use.
>
>For videos and images, you can set the type in express route, something like this: https://stackoverflow.com/questions/51661744/how-to-set-content-type-when-doing-res-send/51661772

Yeah but I'm not sure about the part where I check that the preview was 
removed.

xPaw:
Yeah that is a bit tricky. One solution I can imagine working is creating a `Proxy` on `message.previews` and the first change to the array would be type `loading` as link prefetcher discovers the link, and the second one would be the array emptying as it get removed.

PR #3995, opened 2020-07-20
Cleanup vue router route guards

xPaw:
This is also needed for Vue 3 porting.

cc @timmw

### DIFF
--- a/client/js/router.js
+++ b/client/js/router.js
@@ -32,12 +32,58 @@ const router = new VueRouter({
 				next();
 			},
 		},
+		{
+			name: "Connect",
+			path: "/connect",
+			component: Connect,
+			props: (route) => ({queryParams: route.query}),
+		},
+		{
+			name: "Settings",
+			path: "/settings",
+			component: Settings,
+		},
+		{
+			name: "Help",
+			path: "/help",
+			component: Help,
+		},
+		{
+			name: "Changelog",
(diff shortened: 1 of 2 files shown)


client/js/router.js
```diff
@@ -32,12 +32,58 @@ const router = new VueRouter({
...
 	],
 });
 
+router.beforeEach((to, from, next) => {
+	// If user is not yet signed in, wait for appLoaded state to change
+	// unless they are trying to open SignIn (which can be triggered in auth.js)
+	if (!store.state.appLoaded && to.name !== "SignIn") {
+		store.watch(
+			(state) => state.appLoaded,
+			() => next()
```
xPaw, line 75: The only "issue" here is that it is possible to navigate to signin before this `next` happens.

If you reload the page with `/settings` and your auth fails, it will technically happen later after sign in here.

I haven't noticed it actually breaking anything though.

client/js/router.js
```diff
@@ -48,6 +94,12 @@ router.beforeEach((to, from, next) => {
 		return;
 	}
 
+	// Disallow navigating to invalid networks
+	if (to.name === "NetworkEdit" && !store.getters.findNetwork(to.params.uuid)) {
```
timmw, line 98: How about moving this to a before guard on the NetworkEdit route?

PR #3998, opened 2020-07-24
Highlight exceptions

### LINKED ISSUE #3994: highlights also needs an exclude list
Currently it is impossible to silence specific notifications that include your nick because there are no negative-highlight-matches.

We need a list of words to be excluded from notifications that passed the current include list, so we remove the false positives.

Imagine that you have bots that may notify about build status or commit messages made by yourself, and you are high-likely to not want to be notified about your own actions.

Jay2k1:
This adds an exception/exclude list for strings that should _not_ trigger highlights. The code strips IRC formatting before testing messages for the exceptions, so that the given strings match colored/styled messages as well. Fixes #3994.

I basically duplicated the highlight code and reversed it.

Both custom highlight settings fields now have labels and help text:

image

### DIFF
--- a/client/components/Windows/Settings.vue
+++ b/client/components/Windows/Settings.vue
@@ -351,8 +351,15 @@ This may break orientation if your browser does not support that."
 
 			<div v-if="!$store.state.serverConfiguration.public && $store.state.settings.advanced">
 				<label class="opt">
-					<label for="highlights" class="sr-only">
-						Custom highlights (comma-separated keywords)
+					<label for="highlights" class="opt">
+						Custom highlights
+						<span
+							class="tooltipped tooltipped-n tooltipped-no-delay"
+							aria-label="If a message contains any of these comma-separated 
+expressions, it will trigger a highlight."
+						>
+							<button class="extra-help" />
+						</span>
 					</label>
 					<input
 						id="highlights"
@@ -360,7 +367,31 @@ This may break orientation if your browser does not support that."
 						type="text"
 						name="highlights"
 						class="input"
-						placeholder="Custom highlights (comma-separated keywords)"
+						placeholder="Comma-separated, e.g.: word, some more words, anotherword"
+					/>
+				</label>
+			</div>
+
+			<div v-if="!$store.state.serverConfiguration.public && $store.state.settings.advanced">
+				<label class="opt">
+					<label for="highlightExceptions" class="opt">
+						Highlight exceptions
+						<span
+							class="tooltipped tooltipped-n tooltipped-no-delay"
+							aria-label="If a message contains any of these comma-separated 
+expressions, it will not trigger a highlight even if it contains 
+your nickname or expressions defined in custom highlights."
+						>
+							<button class="extra-help" />
+						</span>
+					</label>
+					<input
+						id="highlightExceptions"
+						:value="$store.state.settings.highlightExceptions"
+						type="text"
+						name="highlightExceptions"
+						class="input"
+						placeholder="Comma-separated, e.g.: word, some more words, anotherword"
 					/>
 				</label>
 			</div>
--- a/src/client.js
+++ b/src/client.js
@@ -62,6 +62,7 @@ function Client(manager, name, config = {}) {
 		manager: manager,
 		messageStorage: [],
 		highlightRegex: null,
+		highlightExceptionRegex: null,
 	});
 
 	const client = this;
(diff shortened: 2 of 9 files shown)


src/client.js
```diff
@@ -447,6 +449,34 @@ Client.prototype.compileCustomHighlights = function () {
 	);
 };
 
+Client.prototype.compileCustomHighlightExceptions = function () {
```
xPaw, line 452: Would be nice to have a single function so it doesn't duplicate all the code. You can pass `client.config.clientSettings.highlightExceptions` into it, and make it return null or the compiled regex.

src/plugins/irc-events/message.js
```diff
@@ -126,6 +126,14 @@ module.exports = function (irc, network) {
 			}
 		}
 
+		// if highlight exceptions match, do not highlight at all
```
xPaw, line 129: This comment should be below, where the `if` is.

As for `cleanMessage`, should we make `client.highlightRegex` also test against it?

client/components/Windows/Settings.vue
```diff
@@ -351,16 +351,45 @@ This may break orientation if your browser does not support that."
...
+				</label>
+			</div>
+
+			<div v-if="!$store.state.serverConfiguration.public && $store.state.settings.advanced">
+				<label class="opt">
+					<label for="highlightExceptions" class="opt">
+						Highlight exceptions
+						<span
+							class="tooltipped tooltipped-n tooltipped-no-delay"
+							aria-label="If a message contains any of these comma-separated expressions, it will not trigger a highlight
```
xPaw, line 380: This probably needs new lines as it will overflow (and css tooltips are kinda terrible)

xPaw:
cc @ssbarnea to give this a test.

client/components/Windows/Settings.vue
Jay2k1: Yeah, where you said I should add a line break, can't that tooltip get a max-width or something so we don't have to care about inserting line breaks?

client/components/Windows/Settings.vue
```diff
@@ -351,16 +351,45 @@ This may break orientation if your browser does not support that."
...
+						</span>
 					</label>
 					<input
 						id="highlights"
 						:value="$store.state.settings.highlights"
 						type="text"
 						name="highlights"
 						class="input"
-						placeholder="Custom highlights (comma-separated keywords)"
+						placeholder="word, some more words, anotherword"
```
xPaw, line 369: Shall we keep comma-separated here to make it clearer?

Like `Comma-separated, for example: word, some more words, anotherword` or something.

Jay2k1: Maybe. We can use "e.g." so the text fits the field size:
image

src/plugins/irc-events/message.js
Jay2k1: > As for cleanMessage, should we make client.highlightRegex also test against it?
Actually, why not. Things like inserting a control character into nicknames in bot messages will not affect this, and if you specifically add things to custom highlights, you probably prefer if that highlights regardless of formatting.

test/tests/customhighlights.js
```diff
@@ -96,4 +96,56 @@ describe("Custom highlights", function () {
 		client.compileCustomHighlights();
 		expect(client.highlightRegex).to.be.null;
 	});
+
+	// tests for highlight exceptions
+	client.config.clientSettings.highlights = "foo, @all,   sp ace   , 고";
+	client.config.clientSettings.highlightExceptions = "foo bar, bar @all, test sp ace test";
```
xPaw, line 102: You set these but don't call `compileCustomHighlights`?

test/tests/customhighlights.js
```diff
@@ -96,4 +96,56 @@ describe("Custom highlights", function () {
...
+
+	it("should NOT highlight due to highlight exceptions", function () {
```
xPaw, line 104: this `NOT` should be in next test i think?

test/tests/customhighlights.js
Jay2k1: Oh, I forgot to add that, yes. The custom highlights are cleared in line 91 and 95, so I need to recompile anyway.

test/tests/customhighlights.js
Jay2k1: No, it's not the regex where the negation happens but the code, so the regex needs to match things that actually should not highlight and vice versa.
Or in other words, if this test fails, the user would get a highlight because the exception code fails.

xPaw:
Can you rebase please, I will merge it.

PR #4014, opened 2020-08-04
Disable link prefetching for urls with no schema specified

xPaw:
image

People find it annoying when things like "readme.md", any python module name, or a .sh script gets a link preview. This patch keeps these links clickable, but will not generate a link preview.

Previews on bare domains are generally not that useful anyway.

*On a side note, I was thinking whether we should skip link previews for links that are wrapped with angle brackets. Discord does that for example.*


igalic:
> On a side note, I was thinking whether we should skip link previews for links that are wrapped with angle brackets. Discord does that for example.

as in, when someone writes in markdown?

igalic:
is this going to be a new default? without configuration?

(i highly approve of both of those!)

xPaw:
> as in, when someone writes in markdown?

I guess? This PR does not do that by the way, that was just a side note.

PR #4029, opened 2020-08-13
Replace dashes to underscores in emoji autocompletion

xPaw:
Replace dashes to underscores except for `:-1:` This removes autocompletion for `:-P` prompting for `:non-potable_water:`

Discord also uses underscores for example.


jojo0273:
- []()

PR #4051, opened 2020-08-28
Add HTML lang and labelled-by field to upload

MaxLeiter:
Found the errors using the [WAVE extension](https://wave.webaim.org/extension/).

Until we have localization, the UI language is always English, so I think it's fine to set the lang to en.

### DIFF
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -1,5 +1,5 @@
 <template>
-	<div id="chat-container" class="window" :data-current-channel="channel.name">
+	<div id="chat-container" class="window" :data-current-channel="channel.name" lang="">
 		<div
 			id="chat"
 			:class="{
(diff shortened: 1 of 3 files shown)


xPaw:
While it's true UI is English, the chat messages may not be, that's why I avoided adding the lang attribute.

MaxLeiter:
Then I think it should be set to empty string so it's explicitly unknown.

However, I believe if the provided content is English the language should be en

xPaw:
Is there an attribute we could set on chat content to indicate unknown language?

MaxLeiter:
We could set lang=“” (unknown) on the chat window or a container if they
don’t contain English elements. If not, we could declare it on each message.

xPaw:
Does that actually work, if you declare it on the chat window?

MaxLeiter:
From the spec:

> To determine the language of a node, user agents must look at the nearest ancestor element (including the element itself if the node is an element) that has a lang attribute in the XML namespace set or is an HTML element and has a lang in no namespace attribute set

https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes

So I think it would work fine

xPaw (approved):

PR #4057, opened 2020-08-29
Add kick reason dialog

GewoonYorick:
This PR makes the following changes:
- Add an inputDialog
- Modify the kick item in the context menu to request a kickreason from the user

Impression of what it looks like:
image

The InputDialog can also later be used for other things such as plugins


xPaw:
Wouldn't it be simpler to just add an optional input field into `ConfirmDialog`, which for kick would show up?

client/components/InputDialog.vue
```diff
@@ -0,0 +1,136 @@
...
+						type="text"
+						:value="reason"
+						@input="setReason"
+						@keypress.enter.exact.prevent="close"
+					/>
+				</form>
+			</div>
+			<div class="input-buttons">
+				<button class="btn btn-cancel" @click="close()">Cancel</button>
+				<button class="btn btn-confirm" @click="close(true)">{{ data.button }}</button>
```
MaxLeiter, line 28: I think the `close(true)` should be switched.

client/components/InputDialog.vue
```diff
@@ -0,0 +1,136 @@
...
+			</div>
+		</div>
+	</div>
+</template>
+
+<style>
+#input-dialog {
+	background: var(--body-bg-color);
+	color: #fff;
```
MaxLeiter, line 37: `var(--body-color)`?

client/components/InputDialog.vue
```diff
@@ -0,0 +1,136 @@
...
+	margin: 10px;
+	border-radius: 5px;
+	max-width: 500px;
```
MaxLeiter, line 40: Do we want this to still be modal-like on mobile? Maybe worth full-screening or something.

client/css/style.css
```diff
@@ -2756,13 +2758,18 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
+#input-dialog-overlay.opened,
 #confirm-dialog-overlay.opened,
 #upload-overlay.is-dragover,
 #image-viewer.opened {
 	visibility: visible;
 	opacity: 1;
 }
 
+#input-dialog-overlay {
+	background: rgba(0, 0, 0, 0.4);
```
MaxLeiter, line 2770: I think we should use the same overlay as other dialogs.

xPaw:
@MaxLeiter most of your comments would apply to `ConfirmDialog` too (since its copied over, and why I'm suggesting just adding input to it), so not strictly related to this PR.

MaxLeiter:
Closing due to inactivity

PR #4090, opened 2020-10-04
Make `add` and `reset` CLI commands scriptable

### LINKED ISSUE #3913: Make `add` and `reset` CLI commands scriptable
### Feature Description
I'd like to plug these commands into a script which manages users on a shared server, and I'd like to be able to do this in a clean way. I've seen this done on other scripts like nextcloud's `occ` through an environment variable, but a parameter would also work.

Currently I'm doing this by forwarding into the STDIN but this seems a bit sketchy as the format of those commands can change in the future, and it doesn't _just_ work for the `add` command. Any suggestions on how to do this are welcome

supertassu:
Add CLI options `--password` and `--save-logs` (for `add` only) in order to make adding users and changing user passwords scriptable.

Closes #3913.

### DIFF
--- a/src/command-line/users/add.js
+++ b/src/command-line/users/add.js
@@ -11,7 +11,9 @@ program
 	.command("add <name>")
 	.description("Add a new user")
 	.on("--help", Utils.extraHelp)
-	.action(function (name) {
+	.option("--password [password]", "new password, will be prompted if not specified")
+	.option("--save-logs", "if password is specified, this enables saving logs to disk")
+	.action(function (name, cmdObj) {
 		if (!fs.existsSync(Helper.getUsersPath())) {
 			log.error(`${Helper.getUsersPath()} does not exist.`);
 			return;
@@ -28,6 +30,11 @@ program
 
 		if (users.includes(name)) {
 			log.error(`User ${colors.bold(name)} already exists.`);
+			return;
+		}
+
+		if (cmdObj.password) {
+			add(manager, name, cmdObj.password, !!cmdObj.saveLogs);
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


xPaw:
@flying-sausages Can you give this a try to see if it works correctly for you?

src/command-line/users/add.js
```diff
@@ -11,7 +11,15 @@ program
...
 	.description("Add a new user")
 	.on("--help", Utils.extraHelp)
-	.action(function (name) {
+	.option(
+		"--password [password]",
+		"new password, will be prompted if not specified on command line"
+	)
+	.option(
+		"--save-logs",
+		"if password is specified on command line, this will effect if logs for this user will be saved on disk"
```
igalic, line 20: i think this should be "affect", not "effect"
but even so, i still don't understand what this option does

igalic (commented):
🙋🏻‍♀️

src/command-line/users/add.js
xPaw: `if password is specified on command line, this specified whether logs should be saved to disk (defaults to false)`

src/command-line/users/add.js
```diff
@@ -11,7 +11,15 @@ program
 	.command("add <name>")
 	.description("Add a new user")
 	.on("--help", Utils.extraHelp)
-	.action(function (name) {
+	.option(
+		"--password [password]",
+		"new password, will be prompted if not specified on command line"
```
xPaw, line 16: `on command line` feels redundant.

src/command-line/users/add.js
igalic: only if a password is specified, this enables saving (a user's) logs to disk

supertassu:
Oops, didn't mean to click on close. I've improved the wording on both of the options based on your feedback.

igalic (approved):

xPaw (approved):

xPaw:
Do you mind squashing your commits into one please.

supertassu:
> 
> 
> Do you mind squashing your commits into one please.

done!

PR #4093, opened 2020-10-30
Do not generate and send client certificate unless SASL EXTERNAL is requested

### LINKED ISSUE #4092: Don't send SASL certificate unless requested by the user
- _Node version:_ N/A
- _Browser version:_ N/A
- _Device, operating system:_ N/A
- _The Lounge version:_ 4.2.0

---

It seems that by default, thelounge will send the server a SASL certificate, even if no authentication is selected by the user. This is a privacy/tracking concern since this certificate acts as an identifier and allows the server to track a user over many sessions and many IP addresses, potentially forever. If the user does not select "Client certificate (SASL EXTERNAL)", then the certificate should not be sent.

xPaw:
Fixes #4092

Disabling SASL EXTERNAL will not remove the certificate, in case user wants to re-enable it again.

@stevenengler care to test?

### DIFF
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -204,8 +204,7 @@ Network.prototype.setIrcFrameworkOptions = function (client) {
 	this.irc.options.tls = this.tls;
 	this.irc.options.rejectUnauthorized = this.rejectUnauthorized;
 	this.irc.options.webirc = this.createWebIrc(client);
-
-	this.irc.options.client_certificate = this.tls ? ClientCertificate.get(this.uuid) : null;
+	this.irc.options.client_certificate = null;
 
 	if (!this.sasl) {
 		delete this.irc.options.sasl_mechanism;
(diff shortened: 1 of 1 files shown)


stevenengler:
Awesome, thanks for the quick response! I'll assume I should wait for a passing CI test, so please ping me when I should test it, or I'll check back periodically.

xPaw:
The CI failure was unrelated to this PR.

src/models/network.js
```diff
@@ -204,15 +204,15 @@ Network.prototype.setIrcFrameworkOptions = function (client) {
...
 	this.irc.options.rejectUnauthorized = this.rejectUnauthorized;
 	this.irc.options.webirc = this.createWebIrc(client);
-
-	this.irc.options.client_certificate = this.tls ? ClientCertificate.get(this.uuid) : null;
+	this.irc.options.client_certificate = null;
 
 	if (!this.sasl) {
 		delete this.irc.options.sasl_mechanism;
 		delete this.irc.options.account;
 	} else if (this.sasl === "external") {
```
xPaw, line 212: I'm thinking this if check needs a `this.tls` here.

stevenengler:
@xPaw I have tested this PR and it seems to work as I would expect it to. If you select "No authentication" or "Username + password", then `/whois` does not return certification information. Then if you select "Client certificate", `/whois` does return the certificate fingerprint. Toggling the setting requires you to disconnect and reconnect from the network before it takes effect. Every time when changing back to "Client certificate", the certificate fingerprint never changed.

xPaw:
> Toggling the setting requires you to disconnect and reconnect from the network before it takes effect.

That is correct, as the client certificate is sent as part of TLS handshake.

PR #4104, opened 2020-11-18
Add ignore option to contextmenu

GewoonYorick:
PR adds ignore option to the context menu. Icon can easily be changed if necessary.

Fixes https://github.com/thelounge/thelounge/issues/4097

image


brettgilio:
This change seems simple enough. I don't see any issues.

PR #4110, opened 2020-11-24
Fix breaking GIFs while removing metadata

### LINKED ISSUE #4109: Don't break GIFs with metadata removal
When uploading files thelounge checks whether it is an image type file and if so it renders the image in a canvas to get rid of all the metadata. This is all nice and well but breaks GIFs as it only renders the first frame. 

https://github.com/thelounge/thelounge/blob/5a1963647e9ee095a0b0bc585bf23f11169ca48a/client/js/upload.js#L141-L149

I propose to add a second exception for GIFs as for SVGs. GIFs do not support any EXIF data nor do they have any other kind of metadata that isn't important for the GIF to function properly.

Nachtalb:
Closes #4109
GIFs can't contain EXIF data and do not contain any other metadata that isn't operational

### DIFF
--- a/client/js/upload.js
+++ b/client/js/upload.js
@@ -141,7 +141,8 @@ class Uploader {
 		if (
 			store.state.settings.uploadCanvas &&
 			file.type.startsWith("image/") &&
-			!file.type.includes("svg")
+			!file.type.includes("svg") &&
+			file.type !== "image/gif"
 		) {
 			this.renderImage(file, (newFile) => this.performUpload(token, newFile));
 		} else {


MaxLeiter (approved):

client/js/upload.js
```diff
@@ -141,7 +141,8 @@ class Uploader {
 		if (
 			store.state.settings.uploadCanvas &&
 			file.type.startsWith("image/") &&
-			!file.type.includes("svg")
+			!file.type.includes("svg") &&
+			!file.type.includes("gif")
```
xPaw, line 145: Can't this just be `file.type === 'image/gif'`?

Nachtalb: I just went with the way it has been done for SVGs 🤷 , either way would work. Should I change it?

xPaw: Sure. The svg uses `includes` because the mime type is incosistent as far as I remember.

Nachtalb: Not according to https://www.iana.org/assignments/media-types/media-types.xhtml, shall I change both?

xPaw: I remember there was something specific to svg that had to be this way, I can't remember now though, so just keep it.

xPaw (approved):

PR #4114, opened 2020-11-25
Add a configuration option to set the socket file permissions

moritzbuhl:
In my FreeBSD Jail I have the problem that the UDS has wrong file permissions after starting this app.
`660` works such that the thelounge user can use the socket as well as the www group of the nginx proxy.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -35,6 +35,13 @@ module.exports = {
 	// This value is set to `9000` by default.
 	port: 9000,
 
+	// ### `socketPermissions`
+	//
+	// Set the permissions of the UNIX domain socket.
+	//
+	// This value is set to `0o660` by default.
+	socketPermissions: 0o660,
+
 	// ### `bind`
 	//
 	// Set the local IP to bind to for outgoing connections.
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -38,6 +38,7 @@ module.exports = {
 	// ### `socketPermissions`
 	//
 	// Set the permissions of the UNIX domain socket.
+	// Only applies if the `host` option is a UNIX domain socket.
 	//
 	// This value is set to `0o660` by default.
 	socketPermissions: 0o660,
--- a/src/server.js
+++ b/src/server.js
@@ -148,6 +148,12 @@ module.exports = function (options = {}) {
 	server.listen(listenParams, () => {
 		if (typeof listenParams === "string") {
 			log.info("Available on socket " + colors.green(listenParams));
+			fs.chmod(listenParams, Helper.config.socketPermissions, (err) => {
+				if (err) {
+					log.error(`fs chmod error: ${err}. Stopping server...`);
+					process.exit(1);
+				}
+			});
 		} else {
 			const protocol = Helper.config.https.enable ? "https" : "http";
 			const address = server.address();


moritzbuhl:
It runs on my box as expected. I am not sure if the current default permissions would work on linux or anywhere else.
Also should it really be called `chmod`?

src/server.js
```diff
@@ -147,6 +147,12 @@ module.exports = function (options = {}) {
 	server.listen(listenParams, () => {
 		if (typeof listenParams === "string") {
 			log.info("Available on socket " + colors.green(listenParams));
+			fs.chmod(listenParams, Helper.config.chmod, (err) => {
```
MaxLeiter, line 150: We can't assume that `Helper.config.chmod` exists because people who update may not add the value to their config.

MaxLeiter (changes requested):

MaxLeiter:
Hey @moritzbuhl, sorry for the delay and thanks for the contribution, but can you remove the Pushover support? It doesn't belong in this PR (and I don't think this project, but that belogs in another issue/PR).

moritzbuhl:
Ah Sorry. I didn't know that branch was still linked to this PR.

defaults/config.js
```diff
@@ -35,6 +35,13 @@ module.exports = {
 	// This value is set to `9000` by default.
 	port: 9000,
 
+	// ### `chmod`
+	//
+	// Set the permissions of the UNIX domain socket.
+	//
+	// This value is set to `0o660` by default.
+	chmod: 0o660,
```
brunnre8, line 43: chmod doesn't seem to be a good name for it in my view... chmod is the command that changes permissions, not the name of the thing.

socketPermissions would by my preference, as that's what it is. Also needs an easy to understand comment that it only applies if the host option is a unix socket and ignored otherwise

src/server.js
```diff
@@ -148,6 +148,15 @@ module.exports = function (options = {}) {
 	server.listen(listenParams, () => {
 		if (typeof listenParams === "string") {
 			log.info("Available on socket " + colors.green(listenParams));
+
+			if (Helper.config.chmod) {
+				fs.chmod(listenParams, Helper.config.chmod, (err) => {
+					if (err) {
+						log.error(`fs chmod error: ${err}. Stopping server...`);
+						process.exit();
```
brunnre8, line 156: this needs an exit(1) or whatever, anything but 0 as that's an error

defaults/config.js
```diff
@@ -35,6 +35,14 @@ module.exports = {
 	// This value is set to `9000` by default.
 	port: 9000,
 
+	// ### `socketPermissions`
+	//
+	// Set the permissions of the UNIX domain socket.
+	// Only applies if the `host` option is a UNIX domain socket.
+	//
+	// This value is set to `0o660` by default.
+	socketPermissions: 0o660,
```
brunnre8, line 44: Not sure what FreeBSD does, but on Linux this isn't actually enough to set the permissions up.
Chances are that you want to restrict access to a certain group, meaning you need a way to specify owner / group as well as the permission bits.

Meaning, we should probably make this an object no?
I'm fine with just merging it with the current ability to set permission bits, but should someone want user / group in the future it should be extensible without introducing yet another top level option.

Meaning, we could have
```js
socketPermissions: {
    mode: 0o660,
    user: undefined,
    group: undefined
}
```
eventually, starting with only mode.

Does this make sense or am I completely of the right track here?

src/server.js
```diff
@@ -148,6 +148,12 @@ module.exports = function (options = {}) {
 	server.listen(listenParams, () => {
 		if (typeof listenParams === "string") {
 			log.info("Available on socket " + colors.green(listenParams));
+			fs.chmod(listenParams, Helper.config.socketPermissions, (err) => {
+				if (err) {
+					log.error(`fs chmod error: ${err}. Stopping server...`);
```
brunnre8, line 153: s/fs/socket/

That it is using the fs module is really not relevant, what it tried to chmod is.
Bonus points if you also add the path, assuming that it may not be in the error message itself, depending on what OS and what error you get.

brunnre8 (commented):
After reading unix(7) I think we might need to restructure the option a bit, but I might be wrong.

defaults/config.js
brunnre8: Relevant section:

> Pathname socket ownership and permissions
       In the Linux implementation, pathname sockets honor the permissions of the directory they are in.  Creation of a new socket fails if the process does not have write and search (execute) permission on the directory in which the
       socket is created.
>
> On  Linux,  connecting to a stream socket object requires write permission on that socket; sending a datagram to a datagram socket likewise requires write permission on that socket.  POSIX does not make any statement about the
       effect of the permissions on a socket file, and on some systems (e.g., older BSDs), the socket permissions are ignored.  Portable programs should not rely on this feature for security.
>
>When creating a new socket, the owner and group of the socket file are set according to the usual rules.  The socket file has all permissions enabled, other than those that are turned off by the process umask(2).  
>
>The owner, group, and permissions of a pathname socket can be changed (using chown(2) and chmod(2)).

moritzbuhl: Looking into different projects, I only found gogs/gitea, that have a UNIX_SOCKET_PERMISSION option https://docs.gitea.io/en-us/config-cheat-sheet/ or https://github.com/gogs/gogs/blob/main/conf/app.ini.

moritzbuhl:
I managed to resolve my problem by calling `umask` in the rc file.

PR #4115, opened 2020-11-25
Add command to start local development IRC server

Nachtalb:
The server is started on localhost:6667 with `yarn start-irc`

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ The Lounge is the official and community-managed fork of [Shout](https://github.
 ## Installation and usage
 
 The Lounge requires latest [Node.js](https://nodejs.org/) LTS version or more recent.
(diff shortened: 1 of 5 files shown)


Nachtalb:
I added this for a faster development workflow. Though I just noticed that irc-server npm package does not support the whole IRC spec. A viable alternative would be to use something like this https://hub.docker.com/r/linuxserver/ngircd

Together with docker-compose, you'd have a server in no time with `docker-compose pull && docker-compose up -d`

xPaw:
While a cool idea, this is rather pointless if you want to test anything advanced (especially against IRCv3 specifications), it also puts a burden on TL to keep this server updated (which itself doesn't appear to be).

I would just suggest getting oragano server, it's trivial to run it.

After all, this is a client.

PR #4116, opened 2020-11-25
Add network specific leave message

### LINKED ISSUE #4059: Optional Per-Network Away Message
I don't think I saw anyone raise this feature. I saw that we have away/auto-away as an option. I think it'd be nice to be able to set the away message per-network. I can do this on ZNC, but ZNC can't detect when I'm idle or away with the Lounge (that I know of).

I think this might be a factor in some other issues. This could build on #1996? Possibly allow for a "system-wide default" away and a per-network over-ride to be available?

Nachtalb:
I do find his idea quite reasonable and would like to see this happening

### DIFF
--- a/client/components/NetworkForm.vue
+++ b/client/components/NetworkForm.vue
@@ -174,7 +174,7 @@
 					v-model="defaults.leaveMessage"
 					class="input"
 					name="leaveMessage"
-					placeholder="Smell ya later"
+					placeholder="The Lounge - https://thelounge.chat"
 				/>
 			</div>
 			<template v-if="defaults.uuid && !$store.state.serverConfiguration.public">
--- a/client/components/NetworkForm.vue
+++ b/client/components/NetworkForm.vue
@@ -167,6 +167,16 @@
 					maxlength="300"
 				/>
 			</div>
+			<div class="connect-row">
+				<label for="connect:leaveMessage">Away message</label>
+				<input
+					id="connect:leaveMessage"
+					v-model="defaults.leaveMessage"
+					class="input"
+					name="leaveMessage"
+					placeholder="Smell ya later"
+				/>
+			</div>
 			<template v-if="defaults.uuid && !$store.state.serverConfiguration.public">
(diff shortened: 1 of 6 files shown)


client/components/NetworkForm.vue
```diff
@@ -167,6 +167,16 @@
 					maxlength="300"
 				/>
 			</div>
+			<div class="connect-row">
+				<label for="connect:leaveMessage">Away message</label>
```
Jay2k1, line 171: I suppose the label "Away message" is a mistake and should probably read "Leave message"? :)

client/components/NetworkForm.vue
```diff
@@ -167,6 +167,16 @@
...
+				<input
+					id="connect:leaveMessage"
+					v-model="defaults.leaveMessage"
+					class="input"
+					name="leaveMessage"
+					placeholder="Smell ya later"
```
Jay2k1, line 177: I feel like something a bit more serious (like a simple "Bye!") would be more appropriate here.

client/components/NetworkForm.vue
Nachtalb: True i'll change it

richrd (approved):

xPaw:
What does this have to do with #4059? That issues talks about the away message, and you implemented leave message.

Nachtalb:
@xPaw Welp, I guess I interpreted it wrong 😂 sorry...

Anyway, I'd still like this feature here as well. If I find the time to do an actual netwrok specific AFK message I will do so, no promises though.

Nachtalb:
I removed the reference to the issue so that it won't be autoclosed in case this get's merged.

client/components/NetworkForm.vue
MaxLeiter, line 177: I agree with @Jay2k1, can we adjust this to be something more serious than Smell ya later, like `The Lounge - https://thelounge.chat` (which is the default global leaveMessage)

Issue #4123, opened 2020-12-02
Recognize ZNC buffextras messages


strugee:
### Feature Description

I have my instance of The Lounge connected to ZNC, where I have the [buffextras](https://wiki.znc.in/Buffextras) plugin enabled. I have it enabled because I like to have the extra information it gives me, but that also means that when ZNC restarts or reconnects, I see a mention in every single channel I'm in because the `*buffextras` user says `strugee!~strugee@strugee.net joined`.

It could be great if The Lounge could recognize the `*buffextras` "user" automatically and convert those messages into native actions. Apparently a few clients already have plugins that do this; see the linked wiki page.

PR #4125, opened 2020-12-03
Intercept ZNC *buffextras join messages

strugee:
This PR begins implementing the feature suggested in #4123. It only does join messages for now because it's actually surprisingly tedious to write these conversions (I had to read a bunch of ZNC source code) and I wanted feedback on whether this is a good direction to go. If yes, then I'll polish up the WIP code I have to handle mode change messages, look into writing some tests, and maybe write a few more conversions if I have time.

Note that despite (because of?) the lack of test suite updates, this code _has_ been manually tested.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -60,6 +60,27 @@ module.exports = function (irc, network) {
 				return;
 			}
 
+			// Intercept ZNC *buffextras messages and turn them into native messages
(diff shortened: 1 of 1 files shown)


brunnre8:
while I can't tell you if the feature is desired, this certainly would need to be a config switch. The nick is not necessarily name *bufextras, as "*" can be set to any other prefix char (normally something outside of the valid IRC range) in ZNC

Issue #3626, opened 2019-12-22
Allow setting permission bits of created socket


cinderblock:
### Feature Description

Following up on #686, it should be possible to set permissions for the created socket file.

As it stands, socket files created by `thelounge` are given default permissions:

```txt
cameron@myserver:~$ ls -l /run/thelounge/sock
srwxr-xr-x 1 thelounge thelounge 0 Dec 22 14:24 /run/thelounge/sock
```

This means that even if a webserver is in the `thelounge` group, it cannot write to the created socket, negating the usefulness of a socket file.

Please add the ability to control permission bits to socket on startup.

----

I have a tiny npm module that is designed to help with this. It allows configs to specify host or socket and option ports or permission bits. It also removes any existing socket file if the socket is dead. https://www.npmjs.com/package/server-starter Unfortunately this would be a breaking change to the current `config.js`.

xPaw:
It should default to taking the permissions from the folder, have you tried setting correct permissions on it?

cinderblock:
There are two issues that relate to getting permissions set correctly.

The first is the owner/group of the file. That, as I understand it, uses the user's default group unless the "sticky bit" is set on the folder which changes which group is set on file creation.

I'm using systemd's `RuntimeDirectory=thelounge` option (which creates `/run/thelounge` on startup, owned by `thelounge` user, and cleaned on shutdown). I could add an `ExecStartPre=` script that sets the sticky bit, but that's getting excessively confusing compared to other options, imho.

The second problem is setting the file permission bits. My system `umask` defaults to `0002` which should make created files have `rw` permissions for user & group. In practice, I need to add `w` permissions for group. (Not sure why there is a discrepancy there). I'd also like to remove any permissions for any other user/group.

I'm currently getting around these issues by adding `nginx` to the `thelounge` group **and** adding this to my systemd service configuration:
```ini
[Service]
RuntimeDirectory=thelounge
ExecStartPost=sleep 2
ExecStartPost=chmod g+w,o-rwx /run/thelounge/sock
Type=exec # This doesn't work for some reason. Separate problem, but listed here for completeness
```

In any case, controlling the permissions of a created socket is, imho, an expected feature of creating a socket for other programs to interact with. As long as we're not changing the user/group of the created file, `root` is not necessary.

xPaw:
What permission umask do you expect lounge to set on the socket?

cinderblock:
I personally want permissions to be set like this after startup:

```txt
[...]

PR #4128, opened 2020-12-13
allow setting permissions bits of created UNIX sockets

svenpeter42:
This PR adds a new configuration option `socketPermissions` which can be used to change the permissions bits on created UNIX sockets.
I personally run thelounge behind an nginx proxy through a UNIX socket and would like to lock down its permissions to 0660. I can then add the `nginx` user to the `thelounge` group. Right now the socket is created with 0776 by default my machine.

This change is backwards compatible since a default value of `undefined` for `socketPermissions` will just skip the chmod. I have additionally added a warning if someone tries to configure `socketPermissions` together with a HTTP(S) socket.

This has originally been requested in #3626.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -35,6 +35,18 @@ module.exports = {
 	// This value is set to `9000` by default.
 	port: 9000,
 
+	// ### `socketPermissions`
+	//
+	// Set the permissions for the UNIX domain socket. Ignored when `host` is set
+	// to an IP address or hostname.
+	// When the value is not set (i.e. `undefined`) the permissions will not be
+	// changed and will default to whatever permissions the operating system uses.
+	//
+	// Example: Set this to `0o660` to allow read/write access for the owner and
+	//          the group and disallow access for anyone else.
+	//
+	socketPermissions: undefined,
+
 	// ### `bind`
 	//
 	// Set the local IP to bind to for outgoing connections.
--- a/src/server.js
+++ b/src/server.js
@@ -132,6 +132,7 @@ module.exports = function (options = {}) {
 	}
 
 	let listenParams;
+	const socketPermissions = Helper.config.socketPermissions;
 
 	if (typeof Helper.config.host === "string" && Helper.config.host.startsWith("unix:")) {
 		listenParams = Helper.config.host.replace(/^unix:/, "");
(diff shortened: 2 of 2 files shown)


defaults/config.js
```diff
@@ -35,6 +35,18 @@ module.exports = {
...
+	//
+	// Set the permissions for the UNIX domain socket. Ignored when `host` is set
+	// to an IP address or hostname.
+	// When the value is not set (i.e. `undefined`) the permissions will not be
+	// changed and will default to whatever permissions the operating system uses.
+	//
+	// Example: Set this to `0o660` to allow read/write access for the owner and
+	//          the group and disallow access for anyone else.
+	//
+	socketPermissions: undefined,
```
MaxLeiter, line 48: Can we be explicit the default value is undefined, similar to the other config options? Should we explicitly set it to 0o776 (or some other value, probably without execute)?

brunnre8: I don't think setting it to a dummy value is something we should do... after all that makes some modes invalid.

I'd rather see undefined == respect umask

brunnre8: in fact, do we enforce any perms at all? If not that could already be solved without an additional config option by setting the umask appropriately no?

MaxLeiter: @brunnre8 I don't think the current default of 776 is correct, though (based on `Right now the socket is created with 0776 by default my machine.`)

brunnre8: what's your umask? That permissions would imply o001 which is a very strange one to have

src/server.js
```diff
@@ -140,12 +141,20 @@ module.exports = function (options = {}) {
 			port: Helper.config.port,
 			host: Helper.config.host,
 		};
+
+		if (socketPermissions !== undefined) {
+			log.warn("socketPermissions has no effect when a HTTP(S) socket is used.");
```
moritzbuhl, line 146: tcp socket

src/server.js
```diff
@@ -140,12 +141,20 @@ module.exports = function (options = {}) {
...
+		}
 	}
 
 	server.on("error", (err) => log.error(`${err}`));
 
 	server.listen(listenParams, () => {
 		if (typeof listenParams === "string") {
+			if (socketPermissions !== undefined) {
+				fs.chmodSync(listenParams, socketPermissions);
```
moritzbuhl, line 155: error handling?

brunnre8:
Closed in favor of https://github.com/thelounge/thelounge/pull/4114

Both want to do the same thing. Thanks anyhow for your contribution

PR #4134, opened 2020-12-28
Visually distinguish channel names

Nachtalb:
It's hard to distinguish inline channel mentions whereas users are highlighted with colours. Because channels don't have colours assigned I thought that making them bold is the most forward way to highlight them. 

Before
---
image

After
---
image

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1364,6 +1364,10 @@ textarea.input {
 
 /* Parsed nicks and channels */
 
+.inline-channel {
+	font-weight: bold;
+}
+
 #chat .user,
 .inline-channel {
 	cursor: pointer;


GewoonYorick:
Personally I'd prefer to keep this as a css snippet that can be used in a per user basis (custom css settings field) or to a theme, instead of making this the default.

fnutt:
I think this is a nice feature that would fit right into the core. If you want to change it in a theme, please feel free to do so. Thanks for the PR, @Nachtalb.

MaxLeiter:
Thanks for the contribution @nachtalb! Closing after some discussion in #thelounge. If users want this change they can add it to their themes or custom CSS.

xnaas:
For future reference (for anyone who might stumble across this issue) you can do something like this:

```css
#chat .inline-channel {
  color: #268bd2;
}
```

That colors `#channels` in chat messages the same color as a link. Obviously you could pick any color you like. If you specifically want them **bold**, you could do `font-weight: bold;` or whatever as well.

PR #4135, opened 2020-12-30
Add prefetchMaxSearchSize to override limit for link previews

brunnre8:
YouTube puts the opengraph tags needed for the preview after ~300KB in the body
instead of the beginning of the <head> tag.
Instead of hardcoding the value, allow the server admin to set the policy as
they prefer.

Youtube was also kind enough to put the tags in the body instead of the header, so we can't stop once we reach \</head\> either

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -150,6 +150,14 @@ module.exports = {
 	// This value is set to `2048` kilobytes by default.
 	prefetchMaxImageSize: 2048,
 
+	// ### prefetchMaxSearchSize
+	//
+	// This value controls the maximum request size made to find the link preview
+	// Open Graph tags. For some sites like YouTube this can easily exceed 300 kilobytes.
+	//
+	// This value is set to 50 kilobytes by default.
+	prefetchMaxSearchSize: 50,
+
 	// ### `fileUpload`
 	//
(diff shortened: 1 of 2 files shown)


sn0opy:
Works just fine on my setup 👍

defaults/config.js
```diff
@@ -150,6 +150,13 @@ module.exports = {
 	// This value is set to `2048` kilobytes by default.
 	prefetchMaxImageSize: 2048,
 
+	// ### `prefetchMaxSearchSize`
+	//
+	// This value controls the maximum request size made to find the link preview
+	// opengraph tags. For some sites like YouTube this can easily exceed 300KB
+	// set to 50 kilobytes as to not generate a lot of traffic on the server
```
MaxLeiter, line 157: Can this be something like:
// ### `prefetchMaxSearchSize`
//
// This value controls the maximum request size made to find the link preview
// Open Graph tags. For some sites like YouTube this can easily exceed 300 kilobytes.
//
// This value is set to `50` kilobytes by default.

MaxLeiter (changes requested):

defaults/config.js
brunnre8: ack, done

PR #4155, opened 2021-02-09
Update dependencies

MaxLeiter:


xPaw:
Did you test locally whether these updates worked?

Because current master appears to be broken due to socket.io change. https://github.com/socketio/socket.io/discussions/3741

PR #4156, opened 2021-02-09
Updated SQLite3, with windows build-tools (branch depUpdates)

SimplyLinn:


MaxLeiter:
As discussed on IRC, going to hold off and wait to see if upstream addresses this problem before we add another installation step for Windows users. Thanks!

PR #4171, opened 2021-03-01
Configure server ping timeout to 60 seconds

emilyst:
The default socket.io server-side ping timeout was changed from 60 seconds to 5 seconds. In browsers based on Chrome, this is not enough time to respond when the browser is idle. The end result is that the server sets the user away and then back approximately once every minute if the client window is idle, which is undesirable.

This change restores the previous timeout value.

See https://github.com/socketio/socket.io/issues/3259#issuecomment-474523271.


MaxLeiter:
Thanks!

Issue #3636, opened 2019-12-26
Implement multiple file share backends


AlexGustafsson:
### Feature Description

I’m thinking of a feature where one can upload a file to a number of backends such as Firefox Share and Dropbox Transfer.

Currently the only way to share files is by uploading them and making them available via the lounge.

I’m willing to implement this. Would it be a welcome contribution or is it not a direction you’re looking for?

xPaw:
This was discussed and thought about before. The Lounge itself will only support upload files locally (otherwise people ask to implement their favourite service).

This is something that could be done via plugins, but the plugin system is not at the stage where this is possible just yet.

Issue #1422, opened 2017-08-16
Add plugin support


YaManicKill:
There are many things that people want that we don't want to put in core. So we should have a plugin API to allow people to do these things.

I'm looking to start work on this, so I'm going to catalogue things that people will want to do with this. Once I have a first version working, I'll probably separate this into different issues for the other APIs that haven't been implemented.

Required APIs:

* Command aliases - this is a surprisingly common thing, and would just involve a plugin telling us what the alias is called, and what command it should call. This could easily include 
* Database - a plugin could provide a new database connection which will provide a way to give it new log lines to store, and a way to get all the logs at startup (could be expanded in the future)
* Settings - a way for plugins to give a list of settings, with different types, that will be put in the settings page and we will pass the user input to the plugin

I'll flesh out the ideas for implementation later. Other ideas for APIs, stick them in comments on this post for just now, until we have the first plugin API.

I'll base the structure of a plugin on #1228

MiniDigger:
having never touched node.js this might sound stupid but is it feasible to make an event driven api?
like have events for messages receive in channels, pms, events for send messages, events for closing and opening a window or the whole client, stuff like that.

it also would be cool to have an api to provide new options for context menus (on users, on channels, on networks, on messages, ...)

those are the two things I would propose additionally to the things mentioned above.

miwagner1:
It shouldn’t be too hard for acting on anything coming in from the irc server. A regex filter plugin like setup would work. This framework can also be used then for ignoring people, auto op people and more automated things.
I’m going to have to look into getting started with node myself.

YaManicKill:
I'm going to next implement a quick plugin support for custom commands.

This will have an endpoint for the plugins to call that will take the name of the command and a function to call when the command is run, plus a way to send other commands in it's place (I guess we'll probably want this to be a decorator pattern so that it will return the command that is sent to the server rather than this being as well as), probably in the same style as we would send to irc fw (so command and args as an array). 

It should ideally add the command to the client autocomplete, so that'll need to be passed from the server to the client on setup otherwise it'll have to rebuild the client which is ... not ideal. Any thoughts on this?

rain-1:
Would it be possible to divide this work into parts so that we can help?

McInkay:
As discussed in IRC, yes, the base server plugins exist. I'll be working on the plugin feature you want soon though.

MiniDigger:
just sharing some ideas what apis that could be added:

-  commands
    - custom commands
[...]

PR #4177, opened 2021-03-12
Support file uploads to dumpinen.com

osm:
This PR makes it possible to configure the clients to upload files to dumpinen.com instead of the self hosted file storage in the lounge.

dumpinen.com is a lightweight file dumping service, kind of like pastebin.com but without ads and the ability to dump all kind of files.

### DIFF
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -12,7 +12,7 @@ socket.once("configuration", function (data) {
 	store.dispatch("settings/applyAll");
 
 	if (data.fileUpload) {
-		upload.initialize();
+		upload.initialize(data.fileUploadOptions);
 	}
 
 	// If localStorage contains a theme that does not exist on this server, switch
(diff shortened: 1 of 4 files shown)


dgw:
I don't expect building different upload services into core to get much traction with this project (see e.g. #3636). But if you could generalize this to allow building upload services as plugins (see #1422) that would be a _huge_ step!

MiniDigger:
This sounds like something a plugin should provide. 
The lounge should expose an interface a plugin can implement to provide, plugin can implement that provider interface and register it, and the user can select a provider by specifing the name in the config, and additionally options for that provider (username, password, api keys, whatever)

osm:
Yeah that makes sense, I will fix that as soon as I can

MaxLeiter:
Closing, as this won't be accepted before more discussion is had. @osm, feel free to ping me (nick is same as my github name) in #thelounge on irc.freenode.net if you want to work on the plugin implementation.

PR #4178, opened 2021-03-20
Allow animated webp

Nachtalb:
Webp can be animated (just like gifs but better) but up until now, they were run through a canvas and thus the animation was lost. Webp can contain exif/xmp/icc etc (https://developers.google.com/speed/webp/docs/webpmux). and thus we want to be able to remove the data. The problem here is, there is no js library out there to remove the metadata from the, only read. So instead we read the metadata from the image and then decide wether or not we want to run in through a canvas. So this means if:
- animated / metadata => static / no metadata
- animate / no metadata => animated / no metadata

Examples:
animated / metadata: https://files.catbox.moe/lap1sg.webp
animated / no metadata: https://files.catbox.moe/pvdscf.webp

ℹ️ If you know about a JS Metadata removal tool that supports webp please tell me so I can replace this PR.

### DIFF
--- a/client/js/upload.js
+++ b/client/js/upload.js
@@ -2,6 +2,7 @@
 
 import {update as updateCursor} from "undate";
 
+import ExifReader from "exifreader";
 import socket from "./socket";
 import store from "./store";
 
@@ -141,10 +142,18 @@ class Uploader {
 		if (
 			store.state.settings.uploadCanvas &&
 			file.type.startsWith("image/") &&
-			!file.type.includes("svg") &&
-			file.type !== "image/gif"
+			!file.type.includes("svg")
 		) {
-			this.renderImage(file, (newFile) => this.performUpload(token, newFile));
+			file.arrayBuffer().then((buffer) => {
+				try {
+					ExifReader.load(buffer);
+				} catch (e) {
+					this.performUpload(token, file);
+					return;
+				}
+
+				this.renderImage(file, (newFile) => this.performUpload(token, newFile));
+			});
 		} else {
 			this.performUpload(token, file);
 		}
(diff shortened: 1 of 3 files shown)


Nachtalb:
One solution would also be to do that metadata removal serverside and use a tool like `exiv2` in which you can do `exiv2 rm myfile.jpg' to remove all metadata. Though like this you would send the file to the server with the metadata included which may concern some users. Also, we would depend on a third-party tool outside of thelounge (though this could be made optional). Either way animated WEBPs should be allowed as they act the same as gifs but are much smaller.

MaxLeiter:
Could we use ExifReader on the server ([src/plugins/uploader.js](https://github.com/thelounge/thelounge/blob/9d550d9364620d4b38bc21305ae95eb6595aedf2/src/plugins/uploader.js)) to avoid increasing the client bundle size? In general, I don't think users will care if the files have metadata when they're uploaded as long as the stored/accessible file is clean.

Nachtalb:
@MaxLeiter 
> In general, I don't think users will care

The problem I see here is that the user cannot control what is done on the server. So a malicious hoser could copy the file and extract the metadata for malicious use without the users' consent. Whereas if it's done client-side the user has full control and there are no privacy concerns. 

But if you want I could do a server-side thingy. As said in the updated comment above I found https://sharp.pixelplumbing.com/ which could remove all the metadata including WEBP using the libvips library.

MaxLeiter:
@Nachtalb are you in #thelounge on Freenode? If not, could you join? Easier to discuss

Nachtalb:
With #4186 (serverside metadata removal) this PR here will be obsolete

igalic:
> The problem I see here is that the user cannot control what is done on the server. So a malicious hoser could copy the file and extract the metadata for malicious use without the users' consent. Whereas if it's done client-side the user has full control and there are no privacy concerns.

a malicious hoster could modify the client code, just as easily as the server code, given that it's served from the same source

PR #4187, opened 2021-04-03
Add proper filename to the content-disposition header

Nachtalb:
By default we take the slug given in the request, if this is not set we try to give a filename from known types.
If we still have no filename we fallback to the previous method of setting no filename.

If the filename is non ascii we will only create the encoded "filename*" and not the ascii only "filename". This is to prevent other applications to save a file like "?????.png" if the filename contains non ascii chars.

For the browsers nothing will really change compared to the behaviour before this change as good fallbacks if no content-disposition filename is set. But that is not the case for all application, thus it makes sense to include the proper way to set the filename.

Examples: 
```
# Non-ASCII
http://localhost:9999/uploads/3dc8f9633fa0ef19/るしあちゃん　twitter落書き　まとめ_84551246_p1.jpg
inline; filename*=UTF-8''%E3%82%8B%E3%81%97%E3%81%82%E3%81%A1%E3%82%83%E3%82%93%E3%80%80twitter%E8%90%BD%E6%9B%B8%E3%81%8D%E3%80%80%E3%81%BE%E3%81%A8%E3%82%81_84551246_p1.jpg

# ASCII
http://localhost:9999/uploads/3dc8f9633fa0ef19/some_ascii_name.jpg
inline; filename="some_ascii_name.jpg"

# No name
http://localhost:9999/uploads/3dc8f9633fa0ef19/
inline; filename="image.png"

# Not known type without a name
http://localhost:9999/uploads/516bb50c9388dc1c/
attachment
```

### DIFF
--- a/src/plugins/uploader.js
+++ b/src/plugins/uploader.js
@@ -10,26 +10,27 @@ const readChunk = require("read-chunk");
 const crypto = require("crypto");
 const isUtf8 = require("is-utf8");
 const log = require("../log");
+const contentDisposition = require("content-disposition");
 
-// List of allowed mime types that can be rendered in browser
-// without forcing it to be downloaded
-const inlineContentDispositionTypes = [
-	"application/ogg",
-	"audio/midi",
-	"audio/mpeg",
-	"audio/ogg",
-	"audio/vnd.wave",
-	"image/bmp",
-	"image/gif",
-	"image/jpeg",
-	"image/png",
-	"image/webp",
-	"image/avif",
-	"text/plain",
-	"video/mp4",
-	"video/ogg",
-	"video/webm",
-];
+// Map of allowed mime types to their respecive default filenames
+// that will be rendered in browser without forcing them to be downloaded
+const inlineContentDispositionTypes = {
+	"application/ogg": "media.ogx",
+	"audio/midi": "audio.midi",
+	"audio/mpeg": "audio.mp3",
+	"audio/ogg": "audio.ogg",
+	"audio/vnd.wave": "audio.wav",
+	"image/bmp": "image.bmp",
+	"image/gif": "image.gif",
+	"image/jpeg": "image.jpg",
+	"image/png": "image.png",
+	"image/webp": "image.webp",
+	"image/avif": "image.avif",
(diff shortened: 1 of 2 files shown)


src/plugins/uploader.js
```diff
@@ -10,26 +10,27 @@ const readChunk = require("read-chunk");
 const crypto = require("crypto");
 const isUtf8 = require("is-utf8");
 const log = require("../log");
+const contentDisposition = require("content-disposition");
 
 // List of allowed mime types that can be rendered in browser
 // without forcing it to be downloaded
```
brunnre8, line 16: This needs an updated comment.
Reading this, it's hard to see what the map is for and what the mapping is supposed to do, considering that the comment says what it says (it's not a list any longer).

src/plugins/uploader.js
```diff
@@ -92,17 +93,28 @@ class Uploader {
...
 		// Force a download in the browser if it's not an allowed type (binary or otherwise unknown)
-		const contentDisposition = inlineContentDispositionTypes.includes(detectedMimeType)
-			? "inline"
-			: "attachment";
+		let slug = req.params.slug;
+		let disposition =
+			detectedMimeType in inlineContentDispositionTypes ? "inline" : "attachment";
+
+		if (!slug && detectedMimeType in inlineContentDispositionTypes) {
+			slug = inlineContentDispositionTypes[detectedMimeType];
```
brunnre8, line 101: having the same lookup twice (`detectedMimeType in inlineContentDispositionTypes`) in such a short order seems strange to me... why not extracting this to a const so that the thing is obvious?

src/plugins/uploader.js
Nachtalb: tbh I didn't even notice the comment lel, will change it

src/plugins/uploader.js
Nachtalb: I don't really care tbh. Gonna change it :)

Nachtalb:
Update code according to review and added flac inline audio support

MaxLeiter (approved):

PR #4189, opened 2021-04-03
Connect on lobby disconnected icon click

Nachtalb:
image
Us the little disconnected icon to reconnect instead of "right-click => connect"

### DIFF
--- a/client/components/NetworkLobby.vue
+++ b/client/components/NetworkLobby.vue
@@ -20,13 +20,15 @@
 			>
 				<span class="not-secure-icon" />
 			</span>
-			<span
+			<button
 				v-if="!network.status.connected"
 				class="not-connected-tooltip tooltipped tooltipped-w"
-				aria-label="Disconnected"
+				aria-label="Reconnect"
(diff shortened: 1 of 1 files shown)


brunnre8:
I'm not really fond of this UI wise... seems counter intuitive to have a button that is a disconnect indicator to reconnect.

Considering that this isn't a case that happens often (only if your network really had issues and the auto reconnect failed) I'd prefer the right click connect menu

Nachtalb:
I had this done cuz it's hella annoying in a testing/dev setup to always use right-click. Also, it does seem like a button IMHO. My initial reaction to seeing this was to click on it to reconnect. It is next to a working button and has a symbol without any text... in other words, it looks interactive. 

Secondly I don't see the harm in having it as a button.

MaxLeiter:
This was intended as a status icon and I'd rather not make the change without some good reason. It's pretty rare you need to re-connect to a network, and the insecure icon isn't a button so it wouldn't be very consistent. For example, can you click the disconnected icons next to channel names?

PR #4191, opened 2021-04-03
Networks specific media preview disabler blacklist

Nachtalb:
Disable media previews based on text in a message, by the sender or by the channel. 

Like this you can eg. hide all media previews that contain the URL or disable all pics in a channel that includes NSFW content. 

Disabling here means showing them closed. The user is still able to open them afterwards.


src/plugins/irc-events/link.js
```diff
@@ -207,11 +207,22 @@ function disablePreviewDueBlacklist(msg, chan, network) {
...
+	}
+
 	for (const query of network.mediaPreviewBlacklist) {
 		if (
 			(query[0] === "#" && query === channelName) ||
 			(query[0] === "@" && query.slice(1) === sender) ||
-			(!["@", "#"].includes(query[0]) && text.indexOf(query) !== -1)
+			(!["@", "#"].includes(query[0]) && text.indexOf(query) !== -1) ||
+			(query[0] === "/" && testAgainstRegex(query, text))
 		) {
```
brunnre8, line 226: Can we restructure this so that we parse the blacklist exactly once?
As is, we parse the regex on every single message that contains a url.
Also there's no way to tell the user that the regex the entered is wrong (we are simply ignoring the error here).

In my view it would be better to parse this exactly once, when the settings are entered. This allows us to give an error message back to the user for anything that starts with the regex trigger but doesn't parse.

brunnre8:
Thanks a lot for the work, you make review surprisingly easy with the way you are structuring the commits, much appreciated.

Besides the code review I'd like to start some general discussions in regards to the functionality itself.

**How easy should TL be**?
Xpaw repeatedly rejected features that added complexity to the client.
I'm a techy, so I like the proposed feature of auto hiding stuff without doing it in CSS, however it does add considerable complexity as users need to understand the parsing logic and we aren't exactly consistent there (we would have different ways of handling custom highlights, highlight exceptions and the blacklist).
Blocking previews per channel or nick is already possible today via CSS, so the added feature would be easier configuration and the matching based on the message content.

**Security considerations**
Now, excuse me if my understanding is wrong, but the regex matching as suggested in the PR is server side yes?
If not disregard this section completely.

Assuming it is server side, this is a DOS vector: https://en.wikipedia.org/wiki/ReDoS
Considering that both the regex and the content of the parsed message is controlled by the user they can use up considerable resources on the server, effectively DOSing it.
This is only an issue on those instances that don't trust their users, but we shouldn't make it trivial to do so I think. Even if the user is not actively malicious this can still be an issue when they mess up the regex.

What are your opinions there?

fnutt:
Thanks for all the pull requests you've been doing lately. Much appreciated!

Regarding this pull request:
I do like the way xPaw thought about the client - keep it simple, easy to use. The Lounge is a modern web IRC client which should fit most people, but not adding the complexity to fit the few people that wants 100000-settings.

If this can be made "right-click-to-ignore-previews" and reverse, please go ahead. If not I think this is overly complicated and only a very few people will use it.

Nachtalb:
As discussed in the IRC  channel I replaced the textarea with a context menu entry. But I still think we should have some way to hide media previews by keywords. It doesn't necessarily need regex functionality, but so that you at least can filter out stuff like NSFW media. Often it IRC channels have rules where you have to mark certain content with keywords like nsfw when sending images.

MaxLeiter:
@Nachtalb one potential option is to provide a new client setting (only toggleable if either `auto-expand websites|media` is enabled), "Don't expand links with messages containing `NSFW`". This way we don't need a regular expression and avoid what @brunnre8 mentioned. I'm not a huge fan of adding more settings though.

brunnre8:
>one potential option is to provide a new client setting (only toggleable if either `auto-expand websites|media` is enabled), "Don't expand links with messages containing `NSFW`".

Hard coded? That sounds strictly worse than having  text matching from a free text field. As it's the case prior to the regex commit, which by now I think is off the table.

fnutt:
Please change mediaPreviewBlacklist to something more suited in 2021.

Suggestions: mediaPreviewExcludeList, mediaPreviewBlockList

PR #4192, opened 2021-04-04
Restrict what the browser should try to autocomplete

Nachtalb:
Browsers often autocomplete into wrong fields. Eg. wanting to put the password in a fields in the settings screen and then use a sudo random other fields for username etc.
This is rather annoying and can break someones configuration, thus we should only enable it on fields where it somewhat makes sense (name, server & password fields).

### DIFF
--- a/client/components/Windows/Settings.vue
+++ b/client/components/Windows/Settings.vue
@@ -3,7 +3,13 @@
 		<div class="header">
 			<SidebarToggle />
 		</div>
-		<form ref="settingsForm" class="container" @change="onChange" @submit.prevent>
+		<form
+			ref="settingsForm"
+			class="container"
+			autocomplete="off"
+			@change="onChange"
+			@submit.prevent
+		>
 			<h1 class="title">Settings</h1>
 
 			<div>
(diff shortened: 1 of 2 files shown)


client/components/Windows/Settings.vue
```diff
@@ -409,6 +415,7 @@ your nickname or expressions defined in custom highlights."
 					<RevealPassword v-slot:default="slotProps">
 						<input
 							id="old_password_input"
+							autocomplete="on"
 							:type="slotProps.isVisible ? 'text' : 'password'"
```
brunnre8, line 419: while we are at it, can we please give the proper hint for the thing we actually want? https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#:~:text=current-password in that case, and the other password fields should probably be `new-password` so that password manager trigger.

Nachtalb: Neat, didn't know this existed.

MaxLeiter (changes requested):

client/components/Windows/Settings.vue
```diff
@@ -408,7 +414,8 @@ your nickname or expressions defined in custom highlights."
 					<label for="old_password_input" class="sr-only"> Enter current password </label>
```
MaxLeiter, line 414: `for="current-password"`

client/components/Windows/Settings.vue
```diff
@@ -434,9 +442,10 @@ your nickname or expressions defined in custom highlights."
 					<label for="verify_password_input" class="sr-only"> Repeat new password </label>
```
MaxLeiter, line 442: `for="new-password-verify"`

MaxLeiter (changes requested):

MaxLeiter (approved):

PR #4195, opened 2021-04-06
Settings: show label for nick autocompletion postfix

brunnre8:
Without a visible label it's impossible to determine what the option does, as the placeholder is not visible if there's anything in the input field.

The label now looks like the other options like the highlight settings:
example


MaxLeiter:
Thanks

PR #4197, opened 2021-04-13
Message Search: Re-Rebase + Fixes / Adjustments

Nachtalb:
This builds on top of https://github.com/thelounge/thelounge/pull/4148 which builds on top of https://github.com/thelounge/thelounge/pull/3664;

My commits should be fairly self-explanatory as to what they do, but here some more info:

Why did I not do implemented this and that and the other thing?
---

We don't need a feature-complete search yet. 

If we do all that a search should be able to do in one single big af PR, it's never gonna get merged and released. No one likes to review big PRs (and this one is already at the limit). Additionally, after a few weeks, it's already outdated again and has to be rebased over and over again. Thus I tried to just fix only crucial bugs as well as make it somewhat more usable so that we can merge it and continue on. 

I suggest that we merge this in the near future and start a new project in Github "Projects" tab to improve the search. @cottongin had some good ideas in the first PR and there are many things missing: search actions (ban, leave etc.), from-user filter, global search, network search, full text (eg. FTS5) or fuzzy search, speed optimisations... We can do this after we have an initial search that just simply searches messages.

Main changes compared to previous mentioned PRs:
---

- > Need to handle the case when sqlite is disabled.
  > ~ https://github.com/thelounge/thelounge/pull/3664#discussion_r422520401 
  
  Disable Search when we got no message provider
- > `searchNicks` should use AND instead of OR 
  > ~ https://github.com/thelounge/thelounge/pull/3664#discussion_r397098534 
 
  Remove `searchNick` as we have no filters/facets implemented yet
- > 1 Clicking on a username or channel throws error
  > ~ https://github.com/thelounge/thelounge/pull/3664#issuecomment-603200143 (2,3,5 already fixed)

  Correctly set active channel. Fixes any username/channel links as well as highlights current channel in the nav
- Replace `/search/network-uuid/channel/search terms` with `/search/network-uuid/channel?q=search terms`
  The search term is not an asset/resource but a dynamic string, so it should be a query.
- Keep search term in the search input field when first opening the search.
- > 4 The search box on default theme stands out and doesn't match any other existing fields color wise.
  > ~ https://github.com/thelounge/thelounge/pull/3664#issuecomment-603200143 (2,3,5 already fixed)

  Various restyling to fit it more to our current theme as well as usability wise

Opened Search
---
image

Search View
---
image


src/client.js
```diff
@@ -72,7 +73,8 @@ function Client(manager, name, config = {}) {
 
 	if (!Helper.config.public && client.config.log) {
 		if (Helper.config.messageStorage.includes("sqlite")) {
-			client.messageStorage.push(new MessageStorage(client));
+			client.messageProvider = new MessageStorage(client);
```
MaxLeiter, line 76: The user can have multiple message providers active at once, so this won't work. I think we should sent `searchPossible` (or some equivalent) in the `init` event when a client connects. What do you think of that, @Nachtalb?

src/models/network.js
```diff
@@ -189,7 +189,7 @@ Network.prototype.createIrcFramework = function (client) {
 
 	// Request only new messages from ZNC if we have sqlite logging enabled
 	// See http://wiki.znc.in/Playback
-	if (client.config.log && client.messageStorage.find((s) => s.canProvideMessages())) {
+	if (client.messageProvider) {
```
MaxLeiter, line 192: Revert this so we maintain support for multiple message storage plugins

MaxLeiter (changes requested):

src/client.js
Nachtalb: At the moment the only message storages are sqlite and text from which sqlite is the only provider. I wanna implement Postgress later on end this would need a rework anyways. IMHO for the time being it's ok.

src/models/network.js
Nachtalb: Refer to https://github.com/thelounge/thelounge/pull/4197#discussion_r612052240

fnutt:
Nachtalb: Please take a look at the commits done march 5th on this branch: https://github.com/thelounge/thelounge/commits/richrd/message-search-rebase

We probably want some or all of those fixes in this PR as well. :)

Nachtalb:
@fnutt As I have built this PR on top of https://github.com/thelounge/thelounge/pull/4148 which is based on the original https://github.com/thelounge/thelounge/pull/3664 I already have these commits. eg: [`f323653` (#4197)](https://github.com/thelounge/thelounge/pull/4197/commits/f3236538a014c86e32117186cdce692bb105b5d1)

client/js/router.js
```diff
@@ -63,6 +64,11 @@ const router = new VueRouter({
 			path: "/chan-:id",
 			component: RoutedChat,
 		},
+		{
+			name: "SearchResults",
+			path: "/search/:uuid/:target",
```
MaxLeiter, line 69: Can we use `chan-:id/search`? Once [maxleiter/improvedURLs](https://github.com/thelounge/thelounge/tree/maxleiter/improvedURLs) is merged we'll need to change it again, but this makes more sense than throwing the UUID in the address bar, imo.

MaxLeiter (changes requested):
Seems to work well! Can you add a button for closing the Search? Once that + my nitpick comment are addressed I think we'll be good to go after some dogfooding 🎉

client/js/store.js
```diff
@@ -17,7 +17,23 @@ function detectDesktopNotificationState() {
 	return "blocked";
 }
 
-const store = new Vuex.Store({
+let store = null;
+
+const setMessageNetworkChannel = (message) => {
```
MaxLeiter, line 22: I would rather this not exist in `store.js`, or have it move elsewhere in the file and leave `store` as const

src/client.js
```diff
@@ -106,6 +108,8 @@ function Client(manager, name, config = {}) {
 		client.awayMessage = client.config.clientSettings.awayMessage;
 	}
 
+	client.config.clientSettings.searchEnabled = client.messageProvider !== undefined;
```
MaxLeiter, line 111: `&& client.config.log`

MaxLeiter: I can't read 🤷

Nachtalb:
@MaxLeiter I reverted the unneeded changes to the store.js, changed the search so it properly stays in the channel with /chan-:id/search as well as add the close button.

richrd:
@Nachtalb many thanks for resurrecting this! I went through it and it looks really good! Let's finally get this in :)

richrd (approved):

Nachtalb:
20180716084317-b81a7a09

MaxLeiter (approved):

brunnre8:
@NightMachinary please don't comment on random bugs with a me too comment or support request, that's not what the issue tracker is for.

Hop into IRC, #thelounge on libera.chat and we'll sort you out

PR #4205, opened 2021-04-30
Only scroll history when cursor is on first or last row

Nachtalb:
- Needs to be on first to go up and on last to go down
- Preserve position when scrolling through input history

https://user-images.githubusercontent.com/9467802/116762041-85c33800-aa19-11eb-95e8-eb8f5996ea0b.mp4


MaxLeiter:
This works great. thanks!

PR #4206, opened 2021-04-30
Fill inputhistory on channel load and more message load

Nachtalb:
as it says in the title. Pretty simple but worth a ton :)

### DIFF
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -6,7 +6,7 @@ import socket from "../socket";
 import store from "../store";
 
 socket.on("more", function (data) {
-	const channel = store.getters.findChannel(data.chan).channel;
+	const channel = store.getters.findChannel(data.chan)?.channel;
 
 	if (!channel) {
 		return;
--- a/client/js/socket-events/more.js
+++ b/client/js/socket-events/more.js
@@ -6,17 +6,17 @@ import socket from "../socket";
 import store from "../store";
 
 socket.on("more", function (data) {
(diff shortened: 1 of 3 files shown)


client/js/socket-events/more.js
```diff
@@ -6,17 +6,24 @@ import socket from "../socket";
 import store from "../store";
 
 socket.on("more", function (data) {
-	const channel = store.getters.findChannel(data.chan);
+	const channel = store.getters.findChannel(data.chan).channel;
```
xPaw, line 9: Do not do this, findChannel _can_ return null.

Nachtalb: Ahh RIP thx, I'll change it

Nachtalb: @xPaw would you agree to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```suggestion
	const channel = store.getters.findChannel(data.chan)?.channel;
```

xPaw: If webpack/babel can deal with it, why not.

Nachtalb: Should I lift up the `ecmaVersion` to `2020` or use a comment to tell eslint to ignore it? ==> https://eslint.org/blog/2020/07/eslint-v7.5.0-released#optional-chaining-support

MaxLeiter: Yeah, let's bump it to 2020

MaxLeiter (approved):

PR #4210, opened 2021-05-06
Improve inline audio file support

Nachtalb:
Tested on latest Chromium / Firefox. In case of .m4a files they want audio/x-m4a and not audio/m4a, in case of .flac files they want audio/flac and not audio/x-flac. The module we used to detect the types however detects them only as audio/x-m4a and audio/x-flac as they are not official IANA supported mime types (not in IANA spec == "x-" prefix): https://www.iana.org/assignments/media-types/media-types.xhtml Though flac is not in the IANA spec many programs such as the file command (https://man7.org/linux/man-pages/man1/file.1.html) and Chromium (flac) / Firefox (x-flac and flac) support audio/flac only or both.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -249,6 +249,8 @@ function parse(msg, chan, preview, res, client) {
 		case "audio/x-midi":
 		case "audio/x-mpeg":
 		case "audio/x-mpeg-3":
+		case "audio/flac":
+		case "audio/x-m4a":
 			if (!preview.link.startsWith("https://")) {
 				break;
 			}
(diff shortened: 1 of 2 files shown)


MaxLeiter (approved):

xnaas:
Unless the FLAC is uploaded directly to thelounge, the behavior is all over the place. Most FLACs show no attempt at embedding whatsoever, but some will show a filesize and no player...

Some file links you're welcome to try (tested on Firefox and Edge):

* https://p.actionsack.com/a/Gillette.flac — no embed
* https://irc.xnaas.info/uploads/830e7aaa7395cd4e/Gillette.flac — works as expected
* https://filesamples.com/samples/audio/flac/Symphony%20No.6%20(1st%20movement).flac — no embed
* http://www.lindberg.no/hires/test/2L-145/2L-45_stereo_01_FLAC_88k_24b.flac — displays a filesize, but no player

(the m4a stuff all seems to work fine)

xnaas:
I was actually able to get https://p.actionsack.com/a/Gillette.flac to embed by editing nginx's `mime.types`, but that seems to also cause some other weird behavior (outside of thelounge), so...you know...YMMV, I guess.

Issue #4161, opened 2021-02-21
Apply a class to channel list items when they have unread messages/pings


andrewphorn:
For example, so that themes can make style changes to channels with unread pings or messages. For extra credit, it could apply an additional class if there are unread pings.

My use case: embolden channel names when there are unread messages. In weechat I used bolded text to indicate unread, and changed the color based on whether there were unread pings. Made it very easy to find channels quickly without having to look at the much smaller badges.
![image](https://user-images.githubusercontent.com/1857241/108636359-8cda3280-7439-11eb-928c-02f4731ee2ed.png)

As it is right now, since there are no supported previous-adjacency functions yet, this is not possible as far as I can tell.

PR #4214, opened 2021-05-06
Classes for channels in list with unread counts and highlights

sha1sum:
This adds `has-unread` class to the channel classes in the list if the channel has unread messages, and `has-highlight` if the channel has an active highlight.

Resloves #4161


MaxLeiter:
Thanks!

PR #4215, opened 2021-05-06
 Mark search query in result messages

Nachtalb:
image

image

Also, fix search input overlap.


Nachtalb:
I'll close that for now in hope to find a better approach

PR #4218, opened 2021-05-07
Print the response time between two clients with CTCP PING

Nachtalb:
According to [spec](https://tools.ietf.org/id/draft-oakley-irc-ctcp-01.html#rfc.appendix.A.5), PING is used as a reachability check and to measure the time
delay between two clients. Other clients such as IRSSI and HexChat give
the user an easy way to check this delay with `/ping {nick}`

As such we implement this here as well. In the background `/ping {nick}` will
do a simple `/ctcp {nick} PING {timestamp}` to which we will print the
time delay as an answer.

> **A.5. PING**
> This extended query is used to confirm reachability with other clients and to check latency.
> ~ https://tools.ietf.org/id/draft-oakley-irc-ctcp-01.html#rfc.appendix.A.1


client/components/Windows/Help.vue
```diff
@@ -623,6 +623,24 @@
...
 			</div>
 
+			<div class="help-item">
+				<div class="subject">
+					<code>/ping nick</code>
+				</div>
+				<div class="description">
+					<p>
+						Send a <abbr title="Client-to-client protocol">CTCP</abbr> PING request and
+						calculate delay between the clients. Read more about CTCP PING in this
```
brunnre8, line 633: except that's not what we are doing...
We don't measure the delay between the clients but from the TL server, via the irc server to the client and back to the TL server

In essence that
```
[thelounge client] --> [thelounge server] __timer starts__ --> [IRC server] --> [victim] --> [thelounge server] __timer stops__
```

The wording here (and on wikipedia for that matter) is simply misleading, as the communication is asymmetrically via the IRC server.

It's already to have it as a gimmick I guess, but at least mention what it does without people having to look at the code ;)

src/client.js
```diff
@@ -64,6 +64,7 @@ function Client(manager, name, config = {}) {
 		highlightRegex: null,
 		highlightExceptionRegex: null,
 		messageProvider: undefined,
+		expectedPings: {},
```
brunnre8, line 67: this seems to be the wrong level... we need to track such things per the network they originate from and not globally per client

src/plugins/irc-events/ctcp.js
```diff
@@ -35,6 +65,13 @@ module.exports = function (irc, network) {
 			chan = lobby;
 		}
 
+		const action = ctcpResponseActions[data.type];
+		const prventFurtherAction = action ? action(data, chan) : null;
+
+		if (prventFurtherAction) {
+			return;
+		}
+
```
brunnre8, line 74: this reads strange... you have a perfectly healthy if statement, why do you use an additional ternary operator?

PR #4221, opened 2021-05-09
Fix missing users in userlist after removing searchinput

### LINKED ISSUE #4220: Users disappear after searching in userlist
- _Node version:_ v12.20.0
- _Browser version:_ Microsoft Edge Version 90.0.818.56 (But also on chrome & firefox)
- _Device, operating system:_ Android (Phone), Windows (PC), Linux (PC)
- _The Lounge version:_ v4.3.0-pre.1 / v4.2.0

---
After searching for a user in the user list, users have disappeared from the list. After switching channel the users are back again.

How to reproduce:
1. Search for something in the user list
2. Empty the search box

cyzalyjcYl

Nachtalb:
Because the "Username" components still had the same ":key" vue tried to in-place update them. This doesn't quite work for objects (in this case "user" or "user.original"). Thus we change the key for the search so that it actually inits a new component and thus evaluates its content correctly.

fixes #4220

### DIFF
--- a/client/components/ChatUserList.vue
+++ b/client/components/ChatUserList.vue
@@ -28,7 +28,7 @@
 				<template v-if="userSearchInput.length > 0">
 					<Username
 						v-for="user in users"
-						:key="user.original.nick"
+						:key="user.original.nick + '-search'"
 						:on-hover="hoverUser"
 						:active="user.original === activeUser"
 						:user="user.original"


Nachtalb:
I didn't intend to close it, must have hit the close button accidentally

richrd (approved):

MaxLeiter (approved):

PR #4225, opened 2021-05-14
Add X-Frame-Options header

alekna:
Protects against clickjacking attacks.

For more information see https://scotthelme.co.uk/hardening-your-http-response-headers/#x-frame-options

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -283,6 +283,7 @@ function getClientSecure(socket) {
 
 function allRequests(req, res, next) {
 	res.setHeader("X-Content-Type-Options", "nosniff");
+	res.setHeader("X-Frame-Options", "DENY");
 	return next();
 }
 


brunnre8:
Might as well add the corresponding CSP header too https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

brunnre8:
Note however that this will interfere with legitimate usage of an embedded IRC client as well, say for conferences.
Not sure how the 2 recently performed ones handled the TL integration

alekna:
> Might as well add the corresponding CSP header too https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

It's there already:

https://github.com/thelounge/thelounge/blob/db807d0c5611872de075533280a661b1bccef32c/src/server.js#L321

brunnre8:
It's not no.

The policies we have are https://github.com/thelounge/thelounge/blob/db807d0c5611872de075533280a661b1bccef32c/src/server.js#L298-L307

plus image source, depending and some others changes a few lines below.

frame-ancestors isn't one of them.

alekna:
indeed. would `frame-ancestors 'none'` be appropriate value for it?

brunnre8:
Assuming we actually do want to prohibit the framing (and that's an if, again there may be valid use cases for i-framing an IRC client for things like personal websites or conferences) then yes.

The better question is, do we need that?
IRC isn't exactly *that* sensitive is it?

We could shift that to the responsibility of the admin if they feel paranoid, after all such headers can trivially be set by the reverse proxy.
But that also applies for the other policies we do have, doesn't it? And there's a case of having sane defaults.

alekna:
Your arguments convinced me there's no need for that.

PR #4230, opened 2021-05-19
Chat input link preview 

Nachtalb:
Show link preview like in other instant messengers like Telegram. For each link, we try to get a preview (with our usual method) and show them sorted in a quick popup above the chat input. We reuse the already existing previews with slightly different size boundaries.  

localhost_9990_
localhost_9990_ (3)
localhost_9990_ (2)
localhost_9990_ (1)

https://user-images.githubusercontent.com/9467802/118892969-ff0bc780-b901-11eb-95ec-e56d54ff99f6.mp4

### DIFF
--- a/client/components/ChatInput.vue
+++ b/client/components/ChatInput.vue
@@ -1,5 +1,14 @@
 <template>
 	<form id="form" method="post" action="" @submit.prevent="onSubmit">
+		<div v-if="channel.pendingMessagePreviews.length > 0" id="input-previews">
+			<LinkPreview
+				v-for="preview in channel.pendingMessagePreviews"
+				:key="preview.link"
+				:keep-scroll-position="() => {}"
+				:link="preview"
+				:channel="channel"
+			/>
+		</div>
 		<span id="upload-progressbar" />
 		<span id="nick">{{ network.nick }}</span>
 		<textarea
@@ -59,6 +68,9 @@ import commands from "../js/commands/index";
 import socket from "../js/socket";
 import upload from "../js/upload";
 import eventbus from "../js/eventbus";
+import {findLinks} from "../js/helpers/ircmessageparser/findLinks";
+import throttle from "lodash/throttle";
+import LinkPreview from "./LinkPreview.vue";
 
 const formattingHotkeys = {
 	"mod+k": "\x03",
@@ -89,6 +101,9 @@ let autocompletionRef = null;
 
 export default {
 	name: "ChatInput",
+	components: {
+		LinkPreview,
+	},
 	props: {
 		network: Object,
 		channel: Object,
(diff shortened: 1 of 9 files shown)


richrd:
I love the fact that this works for image uploads! For me personally that is probably the most compelling use case. All in all looks like a nice feature. I'd like to hear other peoples views on this.
And also thanks for providing the video preview, helps a lot to see how this feature actually works.

termermc:
I like this, although preferably it is something that can be disabled for people who don't want it

brunnre8:
I'd couple it to the link previews, if those are enabled chances are that people want to see those too.

As for the feature itself, the url preview is really nice.
The pictures need to be styled a bit differently I think, it's not exactly looking like they belong to the input field (they don't integrate like the url bubbles do in my opinion)

Nachtalb:
> I'd couple it to the link previews, if those are enabled chances are that people want to see those too.

I totally forgot about that. Had plans to add it actually rip. 😂 

> The pictures need to be styled a bit differently I think, it's not exactly looking like they belong to the input field (they don't integrate like the url bubbles do in my opinion)

I just used the default preview. It would change a bit anyways with the filename preview. Maybe the filename preview would make it look more "integrated"

fnutt:
There's something funky going on with this PR. I've managed to make the url preview "stuck" right above chat input several times now. It seems like it happens if I press enter right before the url preview has been loaded and viewed above chat input, but it's not easy to replicate.

fnutt:
Bilde_2021-06-17_092928

progval:
@fnutt's issue happens when deleting the link after sending the request to the server but before the server sends the `input:preview` event.

<details><summary>it can be reproduced with this script, which spawns an HTTP server which always takes 2s to reply (click to unfold)</summary>

```python
import random
import time
import http.server
import socketserver

PORT = 12345

URL = "http://"  # replace this

class Handler(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        time.sleep(2)
        self.send_response(200)
        self.send_header("Content-Type", "text/html")
        self.end_headers()
        self.wfile.write(b"<title>test title</title><meta name='description' content='simple description'><meta property='og:title' content='opengraph test'><meta property='og:image' content='" + URL.encode() + b"?" + random.randbytes(20).hex().encode() + b"'>")


with socketserver.TCPServer(("0.0.0.0", PORT), Handler) as httpd:
    assert not httpd.allow_reuse_address
    httpd.allow_reuse_address = True
    print("serving at port", PORT)
    httpd.serve_forever()
```
</details>

It can be resolved with this fix:

```diff
diff --git a/client/js/socket-events/input_preview.js b/client/js/socket-events/input_preview.js
index edfee664..046b6a4b 100644
--- a/client/js/socket-events/input_preview.js
+++ b/client/js/socket-events/input_preview.js
@@ -14,6 +14,11 @@ socket.on("input:preview", function (data) {
 
 	const links = findLinks(channel.pendingMessage).map((obj) => obj.link);
 
+	if (!links.includes(data.preview.link)) {
+		// Link was removed from input box while fetching it
+		return;
+	}
+
 	channel.pendingMessagePreviews.push(data.preview);
 	channel.pendingMessagePreviews.sort((a, b) => links.indexOf(a.link) - links.indexOf(b.link));
 });
```

Issue #4200, opened 2021-04-20
Channel Hotkey Navigation with Screen Reader


LizHareDogs:
### Feature Description

Hi again,

Can theLounge notify the screen reader of the new channel name when using the hotkey (option-up/down arrow on Mac) takes you to the next/previous channel?

Now, when I change channels, I have to go back linearly through everything on the page to get back up to where the channel name is displayed.

Thanks,
Liz

PR #4231, opened 2021-05-20
Add shorcut to navigate between channels with undread msgs

Nachtalb:
IMHO this makes `alt`+`a` obsolete and is easier to familiarize with as the other navigation shortcuts also go directed through the list and not solely from the top down. 

https://user-images.githubusercontent.com/9467802/119032666-bbba6300-b9ac-11eb-90c8-711f8e0b03d6.mp4

### DIFF
--- a/client/js/keybinds.ts
+++ b/client/js/keybinds.ts
@@ -83,6 +83,35 @@ Mousetrap.bind(["alt+shift+up", "alt+shift+down"], function (e, keys) {
 	return false;
 });
 
+// Switch to the next/previous unread chat
+Mousetrap.bind(["alt+mod+up", "alt+mod+down"], function (e, keys) {
+	if (isIgnoredKeybind(e)) {
+		return true;
+	}
+
+	const channels = store.state.networks
+		.map((net) =>
+			net.channels.filter(
+				(chan) => chan.unread || chan === store.state.activeChannel?.channel
+			)
+		)
+		.flat();
+
+	if (channels.length === 0) {
+		return;
+	}
+
+	let index = channels.findIndex((chan) => chan === store.state.activeChannel?.channel);
+
+	const length = channels.length;
+	const direction = keys.split("+").pop() === "up" ? -1 : 1;
+	index = (((index + direction) % length) + length) % length;
+
+	jumpToChannel(channels[index]);
+
+	return false;
+});
+
 // Jump to the first window with a highlight in it, or the first with unread
 // activity if there are none with highlights.
 Mousetrap.bind(["alt+a"], function (e) {
--- a/client/components/Windows/Help.vue
+++ b/client/components/Windows/Help.vue
@@ -181,6 +181,26 @@
 
 			<div class="help-item">
 				<div class="subject">
+					<span v-if="!isApple"><kbd>Alt</kbd> <kbd>Ctrl</kbd> <kbd>↓</kbd></span>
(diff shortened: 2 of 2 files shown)


brunnre8:
LGTM

MaxLeiter:
@Nachtalb can you also look into fixing #4200?

client/js/keybinds.js
```diff
@@ -78,6 +78,33 @@ Mousetrap.bind(["alt+shift+up", "alt+shift+down"], function (e, keys) {
 	return false;
 });
 
+// Switch to the next/previous unread chat
+Mousetrap.bind(["alt+mod+up", "alt+mod+down"], function (e, keys) {
```
MaxLeiter, line 82: @Nachtalb can we test this? If you extract it to a new function it should be easy to call and mock jumpToChannel to make sure the right channel is passed.

MaxLeiter (changes requested):

client/js/keybinds.js
Nachtalb: Quite some time since I worked on this. What did you want me to do exactly, put it in its own function so we can write a test for it? If so, where should I put the function?

client/js/keybinds.ts
```diff
@@ -83,6 +83,35 @@ Mousetrap.bind(["alt+shift+up", "alt+shift+down"], function (e, keys) {
 	return false;
 });
 
+// Switch to the next/previous unread chat
+Mousetrap.bind(["alt+mod+up", "alt+mod+down"], function (e, keys) {
```
brunnre8, line 87: hm, on a second thought... I assume "mod" is the windows key?
Can we leave that one alone and use Ctrl+Alt+{up,down} ?

Windows is frequently already used by the window manager and executes global hotkeys rather than local ones.

Would also make it symmetrical with the rest

* Alt+key == window
* Alt+shift+key == lobby
* Alt+ctrl+key == activity

Bad idea? What made you pick mod?

client/js/keybinds.js
brunnre8: > put it in its own function so we can write a test for it?

Guess that'd be the idea yeah...
As for where to put it... we could make it a method of the client, rather than open coding it in the keybinds.
Can't put it at the network level as we want to jump between them, so that'd be the next one up.

client/js/keybinds.ts
Nachtalb: windows key would be `meta`

`mod` is `ctrl` on windows and `cmd` on osx. 
so on windows this is `alt+ctrl+up/down` and on mac this is `opt+cmd+up/down`

client/js/keybinds.js
Nachtalb: no worries. Haven't touched this project in years. So I can't remember all the stuff either. XD

client/js/keybinds.ts
brunnre8: ack, thanks for the explanation

client/js/keybinds.js
Nachtalb: If I were to put it in its own function rather than as it is now. We should also do this for the rest of the keybinds. This should be done in its own PR thought. To keep everything consistent. Doing this in this PR. will make it grand instead of as small as PR should be.

brunnre8: IIRC our client tests are also broken since the TS rewrite... so it'll be some effort we shouldn't punt on this

brunnre8:
Considering that our test framework for the client is borken I'll overrule Max here.
We'll fix it in post :wink:

PR #4235, opened 2021-05-22
Render styling for colored host masks

angerson:
On some IRC networks, users have vanity host masks with colors or other text styling.
Rizon is one such network.

For example, a user connecting from `127.0.0.1` could instead have the host 
`angerson@this.is.my.host.mask`. `this.is.my.host.mask` may have IRC color code
characters in it, which without this change would be displayed as a bunch of jumbled
garbage in the `/whois` response or join/part messages.

Resolves https://github.com/thelounge/thelounge/issues/4232.

### DIFF
--- a/client/components/MessageTypes/chghost.vue
+++ b/client/components/MessageTypes/chghost.vue
@@ -6,17 +6,20 @@
 			>username to <b>{{ message.new_ident }}</b></span
 		>
 		<span v-if="message.new_host"
-			>hostname to <i class="hostmask">{{ message.new_host }}</i></span
-		>
+			>hostname to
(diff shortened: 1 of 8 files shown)


angerson:
Most of the other components (joins, parts, etc.) also lack this styling. Would the correct thing to do be to wrap every `{{ message.hostmask }}` with a `ParsedMessage`? I don't think a more abstract solution is necessary, because it's only hostmasks that are likely to be like this.

MaxLeiter:
Thanks for the PR, @angerson! It seems like there's a small linting error in `client/components/MessageTypes/chghost.vue`. Can you address that and squash your commits?

angerson:
Done, thanks @MaxLeiter.

Issue #4194, opened 2021-04-06
`/quit` behaves differently from `/disconnect`


optix2000:
- _Node version:_ v12.18.3
- _Browser version:_ Chrome 89.0.4389.114
- _Device, operating system:_ Debian bullseye
- _The Lounge version:_ v4.2.0

---

This may be a different bug depending on what the intent is.

1) The documentation for `/quit` and `/disconnect` are the same. But not identical.
```
/disconnect [message]	
Disconnect from the current network with an optionally-provided message.
/quit [message]	
Disconnect from the current network with an optional message.
```
2) `/quit` will delete the server instead of just disconnecting (and leaving the server on the server list).

If the _documentation_ is canon, then `/quit` should just be merged with `/disconnect` as the same command and the documentation should be made to read the same.

If the _code_ is canon, then the documentation needs to be updated. Though I'd be curious to know why, since most clients will treat `/quit` and `/disconnect` the same.

---
Other notes:

Looks like `/disconnect` was forked off as a separate command here: https://github.com/thelounge/thelounge/commit/047a79ead6b804d5b85bd91b5aada5c3ea3686d1
But `/quit` has some extra code to delete the network here (which has apparently been there for years): https://github.com/thelounge/thelounge/blob/8dd9bc0e9813af29fd1a42a33f19c7ffea820e26/src/plugins/inputs/quit.js#L12

fnutt:
If I understand you correctly - this is sort of fixed in right-click context menus, where you have disconnect and remove. Maybe we should also use the same words in the client help?

"/quit [message]
Disconnect and remove the current network with an optional message."

optix2000:
> this is sort of fixed in right-click context menus, where you have disconnect and remove. Maybe we should also use the same words in the client help?
This is a non-issue with the right-click context menus, since they behave intuitively.

How this should get fixed depends on how `/quit` is _supposed_ to work. If it's meant to delete/remove the server, which I think is the wrong choice, then the docs/help strings should be fixed.
If `/quit` is supposed to do what it says in the docs, then it should be merged with `/disconnect`, and should _not_ delete the server.

dgw:
QUIT is a standard IRC command. It disconnects your client from the network (ircd). I've not used any other client in which `/quit` removed the entire network from the settings.

Some clients only show active networks, yes, but QUITting still leaves the connection settings intact so it's easy to reconnect later, usually through a menu or dialog that presents a list of saved networks. Deleting those saved networks is a separate step.

[...]

PR #4236, opened 2021-05-22
Show quit dialog when using /quit

Nachtalb:
Beforehand it only showed the dialogue when quitting via the context menu. This also adds the possibility for the server to ask for dialogues.

https://user-images.githubusercontent.com/9467802/119228481-7b292980-bb13-11eb-97b2-e5c2fc9ad2b3.mp4

### DIFF
--- a/client/components/ConfirmDialog.vue
+++ b/client/components/ConfirmDialog.vue
@@ -3,7 +3,7 @@
 		<div v-if="data !== null" id="confirm-dialog">
 			<div class="confirm-text">
 				<div class="confirm-text-title">{{ data.title }}</div>
-				<p>{{ data.text }}</p>
+				<p v-html="data.text"></p>
 			</div>
 			<div class="confirm-buttons">
 				<button class="btn btn-cancel" @click="close(false)">Cancel</button>
@@ -52,6 +52,7 @@
 
 <script>
 import eventbus from "../js/eventbus";
+import socket from "../js/socket";
 
 export default {
 	name: "ConfirmDialog",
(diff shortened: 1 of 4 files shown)


MaxLeiter:
I believe we never added it to /quit because typing /quit is very deliberate, whereas it's pretty easy to fat-thumb the disconnect. I'm not against this PR, just worried it adds unnecessary complexity

Nachtalb:
> I believe we never added it to /quit because typing /quit is very deliberate, whereas it's pretty easy to fat-thumb the disconnect. I'm not against this PR, just worried it adds unnecessary complexity

The thing is you can very easily accidentally type `/quit` and hit enter due to autocomplete. `/quit` is quite a strong action as it removes all network settings and the whole message backlog without any way to undo it.

brunnre8:
ok, here's my two cents:

In my opinion we should just remove /quit entirely.
You very very rarely want to delete a network. That's a thing that doesn't need to be done quickly.

I'd rather have a single command that leaves absolutely no confusion as to what it does and we have that in `/disconnect` for the use case which might be more frequent, naming to disconnect from a network.

dgw:
#4194 seems relevant here. I'll say again, I don't think `/quit` ever _should_ have deleted all network settings from TL. It should only be an alias to `/disconnect`.

fnutt:
Some people likes commands instead of going through menus to remove networks and logs related to it. I'm perfectly fine with this PR, and would very much like to see it merged, please. It will prevent us from accidently doing something that we don't want, but still have the /quit command working as intended in the first place.

If we want to remove, adjust or find other (hopefully better...) solutions later down the road, we can do that later.

brunnre8:
I don't get that argument fnutt, when do you ever want to delete a network from the chat input?

fnutt:
When you don't want to be on the network anymore? Why keep logs for a network you are not on anymore?

Instead of right clicking on the network name -> choose "remove", you can instead type /quit.

PR #4237, opened 2021-05-22
Refactor message storages and replace node-sqlite3

Nachtalb:
Refactor message storages and replace node-sqlite3 in preparation for more refactoring and plugin functionality for message storages. 

I mainly did 3 things:
- harder checks during enabling of the storages 
- remove and cleanup code
- replace node-sqlite3 with better-sqlite3 

Reasons as to why I chose to replace node-sqlite3 with better-sqlite3:
- Which provides transaction/rollback wrapper (I heard talk on IRC that we do everything transactional... we didn't do anything transactional)
- Is synchronous, so we can test the state of the DB beforehand (instead of after we already told everyone it's ok, like wtf why did we even check for schema version when we don't do anything about mismatches??!)
- much much faster
- provides more functionality we might use in the future, you never know ¯\_(ツ)_/¯
- fewer bugs (according to better-sqlite3 team)

Also read:
https://github.com/JoshuaWise/better-sqlite3#why-should-i-use-this-instead-of-node-sqlite3
https://github.com/JoshuaWise/better-sqlite3/issues/262#issuecomment-490536241

### DIFF
--- a/client/components/Windows/SearchResults.vue
+++ b/client/components/Windows/SearchResults.vue
@@ -124,7 +124,7 @@ export default {
 				return [];
 			}
 
-			return this.search.results.slice().reverse();
+			return this.search.results.slice();
 		},
 		chan() {
 			const chanId = parseInt(this.$route.params.id, 10);
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "yarn": "1.22.10"
   },
   "optionalDependencies": {
-    "sqlite3": "5.0.2"
+    "better-sqlite3": "7.4.0"
   },
   "devDependencies": {
     "@babel/core": "7.14.0",
--- a/src/plugins/messageStorage/sqlite.js
+++ b/src/plugins/messageStorage/sqlite.js
@@ -35,19 +35,21 @@ class MessageStorage {
 	}
 
 	enable() {
+		if (!sqlite3) {
+			return false;
+		}
+
 		const logsPath = Helper.getUserLogsPath();
-		const sqlitePath = path.join(logsPath, `${this.client.name}.sqlite3`);
 
 		try {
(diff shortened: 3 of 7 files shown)


Nachtalb:
better-sqlite3 also adds a [backup](https://github.com/JoshuaWise/better-sqlite3/blob/HEAD/docs/api.md#backupdestination-options---promise) functionality which can be used as an additional safe point when we add any kind of migration (eg. adding a primary key which will pave the way for jump to message)

brunnre8:
> better-sqlite3 also adds a [backup](https://github.com/JoshuaWise/better-sqlite3/blob/HEAD/docs/api.md#backupdestination-options---promise) functionality which can be used as an additional safe point when we add any kind of migration (eg. adding a primary key which will pave the way for jump to message)

I mean, that's no argument. The run of the mill copy function that handles OS files does just the same.
Backups of the db really aren't the problem

Nachtalb:
> > better-sqlite3 also adds a [backup](https://github.com/JoshuaWise/better-sqlite3/blob/HEAD/docs/api.md?rgh-link-date=2021-05-22T23%3A19%3A42Z#backupdestination-options---promise) functionality which can be used as an additional safe point when we add any kind of migration (eg. adding a primary key which will pave the way for jump to message)
> 
> I mean, that's no argument. The run of the mill copy function that handles OS files does just the same.
> Backups of the db really aren't the problem

That's more of an FYI. Otherwise, I would have listed it in the main list in the OP... Why focus on that when it's clearly not the main reason or argument?

brunnre8:
I didn't focus on it, you made the argument, I have an opinion about it and hence made the comment

MaxLeiter:
@Nachtalb can you rebase?

PR #4251, opened 2021-06-01
install: allow installation of local packages

brunnre8:
It may not be desirable to host all plugins on npm, allow for local packages to
be installed given a package name with a `file:` prefix.

This is still more restrictive than what yarn would support but allows us to still
verify the thelounge compatibility by reading the package.json file.

`yarn add` messes up with local filepaths and generates a lockfile that is
"outdated" as far as any other yarn commands go, which makes them error out.

For some reason `yarn install` fixes that and hence we run that after an install.
Here's the diff of yarn.lock between the broken state after `yarn add file:$path`
and `yarn install`
	--- yarn.lock.2.afterAdd	2021-06-02 00:10:52.365134018 +0200
	+++ yarn.lock.3.afterinstall	2021-06-02 00:13:27.122760442 +0200
	@@ -2194,7 +2194,7 @@
	     safe-buffer "^5.1.2"
	     yallist "^3.0.3"

	-thelounge-plugin-shortcuts@/home/reto/sourcecode/thelounge-plugin-shortcuts:
	+"thelounge-plugin-shortcuts@file:../../sourcecode/thelounge-plugin-shortcuts":
	   version "1.0.12"
	   dependencies:
	     thelounge "4.2.0"

The only thing it does is switch an absolute path to a relative one for whatever
reason.

### DIFF
--- a/src/command-line/install.js
+++ b/src/command-line/install.js
@@ -13,6 +13,8 @@ program
 	.on("--help", Utils.extraHelp)
 	.action(function (packageName) {
 		const fs = require("fs");
+		const fspromises = fs.promises;
+		const path = require("path");
 		const packageJson = require("package-json");
 
 		if (!fs.existsSync(Helper.getConfigPath())) {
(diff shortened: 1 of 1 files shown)


brunnre8:
rebased your fix and let the (proper) formatter run this time around... sorry for that.

MaxLeiter (approved):

PR #4256, opened 2021-06-12
Provide the option to hide NSFW previews

storbeck:
Many times people post NSFW images in IRC and they are usually marked as such otherwise they could be reprimanded. We should provide an option to disable showing NSFW previews while still allowing SFW previews

### DIFF
--- a/client/components/LinkPreview.vue
+++ b/client/components/LinkPreview.vue
@@ -132,6 +132,7 @@
 <script>
 import eventbus from "../js/eventbus";
 import friendlysize from "../js/helpers/friendlysize";
+import messageMatchFilters from "../js/helpers/messageMatchFilters";
 
 export default {
 	name: "LinkPreview",
@@ -233,22 +234,16 @@ export default {
 					this.$refs.content.offsetWidth >= this.$refs.container.offsetWidth;
 			});
 		},
+		shouldHidePreview() {
+			return messageMatchFilters(
+				this.$parent.message.text,
+				this.$store.state.settings.hidePreviewFilters
+			);
+		},
 		updateShownState() {
-			// Ensure we don't have empty strings in the list of preview filters
-			const filterTokens = this.$store.state.settings.hidePreviewFilters
-				.split(",")
-				.map((filter) => filter.trim())
-				.filter((filter) => filter.length > 0);
-
-			const messageText = this.$parent.message.text.toLowerCase();
-
-			const caughtFilter = filterTokens.some((filter) => {
-				return messageText.includes(filter.toLowerCase());
-			});
-
 			// User has manually toggled the preview, or the message has been filtered
 			// by the preview filter settings, do not apply default
-			if (this.link.shown !== null || caughtFilter) {
+			if (this.link.shown !== null || this.shouldHidePreview()) {
 				return;
 			}
 
--- a/client/components/LinkPreview.vue
+++ b/client/components/LinkPreview.vue
@@ -234,8 +234,13 @@ export default {
 			});
 		},
 		updateShownState() {
-			// User has manually toggled the preview, do not apply default
-			if (this.link.shown !== null) {
+			// User has manually toggled the preview, or the image is marked as nsfw
+			// and the setting has been enabled do not apply default
+			const hideNsfw =
+				this.$parent.message.text.toLowerCase().includes("nsfw") &&
+				this.$store.state.settings.hideNsfw;
+
+			if (this.link.shown !== null || hideNsfw) {
 				return;
 			}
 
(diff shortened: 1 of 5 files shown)


termermc:
Perhaps you can be more explicit in the option name, like `Hide previews for links containing "NSFW"`?

brunnre8:
This is essentially a duplicate of https://github.com/thelounge/thelounge/pull/4191 this time with a hardcoded vocabulary which I find suboptimal.

Not everyone speaks English, nsfw might just be one tag applied but it could also be lewed, pr0n, porn... take your pick.
Depends on the channel guidelines.

I don't like this hardcoded list as for the rest of the arguments please read the comments on that PR of Nachtalb

storbeck:
@brunnre8 thanks for the feedback! I've changed it to just be a textbox similar to highlights to be more customizable and less specific to my use-case

client/components/LinkPreview.vue
```diff
@@ -234,8 +234,21 @@ export default {
 			});
 		},
 		updateShownState() {
-			// User has manually toggled the preview, do not apply default
-			if (this.link.shown !== null) {
+			// Ensure we don't have empty strings in the list of preview filters
+			const filterTokens = this.$store.state.settings.hidePreviewFilters
```
MaxLeiter, line 238: can you make this its own function and write tests for edge cases?

MaxLeiter (changes requested):

client/components/LinkPreview.vue
storbeck: Absolutely I can do that

brunnre8:
If you want to add this, can we at least make the tag a freetext field with a plain text matcher as Nachtalb implemented it?

Hardcoding NSFW is really worse than having it configurable and you're introducing a config anyhow

brunnre8:
I think though that #4191 is better as it also includes the ability to hide previews of a nick / chan with a right click menu...

After all that's what was discussed on IRC no?

storbeck:
#4191 has been changed since it's original pull request and it no longer offers this functionality. It seems to only hide based on nick or channel and is only on or off. These two don't appear to be competitors and could co-exist. @Nachtalb mentioned that they do still wish to provide a keyword based filter which this solves. Do you think the pull requests should be merged or are you against the keyword based filter?

brunnre8:
I can see why one would like to do both, channel/user and keyword.

Nachtalb only changed his PR after getting push back in the IRC channel, I still have his original code locally iirc

fnutt:
I'm against adding lots of advanced configuration options. I bet this would not have been merged less than a year ago...

Make it simple! Right click and hide, as #4191 does. Regex and other stuff is advanced stuff that could be put into a plugin in the feature.

Nachtalb:
TBH I still don't get why a keyword list filter would be too "complex" or "advanced" for the upstream and the need to move it to an own plugin. It doesn't have to be anything fancy like the regex match proposed at the start or anything. Just a simple list of words that cause a message to hide its preview. It's miles better than any hardcoded thing (that btw was once proposed by @MaxLeiter (https://github.com/thelounge/thelounge/pull/4191#issuecomment-819033893)), is less complex than any regex handling and enables every user to easily hide whatever they want. 

I mean we have a field for custom CSS. This is on its way more "advanced" than a simple blacklist, as the user has to know CSS syntax and so on. So I really don't see the ground of that argument.


Click'n'hide only allows a more lean way of `/ignore` and nothing more. With a simple list, you could just do if the message contains any blacklisted word then do stuff. Like so you can blacklist any kind of website you don't want to get content from or any kind of NSFW//NSFL content if the channels rule is to mark the content. 

If this is "too complex" for TL then IMHO you could just as well remove: Custom Highlights, Highlight exceptions, custom CSS, nick autocomplete postfix and maybe even more. 
And don't forget that the demographic of IRC is not your average Joe, but someone who is already techy to some degree. A person that grew up with Whatsapp and isn't techy will not ever use IRC as it is not a rich chat platform.

But all in all, we may disagree on all that. And if you guys really are that much against then so be it. ¯\_(ツ)_/¯

brunnre8:
I agree with Nachtalb here... A keyword list (without the regex shenanigans, as discussed) is preferable over a hardcoded list of any kind in my view.

storbeck:
> I agree with Nachtalb here... A keyword list (without the regex shenanigans, as discussed) is preferable over a hardcoded list of any kind in my view.

That is what this PR aims to accomplish

test/client/js/helpers/messageMatchFilters.js
```diff
@@ -0,0 +1,41 @@
...
+		];
+
+		for (const teststring of teststrings) {
+			expect(messageMatchFilters(teststring, hidePreviewFilters)).to.be.false;
+		}
+	});
+
+	it("should match a filter", function () {
+		const teststrings = [
+			"this http://www.bar.com/link.png should be filtered",
```
brunnre8, line 28: I'm wondering if that's really desired.
A word filter should match on words, not arbitrary substrings.

Meaning "this bar should be filtered" while "this foobar should not"

PR #4257, opened 2021-06-12
Replace sharp and libvips with gm and ImageMagick

bl1nk:
:wave: This PR replaces the `sharp` dependency with `gm` which makes use of ImageMagick to process images.

Currently `sharp` is used to strip metadata from images and rotate them accordingly. Due to absence of precompiled binaries of `libvips` for the armv7 architecture, which are needed by `sharp`, our only option to provide Docker container images for all platforms is to compile libvips from source.

The alternative is to replace `sharp` with something else. I tried the first thing that came to my mind when it comes to image processing which is ImageMagick. Packages are available in [Debian](https://packages.debian.org/stretch/imagemagick) and [Alpine Linux](https://pkgs.alpinelinux.org/packages?name=imagemagick&branch=v3.11) which would allow us to include all needed dependencies in [thelounge-docker](https://github.com/thelounge/thelounge-docker) and [thelounge-deb](https://github.com/thelounge/thelounge-deb).

`image/heif` type images will be skipped. Debian stretch does not include libheif which would be needed to work with HEIF files. For better UX the image will be uploaded as is instead of showing an error to the user.


termermc:
I'd actually really like the addition of ImageMagick, since in the future it could be used to thumbnails, or something else. It's a good tool to have.

fnutt:
I am against using ImageMagic since heif images would have to be skipped. This mean all Apple iPhone users will not have support for stripping exif data, which is rather bad if you ask me (if I have understood it correctly).

bl1nk:
> This mean all Apple iPhone users will not have support for stripping exif data, which is rather bad if you ask me (if I have understood it correctly).

I thought the same and tested it using my iPhone. iOS will convert HEIF images to jpeg before uploading and sharing, even when using the "Actual Size" option. The [comment in uploader.js](https://github.com/thelounge/thelounge/pull/4257/files#diff-d70ef2faf4832e851ed4e2fb5b66c16328d6a79d1e0e751d41f26241beb97768R253-R255) also clarifies this:

``` javascript
// On some OS ImageMagick is too old to deal with HEIF. Instead of
// erroring and not uploading any picture HEIF files will be
// uploaded as is.
//
// When sharing HEIF pictures from iOS devices the OS takes care of
// converting the picture to JPEG beforehand which then will be
// successfully rotated and stripped from metadata.
const isImage = contentType.startsWith("image/") && !contentType.endsWith("heif");
```

brunnre8:
instead of adapting to debian ancient, can we ask image magik if heif is supported and use it where available?

If not it might be better to try anyhow and only fallback to not doing it if it errors out.

But we should not arbitrary restrict the functionality if the underlying library actually supports it I think.

MaxLeiter:
Thanks @bl1nk! I don't think I'm a fan of adding ImageMagick as a new (unpackaged) dependency for users running from npm/source. xPaw suggested (and I agree) that our best option is trying to get sharp/libvips (https://github.com/lovell/sharp-libvips) to provide more prebuilt binaries. Do you have any thoughts on this @bl1nk?

bl1nk:
Sounds good to me. I agree that getting more prebuilt binaries of libvips is the better option.

Issue #4175, opened 2021-03-04
Keybindings for highlights


ronilaukkarinen:
### Feature Description

One of our users brought out two brilliant ideas:

1. Add hotkey for viewing the highlights popup "@-icon", for example alt/opt + M
2. Make it possible to jump to highlights for example by pressing and holding ctrl and pressing up/down keys

As and front-end developer I might look into it in some point (although my strongest suit is CSS/style side and not necessarily super technical JS side) but decided to throw a feature req first in case if you had some ideas.

PR #4258, opened 2021-06-12
Toggle recent mentions popup with alt+m

bl1nk:
This adds a keybind to toggle the recent mentions popup using `alt`+`m`(or `opt`+`m` on macOS). 

The top bar is only shown if the user is connected to at least one network. Only then it is possible to open the recent mentions popup so the keybind only toggles under the same conditions.

Relates to #4175 as it implements the first idea mentioned in that issue.

### DIFF
--- a/client/components/App.vue
+++ b/client/components/App.vue
@@ -101,7 +101,9 @@ export default {
 			return false;
 		},
 		toggleMentions() {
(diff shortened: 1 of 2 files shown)


MaxLeiter:
Great, thanks @bl1nk!

PR #4260, opened 2021-06-19
Display server-originated notices to channels in the channel window

### LINKED ISSUE #4180: Server-originated notices to channels should be displayed in the channel
### Feature Description

Currently if a server sends a notice to a channel it gets displayed in the server window. This is problematic as it can result in context being lost. e.g. if I was to `/REMOVE` Tr0ll123 from a channel the following notice is sent to channel members:

```
Notice: -irc.example.com- Sadie removed Tr0ll123 from the channel
```

This being shown in the server window is problematic as it is loses the context of the channel that it should be shown in.

Ideally channel notices should be displayed in the channel window to avoid losing this context.

BradleyShaw:
Closes #4180 

This should preserve the current behaviour of displaying all server-originated messages to the status window with the exception of those targeted at a channel which will now be displayed in the channel window

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -51,8 +51,13 @@ module.exports = function (irc, network) {
 				return Helper.compareHostmask(entry, data);
 			});
 
-		// Server messages go to server window, no questions asked
-		if (data.from_server) {
+		// Server messages that aren't targeted at a channel go to the server window
+		if (
+			data.from_server &&
+			(!data.target ||
+				!network.getChannel(data.target) ||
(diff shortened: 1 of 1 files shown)


fnutt:
Just want to make sure - notices targeted a user from e.g. UNO-bot must be displayed in the channel where UNO-bot is. If not it is very difficult to play the game.

Does this pr change this behaviour?

BradleyShaw:
It will only affect how messages from a server (e.g `:irc.example.org NOTICE target :message`) are handled and should only change the behaviour if the target is a channel

MaxLeiter (approved):

MaxLeiter:
Thanks!

Issue #2395, opened 2018-04-27
Enforce unique network names, use them for URLs & paths


dgw:
I started thinking about this in the comments of #2366, and thought I'd try to pull out the potential issues with doing this in a separate discussion to avoid too much cross-talk in that thread. Ultimately, this represents an alternate solution to #1392, instead of using (parts of) UUIDs for log folder names.

Since The Lounge's maintainers agree (I'm taking [@McInkay's word](https://github.com/thelounge/thelounge/pull/2366#issuecomment-384574806) on that) unique network names are, in principle, a good goal, I would like to get the thought process going. The 3.0 update will be the best opportunity to introduce a major change like this for probably a long time, after all. It's also probably best to change the logging path generation as few times as possible, so it's not ideal to go with the (currently fluctuating) solution involving (parts of) UUIDs in #2366 for 3.0 and then change it again in, say, 4.0.

### Handling existing duplicates

This is probably the main pitfall that will face users hypothetically upgrading to The Lounge 3.0 if it begins enforcing unique network names, as some users might already have multiple networks with the same name. The obvious answer is "add a migration path", but it's not so simple to come up with how said migration should work.

One option is to perform an upgrade on each user's JSON file at first start of the new TL version, and iterate through the saved networks, renaming duplicates automatically. It's brute-force-y, and probably won't come up with particularly helpful names ("Freenode", "Freenode 2", "Freenode 3", etc., anyone?), but it's simple. Neither the server administrator performing the upgrade nor a user with duplicate network names needs to do anything to make sure TL 3.0 can start. The naïvely renamed networks can be fixed with network editing (which The Lounge 3.0 will support—hooray!).

Another option is to do some renaming interactively in an upgrade script. The server administrator would provide human intelligence instead of a "dumb" auto-rename algorithm, but it would cost admin time. It also means actively presenting at least some users' network names to the server admin, which could make users uncomfortable. (Even though the server admin can already read everything, this is less theoretical.) The main reason I don't like it, though, is that it makes the server admin rename networks previously unseen, and never used personally.

The option I like the most is delayed renaming, whereby the code to check whether duplicate network names exist is enabled but the whole system tries to put up with duplicate names for some number of releases. It could come with a warning to affected users when they log in that some features may not work correctly unless they fix the duplicated network names. (I have no UX or UI design in mind at this time, however. This is still idea phase.) This option's downside is that it might delay resolution of e.g. #1392 for another version or three.

### Stop using server addresses in URIs and paths

[...]

Issue #2342, opened 2018-04-10
Use 'speaking' urls for networks and channels


MiniDigger:
use
https://demo.thelounge.chat/Freenode/#thelounge (for channel)
https://demo.thelounge.chat/Freenode (for network)
https://demo.thelounge.chat/Freenode/MiniDigger (for pm)
instead of
https://demo.thelounge.chat/#chan-42

McInkay:
Few questions about this:

1. When we have multiple servers named the same thing. I guess we would just go to the first in the list? Or find the first that has the channel/query in it 

2. When we have none open, do we automatically join/query? If not, do we open the history for that channel/query if we have any logs for it? That might be nice.

creesch:
> When we have multiple servers named the same thing. I guess we would just go to the first in the list? Or find the first that has the channel/query in it

Irccloud just adds a number to channels with duplicate names, that could work. 

So if you have two freenodes you'd get something like this. 

- https://demo.thelounge.chat/Freenode 
- https://demo.thelounge.chat/Freenode:1

christer88:
https://github.com/thelounge/thelounge/issues/1402

PR #4261, opened 2021-06-21
Remove channel IDs from URLs, use network host and channel name instead

### LINKED ISSUE #4179: Static URLs in Private Mode
### Feature Description

Hi! We are looking at using TheLounge to facilitate communication during the UseR! 2021 conference.

We've set up a private instance to experiment with, since the conference will have to be private.

We hope to be able to provide links on the conference website to specific channels, for example, there might be a channel for questions about each talk. Currently we are finding that the URLs for the channels are not static.

We have read about these issues in [888](https://github.com/thelounge/thelounge/issues/888)
and [674](https://github.com/thelounge/thelounge/issues/674).

Thank you!

MaxLeiter:
Links take the following forms:

https://demo.thelounge.chat/irc.libera.chat/thelounge => irc.libera.chat#thelounge
https://demo.thelounge.chat/irc.libera.chat/%23thelounge => irc.libera.chat#thelounge

I couldn't figure out a non-hacky way to handle anchor tags in the URL (to support `/#channel` style URLs), but imo that can be tackled in a future PR, if at all.

If the network exists, the Join Channel UI is populated. If the network doesn't, the connect UI is opened and the channel is pre-filled into the channels input box. 

Closes #4179 #2342 #2395


client/js/router.js
```diff
@@ -94,9 +94,57 @@ router.beforeEach((to, from, next) => {
...
+
+		if (existingNetwork) {
+			// Join Channel UI
+
+			const activeChannel = store.state.activeChannel;
+
+			// if the active channel is in the network, send the user back to that channel, else to the lobby
+			if (activeChannel && activeChannel.network.uuid === existingNetwork.uuid) {
+				next({
+					path: `/${to.params.networkHost}/${encodeURIComponent(
```
xPaw, line 128: This seems like the only place you called `encodeURIComponent`. If I had to guess there will be problems with unconventional channel names (like containing `/` and other fancy characters that may break the url).

xPaw:
How will this work if you have multiple networks with same host? That's very common with bouncers.

fnutt:
> How will this work if you have multiple networks with same host? That's very common with bouncers.

Suggestion: If more than one connection to the same network host, then add a prefix number in front of the host. E.g. http://<domain>/#/irc.libera.chat/<channel> and http://<domain>:9000/#/<prefix number>/irc.libera.chat/<channel>

It will solve the issue for advanced users with bouncers. For most other users, especially public instances of The Lounge, the url will be as intended.

dgw:
Aside from it looking like some bits of those examples got lost (did you surround things in `&lt;`/`&gt;`?), I say…

Why not just use the (slugified) network name?

fnutt:
> Aside from it looking like some bits of those examples got lost (did you surround things in `<`/`>`?), I say…
> 
> Why not just use the (slugified) network name?

Fixed :)

brettgilio:
What is the status on this being merged?

PR #4267, opened 2021-06-29
properly track user modes for context menu

brunnre8:
The RPL_ISUPPORT response contains a PREFIX element, which not only tracks the
prefix chars ("@", "+" etc) but also their corresponding mode chars (+O, +v)
This commit changes the context menu to not rely on a hardcoded list but rather
user the one given in the prefix response by the server.

@fnutt ping, please test this if you can and find the time

### DIFF
--- a/client/js/helpers/contextMenu.js
+++ b/client/js/helpers/contextMenu.js
@@ -275,7 +275,7 @@ export function generateUserContextMenu($root, channel, network, user) {
 		},
 	};
 
-	const networkModeSymbols = network.serverOptions.PREFIX.map((m) => m.symbol);
+	const networkModeSymbols = network.serverOptions.PREFIX.symbols;
 
 	/**
 	 * Determine whether the prefix of mode p1 has access to perform actions on p2.
@@ -296,7 +296,7 @@ export function generateUserContextMenu($root, channel, network, user) {
 			: networkModeSymbols.indexOf(p1) < networkModeSymbols.indexOf(p2);
 	}
 
-	network.serverOptions.PREFIX.forEach((mode) => {
+	network.serverOptions.PREFIX.prefix.forEach((mode) => {
 		if (!compare(currentChannelUser.modes[0], mode.symbol)) {
 			// Our highest mode is below the current mode.  Bail.
 			return;
--- a/client/js/helpers/contextMenu.js
+++ b/client/js/helpers/contextMenu.js
@@ -252,22 +252,30 @@ export function generateUserContextMenu($root, channel, network, user) {
 		return items;
 	}
 
-	// Names of the modes we are able to change
-	const modes = {
-		"~": ["owner", "q"],
-		"&": ["admin", "a"],
-		"@": ["operator", "o"],
-		"%": ["half-op", "h"],
-		"+": ["voice", "v"],
+	// Names of the standard modes we are able to change
+	const modeCharToName = {
+		"~": "owner",
+		"&": "admin",
+		"@": "operator",
+		"%": "half-op",
+		"+": "voice",
 	};
 
 	// Labels for the mode changes.  For example .rev(['admin', 'a']) => 'Revoke admin (-a)'
 	const modeTextTemplate = {
-		revoke: (m) => `Revoke ${m[0]} (-${m[1]})`,
-		give: (m) => `Give ${m[0]} (+${m[1]})`,
+		revoke(m) {
+			const name = modeCharToName[m.symbol];
+			const res = name ? `Revoke ${name} (-${m.mode})` : `Mode -${m.mode}`;
+			return res;
+		},
+		give(m) {
+			const name = modeCharToName[m.symbol];
+			const res = name ? `Give ${name} (+${m.mode})` : `Mode +${m.mode}`;
+			return res;
+		},
 	};
 
-	const networkModes = network.serverOptions.PREFIX;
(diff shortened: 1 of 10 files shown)


fnutt:
Works very well! No errors as before.

Thanks for the quick fix. 👍

src/plugins/irc-events/connection.js
```diff
@@ -202,15 +202,12 @@ module.exports = function (irc, network) {
 		data.options.PREFIX.forEach((mode) => {
 			network.prefixLookup[mode.mode] = mode.symbol;
```
xPaw, line 203: This can probably be removed and use `serverOptions.PREFIX` instead.

client/js/helpers/contextMenu.js
```diff
@@ -252,22 +252,28 @@ export function generateUserContextMenu($root, channel, network, user) {
...
+		"%": "half-op",
+		"+": "voice",
 	};
 
 	// Labels for the mode changes.  For example .rev(['admin', 'a']) => 'Revoke admin (-a)'
 	const modeTextTemplate = {
-		revoke: (m) => `Revoke ${m[0]} (-${m[1]})`,
-		give: (m) => `Give ${m[0]} (+${m[1]})`,
+		revoke(m) {
+			const name = modeCharToName[m.symbol] ? modeCharToName[m.symbol] : "unkown mode";
```
xPaw, line 267:
```suggestion
			const name = modeCharToName[m.symbol] || "unknown mode";
```

src/plugins/irc-events/connection.js
brunnre8: tried that, the problem is that it makes the compare function awkward. All callsites only have the prefix string (@, + etc) and then you'd need to shoehorn those strings into the expected object which looked very bad when I tried it

brunnre8:
Thanks for the review xPaw, appreciate it.
Bear with me with the style choices and please do correct me if something is not idiomatic. I'm at home in python and go so this will bleed in.

src/plugins/irc-events/connection.js
brunnre8: wait, wrong part of the code.

Let me check where we even use the prefixLookup

client/js/helpers/parse.js
```diff
@@ -79,7 +79,10 @@ function parse(createElement, text, message = undefined, network = undefined) {
 	// arrays of objects containing start and end markers, as well as metadata
 	// depending on what was found (channel or link).
 	const channelPrefixes = network ? network.serverOptions.CHANTYPES : ["#", "&"];
-	const userModes = network ? network.serverOptions.PREFIX : ["!", "@", "%", "+"];
+	const userModes =
+		network && network.serverOptions && network.serverOptions.PREFIX
```
MaxLeiter, line 83: network?.serverOptions?.PREFIX should work

brunnre8: do we have that compatibility level? That syntax is pretty new.
Will do that then

MaxLeiter: yeah we should support es2020, which supports it

brunnre8:
This should be done I think, I've refactored the prefixLookup map to get rid of the separate map that was passed around, should make things slightly easier server side.
On the client, not so much I'm afraid but I don't see a way to make this cleaner.

client/js/helpers/contextMenu.js
```diff
@@ -252,22 +252,28 @@ export function generateUserContextMenu($root, channel, network, user) {
...
-		revoke: (m) => `Revoke ${m[0]} (-${m[1]})`,
-		give: (m) => `Give ${m[0]} (+${m[1]})`,
+		revoke(m) {
+			const name = modeCharToName[m.symbol] || "unknown mode";
+			return `Revoke ${name} (-${m.mode})`;
+		},
+		give(m) {
+			const name = modeCharToName[m.symbol] || "unknown mode";
+			return `Give ${name} (+${m.mode})`;
+		},
```
JeDaYoshi, line 273: I'd personally prefer if it just fell back to `Revoke -${m.mode}`.
Saying `Revoke/Give unknown mode` makes the context menu unnecessarily longer and doesn't looks as well.

brunnre8: "give +$mode" is a redundant phrasing. The plus or minus already tells you what it is.

So either we should get rid of the prefix (give/revoke) or the symbols in my opinion.

I'm not hung up on the unknown mode thing, was just the first thing that came to mind

dgw: The plus or minus bit with the actual mode character is basically letting the protocol leak through into the GUI. "Give" and "Revoke" (or "Remove") are the human-friendly expressions of those actions. It belongs in parentheses (current), or should be omitted if there mode's friendly name is known.

Strong agree on skipping the "unknown mode" name fallback, though. If the modechar is all that's available, having that fallback doesn't tell the user anything more than not having it.

brunnre8: you still need some human parsable thing and that will still be constant.

How about "mode +Q" or what have you and correspondingly "mode -Q" ?

That (while not being symmetric to the human friendly names) at least makes it obvious what the options does.

brunnre8:
Got rid of the "Unknown mode" thing as per user feedback.
PTAL

MaxLeiter (changes requested):

fnutt:
As requested by MaxLeiter, I have tested this PR again after the recent changes. I can confirm it works wonderfully!

PR #4274, opened 2021-07-03
Add /umode support

jeda:
This PR adds the `/umode` command to set user modes without needing to be focused in the server tab.


MaxLeiter:
Thanks!

PR #4275, opened 2021-07-03
Optimise modes based on ISUPPORT

JeDaYoshi:
This will see the maximum allowed of modes that are allowed at once as sent in RPL_ISUPPORT and will send multiple batches while using /op, /voice, etc.

This also fixes a minor issue where it would try sending an empty voice if it had an extra space on arguments (such as using `/voice  `)

### DIFF
--- a/src/plugins/inputs/mode.js
+++ b/src/plugins/inputs/mode.js
@@ -42,7 +42,7 @@ exports.input = function ({irc, nick}, chan, cmd, args) {
 			devoice: "-v",
 		}[cmd];
 
-		const limit = parseInt(irc.network.supports("MODES")) || 1;
+		const limit = parseInt(irc.network.supports("MODES")) || target.length;
 
 		for (let i = 0; i < target.length; i += limit) {
 			const targets = target.slice(i, i + limit);
--- a/src/plugins/inputs/mode.js
+++ b/src/plugins/inputs/mode.js
@@ -42,7 +42,7 @@ exports.input = function ({irc, nick}, chan, cmd, args) {
 			devoice: "-v",
 		}[cmd];
 
-		const limit = parseInt(irc.network.supports("modes")) || 1;
+		const limit = parseInt(irc.network.supports("MODES")) || 1;
 
 		for (let i = 0; i < target.length; i += limit) {
 			const targets = target.slice(i, i + limit);
--- a/src/plugins/inputs/mode.js
+++ b/src/plugins/inputs/mode.js
@@ -19,7 +19,9 @@ exports.input = function ({irc, nick}, chan, cmd, args) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


JeDaYoshi:
Oh, oops. That seems to have broken tests since they don't account for `.supports`... I can take a look at that.

test/commands/mode.js
```diff
@@ -20,6 +20,13 @@ describe("Commands", function () {
 			lastCommand: null,
 			nick: "xPaw",
 			irc: {
```
MaxLeiter, line 22: Can you add a test for servers without ISUPPORT?

MaxLeiter (changes requested):

src/plugins/inputs/mode.js
```diff
@@ -40,9 +42,13 @@ exports.input = function ({irc, nick}, chan, cmd, args) {
 			devoice: "-v",
 		}[cmd];
 
-		args.forEach(function (target) {
-			irc.raw("MODE", chan.name, mode, target);
-		});
+		const limit = parseInt(irc.network.supports("modes")) || 1;
```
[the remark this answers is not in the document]
JeDaYoshi: Already fixed on the second commit.

test/commands/mode.js
JeDaYoshi: Sure thing.

test/commands/mode.js
```diff
@@ -81,10 +101,29 @@ describe("Commands", function () {
...
+			]);
+			expect(testableNetwork.lastCommand).to.equal("MODE #thelounge -v thelounge");
+		});
+
+		it("should fallback to default limit of 1 mode for shorthand commands", function () {
+			ModeCommand.input(testableNetworkNoSupports, channel, "voice", ["xPaw"]);
+			expect(testableNetworkNoSupports.lastCommand).to.equal("MODE #thelounge +v xPaw");
 
-			// Multiple arguments are supported, sent as separate commands
-			ModeCommand.input(testableNetwork, channel, "devoice", ["xPaw", "Max-P"]);
```
MaxLeiter, line 86: based on https://defs.ircdocs.horse/defs/isupport.html there should be no limit if `MODES` is undefined, not 1, but that could be inaccurate. Did you choose 1 for a reason?

MaxLeiter (changes requested):

test/commands/mode.js
JeDaYoshi: I chose 1 mostly for safety. Since I don't know whether the IRCd will accept all my modes or not (and since there's no mechanism for checking up if all were indeed set), I went that way. I can change that to no limit if desired, though.

MaxLeiter (approved):

MaxLeiter:
Thanks!

PR #4277, opened 2021-07-03
Fix nick-less messages from servers

JeDaYoshi:
There are cases where it's slightly confusing from which server things like, for example, InspIRCd server notices, are sent from, since they are sent as messages, and there's not a handler for setting the nick in these cases (resulting in an empty sender).

This PR fixes it.

### DIFF
--- a/src/plugins/irc-events/message.js
+++ b/src/plugins/irc-events/message.js
@@ -11,12 +11,6 @@ module.exports = function (irc, network) {
 	const client = this;
 
 	irc.on("notice", function (data) {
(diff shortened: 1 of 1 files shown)


BradleyShaw:
I think this code might be better off somewhere like `handleMessage()` since a server *could* send any of the message types handled here

JeDaYoshi:
Hmm, I can do that either. I'll take a look in a minute.

JeDaYoshi:
Should be good now.

MaxLeiter (approved):

PR #4278, opened 2021-07-03
Disable /search if not possible and move clientCommands to client

### LINKED ISSUE #4273: Disable `/search query` if search is not possible.
Search only works if sqlite `messageStorage` is enabled. Right now, when this is not enabled, you don't get the magnifying glass icon but `/search query` is still available. Searching for something opens up the search window but it gets stuck in the "searching.." state

Closing that window takes you back to the channel and reveals an error VUE ERROR: CANNOT READ PROPERTY 'SLICE' OF UNDEFINED:

I guess the sane solution would be to disable the `/search` command if it's not possible (either because logging to sqlite is disabled globally or for the user)

JeDaYoshi:
Fixes #4273

### DIFF
--- a/client/js/autocompletion.js
+++ b/client/js/autocompletion.js
@@ -6,7 +6,6 @@ import Mousetrap from "mousetrap";
 import {Textcomplete, Textarea} from "textcomplete";
 import fuzzy from "fuzzy";
 
-import commands from "./commands/index";
 import emojiMap from "./helpers/simplemap.json";
 import store from "./store";
 
@@ -313,14 +312,17 @@ function completeNicks(word, isFuzzy) {
 }
 
 function getCommands() {
-	const clientCommands = Object.keys(commands).map((cmd) => `/${cmd}`);
-	const cmds = [...new Set(Array.from(constants.commands).concat(clientCommands))];
+	const cmds = constants.commands;
 
-	if (!store.state.settings.searchEnabled) {
-		cmds.pop("/search");
+	if (store.state.settings.searchEnabled === false) {
+		const search = cmds.indexOf("/search");
+
+		if (search !== -1) {
+			cmds.splice(search, 1);
+		}
 	}
 
-	return cmds.sort();
+	return cmds;
 }
 
 function completeCommands(word) {
--- a/client/js/autocompletion.js
+++ b/client/js/autocompletion.js
@@ -6,6 +6,7 @@ import Mousetrap from "mousetrap";
 import {Textcomplete, Textarea} from "textcomplete";
 import fuzzy from "fuzzy";
 
+import commands from "./commands/index";
 import emojiMap from "./helpers/simplemap.json";
 import store from "./store";
 
@@ -311,8 +312,19 @@ function completeNicks(word, isFuzzy) {
 	return users.filter((w) => !w.toLowerCase().indexOf(word));
 }
 
+function getCommands() {
+	const clientCommands = Object.keys(commands).map((cmd) => `/${cmd}`);
+	const cmds = [...new Set(Array.from(constants.commands).concat(clientCommands))];
+
+	if (!store.state.settings.searchEnabled) {
+		cmds.pop("/search");
+	}
+
+	return cmds.sort();
+}
+
 function completeCommands(word) {
-	const words = constants.commands.slice();
+	const words = getCommands().slice();
 
 	return fuzzyGrep(word, words);
 }
--- a/client/js/autocompletion.js
+++ b/client/js/autocompletion.js
@@ -312,7 +312,7 @@ function completeNicks(word, isFuzzy) {
 }
 
 function getCommands() {
-	const cmds = constants.commands;
+	const cmds = constants.commands.slice();
 
 	if (store.state.settings.searchEnabled === false) {
 		const search = cmds.indexOf("/search");
@@ -326,7 +326,7 @@ function getCommands() {
 }
 
 function completeCommands(word) {
-	const words = getCommands().slice();
+	const words = getCommands();
 
 	return fuzzyGrep(word, words);
 }
--- a/client/js/commands/search.js
+++ b/client/js/commands/search.js
@@ -4,15 +4,19 @@ import store from "../store";
 import {router} from "../router";
 
 function input(args) {
-	router.push({
-		name: "SearchResults",
-		params: {
-			id: store.state.activeChannel.channel.id,
-		},
-		query: {
-			q: args.join(" "),
-		},
-	});
+	if (!store.state.settings.searchEnabled) {
+		store.commit("currentUserVisibleError", "Search is currently not enabled.");
+	} else {
+		router.push({
+			name: "SearchResults",
+			params: {
+				id: store.state.activeChannel.channel.id,
+			},
+			query: {
+				q: args.join(" "),
+			},
+		});
+	}
 
 	return true;
 }
--- a/client/js/commands/search.js
+++ b/client/js/commands/search.js
@@ -2,22 +2,22 @@
 
 import store from "../store";
 import {router} from "../router";
+const Msg = require("../../../src/models/msg");
 
 function input(args) {
+	const {channel} = store.state.activeChannel;
+
 	if (!store.state.settings.searchEnabled) {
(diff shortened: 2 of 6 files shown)


JeDaYoshi:
*blinks* That test failure seems unrelated to my changes...

MaxLeiter:
I think it;'d be best if we also want to remove /search from the autocomplete: https://github.com/thelounge/thelounge/blob/fc6c916e7cb5582fb6e7e26f1c468d7ac9417be0/src/plugins/inputs/index.js#L50

JeDaYoshi:
That will work.

test/plugins/inputs/indexTest.js
```diff
@@ -4,16 +4,34 @@ const expect = require("chai").expect;
...
 
 		it("should only return strings with no whitespaces and starting with /", function () {
-			inputs.getCommands().forEach((command) => {
+			inputs.getCommands(client).forEach((command) => {
 				expect(command).to.be.a("string").that.does.not.match(/\s/);
 				expect(command[0]).to.equal("/");
 			});
 		});
+
+		it("should include /search without a message provider", function () {
```
MaxLeiter, line 27: should not* include

MaxLeiter (changes requested):

JeDaYoshi:
What worries me is that the current approach of running `getCommands()` every auto-completion for commands might hurt somewhat in performance. My other idea was moving `getCommands` to when the socket emits `commands`, but at that point, I don't have access to whether search is enabled from the store.

client/js/autocompletion.js
```diff
@@ -311,8 +312,19 @@ function completeNicks(word, isFuzzy) {
 	return users.filter((w) => !w.toLowerCase().indexOf(word));
 }
 
+function getCommands() {
+	const clientCommands = Object.keys(commands).map((cmd) => `/${cmd}`);
+	const cmds = [...new Set(Array.from(constants.commands).concat(clientCommands))];
```
MaxLeiter, line 317: move these outside of the function to make it easier for v8 to optimize

ref https://stackoverflow.com/a/58436106

MaxLeiter (changes requested):

JeDaYoshi:
I'm still not 100% on the way it's being done, but I'm unsure of where else to handle the `/search` command removal.

MaxLeiter:
I was going to suggest adding a check in where we emit search:results, but @Nachtalb handled that here https://github.com/thelounge/thelounge/blob/26a38b12abeb61fce10f412872a044c6fcf619da/src/client.js#L548

Request Nachtalbs review if they can get around to it, otherwise will merge in a few days. Thanks for another PR, @JeDaYoshi

MaxLeiter (approved):

client/js/autocompletion.js
```diff
@@ -311,8 +311,22 @@ function completeNicks(word, isFuzzy) {
...
 }
 
+function getCommands() {
+	const cmds = constants.commands;
+
+	if (store.state.settings.searchEnabled === false) {
+		const search = cmds.indexOf("/search");
+
+		if (search !== -1) {
+			cmds.splice(search, 1);
```
Nachtalb, line 321: ⚠️ This will edit `constants.commands`. We do not copy the list here but just assign the reference to it to another variable. So with the splice on line 321, we remove it from `constants.commands`. What you want is a copy: `const cmds = constants.commands.slice();

client/js/autocompletion.js
```diff
@@ -311,8 +311,22 @@ function completeNicks(word, isFuzzy) {
...
+		}
+	}
+
+	return cmds;
+}
+
 function completeCommands(word) {
-	const words = constants.commands.slice();
+	const words = getCommands().slice();
```
Nachtalb, line 329: With the change proposed above, you don't have to do a slice here anymore as the array was already copied.

client/js/commands/search.js
```diff
@@ -4,15 +4,19 @@ import store from "../store";
...
-		name: "SearchResults",
-		params: {
-			id: store.state.activeChannel.channel.id,
-		},
-		query: {
-			q: args.join(" "),
-		},
-	});
+	if (!store.state.settings.searchEnabled) {
+		store.commit("currentUserVisibleError", "Search is currently not enabled.");
```
Nachtalb, line 8: I am not so keen on that one. The `/search` command is made for people who like to do everything with the keyboard. And these errors can only be removed by clicking on them (so the user has to reach the mouse). 

Maybe we could add a timeout functionality to these errors in general (or put a simple `setTimeout` in which we check if it's still the same error and if so remove the error (if we don't add a check we might remove another error which we definitely don't want))

client/js/socket-events/commands.js
```diff
@@ -1,8 +1,12 @@
 const constants = require("../constants");
+import localCommands from "../commands/index";
 import socket from "../socket";
 
+const clientCommands = Object.keys(localCommands).map((cmd) => `/${cmd}`);
```
Nachtalb, line 5: This just reminds me how inconsistent JS is again 😂 the need to call a class method on and passing in an instance of the said class to get the desired effect `Object.keys(obj)` instead of `obj.keys()` like in any other language. Well tbf, it's wrong to call it class methods in the first place as classes don't technically exist in JS anyways. But welp ¯\_(ツ)_/¯

Nachtalb (changes requested):

client/js/autocompletion.js
JeDaYoshi: Ah, I completely missed that I could do the `.slice()` there. I was aware `constants.commands` was being modified and was one of the points that worried me - but yeah... I had to think further about that. Thank you!

client/js/commands/search.js
JeDaYoshi: Applied a setTimeout for 5 seconds. That said, it would be inconsistent with other things (say, HTTP upload errors and so on), which stay forever until you click on them.

JeDaYoshi: To be fair, I think the best action here is that instead of showing an error, adding an information message on the currently open channel. Opinions? (cc @MaxLeiter)

MaxLeiter: I agree currentUserVisibleError shouldn't be used for this. Make it a message so it functions more like responses to other commands

client/js/commands/search.js
```diff
@@ -2,17 +2,28 @@
 
 import store from "../store";
 import {router} from "../router";
+const Msg = require("../../../src/models/msg");
```
Nachtalb, line 5: Should we really include server/backend code in the front end? This will be the first case of importing anything from `/src/` inside the `/client/`
CC/ @MaxLeiter

JeDaYoshi: Yeah... I need input there. I didn't want to make a fake message and just opted in for using the message model, but yes, that'd indeed be the first case.

MaxLeiter: Yeah, we shouldn't import msg. We don't currently have a nice way of generating client-side messages, so we'd need to pass it to the server. Regardless, when thinking about accessibility, a chat message isn't the best way to notify the user of an error (at least, with our current set-up). With that in mind, maybe we should use an alert or modal.

client/js/commands/search.js
```diff
@@ -2,17 +2,28 @@
...
+	const {channel} = store.state.activeChannel;
+
+	if (!store.state.settings.searchEnabled) {
+		const message = new Msg({
+			type: Msg.Type.ERROR,
+			text: "Search is currently not enabled.",
+		});
+		channel.messages.push(message);
+	} else {
+		router.push({
```
MaxLeiter, line 17: There's an existing bug here where the chat input box disappears when we use `/search`. We shouldn't use the router like this to activate the change; maybe we commit to store?

MaxLeiter:
Hey @JeDaYoshi. Any updates on this? Just trying to plan 4.3.

JeDaYoshi:
Right, very sorry for the big delay @MaxLeiter. I'll get back to this on this weekend so there's no stopper :)

JeDaYoshi:
I want to apologise for not taking a look at this PR in a while, I'm going to get to it today and push changes.

PR #4282, opened 2021-07-07
Add the option to mute channels, queries, and networks

MaxLeiter:
Closes https://github.com/thelounge/thelounge/issues/2490

Muted channels have 0.5 opacity applied and notifications aren't sent/no pop is played when a highlight is sent to a muted channel.


fnutt:
What is "query"? I know, but it's not a word I would use for private messages / direct messages.

Would it please be possible to use some other wording that makes more sense to most people?

image

brunnre8:
conversation? Might be better

xnaas:
I commented in IRC, but I'll comment on this PR as well: I'm also pro reduced opacity (vs another option like an icon).

NightMachinary:
Muting networks is great. 👍

brunnre8:
There's at least one bug left.
The hamburger menu icon of the chanlist shows the activity dot when a mention is present in a muted channel.

brunnre8:
Creating a new network mutes it by default apparently, another issue we need to address

MaxLeiter:
@brunnre8 fixed your two bugs

brunnre8 (approved):
Issues seem fixed, thanks!

brunnre8 (changes requested):
New bug, setting the mute doesn't propagate to other clients

src/server.js
```diff
@@ -669,7 +669,7 @@ function initializeClient(socket, client, token, lastMessage, openChannel) {
 				chan.setMuteStatus(setMutedTo);
 			}
 
-			socket.emit("mute:changed", {
+			manager.sockets.emit("mute:changed", {
```
brunnre8, line 672: This seems wrong.
You aren't sending the mute change event to the User are you?
Rather you send it to every connected user (manager.sockets is the global io object)

Or do I misunderstand how our transport layer works?

MaxLeiter: Good catch, I rushed. Proper method is to use the socket IDs and find the relevant one in manager.sockets.

itsjohncs:
Just pulled this into my dogfood branch. Will run it and play around with it for the next few days. Will hopefully have some time for a code review soon as well.

src/plugins/inputs/index.js
```diff
@@ -35,6 +35,7 @@ const userInputs = [
 	"rejoin",
 	"topic",
 	"whois",
+	"mute",
```
itsjohncs, line 38: Should unmute be present too?

src/server.js
```diff
@@ -655,6 +656,28 @@ function initializeClient(socket, client, token, lastMessage, openChannel) {
...
+			// If the user mutes the lobby, we mute the entire network.
+			if (chan.type === Chan.Type.LOBBY) {
+				for (const channel of network.channels) {
+					channel.setMuteStatus(setMutedTo);
+				}
+			} else {
+				chan.setMuteStatus(setMutedTo);
+			}
+
+			for (const attachedClient of Object.keys(client.attachedClients)) {
```
itsjohncs, line 672: I think we'll send a `mute:changed` event back down to the client who just sent one to us. Might be best to exclude them?

itsjohncs (commented):
Did a quick look through. I'll take some time to try and break it before long as well to see if I can shake out any bugs.

itsjohncs (changes requested):
I've been running this for a few days now and I haven't shaken out any bug. Seems alright to me, though I'll wait to approve until we've worked through some of the questions I brought up.

client/js/helpers/contextMenu.js
```diff
@@ -170,6 +170,26 @@ export function generateChannelContextMenu($root, channel, network) {
 		});
 	}
 
+	const humanFriendlyChanTypeMap = {
```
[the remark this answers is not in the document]
brunnre8: special are things like the banlist or the output of /list or whatever the command is, doesn't make sense to mute them I think

src/plugins/inputs/index.js
MaxLeiter: unmute is in `src/plugins/inputs/mute.js`, and this array is referring to the filenames. I can split them into two files if that's preferred, I don't really care.

src/server.js
MaxLeiter: the client (even the one that emits the `mute:change` event) waits for the `mute:changed` acknowledgement from the server, so we want all clients to receive the event. This let me not repeat the 'mute channel' code on the client and just mute channels when the event is received.

src/plugins/inputs/index.js
brunnre8: single file is preferred I think, we have that for others too

client/js/helpers/contextMenu.js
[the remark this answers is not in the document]
itsjohncs: That makes sense. Looks like we'll need to specially exclude them somewhere then:

Screen Shot 2022-02-04 at 2 37 00 PM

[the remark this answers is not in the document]
MaxLeiter: Fixed. Also stopped muting lobbies from muting special channels.

client/components/Windows/Help.vue
```diff
@@ -780,6 +794,18 @@
...
 			</div>
 
+			<div class="help-item">
+				<div class="subject">
+					<code>/unmute [...channel]</code>
+				</div>
+				<div class="description">
+					<p>
+						Un-mutes the given channel(s) or the current channel if no channel is
+						provided. See `/mute` for more information.
```
itsjohncs, line 804: Markdown isn't parsed here. This should be `<code>/mute</code>`.

itsjohncs (approved):
Looks good! Just found one small formatting issue in the docs.

src/plugins/inputs/mute.js
```diff
@@ -0,0 +1,63 @@
...
+			})
+		);
+		return;
+	}
+
+	for (const target of targets) {
+		change_mute_state(client, target, valueToSet);
+	}
+
+	return true;
```
brunnre8, line 62: why return true if you returned nothing in the other code paths?

brunnre8 (approved):
Looks great, small nitpick but other than that, I'd say merge it

the-lazy-fox:
Hi,
Great to see this feature.
Is it managed to not see the badge when muted except if you are mentionned?
Thank you @MaxLeiter

dgw:
@the-lazy-fox You can try it out right now by installing the prerelease package. https://github.com/thelounge/thelounge/releases/tag/v4.3.1-rc.1

brunnre8:
Muting only inhibits notifications due to mentions.
It doesn't quite do what you want I'm afraid.
There are several options of how it could work, my use case is for example that I don't want to get notified, but still want to see channel activity.

That goes exactly counter to your preference, which is that you don't want to see the counter and only see highlights. That's not what muting is about.

the-lazy-fox:
Ok then I will rename my feature request :)

PR #4294, opened 2021-08-03
Fix /away handling no arguments

MaxLeiter:
/away with no arguments removes a users away status. Currently, we're setting the status to an empty string if no arguments are passed.

### DIFF
--- a/src/plugins/inputs/away.js
+++ b/src/plugins/inputs/away.js
@@ -6,7 +6,7 @@ exports.input = function (network, chan, cmd, args) {
 	let reason = "";
 
 	if (cmd === "away") {
-		reason = args.join(" ") || " ";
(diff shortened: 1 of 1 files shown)


brunnre8:
I think that's done on purpose... Setting an empty away status that is, it isn't a bug as such.

It also matches the docs in our help section: https://github.com/thelounge/thelounge/blob/610ae6805aef2a104a12113efde5455761d1d060/client/components/Windows/Help.vue#L374

><p>Mark yourself as away with an optional message.</p>

Resetting the away status is supposed to be done with `/back`

So, I'm fine either way (die hard IRC users probably want the your version). The current behavior though makes more sense for a newbie.

If we go your version though, please fix the docs to mention the new behavior.

PR #4300, opened 2021-08-30
Feature/per user outgoing address binding

sarah501:
User creation will additionally prompt for an outgoing address to bind to. This will bind the user to the specified address which overrides the bind defined in config.js (for the user), if undefined, the bind address specified in config.js will takes effect.

### DIFF
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -187,10 +187,9 @@ Network.prototype.validate = function (client) {
 };
 
 Network.prototype.createIrcFramework = function (client) {
-	var outgoing_addr = Helper.config.bind;
-	if (client.config.bind !== 'undefined') {
-		outgoing_addr = client.config.bind;
-	}
+	const outgoing_addr = client.config.bind !== 'undefined'
+							?  client.config.bind
+							: Helper.config.bind;
 	this.irc = new IrcFramework.Client({
 		version: false, // We handle it ourselves
 		outgoing_addr: outgoing_addr,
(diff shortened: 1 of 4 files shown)


MaxLeiter:
Thanks @sarah501! Can you fix the linting and adjust the documentation to reflect this change? I'm not entirely sure this is how we want to do this, but some documentation would help discussion.

sarah501:
> 
> 
> Thanks @sarah501! Can you fix the linting and adjust the documentation to reflect this change? I'm not entirely sure this is how we want to do this, but some documentation would help discussion.

Which documentation do I need to update?

src/models/network.js
```diff
@@ -189,7 +189,7 @@ Network.prototype.validate = function (client) {
 Network.prototype.createIrcFramework = function (client) {
 	this.irc = new IrcFramework.Client({
 		version: false, // We handle it ourselves
-		outgoing_addr: Helper.config.bind,
+		outgoing_addr: client.config.bind !== "undefined" ? client.config.bind : Helper.config.bind,
```
xPaw, line 192: Don't pass in undefined as string. Make it default to `null` (not string).

The prompt also doesn't really explain anything "Outgoing address?" is just ambiguous.

MaxLeiter:
hey @sarah501! Are you able to pick this up and address xPaws comments?

MaxLeiter:
Closing due to inactivity. @whoever sees this: Feel free to take it up if you're interested. Thanks for your effort, @sarah501!

PR #4308, opened 2021-09-05
Always chmod home dir on startup

awfulcooking:
Helps to prevent and retrospectively address issues like
thelounge/thelounge-deb#72

Obviously some sysadmins may want the directory to be world-readable,
but presently, that isn't safe.

This commit could be omitted or later reverted if sensitive data
protection is ensured through other means - e.g. with patches for
existing files, and an emphasis at review time to catch over-exposure in
the future ✌

Currently world-readable without this patch are: logs/**/*, vapid.json

### DIFF
--- a/src/command-line/start.js
+++ b/src/command-line/start.js
@@ -23,7 +23,6 @@ program
 function initalizeConfig() {
 	if (!fs.existsSync(Helper.getConfigPath())) {
 		fs.mkdirSync(Helper.getHomePath(), {recursive: true});
-		fs.chmodSync(Helper.getHomePath(), "0700");
 		fs.copyFileSync(
 			path.resolve(path.join(__dirname, "..", "..", "defaults", "config.js")),
 			Helper.getConfigPath()
@@ -31,5 +30,6 @@ function initalizeConfig() {
 		log.info(`Configuration file created at ${colors.green(Helper.getConfigPath())}.`);
 	}
 
+	fs.chmodSync(Helper.getHomePath(), "0700");
 	fs.mkdirSync(Helper.getUsersPath(), {recursive: true});
 }


brunnre8:
I don't think that's a good idea, fixing the permissions at creation time is fine, but if the admin changes the permissions we ought to leave them alone.

togetherbeer:
Currently, the permissions look like this:

```sh
drwxrwxr-x   5 thelounge thelounge  4096 Sep  1 18:57 .
drwxr-xr-x 145 root      root      12288 Sep  1 18:36 ..
-rw-rw----   1 thelounge thelounge 16815 Sep  1 18:44 config.js
drwxr-xr-x   3 thelounge thelounge  4096 Sep  1 18:57 logs
drwxr-xr-x   4 thelounge thelounge  4096 Sep  4 15:26 packages
drwxrwx---   2 thelounge thelounge  4096 Sep  5 17:10 users
-rw-r--r--   1 thelounge thelounge   169 Sep  1 18:50 vapid.json
```

And so if we address the permissions on vapid.json and logs (which I would advocate doing retrospectively because of their sensitivity), the only remaining entry that will be readable outside of thelounge:thelounge will be:

```
drwxr-xr-x   4 thelounge thelounge  4096 Sep  4 15:26 packages
```

I appreciate the old Unix desire to make the server a party and let everyone read as much as possible, but config.js already being considered secret, and the reality that this is handling people's communication history and identities. I think a retrospective chmod becomes pragmatic

MaxLeiter:
After some discussion with @maxpoulin64 and others on IRC, I don't think this is the right way to solve this. We could replace the chmod with a warning log, but the TL shouldn't fix admin errors. The deb package (and others) can address this on their own in the postinst step or equivalent.

PR #4314, opened 2021-09-24
Properly mute a channel

brunnre8:
When a channel is muted it should neither generate a mention,
nor should it be highlighted specially.
This is done in a "soft" manner, meaning unmuting a channel brings it
back.

### DIFF
--- a/client/components/Channel.vue
+++ b/client/components/Channel.vue
@@ -1,9 +1,12 @@
 <template>
 	<ChannelWrapper ref="wrapper" v-bind="$props">
 		<span class="name">{{ channel.name }}</span>
-		<span v-if="channel.unread" :class="{highlight: channel.highlight}" class="badge">{{
-			unreadCount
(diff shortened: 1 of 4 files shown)


brunnre8:
@MaxLeiter ping, please take a look.
I restructured approximately everything, sorry for that.
But the conditionals were nested *way* too deep, I stared at the code for way too long to figure out all the edge cases.

While the context menu version worked very well, the cli version /mute was buggy.
I've fixed those up.

Further the cli didn't emit the change event, so the client never updated. That's fixed too but please take a look if it makes sense.
I don't like the code duplication very much though, seems brittle.
What's the usual approach there? Generate a helper function somewhere and call it from server.js as well as the mute.js command?

brunnre8:
(sorry, now I'm really done, forgot to run prettier)

MaxLeiter (approved):

PR #4315, opened 2021-09-30
Add keyboard shortcut for help screen

### LINKED ISSUE #4286: Feature Request: Keyboard shortcut for help screen
I'd love to have a shortcut like <kbd>⌘</kbd>+<kbd>?</kbd> to open the keyboard shortcuts (aka help screen). Even more awesome would also be to be able to press <kbd>esc</kbd> to get out of the menu also.

***

Request aside, thanks a lot for thelounge, it's a slick IRC client for the modern times, really appreciate it. ❤️

NoahvdAa:
Fixes #4286.

### DIFF
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -3,7 +3,7 @@
 import Mousetrap from "mousetrap";
 
 import store from "./store";
-import {switchToChannel} from "./router";
+import {switchToChannel, navigate} from "./router";
 import isChannelCollapsed from "./helpers/isChannelCollapsed";
 import isIgnoredKeybind from "./helpers/isIgnoredKeybind";
 
@@ -103,6 +103,17 @@ Mousetrap.bind(["alt+a"], function (e) {
 	if (targetChannel) {
 		jumpToChannel(targetChannel);
 	}
+
+	return false;
+});
+
+// Show the help menu
+Mousetrap.bind(["command+/", "ctrl+/"], function (e) {
+	if (isIgnoredKeybind(e)) {
+		return true;
+	}
+
+	navigate("Help");
 
 	return false;
 });
--- a/client/js/keybinds.js
+++ b/client/js/keybinds.js
@@ -3,7 +3,7 @@
 import Mousetrap from "mousetrap";
 
 import store from "./store";
-import {switchToChannel, navigate} from "./router";
+import {switchToChannel, router, navigate} from "./router";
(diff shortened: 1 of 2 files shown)


client/js/keybinds.js
```diff
@@ -156,6 +169,13 @@ const ignoredKeys = {
 };
 
 document.addEventListener("keydown", (e) => {
+	// Allow navigating back to the previous page when on the help screen.
+	if (e.which === 27 && router.currentRoute.name === "Help" && previousRoute) {
```
MaxLeiter, line 173: iirc, `e.which` isn't recommended. Can we use `e.key === "Escape"`? It should have more support for non-standard keyboards.

MaxLeiter (changes requested):

client/js/keybinds.js
NoahvdAa: 👍 I used which because the rest of the file also seemed to use it. This makes more sense, fixed.

client/js/keybinds.js
```diff
@@ -107,6 +107,17 @@ Mousetrap.bind(["alt+a"], function (e) {
 	return false;
 });
 
+// Show the help menu
+Mousetrap.bind(["command+/", "ctrl+/"], function (e) {
```
brunnre8, line 112: All other navigation keys are Alt based ones, why does this use control?

Meaning Alt-/ would be more in line with what we currently have

client/js/keybinds.js
```diff
@@ -107,6 +107,19 @@ Mousetrap.bind(["alt+a"], function (e) {
 	return false;
 });
 
+// Show the help menu.
+let previousRoute;
+Mousetrap.bind(["command+/", "ctrl+/"], function (e) {
```
brunnre8, line 112: This is hacky.
Is there a particular reason we can't use our router?
https://router.vuejs.org/guide/essentials/navigation.html#router-go-n

brunnre8: line 111 that is, not sure why the other got included again

NoahvdAa: Don't know how I managed to miss that, fixed!

client/js/keybinds.js
NoahvdAa: The original issue (#4286) suggested cmd/ctrl, alt definitely fits in better.

MaxLeiter (approved):

MaxLeiter:
Thanks!

PR #4317, opened 2021-09-30
Show quit dialog when using /quit

MaxLeiter:
Supersedes https://github.com/thelounge/thelounge/pull/4236 by @Nachtalb

### DIFF
--- a/client/components/ConfirmDialog.vue
+++ b/client/components/ConfirmDialog.vue
@@ -3,7 +3,7 @@
 		<div v-if="data !== null" id="confirm-dialog">
 			<div class="confirm-text">
 				<div class="confirm-text-title">{{ data.title }}</div>
-				<p>{{ data.text }}</p>
+				<p v-html="data.text"></p>
 			</div>
 			<div class="confirm-buttons">
 				<button class="btn btn-cancel" @click="close(false)">Cancel</button>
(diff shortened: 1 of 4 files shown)


client/components/ConfirmDialog.vue
```diff
@@ -3,7 +3,7 @@
 		<div v-if="data !== null" id="confirm-dialog">
 			<div class="confirm-text">
 				<div class="confirm-text-title">{{ data.title }}</div>
-				<p>{{ data.text }}</p>
+				<p v-html="data.text"></p>
```
brunnre8, line 6: With this you introduce a security failure... data.text contains network.name which is user controlled.
Meaning this enables arbitrary injection of html with all that entails.
You should never use v-html for user controlled content

brunnre8:
Besides the html injection I've several issues with this.
* The popup uses btn-danger, btn-danger by default seems to be styled green (for the morning theme, danger should be some sort of red"
* Cancel should be the filled button / have an outline, as it is the "safe" thing to do, the actual "delete" should be the non obvious one. Guide the user to click the proper button while you are at it
* Can't navigate by keyboard. If the user typed /quit from the "cli" I expect that you can then do the operation without touching the mouse simply by selecting the button with the arrow keys / hitting enter

PR #4322, opened 2021-10-05
client/Mentions: change button copy to "Dismiss all"

williamboman:
```
14:56 <williamboman> Minor nitpick of mine: Am I the only one who finds the copy for the "Hide all" button in the Recent mentions popup a bit misleading? In my mind, hiding something suggests you can unhide it. What about "Dismiss all"?
15:00 <+bookworm> makes sense to me
15:00 <+bookworm> (s/hide/dismiss/ that is)
15:07 <xnaas> I concur on dismiss vs hide as well
```

### DIFF
--- a/client/components/Mentions.vue
+++ b/client/components/Mentions.vue
@@ -13,7 +13,7 @@
 					class="btn hide-all-mentions"
 					@click="hideAllMentions()"
 				>
-					Hide all
+					Dismiss all
 				</button>
 			</div>
 			<template v-if="resolvedMessages.length === 0">


client/components/Mentions.vue
```diff
@@ -13,7 +13,7 @@
 					class="btn hide-all-mentions"
 					@click="hideAllMentions()"
```
williamboman, line 14: Not sure if we should change the terminology from "hide" to "dismiss" across the codebase too

williamboman: https://github.com/thelounge/thelounge/pull/4322/commits/90df2951ac94299e0cb36962011314e1dd7f6127

client/components/Mentions.vue
```diff
@@ -37,11 +37,14 @@
 							</span>
 						</div>
 						<div>
-							<span class="close-tooltip tooltipped tooltipped-w" aria-label="Close">
+							<span
+								class="close-tooltip tooltipped tooltipped-w"
+								aria-label="Dismiss this mention"
```
williamboman, line 42: Seems to be a nested aria-label here, where the outer one has precedence in the UI (tooltip). Figured it'd make more sense to change this to be the same as the button's aria-label?

PR #4323, opened 2021-10-05
Support twitch.tv nick capitalization when sent in the display-name tag

awfulcooking:
Twitch lets users set capitalization for their usernames in Settings, and uses the display-name tag to confer it in chat, with nicks otherwise sent lowercase.

A large portion of Twitch users set it, and using the wrong capitalization to address a Twitch user looks out of place.

This patch promotes the correct capitalization to a message's nick field, when present.

I know that display-name has an ongoing spec [here](https://github.com/ircv3/ircv3-specifications/pull/452), and that Twitch may expand it for other purposes some day.

And that ideally, there would be a broader, more encompassing solution here.

But for the past N years, Twitch has used it solely for capitalization, and explained as such to users on the account settings page.

"Proper" display-name support would a) depend on ircv3/ircv3-specifications#452, b) be different in nature / scope / UX to what is done today by Twitch, c) require more QA and development than this simple change.

I'm hoping something like this can be considered to give better support for chatting on Twitch in the mean time.

A final note: the Twitch app itself, and the 3rd-party app [Chatty](https://chatty.github.io), are also unable to show capitalization for nicks, e.g. in the user list, until they receive a message with this tag.


awfulcooking:
This needs more work, following discussion on IRC, and actually receiving one of the asynchronous JOINs from Twitch in testing:

0. Should update the nickname in the channel user list, if an entry is present.
1. Should create an entry in the channel user list, if no entry is present.
2. Ensure a JOIN received after a user has sent messages to a channel does not overwrite the nick to lowercase in the user list

n.b. Twitch [no longer supports NAMES](https://discuss.dev.twitch.tv/t/irc-update-removing-mode-and-names-capabilities/25568/17) (/quote NAMES #chan -> 'Unknown command')

awfulcooking:
Been testing this for a month or so now and it works, definite QoL improvement for Twitch.

Points 0 and 1 are addressed with the latest commit. Point 2 has been non-essential and could be fixed in a subsequent PR

MaxLeiter:
@awfulcooking can you address the linting issues? https://github.com/thelounge/thelounge/runs/3843733230?check_suite_focus=true

awfulcooking:
Cheers. I didn't notice that was why the build had failed. Hopefully will pass now

MaxLeiter (approved):

brunnre8:
>Should create an entry in the channel user list, if no entry is present.

I fail to see where this happens in the code, care to point it out to me?
The only calls to setUser is from a nick / join event handler as far as I can tell.

Second question, you sure that shows an acceptable performance characteristic?
You are firing off "users" which in turn spawns "names"  which in turn calls client.names which calls "client.getSortedUsers"

Now, the thing does what it says on the tin but also copies all the values from the user map (and looking at twitch, the user list is rather long).

Just from a feeling doing that whenever someone talks seems incredibly slow, after all on twitch almost all people who joined are talking, at least it feels that way and you'll be doing that for every single message which is incoming until you've seen all the nicks once (then your conditional bails early as the user nick is already the display name).

Am I missing something?

brunnre8:
@awfulcooking are you interested in continuing this work?

awfulcooking:
Hi @brunnre8. Thanks for raising your concern about the frequency and overhead of the names sync to the client. A few mitigating things are:

- it only happens once per user, when they first speak with a display nick that differs what's stored in thelounge's user list
- somewhere under half of Twitch users use names that are not lowercase, so it won't trigger for everyone
- some reverse proxies like nginx will offer websocket compression by default, cutting down on transfer

I didn't notice performance trouble with channels like #limmy, and it's been great in the smaller channels I usually hang out in

However, if there's a better way to send a user list update to the client for a single user, it'd be great to use it. What do you think..

PR #4326, opened 2021-10-10
Allow network list reordering via touch.

itsjohncs:
Users can now long touch and drag a channel or network to change its
ordering in the network list.

--

Tested on: recent Chrome, Firefox, and Safari on Mac OS X; recent Safari on an iPhone 8; and Chrome on a several year old Android Nexus.

### DIFF
--- a/client/components/NetworkList.vue
+++ b/client/components/NetworkList.vue
@@ -81,6 +81,9 @@
 				class="network"
 				role="region"
 				aria-live="polite"
+				@touchstart="onDraggableTouchStart"
+				@touchend="onDraggableTouchEnd"
+				@touchcancel="onDraggableTouchEnd"
 			>
 				<NetworkLobby
 					:network="network"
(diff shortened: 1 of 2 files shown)


itsjohncs:
I've resolved the scrolling issue and this is now ready to review.

I made several changes to the original commit. They've been squashed but the full history will be maintained at [itsjohnc/thelounge@gestures-next-channel-unsquashed](https://github.com/itsjohncs/thelounge/tree/gestures-next-channel-unsquashed) while this PR is open.

itsjohncs:
Just updated the second commit with a small cleanup that only occurred to me a short bit ago. Full details within the unsquashed commit https://github.com/itsjohncs/thelounge/commit/9674f5e582ddc07862e799e929dd603914f9eb1d.

MaxLeiter:
It can be *slightly* janky if you try to scroll up/down while dragging or if you drag off to the right, but that can be fixed eventually if it bothers someone enough. This is a massive improvement; thanks!

itsjohncs:
@brunnre8 this is intentional. Those same context menu options also exist in the dropdown menu accessible through the three dots in the topic bar. Functionality isn't lost.

We'd have to implement drag-and-drop using a different method than long-touches if we want to allow long-touch to be "open context menu" on Android.

itsjohncs:
Summarizing discussion in IRC:

* Android long-press to activate context menu is important and shouldn't be removed.
* Android does not have canonical design pattern for reordering lists.
* I'll make a follow-up patch that allows Android's context menu to activate at the same time as the drag starting, but dismisses the context menu when user actually drags a channel some distance. Android's native Discord client uses this design for reordering the network list, so we get a vote of confidence from their design team at least.

PR #4328, opened 2021-10-10
Prevent sporadic rounding of search input on iOS.

itsjohncs:
Safari on iOS would sometimes override the styling of the search input
and set the `border-radius` much higher. The effect would then
dissappear if the DOM was changed.

--

Easiest way to reproduce this (on an iOS device) prior to this change is to open the sidebar, lock/turn your screen off, then turn your screen back on and take a look again. The search input will become rounded:

If you then hid and then reposed the sidebar, it would return to normal:

### DIFF
--- a/client/components/NetworkList.vue
+++ b/client/components/NetworkList.vue
@@ -141,6 +141,7 @@
 	color: #fff;
 	background-color: rgba(255, 255, 255, 0.1);
 	padding-right: 35px;
+	appearance: none;
 }
 
 .jump-to-input .input::placeholder {


aab12345:
John.  This also happens on the new search field on IOS 15.

aab12345:
> It likely affects all inputs we override the default styles for. @aab12345, you have the chance to test and provide feedback via issues. Running via master is not a guarantee for stability or no bugs. That's why we have the master and pre-release process.

Yeh, I understand - all good though. I do build it from (https://github.com/thelounge/thelounge.git) but maybe there's a cache issue some where. Speaking with John, it may be my testing environment so I'll start it from scratch from ground up soon.

All good for testing though!

PR #4332, opened 2021-10-13
Enable Android's context menus in network list.

itsjohncs:
After #4326 Android users could no longer long-touch to bring up the
context menu for channels in the network list. Now they can again.

--

@brunnre8 was concerned with flashing the drag styles for users who are just trying to get the context menu. Implementing the basic functionality in this commit took me awhile unfortunately so I haven't been able to work on that.

But I'll play around with the design to address the flashing problem and I'll share a followup change if I succeed.

### DIFF
--- a/client/components/ChannelWrapper.vue
+++ b/client/components/ChannelWrapper.vue
@@ -82,15 +82,11 @@ export default {
 			this.$root.switchToChannel(this.channel);
 		},
 		openContextMenu(event) {
-			// events.buttons will be 0 when the event is caused by a long
-			// touch on Android.
-			if (event.buttons !== 0) {
-				eventbus.emit("contextmenu:channel", {
-					event: event,
-					channel: this.channel,
-					network: this.network,
-				});
-			}
+			eventbus.emit("contextmenu:channel", {
+				event: event,
+				channel: this.channel,
+				network: this.network,
+			});
 		},
 	},
 };
--- a/client/components/ContextMenu.vue
+++ b/client/components/ContextMenu.vue
@@ -2,6 +2,7 @@
 	<div
 		v-if="isOpen"
 		id="context-menu-container"
+		:class="{passthrough}"
 		@click="containerClick"
 		@contextmenu.prevent="containerClick"
 		@keydown.exact.up.prevent="navigateMenu(-1)"
@@ -49,6 +50,7 @@ export default {
 	data() {
 		return {
 			isOpen: false,
+			passthrough: false,
 			previousActiveElement: null,
 			items: [],
 			activeItem: -1,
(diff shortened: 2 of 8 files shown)


itsjohncs:
`Activ8` on IRC reported that right-clicking on Windows OS w/ Latest Edge didn't work before this change, but does work afterwards. So sounds like my disabling of Android's context menu was triggering more than I thought.

Just more reasons to pull in this change.

aab12345:
Tested this on local env, can confirm right click is back with this commit, pulled and merged to local source.  

Tested in latest versions of Edge, FF and Opera browsers.

itsjohncs:
A user on IRC just confirmed that right click behavior on Windows Chrome is also broken prior to this change, and works afterwards (tested using a patched version of SortableJS). Which makes sense since Chrome and Edge are so closely related.

Since its been a little while since the upstream PR was opened and we haven't heard back, I'll add a commit to this PR implementing option (1) and we can talk through that.

itsjohncs:
Done. AFAIK PR should be mergeable as-is now. Take a look.

brunnre8:
Tested on Android 12, Chrome Beta 96.0.4664.17

I'm afraid we aren't quite there yet, it *almost* works.

However it still drops some of the touch events for no real reason (note I can't reproduce this in simple steps, it's always touching things in the message list until the channel list starts to drop touches).

Sometimes it works, other times it doesn't:
https://www.youtube.com/watch?v=O16I__4ZxMM

brunnre8:
Hm, seems to be reproducible if you focus something else, say the context menu via the three dots, then change to the channel list and try right clicking.

Might be a focus thing?

Yeah, got it reproducible with the steps above

itsjohncs:
> Hm, seems to be reproducible if you focus something else, say the context menu via the three dots, then change to the channel list and try right clicking.

That's a slightly unfortunate interaction, but I don't think a problem. The context menu is laying down a full-screen transparent block so that when you click somewhere else the context menu can be closed. You should be able to reproduce that interaction on master.

itsjohncs:
> However it still drops some of the touch events for no real reason (note I can't reproduce this in simple steps, it's always touching things in the message list until the channel list starts to drop touches).
>
> Sometimes it works, other times it doesn't:
https://www.youtube.com/watch?v=O16I__4ZxMM

I don't know that this is the case, but you might just be moving your finger slightly when touching. The sensitivities to movement for the OS to trigger a context menu event vs the sensitivity of Sortable to trigger a drag are not exactly the same (partly because the OS numbers are not available to us), so you may get one or the other to activate separately.

itsjohncs:
Running on master (with my previous commits reverted) I can get some intermittent right click event eating happening immediately after opening the sidebar. I'm not sure what's causing it but it looks like there's some existing touch-eating behavior 🤷‍♀️.

brunnre8:
>I don't know that this is the case, but you might just be moving your finger slightly when touching

I don't think that's it no.
Even if you are dragging the context menu isn't immediate dismissed after all.

itsjohncs:
I'm able to reproduce a case like the one in your video on `v4.3.0-pre.4` intermittently. I dragged sidebar open, long touched an item precisely and vibration triggers, context menu does not appear, then long touch again and context menu appears.

So I'm fairly sure that one is an existing behavior.

> Even if you are dragging the context menu isn't immediate dismissed after all.

I haven't been able to get this to happen though. You got the context menu to appear while dragging, and the menu was not dismissed?

brunnre8:
>You got the context menu to appear while dragging, and the menu was not dismissed?

No, what I was saying is that I don't think "you wiggled your finger while long pressing" was a likely cause. That's the first thing I tried to rule out 😉

itsjohncs:
Ah gotcha.

I haven't been able to find any behavior I can't reproduce on `v4.3.0-pre.4`. Do you mind smashing at an earlier version too and seeing if you get the same weirdness you're experiencing?

brunnre8:
will do, I'm not on my computer at the moment though and my instance now runs your patch (I test on my live instance because you have my absolute trust 😉 )

brunnre8:
For the record:  If the bug is already there on master I don't see a point in delaying the merge of the fix. After all the current state is broken, so thos PR makes things better, not worse

itsjohncs:
Cool 👍, sounds like this is good to go then.

brunnre8 (approved):

MaxLeiter (approved):

MaxLeiter:
Thanks @itsjohncs and @brunnre8 for digging deep with this. Great work

PR #4333, opened 2021-10-15
Fix sporadic rounding on message search bar.

itsjohncs:
This is fixed in the same way as #4328.

--

Reported by @aab12345. Here's a screenshot of the search bar displaying the same rounding issue:

Image of search bar with its corners overly rounded

And after the fix is applied:

Image of search bar with its corners a normal amount of rounded.


aab12345:
Tested this on local env, can confirm;

no more sporadic rounded with search in IOS with this commit, pulled and merged to local source.

PR #4334, opened 2021-10-16
Resolve issues from review of #4237.

itsjohncs:
This commit fixes any issues raised during the review of thelounge#4237. That
PR's author is no longer available so this split approach is necessary.

Several cosmetic changes were made, as well as these changes/fixes to
functionality:

1. The sort order of our queries was reverted back to its original
   behavior because the new ordering meant we'd retrieve the oldest
   messages from the DB instead of the newest. New tests were added to
   verify this from now on.
2. Failures in `TextFileMessageStorage.enable()` are now handled
   correctly.
3. Promises are returned from our test functions whenever possible.
   This allows the test runner to provide much better error reports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "yarn": "1.22.10"
   },
   "optionalDependencies": {
-    "better-sqlite3": "7.4.0"
+    "better-sqlite3": "7.4.3"
   },
   "devDependencies": {
     "@babel/core": "7.15.5",
--- a/src/plugins/messageStorage/sqlite.js
+++ b/src/plugins/messageStorage/sqlite.js
@@ -87,7 +87,7 @@ class MessageStorage {
 				})();
 			}
 		} catch (error) {
-			log.error(`Failed to initialize sqltie database: ${error}`);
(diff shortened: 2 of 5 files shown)


itsjohncs:
@brunnre8 is hitting some bugs in scroll back.

itsjohncs:
@brunnre8 debugged their issue. Thankfully at least one of the three of us dogfooding had enough messages in their DB for the bug to appear.

We must use the old sort orders we were using before when querying from `sqlite`. `SELECT * FROM tbl ORDER BY time ASC LIMIT 10` would get the first `10` items from `tbl` while ordering `DEC` would get the last `10` items.

We should have tests that catch this so I'll add some before i fix it. Thanks @brunnre8!

itsjohncs:
Fixed ordering issue as well as added unit tests for it. I've squashed the changes but an unsquashed version will be maintained at [itsjohncs/thelounge@finish-up-4237-unsquashed](https://github.com/itsjohncs/thelounge/tree/finish-up-4237-unsquashed) while this PR is open.

PR #4344, opened 2021-10-26
Csp

brunnre8:
In custom themes we want to be able to load remote fonts and pictures, the TL CSP header would in principle allow it, but we didn't actually apply them to the plugins / themes.

Rather than playing whack a mole let's just globally apply the CSP.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -50,6 +50,7 @@ module.exports = function (options = {}) {
 	app.set("env", "production")
 		.disable("x-powered-by")
 		.use(allRequests)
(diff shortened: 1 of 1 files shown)


itsjohncs:
LGTM. Just did some additional testing on an HTTP deployment as well, was curious about media previews but seems to be working well.

itsjohncs (approved):

MaxLeiter (approved):

MaxLeiter:
Going to dogfood this another day or two but seems to be working fine.

PR #4351, opened 2021-11-02
Allow wildcards in hostmask

brunnre8:
According to https://modern.ircdocs.horse/#wildcard-expressions
masks should support "*" and "?" wildcards.
Within TL this only impacts the /ignore functionality.

The reasoning for doing this is to ignore say GuestNNNN!*@* with
guest*!*@* and be done with it if someone spams a gateway.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -43,6 +43,7 @@ const Helper = {
 	getDefaultNick,
 	parseHostmask,
 	compareHostmask,
+	compareWithWildcard,
 
 	password: {
 		hash: passwordHash,
@@ -314,8 +315,27 @@ function parseHostmask(hostmask) {
 
 function compareHostmask(a, b) {
 	return (
-		(a.nick.toLowerCase() === b.nick.toLowerCase() || a.nick === "*") &&
-		(a.ident.toLowerCase() === b.ident.toLowerCase() || a.ident === "*") &&
-		(a.hostname.toLowerCase() === b.hostname.toLowerCase() || a.hostname === "*")
+		compareWithWildcard(a.nick, b.nick) &&
+		compareWithWildcard(a.ident, b.ident) &&
+		compareWithWildcard(a.hostname, b.hostname)
 	);
+}
+
+function compareWithWildcard(a, b) {
+	// we allow '*' and '?' wildcards in our comparison.
+	// this is mostly aligned with https://modern.ircdocs.horse/#wildcard-expressions
+	// but we do not support the escaping. The ABNF does not seem to be clear as to
+	// how to escape the escape char '\', which is valid in a nick,
+	// whereas the wildcards tend not to be (as per RFC1459).
+
+	// The "*" wildcard is ".*" in regex, "?" is "."
+	// so we tokenize and join with the proper char back together,
+	// escaping any other regex modifier
+	const wildmany_split = a.split("*").map((sub) => {
+		const wildone_split = sub.split("?").map((p) => _.escapeRegExp(p));
+		return wildone_split.join(".");
+	});
+	const user_regex = wildmany_split.join(".*");
+	const re = new RegExp(`^${user_regex}$`, "i"); // case insensitive
+	return re.test(b);
 }
--- a/test/tests/hostmask.js
+++ b/test/tests/hostmask.js
@@ -48,7 +48,14 @@ describe("Hostmask", function () {
 
 	it(".compareHostmask (wildcard)", function () {
 		const a = Helper.parseHostmask("nick!user@host");
-		const b = Helper.parseHostmask("nick!*@*");
+		const b = Helper.parseHostmask("n?ck!*@*");
+		expect(Helper.compareHostmask(b, a)).to.be.true;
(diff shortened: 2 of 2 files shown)


src/helper.js
```diff
@@ -314,8 +315,19 @@ function parseHostmask(hostmask) {
...
+		compareWithWildcard(a.nick, b.nick) &&
+		compareWithWildcard(a.ident, b.ident) &&
+		compareWithWildcard(a.hostname, b.hostname)
 	);
 }
+
+function compareWithWildcard(a, b) {
+	a = a.toLowerCase();
+	b = b.toLowerCase();
+	const wildcars_split = a.split("*").map((sub) => {
```
itsjohncs, line 327: Looks like this doesn't support escaping. The horse docs say `*` and `?` should be escapable with a `\`.

But TBH I'm not sure how important of a feature it is to support escaping so I wouldn't be too distraught if you just don't want to. Seems like it shouldn't be too difficult to support though.

src/helper.js
```diff
@@ -314,8 +315,19 @@ function parseHostmask(hostmask) {
...
+		const parts = sub.split("?").map((p) => _.escapeRegExp(p));
+		return parts.join(".");
+	});
+	const re = new RegExp(wildcars_split.join(".*"));
```
itsjohncs, line 331: If you split out the part of this function that generates the regex into its own function, you could write tests for that function that would be a bit easier to verify the correctness of.

If you pass `"i"` as a flag to `RegExp` it'll do a case insensitive comparison. Since we're using the regex engine now anyways we might as well let it handle that part of it.

itsjohncs (commented):
Had a few minor thoughts, nothing blocking or particularly important though.

src/helper.js
brunnre8: Yeah... thought about it but then dismissed it as being probably overkill... after all who uses a wildcard in a nick / ident / host.

But while we are at it we might just do it properly don't we?

src/helper.js
brunnre8: ack, will do. Thanks

src/helper.js
brunnre8: FTR: the spec is very much unclear how you'd then escape the escape char.
Considering that nicks can contain the \ char but not the wildcards we refrain from implementing the escaping.

src/helper.js
brunnre8: I didn't split the function but added the regex modifier (and a huge comment).
Considering that there are already extensive tests for the comparison function with quite a bunch of edge case scenarios I don't see a benefit of testing for the raw regex, we already know that we get the desired result. Hope that's ok.

test/tests/hostmask.js
```diff
@@ -60,3 +67,38 @@ describe("Hostmask", function () {
...
 });
+
+describe("compareWithWildcard", function () {
+	const goodPairs = [
+		["asdf", "asdf"],
+		["a?df*", "asdf"],
+		["*asdf*", "asdf"],
+		["*asdf", "asdf"],
+		["asd?", "asdf"],
+		["asd?*", "asdf"],
```
itsjohncs, line 78: Could you add these cases?

```js
["AsDf", "asdf"],
["a??f", "asdf"],
["a*", "asdf"],
["*f", "asdf"],
["*s*", "asdf"],
```

If `String.split()` worked like it did in Python, the last 4 cases would've failed. Thought I'd found a bug at first!

test/tests/hostmask.js
```diff
@@ -60,3 +67,38 @@ describe("Hostmask", function () {
...
+		});
+	}
+
+	const badPairs = [
+		["asdf", "fdsa"],
+		["a?df*", "adfg"],
+		["?", ""],
+		["?asdf", "asdf"],
+		["?*", ""],
+		["*?*", ""],
```
itsjohncs, line 95: Could you add these cases?

```js
["asd", "asdf"],
["sdf", "asdf"],
["sd", "asdf"],
["", "asdf"],
```

They currently fail. You'll want to add `^` and `$` to the regex to make sure it matches the whole string.

itsjohncs (changes requested):
I think it's fine as one function and not handling escaping sounds fine as well 👍.

I did find one bug that we ought to fix though. At least I think it's a bug. I don't think the mask `asd` is meant to match `asdf`.

test/tests/hostmask.js
brunnre8: Right, nice catch! Thanks

test/tests/hostmask.js
brunnre8: what do you mean?
```python
>>> "*".join("*s*".split("*"))
'*s*'
```

itsjohncs: Oh! I just got mixed up.

`split` without arguments will collapse empty splits but apparently not if you give a delimiter explicitly.

brunnre8:
PTAL, added the tests you suggested and fixed the code

itsjohncs (approved):
Thanks! LGTM

MaxLeiter (approved):

PR #4352, opened 2021-11-03
Prevent round and white search styling in iOS 15.

itsjohncs:
This is fixed the same way as in #4333.

--

Before fix is applied:

before

After fix is applied:

after


fnutt:
Tested on Apple iPhone 8 Plus running iOS 15.1. It works well and does in fact fix the issue with white areas and rounded shapes.

Thanks for the PR, much appreciated!

MaxLeiter:
Merging, but in the future someone can feel free to create a PR replacing this with a global ` input[type=text] { appearance: none }` after some testing

PR #4360, opened 2021-11-15
Disable /search and hide help item if not enabled

### LINKED ISSUE #4273: Disable `/search query` if search is not possible.

MaxLeiter:
Fixes #4273 and supersedes #4278.

Removes `/search` from the client side commands if it's not enabled and hides its help item


client/js/commands/search.js
```diff
@@ -4,10 +4,15 @@ import store from "../store";
 import {router} from "../router";
 
 function input(args) {
+	if (!store.state.settings.searchEnabled) {
+		return true;
+	}
+
+	const {channel} = store.state.activeChannel;
```
itsjohncs, line 11: This seems like a fairly random stylistic change to include here and doesn't seem necessary.

client/js/commands/search.js
```diff
@@ -4,10 +4,15 @@ import store from "../store";
 import {router} from "../router";
 
 function input(args) {
+	if (!store.state.settings.searchEnabled) {
+		return true;
```
itsjohncs, line 8: I wonder if just returning false here would be reasonable. We'd end up behaving as if the command is completely unknown, which seems reasonable given that we're stripping its existence from the UI while its disabled.

src/plugins/inputs/index.js
```diff
@@ -47,7 +45,6 @@ const getCommands = () =>
 	Array.from(userInputs.keys())
 		.concat(Array.from(pluginCommands.keys()))
 		.map((command) => `/${command}`)
-		.concat(clientSideCommands)
```
itsjohncs, line 50: I like that we're not sending the client the commands it already knows now :).

itsjohncs (commented):
Haven't tested this yet, but did a somewhat thorough look-through the code. Hopefully this is the sort of review you're after. I'm happy to focus solely on functionality instead of softer/opinionated qualities.

src/plugins/inputs/index.js
MaxLeiter: This is now done in `client/js/socket-events/commands.js`, but thats a change from the previous PR and not strictly necessary.

MaxLeiter:
@itsjohncs can't re-request review for some reason, but I addressed your comments.

client/js/autocompletion.js
```diff
@@ -311,10 +311,19 @@ function completeNicks(word, isFuzzy) {
 	return users.filter((w) => !w.toLowerCase().indexOf(word));
 }
 
-function completeCommands(word) {
-	const words = constants.commands.slice();
+function getCommands() {
+	let cmds = constants.commands.slice();
 
-	return fuzzyGrep(word, words);
+	if (store.state.settings.searchEnabled === false) {
```
itsjohncs, line 317: why is this being compared to false rather than checking to see if its falsy?

itsjohncs:
digging it, going to quickly dogfood it this evening and see if i can shake out any actual bugs.

itsjohncs (approved):
Shook the tree a bit and didn't see any bugs fall out. Looks great :). Glad you were able to get to this.

PR #4361, opened 2021-11-18
Add /kickban

supertassu:
This patch adds a new command, /kickban, that is a combination of /kick and /ban: it kicks the specific user from the channel and then sets the +b mode to ban the user from the channel.

### DIFF
--- a/client/components/Windows/Help.vue
+++ b/client/components/Windows/Help.vue
@@ -583,10 +583,22 @@
 
 			<div class="help-item">
 				<div class="subject">
-					<code>/kick nick</code>
+					<code>/kick nick [reason]</code>
 				</div>
 				<div class="description">
 					<p>Kick a user from the current channel.</p>
+				</div>
+			</div>
+
+			<div class="help-item">
+				<div class="subject">
+					<code>/kickban nick [reason]</code>
+				</div>
+				<div class="description">
+					<p>
+						Kick and ban (<code>+b</code>) a user from the current channel. Unlike
+						<code>/ban</code>, only nicknames (and not host masks) can be used.
+					</p>
 				</div>
 			</div>
 
--- a/src/plugins/inputs/ban.js
+++ b/src/plugins/inputs/ban.js
@@ -3,7 +3,7 @@
 const Chan = require("../../models/chan");
 const Msg = require("../../models/msg");
 
-exports.commands = ["ban", "unban", "banlist"];
+exports.commands = ["ban", "unban", "banlist", "kickban"];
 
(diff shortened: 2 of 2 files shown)


brunnre8:
please add the docs to the help window

supertassu:
> please add the docs to the help window

Done, thanks.

client/components/Windows/Help.vue
```diff
@@ -590,6 +590,15 @@
 				</div>
 			</div>
 
+			<div class="help-item">
+				<div class="subject">
+					<code>/kickban nick</code>
```
brunnre8, line 595: except your code works differently than what you say here.
So, which one is wrong, doc or the code 😜 ?

supertassu: You're referring to the reason, right? I'll fix the docs if that's the case

brunnre8: Yes, look at other commands as to how to specify optional parameters.

Also worth noting that /ban will work differently, as that one accepts a banmask, kick only works with a nick.

client/components/Windows/Help.vue
```diff
@@ -583,13 +583,25 @@
...
 			</div>
 
+			<div class="help-item">
+				<div class="subject">
+					<code>/kickban nick [reason]</code>
+				</div>
+				<div class="description">
+					<p>
+						Kick and ban (<code>+b</code>) a user from the current channel. Unlike /ban,
+						only user names (and not host masks) can be used.
```
brunnre8, line 600: s/user names/nick/ please. The term nick has a rather specific meaning. "user name" is a bit more up for interpretation (do you mean nick or realname?)

supertassu: Sure, done!

client/components/Windows/Help.vue
```diff
@@ -583,13 +583,25 @@
...
 				</div>
 			</div>
 
+			<div class="help-item">
+				<div class="subject">
+					<code>/kickban nick [reason]</code>
+				</div>
+				<div class="description">
+					<p>
+						Kick and ban (<code>+b</code>) a user from the current channel. Unlike /ban,
```
MaxLeiter, line 599: > Unlike `<code>ban</code>`

MaxLeiter (changes requested):

client/components/Windows/Help.vue
supertassu: Done

brunnre8 (approved):
LGTM, thanks a lot for your persistence.

PR #4363, opened 2021-11-18
Don't download image contents during prefetch if not needed

sfan5:
Very minor thing.
For image mime types only `res.type` or `res.size` are ever accessed (unless of course prefetchStorage is enabled) so `fetch()` can abort early to avoid downloading unneeded data.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -403,7 +403,8 @@ function fetch(uri, headers) {
 					if (imageTypeRegex.test(contentType)) {
 						// response is an image
 						// if Content-Length header reports a size exceeding the prefetch limit, abort fetch
-						if (contentLength > limit) {
+						// and if file is not to be stored we don't need to download further either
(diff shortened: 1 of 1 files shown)


src/plugins/irc-events/link.js
```diff
@@ -406,6 +406,10 @@ function fetch(uri, headers) {
 						if (contentLength > limit) {
 							gotStream.destroy();
 						}
+						// If file is not to be stored we don't need to download further either
+						else if (!Helper.config.prefetchStorage) {
```
itsjohncs, line 410: I'd prefer to see these two branches joined since they have the same body. ie:

```js
if (contentLength > limit || !Helper.config.prefetchStorage) {
    gotStream.destroy();
}
```

itsjohncs (approved):
Did a quick test locally and looks good to me.

brunnre8 (approved):

MaxLeiter:
Thanks for the PR, @sfan5. This will be merged once 4.3 is fully released.

PR #4365, opened 2021-11-18
Mentions esc

brunnre8:
I mean, while we are at it?

### DIFF
--- a/client/components/Mentions.vue
+++ b/client/components/Mentions.vue
@@ -185,10 +185,10 @@ export default {
 		},
 	},
 	mounted() {
-		eventbus.on("mentions:toggle", this.openPopup);
+		eventbus.on("mentions:toggle", this.togglePopup);
 	},
 	destroyed() {
-		eventbus.off("mentions:toggle", this.openPopup);
+		eventbus.off("mentions:toggle", this.togglePopup);
 	},
 	methods: {
 		messageTime(time) {
@@ -211,7 +211,7 @@ export default {
 				this.isOpen = false;
 			}
 		},
-		openPopup() {
+		togglePopup() {
 			this.isOpen = !this.isOpen;
 
 			if (this.isOpen) {
--- a/client/components/Mentions.vue
+++ b/client/components/Mentions.vue
@@ -186,9 +186,11 @@ export default {
 	},
 	mounted() {
 		eventbus.on("mentions:toggle", this.togglePopup);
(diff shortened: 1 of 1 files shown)


brunnre8:
yes, yes... linters. I know. Fixed.

client/components/Mentions.vue
```diff
@@ -211,14 +211,20 @@ export default {
...
-		openPopup() {
+		togglePopup() {
 			this.isOpen = !this.isOpen;
 
 			if (this.isOpen) {
 				this.isLoading = true;
 				socket.emit("mentions:get");
+				eventbus.on("escapekey", this.closePopup);
+			} else {
+				eventbus.off("escapekey", this.closePopup);
```
itsjohncs, line 222: We should to this on destruction as well. Hot reloading can cause destruction at unusual times but I bet there's non-dev-server ways for this component to be destroyed without `togglePopup` being called first as well.

client/components/Mentions.vue
```diff
@@ -211,14 +211,20 @@ export default {
 				this.isOpen = false;
 			}
 		},
-		openPopup() {
+		togglePopup() {
```
itsjohncs, line 214: Love the name change.

client/components/Mentions.vue
```diff
@@ -211,14 +211,20 @@ export default {
...
 			this.isOpen = !this.isOpen;
 
 			if (this.isOpen) {
 				this.isLoading = true;
 				socket.emit("mentions:get");
+				eventbus.on("escapekey", this.closePopup);
```
itsjohncs, line 220: Since `closePopup` is called when escape is hit, but it doesn't unhook the event bus, we can end up adding a bunch of callbacks. Ex:

1. `togglePopup` is called. Callback is added.
2. Escape is pressed. Popup is  closed.
3. `togglePopup` is called. Callback is added.

We now have 2 callbacks added. And we could continue doing this to get as many callbacks as we'd like.

It might be simplest to add the hook when the component is mounted and then remove it when the component is destroyed. This'd take care of my other comment as well.

itsjohncs (changes requested):

brunnre8:
well, what John said :smiley: 

That should address both review comments

itsjohncs (approved):
Didn't actually test this but looks good to me.

MaxLeiter (approved):

PR #4367, opened 2021-11-21
Revert "Preserve location on first and last line when scrolling through inputs"

### LINKED ISSUE #4362: Fix regressions from c5f6b46 or make it a per-user option
### Feature Description
In c5f6b46 a change was made that is rather drastic.

Example scenario: You need some help from ChanServ.

1. `/cs HELP`
2. *read output*
3. `UpArrow`

Now, at this point, the old behavior was to simply press space and add a new argument (e.g. `/cs HELP IDENTIFY`), but now you're at the beginning of the line and have to add an extra key press: `end`.

The above is just one example of a regression caused by the change.

MaxLeiter:
This reverts commit [c5f6b46](https://github.com/thelounge/thelounge/pull/4205/commits), half of PR #4205. The cursor location is now the end of the input when you go up or down. 

Closes #4362.

Note the other commit in the original PR, `Only scroll history when cursor is on first or last row`, is untouched by this.

### DIFF
--- a/client/components/ChatInput.vue
+++ b/client/components/ChatInput.vue
@@ -140,10 +140,11 @@ export default {
 				return;
 			}
 
-			const oldValue = this.$refs.input.value;
-			const oldPosition = this.$refs.input.selectionStart;
-			const onRow = (oldValue.slice(null, oldPosition).match(/\n/g) || []).length;
-			const totalRows = (oldValue.match(/\n/g) || []).length;
+			const onRow = (
+				this.$refs.input.value.slice(null, this.$refs.input.selectionStart).match(/\n/g) ||
+				[]
+			).length;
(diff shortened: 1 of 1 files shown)


xnaas:
Just confirming in-issue what was said via IRC:

>  <xnaas> Tested #4367 MaxLeiter. It does appear to work as it did before.
> 
>  <xnaas> Also, I don't know if "Only scroll history when cursor is on first or last row" is supposed to do what it says...but I haven't seen that behavior in pre.6 or rc.1, in Firefox. At any point I press an up or down arrow, my typed history would cycle.

MaxLeiter:
from xnaas on IRC: `Tested #4367 MaxLeiter. It does appear to work as it did before.`

PR #4372, opened 2021-11-23
Offer to upload text to server if user tries to send a long message

MaxLeiter:
Seems to work okay, except that cancel/escape either needs to A) send the message anyways (current behavior) or B) close the modal and not send. The problem with B is that then there's no way to send the message without uploading it. 

I need to either revamp the ConfirmDialog to support multiple actions, allow the caller to distinguish how it was cancelled, or figure something else out.

### DIFF
--- a/client/components/ChatInput.vue
+++ b/client/components/ChatInput.vue
@@ -260,10 +260,30 @@ export default {
 				autocompletionRef.hide();
 			}
 
-			this.channel.inputHistoryPosition = 0;
-			this.channel.pendingMessage = "";
-			this.$refs.input.value = "";
-			this.setInputSize();
+			const resetInput = () => {
+				this.channel.inputHistoryPosition = 0;
+				this.channel.pendingMessage = "";
+				this.$refs.input.value = "";
+				this.setInputSize();
+			};
+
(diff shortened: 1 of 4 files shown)


client/components/ChatInput.vue
```diff
@@ -260,34 +260,73 @@ export default {
...
-			if (text[0] === "/") {
-				const args = text.substr(1).split(" ");
-				const cmd = args.shift().toLowerCase();
+				socket.emit("input", {target, text});
+			};
+
+			if (this.$store.state.serverConfiguration.fileUpload) {
+				const lines = 1 + (text.match(/\n/g) || "").length;
+
+				if (lines > 3 || text.length > 700) {
```
igalic, line 301: what happens when someone pastes more than 700 lines?

igalic (commented):
❓

client/components/ChatInput.vue
brunnre8: the same thing that happens for text with more than 3?

lines > 3 || text.length > 700

means "either over 3 lines or text longer than 700 characters"

Not sure I understand your issue

igalic: well, reading sure is hard.
sorry for the noise

PR #4373, opened 2021-11-23
Permissions

brunnre8:
We still have some issues related to packaging.

That's essentially the issue raised at https://github.com/thelounge/thelounge-deb/pull/72

However the issue is persistent over all installation methods and should be fixed in core.
Rather than messing with the filesystem and trying to fix the mess we created we simply warn and let the admins handle it.

While not ideal, at least that solves the issue for new installs and we can always fix it in the packages of the more opinionated packages (read Debian) and maybe warn on do it yourself distros (Arch)

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -158,6 +158,34 @@ function setHome(newPath) {
 	// Load theme color from the web manifest
 	const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf8"));
 	this.config.themeColor = manifest.theme_color;
+
+	// log dir probably shouldn't be world accessible.
+	// Create it with the desired permission bits if it doesn't exist yet.
+	let logsStat = undefined;
+
+	try {
+		logsStat = fs.statSync(userLogsPath);
+	} catch {
+		// ignored on purpose, node v14.17.0 will give us {throwIfNoEntry: false}
+	}
+
+	if (!logsStat) {
+		try {
+			fs.mkdirSync(userLogsPath, {recursive: true, mode: 0o750});
+		} catch (e) {
+			log.error("Unable to create logs directory", e);
+		}
+	} else if (logsStat && logsStat.mode & 0o001) {
+		log.warn(
+			"contents of",
+			userLogsPath,
+			"can be accessed by any user, the log files may be exposed"
+		);
+
+		if (os.platform() !== "win32") {
+			log.warn(`run \`chmod o-x ${userLogsPath}\` to correct it`);
+		}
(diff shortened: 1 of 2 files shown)


src/helper.js
```diff
@@ -158,6 +159,30 @@ function setHome(newPath) {
 	// Load theme color from the web manifest
 	const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf8"));
 	this.config.themeColor = manifest.theme_color;
+
+	// log dir probably shouldn't be world accessible.
+	// Create it with the desired permission bits if it doesn't exist yet.
+	let logsStat = undefined;
+
+	try {
+		logsStat = fs.statSync(userLogsPath);
```
MaxLeiter, line 168: can we not `fs.statSync(userLogsPath, { throwIfNoEntry: false })`? I'm confused by the comment in the catch

MaxLeiter (changes requested):

src/helper.js
brunnre8: we can't, that's a feature of v14.17.0 butour minimum version is 12

MaxLeiter (approved):
The comment could be slightly better worded (`node...will give us {}` => `node v14.17.0 will provide an option to { throwIfNoEntry: false } ` maybe) but nbd. Thanks

src/plugins/webpush.js
```diff
@@ -11,7 +11,22 @@ class WebPush {
...
+			vapidStat = fs.statSync(vapidPath);
+		} catch {
+			// ignored on purpose, node v14.17.0 will give us {throwIfNoEntry: false}
+		}
+
+		if (vapidStat) {
+			if (Helper.isWorldReadable(vapidStat)) {
+				log.warn(
+					vapidPath,
+					"is world readable. The file contains secrets. Please fix the permissions"
```
MaxLeiter, line 26: can we provide a link to directions to fix the instructions or include them here?

MaxLeiter: I also think `isWorldReadable` should just live here unless you see another use for it in the nearish future.

MaxLeiter (changes requested):

src/plugins/webpush.js
brunnre8: re inlining, sure will do.
In regards to how to fix it, well... Ignoring windows? No clue how to do it there but for unixy systems chmod will do.

brunnre8: done. PTAL

PR #4427, opened 2021-12-04
handle RPL_UMODEIS

brunnre8:


src/plugins/irc-events/mode.js
```diff
@@ -39,6 +39,17 @@ module.exports = function (irc, network) {
 		targetChan.pushMessage(client, msg);
 	});
 
+	irc.on("user info", function (data) {
+		const serverChan = network.channels[0];
```
MaxLeiter, line 43: do we want to show this in the open channel if it exists?

brunnre8: If you prefer?
I wasn't quite sure what to do with it, so I shoved it into the server buffer

MaxLeiter: Yeah, I think this should go in `activeChannel`, similar to notices targeted at the user.

brunnre8:
Added the corresponding showInActive flag, should be good now.

MaxLeiter (approved):

itsjohncs (approved):
Did a quick test, enough to see the message (and confirmed it'll appear in the active channel). LGTM

PR #4433, opened 2021-12-18
sqlite3: switch to microsoft/vscode fork

brunnre8:
Mapbox apparently lost interest in maintaining the package.
It still forces us to an outdated node-gyp version which also
makes python2 necessary which is slowly getting removed from
various OSes (Fedora, Arch...)

Microsoft started their maintained fork for vscode, so let's use that

Note: this is a young fork and they removed the pre-built binary blobs...
Not sure if we should revive the better-sqlite PR instead and try to fix the load issues rather than hoping for microsoft to bundle binaries... fairly sure they don't care as they vendor everything to their end user.

Meaning let's track: https://github.com/microsoft/vscode-node-sqlite3/issues/14

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "yarn": "1.22.17"
   },
   "optionalDependencies": {
-    "sqlite3": "5.0.2"
+    "@vscode/sqlite3": "^5.0.7"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


brunnre8:
No indication that MS wants to be upstream, bumped sqlite3 instead to the latest commit

PR #4438, opened 2021-12-31
Count number of mode changes, not MODE messages when creating condensed messages

supertassu:
Update the code in MessageCondensed that generates the condensed messages ("X users have joined, Y modes were set") to count the number of actual mode changes instead of the raw count of MODE messages. One MODE message can contain multiple mode changes.


MaxLeiter (approved):

brunnre8:
Thanks a bunch

PR #4446, opened 2022-01-24
sqlite3: update to latest commit

brunnre8:
It doesn't look like upstream wants to release a new version.
However, it forces us to use python2 and a insecure tar version.
So staying on the release is not really an option.
Mitigate it by switching to the latest commit in the repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "yarn": "1.22.17"
   },
   "optionalDependencies": {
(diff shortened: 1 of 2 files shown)


itsjohncs (approved):
I took a look at the commits in mapbox/node-sqlite3 since the last release and there's very few. This seems safe enough to do though I wonder if it'll bork any of our builds when we go to deploy.

Issue #1981, opened 2018-01-14
Make auth plugins more pluggable


merlinthp:
Issue #1953 mentions that in future auth plugins should be shipped separately from lounge itself.  That'll require a more formal auth plugin API, and some code being moved out of the main server code into plugins (mainly the local auth password change code).  Having taken a look at the current code, here's my idea on how to do this:

- Deprecate the public option, replace it with an auth option which specifies which auth plugin to use (a value of 'none' would equate to public).
- Add code to convert the current public and ldap.enable options to the appropriate auth setting, and write log entries encouraging the admin to update their config
- Don't have a hard-coded set of require() statements for auth plugins, just try to require() the plugin specified by the auth option
- Get rid of the isEnabled export in the existing auth modules, as they won't be needed
- Deprecate the ldap.enable option, and write log entries as with the public option
- Add a mechanism for auth modules to specify that they can change user passwords
- Change the ldapEnabled value sent to clients to "password change available" to make it auth module agnostic
- Move the local auth password change code out of the main server.js change-password handler into the local auth plugin

Thoughts?  I can work up a PR based on feedback.

metsjeesus:
I have tried to make my own auth method over a year now, but allways fail on timings. Tried to make Promises, kind of works, but looks nice and has its own problems. Maybe to a event based system is better pluggable then list, names and promises?

poVoq:
Yes this would be appreciated, especially if this would lead to other auth methods becoming available, such as an external user SQL database (for example from a forum) or OpenID Connect / Oauth2 support.

xPaw:
If there was an update someone would have left a comment. If you need this, feel free to work on it.

BrainWart:
I have interest in helping with this. I have not reviewed the entire auth flow yet, but my high level thoughts are to
1. Move auth plugin initialization to server.js
  - This will let us create new endpoints
2. Stack the authentication methods
  - If one fails fallback to the next? or maybe a selection for the auth method
3. Convert the existing auth plugins to "classes" and inject the app, config, etc through the initializer.

After a rough conversion to this, I plan to implement OIDC or import an external OIDC library and tie it in to see what might need to change for passwordless auth.

I am concerned how multiple providers might be handled by the internal user management so I'll be poking around there a bit to see how that might be handled securely.

BrainWart:
[...]

PR #4447, opened 2022-01-25
Allow HTTP User Header Auth

### LINKED ISSUE #2065: Support Oauth with Google
Hey! I really loved testing out thelounge, but the process for creating users is really tedious and they'd have to tell me their passwords for that. Also, I have to manually create users instead of having an Oauth based sign in strategy.

Is there any plan on adding Oauth? Is it a likely to be tough job?

### LINKED ISSUE #3692: Support Auth Proxy Authentication
### Feature Description

Support reading a specific HTTP request header (ex. `X-WEBAUTH-USER`) to authenticate a logged in user.

### Background

I have a reverse proxy sitting at my network edge that routes traffic to the proper frontends. In addition this proxy handles authentication and passes information about the logged-in user via headers to the frontend. This is nice since is absolves applications from need to do anything authentication work other than blindly trusting the header provided to them.

### Prior Art

This FR is very similar to what [Grafana supports](https://grafana.com/docs/grafana/latest/auth/auth-proxy/), however, it can be simplified quite a bit since The Lounge really just needs the username of the logged-in user -- nothing else.

### Initial Investigation

I poked around the code a little bit to see how feasible this. My notes are:

- HTTP headers from the initial WS upgrade request are persisted on the server side so we should be able to read the extra headers.
- Unfortunately the current authentication code assumes that user interaction will take place (i.e. filling out the login form). It doesn't currently have a mechanism for automatically logging someone in via an authentication plugin.

With the second point resolved this should be pretty trivial to implement since we basically just need configuration to tell us what header to read and then The Lounge should just read it and blindly trust it.

Should this be implemented we should document best practices here, in particular that this should only be use if your app is only accessible via a reverse proxy and that the proxy strips the header from user requests.

### LINKED ISSUE #4143: SSO with SAML or OpenID Connect
I would like to request a feature, and that is SSO. 
I would love to see OAuth2 (OpenID Connect or SAML2) implemented in the application. 
Having this feature would benefit others that share their applications with their family and friends. 

In my case, it would allow my friends that already have a Keycloak account in my environment, to use this application in Private Mode, and automatically provision the accounts by using their own credentials, without having me to create a username and password for them. 

For my current situation, to avoid user creation, I would just enable it in private mode, but this will not preserve chat history.

### LINKED ISSUE #4227: Web Manifest not loaded when using HTTP basic auth
- _Node version: 12.18.3_
- _Browser version: Chromium 90.0.4430.93_
- _Device, operating system: Desktop, manjaro_
- _The Lounge version: 4.2.0_

---
The file 'thelounge.webmanifest' fails to load with a 401 Unauthorized error when hosted behind HTTP basic auth.
After adding 'crossorigin="use-credentials"' as described here https://github.com/aws-amplify/amplify-console/issues/195 the manifest is loaded as expected and it's possible to install thelounge as a PWA.

I'm running thelounge in a docker container and this worked for me as a temporary fix: `sed -i 's/link rel="manifest"/link rel="manifest" crossorigin="use-credentials"/g' /usr/local/share/.config/yarn/global/node_modules/thelounge/client/index.html.tpl`

9p4:
By allowing HTTP user header authentication, we allow for TL to run behind an SSO implementation (such as Keycloak or Authelia or Gluu or even Google and Microsoft), or other forms of authentication done at the reverse proxy level.

Todos:

- [x] Tests
- [x] Docs

Here is the Apache configuration I used to test. I used [htpasswd](https://www.interserver.net/tips/kb/apache-htpasswd-authentication-ubuntu/) as an authentication method and converted the `REMOTE_USER` header to `Proxy-User`.

```
RewriteEngine On
RewriteCond %{REQUEST_URI}  ^/socket.io            [NC]
RewriteCond %{QUERY_STRING} transport=websocket    [NC]
RewriteRule /(.*)           ws://localhost:9000/$1 [P,L]

ErrorDocument 401 'Sign in please'

RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
ProxyVia On
ProxyRequests Off
ProxyAddHeaders On
ProxyPass / http://localhost:9000/
ProxyPassReverse / http://localhost:9000/
# If your auth plugin doesn't send a header (and instead has a variable)
RewriteRule .* - [E=PROXY_USER:%{LA-U:REMOTE_USER}] # note mod_rewrite's lookahead option
RequestHeader set Proxy-User %{PROXY_USER}e
 
<Location />
        AuthType Basic
        AuthName "Restricted Files"
        AuthBasicProvider file
        AuthUserFile "/srv/http/.htpasswd"
        Require valid-user
</Location>
<LocationMatch ^/img/.*$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>
<LocationMatch ^/thelounge\.webmanifest$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>
<LocationMatch ^/favicon\.ico$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>
<LocationMatch ^/robots\.txt$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>
 
# By default Apache times out connections after one minute
ProxyTimeout 86400
```

For [`mod_auth_openidc`](https://github.com/zmartzone/mod_auth_openidc):

```
RewriteEngine On
RewriteCond %{REQUEST_URI}  ^/socket.io            [NC]
RewriteCond %{QUERY_STRING} transport=websocket    [NC]
RewriteRule /(.*)           ws://localhost:9000/$1 [P,L]

RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
ProxyVia On
ProxyRequests Off
ProxyAddHeaders On
ProxyPass / http://localhost:9000/
ProxyPassReverse / http://localhost:9000/

OIDCSessionType server-cache:persistent
OIDCSessionCacheFallbackToCookie Off
OIDCStateMaxNumberOfCookies 10 true

OIDCProviderMetadataURL https://keycloak.example.com/realms/test/.well-known/openid-configuration
OIDCClientID thelounge
OIDCClientSecret aaa

# OIDCRedirectURI is a vanity URL that must point to a path protected by this module but must NOT point to any content
OIDCRedirectURI http://127.0.0.1/r
OIDCCryptoPassphrase aaa
OIDCRemoteUserClaim preferred_username
ErrorDocument 401 'Sign-in failed. Are you allowed access?'

OIDCSessionInactivityTimeout 10

<Location />
        AuthType openid-connect
	<RequireAll>
		Require valid-user
		Require claim roles:thelounge
	</RequireAll>
</Location>
<LocationMatch ^/img/.*$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>
<LocationMatch ^/thelounge\.webmanifest$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>
<LocationMatch ^/favicon\.ico$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>
<LocationMatch ^/service-worker\.js$>
    <Limit GET>
        Satisfy Any
    </Limit>
</LocationMatch>

# By default Apache times out connections after one minute
ProxyTimeout 86400
```

~~Currently, this seems to break PWAs (due to PWAs not being allowed to redirect the client)~~ PWA support seems to work (UX isn't perfect, the client has to reload more times than needed, but it works fairly well).

Fixes #2065; Fixes #3692; Fixes #4143; Fixes #4227


9p4:
Here is the Apache configuration I used to test. I used [htpasswd](https://www.interserver.net/tips/kb/apache-htpasswd-authentication-ubuntu/) as an authentication method and converted the `REMOTE_USER` header to `Proxy-User`.

```
RewriteEngine On
RewriteCond %{REQUEST_URI}  ^/socket.io            [NC]
RewriteCond %{QUERY_STRING} transport=websocket    [NC]
RewriteRule /(.*)           ws://localhost:9000/$1 [P,L]
 
RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
ProxyVia On
ProxyRequests Off
ProxyAddHeaders On
ProxyPass / http://localhost:9000/
ProxyPassReverse / http://localhost:9000/
RewriteRule .* - [E=PROXY_USER:%{LA-U:REMOTE_USER}] # note mod_rewrite's lookahead option
RequestHeader set Proxy-User %{PROXY_USER}e
 
<Proxy *>
        AuthType Basic
        AuthName "Restricted Files"
        AuthBasicProvider file
        AuthUserFile "/srv/http/.htpasswd"
        Require valid-user
</Proxy>
 
# By default Apache times out connections after one minute
ProxyTimeout 86400
```

9p4:
Should solve #4143

brunnre8:
Reading #1981 this approach (integrating in core) was already dismissed.

9p4:
I'd refrain from merging then. Any eta for when the restructuring will be completed (or started)? It seems like the last bit of activity was two years ago.

brunnre8:
No one is working on it as far as I know.

itsjohncs:
TBH this header approach seems more practical and I'd prefer it over the solution in #1981. Turning the auth system into something that's entirely plugin-based is sure to be a large effort and I doubt we'll actually end up with a less complex system for our trouble.

Notably most things you could do with the plugin system you could also do via this header auth based system.

This header approach also has a dev who's actively interested in pursuing it in @9p4, versus the plugin-auth which may never attract such a person.

9p4:
Apache configs have to be updated with 
```
OIDCSessionMaxDuration 60
OIDCStateMaxNumberOfCookies 10 true
```

9p4:
After updating the Apache config (see original comment), PWAs and reauth all seem to be working. I'll daily drive this for a few days and let you all know how this goes.

9p4:
The UX for PWAs seem to be at what is expected: sign in on Safari, add to home screen, sign in again in the PWA. Occasional reauths required (done in the PWA) based on the reverse proxy level.

Some issues currently: the PWA has to refresh every time there is a 401 because of the reauth required (so when the app is closed and reopened, it loads, refreshes, then loads again). This, while minor, is not the best solution.

Ignoring that, this seems to be suitable for merge for the use cases I have tested with. Once this is merged, I'll add some documentation on https://github.com/thelounge/thelounge.github.io

itsjohncs:
Just pulled this into my dogfood branch. Will run it and play around with it for the next few days. Will hopefully have some time for a code review soon as well.

brunnre8 (changes requested):
This needs some work in my opinion.
The code misses quite a few places where ldap does something special, yet header auth doesn't.
Pretty sure all those places are now buggy for the header auth codepath.

Generally, what you are doing is pretty much the same as ldap so chances are you need to handle things the same way.

That in essence means that you need to refactor ldap and header auth to extract the common behavior (don't just litter additional conditionals in there).

I personally don't like integrating another auth plugin. John seems to disagree on that. So I guess we need Max as a tie breaker.

brunnre8:
>Once this is merged

Just to make it clear. Should this be considered for merging it certainly needs tests first for a feature that's this intrusive and security sensitive.

client/js/socket-events/auth.js
```diff
@@ -70,7 +73,11 @@ socket.on("auth:start", function (serverHash) {
 			hasConfig: store.state.serverConfiguration !== null,
 		});
 	} else {
-		showSignIn();
+		if (headerAuthEnabled) {
```
itsjohncs, line 76: This can be folded up a level.

```js
} else if (headerAuthEnabled) {
    ...
} else {
    ...
}
```

client/js/socket-events/connection.js
```diff
@@ -42,6 +42,11 @@ function handleDisconnect(data) {
 	store.commit("currentUserVisibleError", `Waiting to reconnect… (${message})`);
 	updateLoadingMessage();
 
+	if (message === "xhr poll error" && store.state.serverConfiguration.headerAuthEnabled) {
+		socket.disconnect();
+		window.location.reload(true);
```
itsjohncs, line 47: Is this being done because when we try to reconnect the auth header will no longer be included, so we need to reload the page to reconnect?

client/js/socket-events/connection.js
```diff
@@ -42,6 +42,11 @@ function handleDisconnect(data) {
 	store.commit("currentUserVisibleError", `Waiting to reconnect… (${message})`);
 	updateLoadingMessage();
 
+	if (message === "xhr poll error" && store.state.serverConfiguration.headerAuthEnabled) {
```
itsjohncs, line 45: I think `store.state.serverConfiguration.headerAuthEnabled` can be `true` while `public` is `false`. Which means that this condition is not properly checking whether header auth is enabled.

src/server.js
```diff
@@ -178,7 +178,10 @@ module.exports = function (options = {}) {
 				performAuthentication.call(socket, {});
 			} else {
 				socket.on("auth:perform", performAuthentication);
-				socket.emit("auth:start", serverHash);
+				socket.emit("auth:start", {
+					serverHash: serverHash,
+					headerAuthEnabled: Helper.config.headerAuth.enable,
```
itsjohncs, line 183: I think this is another case where `public` would need to be checked as well.

I feel like we should really be erroring if `public` is set while ldap or header auth is enabled, since the latter settings will be totally ignored... then we could avoid this error-prone nonsense. But since that's a backwards incompatible change it'd probably be best to do that in a separate PR...

src/server.js
```diff
@@ -178,7 +178,10 @@ module.exports = function (options = {}) {
 				performAuthentication.call(socket, {});
 			} else {
 				socket.on("auth:perform", performAuthentication);
-				socket.emit("auth:start", serverHash);
+				socket.emit("auth:start", {
+					serverHash: serverHash,
```
itsjohncs, line 182: This can just be `serverHash` rather than `serverHash: serverHash`

src/server.js
```diff
@@ -92,9 +92,9 @@ module.exports = function (options = {}) {
 
 	let server = null;
 
-	if (Helper.config.public && (Helper.config.ldap || {}).enable) {
+	if (Helper.config.public && (Helper.config.ldap || Helper.config.headerAuth || {}).enable) {
```
itsjohncs, line 95: This isn't checking what we intend to be checking. As an example:

```js
const a = {enable: false};
const b = {enable: true};
console.log((a || b || {}).enable);
```

That will print false, even though `b` is enabled.

src/plugins/auth/header.js
```diff
@@ -0,0 +1,22 @@
...
+function headerAuth(manager, client, user, password, callback) {
+	// If no user is found, create it
+	if (!client) {
+		manager.addUser(user, null, true);
+	}
+
+	return callback(true);
+}
+
+function isHeaderAuthEnabled() {
```
itsjohncs, line 14: It seems like LDAP and Header Auth can be enabled at the same time. What happens in that case?

itsjohncs (commented):
Did a quick look through. I suspect there'll be a few rounds of back and forth here. Thank you for doing this, I hope the review process doesn't burn you out too much and that we're able to get this in!

src/plugins/auth/header.js
9p4: In that case, the application should use the header-based authentication.

client/js/socket-events/connection.js
9p4: The whole "reload/reauth on 401" system is a mess at the moment. I'll have to redo this bit and try to move everything into the service worker.

client/js/socket-events/connection.js
9p4: Yep. There's no easy cross-platform way to do this.

9p4:
Currently, everything works except for PWAs. The current issue is this: on a soft reload of the PWA (it's closed and reopened, etc), authentication fails and returns a 401 when the service worker is trying to get the resources. A hard refresh is required to fix this, since it bypasses the service worker and allows for the redirect / auth / whatever to load. It seems like service workers can't properly work with redirects, so it seems like hacky solutions are needed unfortunately.

9p4:
Also, I do not want to have the service worker be critical. In stricter / more secure FF configs, service workers aren't loaded, so automatic authentication never occurs.

9p4:
Seems like the auth/reauth bugs in the PWA are all gone now.

brunnre8:
Please leave discussions be and stop resolving them unless you actually fixed it up in code. The answer is important to every other reviewer as well and constantly having to expand to check whether you've actually changed something is tedious

src/plugins/auth/header.js
brunnre8: ehm, why?
Why do you think header based auth trumps any other?

It should not. The only sane thing to do when you have both ldap and header auth enabled is to fail as that's clearly a nonsensically configuration

brunnre8: unless you truly want to enable both, but is there a world where this really makes sesne?

client/service-worker.js
```diff
@@ -3,7 +3,7 @@
 "use strict";
 
 const cacheName = "__HASH__";
-const excludedPathsFromCache = /^(?:socket\.io|storage|uploads|cdn-cgi)\//;
+const includedPathsInCache = /^(js|css|img|themes|favicon\.ico|fonts|#)\/*/;
```
brunnre8, line 6: Why are you doing this?
Why can't you just use the same blacklist?

What are you now excluding that was working before that makes your issues go away as far as you are concerned?

src/command-line/users/index.js
```diff
@@ -1,6 +1,9 @@
 "use strict";
 
-if (!require("../../helper").config.ldap.enable) {
+if (
+	!require("../../helper").config.ldap.enable &&
+	!require("../../helper").config.headerAuth.enable
```
brunnre8, line 5: if you use it multiple times, import it once at the toplevel please, don't repeat require like this.

src/server.js
```diff
@@ -824,6 +832,11 @@ function performAuthentication(data) {
 		return;
 	}
 
+	// Prevent empty headers from creating users with blank / no username
+	if (data.user === "") {
```
brunnre8, line 836: As far as I can tell this only applies in the case where headerauth is active. Anything else simply fails the auth as it should, right?

So then put it where it belongs. A few lines above when you actually get the header value, if it's empty respond with a failure in the log and return. That's a misconfiguration most likely so it would be nice to give the admin an indication what's wrong.

brunnre8: Better yet, ignore it here, fail the auth inside of header auth as that's an invalid input value to that.
That means every backend works the same

client/service-worker.js
9p4: Some providers (such as for SAML or OpenID) have a redirect URL which can be arbitrary. This would end up being cached by the PWA, causing all kinds of issues. Therefore, using an allowlist instead of a blocklist would make sure that the arbitrary redirect URL is not cached.

brunnre8: The login redirect is to the same domain TL is hosted on?
Because if not it should not be an issue as the service worker wouldn't be in scope? Or what am I missing?

brunnre8: note to myself, example.com/thelounge is a thing

9p4: When using SAML or OpenID, there must be a redirect URL configured. This URL must be something that TL isn't using: tl.example.com/r is one example. It can be anything. When authentication occurs, the user is redirected to the provider (ie auth.example.com) which redirects the user back to the app with (tl.example.com/r?callback_data). The PWA can cache this URL which would cause issues with login. Because the callback URL doesn't have to be /r (it can be /redirect, /jfkdkdkd, or anything else), an allowlist would prevent it from being cached.

9p4: Authentication happens at the reverse proxy level: a cookie is dropped which is checked by the proxy on connection. TL has nothing to do with auth. When using a PWA, it opens an internal web browser to handle the redirect.

9p4: For HTTP auth to work, the whole of TL must be protected by it. For OpenID, there must be an endpoint that is used for redirect that is unused by the application and also protected by the authentication (so if auth is protecting /tl, then the redirect url must be at /tl/something or /tl/a/b/c)

For reference, here's what the auth looks like when using OpenID:
https://user-images.githubusercontent.com/17993169/151808953-f3966da5-c390-48f8-a86c-04d2b98add9a.MOV

9p4: The flow for this is as follows:

1. User loads TL.
2. Apache (or whatever) asks for the auth cookie
3. If the user doesn't have it, the user is told to authenticate
4. (For OpenID/SAML) The user is redirected to the provider
5. The provider authenticates the user
6. The provider redirects back to the application to an endpoint that Apache consumes but not TL
7. Apache parses the data given to it
8. Apache gives the user the cookie if all the conditions are met (user has right role, etc)
9. User loads original page, is authenticated
10. Apache looks at user cookie and passes user into to TL via header

9p4: By mod_auth_openidc (by far the most popular OpenID implementation for Apache), the redirect URL must be under the protected URL.

src/plugins/auth/header.js
9p4: https://github.com/thelounge/thelounge/pull/4447#discussion_r795267583

client/service-worker.js
```diff
@@ -38,12 +38,13 @@ self.addEventListener("fetch", function (event) {
...
 	const path = url.substring(scope.length);
 
-	// Skip ignored paths
-	if (excludedPathsFromCache.test(path)) {
+	if (!includedPathsInCache.test(path)) {
 		return;
 	}
 
-	event.respondWith(networkOrCache(event));
+	const response = networkOrCache(event);
```
9p4, line 45: Why did I split this up?

brunnre8: Because there were like a million conditionals in there :stuck_out_tongue_winking_eye: 
You've eliminated ~~all~~ most of them since then

itsjohncs:
Having trouble determining whether I should take another proper look at this, seems like there's a lot of movement. Lemme know if you'd like me to take a look.

9p4:
The more eyes I can get to look at this, the better. Thanks for all of the help so far!

itsjohncs:
Do you have any more changes planned for this PR?

9p4:
Currently, I'm just testing it out in prod to find any more bugs. There are some minor UX bugs with PWAs but other than that it works just fine. I also have to put together some documentation but that can be done once this PR is finalized.

client/js/socket-events/connection.js
```diff
@@ -39,6 +42,18 @@ function handleDisconnect(data) {
 		return;
 	}
 
+	// Sorry brunnre8, but I have to consume the error message since there appears to be no way to
+	// get the error from anywhere else
+	// We are checking if the server configuration is null because if it is null, then the client
+	// never loaded properly and there's a different issue than 401 on the header auth
+	if (
+		store.state.serverConfiguration !== null &&
+		message === "xhr poll error" &&
```
brunnre8, line 51: This is still not okay, there has to be a better way.

Run TL with the following config:

```json
// ### `transports`
	//
	// Set `socket.io` transports.
	//
	// This value is set to `["polling", "websocket"]` by default.
	transports: ["websocket"],
```

This will break the above as far as I can tell.
Every single update of socket.io may break that and we'd have nothing to complain about.
Error messages are not part of the API.

9p4: It seems like error messages are used for handing disconnect errors already: https://github.com/thelounge/thelounge/blob/master/client/js/socket-events/connection.js#L48

9p4:
I've made some more changes. Dependencies on error messages should all be gone. I'm going to test out these changes for the next few days and find and hopefully iron out any other bugs.

client/js/socket-events/connection.js
itsjohncs: I had the same thought as @brunnre8 and then the same follow-up thought as @9p4. You should make sure there's no error code on the object that we can use instead, but checking the error message is sometimes the only option 🤷‍♀️ .

MaxLeiter (changes requested):

defaults/config.js
```diff
@@ -365,6 +365,15 @@ module.exports = {
 	// This is set to `null` by default to disable `oidentd` support.
 	oidentd: null,
 
+	// Header authentication
```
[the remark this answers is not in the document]
9p4: Should be good. Let me know if it has to be changed more.

MaxLeiter (changes requested):
I left some more config comments. The reason I'm being strict about it is we use the config to generate the [documentation on the website](https://thelounge.chat/docs/configuration), hence the markdown.

test/plugins/auth/header.js
```diff
@@ -0,0 +1,64 @@
...
+		headerAuth.auth(manager, client, user, null, function (valid) {
+			expect(valid).to.equal(true);
+			done();
+		});
+	});
+}
+
+describe("Header authentication plugin", function () {
+	// Increase timeout due to unpredictable I/O on CI services
+	this.timeout(TestUtil.isRunningOnCI() ? 25000 : 5000);
```
MaxLeiter, line 29: let's not do this unless we see it being an issue?

pcolmer:
Does the authenticating user need to exist in The Lounge before this works or will the user be created "on the fly" as happens with LDAP authentication? I'm asking because I'm testing this against Auth0 with `mod_auth_openidc` and I'm hitting some issues. I'm trying to figure out which part of the stack is failing at the moment.

9p4:
It should automatically create the users.

src/server.js
```diff
@@ -773,6 +777,10 @@ function performAuthentication(data) {
 	let client;
 	let token = null;
 
+	if (!Helper.config.public && Helper.config.headerAuth.enable) {
+		data.user = socket.handshake.headers[Helper.config.headerAuth.header];
```
pcolmer, line 781: If the header isn't being set or the header name is incorrect, `data.user` will be set to `unknown` but The Lounge will get stuck because it can't work with that as a username. This can be tested/reproduced by setting `config.js` to look for a header name that doesn't exit.

Later on in the code, there is a check if `data.user` is not a string but I think that at this ^^ point, the code should behave as if the authorisation has failed, i.e. `socket.emit("auth:failed);`

pcolmer:
@9p4 Thank you for sharing your `mod_auth_openidc` setup.

Having spent a couple of days working on getting this set up with `mod_auth_openidc`, Auth0 and claims checking, can I please share the following?

```
		Require valid-user
		Require claim roles:thelounge
```

In case anyone else comes across this, the first `Require` will let *everyone* in if they authenticate successfully. The `claim` requirement never gets checked. If you only want to allow users in that have a specific role, it is necessary to leave out the `valid-user` claim.

The name for the header to be used when retrieving the user's name must be in lower-case. The documentation gives two examples that include upper-case letters. Is it worth adding a bit to the code so that it forces the configuration value to lower case?

Also, it is *very* confusing to give `REMOTE_USER` as an example as this is actually exported as an environment variable by `mod_auth_openidc` and not as a header. I have tried unsuccessfully to get your example:

```
RewriteRule .* - [E=PROXY_USER:%{LA-U:REMOTE_USER}] # note mod_rewrite's lookahead option
RequestHeader set Proxy-User %{PROXY_USER}e
```

to work. It causes a browser "Submitting" loop. I've ended up having to set `config.js` to specify a header value that is exported directly from `mod_auth_openidc`, e.g. `oidc_claim_email`.

But thank you for your contribution. Now that I've got it working (!), this will be a life-saver and I look forward to it being added to the main release.

9p4:
> In case anyone else comes across this, the first Require will let everyone in if they authenticate successfully. The claim requirement never gets checked. If you only want to allow users in that have a specific role, it is necessary to leave out the valid-user claim.

It's wrapped in a `RequireAll` so it will check all of the entries in it. I've tested it out (removing user from `thelounge` role)

I've also updated the PR and updated to master.

src/server.js
9p4: hopefully I've added enough checks

pcolmer:
@9p4 
> It's wrapped in a RequireAll so it will check all of the entries in it. I've tested it out (removing user from thelounge role).

Thanks! I'd missed that. Something else I've learnt today 😀

pcolmer:
I have a question about how this works: if `mod_auth_openidc` is configured to require something like a specific claim, and the user gains access to The Lounge, what happens if that "something specific" is removed from the user? So, if the user's access is removed from (say) their account on Auth0, is there a way to log them out/remove them from The Lounge?

I *think* that the auth token should be getting refreshed/renewed periodically so the claims *should* also get updated but I'm not sure if that would trickle down this stack to the point where they get kicked off The Lounge.

9p4:
It should invalidate on refresh of the page. Until then the authentication remains.

9p4:
Because this uses header authentication, there's no way to have a platform-agonistic endpoint in TL without making the scope of this project much larger (supporting OpenID natively, etc)

MaxLeiter:
@9p4 is this ready for review or are you still working on it? The PR mentions tests aren't finished yet, is that still true?

9p4:
The basic tests are complete, but I'm trying to make this work for just the polling transport mode (for which the PWA does not behave as expected)

9p4:
After a long while, ready for review!

9p4:
The issues regarding the PWA were caused by my SSO server and session expiry.

9p4:
Any updates?

MaxLeiter:
@9p4 Do you mind updating your PR to typescript? It should be fairly straight-forward and I'm happy to assist on IRC if you need it. Otherwise I can do the update for you, but that will probably take me a bit longer.

9p4:
Unfortunately, I won't have access to a computer for the next couple of weeks. I can only do small drive-by contributions at this point from a flip phone over 3G haha

If you're willing to wait, I can get this done mid-July.

9p4:
I have begun the conversion to TS

9p4:
Alright, it should all be updated to TS.

sinavir:
Hello, I was wondering if this PR will be merged one day. I'm really interested in SSO features like that one and merging this would be very nice I think !

platta:
I'm currently running this PR on top of 4.3.1 because it fixes a re-auth issue with my simple Authelia configuration (I'm not doing any auth passthrough).

If there's anything I can do to assist with testing, I'd be happy to lend a hand.

MaxLeiter:
@sinavir yes, it will be merged! The dev team is just busy and has lives. If you want to help, you can be like @platta (thanks platta!) and report back your experiences with it.

platta:
I've got TL behind NGINX as a reverse proxy, and my configuration uses Authelia to give me 2FA. So, I have to log into Authelia, enter the 2FA code, and then still log into TL.

In the past, when my Authelia auth timed out...

- **In OS X Safari** - Refreshing or reloading the page would just show the TL logo and this error message: `Service worker failed to fetch an url: Request failed with HTTP 0`. If I open the dev console, I could see error messages about refusing to connect to my Authelia URL due to access control checks. The errors said the Authelia URL "does not appear in the connect-src directive of the Content Security Policy." I could open another browser tab and log back in, then refresh and all would be fine (no need to log back into TL). I never tried quitting Safari and re-launching, so I'm not sure what effect, if any, that would have had.
- **As an iOS PWA** - Quitting and re-launching the "app" wouldn't change anything, and since the PWA lives in its own sandbox there was no way to log in "on another tab" to fix it. I would have to go into Safari and add a new Home Screen shortcut every time my auth timed out. From there, log into Authelia, and log back into TL.

Now that I'm running on this PR, and I've set my timeout on Authelia to 5 minutes, so I constantly have to re-auth. What I've observed now...
- **In OS X Safari** - Same behavior as before, however if I quit and re-launch Safari I am properly redirected to Authelia's login page, and then back into TL (not asked to log back into TL).
- **As an iOS PWA** - Same behavior as before, however if I quit and re-launch the "app" I am properly redirected to Authelia's login page, and then back into TL (not asked to log back into TL).

For my simple configuration, this solves the issues I had. Like @9p4 mentioned in an earlier comment, having to quit and re-launch isn't ideal but that can easily be explained in documentation.

If there is any more digging/testing/reporting I can do, just ask. I haven't worked in Node in ages, and I don't know Vue, but I'll do what I can!

9p4:
I'm so glad others are testing this out! Keep the feedback coming please!

9p4:
> In OS X Safari - Same behavior as before, however if I quit and re-launch Safari I am properly redirected to Authelia's login page, and then back into TL (not asked to log back into TL).

Refreshing the tab ought to fix it too, unless if something has gone very wrong in my code

9p4:
Try doing a hard reload (command + shift + r)

xadammr:
This works perfectly, 99% of the time. Sometimes on a reload I will get a "Safari Can't Open the Page" error: 

Screen Shot 2022-08-22 at 20 35 33

To resolve, I need to remove any fragments (eg. /#/settings or /#/channel-12) and load the root URL. I am then logged back in. 

It is intermittent, and I can't always reproduce. 

I am using Apache as my reverse proxy.

pinpox:
Any updates on this? Would be great to delegate TL account management to a central SSO service (currently using [dex](dexidp.io/))

9p4:
It's nearly been a year since this PR was opened. Since then, I no longer believe that HTTP header authentication is the most effective nor compatible way to extend the authentication capabilities of TL. TL's architecture is fundamentally incompatible with using HTTP header authentication to sign in. In the coming month, I will be working on a new PR to add OpenID support in this PR's stead.

This code works just fine—but it's just that. I don't believe that the code in this PR is good enough to be added to this project. Others are free, however, to use this fork if they wish to. I would advise against it since it is not maintained.

pinpox:
OpenID support  seems like a good call. If/when there is a PR ongoing, let us know where to track progress.

PR #4449, opened 2022-01-26
Remove node 15.x from build matrix

brunnre8:
EOL was June 2021, this time has passed long ago.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,10 +17,6 @@ jobs:
           - os: ubuntu-latest
             node_version: 14.x
 
-          # EOL: June 2021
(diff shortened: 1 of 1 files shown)


brunnre8:
This prevents me from updating one of our outdated deps and I'd rather not hold it back if the only reason is an outdated node version.

We don't support the ubuntu / debian shipped node version anyhow (not even supported by node) so it doesn't really matter if people have to update to a LTS rather than a short lived version in between.

itsjohncs (approved):
This seems fine, especially since 14 is still supported. No complaints here.

MaxLeiter (approved):

PR #4451, opened 2022-01-30
Replace deprecated Renovate config options.

itsjohncs:
This was an automatically requested migration that the
`renovate-config-validator` asked for. It's hard to tell because the
config options it asked to be removed are no longer in their docs, but
I believe this will not change the behavior of the renovate bot at all.


MaxLeiter (approved):

brunnre8:
merged via the combined PR

PR #4453, opened 2022-01-31
Bump most deps

brunnre8:
This bumps most deps that we can, except
* anything maintained by sindresorhus (file-type, filenamify, got, read-chunk, chalk) due to them [forcing ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c)
* emoji-regex: can't update as it breaks our tests (https://github.com/mathiasbynens/emoji-regex/issues/61)

The new auto formatter and linters pulled in new rules, I've disabled some because they were very noisy. Please have a look if you agree with the choice.


itsjohncs:
Have you been running on this changeset for a bit? I feel like as long as it runs without issue for a bit, it's probably fine to merge.

brunnre8:
Considering that I literally just commited a few seconds ago probably not :stuck_out_tongue_winking_eye:

brunnre8:
I just wanted the CI to trigger, will dogfood that as well but would also appreciate some testing on your side

itsjohncs:
Haha fair enough, I'll pull this into my dogfood branch at some point. I wanna finish poking at the muting feature before I do though and not sure when I'll do that.

client/css/style.css
```diff
@@ -17,7 +17,7 @@
 	--button-text-color-hover: #fff;
 
 	/* Color for sidebar overlay and other things that dim the viewport when something else is on top */
-	--overlay-bg-color: rgba(0, 0, 0, 0.5);
+	--overlay-bg-color: rgb(0 0 0 / 50%);
```
MaxLeiter, line 20: I prefer rgba but 🤷‍♂️

client/css/style.css
```diff
@@ -2083,7 +2084,7 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 
 .window#changelog h3 {
 	font-size: 20px;
-	border-bottom: 1px solid currentColor;
+	border-bottom: 1px solid currentcolor;
```
MaxLeiter, line 2087: The spec defines currentColor https://www.w3.org/TR/css-color-3/#currentcolor

MaxLeiter (commented):
Left two nit pick comments but the rest looks good. Haven't actually tested it yet. Thanks for tackling this!

client/css/style.css
itsjohncs: I've started using the `r g b / a` syntax myself since the `stylelint` project suggests doing so. I figure if a group who's taken the time to write a CSS linter thinks its the best syntax to use, I can probably just go with it.

client/css/style.css
brunnre8: That's done by the auto fix option, I didn't do so manually

brunnre8: https://github.com/stylelint/stylelint/issues/5863

brunnre8: spec version 4 disagrees https://www.w3.org/TR/css-color-4/#currentcolor-color

client/css/style.css
brunnre8: I try to not fight the linter if I don't have to but don't have any opinion in this case.
If any of you feel strongly say so and I'll fix up the linter to do what you want.

brunnre8:
sorry for the noise... git, myself, yarn and npx browserlist all had different opinions on how the merge resolution should look like.

I'm done now, please take a look

MaxLeiter (approved):

Issue #3921, opened 2020-05-21
Emit a message for SASL loggedin/loggedout events


xPaw:
irc-framework emits them, but we do not handle them so it's "lost".

https://github.com/kiwiirc/irc-framework/blob/2321afaa4ef7a2a670738cc8eb79092a3e4d7e1e/src/commands/handlers/registration.js#L289-L296

https://github.com/kiwiirc/irc-framework/blob/2321afaa4ef7a2a670738cc8eb79092a3e4d7e1e/src/commands/handlers/registration.js#L314-L321

RobustProgram:
Hi, I would like to give this a go

dipayanDebTheCoder:
Is anyone working on this. I am new to open source so I am looking for some easy coding job to boost my confidence and get started in contributing to open source.  If no one is working on this in that case I can give it a try

xPaw:
Pull requests welcome.

PR #4466, opened 2022-02-08
Emit a message for SASL loggedin/loggedout events

progval:
Closes GH-3921

I didn't write any test as I don't see any for existing modules in `src/plugins/irc-events/`


src/plugins/irc-events/sasl.js
```diff
@@ -0,0 +1,27 @@
...
+	const client = this;
+
+	irc.on("loggedin", (data) => {
+		const lobby = network.channels[0];
+
+		const msg = new Msg({
+			type: Msg.Type.AUTH,
+			text: "Logged in as: " + data.account,
+		});
+		lobby.pushMessage(client, msg, false);
```
MaxLeiter, line 15: Code looks good, but I'm wondering if we want this to be `true` (for those reading here, `increasesUnread`)

progval: I set this to `false` because I don't think it's worth reading, but I also think that about MOTD and the like (which is `true`), so it's your call

MaxLeiter: Yeah, for explicitness's sake we should increase the unread to decrease potential confusion. New message in window === increase the unread, generally. For example, I want to know if I'm logged out by the network. We already spam the user with messages on connection, so one more about logging in is nbd imo

progval: done

MaxLeiter (approved):

progval:
@brunnre8 any suggestion about how to style it?

brunnre8:
You can either be cute and find some font awesome icon that fits this specific message or simply add the type to the list of things that are already there and use the green info bubble or some such

https://github.com/thelounge/thelounge/blob/3fb18717a7a839af75599e40f257df5fc8c8e1a2/client/css/style.css#L428

Here's the CSS for all the other such message types

MaxLeiter (approved):

brunnre8:
This is merged via 1e3a7b12500d8898500eaf54c01e52f8d5a0b3fd

GH apparently has issues recognizing manually squashed PRs

PR #4467, opened 2022-02-08
README: suggest running 'yarn format:prettier' when linting fails

progval:
When running 'yarn test', prettier just lists different files, but does not
suggest what changes to make.


MaxLeiter (approved):

MaxLeiter:
Thanks!

Issue #521, opened 2016-07-20
Reload config on SIGHUP


nirvdrum:
This is more of a wishlist item than anything else. I wanted to change the stylesheet for the client and was curious to see if there were a way to reload the config.js file without restarting the process. I sent a `SIGHUP` to the process and it was summarily killed. I'm not sure that there's a "right way" to handle it (node seems to have its own defaults), but many daemonized applications gracefully handle the signal.

fuzzy76:
I just missed this for reloading renewed certificates as well.

finalhit:
This is especially important for Let's Encrypt certificates expiring every 90 days, need to reload the config on renewal hook.

paride:
I think a `thelounge reload` command would also work.

bobey6:
I'm a bit surprised that with so many people using Let's Encrypt certificates that it's still not possible to reload/rehash the config for new certificates without restarting the entire application.

PR #4468, opened 2022-02-09
Add partial support for reloading configuration

progval:
Partially implements GH-521

Supported:

* reloading default network config
* uploader config
* triggering a reload of all clients on change
* probably LDAP? I didn't test

Not supported yet:

* reloading TLS config and certificates
* public/private mode (probably not worth supporting anyway)
* staggered client restart to avoid overloading the server
  (is this an issue? it's still better than just restarting the server
  anyway)

### DIFF
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -33,7 +33,7 @@ socket.once("configuration", function (data) {
 	}
 });
 
-socket.on("configuration:reload", function () {
+socket.on("reload", function () {
 	socket.disconnect();
 	location.reload(true);
 });
--- a/client/js/socket-events/configuration.js
+++ b/client/js/socket-events/configuration.js
@@ -32,3 +32,8 @@ socket.once("configuration", function (data) {
 		});
 	}
 });
+
+socket.on("configuration:reload", function () {
+	socket.disconnect();
+	location.reload(true);
+});
--- a/src/helper.js
+++ b/src/helper.js
@@ -22,6 +22,7 @@ let clientCertificatesPath;
 
 const Helper = {
 	config: null,
+	reloadConfig,
 	expandHome,
 	getHomePath,
 	getPackagesPath,
@@ -52,9 +53,22 @@ const Helper = {
 	},
 };
 
-module.exports = Helper;
+function requireConfig(configPath_) {
+	// Clear the cached config
+	delete require.cache[configPath_];
 
-Helper.config = require(path.resolve(path.join(__dirname, "..", "defaults", "config.js")));
+	return require(configPath_);
+}
+
+function systemConfig() {
+	configPath = path.resolve(path.join(__dirname, "..", "defaults", "config.js"));
+
+	return requireConfig(configPath);
+}
+
+Helper.config = systemConfig();
+
+module.exports = Helper;
 
 function getVersion() {
 	const gitCommit = getGitCommit();
@@ -101,6 +115,10 @@ function getVersionCacheBust() {
 }
 
 function setHome(newPath) {
+	loadAndMergeConfig(this.config, newPath);
+}
+
+function loadAndMergeConfig(config, newPath) {
 	homePath = expandHome(newPath);
 	configPath = path.join(homePath, "config.js");
 	usersPath = path.join(homePath, "users");
@@ -112,7 +130,7 @@ function setHome(newPath) {
 
 	// Reload config from new home location
 	if (fs.existsSync(configPath)) {
-		const userConfig = require(configPath);
+		const userConfig = requireConfig(configPath);
 
 		if (_.isEmpty(userConfig)) {
 			log.warn(
@@ -128,14 +146,14 @@ function setHome(newPath) {
 			log.warn("Using default configuration...");
 		}
 
-		mergeConfig(this.config, userConfig);
+		mergeConfig(config, userConfig);
 	}
 
-	if (this.config.fileUpload.baseUrl) {
+	if (config.fileUpload.baseUrl) {
 		try {
-			new URL("test/file.png", this.config.fileUpload.baseUrl);
+			new URL("test/file.png", config.fileUpload.baseUrl);
 		} catch (e) {
-			this.config.fileUpload.baseUrl = null;
+			config.fileUpload.baseUrl = null;
 
 			log.warn(`The ${colors.bold("fileUpload.baseUrl")} you specified is invalid: ${e}`);
 		}
@@ -157,7 +175,7 @@ function setHome(newPath) {
 
 	// Load theme color from the web manifest
 	const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf8"));
-	this.config.themeColor = manifest.theme_color;
+	config.themeColor = manifest.theme_color;
 
 	// log dir probably shouldn't be world accessible.
 	// Create it with the desired permission bits if it doesn't exist yet.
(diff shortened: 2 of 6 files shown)


MaxLeiter:
Thanks for the PR, @progval!
 
> probably LDAP? I didn't test
We have some users who generally test/try out LDAP-related stuff for us if you want me to ping them

> public/private mode (probably not worth supporting anyway)
Not really ideal but I think I agree, although we need to make this clear in the docs

>staggered client restart to avoid overloading the server
(is this an issue? it's still better than just restarting the server
anyway)
This can already be an issue on servers with hundreds of users so it should definitely be implemented in some way.

client/js/socket-events/configuration.js
```diff
@@ -32,3 +32,8 @@ socket.once("configuration", function (data) {
 		});
 	}
 });
+
+socket.on("configuration:reload", function () {
```
MaxLeiter, line 36: I'm not sure if this should be scoped to `configuration` just because it's only used for it, as the actual content isn't at all config related. Personal preference atm, I guess. Seeing `configuration:reload` in ClientManager feels a little awkward.

src/helper.js
```diff
@@ -52,9 +53,22 @@ const Helper = {
 	},
 };
 
-module.exports = Helper;
+function requireConfig(configPath_) {
```
MaxLeiter, line 56: why `configPath_`? Can we not just use `configPath` again? If it's to signal that it's private I'm used to seeing it prefixed with _, like `_configPath`, and that's not generally consistent with our codebase for function params

src/helper.js
```diff
@@ -188,6 +206,25 @@ function setHome(newPath) {
 	}
 }
 
+// Returns whether there was any change to the configuration.
```
MaxLeiter, line 209: Can we split this up into `reloadConfig` and (as an example) `doesConfigHaveChanges`? I don't see why reloadConfig() should return a boolean of having changes.

src/plugins/uploader.js
```diff
@@ -77,6 +77,10 @@ class Uploader {
 	}
 
 	static async routeGetFile(req, res) {
+		if (!Helper.config.fileUpload.enable) {
+			return res.status(404).send("Not found (file upload is disabled");
```
MaxLeiter, line 81: `...file uploading...`

src/plugins/uploader.js
```diff
@@ -127,6 +131,10 @@ class Uploader {
 	}
 
 	static routeUploadFile(req, res) {
+		if (!Helper.config.fileUpload.enable) {
+			return res.status(404).send("Not found (file upload is disabled");
```
MaxLeiter, line 135: `...file uploading...`

MaxLeiter: maybe we want an isFileUploading middleware?

src/server.js
```diff
@@ -233,6 +231,14 @@ module.exports = function (options = {}) {
 
 		process.on("SIGINT", exitGracefully);
 		process.on("SIGTERM", exitGracefully);
+		process.on("SIGHUP", function () {
+			const hasChanges = Helper.reloadConfig();
```
MaxLeiter, line 235: My above comments would transform this to `const hasChanges = Helper.doesConfigHaveChangs()` which would call `reloadConfig()`

MaxLeiter (changes requested):

src/helper.js
progval: the linter complains about both `configPath` and `path` because they shadow global variables, and I couldn't think of a better name

src/helper.js
progval: `doesConfigHaveChanges` would also have to `require()` the config, which feels wasteful

src/helper.js
MaxLeiter: `filePath`?

src/helper.js
MaxLeiter: the require would be in reloadConfig, doesConfigHaveChanges would call that, right? Just this way the diffing can be left for if it's needed.

progval: Are you saying `doesConfigHaveChanges()` should actually perform the reload? This is very unclear from the name.

MaxLeiter: 👍

MaxLeiter (approved):

MaxLeiter:
> staggered client restart to avoid overloading the server 

This would be nice but we don't have it at initial startup so after some thought I don't think it should block this. In the future it's something we should implement, though.

brunnre8:
Just to augment val's list:

Not supported yet (in addition):
* socket.io transports
  (Or well, you end up having the client and server disagree, which might be bad... say flip from websocket only to polling only and suddenly the client can't possibly connect as it gets the new one via the force refresh of the index page, but the server doesn't actually change it's listening params)
* host / port / bind (neither identd, irc-framework nor TL)
* Message storage
  The code never re-creates connected clients as far as I can tell, meaning the message storage is now out of sync with the config.
* UseHexIP / webirc or similar irc-framework options

src/plugins/uploader.js
```diff
@@ -72,6 +72,13 @@ class Uploader {
 	}
 
 	static router(express) {
+		express.use(function (req, res, next) {
```
brunnre8, line 75: This is bad, that registers a **global** handler for every single route path.
Meaning whatever route you register after that call will always fail if file uploads are disabled.

You have to scope this to the /uploads path

brunnre8 (changes requested):
Thanks for working on this Val, much appreciated.

It does have some logic bugs and some things which seemed strange to me while reading the code, the latter being personal preference so feel free to push back if you disagree there.

brunnre8:
Looking at the complexity involved with only some of the settings reloaded, do we really want to partially Merge this though @MaxLeiter ?
From a user perspective we should either load everything anew or fail the reload if options changed that we can't actually guarantee to have a sane effect.

src/helper.js
```diff
@@ -188,6 +206,25 @@ function setHome(newPath) {
...
+function reloadConfigIfChanged() {
```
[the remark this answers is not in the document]
progval: I agree with you.

src/plugins/uploader.js
```diff
@@ -72,6 +72,13 @@ class Uploader {
...
+			if (Helper.config.fileUpload.enable) {
+				next();
+			} else {
+				res.status(404).send("Not found (file uploading is disabled");
```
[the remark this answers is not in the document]
progval: So what should I send as error message?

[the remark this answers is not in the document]
brunnre8: I'd just send the usual 404, but that's me.
The additional information is nice for the admin if they misconfigured TL, but for most people who aren't actually the admin this doesn't make sense I think.

MaxLeiter:
@brunnre8 It's an improvement, so I'm comfortable half-merging it as long as the ramifications and shortcomings are clear. But that's just my opinion, it might make more sense to hold-off. I'm happy to ship 4.3.1 without it.

progval:
Where should I document what is reloadable and what isn't?

1. Only in the website (ie. in https://github.com/thelounge/thelounge.github.io/blob/master/_docs/configuration.md)
2. At the beginning or the end of config.js
3. On each setting's comment section?

brunnre8:
I personally favor 3. as this is the least likely to drift out of date... but sadly also the most effort probably.
Unless someone else chips in, pick your favorite, I don't have a strong opinon.

MaxLeiter:
1. is impossible, as it's generated from the docs. I recommend going with 3.

progval:
Done. I didn't check most of the values (only the ones mentioned when opening the PR) because there are many, but the code seemed straightforward enough. I can double-check if there is any you are not confident about.

MaxLeiter (approved):

Issue #4429, opened 2021-12-06
File Uploader: MIME type issue


xnaas:
- _Node version:_ [official alpine docker container](https://github.com/thelounge/thelounge-docker/blob/31785401f6e7fce0b70ae28c60377d425c4b100b/alpine/Dockerfile#L1)
- _Browser version:_ Firefox 95
- _Device, operating system:_ Windows 10 64-bit
- _The Lounge version:_ 4.3.0

---

When uploading an ogg file that container opus audio, the MIME type doesn't appear to get set correctly. If the .ogg contains vorbis audio, the mime type appears to be set correctly and the audio embeds.

This ZIP file contains an example of each to test uploading to your own The Lounge: [Toss A Coin To Your Witcher (2020).zip](https://github.com/thelounge/thelounge/files/7663148/Toss.A.Coin.To.Your.Witcher.2020.zip) (credit to [Dan Vasc](https://www.youtube.com/watch?v=bS4Q-WWyl3Q) for this sick cover, btw) If you don't want to download the ZIP, these files will be available to test with for a short time: [OPUS .ogg](https://irc.xnaas.info/uploads/14b35251dfe80dfc/%5BOPUS%5D%20Toss%20A%20Coin%20To%20Your%20Witcher%20(Metal%20Cover).ogg), [VORBIS .ogg](https://irc.xnaas.info/uploads/509cb1f0d5dac516/%5BVORBIS%5D%20Toss%20A%20Coin%20To%20Your%20Witcher%20(Metal%20Cover).ogg).

If you upload the `[OPUS]` file, you will not get an embed in The Lounge. If you load the file directly, you'll notice that the mime type being set is `audio/opus`, which doesn't appear to be supported by any of the major browsers.

If you upload the `[VORBIS]` file, you will get an embed in The Lounge. If you load the file directly, you'll notice that the mime type being set is `audio/ogg`, which is supported by at least Chromium and Firefox (Safari is special).

xnaas:
Tangentially related...but had some more fun testing with FLAC (not uploads, just links; see also: #4210):

* If the server is configured to set FLAC files to `audio/x-flac`, The Lounge won't embed it.
* If the server is configured to set FLAC files to `audio/flac`, The Lounge will embed it.

[...]

PR #4470, opened 2022-02-11
uploader.js: upload m4a as audio/mp4 | link.js: embed audio/mp4, x-flac, and x-m4a

xnaas:
Trying to follow the MDN, it says that `.m4a` should be set to `audio/mp4`.

This PR makes it so that an `.m4a` file uploaded through The Lounge has the mime type `audio/mp4` set. This also ensures that files being served with `audio/mp4` are embedded by The Lounge.

The only sample `.m4a` files I could find to test with that wasn't something I already have or uploaded were:

* https://dl.espressif.com/dl/audio/ff-16b-2c-44100hz.m4a
* https://dl.espressif.com/dl/audio/ff-16b-1c-44100hz.m4a

but these set `audio/mpeg`, which TL already handles. 😅 It would seem that `.m4a`s are quite rare to just be openly hosted around.

Edit: I found https://filesamples.com/samples/audio/m4a/Symphony%20No.6%20(1st%20movement).m4a which seems to serve as `audio/x-m4a`.

---

Source: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Containers#browser_compatibility

Relevant previous PR: #4210.

### DIFF
--- a/src/plugins/uploader.js
+++ b/src/plugins/uploader.js
@@ -117,6 +117,8 @@ class Uploader {
 			detectedMimeType = "audio/wav";
 		} else if (detectedMimeType === "audio/x-flac") {
 			detectedMimeType = "audio/flac";
+		} else if (detectedMimeType === "audio/x-m4a") {
+			detectedMimeType = "audio/mp4";
 		}
 
 		res.setHeader("Content-Disposition", disposition);
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -251,7 +251,7 @@ function parse(msg, chan, preview, res, client) {
 		case "audio/x-mpeg":
 		case "audio/x-mpeg-3":
 		case "audio/flac":
-		case "audio/x-m4a":
+		case "audio/mp4":
 			if (!preview.link.startsWith("https://")) {
 				break;
 			}
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -251,7 +251,9 @@ function parse(msg, chan, preview, res, client) {
 		case "audio/x-mpeg":
 		case "audio/x-mpeg-3":
 		case "audio/flac":
+		case "audio/x-flac":
 		case "audio/mp4":
+		case "audio/x-m4a":
 			if (!preview.link.startsWith("https://")) {
 				break;
 			}


src/plugins/uploader.js
```diff
@@ -115,8 +115,6 @@ class Uploader {
 		// so that browsers can play them correctly
 		if (detectedMimeType === "audio/vnd.wave") {
 			detectedMimeType = "audio/wav";
-		} else if (detectedMimeType === "audio/x-flac") {
```
itsjohncs, line 118: This is done because the `file-type` module returns `audio/x-flac` rather than `audio/flac`. This sets it to `audio/flac` instead, which I think is generally what you're trying to do.

Removing this seems like it's counterproductive. Unless you noticed that the newer versions of `file-type` return `audio/flac` now?

src/plugins/uploader.js
```diff
@@ -20,8 +20,8 @@ const inlineContentDispositionTypes = {
 	"audio/mpeg": "audio.mp3",
 	"audio/ogg": "audio.ogg",
 	"audio/vnd.wave": "audio.wav",
-	"audio/x-flac": "audio.flac",
-	"audio/x-m4a": "audio.m4a",
```
itsjohncs, line 24: These are compared to the return values of `file-path`, so this should break file-type detection unless `file-path` has changed what mime types it returns for these file types?

itsjohncs (changes requested):
I only took a close look at the `uploader.js` changes, but I'm not sure I understand what's going on. If these changes are working as you expect, could you share your test procedures? (ie: steps to follow, expected result before change, expected result after change)

src/plugins/uploader.js
xnaas: I wonder if this `file-type` plugin is the one responsible for #4429? 🤔

xnaas:
I apologize for not mentioning this in the original PR message.

The goal of this PR is for The Lounge to embed both uploaded `.m4a` files (which get `audio/x-m4a`) as well as `.m4a` files that are served with `audio/mp4` from webservers.

The current situation is that The Lounge will only embed a `.m4a` that you uploaded directly to it or if a webserver is configured to add the `audio/x-m4a` mime type to a `.m4a` file instead of `audio/mp4`.

xnaas:
Without this PR: bad

With this PR: good

xnaas:
I built [this container image](https://github.com/xnaas/docker-thelounge/pkgs/container/thelounge/14838682) off of https://github.com/thelounge/thelounge/pull/4470/commits/daaf0c3e9a843608ef4ac03a6cad00bcff099544 to make testing easy for anyone already versed in Docker.

xnaas:
Also, thank you for the feedback bookworm. :)

itsjohncs (approved):
Looks good to me. Didn't test it myself but took a close look at the code and feel reasonably confident this'll do what's advertised. Thank you! I'll merge this tomorrow if no other maintainers chime in (and if I don't forget 😅).

MaxLeiter (approved):

MaxLeiter:
Thanks @xnaas!

PR #4472, opened 2022-02-12
Allow configuring multiple default networks

### LINKED ISSUE #4299: Make Default host plural
I suggest making the default host configuration value plural and change its type to an array of default hosts. Then inside the front end allow a drop down box for selection of the default hosts array.

This will add more customization ability for those who wish to allow users to connect to one or more networks that are configured.

progval:
Closes GH-4299

## When `lockNetwork` is false

This adds a datalist to the network name. Picking a name equal to one of the items of the datalist auto-fills all the fields in the form with whatever was configured for that network. (This overwrites what the user may have written, but I don't think this is an issue, as the network name is the first field in the form.)

The default network name (and details) are the first network in the list (the current behavior).

There are probably improvements to be made here, as browsers make `<input>` datalists really not obvious to users (they look like regular autocompletes), but this is the simplest way to implement it.

<details>
<summary>Screenshots for `lockNetwork` being false</summary>

private_unlocknetwork_multi

private_unlocknetwork_multi2

(these screenshots are in private mode, but they look the same in public mode)
</details>

## When `lockNetwork` is true and there is a single network configured

this changes nothing

<details>
<summary>Screenshots for `lockNetwork` being true with a single network configured</summary>

public:

public_locknetwork_single

private:

private_locknetwork_single
</details>

## When `lockNetwork` is true and there is more than one network configured

this adds a `<select>` box, which defaults to the first network configured

<details>
<summary>Screenshots for `lockNetwork` being true with two networks configured</summary>

public:

public_locknetwork_multi

private:

private_locknetwork_multi

</details>


src/helper.js
```diff
@@ -271,12 +272,24 @@ function passwordCompare(password, expected) {
...
+function getDefaultNetworks() {
+	if (this.config.defaults === undefined) {
+		return [];
+	}
+
+	if (Array.isArray(this.config.defaults)) {
+		return this.config.defaults;
+	}
+
+	return [this.config.defaults];
```
MaxLeiter, line 292: Can we also add a log warning here that the object format is deprecated?

Something like https://github.com/thelounge/thelounge/pull/4472/files#diff-85b18f65b01cea7762ab044d7df73c7d3e6de0c9e3502cce8540cad8f2534039R283

MaxLeiter (changes requested):

src/helper.js
progval: Oh yes, of course. I don't see what your link is supposed to point to, so I improvised. The output format looks like this:

```
2022-02-14 18:32:26 [WARN] Key "defaults" should now be an array of networks, please update your config. It should look like this:
defaults: [
	{
		"name": "Libera.Chat",
		"host": "irc.Libera.Chat",
		"port": 6697,
		"password": "",
		"tls": true,
		"rejectUnauthorized": true,
		"nick": "thelounge%%",
		"username": "thelounge",
		"realname": "The Lounge User",
		"join": "#progval",
		"leaveMessage": ""
	}
],
```

I can change it, of course

MaxLeiter: Ah, I used the wrong link. This message works, but I'd include "support for the old object format will be removed in a future version". Not sure we need to print the defaults, maybe a link to the docs?

progval: Done. I pluralized the anchor in the link, as I expect the title to be made plural when releasing the next version.

MaxLeiter: Yep, the docs are auto-generated from the config file comments so you're change here will change the anchor:
https://github.com/thelounge/thelounge/pull/4472/files#diff-2a8a9985100687b211e3cf1123b40796d8091ac34db900f6683a81ef06046219R207

brunnre8:
Can we please always make this a select box?
The thing that looks like an autocomplete input is something I personally really dislike as a user and I can't be the only one.

We can just call them "presets" when the network isn't locked and rename it to "Network" if the mode is locked.
("Name" isn't terribly helpful I think and one can probably get rid of a separate section for it considering that it's only a single field)

progval:
Agreed, thanks!

Here is how it looks when `lockNetwork` is false, now:

screenshot-2022-03-05_12-09-42

MaxLeiter (changes requested):

src/models/network.js
```diff
@@ -138,27 +138,35 @@ Network.prototype.validate = function (client) {
 	}
 
 	if (Helper.config.lockNetwork) {
+		// Get the first configured network that matches this one, if any.
+		let defaultNetwork = Helper.getDefaultNetworks().filter(
```
MaxLeiter, line 142: Can we store `getDefaultNetworks` so we don't print the `[WARN] Key "defaults" should now be` multiple times if they're using an old config?

Testing showed me

```sh
2022-04-20 00:32:40 [WARN] Key "defaults" should now be an array of networks. Support for the old object format will be removed in a future version, please update your config. https://thelounge.chat/docs/configuration#default-networks
2022-04-20 00:32:46 [WARN] Key "defaults" should now be an array of networks. Support for the old object format will be removed in a future version, please update your config. https://thelounge.chat/docs/configuration#default-networks
2022-04-20 00:32:48 [WARN] Key "defaults" should now be an array of networks. Support for the old object format will be removed in a future version, please update your config. https://thelounge.chat/docs/configuration#default-networks
2022-04-20 00:32:52 [WARN] Key "defaults" should now be an array of networks. Support for the old object format will be removed in a future version, please update your config. https://thelounge.chat/docs/configuration#default-networks
```

src/helper.js
```diff
@@ -271,12 +272,28 @@ function passwordCompare(password, expected) {
...
+		return [];
+	}
+
+	if (Array.isArray(this.config.defaults)) {
+		return this.config.defaults;
+	}
+
+	const key = colors.bold("defaults");
+	log.warn(
+		`Key "${key}" should now be an array of networks. Support for the old object format will be removed in a future version, please update your config. https://thelounge.chat/docs/configuration#default-networks`
```
MaxLeiter, line 294: `Configuration key...`? I think Key is too ambiguous

MaxLeiter (changes requested):
If defaults isnt an array, the preset option isn't selected even though it should be:

Image showing the network template dropdown without a default value selected

progval:
I can't find a way to select the first option. I tried this:

```diff
diff --git a/client/components/NetworkForm.vue b/client/components/NetworkForm.vue
index 0308b0be..91e8acfa 100644
--- a/client/components/NetworkForm.vue
+++ b/client/components/NetworkForm.vue
@@ -23,9 +23,10 @@
 						<label for="connect:presetName">Preset</label>
 						<select id="connect:presetName" v-model="presetName" name="presetName">
 							<option
-								v-for="defaultNetwork in config.defaults"
+								v-for="(defaultNetwork, i) in config.defaults"
 								:key="defaultNetwork.name"
 								:value="defaultNetwork.name"
+								:selected="i === 0"
 							>
 								{{ defaultNetwork.name }}
 							</option>
```

and a few variations, but they have no effect. Any idea?

MaxLeiter:
@progval I believe you need to update `presetName`, as that's referenced in the `<select>` `v-model`. `:selected` on the option doesn't do anything, the state lives in the select.

progval:
Now I don't understand why it works for other fields...

Anyway, done. Do you want to deal with the merge conflicts?

progval:
ping?

brunnre8:
https://github.com/progval/thelounge/pull/1 / https://github.com/thelounge/thelounge/tree/bookworm/multiDefaultNets <-- my version of a merge resolution, GH doesn't seem to pick it up automatically

Issue #4369, opened 2021-11-21
minor issues with copy/paste


xnaas:
- _Node version:_ 12.22.7
- _Browser version:_ Firefox 95.0b9 (64-bit)
- _Device, operating system:_ Windows 10
- _The Lounge version:_ The Lounge source (dd266c60 / v4.3.0-rc.2) (Node.js 12.22.7 on linux x64)

---

## Missing Preceding `<`

If you copy and paste a line from only the user's nickname, without timestamp, you do not get a preceding `<`.

### Example:

Test Line:
![image](https://user-images.githubusercontent.com/8271327/142747676-3b9aef1d-9e4e-48e0-8f4a-f53dce9ca3aa.png)

Result:
```
+MaxLeiter> 32 contributors in this release. very cool
```

Desired Output:
```
<+MaxLeiter> 32 contributors in this release. very cool
```

## Random Extra Space

If you're copying multiple lines, *including timestamps*, the first line will randomly have a space before it.

Test Lines:
![image](https://user-images.githubusercontent.com/8271327/142747733-d594c458-47d9-47f4-968f-cc361b1a69a5.png)

Result:
```
 20:53:32 <xnaas> of course, just a general note
20:53:40 <xnaas> testing the behavior reminded me to complain
20:53:47 <xnaas> s/behavior/feature
20:53:50 <xnaas> :D
20:54:42 <+MaxLeiter> 32 contributors in this release. very cool
```

Desired Output:
```
20:53:32 <xnaas> of course, just a general note
20:53:40 <xnaas> testing the behavior reminded me to complain
20:53:47 <xnaas> s/behavior/feature
20:53:50 <xnaas> :D
20:54:42 <+MaxLeiter> 32 contributors in this release. very cool
```

## Additional Notes
This behavior is identical when dragging and dropping text as well.

PR #4473, opened 2022-02-12
Include leading '<' + remove leading ' ' when copying messages

progval:
Closes GH-4369.

1. Fix missing leading '<'

Firefox does not seem to select leading (or trailing) characters that are
too small; so this commit sets a very small width, that is still large
enough to be selected.

This commit also adds `display: inline-block`, so the width is not
ignored; but this causes Chrome to ignore the space after `>`, so I made
it a non-breakable space.

An alternative is to make only the leading `only-copy` an
`inline-block`, but I think the non-breakable space is a good idea
regardless.

2. Remove leading ' ' when copying multiple messages

### DIFF
--- a/client/components/Message.vue
+++ b/client/components/Message.vue
@@ -27,11 +27,11 @@
 			</span>
 		</template>
 		<template v-else-if="isAction()">
-			<span class="from"><span class="only-copy">*** </span></span>
+			<span class="from"><span class="only-copy">***&nbsp;</span></span>
 			<component :is="messageComponent" :network="network" :message="message" />
 		</template>
 		<template v-else-if="message.type === 'action'">
-			<span class="from"><span class="only-copy">* </span></span>
+			<span class="from"><span class="only-copy">*&nbsp;</span></span>
 			<span class="content" dir="auto">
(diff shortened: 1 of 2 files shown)


MaxLeiter:
@xnaas as you reported the bug I'd appreciate your approval here

xnaas:
Brief testing shows that this solves the `<` copy issue on Chromium and Firefox, but the empty space before timestamps (when copying multiple lines) still seems to be present at least when testing on Firefox 98beta and 91.5.0esr (Windows 10). Chromium seems fully sorted out, though.

progval:
oh yeah I missed the second part. Fixed as well now.

MaxLeiter (approved):

xnaas:
I don't know that there's an solution to is, but I noticed that iOS Safari doesn't get the `<`. Highlighting stuff in general is a massive pain in the ass on iOS though, so probably outside of the scope of this PR.

progval:
You could try replacing `0.01px` with something a little bit bigger

brunnre8:
Thanks Val, as always.

Issue #4150, opened 2021-01-28
Using a DNS hostname in defaults/host does not work


FlyveHest:
- _Device, operating system:_ Docker
- _The Lounge version:_ Latest docker image pr. 28th of January

If I use a DNS based hostname in the defaults/host field, I get the following error when starting the docker image

`[ERROR] Failed to load messages: TypeError: Cannot read property 'network' of null`

Uncommenting and using the default from config.js (chat.freenode.net), results in the same error.

If I use an IP, it works as expected.

---

FlyveHest:
I uncommented the freenode section that is included in the default config.js, which I assume should work.

It was in quotes, yes

brunnre8:
can you paste your config in full (redact any passwords from say the webirc section) but leave it as close to your actual config as you can

FlyveHest:
Sure, looks like the comments parser fudges things up a bit though.

If I change the `host: "172.18.0.8"` to, for instance, `host: "chat.freenode.net"` I get the error.

If I remove my defaults section completely, and uncomment the defaults that is already in the config.js file, I get the error.

-- CONFIG START --
"use strict";

module.exports = {

	// ## Server settings
	// ### `public`
	//
	// When set to `true`, The Lounge starts in public mode. When set to `false`,
	// it starts in private mode.
	//
	// - A **public server** does not require authentication. Anyone can connect
	//   to IRC networks in this mode. All IRC connections and channel
	//   scrollbacks are lost when a user leaves the client.
	// - A **private server** requires users to log in. Their IRC connections are
	//   kept even when they are not using or logged in to the client. All joined
	//   channels and scrollbacks are available when they come back.
	//
	// This value is set to `false` by default.
	public: false,

	// ### `host`
	//
	// IP address or hostname for the web server to listen to. For example, set it
	// to `"127.0.0.1"` to accept connections from localhost only.
	//
	// For UNIX domain sockets, use `"unix:/absolute/path/to/file.sock"`.
	//
	// This value is set to `undefined` by default to listen on all interfaces.
	host: undefined,

	// ### `port`
	//
	// Set the port to listen to.
	//
	// This value is set to `9000` by default.
	port: 9000,

	// ### `bind`
	//
	// Set the local IP to bind to for outgoing connections.
	//
	// This value is set to `undefined` by default to let the operating system
	// pick its preferred one.
	bind: undefined,

	// ### `reverseProxy`
	//
[...]

PR #4475, opened 2022-02-13
Network.validate: Deduplicate code + tell users what the invalid hostname is

progval:
Along with GH-4476, this should help debug issues like GH-4150.

### DIFF
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -114,6 +114,17 @@ Network.prototype.validate = function (client) {
 	this.proxyPassword = cleanString(this.proxyPassword);
 	this.proxyEnabled = !!this.proxyEnabled;
 
+	const error = function (network, text) {
+		network.channels[0].pushMessage(
+			client,
+			new Msg({
+				type: Msg.Type.ERROR,
+				text: text,
+			}),
+			true
+		);
+	};
+
 	if (!this.port) {
 		this.port = this.tls ? 6697 : 6667;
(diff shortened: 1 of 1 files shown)


src/models/network.js
```diff
@@ -134,15 +145,7 @@ Network.prototype.validate = function (client) {
...
-			this.channels[0].pushMessage(
-				client,
-				new Msg({
-					type: Msg.Type.ERROR,
-					text: "Hostname you specified is not allowed.",
-				}),
-				true
-			);
-
+			error(this, `Hostname you specified (${this.host}) is not allowed.`);
```
MaxLeiter, line 148: Can we change this to `The hostname`?

MaxLeiter (changes requested):

MaxLeiter (approved):

src/models/network.js
```diff
@@ -114,6 +114,17 @@ Network.prototype.validate = function (client) {
 	this.proxyPassword = cleanString(this.proxyPassword);
 	this.proxyEnabled = !!this.proxyEnabled;
 
+	const error = function (network, text) {
+		network.channels[0].pushMessage(
```
MaxLeiter, line 118: Are you against removing `network` and just using `this.channels`?

progval: iirc that didn't work for some reason, I think `this` gets bound to something else

PR #4476, opened 2022-02-13
Show a nicer error in Chan.loadMessages() when network is misconfigured

progval:
ie. an actual error message instead of crashing on a `null` value, with the error reported in GH-4150.

The real (and helpful) error message will be displayed in the client's server buffer, thanks to GH-4475

### DIFF
--- a/src/models/chan.js
+++ b/src/models/chan.js
@@ -241,6 +241,12 @@ Chan.prototype.loadMessages = function (client, network) {
 		return;
 	}
 
+	if (!network.irc) {
+		// Network created, but misconfigured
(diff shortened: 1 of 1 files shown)


itsjohncs (approved):
Didn't test this but it looks fine. We'd get a bunch of "can't access bla of undefined" errors if we didn't return early. So at worst we'll be causing some failures that previously gave error messages to simply fail silently, but I'll trust that 4475 provides good error reporting.

Thanks for the PR! I'll merge this in if no one else says anything by tomorrow.

progval:
I re-added a log line, as suggested by @MaxLeiter

MaxLeiter (approved):

PR #4477, opened 2022-02-13
Preserve client certificate when TLS is indirectly enabled by a STS policy

### LINKED ISSUE #4152: SASL certificate gets recreated every time if server uses STS
- _Node version:_ v13.14.0
- _The Lounge version:_ v4.2.0

Running v4.2.0 when I edit settings of an IRC network using STS (irc.orpheus.network:7000) every time I click "Save network" the certificate gets recreated. The bug does not appear on non-STS networks.

progval:
Closes GH-4152

Only the last commit directly addresses the issue; the first four commits are only loosely related test refactoring (needed to test the fix) and additions (which test the basics this fix's tests rely on)

### DIFF
--- a/test/models/network.js
+++ b/test/models/network.js
@@ -6,6 +6,7 @@ const Msg = require("../../src/models/msg");
 const User = require("../../src/models/user");
 const Network = require("../../src/models/network");
 const Helper = require("../../src/helper");
+const STSPolicies = require("../../src/plugins/sts");
 
 describe("Network", function () {
 	describe("Network(attr)", function () {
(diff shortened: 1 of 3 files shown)


MaxLeiter (approved):

MaxLeiter:
Looks good. My one potential change would be removing the [test/fixtures/env.js](https://github.com/thelounge/thelounge/pull/4477/files#diff-a0a3dc4f01870d9b985fe08b97efa26a8d7bc4e64d0bd20bb9c16a7dd8eff109) changes so the STS policy stuff is only enforced on the relevant tests .

progval:
I don't understand. Are you saying I should add

```
STSPolicies.refresh.cancel();
fs.unlinkSync(STSPolicies.stsFile);
```

to each of the three tests that use STS stuff?

MaxLeiter:
Yeah, maybe I'm misunderstanding but we only want that code to run after certain tests, right?

test/models/network.js
```diff
@@ -223,6 +223,28 @@ describe("Network", function () {
 			ClientCertificate.remove(network.uuid);
 			Helper.config.public = true;
 		});
+
+		it("should remove client certs if there is a STS policy", function () {
```
brunnre8, line 227: it **should remove** client certs?

>expect(ClientCertificate.get(network.uuid)).to.deep.equal(client_cert); // Should be unchanged

Well :wink: I think there's a not missing no?

brunnre8:
Other than the test naming, this PR looks good I think.

But what I don't understand (and this has nothing to do with the PR itself, just poking your minds here).
Why on earth are we deleting the cert just because someone sets TLS to no TLS?

We shouldn't be **sending** the cert over plain text, but deleting it?
Say someone was doing troubleshooting over the plain text port, only to set it back to TLS and find their auth broken because the cert they setup is gone.

Anything I'm missing that warrants that behavior?

progval:
I completely agree. I assumed there was a good reason for this, but I'll gladly drop the cert removal entirely

MaxLeiter (approved):

brunnre8:
Thanks a lot Val :)

PR #4480, opened 2022-02-14
Don't truncate long passwords

progval:
No need for that; irc-framework properly handles linewrapping long passwords; and 300 characters is more than enough to trigger line-wrapping anyway.

Technically there is also no need for the sanitization, as passwords are base64-encoded, but it probably doesn't hurt to do it.

### DIFF
--- a/src/models/network.js
+++ b/src/models/network.js
@@ -90,7 +90,7 @@ Network.prototype.validate = function (client) {
 	const cleanNick = (str) => str.replace(/[\x00\s:!@]/g, "_").substring(0, 100);
 
 	// Remove new lines and limit length
-	const cleanString = (str) => str.replace(/[\x00\r\n]/g, "").substring(0, 300);
+	const cleanString = (str, length = 300) => str.replace(/[\x00\r\n]/g, "").substring(0, length);
 
 	this.setNick(cleanNick(String(this.nick || Helper.getDefaultNick())));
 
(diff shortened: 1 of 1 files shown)


src/models/network.js
```diff
@@ -106,7 +106,7 @@ Network.prototype.validate = function (client) {
 	this.host = cleanString(this.host).toLowerCase();
 	this.name = cleanString(this.name);
 	this.saslAccount = cleanString(this.saslAccount);
-	this.saslPassword = cleanString(this.saslPassword);
+	this.saslPassword = this.saslPassword.replace(/[\x00\r\n]/g, ""); // Clean but don't truncate
```
brunnre8, line 109: Hm... we should probably not open code this here, but rather split the cleanString method into cleanString and cleanStringAndTruncate.
Problem being that should this ever change you now need to fix it twice.

Then again, does this ever change?

MaxLeiter: We can also test this + cleanString if we split it into a new function and move them to a new file/export them, but that may be overkill for a fairly simple regex.

xPaw: What I would do is add length param to `cleanString`, and just provide a higher length for password (I mean right now do people seriously have passwords longer than 300 chars? that's already silly).

itsjohncs:
What's the motivation for doing this? Similarly to xpaw, 300 already seems enormous for a password to me.

itsjohncs:
I see! Thank you for taking the time to try prevent a similar error from catching someone else out.

But I'd rather see this solved by making the truncation not-silent. Could also prevent the password from being that long at all (via a length limit on the password input for example). Raising the limit seems like we're just passing the buck (I'm also suspicious that this won't work across all IRC servers as implemented).

I'm going to close this PR but if anyone else wants to see this PR continue please feel free to reopen.

progval:
> I'm also suspicious that this won't work across all IRC servers as implemented

FWIW, it does work on Atheme and every major IRCd. The only software I found not to support it is Anope; but The Lounge can already trigger that bug with the current length limit, because The Lounge truncates on number of characters instead of number of bytes.

PR #4484, opened 2022-02-14
Use the DNS result order returned by the OS

### LINKED ISSUE #3420: IPv4 precedence over IPv6 when connecting to network
- _Node version:_ v12.10.0
- _Browser version:_ Firefox 69.0
- _Device, operating system:_ Archlinux (on server and client)
- _The Lounge version:_ v3.2.0

My internet connection has an IPv6 and a CGNAT ipv4.
When connecting to chat.freenode.net, thelounge tries to connect via ipv4 by default.
If I edit the Network and put the IPv6 on the server, it does not connect due to ssl.

TheLouge should try to connect first with IPv6 when available

Since the public ipv4 is shared among other people, it got listed on DroneBL.

sfan5:
Calling [`setDefaultResultOrder('verbatim')`](https://nodejs.org/api/dns.html#dnssetdefaultresultorderorder) will make Node prefer IPv6 for socket connections (if possible) just like most other applications.
IMO this should be the default for any application in 2022 anyway but this is particularly relevant for IRC connections as highlighted by #3420.
The function was added in some version of Node 16/14 so there's an `if` for now.

closes #3420

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,4 +25,11 @@ if (!require("semver").satisfies(process.version, pkg.engines.node)) {
 	process.exit(1);
 }
 
+const dns = require("dns");
+
+// Set DNS result order early before anything that may depend on it happens.
+if (dns.setDefaultResultOrder) {
+	dns.setDefaultResultOrder("verbatim");
+}
+
 require("./src/command-line");


MaxLeiter (approved):

sfan5:
True, I thought about mentioning RFC 6724 but I doubt that means anything to most people.
"Set default DNS result order to order returned by the OS" maybe?

MaxLeiter:
`Use the default result order provided by the operating system`?

brunnre8:
s/os/dns resolver/

why gloss over it? Just tell it the way it is .
It's not a knob, people that care know what it means and for the rest it's irrelevant

brunnre8:
Thanks  :smiley:

PR #4485, opened 2022-02-14
Automatically update ignored list when ignoring / unignoring

GewoonYorick:
Title says it all ;)

When the ignore window is opened on the client it'll automatically update the ignored-user list. 

Added an additional message when the ignorelist is empty.

Fixed bug where an open ignorelist would not update when all ignore entries were removed.

### DIFF
--- a/client/components/Special/ListIgnored.vue
+++ b/client/components/Special/ListIgnored.vue
@@ -29,8 +29,8 @@ export default {
 		channel: Object,
 	},
 	methods: {
-		getLocaletime(date) {
-			return localetime(date);
+		getLocaletime(datetime) {
+			return localetime(datetime);
 		},
 	},
 };
--- a/client/components/Special/ListIgnored.vue
+++ b/client/components/Special/ListIgnored.vue
@@ -9,7 +9,7 @@
 		<tbody>
 			<tr v-for="user in channel.data" :key="user.hostmask">
 				<td class="hostmask"><ParsedMessage :network="network" :text="user.hostmask" /></td>
-				<td class="when">{{ localetime(user.when) }}</td>
+				<td class="when">{{ getLocaletime(user.when) }}</td>
 			</tr>
 		</tbody>
 	</table>
@@ -29,7 +29,7 @@ export default {
 		channel: Object,
 	},
 	methods: {
-		localetime(date) {
+		getLocaletime(date) {
 			return localetime(date);
 		},
 	},
--- a/client/components/Special/ListIgnored.vue
+++ b/client/components/Special/ListIgnored.vue
@@ -1,5 +1,5 @@
 <template>
-	<div v-if="channel.data.length === 0" class="empty-list">Your ignorelist is empty.</div>
+	<div v-if="channel.data.length === 0" class="empty-ignore-list">Your ignorelist is empty.</div>
 	<table v-else class="ignore-list">
 		<thead>
 			<tr>
@@ -17,7 +17,7 @@
 </template>
 
 <style scoped>
-.empty-list {
+.empty-ignore-list {
 	padding: 0 0.5rem;
 }
 </style>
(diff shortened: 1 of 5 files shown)


client/components/Special/ListIgnored.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-	<table class="ignore-list">
+	<div v-if="channel.data.length === 0" class="empty-list">Your ignorelist is empty.</div>
```
MaxLeiter, line 2: `empty-ignore-list`

client/components/Special/ListIgnored.vue
```diff
@@ -9,12 +10,18 @@
 		<tbody>
 			<tr v-for="user in channel.data" :key="user.hostmask">
 				<td class="hostmask"><ParsedMessage :network="network" :text="user.hostmask" /></td>
-				<td class="when">{{ localetime(user.when) }}</td>
+				<td class="when">{{ getLocaletime(user.when) }}</td>
```
MaxLeiter, line 13: Not sure this is a necessary name change

MaxLeiter (changes requested):

client/components/Special/ListIgnored.vue
GewoonYorick: I changed it because we import `localetime` and also use it as method name. Technically this works fine, but to make it easier to distinguish, a different name would be better imo. :)

MaxLeiter (approved):

MaxLeiter:
Tested and worked well. Thanks @GewoonYorick!

src/plugins/inputs/ignore.js
```diff
@@ -10,6 +10,7 @@ exports.input = function (network, chan, cmd, args) {
 	const client = this;
 	let target;
 	let hostmask;
+	let create_new_ignored_window = false;
```
brunnre8, line 13: You don't really need this bool.
The structure that you propose is rather strange, prior to your changes the case statement was the last thing that happened, that should stay this way.

src/plugins/inputs/ignore.js
```diff
@@ -110,35 +111,41 @@ exports.input = function (network, chan, cmd, args) {
...
-						data: ignored,
-					});
-				}
+				create_new_ignored_window = true;
 			}
 
 			break;
 	}
+
+	const chanName = "Ignored users";
```
brunnre8, line 120: Now, suddenly ignore + unignore end up here, even though you already know that they shouldn't.
The only code path that should ever see this is ignorelist, so scope it to that.
If it gets too long to inline it into the switch (which may very well be the case), write a function for it.

src/plugins/inputs/ignore.js
```diff
@@ -110,35 +111,41 @@ exports.input = function (network, chan, cmd, args) {
...
+	const channel = network.getChannel(chanName);
+
+	if (typeof channel === "undefined" && !create_new_ignored_window) {
```
brunnre8, line 123: The only point of this flag is to get rid of ignore + unignore again, you wouldn't even have to have this flag if you scope it properly.

brunnre8 (changes requested):
Thanks a lot for your work, would you mind re-structuring this a bit?
No offense intended, I simply don't understand why you want to do this after the switch statement, with a bool guard where you could just make this a function and call that in the single place that cares.

brunnre8:
Nevermind, found my misconception... but that code as is is way to subtle for my taste... give me a sec, trying to tell you what I mean in code rather than English

brunnre8:
So, here's what I mean in code: https://github.com/GewoonYorick/thelounge/pull/1

brunnre8:
@MaxLeiter ping, please have a look at the changes from my side there, which do you find easier to read?

MaxLeiter:
whoops, missed your comment @bookworm. @GewoonYorick: if you want to continue this, I think bookworm's solution is ideal. If you run into any issues updating to typescript, or don't have the time / want to, please let me know and I can take it up / help out.

PR #4487, opened 2022-02-16
sqlite: Escape '%' and '_' in search queries.

### LINKED ISSUE #4481: History search queries are not sanitized
The Lounge source (1ed4f57a / v4.3.0) (Node.js 17.3.1 on linux x64)

---

SQL [LIKE](https://mariadb.com/kb/en/like/) has two wildcard characters and these are not sanitized, consequently this works:
Bildschirmfoto_2022-02-14_21-44-23

progval:
I picked '@' arbitrarily, it doesn't matter much.
I just don't like '\' because it needs to be escaped itself in the JS code,
which is annoying.

Closes GH-4481

### DIFF
--- a/src/plugins/messageStorage/sqlite.js
+++ b/src/plugins/messageStorage/sqlite.js
@@ -205,9 +205,12 @@ class MessageStorage {
 			return Promise.resolve([]);
 		}
 
+		// Using the '@' character to escape '%' and '_' in patterns.
+		const escapedSearchTerm = query.searchTerm.replace(/([%_@])/g, "@$1");
+
 		let select =
-			'SELECT msg, type, time, network, channel FROM messages WHERE type = "message" AND json_extract(msg, "$.text") LIKE ?';
-		const params = [`%${query.searchTerm}%`];
+			'SELECT msg, type, time, network, channel FROM messages WHERE type = "message" AND json_extract(msg, "$.text") LIKE ? ESCAPE \'@\'';
+		const params = [`%${escapedSearchTerm}%`];
 
 		if (query.networkUuid) {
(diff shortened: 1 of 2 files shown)


itsjohncs:
Why do you think these characters ought to be escaped?

MaxLeiter:
@itsjohncs there's some discussion in #4481, here's my comment:

> I agree with @brunnre8, there's a chance you want to search for % or _. We can implement more user-friendly and/or transparent methods that use them on the back-end in the future. I personally like the idea of github-style search, i.e. from:MaxLeiter in:#thelounge but that's for another issue

I think we might want to escape them with \, which is at least consistent with javascript

itsjohncs:
What an obvious thing to miss, of course there's more explanation in the linked issue 😅. Thank you @MaxLeiter. Sorry @progval for not looking at that issue.

> I think we might want to escape them with \, which is at least consistent with javascript

Since it's an implementation detail that isn't exposed to the user, I'm pretty indifferent to the actual escape character that's used. The forward slash would certainly be a bit more normal though so might cause a bit less confusion for folks reading the code 🤷‍♀️.

brunnre8 (approved):
Looks good to me, thanks Val.

I don't care what the escape is, @ is fine and the variable naming doesn't leave much room for interpretation so :shrug:

itsjohncs (approved):
Sorry for the delay. We're waiting to merge any more PRs until we make a release so this'll be sitting until then, but we'll merge this afterwards.

I did some quick testing just now to verify that `%` and `_` are treated literally in search queries. LGTM.

MaxLeiter:
LGTM 🎉 

thanks for your patience @progval!

Issue #1085, opened 2017-04-24
Clicking somewhere outside of input/autocomplete still keeps the menu around


astorije:
*This will only be relevant when #787 gets merged as I don't expect it to fix this minor issue.*

If the autocomplete dropdown gets opened, and a right click is pressed on channel/network/user, the context menu opens but the autocomplete menu does not close.

astorije:
@yashsriv, do you think you'd be able to take a look into that once your PR gets merged? It's not critical so don't try to fix it on your PR directly or that might delay the merging again, while your PR is essentially ready.

richrd:
Is this actually a bug? I don't see it being much of a problem actually.

xPaw:
I think the actual issue is that clicking somewhere outside of input/autocomplete still keeps the menu around.

PR #4493, opened 2022-02-23
Autocomplete: update to @textcomplete package and close on blur

MaxLeiter:
The library we use, https://github.com/yuku/old-textcomplete, was archived in favor of this new (almost) drag-and-drop replacement: https://github.com/yuku/textcomplete. I also addressed #1085 as it was a simple fix.


client/js/autocompletion.js
```diff
@@ -214,29 +216,21 @@ function enableAutocomplete(input) {
...
 		emojiStrategy,
 		nicksStrategy,
 		chanStrategy,
 		commandStrategy,
 		foregroundColorStrategy,
 		backgroundColorStrategy,
-	]);
+	];
 
-	// Activate the first item by default
```
MaxLeiter, line 234: Having tested on 4.3.0, I don't think this does anything anymore, and the issue is hidden / not archived on Wayback Machine, so I removed this code. The Textcomplete types don't point to `dropdown` valid anymore.

client/js/autocompletion.js
```diff
@@ -3,7 +3,9 @@
 const constants = require("./constants");
 
 import Mousetrap from "mousetrap";
-import {Textcomplete, Textarea} from "textcomplete";
+const {Textcomplete} = require("@textcomplete/core");
+const {TextareaEditor} = require("@textcomplete/textarea");
```
itsjohncs, line 7: Feels odd to mix require and import here. Does this pattern work instead?:

```js
import textcomplete from "@textcomplete/core";
const {Textcomplete} = textcomplete;
```

itsjohncs (approved):
LGTM. Tested locally (briefly) on Safari on Mac OS and the emoji picker works as before afaict.

client/js/autocompletion.js
MaxLeiter: yeah this will work, I just used vscodes auto import tool and didn't bother checking.

PR #4494, opened 2022-02-24
Break up style.css and move styles to components

### LINKED ISSUE #1551: Split style.css into subcomponents and preprocess css as a build step

MaxLeiter:
Closes #1551

Big WIP, will require a lot of testing before merge/release, but I'm pretty happy with it so far. Some styles are shared between components, like SearchResults and Chat, so for those I opted to create a css file for their shared styles. I'm not too sure I'm happy with that, though. I also haven't finished moving everything over yet, but maybe that shouldn't be the goal of this PR.

I used scoped styles where possible: https://vue-loader.vuejs.org/guide/scoped-css.html, except for `App.vue`


client/components/App.vue
```diff
@@ -16,6 +16,185 @@
 	</div>
 </template>
 
+<style>
+@import "../css/icons.css";
+@import "../css/chat-and-search.css";
```
MaxLeiter, line 21: I feel like a better distinction can be made between chat-and-search and chat-messages but I don't know it. suggestions welcome

PR #4495, opened 2022-02-24
set 'video/quicktime' to 'video/mp4' for uploads

ghost:
`video/quicktime` only plays in Firefox and Safari.
`video/mp4` plays in Firefox, Safari, and Chromium-based browsers.

### DIFF
--- a/src/plugins/uploader.js
+++ b/src/plugins/uploader.js
@@ -119,6 +119,8 @@ class Uploader {
 			detectedMimeType = "audio/flac";
 		} else if (detectedMimeType === "audio/x-m4a") {
 			detectedMimeType = "audio/mp4";
+		} else if (detectedMimeType === "video/quicktime") {
+			detectedMimeType = "video/mp4";
 		}
 
 		res.setHeader("Content-Disposition", disposition);


xnaas:
Test:

image

The top link was uploaded from The Lounge with this PR. The bottom link was uploaded from stable (4.3.0).

xnaas:
I suppose this merits some additional discussion.

There are still `.mov` files that are served with `video/quicktime` around the internet, primarily Discord does this. In all my testing, a `.mov` with `video/quicktime` would not play in any Chromium-based browser (but played fine in Firefox and Safari), but switching it to `video/mp4` made it so the files would play in all 3 major browser types.

Should this PR also be expanded so TL embeds `video/quicktime` even though Chromium-based browsers do not support it?

xnaas:
Test be8a853:

image

Left: Edge | Right: Firefox

Vivaldi also a no-go on embedding video/quicktime:

image

itsjohncs:
Took me some time to understand what's happening here, so I want to check my understanding with you @xnaas:

Currently on `master` we won't show previews for any videos with the `video/quicktime` mimetype. This change aims to support previewing such videos.

Because the `video/quicktime` mime type is poorly supported by browsers, we'll avoid serving videos that our filetype detector marks with that mimetype and we'll instead serve them as `video/mp4`.

External links served with `video/quicktime` will show previews as well but this will only work on Firefox and Safari. On other browsers the embed arrow will be shown but the embed will not work.

--

Am I understanding everything you're trying to do here @xnaas?

xnaas:
That is the route I went with this, yes, @itsjohncs.

itsjohncs:
Thanks. I'm hesitant about trying to preview external `video/quicktime` links. Chromium browsers are pretty popular and this change would be knowingly introducing a bug for them.

Is there a way to hide the preview on the client once it's clear that the preview can't be loaded?

xnaas:
> Chromium browsers are pretty popular and this change would be knowingly introducing a bug for them.

All video embeds have this bug on mobile Safari, which could be up to 20% of users. 😜 I get the point, though. Would be nice for anyone who cares about Chromium to maybe bump [Chromium issue #1120292](https://bugs.chromium.org/p/chromium/issues/detail?id=1120292) and mention this, dunno.

> Is there a way to hide the preview on the client once it's clear that the preview can't be loaded?

That, unfortunately, will require feedback from another maintainer. I wouldn't have a clue. 😅 (and per above mention of iOS issue, would love something for this)

itsjohncs (changes requested):
I think we ought to remove the part of this PR that allows external links served with `video/quicktime` to be served then. That seems like something that ought to be done only after we can handle broken embeds on the client.

If you'd like to pass the torch to someone else to implement that at some point: making an issue and hoping it attracts someone seems like a good enough route.

xnaas:
I "un-did" the 2nd commit, which contained the "offending" code.

itsjohncs (approved):
Thanks, looks good to me. Though I didn't test this myself since it seems like the testing you've described is sufficient.

We're waiting to merge anything until after we make a release, but we'll merge this after that.

MaxLeiter (approved):

Issue #4504, opened 2022-03-04
Cannot load some plugins when on an rc build of TL


xnaas:
- _Node version:_ `v16.14.0`
- _Browser version:_ N/A
- _Device, operating system:_ N/A
- _The Lounge version:_ `4.3.1-rc.1`

---

> [ERROR] Package thelounge-plugin-shortcuts could not be loaded: v1.1.1 does not support this version of The Lounge. Supports: >=4.3.0

The Lounge fails to load some plugins when you're using an rc build, presumably the `-rc.1` is causing a problem with version comparison.

xnaas:
Closed by #4506.

PR #4506, opened 2022-03-07
Plugins: include pre-releases in compatibility lookup

brunnre8:
Semver doesn't treat pre-release versions as upgrades, meaning >4.3.0 isn't satisfied
by 4.3.1-rc.1.
For the purpose of TL plugins however, we are only interested in the semantic version and
expect that rc's adhere to the compatibility promise.

### DIFF
--- a/src/plugins/packages/index.js
+++ b/src/plugins/packages/index.js
@@ -109,7 +109,9 @@ function loadPackage(packageName) {
 
 		if (
 			packageInfo.thelounge.supports &&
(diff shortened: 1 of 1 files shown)


xnaas:
Closes #4504.

itsjohncs (approved):
LGTM. I didn't do any testing myself but I looked through `sermver`'s docs to make sure I understand what's happening here and everything seems fine.

PR #4507, opened 2022-03-07
User: fix file permissions on create

brunnre8:
User files contain secrets and should be protected.
Chances are that the user folder can be protected as well,
so let's do that if TL is creating the folder.

### DIFF
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -169,7 +169,9 @@ ClientManager.prototype.addUser = function (name, password, enableLog) {
 	};
 
 	try {
-		fs.writeFileSync(userPath, JSON.stringify(user, null, "\t"));
+		fs.writeFileSync(userPath, JSON.stringify(user, null, "\t"), {
+			mode: 0o600,
+		});
 	} catch (e) {
 		log.error(`Failed to create user ${colors.green(name)} (${e})`);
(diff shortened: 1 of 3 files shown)


itsjohncs:
Did you test this by bringing up a clean install? Seems fine looking at the code but that seems like a worthwhile check to make sure nothing falls down.

brunnre8:
Yes:

```
[drwx------ reto     reto    ]  /tmp/theloungedir
├── [-rw-r--r-- reto     reto    ]  config.js
├── [drwxr-x--- reto     reto    ]  logs
├── [drwxr-xr-x reto     reto    ]  packages
│   ├── [drwxr-xr-x reto     reto    ]  node_modules
│   └── [-rw-r--r-- reto     reto    ]  package.json
├── [drwx------ reto     reto    ]  users
│   └── [-rw------- reto     reto    ]  dummy.json
└── [-rw------- reto     reto    ]  vapid.json
```

After setting THELOUNGE_HOME to a new dir, starting the server and a single user add.

Only thing that's not currently protected is the packages folder, which might need to be protected if plugins do something sensitive, but I think they can do it themselves if they care.

itsjohncs (approved):
Thanks for confirming.

PR #4508, opened 2022-03-08
cli: don't error if the user folder doesn't exist

brunnre8:
The user folder gets created on demand, thelounge list should not
fail if the folder doesn't exist.
This just means that no users are present, so report that instead.

### DIFF
--- a/src/command-line/users/list.js
+++ b/src/command-line/users/list.js
@@ -3,8 +3,6 @@
 const log = require("../../log");
 const colors = require("chalk");
 const program = require("commander");
-const fs = require("fs");
-const Helper = require("../../helper");
 const Utils = require("../utils");
 
 program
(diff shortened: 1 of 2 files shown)


src/command-line/users/list.js
```diff
@@ -25,16 +18,17 @@ program
...
-				log.info(`${i + 1}. ${colors.bold(user)}`);
-			});
-		} else {
+		if (users.length === 0) {
 			log.info(
 				`There are currently no users. Create one with ${colors.bold(
 					"thelounge add <name>"
 				)}.`
 			);
+			return;
```
brunnre8, line 27: I'm a fan of return early, keep the happy path to the left.
If you disagree heavily please tell me and I'll adapt my style.

itsjohncs: I think either way is fine. Code style seems like a fairly low stakes thing in an OSS code case with an auto formatter too.

itsjohncs (approved):
LGTM. Didn't test this myself but the code paths are straightforward so I think the testing you've done in development should be sufficient.

PR #4514, opened 2022-03-14
add config files for local docker development

kinghat:
to make it easier to get up and going with developing in a container for all of the benefits that can provide.

this should be fully working from my testing. i have not tested if the `${localEnv:USERPROFILE}/.thelounge` part of the mount works on windows.

there might be push back on this being added to the repo and no hard feels if so. there are other vscode specific config files in the repo so 🤷‍♂️

this could be modified to use a compose file which might make it less vscode specific but these config files would still be needed in any case.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:lts-alpine
+
+ENV NODE_ENV production yarn build
```
brunnre8, line 3: that's not doing what you think it's doing, ENV sets an env var and that's it. In your case `NODE_ENV="production yarn build"`

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+ENV THELOUNGE_HOME "~/.thelounge"
+VOLUME "${THELOUNGE_HOME}"
+
+ENV PORT 9000
+EXPOSE ${PORT}
+
+RUN apk --update --no-cache --virtual add git openssh
```
brunnre8, line 11: Missing python, build-essentials or whatever it's named in musl land.
You'll have trouble getting the sqlite3 deb otherwise.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "The Lounge",
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
+  "extensions": [
```
brunnre8, line 6: Only pull in the minimum extensions actually neccessary. Gruntfuggly.todo-tree isn't it.
In fact, you probably shouldn't differ from https://github.com/thelounge/thelounge/blob/master/.vscode/extensions.json overly much.

brunnre8 (changes requested):

.devcontainer/Dockerfile
kinghat: sorry for being dense but afaict its building and querying the sqlite db with `["sqlite", "text"]` as my config setting :thinking:

.devcontainer/devcontainer.json
kinghat: what i had originally but got greedy: https://github.com/thelounge/thelounge/pull/4514/commits/9a11d76d63620048f4a4eaa009859c37d743d9d9

.devcontainer/Dockerfile
kinghat: removed: https://github.com/thelounge/thelounge/pull/4514/commits/0bb0e2f10baaee9471f31d73ec159027768a4d5e

.devcontainer/Dockerfile
brunnre8: On the master branch?

We pin a specific commit and the way it works is that sqlite3 tries to find a binary blob with the current Version/Arch/nodeAbi/platform and if it can't find that it falls back to building it.

If it still works if you don't actually have a compiler that means that you've pulled the blob. Which surprises me a tad.
Then it would look like it's only fetching the latest release version of the blob, no matter that it could've changed in the repo.

And considering that they changed the compilation flags but didn't bump any versions afaict and hence can't ID a binary blob with those mods... well.

https://github.com/mapbox/node-sqlite3/commit/918052b538b0effe6c4a44c74a16b2749c08a0d2

kinghat: thanks for the explanation. ive added `sqlite` to the image: https://github.com/thelounge/thelounge/pull/4514/commits/a90c59f01037c074c3841d84c3e400ad35d1ea1e

kinghat: undid the previous commit.

kinghat:
should this be tested on a windows environment? i dont have access to that. maybe a mac environment too, but id imagine its inline with linux.

kinghat:
updated the devcontainer configs to properly account for consumers who were already developing on host and want to use their `~/.thelounge` config inside the container. if starting fresh, that dir from inside the container will be bind mounted to the host machine.

tested and seems to be fully functioning on windows too.

kinghat:
added a little documentation: https://github.com/thelounge/thelounge/pull/4514/commits/496c00369f8195473f67a78c43b03e641d8c0811 lmk what you think. im personally not sure it needs to be mentioned in `CONTRIBUTING`. i feel like all of that could be moved to the development section of the `README`. or dropped all together. whatever works.

.devcontainer/docker-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+chown -R node:node "${THELOUNGE_HOME}"
+
+sleep infinity
```
brunnre8, line 5: ehm, why?

.github/CONTRIBUTING.md
```diff
@@ -35,3 +35,13 @@ your contributions.
...
+- When using VS Code, configuration files have been provided to take advantage of their [container development infrastructure](https://code.visualstudio.com/docs/remote/containers). touted benefits:
+
+    > The Remote - Containers extension lets you use a Docker container as a full-featured development environment. Whether you deploy to containers or not, containers make a great development environment because you can:
+    >
+    > - Develop with a consistent, easily reproducible toolchain on the same operating system you deploy to.
+    > - Quickly swap between different, separate development environments and safely make updates without worrying about impacting your local machine.
+    > - Make it easy for new team members / contributors to get up and running in a consistent development environment.
+    > - Try out new technologies or clone a copy of a code base without impacting your local setup.
+  
+  To get started add the [Remote Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension and follow installation instructions provided there.
```
brunnre8, line 47: Get rid of all but the first sentence.
> When using VS Code, configuration files have been provided to take advantage of their [container development infrastructure](https://code.visualstudio.com/docs/remote/containers).

Is fine, the rest doesn't belong in our readme.

README.md
```diff
@@ -87,3 +87,5 @@ Before submitting any change, make sure to:
   - Run `yarn format:prettier` if linting fails
 - Run `yarn build` if you change or add anything in `client/js` or `client/components`
 - `yarn dev` can be used to start The Lounge with hot module reloading
+
+If you would like to work on The Lounge in a containerized development environment, VS Code specific configuration files have been provided. If not prompted, use the `Rebuild and Reopen in Container` command to get going. More info in [Contributing](.github/CONTRIBUTING.md).
```
brunnre8, line 91: s/containerized development environment, VS Code/VS Code devcontainer/

way shorter and if you don't know what a devcontainer is, there's the link to the explanation in CONTRIBUTING.md.

That's just a suggestion though, bikeshedding a bit here. It's just that containerized development environment doesn't mean anything, can be a million of things (nspawn, mkosi, docker, fancy mount points...)

.devcontainer/docker-entrypoint.sh
kinghat: this dir needs to be owned by the same user running the container so that it doesnt become owned by root on the host when mounting into the container.

brunnre8: doesn't matter if it's owned by root on the host, if you use namespace mapping properly (as in set up uid maps) root in the container will not be uid 0 on the host.
And all example containers of vscode get away without any such chowning or in fact any entryfile at all

kinghat: the container is being run by user `node` with id of `1000`, but because docker is creating the config file on start of the container if it doesnt exists(a fresh dev setup) its created as root. this is what is done in the official docker: https://github.com/thelounge/thelounge-docker/blob/cfd776abbf841dc561608e4765b3774d24562232/docker-entrypoint.sh#L4

.github/CONTRIBUTING.md
kinghat: addressed in: https://github.com/thelounge/thelounge/pull/4514/commits/30c50438f4ea7b2045d30418a5ce10aed9ff8721

README.md
kinghat: > s/containerized development environment, VS Code/VS Code devcontainer/
> 
> way shorter and if you don't know what a devcontainer is, there's the link to the explanation in CONTRIBUTING.md.
> 
> That's just a suggestion though, bikeshedding a bit here. It's just that containerized development environment doesn't mean anything, can be a million of things (nspawn, mkosi, docker, fancy mount points...)

addressed: https://github.com/thelounge/thelounge/pull/4514/commits/30c50438f4ea7b2045d30418a5ce10aed9ff8721#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

kinghat:
going to update the docs for this as well

PR #4524, opened 2022-04-08
Add `prefetchTimeout` config setting

emilyst:
The problem I'm solving is that of uploading an image that's about 8 MB and other users being unable to preview it.

Capture 2022-04-08 at 15 52 41

This pull request adds a new configuration setting called `prefetchTimeout`. To my mind, it makes sense to add this alongside the existing configurations for prefetch size and search size, since changing those would result in longer requests.

I set the default value to 5000 ms, which is what it is currently hard-coded to be.

Security note: In my description for the default value, I caution against changing it unwarily, since it can hypothetically make The Lounge more vulnerable to denial of service attacks (e.g., by posting a huge number of large links in a short time, causing The Lounge to run out of available sockets, or by tying up a large amount of bandwidth). However, it's unclear to me how much functionality would be impacted by prefetch failing, and in any case, such a hypothetical vulnerability already exists (it's just harder to exploit with the default settings).

So I might be worried about nothing.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -159,7 +159,7 @@ module.exports = {
 	//
 	// This value is set to `50` kilobytes by default.
 	prefetchMaxSearchSize: 50,
-	
+
 	// ### `prefetchTimeout`
 	//
 	// When `prefetch` is enabled, this value sets the number of milliseconds
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -159,6 +159,18 @@ module.exports = {
 	//
 	// This value is set to `50` kilobytes by default.
 	prefetchMaxSearchSize: 50,
+	
+	// ### `prefetchTimeout`
+	//
+	// When `prefetch` is enabled, this value sets the number of milliseconds
+	// before The Lounge gives up attempting to fetch a link. This can be useful
+	// if you've increased the `prefetchMaxImageSize`.
+	//
+	// Take caution, however, that an inordinately large value may lead to
+	// performance issues or even a denial of service, since The Lounge will not
+	// be able to clean up outgoing connections as quickly. Usually the default
+	// value is appropriate, so only change it if necessary.
+	prefetchTimeout: 5000,
 
 	// ### `fileUpload`
 	//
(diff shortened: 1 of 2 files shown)


emilyst:
Not sure what the deal is with the tests, since I cannot reproduce the same result locally.

src/plugins/irc-events/link.js
```diff
@@ -390,7 +390,7 @@ function fetch(uri, headers) {
 		try {
 			const gotStream = got.stream(uri, {
 				retry: 0,
-				timeout: 5000,
+				timeout: Helper.config.prefetchTimeout, // milliseconds
```
MaxLeiter, line 393: Can you add a warning and fallback for if `prefetchTimeout` isn't present? 

Something like this a few lines up:
```js
const prefetchTimeout = Helper.config.prefetchTimeout;
if (!prefetchTimeout) {
    console.warn("prefetchTimeout is missing from your The Lounge configuration, defaulting to 5000");
}
```

defaults/config.js
```diff
@@ -159,6 +159,18 @@ module.exports = {
...
+	//
+	// When `prefetch` is enabled, this value sets the number of milliseconds
+	// before The Lounge gives up attempting to fetch a link. This can be useful
+	// if you've increased the `prefetchMaxImageSize`.
+	//
+	// Take caution, however, that an inordinately large value may lead to
+	// performance issues or even a denial of service, since The Lounge will not
+	// be able to clean up outgoing connections as quickly. Usually the default
+	// value is appropriate, so only change it if necessary.
+	prefetchTimeout: 5000,
```
MaxLeiter, line 173: Can you add a line specifying the default like the other options? 

e.g.
```
// This value is set to `50` kilobytes by default.
```

MaxLeiter (changes requested):

src/plugins/irc-events/link.js
emilyst: Take a glance at the new commits and see if those changes suit?

MaxLeiter (approved):

PR #4536, opened 2022-04-13
Update sqlite3 to 5.0.3

xPaw:
See https://github.com/TryGhost/node-sqlite3/releases/tag/v5.0.3

The package is now maintained by the Ghost team, most issues should be resolved now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "yarn": "1.22.17"
   },
   "optionalDependencies": {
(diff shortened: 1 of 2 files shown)


brunnre8:
Thanks, but this PR is missing the lockfile entry

xPaw:
Tested on raspberry without python2 (and no python-is-python3) using node 16, it compiles fine.

PR #4537, opened 2022-04-13
renovate: Disable digest updates

brunnre8:
There's a reason one pins a commit and it is because we
want that specific commit.
Renovate gets *really* noisy if commits are made frequently
to a repo, so let's disable it.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,5 @@
 {
-  "extends": ["config:base"],
+  "extends": ["config:base", ":disableDigestUpdates"],
   "ignorePaths": ["test/"],
   "labels": ["Type: Dependencies"],
   "lockFileMaintenance": {"enabled": true},


itsjohncs:
Sorry if there was talk about this somewhere that I missed. What exactly are renovate's digest updates? Searching online hasn't given me a very good understanding.

dgw:
@itsjohncs e.g. #4530 and #4531, trying to update sqlite3 to the latest commit even though it's pinned to a specific older one on purpose.

itsjohncs:
I see, I understand now. Renovate will try to update the git revision we specify for a dependency in `package.json` to a more recent one, but if we went to the trouble of specifying a revision we probably want that one in particular. Sounds good to me!

itsjohncs (approved):

PR #4542, opened 2022-04-19
Use nick as a realname fallback

brunnre8:
Currently the realname is set to an advertisement if it isn't explicitly
set by the user.
Some clients started to show the realname as a display name in their
UI, which makes this tedious as you'll end up with gazillion "The Lounge
User" entries.

To avoid this, set the realname to the nick on first connect, so that
it is useful.
Note that this isn't done on nick changes, but only on the initial
connect step.

Fixes: https://github.com/thelounge/thelounge/issues/4527

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -241,7 +241,7 @@ module.exports = {
 	// - `nick`: Nick name. Percent signs (`%`) will be replaced by random
 	//   numbers from 0 to 9. For example, `Guest%%%` may become `Guest123`.
 	// - `username`: User name.
-	// - `realname`: Real name.
+	// - `realname`: Real name displayed by some clients. Defaults to the nick if set to ""
 	// - `leaveMessage`: Network specific leave message (overrides global leaveMessage)
 	// - `join`: Comma-separated list of channels to auto-join once connected.
 	//
@@ -271,7 +271,7 @@ module.exports = {
 		rejectUnauthorized: true,
 		nick: "thelounge%%",
 		username: "thelounge",
-		realname: "The Lounge User",
+		realname: "",
 		join: "#thelounge",
 		leaveMessage: "",
 	},
--- a/server/models/network.ts
+++ b/server/models/network.ts
@@ -190,7 +190,7 @@ class Network {
 		}
 
 		this.username = cleanString(this.username) || "thelounge";
-		this.realname = cleanString(this.realname) || "The Lounge User";
+		this.realname = cleanString(this.realname) || this.nick;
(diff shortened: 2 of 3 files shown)


brunnre8:
This ignores the case where the server doesn't accept the nick as is, but immediately changes it to something else... not sure what the best behavior is there.

Example being a protected nick, where it allows you to connect as dummy for say 30s and then force changes the nick to something else if you don't identify.

Opinions?

dgw:
Once registration is complete, the realname can't be changed unless the server supports [`setname`](https://ircv3.net/specs/extensions/setname) (obviously TL/irc-fw would have to support it, too). Sending `USER` again after registration is supposed to return a 462 ERR_ALREADYREGISTERED.

Before registration finishes, if the server sends back an error in response to TL's `NICK` command, the automatic realname could be updated to match the fallback nick TL generates, but it doesn't seem worth the effort/maintenance. Even if the server forcibly changes the TL user's nick, that nick is still the user's chosen identity. I don't see a whole lot of reason not to leave the original nick in the realname field if it was set automatically.

test/models/network.js
```diff
@@ -83,6 +83,29 @@ describe("Network", function () {
 			expect(network2.username).to.equal("InvalidNick");
 		});
 
+		it("realname should be set to nick only if realname is empty", function () {
+			Helper.config.defaults.nick = "";
```
itsjohncs, line 87: This is pretty mysterious to me. Might be worth a comment explaining why this is necessary (if it is). I'm hoping `Helper` knows how to reset its configuration between test runs too since we're doing this in other tests as well.

itsjohncs (approved):
Looks good to me. Didn't test myself but code seems straightforward enough.

MaxLeiter (approved):

test/models/network.js
brunnre8: Honestly? No clue, I just copy pasted from another test.

I guess the reason is to make it less likely to make the test pass by chance due to the defaults.
But I on purpose took stupid names to avoid that.

I'll remove it and see what happens

brunnre8:
@itsjohncs seems like the line you objected at isn't necessary in my case, least the tests pass and I can't say I see something wrong with them.

@all I've also set the default to the empty string, else the modification would probably not have an effect at all.
Even now, all existing installations will continue to use the other thing and it's hard to change that considering that we copy paste all configuration option to a different location, rather than just merging it.

That's nice for the user, as they don't have settings changing from under them, but bad for us if we have a change like this where a better default would be available.

emersion:
LGTM!

PR #4547, opened 2022-04-27
Add /who handling

MaxLeiter:
irc-framework sends the WHOX `cuhsnfdaor` flags and doesn't support receiving others, so we manually remove parameters the user doesn't request.

I also added an uncaught exception handler on the server to stop a subset of crashes. The problem is people won't look at their server logs often, but bookworm and I agreed that's better than crashing. When we add an admin role we can report it to them.

https://ircv3.net/specs/extensions/whox

ref https://github.com/kiwiirc/irc-framework/issues/331

### DIFF
--- a/client/components/MessageTypes/whois.vue
+++ b/client/components/MessageTypes/whois.vue
@@ -1,6 +1,6 @@
 <template>
 	<span class="content">
-		<p>
+		<p v-if="message.whois.nick">
 			<Username :user="{nick: message.whois.nick}" />
 			<span v-if="message.whois.whowas"> is offline, last information:</span>
 		</p>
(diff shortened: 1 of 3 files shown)


src/plugins/inputs/who.js
```diff
@@ -0,0 +1,83 @@
+"use strict";
+const Msg = require("../../models/msg");
+
+exports.commands = ["who"];
+
+const parseWhoxResponse = (user, args) => {
+	const parseFieldNameForLounge = (field) => {
+		switch (field) {
+			case "channel":
+				return "channel"; // no-op for now
```
itsjohncs, line 10: This is the same thing that the default branch of this switch statement would do. What's the significance of having it separately called out here? Are you expecting it to not be a no-op in the future?

itsjohncs (commented):
Looks like there's no help material being added and there's no error messages we give back from the API (though maybe the server is doing some of that for us). Are you planning on doing another pass after this lands to add help text and error messages?

Though I'm also not sure I understand why support for `/who` is being added.

> I also added an uncaught exception handler on the server to stop a subset of crashes. The problem is people won't look at their server logs often, but bookworm and I agreed that's better than crashing.

I'll defer to you two. But TBH I'd prefer we continue to crash. I'd prefer failures make as much noise as possible especially with that huge TypeScript change looming.

src/plugins/inputs/who.js
MaxLeiter: I think I liked the explicitness at the time, but now 🤷

brunnre8 (changes requested):

PR #4548, opened 2022-04-27
Remove whois irc-event event handler, send whois/whowas response to current buffer

### LINKED ISSUE #492: `/whois nick` should go to currently active window

MaxLeiter:
We can remove the event handler now that we use the irc-framework callback similar to #4547 

Closes #492

Updated to Typescript/vue 3


brunnre8:
/whowas on a user without an entry doesn't send the message to the current channel, but to the server buffer and only in debug:

17:40 *** WHOWAS asdf
17:40 *** :copper.libera.chat 406 thisisatestthing asdf :There was no such nickname

src/plugins/inputs/whois.js
```diff
@@ -1,15 +1,39 @@
...
-		// This queries server of the other user and not of the current user, which
-		// does not know idle time.
-		// See http://superuser.com/a/272069/208074.
-		irc.raw("WHOIS", args[0], args[0]);
-	} else {
-		// Re-assembling the command parsed in client.js
-		irc.raw(`${cmd} ${args.join(" ")}`);
+	const client = this;
+	const target = args[0];
+	const targetNick = args[1] ? args[1] : target;
```
brunnre8, line 9: This I don't understand, args[0] should be the nick.
Why are you using args[1] ?

That made sense originally when we were querying the idle time, but we aren't doing that anymore.

brunnre8: Which probably brings me to the second point, should we add that back? Chances are yes right? Because we do want the idle time?

src/plugins/inputs/whois.js
```diff
@@ -1,15 +1,39 @@
...
+				client,
+				new Msg({
+					type: Msg.Type.WHOIS,
+					whois: data,
+				})
+			);
+		}
+	};
+
+	if (cmd === "whois") {
```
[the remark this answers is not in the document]
MaxLeiter: For two options prefer if-else but fine with changing it

src/plugins/inputs/whois.js
MaxLeiter: I kept the idle time code, but forgot to whois the nick and targetNick. Fixed.

brunnre8:
Considering that we always send the two value whois, there's no error reply when the nick is invalid, opened https://github.com/kiwiirc/irc-framework/pull/334 so that we can then add a proper handler

PR #4551, opened 2022-04-30
Disable connecting to the same network multiple times if lockNetwork is enabled

### LINKED ISSUE #4107: Server limit when in private mode
Greetings,

I've got thelounge running in private mode with "locknetwork" enabled. It seems though that users can still add additional servers/networks even though it's locked to the configured network. But it lets people join the network with multiple nicknames simultaneously. How would I go about limiting the server connections to just 1 server session per user account?

Thanks so much.

MaxLeiter:
Closes #4107

e connect form with the message 'You are you already connected and cannot connect again.' above the connect button

### DIFF
--- a/client/components/NetworkForm.vue
+++ b/client/components/NetworkForm.vue
@@ -400,6 +400,9 @@ the server tab on new connection"
 			</template>
 
 			<div>
+				<div v-if="disabledReason && disabled" class="disabled-reason">
+					{{ disabledReason }}
+				</div>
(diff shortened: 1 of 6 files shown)


GewoonYorick:
Could this be turned into a config option?

 I use locknetwork to only allow connections to my znc, however we do have multiple znc networks.

MaxLeiter:
I added an `allowMultipleSameHostConnections` but that seems long winded. Suggestions appreciated.

Issue #4078, opened 2020-09-15
UI: Allow hiding or collapsing of less used channels


ssbarnea:
### Feature Description

I am heavy IRC user and I am joined in >50 channels related to open-source and the channel left sidebar no longer fits a full screen, making quite hard to spot the channels that matters from those where I am only lurking (or waiting for someone to contact me, to to mention a keyword).

I am not yet sure which is the easiest solution for addressing this issue so I will list few ideas, letting other review them, hopefully we can find a solution for **scaling UI beyond ~15 channels**:

- **grouping channels** : be able to right click on a channel and select a pick a group. A group would be listed at root level with `[+]` icon and should show the total number of messages it contains. Selecting the group should expand it, clicking again on it should collapse it.
- **colors** : ability to change the color used to display channels, this would make easier to visually group channels. Color should be selectable using right click on channel, having a preset of ~6-10 colors should be more than enough.
- **auto-hide channel** : ability to hide channels without any activity in them. This can go at two levels: hide if there are no messages and hide if there are no mentions.

xPaw:
> auto-hide channel  [..] hide if there are no mentions.

This is already how collapsing works, click the arrow button next to the network name.

ssbarnea:
> This is already how collapsing works, click the arrow button next to the network name.

Yes, but we need collapse for groups of channels, so we de-clutter the UI.

venatiodecorus:
I like the way IRC Cloud handles it, where you can pin your favorite channels to the top of the list. I'd love to see that feature.

ssbarnea:
@xPaw Is there a way to hide specific channels? The option to collapse a network does not do any good if you are active on some channels on that network but not on others. Probably I could use two connections with separated nicks for freenode, one for my active channels and one for occasional ones, ones that I want to be joined into but where I do not expect to chat very often, and keep the second one collapsed. It would be a PITA and confuse people that discover I am using two nicks.

brunnre8:
@ssbarnea please don't ping random maintainers, it's rather rude.

Second, that's a support question and not specific to this bug per se, please use our IRC chan on freenode.

Hop in there and say hi (I'm bookworm)

You can do what you want with a single nick by using ZNC as long as TL doesn't natively support this.
Again, hop into IRC and I can explain the details

brettgilio:
It is really not the most elaborate work around, but in the mean time

```
div[data-name="#channel-name"] {
    display: none;
}
```

PR #4556, opened 2022-05-01
Add favoriting/pinning channels

### LINKED ISSUE #180: Favourite/Unified channels list
Would it be possible to have a Favorites list that we can be activated via a setting and when activated it can contain our favourite channels from various networks where tend to check the most?

Some people does have a lot of networks in their repository and it would be a drag to check the whole list whenever they want to check this and that.

MaxLeiter:
Closes #180, #4078(?) 

It effectively adds a new network, 'favorites' that can contain channels from any network.

If there's a name conflict, the server name is appended to the channel name while it's favorited.

The functionality all works except the command, as I'm not sure of how that should work.

I was thinking:

`/favorites add [target] [...targets]`
`/favorites remove [target] [...targets]`
`/favorites clear`

If no target is provided, it's assumed to be the current channel.

Here you can see two `#testing123` channels with the server name attached. It's also added to the topic line, so (for example) on mobile you can be confident which you pressed
A screenshot showing the Favorites 'network'

The Favorites 'network' can't be opened (clicking it does nothing), but does have a context menu:
The context menu has one option: 'clear favorites'

You can add channels/queries to your favorites with the context menu (or aforementioned command):
Context menues have an 'add to favorites' option now

### DIFF
--- a/client/components/Channel.vue
+++ b/client/components/Channel.vue
@@ -1,6 +1,6 @@
 <template>
 	<ChannelWrapper ref="wrapper" v-bind="$props">
-		<span class="name">{{ channel.name }}</span>
+		<span class="name">{{ name() }}</span>
 		<span
 			v-if="channel.unread"
 			:class="{highlight: channel.highlight && !channel.muted}"
@@ -50,6 +50,9 @@ export default {
 	methods: {
 		close() {
 			this.$root.closeChannel(this.channel);
+		},
+		name() {
+			return this.channel.displayName ? this.channel.displayName : this.channel.name;
 		},
 	},
 };
--- a/client/components/Channel.vue
+++ b/client/components/Channel.vue
@@ -15,13 +15,27 @@
 			>
 				<span class="parted-channel-icon" />
 			</span>
-			<span class="close-tooltip tooltipped tooltipped-w" aria-label="Leave">
-				<button class="close" aria-label="Leave" @click.stop="close" />
+			<span
+				class="close-tooltip tooltipped tooltipped-w"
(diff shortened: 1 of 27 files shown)


xPaw:
> The functionality all works except the command, as I'm not sure of how that should work.

Sounds like the channel object should know which network it belongs to, so it should just work™.

MaxLeiter:
@xPaw I meant the command interface:

> I was thinking:
> 
> /favorites add [target] [...targets]
> /favorites remove [target] [...targets]
> /favorites clear

xPaw:
Ah right, misunderstood.

The (un)favorite syntax would match all the existing commands.

igalic:
it's always possible to have those be aliases
some people may find the first more intuitive and others the second

i for instance add aliases in git for `tags`, `branches` and `remotes`, which by default print those in a sensible (to me) format

itsjohncs:
I'd give a vote for `unfavorite` and `favorite`. I think subcommands are nice as a way to do namespacing when there's a really huge number of commands (like `git`), but that's not the case here.

brunnre8:
Let's go for consistency and /favorite + /unfavorite then? Or what's the consensus now?

fnutt:
I agree about /favorite and /unfavorite. 👍  Good work @MaxLeiter!

fionera:
What is required to get this merged? I would really love to see this :)

FingerlessGlov3s:
Any update on this PR? Would love having this "favorites" feature.

sevofluranegiver:
Is there anything blocking this PR from going through?

PR #4557, opened 2022-05-01
Allow adding commands when creating a connection, support delays between commands

### LINKED ISSUE #1390: Enabling users to set delays between connect commands
One neat feature would be to be able to set a delay of a few seconds between two commands in order to give enough time to invite bots for example.

MaxLeiter:
I'm fine with only accepting one commit or neither if they're contentious or I'm missing something.

This allows you to add commands on the new network form if webirc isn't enabled. It also introduces a `/wait` command, so you can wait 10 seconds between executing commands with 

```
/wait 10
/msg bot command
``` 

Closes #1390

### DIFF
--- a/client/components/NetworkForm.vue
+++ b/client/components/NetworkForm.vue
@@ -252,7 +252,18 @@
 					placeholder="The Lounge - https://thelounge.chat"
 				/>
 			</div>
-			<template v-if="defaults.uuid && !$store.state.serverConfiguration.public">
+			<template v-if="!defaults.uuid">
+				<div class="connect-row">
+					<label for="connect:channels">Channels</label>
+					<input
+						id="connect:channels"
+						v-model="defaults.join"
+						class="input"
+						name="join"
+					/>
+				</div>
+			</template>
+			<template v-if="!$store.state.serverConfiguration.public">
 				<div class="connect-row">
 					<label for="connect:commands">
 						Commands
@@ -260,7 +271,7 @@
 							class="tooltipped tooltipped-ne tooltipped-no-delay"
 							aria-label="One /command per line.
 Each command will be executed in
-the server tab on new connection"
+the server tab on every connection."
 						>
 							<button class="extra-help" />
 						</span>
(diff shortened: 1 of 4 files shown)


src/plugins/irc-events/connection.js
```diff
@@ -39,15 +41,34 @@ module.exports = function (irc, network) {
...
+								if (delay > MAX_DELAY_LENGTH_SECONDS * 1000) {
+									delay = MAX_DELAY_LENGTH_SECONDS * 1000;
+								}
+							} catch {
+								delay = 1000;
+							}
+
+							resolve();
+						} else {
+							setTimeout(() => {
```
MaxLeiter, line 60: need to reset delay

brunnre8:
why would you want to prevent commands if webirc is enabled?
That should have nothing to do with each other no?

MaxLeiter:
@brunnre8 I did it based on the existing comment. @xPaw, I believe you wrote this code initially, do you know why you excluded commands?

src/plugins/irc-events/connection.js
```diff
@@ -39,15 +41,34 @@ module.exports = function (irc, network) {
...
-						target: network.channels[0].id,
-						text: cmd,
-					});
-				}, delay);
-				delay += 1000;
-			});
+			const commands = network.commands.map(
+				(cmd) =>
+					new Promise((resolve) => {
+						if (cmd.toLowerCase().startsWith("/wait")) {
```
xPaw, line 47:
```suggestion
						if (cmd.toLowerCase().startsWith("/wait ")) {
```

client/components/NetworkForm.vue
```diff
@@ -252,15 +252,30 @@
...
+					<label for="connect:channels">Channels</label>
+					<input
+						id="connect:channels"
+						v-model="defaults.join"
+						class="input"
+						name="join"
+					/>
+				</div>
+			</template>
+			<template v-if="!$store.state.serverConfiguration.public">
```
xPaw, line 266: Does this need a check for webirc too?

xPaw:
I can't remember the exact reason it was disabled, probably to not deal with public mode.

prenetic:
Just joined an instance of The Lounge and this feature is highly desirable. Quassel will provide delays for on-connect commands if needed (`/wait n`), which is the case with certain bots that don't always react in a timely fashion.

PR #4558, opened 2022-05-01
Config

brunnre8:
This is a smaller change than it looks like... all it does really is pull out the config from the Helper namespace... it doesn't belong there.


igalic:
funny how often Helper was used to mean Config.

MaxLeiter (approved):

PR #4560, opened 2022-05-03
Upgrade husky to v7

nemchik:


brunnre8:
Actually, I'd rather get rid of Husky all together.
https://github.com/thelounge/thelounge/issues/4452

Still, thank you for the effort

PR #4565, opened 2022-05-05
Cleanup configs (WIP)

nemchik:
I've labeled this as a WIP because it doesn't complete everything needed for the typescript branch to merge to master, but it's moving in the right direction. Some highlights:

* In eslint, define file extensions for ts and vue as overrides, and only apply the specific rules they need respectively. This works as intended for ts, but I have not yet fully figured out how to get it working for vue (because vue needs to be able to handle inline ts)
* Include a handful of ts specific eslint rules to be turned off globally until more issues can be resolved (tsc doesn't complain but eslint does with some things)
* Rearrange and clean up all tsconfig files (remove some duplication)
* Add some TODO comments (like eventually remove ...)

This should be able to merge into the typescript branch now and work can continue on the making typescript happy.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -2,12 +2,15 @@ module.exports = {
 	root: true,
 	overrides: [
 		{
-			files: ["**/*.ts"],
+			files: [
+				"**/*.ts",
+				// "**/*.vue"
+			],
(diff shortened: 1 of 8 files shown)


MaxLeiter:
LGTM. Thanks @nemchik

PR #4574, opened 2022-05-16
Fix yarn dev

nemchik:
~~This may significantly change how the webpack development config behaves. Instead of replacing `entry` and `plugins` this will append to the existing values. This makes the dev-server use the webpack development config (previously it used production).~~

Completely rewrote the changes here. Main thing is the mini css extract was needed in the dev webpack config mode.

Also added using `output.clean` https://webpack.js.org/configuration/output/#outputclean instead of removing files from the output directory manually.

### DIFF
--- a/src/plugins/dev-server.js
+++ b/src/plugins/dev-server.js
@@ -6,7 +6,7 @@ module.exports = (app) => {
 	log.debug("Starting server in development mode");
 
 	const webpack = require("webpack");
-	const webpackConfig = require("../../webpack.config.js")(undefined, {mode: "development"});
+	const webpackConfig = require("../../webpack.config.js")(undefined, {mode: "production"});
 
 	webpackConfig.plugins.push(new webpack.HotModuleReplacementPlugin());
 	webpackConfig.entry["js/bundle.js"].push(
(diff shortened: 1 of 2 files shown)


MaxLeiter:
Tests dont seem to be happy with this change

src/plugins/dev-server.js
```diff
@@ -6,7 +6,7 @@ module.exports = (app) => {
 	log.debug("Starting server in development mode");
 
 	const webpack = require("webpack");
-	const webpackConfig = require("../../webpack.config.js");
+	const webpackConfig = require("../../webpack.config.js")(undefined, {mode: "production"});
```
brunnre8, line 9: mode production?

nemchik: `src/plugins/dev-server.js` does in fact use the production webpack config. This is one commit before the cjs changes i PR'd https://github.com/thelounge/thelounge/blob/5f7acbf9948446289054527ee9385f938a116386/src/plugins/dev-server.js#L9 and browsing the repo at that commit shows there is a webpack.config-test.js but that's not what the dev-server is using.

MaxLeiter (approved):

Issue #4578, opened 2022-05-31
TypeScript / Vue 3 TODOs


MaxLeiter:
For after #4559 is merged

- The tests currently have a lot of `as any` casting. We should instead use a factory/constructor object (like `createConfigObject({ ...customArgs })`) and use [chai-subset](https://www.chaijs.com/plugins/chai-subset/) to verify its valid
- SQLite message storage returns an empty array if it's disabled, which leads to some weird typings.  It should probably return a custom object shaped like search results?
- ~In general, remove all `@ts-ignore`s~
- Remove all `eslint-disable-next-line`
- Remove all `require()` and replace with `import()`
- switchToChannel should be async
- continue #4586
- remove all @ts-expect-error's
- switch to Pinia (Vuex 5 rebranded)
- remove awayMessage (ref: https://github.com/thelounge/thelounge/pull/4559#discussion_r892022970)
- drop lodash
- Look into emitting types in the npm package
- all document.querySelector / getElementBy... should be replaced with refs
- type service worker (with JSDoc?)
- type eventbus similar to socketio (see `server/types/socket-events.d.ts`)
- I'll update this as I think of the rest....

brunnre8:
> SQLite message storage returns an empty array if it's disabled, which leads to some weird typings. It should probably return a custom object shaped like search results?

An empty array satisfies the type of an something[] just fine... not sure what's weird typing wise
You do have to specify the interface properly, but that's already fixed in my sqlite cleanup branch.

So I guess, that's solved?

PR #4586, opened 2022-06-04
Rearrange webpack dev/prod plugins

nemchik:
Dev and Prod both use the same common set of plugins. This moves the common set to the main config and then adds the production plugins when running in production mode.

Plugin order matters, so I've added the `NormalModuleReplacementPlugin` last at the very bottom.

This does move the `MiniCssExtractPlugin` above the `DefinePlugin`, but I am not sure if that is important.

Edit: This may not be worth chasing. I slapped it together kind of quick, and there's not a huge benefit to it. I just noticed that `VueLoaderPlugin`, `MiniCssExtractPlugin`, and `NormalModuleReplacementPlugin` are used in prod and dev, and all the rest are only used in prod. Feel free to close this, or I can put more effort into making it work correctly if needed.

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -96,12 +96,6 @@ const config: webpack.Configuration = {
 		new MiniCssExtractPlugin({
 			filename: "css/style.css",
 		}),
-
-		// Client tests that require Vue may end up requireing socket.io
-		new webpack.NormalModuleReplacementPlugin(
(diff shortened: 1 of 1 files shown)


MaxLeiter:
Thanks for the effort but I think we can hold off on this until after the merge. Added to #4578 4586

PR #4590, opened 2022-06-15
Use correct option name (filter vs ldapFilter) in config.js comment

murph:
The comment for the "filter" key under "searchDN" refers to it as "ldapFilter" instead of "filter".


MaxLeiter:
Thank you!

Issue #4593, opened 2022-06-21
Add user is broken.


brettgilio:
See https://github.com/thelounge/thelounge/pull/4592

xPaw:
The fix is to just specify `.default` like so:
```
const ClientManager = require("../../clientManager").default;
```

Ideally of course we wouldn't be using requires.

PR #4595, opened 2022-06-22
server/command-line: remove most instances of require

MaxLeiter:
follow-up to #4594 / #4593


xPaw:
Looks fine. Did you verify that moving clientManager import up there doesn't mess with any of the import order?

I think another reason it was require inside of a function is so that it didn't load all the server code recursively when the command isn't used.

MaxLeiter:
I'll come back to this eventually

PR #4598, opened 2022-06-24
Fix regex escape for prefix patterns

brunnre8:
Our regex escape function escapes proper regexes, however
it isn't meant to be shoved into a char class via string interpolation.

We need to also escape '-' if we do so.


MaxLeiter:
Looks good. Just commenting to leave proof we discussed adding tests / checking code-base for other instances of this on IRC.

brunnre8:
added the test, couldn't find other instances

MaxLeiter (approved):

PR #4600, opened 2022-06-26
Remove uploading event listeners on ChatInput unmount

MaxLeiter:
Not too sure how this worked in Vue 2, but we weren't removing the event listeners in Upload so you could cause an error pasting on a not-chat page.

### DIFF
--- a/client/js/upload.ts
+++ b/client/js/upload.ts
@@ -11,12 +11,18 @@ class Uploader {
 	overlay: HTMLDivElement | null = null;
 	uploadProgressbar: HTMLSpanElement | null = null;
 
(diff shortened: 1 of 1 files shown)


client/js/upload.ts
```diff
@@ -11,12 +11,18 @@ class Uploader {
 	overlay: HTMLDivElement | null = null;
 	uploadProgressbar: HTMLSpanElement | null = null;
 
+	onDragEnter = (e: DragEvent) => this.dragEnter(e);
```
MaxLeiter, line 14: we have to pass a reference as the second arg to removeEventListener for it to work properly.

PR #4616, opened 2022-08-06
Respect bind setting for all outgoing requests

brunnre8:
So far the bind config only impacted the IRC connections.
However, nothing in our doc comment says that this is intentional.

> ### bind
> Set the local IP to bind to for outgoing connections.

This commit fixes the leak and uses it for all outgoing requests
as described by the docstring.

### DIFF
--- a/server/plugins/changelog.ts
+++ b/server/plugins/changelog.ts
@@ -3,6 +3,7 @@ import colors from "chalk";
 import log from "../log";
 import pkg from "../../package.json";
 import ClientManager from "../clientManager";
+import Config from "../config";
 
 const TIME_TO_LIVE = 15 * 60 * 1000; // 15 minutes, in milliseconds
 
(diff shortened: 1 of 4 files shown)


MaxLeiter (approved):

brunnre8:
Don't merge this, it makes node 18 reproducibly fail.
I need to test that locally.

brunnre8:
@MaxLeiter this was broken because our tests were trying to use the ipv6 address to listen on, while the bind address forced ipv4.

So let's fix the test harness instead.

Btw: the tests are all horribly broken, they don't actually work in most failure cases.

Throwing from the event listener is not doing what one would hope

```
		this.irc.once("msg:preview", function (data) {
			expect(data.preview.type).to.equal("link");
			expect(data.preview.head).to.equal("Untitled page");
			expect(data.preview.body).to.equal(loremIpsum.substring(0, 300));
			expect(data.preview.body).to.have.length(300);
			expect(data.preview.link).to.equal(url);

			expect(message.previews).to.deep.equal([data.preview]);
			done();
		});
```
All that does if any of the expects throw is to reject the promise inside of the link() call. It never actually reaches mocca and hence done isn't called either which leads to all kinds of time outs rather then the actual test failure

brunnre8:
Same goes for anything in an `app.get(...)` callback.
Express swallows the thrown exceptions and mocca never sees that either

PR #4617, opened 2022-08-06
linkPreviews: Enforce TLS validity

brunnre8:
When a URL is prefixed with a TLS scheme, we should make sure
that the remote provides a valid cert, even just for prefetches.
Else MITM of such a site is trivial.

This probably breaks some people with self signed cert, but the
age where that was acceptable is past. We have free CAs now like
Let's Encrypt.


MaxLeiter (approved):

MaxLeiter:
@bookworm merge when you want 👍

PR #4618, opened 2022-08-11
Initial work on adding an events API to plugins

antoniomika:
Out of a conversation on IRC, we talked about the potential benefits of an event API to be used for plugins (and potentially other places within thelounge).

This PR includes some initial work in adding an EventEmitter to be used by plugins. This is done at a high level this way:
1. Create an `EventEmitter` on the `Client` objects that is used for broadcasting messages about a specific client
    - The `EventEmitter` is typed, the definitions in `./servers/events` includes the interface that are used to define
       what events are emitted and their arguments.
2. Create an `EventEmitter` on the `packageApi` that is stored by the packages handler
3. For APIs that should be emitted to the package, define a method that is used to iterate through package emitters and 
    broadcast the message
    - An event listener needs to be added to the client definition to emit the messages using the defined method.

This is just an initial crude implementation out of our conversation (@brunnre8 on IRC). I tested it with a simple plugin that does this:

```javascript
module.exports = {
    onServerStart: api => {
        api.Events.on('message', console.log);
        api.Events.on('notification', console.log);
    },
};
```

### DIFF
--- a/server/client.ts
+++ b/server/client.ts
@@ -13,11 +13,13 @@ import constants from "../client/js/constants";
 
 import inputs from "./plugins/inputs";
 import PublicClient from "./plugins/packages/publicClient";
+import packages from "./plugins/packages/index";
 import SqliteMessageStorage from "./plugins/messageStorage/sqlite";
 import TextFileMessageStorage from "./plugins/messageStorage/text";
 import Network, {IgnoreListItem, NetworkWithIrcFramework} from "./models/network";
 import ClientManager from "./clientManager";
 import {MessageStorage, SearchQuery} from "./plugins/messageStorage/types";
+import {ClientEmitter, ClientEvents} from "./events";
 
 type OrderItem = Chan["id"] | Network["uuid"];
 type Order = OrderItem[];
@@ -89,6 +91,11 @@ export type Mention = {
 	from: UserInMessage;
 };
 
+const packageEvents = new Map<keyof ClientEvents, any>([
+	["message", packages.emitMessage],
+	["notification", packages.emitNotification],
+]);
+
 class Client {
 	awayMessage!: string;
 	lastActiveChannel!: number;
@@ -109,6 +116,7 @@ class Client {
 	highlightRegex!: RegExp | null;
 	highlightExceptionRegex!: RegExp | null;
 	messageProvider?: SqliteMessageStorage;
+	events!: ClientEmitter;
 
 	fileHash!: string;
 
@@ -129,9 +137,14 @@ class Client {
 			highlightRegex: null,
 			highlightExceptionRegex: null,
 			messageProvider: undefined,
+			events: new ClientEmitter(),
 		});
 
 		const client = this;
+
+		packageEvents.forEach((listener, event) => {
+			client.events.on(event, listener);
+		});
 
 		client.config.log = Boolean(client.config.log);
 		client.config.password = String(client.config.password);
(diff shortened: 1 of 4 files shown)


server/client.ts
```diff
@@ -129,10 +137,15 @@ class Client {
 			highlightRegex: null,
 			highlightExceptionRegex: null,
 			messageProvider: undefined,
+			events: new ClientEmitter(),
 		});
 
 		const client = this;
 
+		for (const event in packageEvents) {
+			client.events.on(event as keyof ClientEvents, packageEvents[event]);
```
MaxLeiter, line 146: Can you move the `as` cast up to the packageEvents declaration?

server/events.ts
```diff
@@ -0,0 +1,22 @@
...
+		network: NetworkWithIrcFramework,
+		client: Client,
+		data: MessageData
+	) => void;
+	notification: (
+		irc: NetworkWithIrcFramework["irc"],
+		network: NetworkWithIrcFramework,
+		client: Client,
+		data: MessageData,
+		notification: any
```
MaxLeiter, line 18: Can we type Notification? Is it the same as `notificationData` in irc-events/message.ts?

package.json
```diff
@@ -65,6 +65,7 @@
     "read-chunk": "3.2.0",
     "semver": "7.3.5",
     "socket.io": "4.4.1",
+    "tiny-typed-emitter": "2.1.0",
```
MaxLeiter, line 68: This seems like a fairly small library. We try and keep our dependencies to a minimum; is there a way around this? Can we manually type just the features we use?

MaxLeiter (changes requested):

MaxLeiter:
thanks @antoniomika! I've gave it a quick glance-over, but I'd like @bookworm to 👍 it before we merge

server/client.ts
antoniomika: Changed this to be a typed map to take care of this

server/events.ts
antoniomika: Done

package.json
antoniomika: Typed just `on` and `emit` for now.

server/client.ts
```diff
@@ -89,6 +91,11 @@ export type Mention = {
 	from: UserInMessage;
 };
 
+const packageEvents = new Map<keyof ClientEvents, any>([
```
MaxLeiter, line 94: 👌

server/events.ts
```diff
@@ -0,0 +1,26 @@
...
+		client: Client,
+		data: MessageData
+	) => void;
+	notification: (
+		irc: NetworkWithIrcFramework["irc"],
+		network: NetworkWithIrcFramework,
+		client: Client,
+		data: MessageData,
+		notification: NotificationData
+	) => void;
```
brunnre8, line 18: The event structure is fine, but in my opinion I really don't want to expose the irc framework nor any other internal data structures.

Shouldn't we hide that all behind a public API Interface?

If you expose it like that our internals become public API and we can't change it any longer and are exposing the whole of the framework API which we then need to carry should we ever wrap that deeper for dome reason.

Our current API has the same problem and this might be a me thing, but it seems wrong layer wise.

antoniomika: Agreed. I wasn't exactly sure what the best way to handle that would be (and I was just sticking with what current command events fire).

Maybe we should discuss what we're interested in the client exposing and I can stub out a client interface?

PR #4619, opened 2022-08-12
Added client type checking to webpack

antoniomika:
This PR includes changes to add type checking (via TypeScript) for the client-side code base. webpack does not currently report TypeScript errors (which poses an issue).

This was done out of a message on IRC via @MaxLeiter.

The changes here are broken up between two commits, one to enable type checking and another to fix the issues the compiler has found.

### DIFF
--- a/client/js/helpers/collapseNetwork.ts
+++ b/client/js/helpers/collapseNetwork.ts
@@ -1,7 +1,9 @@
 import storage from "../localStorage";
 
 export default (network, isCollapsed) => {
-	const networks = new Set(JSON.parse(storage.get("thelounge.networks.collapsed")));
+	const stored = storage.get("thelounge.networks.collapsed");
+	const networks = stored ? new Set(JSON.parse(stored)) : new Set();
+
 	network.isCollapsed = isCollapsed;
 
 	if (isCollapsed) {
--- a/client/js/helpers/contextMenu.ts
+++ b/client/js/helpers/contextMenu.ts
@@ -250,10 +250,15 @@ export function generateInlineChannelContextMenu(
 			switchToChannel(channel);
 		}
 
-		socket.emit("input", {
-			target: store.state.activeChannel.channel.id,
-			text: "/join " + chan,
-		});
+		if (store.state.activeChannel) {
+			socket.emit("input", {
+				target: store.state.activeChannel.channel.id,
+				text: "/join " + chan,
+			});
+		} else {
+			// eslint-disable-next-line no-console
+			console.error("Unable to join channel: activeChannel is undefined");
+		}
 	};
 
 	const channel = network.channels.find((c) => c.name === chan);
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -6,9 +6,38 @@
   "files": [
     "../package.json",
     "../server/types/socket-events.d.ts",
+    "../server/helper.ts",
+    "../server/log.ts",
+    "../server/config.ts",
+    "../server/client.ts",
+    "../server/clientManager.ts",
+    "../server/identification.ts",
+    "../server/plugins/changelog.ts",
+    "../server/plugins/uploader.ts",
+    "../server/plugins/storage.ts",
+    "../server/plugins/inputs/index.ts",
+    "../server/plugins/messageStorage/sqlite.ts",
+    "../server/plugins/messageStorage/text.ts",
+    "../server/plugins/packages/index.ts",
+    "../server/plugins/packages/publicClient.ts",
+    "../server/plugins/packages/themes.ts",
+    "../server/plugins/dev-server.ts",
+    "../server/plugins/webpush.ts",
+    "../server/plugins/sts.ts",
+    "../server/plugins/clientCertificate.ts",
+    "../server/plugins/auth.ts",
+    "../server/plugins/auth/local.ts",
+    "../server/plugins/auth/ldap.ts",
+    "../server/plugins/irc-events/link.ts",
+    "../server/command-line/utils.ts",
+    "../server/models/network.ts",
+    "../server/models/user.ts",
+    "../server/models/msg.ts",
+    "../server/models/prefix.ts",
     "./js/helpers/fullnamemap.json",
     "./js/helpers/simplemap.json",
-    "../server/helper.ts"
+    "../webpack.config.ts",
+    "../babel.config.cjs"
   ] /* If no 'files' or 'include' property is present in a tsconfig.json, the compiler defaults to including all files in the containing directory and subdirectories except those specified by 'exclude'. When a 'files' property is specified, only those files and those specified by 'include' are included. */,
   // "exclude": [],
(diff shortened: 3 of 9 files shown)


client/tsconfig.json
```diff
@@ -6,9 +6,37 @@
...
+    "../server/plugins/auth.ts",
+    "../server/plugins/auth/local.ts",
+    "../server/plugins/auth/ldap.ts",
+    "../server/plugins/irc-events/link.ts",
+    "../server/command-line/utils.ts",
+    "../server/models/network.ts",
+    "../server/models/user.ts",
+    "../server/models/msg.ts",
+    "../server/models/prefix.ts",
     "./js/helpers/fullnamemap.json",
```
brunnre8, line 36: this works, however that includes way more than I expected.
Why do we need to include files which are purely server side? sqlite for instance is never actually imported in the client.

Is this due to transitive deps from some of the actual necessary imports or are we compiling the whole world and not just the client?

(Generally curious, I assume you had your reasons, so don't take this as a complaint)

server/client.ts
```diff
@@ -337,6 +337,7 @@ class Client {
 		// TODO
 		// eslint-disable-next-line @typescript-eslint/no-misused-promises
 		events.forEach(async (plugin) => {
+			// @ts-ignore
```
brunnre8, line 340: > Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node16', or 'nodenext'.

Why is this safe to ignore?
At the very least it needs a comment as to why that's ok :wink: better yet, solving the actual root problem I guess.

server/plugins/auth.ts
```diff
@@ -13,6 +13,7 @@ export type AuthHandler = (
 
 // The order defines priority: the first available plugin is used.
 // Always keep 'local' auth plugin at the end of the list; it should always be enabled.
+// @ts-ignore
```
brunnre8, line 16: Same thing here, needs a comment as to why it's safe to ignore or a fix of the root cause

server/plugins/inputs/index.ts
```diff
@@ -60,6 +60,7 @@ const builtInInputs = [
 ];
 
 for (const input of builtInInputs) {
+	// @ts-ignore
```
brunnre8, line 63: Same thing here, needs a comment as to why it's safe to ignore or a fix of the root cause

server/server.ts
```diff
@@ -86,6 +86,7 @@ export default async function (
 	const app = express();
 
 	if (options.dev) {
+		// @ts-ignore
```
brunnre8, line 89: Same thing here, needs a comment as to why it's safe to ignore or a fix of the root cause

server/server.ts
```diff
@@ -141,6 +142,7 @@ export default async function (
 	}
 
 	if (!Config.values.https.enable) {
+		// @ts-ignore
```
brunnre8, line 145: ... you get the gist by now I think :smile: 
I'll stop dumping those for the cases further down the list, but all of them have the same issue

brunnre8 (changes requested):
Thanks a bunch for your help here, build system work is not particularly fun to do so that's very much appreciated.

There are some areas where we should be a bit more careful with the error handling though, rather than just asserting things we can't possibly guarantee.

Mind fixing those up?

There's also some questions out of interest.

server/client.ts
antoniomika: The client is being built for `es2015`. I didn't really want to drop support for old clients especially as these dynamic imports only happen in the server side app, and are just included due to transitive dependencies. 

Let me see what I can make happen with moving around some code.

client/tsconfig.json
antoniomika: Of course, that's alright! 

And correct, it's due to the transitive deps from code sharing. This is all a result of the dependency import tree being a bit wonky between client/server side code sharing. Each of these were an import used in the TS project (and the compiler would fail if they weren't provided to tsconfig). I'm going to try and move the code around to limit this list (and therefore fix some of the dynamic import errors).

client/js/helpers/contextMenu.ts
```diff
@@ -251,7 +251,7 @@ export function generateInlineChannelContextMenu(
 		}
 
 		socket.emit("input", {
-			target: store.state.activeChannel.channel.id,
+			target: store.state.activeChannel!.channel.id,
```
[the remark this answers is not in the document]
antoniomika: That would actually still fail as the target could then be undefined. Instead, we'll check that activeChannel is set before attempting to emit an input.

client/js/helpers/collapseNetwork.ts
```diff
@@ -1,7 +1,7 @@
 import storage from "../localStorage";
 
 export default (network, isCollapsed) => {
-	const networks = new Set(JSON.parse(storage.get("thelounge.networks.collapsed")));
+	const networks = new Set(JSON.parse(storage.get("thelounge.networks.collapsed")!));
```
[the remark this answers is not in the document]
antoniomika: I actually wasn't sure what the correct functionality here was supposed to be, so I decided to try and stick with the logic as closely as possible.

Though `!` does assert a none-null/undefined value this is entirely within typescript, meaning it is removed once compiled into javascript and doesn't have an effect, ergo doing what the code was doing previously. We'll make this be an empty set definition instead based on whether the stored value is retrieved though.

client/js/helpers/contextMenu.ts
```diff
@@ -250,10 +250,12 @@ export function generateInlineChannelContextMenu(
...
-			target: store.state.activeChannel.channel.id,
-			text: "/join " + chan,
-		});
+		if (store.state.activeChannel) {
+			socket.emit("input", {
+				target: store.state.activeChannel.channel.id,
+				text: "/join " + chan,
+			});
+		}
 	};
```
brunnre8, line 259: how about some console.error or some such in the else path so that we notice something is off?

client/js/helpers/collapseNetwork.ts
```diff
@@ -1,7 +1,9 @@
 import storage from "../localStorage";
 
 export default (network, isCollapsed) => {
-	const networks = new Set(JSON.parse(storage.get("thelounge.networks.collapsed")));
+	const stored = storage.get("thelounge.networks.collapsed");
+	const networks = stored ? new Set(JSON.parse(stored)) : new Set();
```
MaxLeiter, line 5: new Set<ClientNetwork>?

MaxLeiter (changes requested):

MaxLeiter:
Thank you! This seems to work great. I left one small comment and then I'm okay with merging.

client/js/helpers/collapseNetwork.ts
antoniomika: `networks` here looks to be an array of uuid's (string), isn't it?

PR #4622, opened 2022-08-27
Sqlite cleanup + convert to async

brunnre8:
Tried to clean up the sqlite storage a bit... while working on the msgid thing.

There are no functional changes, other than the fact that the whole message storage interface is now async (has to be, else we can't wait on migrations and so forth)

### DIFF
--- a/server/helper.ts
+++ b/server/helper.ts
@@ -23,6 +23,7 @@ const Helper = {
 	parseHostmask,
 	compareHostmask,
 	compareWithWildcard,
+	catch_to_error,
 
 	password: {
 		hash: passwordHash,
@@ -182,4 +183,18 @@ function compareWithWildcard(a: string, b: string) {
 	const user_regex = wildmany_split.join(".*");
 	const re = new RegExp(`^${user_regex}$`, "i"); // case insensitive
 	return re.test(b);
+}
+
+function catch_to_error(prefix: string, err: any): Error {
+	let msg: string;
+
+	if (err instanceof Error) {
+		msg = err.message;
+	} else if (typeof err === "string") {
+		msg = err;
+	} else {
+		msg = err.toString();
+	}
+
+	return new Error(`${prefix}: ${msg}`);
 }
--- a/server/plugins/messageStorage/sqlite.ts
+++ b/server/plugins/messageStorage/sqlite.ts
@@ -195,10 +195,10 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
 		}) as Message[];
 	}
 
-	search(query: SearchQuery): Promise<SearchResponse> {
+	async search(query: SearchQuery): Promise<SearchResponse> {
 		if (!this.isEnabled) {
 			// this should never be hit as messageProvider is checked in client.search()
-			return Promise.reject(
+			throw new Error(
 				"search called but sqlite provider not enabled. This is a programming error"
 			);
 		}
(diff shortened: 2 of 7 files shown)


server/plugins/messageStorage/sqlite.ts
```diff
@@ -243,31 +227,59 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
...
+	canProvideMessages() {
+		return this.isEnabled;
+	}
+
+	private run(stmt: string, ...params: any[]) {
+		this.serialize_run(stmt, params).catch((err) =>
+			log.error(`failed to run ${stmt}`, String(err))
+		);
+	}
+
```
itsjohncs, line 251: Might as well embrace the `async` way fully:

```typescript
private async run(stmt: string, ...params: any[]) {
    try {
        await this.serialize_run(stmt, params)
    } catch (err) {
        log.error(`failed to run ${stmt}`, String(err));
    );
}
```

server/plugins/messageStorage/sqlite.ts
```diff
@@ -243,31 +227,59 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
...
+	private serialize_run(stmt: string, params: any[]): Promise<void> {
 		return new Promise((resolve, reject) => {
-			this.database.all(select, params, (err, rows) => {
-				if (err) {
+			this.database.serialize(() => {
+				this.database.run(stmt, params, (err) => {
+					if (!err) {
```
itsjohncs, line 256: I think this could be a little simpler.

```typescript
if (err) {
    reject(err);
} else {
    resolve();
}
```

server/plugins/messageStorage/sqlite.ts
```diff
@@ -162,61 +155,52 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
...
+			return newMsg;
+		}) as Message[];
 	}
 
-	search(query: SearchQuery): Promise<SearchResponse | []> {
+	async search(query: SearchQuery): Promise<SearchResponse> {
 		if (!this.isEnabled) {
 			// this should never be hit as messageProvider is checked in client.search()
-			return Promise.resolve([]);
+			throw new Error(
```
itsjohncs, line 201: I'm nearly always a fan of making failures louder, this seems much better.

itsjohncs (approved):
LGTM! I didn't pull it down and test it out, but I did a pretty thorough comb through of the code. Including trying to track down what exactly `serialize` is doing ([docs](https://github.com/TryGhost/node-sqlite3/wiki/Control-Flow)).

server/plugins/messageStorage/sqlite.ts
brunnre8: yes. That went the way of the Dodo since... just didn't push the new changes yet.

server/plugins/messageStorage/sqlite.ts
brunnre8: ack, patched in the patchset that will succeed this one

brunnre8:
mkey, rewrote the thing, PTAL.

This is not getting any better, so I'd like to avoid adding more things on top until that's figured out

brunnre8:
@itsjohncs please remove the approval / re-ack the PR, I'll ignore it until you actually looked over the changes.
GH doesn't seem to auto revoke the approvals apparently, even though new stuff is pushed

server/helper.ts
```diff
@@ -183,3 +184,17 @@ function compareWithWildcard(a: string, b: string) {
 	const re = new RegExp(`^${user_regex}$`, "i"); // case insensitive
 	return re.test(b);
 }
+
+function catch_to_error(prefix: string, err: any): Error {
```
brunnre8, line 188: That thing is bonkers, but TS yells at me if I shove the error in a string literal from a promise catch block... as that doesn't guarantee that I'm actually looking at an Error (the interface)...

So you need to type assert this and I couldn't figure out how to do it without the code in this function... and I'm not going to inline this in every catch block

itsjohncs (changes requested):
> please remove the approval / re-ack the PR

dgw:
> GH doesn't seem to auto revoke the approvals apparently, even though new stuff is pushed

This is available for branches with Branch Protection turned on. The relevant checkbox is: "Optionally, to dismiss a pull request approval review when a code-modifying commit is pushed to the branch, select **Dismiss stale pull request approvals when new commits are pushed.**"

[More info](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule)

brunnre8:
@itsjohncs / @MaxLeiter mind reviewing this?
I'd like to merge it down with the search fixes.

MaxLeiter (approved):

Issue #4637, opened 2022-11-01
Search needs to clear the store


brunnre8:
- _The Lounge version:_ master branch

Currently when we search we add the results to the store.
However this is never cleared again, leading to funny artifacts

client/js/socket-events/search.ts:
```ts
socket.on("search:results", (response) => {
	store.commit("messageSearchInProgress", false);

	if (store.state.messageSearchResults) {
		store.commit("addMessageSearchResults", response);
		return;
	}

	store.commit("messageSearchResults", response);
});
```

After the first search is closed out, state.messageSearchResults is still non empty, meaning a second search will try to add to it, which makes no sense as it may not even be related.

/cc @MaxLeiter that's your bug :stuck_out_tongue_winking_eye:

PR #4640, opened 2022-11-12
Search bug fixes

brunnre8:
While fixing #4637 I noticed some other things that could be cleaned up, plus an off by one bug in our search code.

So here's it all, please tell me if you want to split this into individual PRs.

### DIFF
--- a/server/plugins/messageStorage/sqlite.ts
+++ b/server/plugins/messageStorage/sqlite.ts
@@ -224,7 +224,7 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
 
 		let select =
 			'SELECT msg, type, time, network, channel FROM messages WHERE type = "message" AND json_extract(msg, "$.text") LIKE ? ESCAPE \'@\'';
-		const params = [`%${escapedSearchTerm}%`];
+		const params: any[] = [`%${escapedSearchTerm}%`];
 
 		if (query.networkUuid) {
 			select += " AND network = ? ";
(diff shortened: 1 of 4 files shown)


server/plugins/messageStorage/sqlite.ts
```diff
@@ -240,8 +240,8 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
 
 		select += " ORDER BY time DESC LIMIT ? OFFSET ? ";
 		params.push(maxResults.toString());
-		query.offset = parseInt(query.offset as string, 10) || 0;
-		params.push(String(query.offset));
```
MaxLeiter, line 244: If you assign `String(query.offset)` to something above we can avoid the `toString/String()/as string`s

MaxLeiter (approved):

server/plugins/messageStorage/sqlite.ts
brunnre8: what do you mean exactly? Not sure I follow.

MaxLeiter: `const offset = String(query.offset)` and use `offset` instead of `query.offset as string` and the rest

brunnre8: Probably still need to check for undefined / null (although not if I look a the callers, but you know how it goes) and then it'd be String(query.offset || 0) which doesn't seem to be more readable to me than the above.

Don't have any strong opinions, but why do you prefer the conversion over toString() ?

But I don't know

**A)**
```typescript
const offset = query.offset || 0;
params.push(offset.toString());
```
and **B)**
```typescript
const offset = query.offset || 0;
params.push(String(offset));
```

Does exactly the same, I don't favor one over the other so I picked the same that was already there for the limit clause.

Shall I rewrite it to the B) version?

brunnre8: Or we have **C)**
```typescript
const offset = ;
params.push(String(query.offset || 0));
```
Gets a tad busy but avoids the useless temporary var

brunnre8: actually... TS prevents us from setting it to not a number doesn't it? So ignore it, I don't need to do the or 0 dance at all

brunnre8:
Got rid of the string conversion altogether. Needed some changes to the test as they were violating the type.

PTAL

MaxLeiter (approved):

PR #4642, opened 2022-11-21
remove VueApp from router

brunnre8:
Nothing actually depends on the vue app being monkey patched onto the router, so let's get rid of it.


MaxLeiter:
This was originally in place because of the image previews, but I was able to work around it.

MaxLeiter (approved):

PR #4646, opened 2022-11-26
sqlite: synchronize enable() internally

brunnre8:
TL is stupid and doesn't wait for message{Provider,Storage} to settle before it starts using the store.

While this should be fixed globally, we can hack around the problem by pushing everything onto the call stack and hope that we'll eventually finish the setup before we blow the stack.

Now, how do we want to handle this?
Clearly the API design that was done initially doesn't cut it, but the changes will be rather high as we need to switch to an async model.

Do we merge this workaround for now, or do we roll back?
The problem with rolling back is that we can't do any migrations to our DB whatsoever

### DIFF
--- a/server/plugins/messageStorage/sqlite.ts
+++ b/server/plugins/messageStorage/sqlite.ts
@@ -38,17 +38,30 @@ const schema = [
 	"CREATE INDEX IF NOT EXISTS time ON messages (time)",
 ];
 
+class Deferred {
+	resolve!: () => void;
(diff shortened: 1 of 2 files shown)


brunnre8:
I removed the reject on deferred as it's not needed and make sure that we unlock on failure too.

What do you think?

MaxLeiter (approved):

MaxLeiter:
I hate that this is necessary

PR #4647, opened 2022-11-26
search: ignore searchResults if it isn't the active query

brunnre8:
Prior to this, the search is still racy but one tends to notice this only when the DB is large or network is involved. The user can initiate a search, get bored, navigate to another chan issue a different search.

Now however, the results of the first search come back in and hilarity ensues as we are now confused with the state.

To avoid this, keep track of the last search done and any result that comes in that isn't equal to the active query is garbage and can be dropped.

### DIFF
--- a/client/components/Windows/SearchResults.vue
+++ b/client/components/Windows/SearchResults.vue
@@ -33,18 +33,19 @@
 							<button
 								ref="loadMoreButton"
 								:disabled="
-									store.state.messageSearchInProgress || !store.state.isConnected
+									!!store.state.messageSearchPendingQuery ||
+									!store.state.isConnected
 								"
 								class="btn"
 								@click="onShowMoreClick"
 							>
-								<span v-if="store.state.messageSearchInProgress">Loading…</span>
+								<span v-if="store.state.messageSearchPendingQuery">Loading…</span>
 								<span v-else>Show older messages</span>
 							</button>
 						</div>
 
 						<div
-							v-if="store.state.messageSearchInProgress && !offset"
+							v-if="store.state.messageSearchPendingQuery && !offset"
 							class="search-status"
 						>
 							Searching…
@@ -105,6 +106,7 @@ import type {ClientMessage} from "../../js/types";
 import {useStore} from "../../js/store";
 import {useRoute, useRouter} from "vue-router";
 import {switchToChannel} from "../../js/router";
+import {SearchQuery} from "../../../server/plugins/messageStorage/types";
 
 export default defineComponent({
 	name: "SearchResults",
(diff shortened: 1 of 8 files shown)


brunnre8:
@MaxLeiter quck question, how do you want to handle it when types are defined in the server but the client wants them?
Do we duplicate the type on the client?
Move the type to some common folder?

That's the one thing which still makes this a draft (plus testing on something where you actually notice the race, will use my own personal instance)

client/components/Windows/SearchResults.vue
```diff
@@ -106,6 +107,7 @@ import type {ClientMessage} from "../../js/types";
 import {useStore} from "../../js/store";
 import {useRoute, useRouter} from "vue-router";
 import {switchToChannel} from "../../js/router";
+import {SearchQuery} from "../../../server/plugins/messageStorage/types";
```
brunnre8, line 110: here we need the searchquery type

server/plugins/messageStorage/sqlite.ts
```diff
@@ -230,15 +230,10 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
 		params.push(query.offset);
 
 		const rows = await this.serialize_fetchall(select, ...params);
-		const response: SearchResponse = {
-			searchTerm: query.searchTerm,
-			target: query.channelName,
```
brunnre8, line 235: we renamed this for absolutely no reason...
Hence I propose that we get rid of the mental gymnastics involved and just call it what it really is

server/plugins/messageStorage/types.d.ts
```diff
@@ -29,12 +29,9 @@ export type SearchQuery = {
...
 };
 
-export type SearchResponse =
-	| Omit<SearchQuery, "channelName" | "offset"> & {
-			results: Message[];
-			target: string;
-			offset: number;
-	  };
+export type SearchResponse = SearchQuery & {
+	results: Message[];
```
brunnre8, line 33: if we stop renaming things, this gets much simpler

server/server.ts
```diff
@@ -760,9 +760,8 @@ function initializeClient(
 		});
 
 		socket.on("search", async (query) => {
-			await client.search(query).then((results) => {
-				socket.emit("search:results", results);
-			});
+			const results = await client.search(query);
```
brunnre8, line 763: This function seems to be confused, it's already async but then uses promises as well... simplify

server/types/socket-events.d.ts
```diff
@@ -107,7 +107,7 @@ interface ServerToClientEvents {
 		token: string;
 	}) => void;
 
-	"search:results": (response: {results: ClientMessage[]}) => void;
+	"search:results": (response: SearchResponse) => void;
```
brunnre8, line 110: I don't see why you'd already drop the type information here, we are still on the server side so having more info is good no?

brunnre8:
@MaxLeiter mind having a look at my comments / questions above?

MaxLeiter:
You should be able to import a type across the client / server, just choose one to define it in.

server/plugins/messageStorage/sqlite.ts
MaxLeiter: Pretty sure I copied this all over during the typescript transition. Feel free to improve it

brunnre8:
Well, in that case, please have a look.
I've rebased this onto master

MaxLeiter (approved):
Tested searching and seems good.  Thank you for the type cleanup.

PR #4649, opened 2022-11-27
style: Put user colors into the smallest possible scope

brunnre8:
The only thing that cares about user colors is the user component. Putting a class value on the chat component seems to be the wrong place.

This also allows us to remove various css selectors so that we don't need to be that specific.
After all whatever has that class needs to be colored, we don't care where it is.

### DIFF
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -4,7 +4,6 @@
 			id="chat"
 			:class="{
 				'hide-motd': store.state.settings.motd,
-				'colored-nicks': store.state.settings.coloredNicks,
 				'time-seconds': store.state.settings.showSeconds,
 				'time-12h': store.state.settings.use12hClock,
 			}"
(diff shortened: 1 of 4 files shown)


MaxLeiter:
Are we using this CSS in multiple places or we can move it into a component?

brunnre8:
theoretically component,  but I really don't want to put colors there.

It should be easy for themes to be created and to override colors, as is we are using way to much colors in way to deep selectors.

My actual goal is to remove all colors from the default style and just have layout there 

All colors and fancy styles should be pushed into the themes

brunnre8:
and iiuc component CSS gets mangled no? that's how they do the scoping?

That'd make it hard to override

MaxLeiter:
`<style>` is global by default, `<style scoped>` is scoped

MaxLeiter (approved):

PR #4655, opened 2022-12-15
Adding custom js support

### LINKED ISSUE #4654: Custom JS
### Custom JS
It would be great to add the ability to add custom JS similar to adding custom CSS. Mainly this request is coming about because of the inability to let plugins modify messages, so the next best thing would be to let us be able to inject JS into the client side for processing. 

The goal I'm trying to achieve is a plugin for custom emotes, the built-in emote processing works excellent, but I would like to see something like discords emote handling but that seems to be impossible with the current options for plugins

NickBusey:
Obviously potentially unsafe, happy to discuss.

Closes #4654

### DIFF
--- a/server/server.ts
+++ b/server/server.ts
@@ -380,7 +380,7 @@ function addSecurityHeaders(req: Request, res: Response, next: NextFunction) {
 		"form-action 'self'", // 'self' to fix saving passwords in Firefox, even though login is handled in javascript
 		"connect-src 'self' ws: wss:", // allow self for polling; websockets
 		"style-src 'self' https: 'unsafe-inline'", // allow inline due to use in irc hex colors
-		"script-src 'self'", // javascript
+		"script-src 'self' 'unsafe-eval'", // javascript
 		"worker-src 'self'", // service worker
 		"manifest-src 'self'", // manifest.json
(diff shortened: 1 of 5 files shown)


brunnre8:
I appreciate your contribution. However I don't think this is a good idea.

It is very unsafe and I don't think injecting code is really the core of what TL should do.
It can make everything unstable and we wouldn't have an easy way to troubleshoot issues coming from that.

There's plenty of ways a technical user can already inject custom scripts if they want to with various browser extensions.

Custom CSS is fine-ish but already stretching it a bit. So the presence of it should not serve as an argument why we should add js injection.

server/server.ts
```diff
@@ -380,7 +380,7 @@ function addSecurityHeaders(req: Request, res: Response, next: NextFunction) {
 		"form-action 'self'", // 'self' to fix saving passwords in Firefox, even though login is handled in javascript
 		"connect-src 'self' ws: wss:", // allow self for polling; websockets
 		"style-src 'self' https: 'unsafe-inline'", // allow inline due to use in irc hex colors
-		"script-src 'self'", // javascript
+		"script-src 'self' 'unsafe-eval'", // javascript
```
xPaw, line 383: This is definetely a no-no.

Issue #4615, opened 2022-08-05
[Unread messages count] Abandon "k" (= thousand) when it reaches million+


lucaztz:
### Feature Description
I've noticed that when a channel reaches above million unread messages, the side number stays k (i.e. 1082k), My suggestion is to implement "m" and round for the nearest three or two numbers numbers, so "1082k" rather becomes "1.08m" for three or "1.1m" for two (8 is nearer to 1 than 0).

PR #4659, opened 2022-12-19
Use "M" shorthand for messages >= one million

brettgilio:
Implements [Unread messages count] Abandon "k" (= thousand) when it reaches million+ #4615

### DIFF
--- a/client/js/helpers/roundBadgeNumber.ts
+++ b/client/js/helpers/roundBadgeNumber.ts
@@ -8,5 +8,5 @@ export default (count: number) => {
 	const {divisor, suffix} =
 		suffixes[Math.min(suffixes.length - 1, Math.floor(Math.log10(count) / 3))];
 
-	return (count / divisor).toFixed(2).slice(0, -1) + suffix;
+	return String(Math.ceil((count / divisor) * 10) / 10).concat(suffix);
 };
--- a/client/js/helpers/roundBadgeNumber.ts
+++ b/client/js/helpers/roundBadgeNumber.ts
@@ -1,12 +1,11 @@
 export default (count: number) => {
-	const suffixes = [
-		{divisor: 1, suffix: ""},
-		{divisor: 1000, suffix: "k"},
-		{divisor: 1000000, suffix: "m"},
-	];
+	if (count < 1000) {
+		return count.toString();
+	}
 
-	const {divisor, suffix} =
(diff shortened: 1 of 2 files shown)


brunnre8:
You didn't edit the test.
Not that you can actually run the tests mind you, our client side test framework is broken at the moment but still.

brettgilio:
> You didn't edit the test. Not that you can actually run the tests mind you, our client side test framework is broken at the moment but still.

Done.

test/client/js/helpers/roundBadgeNumberTest.ts
```diff
@@ -6,10 +6,15 @@ describe("roundBadgeNumber helper", function () {
 		expect(roundBadgeNumber(123)).to.equal("123");
 	});
 
-	it("should return numbers above 999 in thousands", function () {
+	it("should return numbers between 1000 and 999999 with a 'k' suffix", function () {
```
xPaw, line 9: add a few more expects, e.g. 999999 is in fact `999k`.

client/js/helpers/roundBadgeNumber.ts
```diff
@@ -3,5 +3,9 @@ export default (count: number) => {
 		return count.toString();
 	}
 
-	return (count / 1000).toFixed(2).slice(0, -1) + "k";
+	const suffixes = ["", "k", "M"];
+	const magnitudeIndex = Math.min(Math.floor(Math.log10(count) / 3), suffixes.length - 1);
```
xPaw, line 7: While I understand the use case here, I think it's simpler to just have two if cases to handle `M/k`.
```
if(count >= 1000000) M
if(count >= 1000) k
count
```

test/client/js/helpers/roundBadgeNumberTest.ts
```diff
@@ -6,10 +6,15 @@ describe("roundBadgeNumber helper", function () {
...
 		expect(roundBadgeNumber(1000)).to.be.equal("1.0k");
 	});
 
+	it("should return numbers above 999999 with a 'M' suffix", function () {
+		expect(roundBadgeNumber(1000000)).to.be.equal("1.0M");
+		expect(roundBadgeNumber(1234567)).to.be.equal("1.2M");
```
xPaw, line 15: Please add a test for much higher magnitude number (with current code it would validate that it `magnitudeIndex` works` as expected.

PR #4662, opened 2022-12-23
Improve readability of maximum upload size error

brettgilio:
The original version was a little clunky and hard to decipher. It would also be useful for rejected uploads to specify the maximum file size.

### DIFF
--- a/client/js/upload.ts
+++ b/client/js/upload.ts
@@ -125,7 +125,7 @@ class Uploader {
 			if (maxFileSize > 0 && file.size > maxFileSize) {
 				const maxFileSizeInMB = (maxFileSize / 1024 / 1024).toFixed(0);
 				this.handleResponse({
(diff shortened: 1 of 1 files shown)


client/js/upload.ts
```diff
@@ -116,6 +116,7 @@ class Uploader {
 
 		const wasQueueEmpty = this.fileQueue.length === 0;
 		const maxFileSize = store.state.serverConfiguration?.fileUploadMaxFileSize || 0;
+		const maxFileSizeInMB = (maxFileSize / 1024 / 1024).toFixed(0);
```
brunnre8, line 119: This should probably be moved into the if guard... there's no need to calculate it when you don't hit the error.

Helps the reader too that the variable has a narrow scope and isn't used elsewhere.

brettgilio: Good catch.

brunnre8:
please squash down the commits to a single one

PR #4666, opened 2022-12-24
Add edge case checks and rework logic friendlysize

brettgilio:
Follows #4662

### DIFF
--- a/client/js/helpers/friendlysize.ts
+++ b/client/js/helpers/friendlysize.ts
@@ -1,8 +1,15 @@
 const sizes = ["Bytes", "KiB", "MiB", "GiB", "TiB", "PiB"];
 
-export default (size: number) => {
-	// Loosely inspired from https://stackoverflow.com/a/18650828/1935861
-	const i = size > 0 ? Math.floor(Math.log(size) / Math.log(1024)) : 0;
-	const fixedSize = parseFloat((size / Math.pow(1024, i)).toFixed(1));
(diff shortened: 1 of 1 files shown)


brunnre8:
why are you rewriting this code?

brettgilio:
> why are you rewriting this code?

Largely to handle edge cases. I noticed them when I was viewing the file to extend @4662.

client/js/helpers/friendlysize.ts
```diff
@@ -1,8 +1,15 @@
...
-	return `${fixedSize} ${sizes[i]}`;
-};
+export default function formatSize(size: number): string {
+	if (size <= 0) {
+		throw new Error("Size must be a positive number");
+	}
+
+	const i = Math.floor(Math.log(size) / Math.log(1024));
+
+	if (i >= sizes.length) {
```
brunnre8, line 10: this must not error, simply emit a large number

brunnre8 (changes requested):

client/js/helpers/friendlysize.ts
brettgilio: To be clear you just want it to emit the KB value?

brunnre8: why KB?
You are guarding the index value, simply pick the highest available one namely sizes.length and do the dance same as you did in the other patch.

PR #4669, opened 2022-12-29
allow away and back to be collapsed

brunnre8:
This means we also apply the collapsing to normal queries, which might also collapse other things like joins / quits which may be undesired by some

### DIFF
--- a/client/components/MessageCondensed.vue
+++ b/client/components/MessageCondensed.vue
@@ -118,6 +118,18 @@ export default defineComponent({
 									(obj[type] > 1 ? " modes were set" : " mode was set")
 							);
 							break;
+						case "away":
+							strings.push(
+								"marked away " +
+									(obj[type] > 1 ? String(obj[type]) + " times" : "once")
+							);
+							break;
+						case "back":
+							strings.push(
+								"marked back " +
+									(obj[type] > 1 ? String(obj[type]) + " times" : "once")
+							);
+							break;
 					}
 				}
 			});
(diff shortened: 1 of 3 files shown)


client/components/MessageCondensed.vue
```diff
@@ -118,6 +118,18 @@ export default defineComponent({
 									(obj[type] > 1 ? " modes were set" : " mode was set")
 							);
 							break;
+						case "away":
+							strings.push(
+								String(obj[type]) +
+									(obj[type] > 1 ? " times marked away" : " time marked away")
```
MaxLeiter, line 124: not sure '5 times marked away' makes sense. How about `Marked away N times`?

client/components/MessageList.vue
```diff
@@ -177,7 +177,7 @@ export default defineComponent({
 		});
 
 		const condensedMessages = computed(() => {
-			if (props.channel.type !== "channel") {
+			if (!(props.channel.type === "channel" || props.channel.type === "query")) {
```
MaxLeiter, line 180: id prefer if type !== channel && type !== query, personally.

client/components/MessageCondensed.vue
brunnre8: ack

client/components/MessageList.vue
brunnre8: makes sense, will change

brunnre8:
flipped to marked away once in the singular case, as marked away 1 time sounds wrong to me.
Happy with that?

MaxLeiter (approved):

PR #4670, opened 2022-12-30
Storage cleanup

brunnre8:
Fixes several issues with the message storage API's

Reason I'm doing the cleanup (besides the fact that it's needed) is that I want to write a manual migration cli command.

For that I need to initialize the sqlite storage for a user, without actually having a client at all.
When we accept accept a client, that's not possible.

### DIFF
--- a/server/client.ts
+++ b/server/client.ts
@@ -138,7 +138,7 @@ class Client {
 
 		if (!Config.values.public && client.config.log) {
 			if (Config.values.messageStorage.includes("sqlite")) {
-				client.messageProvider = new SqliteMessageStorage(client);
+				client.messageProvider = new SqliteMessageStorage(client.name);
 				client.messageStorage.push(client.messageProvider);
 			}
 
--- a/server/client.ts
+++ b/server/client.ts
@@ -143,7 +143,7 @@ class Client {
 			}
 
 			if (Config.values.messageStorage.includes("text")) {
-				client.messageStorage.push(new TextFileMessageStorage(client));
+				client.messageStorage.push(new TextFileMessageStorage(client.name));
 			}
 
 			for (const messageStorage of client.messageStorage) {
--- a/server/plugins/messageStorage/sqlite.ts
+++ b/server/plugins/messageStorage/sqlite.ts
@@ -8,11 +8,7 @@ import Msg, {Message} from "../../models/msg";
 import Client from "../../client";
 import Chan, {Channel} from "../../models/chan";
 import Helper from "../../helper";
-import type {
-	SearchResponse,
-	SearchQuery,
-	SqliteMessageStorage as ISqliteMessageStorage,
-} from "./types";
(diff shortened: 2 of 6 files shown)


server/plugins/messageStorage/sqlite.ts
```diff
@@ -190,13 +185,11 @@ class SqliteMessageStorage implements ISqliteMessageStorage {
...
-	 * Load messages for given channel on a given network and resolve a promise with loaded messages.
-	 *
-	 * @param network Network - Network object where the channel is
-	 * @param channel Channel - Channel object for which to load messages for
-	 */
-	async getMessages(network: Network, channel: Channel): Promise<Message[]> {
+	async getMessages(
+		network: Network,
+		channel: Channel,
+		nextID: () => number
```
MaxLeiter, line 191: it's not immediately clear to me what nextID is. Can we either re-add a comment or give it a better name? In this case, it seems like it's not the nextId so much as the new Id?

server/client.ts
```diff
@@ -138,12 +138,12 @@ class Client {
 
 		if (!Config.values.public && client.config.log) {
 			if (Config.values.messageStorage.includes("sqlite")) {
-				client.messageProvider = new SqliteMessageStorage(client);
+				client.messageProvider = new SqliteMessageStorage(client.name);
```
MaxLeiter, line 141: nice

MaxLeiter (changes requested):

server/plugins/messageStorage/sqlite.ts
brunnre8: well, function that if called gives you the next ID.

What's your favorite bikeshed color?
I don't care much, that was what my brain came up with in a pitch but I'm happy to adapt to whatever you feel works better

brunnre8: I can certainly try a comment, but if you have a descriptive name in mind that might be better

MaxLeiter: I see why we passed the entire `client` in before...

What if we added an `incrementId` function to Client and passed it into the MessageStorage in `server/client.ts`? Then we can leave `getMessages` as is.

brunnre8: How about we call it gen_unique_msg_id() ?

Descriptive enough?

brunnre8: I really really don't want the storage backends to have access to the client, that violates all kinds of layering so I'd rather not keep getmessages as is.

brunnre8: The reason I'm passing it to this function is that this is the only place that thing gets used.

Yes, you could pass it into the constructor of the backend, but it seems very strange to pass it there, After all nothing but a single function uses it, so why add it to the state when we can use a param?

MaxLeiter (approved):

PR #4676, opened 2023-01-12
OpenID Support

### LINKED ISSUE #2065: Support Oauth with Google

9p4:
Work in progress. Should be done soon.

Design decisions: Only one OpenID provider can be used at a time. If a provider is configured, then the client will automatically authenticate with the OpenID provider on page load (given that the token is invalid).

Roadmap:
- [x] Automatic OpenID
- [x] Username claim
- [x] Role checking
- [ ] Front channel logout
- [ ] Tests
- [x] Documentation
- [ ] Better handling of authentication failure (perhaps try again?)

Resolves: #2065, #4143


defaults/config.js
```diff
@@ -470,6 +470,15 @@ module.exports = {
 		},
 	},
 
+	openid: {
+		enable: false,
+		issuerURL: "https://auth.example.com",
+		baseURL: "https://thelounge.example.com",
+		clientID: "clientId",
+		secret: "secret",
+		logout: true,
```
MaxLeiter, line 479: Not sure what this will do but I think we’ll want a more explicit variable name

MaxLeiter (changes requested):
Thanks for taking this on!

9p4:
This is ready for preliminary testing.

Remaining:
- [ ] Role checking
- [ ] Tests

server/plugins/auth/openid.ts
```diff
@@ -0,0 +1,27 @@
...
+import type {AuthHandler} from "../auth";
+
+const openIDAuth: AuthHandler = (manager, client, user, _, callback) => {
+	if (user === "") {
+		log.error(`Authentication failed`);
+		return callback(false);
+	}
+
+	// If no user is found, create it
+	if (!client) {
```
brunnre8, line 12: don't call this client unless it is our irc client, which I guess it isn't.
Using terms used in a specific context all over the place as something else should be avoided

server/plugins/auth/openid.ts
```diff
@@ -0,0 +1,27 @@
+import log from "../../log";
+import Config from "../../config";
+import type {AuthHandler} from "../auth";
+
+const openIDAuth: AuthHandler = (manager, client, user, _, callback) => {
```
brunnre8, line 5: nit: make this a proper function, not sure why you changed it to the inline style.
This is the top level, you don't need the scoping rules of arrow functions here

server/server.ts
```diff
@@ -100,6 +113,21 @@ export default async function (
...
+		redirect_uris: [Config.values.openid.baseURL],
+		response_types: ["code"],
+	});
+	var redirectUrl = openidClient.authorizationUrl({
+		scope: "openid email profile",
+		code_challenge,
+		code_challenge_method: "S256",
+	});
+	issuerURL = redirectUrl;
+
```
brunnre8, line 130: I don't think that the server should be handling any specifics of the openid auth whatsoever.

Shove that into an abstraction layer with a simple UI.
You shouldn't even be resolving the issuer unless openid auth is enabled.

server/server.ts
```diff
@@ -243,7 +271,11 @@ export default async function (
 				performAuthentication.call(socket, {});
 			} else {
 				socket.on("auth:perform", performAuthentication);
-				socket.emit("auth:start", serverHash);
+				socket.emit("auth:start", {
+					serverHash,
+					openidEnabled: Config.values.openid.enable && !Config.values.public,
+					openidInit: issuerURL,
+				});
```
brunnre8, line 278: This doesn't seem right.
that event should know nothing about specific plugins.
I don't think that there should be any kind of openid data in there

This is coupling the server to the openid plugin, which defeats the purpose of plugins in the first place.

The point of using plugins is that the server (ideally, pretty sure we violate it already) doesn't know anything about how auth works and what the special methods need.

brunnre8 (changes requested):
Thanks for working on this.

Didn't have a proper look yet, but here's what I already noticed while starting to read the code.

I think this needs quite some refactoring to encapsulate the logic rather than spreading it inline in various places

server/server.ts
9p4: Auth has to be triggered by the user, not the server. As such, the client has to know if OpenID is enabled and where to redirect the user.

server/server.ts
9p4: >  You shouldn't even be resolving the issuer unless openid auth is enabled.

Whoops, I forgot to do that. I'll get that done.

> I don't think that the server should be handling any specifics of the openid auth whatsoever.

How do you mean? Since OpenID works by communicating securely with the server (the server has to validate the client's tokens with the OpenID server), then the server has to handle specifics and make sure that the encryption algorithms and all are consistent.

brunnre8: Not servers side vs client side.
That you do the auth/basic flow server side is not the point.

What you do however is shove logic related to openid into `server.ts`
Have a look at how say the uploader or ident functionality is wired up. `server.ts` has absolutely no knowledge about those other than that it's wiring them up.

You are inlining openid specific logic.

server/server.ts
brunnre8: The auth is controlled by the server, not the user.
What you are doing there is mixing two unrelated event types, namely openid + the hash auth into one, flagged by a bool.

That's not really how you should do it.

If a auth backend needs special logic, then pass control to the auth backend and let it return the data. Then strongly type it.

The client should get an event and already know what it is, not some grab bag that contains maybe a hash, maybe openid connect, maybe $next_huge_auth_thing etc.

That just proliferates into a mess

progval:
I started looking into this (no promises!)

Out of curiosity, why did you chose the [authorization code flow](https://www.npmjs.com/package/openid-client#user-content-authorization-code-flow) instead of the [implicit flow](https://www.npmjs.com/package/openid-client#user-content-implicit-id-token-flow)? The latter looks simpler and good enough for TL

9p4:
I think it was more of a familiarity with the authorization flow more than anything else.

MaxLeiter:
no need to be sorry @9p4! Thank you for your effort 🙏

kristof-mattei:
@progval 
Implicit flow has issues that need to be mitigated separately.

> In order to avoid these issues, clients SHOULD NOT use the implicit grant (response type "token") or other response types issuing access tokens in the authorization response, unless access token injection in the authorization response is prevented and the aforementioned token leakage vectors are mitigated.

From https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-2.1.2


This is not the case with the Auth Code flow, nor with the PKCE flow.

Morganamilo:
If it's of use to anyone I updated this patch set to work with current master and patched my lounge instance and it seems to work just fine. Here's the patch file if it's of use to anyone: https://bin.src.lu/thelounge-openid

pinpox:
> I updated this patch set to work with current master and patched my lounge instance and it seems to work just fine. Here's the patch file if it's of use to anyone: https://bin.src.lu/thelounge-openid

Could you open a PR for this or continue on this one? I'm still interested in this feature, but wouldn't want to keep messing with external patches every release.

phls:
Hi, any news about 

> > I updated this patch set to work with current master and patched my lounge instance and it seems to work just fine. Here's the patch file if it's of use to anyone: https://bin.src.lu/thelounge-openid
> 
> Could you open a PR for this or continue on this one? I'm still interested in this feature, but wouldn't want to keep messing with external patches every release.

Hi, any news about the PR?

Morganamilo:
> > I updated this patch set to work with current master and patched my lounge instance and it seems to work just fine. Here's the patch file if it's of use to anyone: [bin.src.lu/thelounge-openid](https://bin.src.lu/thelounge-openid)
> 
> Could you open a PR for this or continue on this one? I'm still interested in this feature, but wouldn't want to keep messing with external patches every release.

The PR was closed because the implementation wasn't complete or should be done a different way. I rebased this PR as it was and don't have any interest in actually developing it.

phls:
> Hi, any news about
> 
> > > I updated this patch set to work with current master and patched my lounge instance and it seems to work just fine. Here's the patch file if it's of use to anyone: https://bin.src.lu/thelounge-openid
> > 
> > 
> > Could you open a PR for this or continue on this one? I'm still interested in this feature, but wouldn't want to keep messing with external patches every release.
> 
> Hi, any news about the PR?

I don't know why my commented was marked as spam, because I really like to see this feature implemented and I would like to know If there was any update about that.

Morganamilo:
@pinpox @phls 

No PR because I don't think this meets the standards to be merged and I don't know the codebase at all.

But I've rebased my patch on the latest master so I can update. The patch is based on commit f97c4df and the patch is 
 here: https://bin.src.lu/tl-oid-f97c4df

deevus:
I'm taking this up. I'll create a PR when it's ready.

Issue #4664, opened 2022-12-23
Generate git info at build time, not run time


MaxLeiter:
Currently we fetch the git information for version warnings from the machine. However, this won't work with the typescript changes because file paths have changed. Additionally, we have this knowledge at build so we may as well use it.

xPaw:
I'm thinking version can be generated into a separate json file, which would be combined with package.json.

There's only a handful of properties that are used from package.json, so there's no reason to load entirety of it at runtime.

PR #4678, opened 2023-01-22
Fix git commit not being available in dist build

xPaw:
"temporary" fix, a proper fix would be #4664.

### DIFF
--- a/server/helper.ts
+++ b/server/helper.ts
@@ -44,23 +44,23 @@ function getVersionNumber() {
 	return pkg.version;
 }
 
+let _fetchedGitCommit = false;
 let _gitCommit: string | null = null;
 
 function getGitCommit() {
-	if (_gitCommit) {
+	if (_fetchedGitCommit) {
 		return _gitCommit;
 	}
 
-	if (!fs.existsSync(path.resolve(__dirname, "..", ".git"))) {
-		_gitCommit = null;
-		return null;
-	}
+	_fetchedGitCommit = true;
 
+	// --git-dir ".git" makes git only check current directory for `.git`, and not travel upwards
+	// We set cwd to the location of `index.js` as soon as the process is started
 	try {
 		// eslint-disable-next-line @typescript-eslint/no-var-requires
 		_gitCommit = require("child_process")
 			.execSync(
-				"git rev-parse --short HEAD", // Returns hash of current commit
+				'git --git-dir ".git" rev-parse --short HEAD', // Returns hash of current commit
 				{stdio: ["ignore", "pipe", "ignore"]}
 			)
 			.toString()
--- a/server/server.ts
+++ b/server/server.ts
@@ -9,7 +9,6 @@ import colors from "chalk";
 import net from "net";
 
 import log from "./log";
-import pkg from "../package.json";
 import Client from "./client";
 import ClientManager from "./clientManager";
 import Uploader from "./plugins/uploader";
@@ -885,7 +884,7 @@ function getClientConfiguration(): ClientConfiguration {
 
 	config.isUpdateAvailable = changelog.isUpdateAvailable;
 	config.applicationServerKey = manager!.webPush.vapidKeys!.publicKey;
-	config.version = pkg.version;
+	config.version = Helper.getVersionNumber();
 	config.gitCommit = Helper.getGitCommit();
 	config.themes = themes.getAll();
 	config.defaultTheme = Config.values.theme;


server/helper.ts
```diff
@@ -44,14 +44,27 @@ function getVersionNumber() {
...
+
+	// Typescript build will be in dist/ folder,
+	// but it will not be when running `dev`.
+	if (rootPath.endsWith(path.join("dist", "server"))) {
+		rootPath = path.basename(rootPath);
+	}
+
+	// Check that .git folder exists in the expected place,
+	// to avoid git command mistakenly recursing outside of our root folder
+	if (!fs.existsSync(path.resolve(rootPath, "..", ".git"))) {
```
brunnre8, line 67: Note that this isn't necessarily a folder... so that comment is wrong.
If it's a worktree, .git is a file.

xPaw: TIL, but that doesn't strictly matter.

brunnre8: I know, but I already had to patch out the git folder detection once. And I'd like to avoid having to do so again so I'd appreciate it if we can get rid of the notion that .git is a folder

brunnre8: thanks

server/helper.ts
```diff
@@ -44,14 +44,27 @@ function getVersionNumber() {
...
+
+	// Typescript build will be in dist/ folder,
+	// but it will not be when running `dev`.
+	if (rootPath.endsWith(path.join("dist", "server"))) {
+		rootPath = path.basename(rootPath);
+	}
+
+	// Check that .git folder exists (or a file in case of a worktree) in the expected place,
+	// to avoid git command mistakenly recursing outside of our root folder
+	if (!fs.existsSync(path.resolve(rootPath, "..", ".git"))) {
```
brunnre8, line 67: So looking at this a bit more closely.. why are you doing the path munging dance there?

Forgive me my ignorance, it took me a while to wrap my head around how it works.

Can't this just be simplified to

```js
function getGitCommit() {
	if (_fetchedGitCommit) {
		return _gitCommit;
	}

	_fetchedGitCommit = true;

	// Check that .git folder exists (or a file in case of a worktree) in the expected place,
	// to avoid git command mistakenly recursing outside of our root folder
	if (!fs.existsSync(path.resolve(process.cwd(), ".git"))) {
		_gitCommit = null;
		return null;
	}

	try {
		// eslint-disable-next-line @typescript-eslint/no-var-requires
		_gitCommit = require("child_process")
			.execSync(
				"git rev-parse --short HEAD", // Returns hash of current commit
				{stdio: ["ignore", "pipe", "ignore"]}
			)
			.toString()
			.trim();
		return _gitCommit;
	} catch (e: any) {
		// Not a git repository or git is not installed
		_gitCommit = null;
		return null;
	}
}
```

The whole __dirname + relative dir dance you do there only works if cwd() is set to the repo root anyhow.

Consider this:
__dirname does not point to dist server, say I run `yarn dev` in whatever dir (something seems to set cwd to the repo root no matter where you are in).
So you don't munge the path but simply head up to the repo root. That makes perfect sense to me.

For me, that essentially looks like that
```
rootpath: /home/reto/sourcecode/thelounge.hotfix/server
cwd: /home/reto/sourcecode/thelounge.hotfix
looking in /home/reto/sourcecode/thelounge.hotfix/.git
```

However, if you build the thing and start it via index.js, not yarn, you munge the rootpath to "server"
Meaning your path resolve walks down some random directory, based on cwd() only to immediately walk back up again as you do `server/../.git`

With the same location as the yarn dev example above:

```
rootpath: /home/reto/sourcecode/thelounge.hotfix/dist/server
rootpath munged to: server
cwd: /home/reto/sourcecode/thelounge.hotfix
looking in /home/reto/sourcecode/thelounge.hotfix/.git
```

In other words, that only works because the cwd() is set to index.js in the first place, and that's done by the index.js entrypoint

```js
#!/usr/bin/env node

  "use strict";

  process.chdir(__dirname);
```

Or by yarn or whatever is changing the working directory when you execute yarn dev, so that whole dance you do there seems pointless.

What am I missing?

xPaw: __dirname should point to the helper.js location, and looks like I indeed made a mistake with basename there.

Considering we chdir to index.js, your idea to use cwd should work regardless of dev/dist modes.

xPaw: I simplified it by using `git --git-dir ".git"`

> Specifying the location of the ".git" directory using this option (or GIT_DIR environment variable) turns off the repository discovery that tries to find a directory with ".git" subdirectory (which is how the repository and the top-level of the working tree are discovered), and tells Git that you are at the top level of the working tree.

brunnre8:
LGTM, thanks

Issue #4620, opened 2022-08-14
Cannot restart thelounge when using LDAP with goauthentik's LDAP outpost


gravypod:
I setup thelounge (v4.3.1) using this container image: thelounge/thelounge@sha256:a08396bfa3ab7b26b83c24056c9cc34322596d20fca2216ada461e33d803728e

Once configured for ldap I was able to login with my ldap credentials:

```
                searchDN: {
                        rootDN: "cn=xxx,ou=user,dc=ldap,dc=goauthentik,dc=io",
                        rootPassword: "xxxx",
                        filter: "(objectClass=user)(memberOf=cn=ldap,ou=groups,dc=ldap,dc=goauthentik,dc=io)",
                        base: "DC=ldap,DC=goauthentik,DC=io",
                        scope: "sub",
                },
```

After doing so I was no longer able to restart the service because it would get into a crashloop. It would print out this error:

```
/usr/local/share/.config/yarn/global/node_modules/ldap-filter/lib/index.js:256
      throw new Error('unbalanced parens');
      ^

Error: unbalanced parens
    at Object.parse (/usr/local/share/.config/yarn/global/node_modules/ldap-filter/lib/index.js:256:13)
    at Object.parseString (/usr/local/share/.config/yarn/global/node_modules/ldapjs/lib/filters/index.js:179:27)
    at Client.search (/usr/local/share/.config/yarn/global/node_modules/ldapjs/lib/client/client.js:571:30)
    at /usr/local/share/.config/yarn/global/node_modules/thelounge/src/plugins/auth/ldap.js:179:14
    at callbackWrapper (/usr/local/share/.config/yarn/global/node_modules/ldapjs/lib/client/client.js:294:5)
    at sendResult (/usr/local/share/.config/yarn/global/node_modules/ldapjs/lib/client/client.js:1197:12)
    at messageCallback (/usr/local/share/.config/yarn/global/node_modules/ldapjs/lib/client/client.js:1220:16)
    at Parser.onMessage (/usr/local/share/.config/yarn/global/node_modules/ldapjs/lib/client/client.js:886:14)
    at Parser.emit (node:events:527:28)
    at Parser.write (/usr/local/share/.config/yarn/global/node_modules/ldapjs/lib/messages/parser.js:107:8)
```

I traced it back to this line of code: https://github.com/thelounge/thelounge/blob/11f7ae98be98067b1a957d95d1f67dbf2d8065b6/server/plugins/auth/ldap.ts#L183

[...]

PR #4680, opened 2023-01-27
Fix misleading LDAP filiter in default config

goodspeed34:
The default filter provided in config files is invalid. This may mislead people. Confirm to issue #4620.


brunnre8:
Thanks a lot

brunnre8:
Closes #4620

PR #4681, opened 2023-01-28
update npmignore for ts build

brunnre8:
With the introduction of typescript we need to change the exclude list of the package generation.

This commit does just that, by flipping the logic to an include list. Maintaining an include list is actually easier in our case than the exclude list, but we can't use the "files" package.json field as we need to negate specific patterns in public/.
Patterns like `!public/js/bundle.vendor.js.map` in "files" make `yarn pack` include all the files, as the check they do is buggy. `npm pack` would work fine, but then we mix package managers again.

Using the allow list in .npmignore makes it work with both, npm and yarn, but gets a bit more tedious to read for a human.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,25 +1,34 @@
-# This file must not contain generated assets listed in .gitignore.
-# npm-debug.log and node_modules/ are ignored by default.
+# Patterns specified here are excluded from packages produced by `yarn pack`
 # See https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package
 
-# Ignore all dot files except for .thelounge_home
-.*
-!.thelounge_home
+# in theory we should be using an include list, but yarn can't cope
+# with negative patterns in the package.json's "files" field
+# see https://github.com/yarnpkg/yarn/issues/8332
+# so we monky patch one ourselves
 
-# Ignore client folder as it's being built into public/ folder
-# except for the specified files which are used by the server
-client/**
-!client/js/constants.ts
-!client/js/helpers/ircmessageparser/findLinks.ts
-!client/js/helpers/ircmessageparser/cleanIrcMessage.ts
+# Note some files are protected and always included, so the list is not exhaustive
+
+!.thelounge_home
+!index.js
+# give yarn global add a chance to use the right deps
+!yarn.lock
+
+!dist/package.json
+
+!dist/server/**/*.js
+
+!dist/defaults/**/*.js
+
+!dist/shared/**/*.js
+
+# this doesn't get copied to dist/ so we need to manually include it
+# TODO: this should be fixed by copying it to dist/ in the webpack build
 !client/index.html.tpl
 
+# ignore specific files in public/
 public/js/bundle.vendor.js.map
-coverage/
-scripts/
-test/
-appveyor.yml
-webpack.config*.js
-postcss.config.js
-renovate.json
+# then include the rest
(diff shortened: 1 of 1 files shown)


brunnre8:
@xPaw is this what you had in mind?

brunnre8:
Still contains too much stuff, but that'd be https://github.com/thelounge/thelounge/issues/4682

brunnre8:
Meh, I think I missunderstood the dist/ folder... I'll have another go

xPaw:
The idea looks right, is there an issue?

brunnre8:
I'm bundling the client

.npmignore
```diff
@@ -1,25 +1,34 @@
...
 
-# Ignore client folder as it's being built into public/ folder
-# except for the specified files which are used by the server
-client/**
-!client/js/constants.ts
-!client/js/helpers/ircmessageparser/findLinks.ts
-!client/js/helpers/ircmessageparser/cleanIrcMessage.ts
+# this doesn't get copied to dist/ so we need to manually include it
+# TODO: this should be fixed by copying it to dist/ in the webpack build
 !client/index.html.tpl
```
MaxLeiter, line 26: By running `yarn pack` and using `node index.js` I got

```

Error: Cannot find module '../client/js/constants'
Require stack:
- /Users/maxleiter/Documents/thelounge/package/dist/server/client.js

```

Do we need to include constants here?

MaxLeiter (changes requested):

.npmignore
MaxLeiter: also, `client/` should probably be in `dist/`, right?

MaxLeiter: there's also no `shared/` if I update your branch with main, so I doubt its a new problem introduced here. But I'm pretty stumped.

brunnre8: client/ shouldn't be in dist, no.
That's used in the static handler which serves from the root iirc

brunnre8:
So, someone on the IRC chan was kind enough to run the pack command on a virtualized OSX install of theirs... works for them.

So I don't think the code is to blame here

.npmignore
```diff
@@ -1,25 +1,34 @@
...
+# Note some files are protected and always included, so the list is not exhaustive
+
 !.thelounge_home
+!index.js
+# give yarn global add a chance to use the right deps
+!yarn.lock
+
+!dist/package.json
+
+!dist/server/**/*.js
```
xPaw, line 18: Wouldn't `!dist/**/*.js` work just as well here without specifying each sub directory?

.npmignore
```diff
@@ -1,25 +1,34 @@
...
-client/**
-!client/js/constants.ts
-!client/js/helpers/ircmessageparser/findLinks.ts
-!client/js/helpers/ircmessageparser/cleanIrcMessage.ts
+# this doesn't get copied to dist/ so we need to manually include it
+# TODO: this should be fixed by copying it to dist/ in the webpack build
 !client/index.html.tpl
 
+# ignore specific files in public/
 public/js/bundle.vendor.js.map
```
xPaw, line 29: Perhaps we should just include it?

brunnre8: Bah...
Ordering doesn't seem to matter, putting it in the other way still includes it.

brunnre8: https://github.com/thelounge/thelounge/pull/4707 ?

PR #4686, opened 2023-01-30
Decouple server

brunnre8:
We should eventually decouple both server from client and client from server.
Shared stuff should move to shared/ or some such.

This decouples the server from the client, in a first step as that's the easier one


client/js/helpers/ircmessageparser/merge.ts
```diff
@@ -2,7 +2,7 @@ import anyIntersection from "./anyIntersection";
 import fill from "./fill";
 import {ChannelPart} from "./findChannels";
 import {EmojiPart} from "./findEmoji";
-import {LinkPart} from "./findLinks";
+import {LinkPart} from "../../../../shared/linkify";
```
MaxLeiter, line 5: eventual TODO: typescript path aliases would let us something like `#/shared/linkify`

MaxLeiter (approved):

PR #4690, opened 2023-02-05
Fix Morning theme nick colors

maxpoulin64:
PR #4649 introduced a regression on the Morning theme as the `#chat.colored-nicks` CSS selector was removed from Default but not Morning. The result is that Morning no longer had nick colors. This applies the same change that was applied to the global CSS.

### DIFF
--- a/client/themes/morning.css
+++ b/client/themes/morning.css
@@ -52,42 +52,42 @@ body {
 }
 
 /* Nicknames */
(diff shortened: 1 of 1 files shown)


brunnre8:
whoops, sorry.

Thanks for catching that.

brunnre8 (approved):

PR #4691, opened 2023-02-06
Fix uploader mount/unmount lifecycle

maxpoulin64:
Currently, in `ChatInput.vue` we call `upload.abort()` which removes the event listeners, which are never added back. This effectively permanently disable uploads if the user navigates away to Settings or any other non-chat pages, and back.

Moves the binding to `mounted()` so that they're properly rebound when a chat window is in view, and also adds an `unmounted()` for clarity.

This should also fix an edge case if the page opens up on a non-chat page and there was never a ChatInput to unbind it, such as login page or add network pages.

### DIFF
--- a/client/components/ChatInput.vue
+++ b/client/components/ChatInput.vue
@@ -337,6 +337,7 @@ export default defineComponent({
 				autocompletionRef.value = undefined;
 			}
 
+			upload.unmounted();
 			upload.abort();
 		});
 
--- a/client/js/upload.ts
+++ b/client/js/upload.ts
@@ -18,18 +18,26 @@ class Uploader {
 	onPaste = (e: ClipboardEvent) => this.paste(e);
 
 	init() {
-		document.addEventListener("dragenter", this.onDragEnter);
-		document.addEventListener("dragover", this.onDragOver);
-		document.addEventListener("dragleave", this.onDragLeave);
-		document.addEventListener("drop", this.onDrop);
-		document.addEventListener("paste", this.onPaste);
-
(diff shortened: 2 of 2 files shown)


client/components/ChatInput.vue
```diff
@@ -337,6 +337,7 @@ export default defineComponent({
 				autocompletionRef.value = undefined;
 			}
 
+			upload.unmounted();
 			upload.abort();
```
brunnre8, line 341: Would it make sense to fold abort into unmounted?

As in, calling unmount should make the uploader clean up after itself no?
Seems strange to shit that to the caller after it told the uploader that it should go away

brunnre8: shift* ah, bad typo... sorry

maxpoulin64: I guess we could make `unmounted` call `abort`, but this whole thing kind of needs refactoring honestly. The way the pasting of the link works completely bypasses Vue. As the comment says, abort is a hack already.

One could want to call `abort` on pressing escape, and one could want to stop accepting new uploads but keep the current one going in the background, although right now that's undefined behaviour that crashes.

Ideally one should be able to start pasting files into different channels and not interfere, but right now it just gets pasted into whatever is active right now despite otherwise being able to start typing in one channel, move to another, type, move back and continue typing the original message. If bound the upload to the channel buffer rather than the input field, we wouldn't need abort and could register/deregister as needed.

brunnre8: yeah... I see the problem.

One more thing on the todo list I guess.
Thanks.

PR #4695, opened 2023-02-09
Don't crash on oidentd socket race condition

maxpoulin64:
My instance just crashed on this out of nowhere. It seems like this is a condition that can happen when two or more sockets gets disconnected at the same time. Each one triggers `socket close`, and for each of those we remove the socket from oidentd which in turn refreshes the oidentd config file. Once we're in there refreshing the database, we trip on the other closed sockets we haven't cleaned up yet, which in turn crashes the whole server.

I have a handful of users on my instance, with at least one server in common so one server restarting would cause a handful of closed sockets on the same tick which would trigger this race condition. At last this is the only plausible explanation I can think of for this happening.

Switched it to just log and skip it instead of an exception.

```
/dist/server/identification.js:89
    throw new Error("Socket has no remote or local port");
                ^
Error: Socket has no remote or local port
    at /dist/server/identification.js:89:23
    at Map.forEach (<anonymous>)
    at Identification.refresh (/dist/server/identification.js:87:26)
    at Identification.removeSocket (/dist/server/identification.js:82:18)
    at IrcClient.<anonymous> (/dist/server/plugins/irc-events/connection.js:101:41)
    at IrcClient.emit (/node_modules/eventemitter3/index.js:181:35)
    at Connection.<anonymous> (/node_modules/irc-framework/src/client.js:99:29)
    at Connection.emit (/node_modules/eventemitter3/index.js:202:33)
    at Connection.socketClose (/node_modules/irc-framework/src/connection.js:111:18)
    at Connection.emit (node:events:390:28)
```


server/identification.ts
```diff
@@ -118,7 +118,13 @@ class Identification {
 
 		this.connections.forEach((connection) => {
 			if (!connection.socket.remotePort || !connection.socket.localPort) {
-				throw new Error("Socket has no remote or local port");
+				// Race condition: this can happen when more than one socket gets disconnected at
+				// once, since we `refresh()` for each one being added/removed. This results
+				// in there possibly be one or more disconnected sockets remaining when we get here.
```
igalic, line 123:
```suggestion
				// in there possibly being one or more disconnected sockets remaining when we get here.
```

server/identification.ts
```diff
@@ -118,7 +118,13 @@ class Identification {
...
+				//
+				// Simply skip this socket and not crash the server.
```
igalic, line 125:
```suggestion
				// Simply skip this socket and do not crash the server.
```

igalic (changes requested):
👀

MaxLeiter (approved):

server/identification.ts
```diff
@@ -118,7 +118,13 @@ class Identification {
 
 		this.connections.forEach((connection) => {
 			if (!connection.socket.remotePort || !connection.socket.localPort) {
-				throw new Error("Socket has no remote or local port");
+				// Race condition: this can happen when more than one socket gets disconnected at
+				// once, since we `refresh()` for each one being added/removed. This results
+				// in there possibly being one or more disconnected sockets remaining when we get here.
+				//
+				// Simply skip this socket and do not crash the server.
```
xPaw, line 125: The only reason it crashed is because we had a `throw` here, so this comment isn't really relevant with the log.

I hope the warning won't persist every time it refreshes, which would indicate that the socket is never removed. I'd add `id` to the log so if that does happen it could be caught.

brunnre8: Makes sense, updated the PR

PR #4697, opened 2023-02-20
implement auth plugins

MobiusCraftFlip:
Implements auth plugins

example plugin:
```javascript
module.exports = {
	onServerStart: (api) => {
	},
	auth: [
		{
			moduleName: "testlocal",
			auth: localAuth,
			isEnabled: () => true,

		}
	],
	"1234": "123123",
};
```

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -492,4 +492,5 @@ module.exports = {
 		// server window, displayed on the client.
 		raw: false,
 	},
+	authModule: "local",
 };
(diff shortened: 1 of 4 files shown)


brunnre8:
Please don't do back merges, they aren't necessary.
Rebase your changes instead.

Also, do you plan to actually PR this eventually? If yes that'll need some discussion as to what your plans are, how you think the API should look etc.

PR #4704, opened 2023-03-04
Plugin input

brunnre8:
This types the public command API during the registration phase and reworks the input function to be a bit easier to understand, least for me.

Holler if you disagree, but I hate the pattern of

```js
if(stuff) {
 // ...
} else if (other) {
 // ...
} else if (yet another) {
 // ...
}
// return
```

That (implicit) return right there is not even on screen when you start to read the first few clauses, so you have no idea what the function does at the end only to then scroll down and find out that it indeed just returns.

We already know that we are done, tell the reader accordingly.

### DIFF
--- a/server/client.ts
+++ b/server/client.ts
@@ -465,38 +465,9 @@ class Client {
 		const cmd = args?.shift()?.toLowerCase() || "";
 
 		const irc = target.network.irc;
-		let connected = irc && irc.connection && irc.connection.connected;
+		const connected = irc?.connected;
 
-		if (inputs.userInputs.has(cmd)) {
-			const plugin = inputs.userInputs.get(cmd);
-
-			if (!plugin) {
-				// should be a no-op
-				throw new Error(`Plugin ${cmd} not found`);
-			}
-
-			if (typeof plugin.input === "function" && (connected || plugin.allowDisconnected)) {
-				connected = true;
-				plugin.input.apply(client, [target.network, target.chan, cmd, args]);
-			}
-		} else if (inputs.pluginCommands.has(cmd)) {
-			const plugin = inputs.pluginCommands.get(cmd);
-
-			if (typeof plugin.input === "function" && (connected || plugin.allowDisconnected)) {
-				connected = true;
-				plugin.input(
-					new PublicClient(client, plugin.packageInfo),
-					{network: target.network, chan: target.chan},
-					cmd,
-					args
-				);
-			}
-		} else if (connected) {
-			// TODO: fix
-			irc!.raw(text);
-		}
-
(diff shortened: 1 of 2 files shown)


server/client.ts
```diff
@@ -465,46 +465,54 @@ class Client {
...
+				new Msg({
+					type: MessageType.ERROR,
+					text: "You are not connected to the IRC network, unable to send your command.",
+				})
+			);
+		};
 
 		if (inputs.userInputs.has(cmd)) {
-			const plugin = inputs.userInputs.get(cmd);
+			const plugin = inputs.userInputs.get(cmd)!; // this is only save because we just checked it
```
MaxLeiter, line 481: nit: I meant to suggest changes here, but I'd prefer something like https://github.com/thelounge/thelounge/pull/4714

server/plugins/inputs/index.ts
```diff
@@ -89,9 +101,22 @@ const getCommands = () =>
 		.concat(passThroughCommands)
 		.sort();
 
-const addPluginCommand = (packageInfo: PackageInfo, command, func) => {
-	func.packageInfo = packageInfo;
-	pluginCommands.set(command, func);
+const addPluginCommand = (packageInfo: PackageInfo, command: any, obj: any) => {
```
MaxLeiter, line 104: should `obj: any` be `obj: ExternalPluginCommand`?

brunnre8: nope, it's provided by an external package, we literally don't know what it is, we must kot lie to the compiler / human

server/client.ts
brunnre8: your PR over there fails tze typing check though

server/plugins/inputs/index.ts
MaxLeiter: nit: `unknown` is preferred then

MaxLeiter (approved):

server/plugins/inputs/index.ts
brunnre8: ack, will change

brunnre8: actually... if I do that the typing gets really annoying.

It errors out even if I guard that this is an object, due to it not having the input property.
Now, I can either type assert it or just keep it as any.

Type asserting it seems worse to me, opinons?

server/client.ts
brunnre8: done

server/plugins/inputs/index.ts
MaxLeiter: Then I think its fine as is. Sorry for wasting your time

PR #4710, opened 2023-03-14
Fix load of channels from user config

progval:
Network.export() only writes the "type" key if it's a ChanType.QUERY; so the config on disk has no "type":

https://github.com/thelounge/thelounge/blob/efd24fd12cad9192d6f333c5a3c01c33ad23b0c6/server/models/network.ts#L633-L639

This causes it to be undefined when loading, which breaks various other checks, and then drops it the next time the config is saved:

https://github.com/thelounge/thelounge/blob/efd24fd12cad9192d6f333c5a3c01c33ad23b0c6/server/models/network.ts#L627-L629

### DIFF
--- a/server/client.ts
+++ b/server/client.ts
@@ -6,7 +6,7 @@ import crypto from "crypto";
 import colors from "chalk";
 
 import log from "./log";
-import Chan, {Channel, ChanType} from "./models/chan";
+import Chan, {ChanConfig, Channel, ChanType} from "./models/chan";
 import Msg, {MessageType, UserInMessage} from "./models/msg";
 import Config from "./config";
 import {condensedTypes} from "../shared/irc";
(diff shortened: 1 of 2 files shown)


xPaw:
Can you add test coverage for this?

progval:
I don't think I can do that without refactoring; it's in a the function that instantiates irc-framework

xPaw:
Mhm yeah. I'm trying to think how this worked (before?) and when did it break.

progval:
I just opened a PR with said refactoring: #4711

And merged #4711 into this PR along with a new test so you can see what it looks like.

brunnre8 (approved):

brunnre8:
Can you squash down the fixes? Then I'll merge it.
I'd do it on my own, but then GH is too stupid to let me "merge" it via some other commit and you'd loose the attribution here.

progval:
done

brunnre8:
Thank you val

PR #4711, opened 2023-03-14
Fix config typing and make Client easier to test

progval:
I recommend reviewing commits separately (and avoiding squash-merging), it should be easier to read

### DIFF
--- a/server/client.ts
+++ b/server/client.ts
@@ -180,8 +180,16 @@ class Client {
 				this.registerPushSubscription(session, session.pushSubscription, true);
 			}
 		});
+	}
 
-		(client.config.networks || []).forEach((network) => client.connect(network, true));
+	connect() {
+		const client = this;
+
+		if (client.networks.length !== 0) {
+			throw new Error(`${client.name} is already connected`);
+		}
+
+		(client.config.networks || []).forEach((network) => client.connectToNetwork(network, true));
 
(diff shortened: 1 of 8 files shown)


server/client.ts
```diff
@@ -96,7 +96,7 @@ class Client {
 		[socketId: string]: {token: string; openChannel: number};
 	};
 	config!: UserConfig & {
-		networks?: Network[];
+		networks?: NetworkConfig[];
```
xPaw, line 99: Do you know why this `networks` isn't specified in `UserConfig` directly?

progval: I assume it's optional in the attribute because of:

https://github.com/thelounge/thelounge/blob/efd24fd12cad9192d6f333c5a3c01c33ad23b0c6/server/client.ts#L182-L184

but required as the constructor argument.

but I don't see the point of making it required. Fixed

server/clientManager.ts
```diff
@@ -283,7 +285,7 @@ class ClientManager {
 
 		try {
 			const data = fs.readFileSync(userPath, "utf-8");
-			return JSON.parse(data) as UserConfig;
+			return JSON.parse(data) as UserConfig & {networks: NetworkConfig[]};
```
xPaw, line 288: not needed anymore?

MaxLeiter (approved):

PR #4715, opened 2023-03-17
tests/server: Tear down test fixtures in the order they were setup

progval:
if for whatever reason before() fails to import the server, it causes after() to fail on the first line, so it doesn't restore stubs; causing other errors to be printed in other tests ("TypeError: Attempted to wrap warn which is already wrapped")

### DIFF
--- a/test/server.ts
+++ b/test/server.ts
@@ -39,10 +39,10 @@ describe("Server", function () {
 	});
 
 	after(function (done) {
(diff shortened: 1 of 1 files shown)


brunnre8:
Might be worth a comment so that this doesn't get shuffled back in the future.

brunnre8 (approved):

PR #4716, opened 2023-03-19
Restore constructor props commits

brunnre8:
Val's PR apparently fixes the issue introduced by https://github.com/thelounge/thelounge/commit/e31c95e32d44d709d96c008cd77d6f6aca60ef40 in https://github.com/thelounge/thelounge/pull/4710

Meaning we can re-introduce the commits that were reverted after merging it.

However, I don't understand **why** the commit referenced above exposes the breakage that was apparently already there in the first place.
I'd appreciate it if someone could educate me / point my brain in the right direction

### DIFF
--- a/server/models/chan.ts
+++ b/server/models/chan.ts
@@ -41,19 +41,18 @@ export type ChanConfig = {
 };
 
 class Chan {
-	// TODO: don't force existence, figure out how to make TS infer it.
-	id!: number;
-	messages!: Msg[];
-	name!: string;
-	key!: string;
-	topic!: string;
-	firstUnread!: number;
-	unread!: number;
-	highlight!: number;
-	users!: Map<string, User>;
-	muted!: boolean;
-	type!: ChanType;
-	state!: ChanState;
+	id: number;
+	messages: Msg[];
+	name: string;
+	key: string;
+	topic: string;
+	firstUnread: number;
+	unread: number;
+	highlight: number;
+	users: Map<string, User>;
+	muted: boolean;
+	type: ChanType;
+	state: ChanState;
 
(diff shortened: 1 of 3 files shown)


xPaw:
This is probably why:
```js
var test = { a: 1, b: 2, c: 3 };

Object.assign(test, { b: undefined, c: null });

//test = {a: 1, b: undefined, c: null}
```

server/models/chan.ts
```diff
@@ -63,20 +62,22 @@ class Chan {
...
+		this.name = "";
+		this.key = "";
+		this.topic = "";
+		this.type = ChanType.CHANNEL;
+		this.state = ChanState.PARTED;
+		this.firstUnread = 0;
+		this.unread = 0;
+		this.highlight = 0;
+		this.users = new Map();
+		this.muted = false;
```
xPaw, line 76: Since this is a class now, why don't we assign all of these as defaults on the class properties instead of setting them in the constructors?

server/models/chan.ts
```diff
@@ -63,20 +62,22 @@ class Chan {
...
+
+		if (attr) {
+			Object.assign(this, attr);
```
xPaw, line 79: I think a better fix would be make all of the parameters that chan creation uses explicit parameters of the constructor, such as `constructor(name, type, key, muted)` (and whatever else parameters can be set on creation.

Alternatively, it could be static helpers such as `createNewQuery` which returns an instance of new `Chan` that is set to type of query.

brunnre8:
Let's go with the approach outlined by xpaw there and avoid similar issues for now if we can, or at least log them.

PR #4718, opened 2023-03-25
sqlite: Add rollback support

brunnre8:
This enabled db migrations to be undone, or "down migrated". The down migration shouldn't be done automatically as it could lead to severe data loss if that were done. Hence, we still hard fail if we encounter a version lower than what we have in the DB.

A CLI will be added in a later commit that allows users to explicitly do that.

### DIFF
--- a/server/plugins/messageStorage/sqlite.ts
+++ b/server/plugins/messageStorage/sqlite.ts
@@ -24,30 +24,72 @@ try {
 }
 
 type Migration = {version: number; stmts: string[]};
+type Rollback = {version: number; rollback_forbidden?: boolean; stmts: string[]};
 
-export const currentSchemaVersion = 1672236339873; // use `new Date().getTime()`
+export const currentSchemaVersion = 1679743888000; // use `new Date().getTime()`
 
 // Desired schema, adapt to the newest version and add migrations to the array below
 const schema = [
-	"CREATE TABLE IF NOT EXISTS options (name TEXT, value TEXT, CONSTRAINT name_unique UNIQUE (name))",
-	"CREATE TABLE IF NOT EXISTS messages (id INTEGER PRIMARY KEY AUTOINCREMENT, network TEXT, channel TEXT, time INTEGER, type TEXT, msg TEXT)",
-	"CREATE INDEX IF NOT EXISTS network_channel ON messages (network, channel)",
-	"CREATE INDEX IF NOT EXISTS time ON messages (time)",
+	"CREATE TABLE options (name TEXT, value TEXT, CONSTRAINT name_unique UNIQUE (name))",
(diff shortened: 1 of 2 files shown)


brunnre8:
@itsjohncs well... if we want to have down migrations, that's what it takes I guess.
Bunch of code, CLI is still missing.

I still don't think that auto migration in the downward direction makes any sense, so we hard fail still.

A later PR can add a CLI that calls `storage.downgrade()`, the infra is in place.

Happy with that approach?

server/plugins/messageStorage/sqlite.ts
```diff
@@ -198,6 +251,116 @@ class SqliteMessageStorage implements SearchableMessageStorage {
...
+			join migrations on migrations.id=rollback_steps.migration_id
+			where version > ?
+			order by version desc, step asc`,
+			since_version
+		);
+		const result: Rollback[] = [];
+
+		// convert to Rollback[]
+		// requires ordering in the sql statement
+		res.reduce((combined: Rollback[], cur) => {
```
itsjohncs, line 267: Is this reduction just being used as a for loop? This is pretty hard for me to follow.

server/plugins/messageStorage/sqlite.ts
```diff
@@ -198,6 +251,116 @@ class SqliteMessageStorage implements SearchableMessageStorage {
...
+			await this.serialize_run("ROLLBACK", []);
+			throw err;
+		}
+
+		await this.serialize_run("COMMIT", []);
+		return new_version;
+	}
+
+	async downgrade() {
+		return this.downgrade_to(currentSchemaVersion);
```
itsjohncs, line 335: I think this and `return await this.downgrade_to(currentSchemaVersion)` is equivalent thanks to the fact that returning a promise from an async function implicitly chains that promise with the parent function's promise. But it strikes me as one of the more esoteric facts about JS and I'd think making `downgrade` a non-async function or awaiting `downgrade_to`'s promise would be clearer.

itsjohncs (approved):
Looks good. Happy to punt the work of deeper testing and making the CLI for when it's actually needed.

server/plugins/messageStorage/sqlite.ts
brunnre8: Yes.
There I was, being pleased with myself that I'm a real JS programmer that uses all the functional thingies....

okayokay, I'm gonna rewrite it with a plain boring loop then :wink:

server/plugins/messageStorage/sqlite.ts
brunnre8: ok, I'll await the result then... don't really want to switch to promises if I don't have to

server/plugins/messageStorage/sqlite.ts
brunnre8: done

PR #4719, opened 2023-03-28
push: remove iOS warning

brunnre8:
iOS 16.4 introduced webpush, we can get rid of the special case in our settings panel.


0xallie:
It might be worth noting that you *must* add The Lounge to the home screen for web push notifications to work on iOS.

MaxLeiter (approved):

MaxLeiter:
@brunnre8 I'm happy to keep the warning but change the text to what @0xallie suggested, or merge as-is.

Issue #1830, opened 2017-12-12
Support some markdown in text input


YaManicKill:
So, I thought I would bring this to the group to see what people think. The idea would be that certain things would turn automatically into the formatted messages, so that users don't have to type ctrl+b etc.

So:

* \*test\* would go to *test*
* \*\*test\*\* would go to **test**
* \~test\~ would go to ~test~

I'm obviously not suggesting links and similar stuff, it would only be things that have an equivelant in irc.

Thoughts?

MiniDigger:
I like the idea of that, but it also could be unwanted for some users. would we be able to transparently show that \*test\* will be translated to *test*? like live transforming it in the input field or something like that? I think that would be better UX wise. 
for example, when you input \*test\*, as soon as you type out the second *, it could transform into \**test\**, keeing the * but transforming the content.

christer88:
I'm not so sure about this. It's something which will be shown in TheLounge and maybe a few other clients, but it's not universal. Think we should keep it to the standards for IRC as much as possible.

YaManicKill:
> I like the idea of that, but it also could be unwanted for some users. 

See, I thought about this, but then I realised that Slack has been doing this for ages and I've never seen anyone complain about it

> for example, when you input *test*, as soon as you type out the second *, it could transform into *test*, keeing the * but transforming the content.

Oh wait, *keeping* the *. Actually yeah, that could work. 

> I'm not so sure about this. It's something which will be shown in TheLounge and maybe a few other clients, but it's not universal. Think we should keep it to the standards for IRC as much as possible.

No, it would translate to the IRC standards for making something bold, etc. I'm not suggesting that any messages that come in from other clients that have * around them would be translated, it's just that our input would send it as that.

astorije:
I was thinking about that too, but in a different way: what if The Lounge interprets Markdown in messages received? So that would not temper messages that I send to the server, but would display when people use Markdown.

It's not uncommon for example in programming-related channels that people use backticks around code snippets and such. TL could interpret into code.

This is rather intrusive and it might not work everywhere, so I think this grant a client-side option, but as people are more and more used to type Markdown to send messages (like Slack), that could turn out to look nice.
This needs some experimentations though, so turning off the setting by default, at least at first. I'm curious how my chats would look like on a daily basis with that.

[...]

PR #4725, opened 2023-04-23
feature: Add client-side parsing of inline markdown

xyzshantaram:
This PR introduces a client-side feature to parse inline markdown in messages and render it. I've added a settings toggle for it (Render inline markdown in messages), and the feature is off by default to avoid breaking anyone's workflow.

Settings toggle

It **DOES NOT** insert IRC control characters or otherwise modify the message being sent out, it is purely cosmetic. 

Markdown is quite widely used and there's is some amount of support for adding it to The Lounge as far as I can tell from previous GitHub discussion here (#1830). People also seem to use things like asterisks and backticks a lot on various channels for emphasis and code. Additionally, unlike the IRC control characters to turn text italic, bold, etc, Markdown gracefully degrades to plain text for people on clients which don't support those features, which imo is more sensible than losing the meaning entirely on a client that doesn't support that kind of formatting.

The parsing code is definitely inefficient in my opinion but does not introduce any noticeable delay in using the app even in large buffers.

Here's a screenshot of the normal behaviour, with the setting turned off:

Parsing disabled

Here's a screenshot with parsing turned on:

image


xyzshantaram:
I may also add support for quotes (i.e. messages that start with a > sign). They could be styled similarly to Discord or Slack's quoting.

xyzshantaram:
Quick update: added a settings toggle for showing the raw markdown characters in the rendered output. I think this should make both camps (i.e. "show exactly what was sent" vs "parse the markdown") happy.

Settings toggle

Feature screenshot

brunnre8:
why are you rebasing master *onto* your topic branch?
You don't have to merge back unrelated changes in any way. It's not helpful, especially not for reviewing

xyzshantaram:
@brunnre8 The contributing guidelines mentioned to rebase on the master branch, so that's what I've been doing. Sorry if I've misunderstood in some way.

brunnre8:
When asked to, if that disclaimer isn't there we need to update the docs.

But you didn't do that, you rewrote the commits frommerge_base..master onto your topic branch rather than reparenting your topic branch.

Look at the graph to see what I mean, `git log --oneline --graph --all`

xyzshantaram:
Crap, that's my bad, I'm sorry. I'm quite new to using git. Should I fix it
properly and force push, or cherry pick my commits onto a fresh branch?

brunnre8:
fix it up and force push, yes.
And no worries, happens

client/components/ParsedMessage.vue
```diff
@@ -11,12 +19,50 @@ export default defineComponent({
...
+						}
+					}
+				}
+
+				return result;
+			};
+
+			const toParse = "".concat(...generateStandIns(parsed));
+			return rehydrate(parseMd(toParse, this.store.state.settings.renderMdSrc), htmls);
+		}
```
SimplyLinn, line 63: I feel this logic is a bit malplaced.

I think it'd be more in the spirit of what this component logic already does (call a single parse function) if you were to break out this logic into its own function which takes "parsed" as an input parameter, as well as other necessary inputs:
````js
// FILE: components/ParsedMessage.vue

import parseMarkdown from "../js/helpers/mdHelpers";
// [...]
{
  // [...]
  render(context) {
    const parsed = parse(
	  typeof context.text !== "undefined" ? context.text : context.message.text,
      context.message,
      context.network
	);
	if (this.store.state.settings.parseMd) {
      return parseMarkdown(parsed, context.message, this.store.state.settings.renderMdSrc);
	}
	return parsed;
  }
  // [..]
}

// FILE: js/helpers/mdHelper
// [...]
export default function parseMarkdown(parsed, message, renderMdSrc) {
  // no-op on non-message type messages
  if (message?.type !== "message") return input;
  // [...] - rest of logic
}
```

It declutters the component render function, the new function is responsible for deciding whether a message should go through the parser or not (by returning early), and the only logic remaining in the component is checking the parseMd flag in order to know if we should call the markdown parser at all.

xyzshantaram: Just saw this, think it got lost in my inbox. Great idea, thanks! I split it off like you suggested.

xyzshantaram:
Hi, sorry to ping a dead thread - Is there anything else that's needed from my end to move this PR closer to merging?

brunnre8:
Frankly, I'm not sure I like the functionality much. So I'm unsure if the added complexity would be worth it.
As said though, that's a personal opinion.

What I can say though is that the code is slightly crazy... Why are you doing that "rehydrate" dance and such?
Just hook into an actual parser, where we still are string based and omit html elements, rather than taking the output of parse  as input.

Your parser is also broken... tests might have helped you.
```js
// strips ** when it's not valid markdown
> parseMd('**testytest', false)
"<span class='irc-italic'></span>testytest"
// manages for * though
> parseMd('*testytest', false)
'*testytest'
```

PR #4728, opened 2023-04-29
Client/Settings: Remove bogus settings-form prop

brunnre8:
As is this has no effect, other than looking weird in html:
`<div settings-form="[object HTMLFormElement]"><div><h2>Native app</h2>`


brunnre8:
@MaxLeiter git blame says you are the author, please have a look

MaxLeiter (approved):
can't remember why I added the ref but LGTM

PR #4737, opened 2023-05-17
Prune old events from sqlite database

### LINKED ISSUE #2822: Logging - limit disk usage of logs/<username>.sqlite files, remove old records
### Feature Description
Currently, there's no way how to control logs/<username>.sqlite file size. It can grow forever. Although this is usually not an issue, it will be good to have a mechanism that will keep file size under control.

### Use-cases

- deployments with many users
- deployments in public/private clouds (on OpenShift), where persistent storage for applications is usually limited

### Ideas

- keep max N messages in tables, use maxHistory parameter and run regular cleanups (SQL - delete oldest entries)
- check sqlite vacuum feature and check if this can help to improve disk space usage of sqlite

azeemba:
Adds a new config `dbHistoryDays` which defaults to undefined.

At start up, each database handler reads the config. If the value is set, then a reoccuring task is scheduled to clean up old events. Events older than `dbHistoryDays` are targeted but only a few thousand events are cleaned up per iteration to avoid freezing the server.

Adds unit tests to validate the cleanup logic as well.

Fixes #2822

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -60,14 +60,6 @@ module.exports = {
 	// This value is set to `10000` by default.
 	maxHistory: 10000,
 
-	// ### `dbHistoryDays`
-	//
-	// Defines the maximum number of days of history to store in the database.
-	// Undefined/-1/0 is treated an unlimited.
-	// The limit is seen as a soft target but not an exact goal. Only a few
-	// thousand rows are pruned at a time to avoid slowing down the service.
-	dbHistoryDays: undefined,
-
 	// ### `https`
 	//
 	// These settings are used to run The Lounge's web server using encrypted TLS.
@@ -311,6 +303,16 @@ module.exports = {
 	//
 	// This value is set to `["sqlite", "text"]` by default.
 	messageStorage: ["sqlite", "text"],
+
+	// ### `sqliteConfig`
+	//
+	// `maxDaysHistory`:
+	// 		Defines the maximum number of days of history to keep in the database.
+	// 		Undefined/-1/0 is treated an unlimited.
+	//		Users can overwrite setting to make the duration shorter
+	sqliteConfig: {
+		maxDaysHistory: undefined,
+	},
 
 	// ### `useHexIp`
 	//
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -60,6 +60,14 @@ module.exports = {
 	// This value is set to `10000` by default.
 	maxHistory: 10000,
 
+	// ### `dbHistoryDays`
+	//
+	// Defines the maximum number of days of history to store in the database.
+	// Undefined/-1/0 is treated an unlimited.
+	// The limit is seen as a soft target but not an exact goal. Only a few
+	// thousand rows are pruned at a time to avoid slowing down the service.
+	dbHistoryDays: undefined,
+
 	// ### `https`
 	//
 	// These settings are used to run The Lounge's web server using encrypted TLS.
--- a/server/config.ts
+++ b/server/config.ts
@@ -76,6 +76,12 @@ type Debug = {
 	raw: boolean;
 };
 
+type SqliteConfig =
+	| {
+			maxDaysHistory: number | undefined;
+	  }
+	| undefined;
+
 export type ConfigType = {
 	public: boolean;
 	host: string | undefined;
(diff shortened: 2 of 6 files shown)


brunnre8 (changes requested):
Thanks a lot for taking this on.
Now, I've some general comments as to how I think this could work, plus a few minor comments in the code directly.

I think this is a bit too limited as is.
For starters, I want the history retention to be configurable by the user, not just the admin.
While in many cases that's the same, it isn't in all cases and not having to edit the server config for something like that is better.

What that means is that we need this to be a hierarchical configuration. Admins sets global limits as they see fit, user can lower that limit as they feel like.

We should also put that configuration into its own top level config object so that we can add more related settings in the future, for instance I would like to add the ability to only delete unimportant status messages (leave, part, nick change, away, back etc) while retaining the actual useful stuff indefinitely (privmsg, notice).

Does this make sense?

Also note, you should probably rebase on top of https://github.com/thelounge/thelounge/pull/4718 as I'll merge that soon (after a bugfix release for 4.4.0)

fnutt:
> I think this is a bit too limited as is. For starters, I want the history retention to be configurable by the user, not just the admin. While in many cases that's the same, it isn't in all cases and not having to edit the server config for something like that is better.
> 
> What that means is that we need this to be a hierarchical configuration. Admins sets global limits as they see fit, user can lower that limit as they feel like.

I'm not sure why a user would set it lower than the admin global limit. What is the usecase for this?

brunnre8:
Admin says, yeah well, storage isn't an issue.
User says, out of privacy issues I'm uncomfortable to have history there for longer than a month.

Easy enough to imagine, no?

server/plugins/messageStorage/sqlite.ts
```diff
@@ -44,18 +44,36 @@ class Deferred {
...
-	constructor(userName: string) {
+	constructor(userName: string, clientStorageConfig: SqliteConfig) {
 		this.userName = userName;
 		this.isEnabled = false;
 		this.initDone = new Deferred();
+
+		this.mergedConfig = this.mergeClientConfig(clientStorageConfig);
+	}
+
+	mergeClientConfig(clientStorageConfig: SqliteConfig) {
```
brunnre8, line 66: This isn't the right place for that kind of logic.
The config needs to be merged by the core code, not the sqlite specific storage backend.

Yes, currently sqlite is our only implementation of the interface, but that doesn't mean that we should violate the layers.

server/config.ts
```diff
@@ -76,14 +76,19 @@ type Debug = {
 	raw: boolean;
 };
 
+type SqliteConfig =
+	| {
+			maxDaysHistory: number | undefined;
+	  }
+	| undefined;
+
```
brunnre8, line 84: that type seems odd to me?

You shouldn't set that to an object or undefined, things that want it optional shall say so when they mean it.

Meaning, the user of the type should do 
```typescript
export type ConfigType = {
 // ...
sqliteConfig?: SqliteConfig;
// ...
}
```
Or whatever, but not that funny type that is intrinsically this way.
That's a hassle to use

defaults/config.js
```diff
@@ -312,6 +304,16 @@ module.exports = {
 	// This value is set to `["sqlite", "text"]` by default.
 	messageStorage: ["sqlite", "text"],
 
+	// ### `sqliteConfig`
+	//
+	// `maxDaysHistory`:
+	// 		Defines the maximum number of days of history to keep in the database.
+	// 		Undefined/-1/0 is treated an unlimited.
```
brunnre8, line 311: exactly undefined
We should have one way and only one way of configuring something.

test/plugins/sqlite.ts
```diff
@@ -37,7 +37,7 @@ describe("SQLite Message Storage (stateful tests)", function () {
 		// Delete database file from previous test run
 		await cleanup();
 
-		store = new MessageStorage("testUser");
+		store = new MessageStorage("testUser", undefined as any);
```
brunnre8, line 40: such casts are very brittle... if you can avoid so please do and in this case you trivially could by providing a correct default value

brunnre8 (changes requested):
Thanks for the update.
In my opinion this should still be restructured a bit though, tell me if you disagree with my comments and I try to explain my reasoning a bit better.

azeemba:
I appreciate the feedback but this has slipped in priority for me. Happy for someone to take over the PR or to close the PR.

brunnre8:
thanks for your efforts

PR #4739, opened 2023-05-23
client: fix password change input

brunnre8:
The TS rewrite dropped the form that was expected to be passed as props.
That lead to the password change being borked, as the fields were always set to "null".
We don't need a form, can just use refs here.

### DIFF
--- a/client/components/Settings/Account.vue
+++ b/client/components/Settings/Account.vue
@@ -118,7 +118,7 @@ export default defineComponent({
 		const store = useStore();
 
 		const passwordErrors = {
-			missing_fields: "Please enter a new password",
+			missing_fields: "Please fill in all fields",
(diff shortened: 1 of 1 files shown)


brunnre8:
Would be nice to get this relatively quickly in, so that we can add this to the bug fix before anyone else notices...

brunnre8:
CI runners are having a bad day... checks are passing though

MindTooth:
Just to ask, could this be the case that I can't log back in after I tried changing password using the UI?  Now nothing works even after changing the password using the CLI.  If so, this is quite an important fix.  Now I'm relying on past valid sessions. 🤣

brunnre8:
@MindTooth no, but please use IRC for such questions and not our PRs

PR #4742, opened 2023-05-29
issue ban before kick in kickban

examknow:
Common practice for kickbans is to issue the ban first to avoid a race condition with the kicked person's client auto-rejoining the channel before the banning client is able to set the ban.


brunnre8:
I don't think this matters

We only issue a nick based ban after all which is trivial to circumvent by the client.
For anything you really care about you need to manually ban in any case

brunnre8:
while I appreciate the effort, I'm going to close this as I don't think it really gains much.

It'd be better to actually fix kickban so that it uses a better ban mask, *then* we can make sure we ban before we kick.

PR #4749, opened 2023-06-10
Prevent autocomplete space following username at login

Kufat:
No popup because this usually happens for reasons that aren't the user's fault.


brunnre8:
Thank you for the PR.
I don't think this is the right approach though, we should just switch the user input field to the normal v-model and let vue do it with the .trim modifier on it.
I've already PRed the rewrite https://github.com/thelounge/thelounge/pull/4753 so all that you need to do is add the trim modifier onto it.

PR #4767, opened 2023-07-29
sqlite: add full text search index

brunnre8:
Our search was using a linear search over the message text data.
This is very inefficient for large databases and barely usable.
We can add an FTS index, trading storage for speed.

As it's setup, this only supports English, but then we get fancy
stemming so that say "work" matches itself as well as "working".

This could be reduced to just stripping funny chars, with less good
search in the English case.

### DIFF
--- a/server/plugins/messageStorage/sqlite.ts
+++ b/server/plugins/messageStorage/sqlite.ts
@@ -26,7 +26,7 @@ try {
 type Migration = {version: number; stmts: string[]};
 type Rollback = {version: number; rollback_forbidden?: boolean; stmts: string[]};
 
-export const currentSchemaVersion = 1703322560448; // use `new Date().getTime()`
+export const currentSchemaVersion = 1706445634804; // use `new Date().getTime()`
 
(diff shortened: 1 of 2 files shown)


brunnre8:
Now, this is the bare bones version that just supports normal phrases, no fancy wildcards, no prefix matches / and / or logic etc.

FTS5 supports all that though so it's just a matter of coming up with a sane syntax and writing the parser for it:

https://www.sqlite.org/fts5.html

MaxLeiter (approved):

brunnre8:
Don't merge as is quite yet, this will not merge cleanly with the current master branch due to a security feature iirc...

Need to rebase and relax the checks

brunnre8:
Found a bug, my triggers aren't doing what I hope they would be doing... need to adapt them

brunnre8:
So, fixed the obvious bugs... @MaxLeiter would you mind re-running it on your machine and activate the storage cleaner with "everything" for a bit?

Just to see if pragma trusted_schema might be needed or not?

PR #4776, opened 2023-09-01
Block Referer header from being sent when clicking links

gaymersriseup:
Before this change, any links a user clicks on will be sent the full The Lounge URL the user is coming from. While not too big of an issue, especially since URL fragments (channel IDs) aren't sent, this provides a little bit of a privacy improvement.


brunnre8:
It should most definitely not change a thing, how did you test this?

We already set the policy https://github.com/thelounge/thelounge/blob/9f05a75c39dc2a1ad9d0befafec1e11e1030c20d/server/server.ts#L400

PR #4779, opened 2023-09-02
Added a channel (or network) toggle to notify on all messages

antoniomika:
This PR adds a toggle on channels (or networks) to notify on all messages. 

image
and 
image

I probably should've asked for comment on what to use for the icon and toggle text. As of now, the icon is [volume-up](ttps://fontawesome.com/v5.15/icons/volume-up?style=solid) and the text is:

When all notifications enabled:
```
Don't notify on all messages for ${chanType}
```

When all notifications disabled:
```
Notify on all messages for ${chanType}
```

`${chanType}` is either `channel` or `network`.

### DIFF
--- a/client/js/socket-events/msg.ts
+++ b/client/js/socket-events/msg.ts
@@ -105,7 +105,11 @@ function notifyMessage(
 		return;
 	}
 
-	if (msg.highlight || (store.state.settings.notifyAllMessages && msg.type === "message")) {
+	if (
+		channel.notifyAll ||
+		msg.highlight ||
+		(store.state.settings.notifyAllMessages && msg.type === "message")
+	) {
 		if (!document.hasFocus() || !activeChannel || activeChannel.channel !== channel) {
 			if (store.state.settings.notification) {
 				try {
--- a/server/plugins/irc-events/message.ts
+++ b/server/plugins/irc-events/message.ts
@@ -177,7 +177,11 @@ export default <IrcEventHandler>function (irc, network) {
 		chan.pushMessage(client, msg, !msg.self);
 
 		// Do not send notifications if the channel is muted or for messages older than 15 minutes (znc buffer for example)
-		if (!chan.muted && msg.highlight && (!data.time || data.time > Date.now() - 900000)) {
+		if (
+			!chan.muted &&
+			(msg.highlight || chan.notifyAll) &&
+			(!data.time || data.time > Date.now() - 900000)
+		) {
 			let title = chan.name;
 			let body = cleanMessage;
 
--- a/server/server.ts
+++ b/server/server.ts
@@ -794,6 +794,38 @@ function initializeClient(
 
 			client.save();
 		});
+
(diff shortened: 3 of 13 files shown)


client/js/socket-events/msg.ts
```diff
@@ -105,7 +105,11 @@ function notifyMessage(
 		return;
 	}
 
-	if (msg.highlight || (store.state.settings.notifyAllMessages && msg.type === "message")) {
+	if (
+		channel.notifyAll ||
```
brunnre8, line 109: This triggers on way too many things no?

Do you *really* want a notification on every single join / away / topic / mode and what have you?

server/plugins/irc-events/message.ts
```diff
@@ -177,7 +177,11 @@ export default <IrcEventHandler>function (irc, network) {
 		chan.pushMessage(client, msg, !msg.self);
 
 		// Do not send notifications if the channel is muted or for messages older than 15 minutes (znc buffer for example)
-		if (!chan.muted && msg.highlight && (!data.time || data.time > Date.now() - 900000)) {
+		if (
+			!chan.muted &&
+			(msg.highlight || chan.notifyAll) &&
```
brunnre8, line 182: This conditional was already somewhat large and I think by now we need to structure this a bit differently, or at the very least pull that out into its own function.

We have pretty much a similar conditional up in the socket events

server/server.ts
```diff
@@ -794,6 +794,38 @@ function initializeClient(
...
+			}
+
+			for (const attachedClient of Object.keys(client.attachedClients)) {
+				manager!.sockets.in(attachedClient).emit("notifyAll:changed", {
+					target,
+					status: setNotifyAllTo,
+				});
+			}
+
+			client.save();
```
brunnre8, line 827: Hrm... considering that this function is more or less exactly the muted one... can we clean this up ?

brunnre8 (changes requested):
Thanks a lot for taking this on, much appreciated.

Now... I do think we've outlived the context menu a bit for notification settings. Adding another option makes it too long in my opinion.
Plus the text makes it... somewhat huge don't you think?

I think it'll be better to actually create a UI for those kinds of things that can then also hold other channel specific stuff.

Then... you pretty much copied the muted code, but without deduplicating anything even though all that changes is an event and a field. Maybe this could be improved a bit?

Last, you notify on way too many things. Don't think status messages and such are really something you need a push notification off. What was your reasoning behind this, other than the literal interpretation of "all"?

antoniomika:
I'll look into what the UX could look like for adding it to the `notifications` part of settings.

My desire was that if someone want's all of the notifications, they want _all_ of the notifications pertaining to the model. We could probably make this more modular, but as you've stated there currently isn't a great place for that to be handled.

Regarding dedupe between this and mute, both features effective do the same thing. We could probably thing of a more general way to add a "toggle" to a specific model. I can look into what that'd look like as well. 

Appreciate the review!

PR #4799, opened 2023-11-04
Add sqlite cleanup functionality

### LINKED ISSUE #2822: Logging - limit disk usage of logs/<username>.sqlite files, remove old records

brunnre8:
Once this is getting hooked up, it'll periodically delete old
messages.

The StoragePolicy can be chosen by the user, currently there's
two versions, delete everything based on age is the obvious.

The other is for the data hoarders among us. It'll only delete
message types which can be considered low value... Types with
a time aspect like away / back... joins / parts etc.

It tries to do that in a sensible way, so that we don't block
all other db writers that are ongoing.
The "periodically" interval is by design not exposed to the user.

Fixes: https://github.com/thelounge/thelounge/issues/2822

### DIFF
--- a/server/command-line/storage.ts
+++ b/server/command-line/storage.ts
@@ -3,6 +3,7 @@ import {Command} from "commander";
 import ClientManager from "../clientManager";
 import Utils from "./utils";
 import SqliteMessageStorage from "../plugins/messageStorage/sqlite";
+import {StorageCleaner} from "../storageCleaner";
 
 const program = new Command("storage").description(
 	"various utilities related to the message storage"
@@ -10,7 +11,7 @@ const program = new Command("storage").description(
 
 program
 	.command("migrate")
-	.argument("[user]", "migrate a specific user only, all if not provided")
+	.argument("[username]", "migrate a specific user only, all if not provided")
 	.description("Migrate message storage where needed")
 	.on("--help", Utils.extraHelp)
 	.action(function (user) {
@@ -20,7 +21,19 @@ program
 		});
 	});
 
-async function runMigrations(user: string) {
+program
+	.command("clean")
+	.argument("[user]", "clean messages for a specific user only, all if not provided")
+	.description("Delete messages from the DB based on the storage policy")
+	.on("--help", Utils.extraHelp)
+	.action(function (user) {
+		runCleaning(user).catch((err) => {
+			log.error(err.toString());
+			process.exit(1);
+		});
+	});
+
+async function runMigrations(user?: string) {
 	const manager = new ClientManager();
 	const users = manager.getUsers();
 
@@ -63,6 +76,48 @@ function isUserLogEnabled(manager: ClientManager, user: string): boolean {
 	}
 
 	return conf.log;
+}
+
+async function runCleaning(user: string) {
+	const manager = new ClientManager();
+	const users = manager.getUsers();
+
+	if (user) {
+		if (!users.includes(user)) {
+			throw new Error(`invalid user ${user}`);
+		}
+
+		return cleanUser(manager, user);
+	}
+
+	for (const name of users) {
+		await cleanUser(manager, name);
+		// if any migration fails we blow up,
+		// chances are the rest won't complete either
+	}
+}
+
+async function cleanUser(manager: ClientManager, user: string) {
+	log.info("handling user", user);
+
+	if (!isUserLogEnabled(manager, user)) {
+		log.info("logging disabled for user", user, ". Skipping");
+		return;
+	}
+
+	const sqlite = new SqliteMessageStorage(user);
+	await sqlite.enable();
+	const cleaner = new StorageCleaner(sqlite);
+	const num_deleted = await cleaner.runDeletesNoLimit();
+	log.info(`deleted ${num_deleted} messages`);
+	log.info("running a vacuum now, this might take a while");
+
+	if (num_deleted > 0) {
+		await sqlite.vacuum();
+	}
+
+	await sqlite.close();
+	log.info(`cleaning messages for ${user} has been successful`);
 }
 
 export default program;
--- a/server/storageCleaner.ts
+++ b/server/storageCleaner.ts
@@ -1,0 +1,148 @@
+import SqliteMessageStorage from "./plugins/messageStorage/sqlite";
+import {MessageType} from "./models/msg";
+import Config from "./config";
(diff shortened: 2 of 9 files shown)


brunnre8:
This does not yet contain a way for a client to make the storage policy more stringent (meaning, delete more aggressively).
That is a feature we could easily add though, given the structure of it.

For now, that should be the bare minimum so people can finally turn their cronjobs off that manually mess with the db

brunnre8:
query example, for statusOnly:

```sql
delete from messages where id in (
    select id from messages where
    time <= 1698515398915
    and type in ('away','back','invite','join','kick','mode','mode_channel','mode_user','nick','part','quit','ctcp','ctcp_request','chghost','topic','topic_set_by')
    order by time asc
    limit 200
)
```

Now, this seems to drive up the cpu to 100% for quite a while on every cleaning cycle.

sqlite uses the time index, but the string comparison of the types is... bad.

We can create a new index for messages, on type.
That seems to get rid of the CPU spikes, using more storage and ignoring the time index.

Alternatively we can also stop sorting, we don't really need it. It's just a bit easier for the user to understand, when messages are deleted in order, so that they don't have funny wholes in conversations when the max age is very low (say a day)

brunnre8:
This is now running on my live instance, with the added index:

```sql
CREATE INDEX msg_type_idx on messages (type);
```

server/command-line/storage.ts
```diff
@@ -20,6 +21,18 @@ program
 		});
 	});
 
+program
+	.command("clean")
+	.argument("[user]", "clean messages for a specific user only, all if not provided")
```
MaxLeiter, line 26:
```suggestion
	.argument("[username]", "clean messages for a specific username only, all if not provided")
```

nit: best to be specific

server/command-line/storage.ts
```diff
@@ -65,4 +78,46 @@ function isUserLogEnabled(manager: ClientManager, user: string): boolean {
...
+
+	if (user) {
+		if (!users.includes(user)) {
+			throw new Error(`invalid user ${user}`);
+		}
+
+		return cleanUser(manager, user);
+	}
+
+	for (const name of users) {
```
MaxLeiter, line 93: I'm confused by this logic. We return above if the user is defined, and if no user is passed we clean all users?

server/storageCleaner.ts
```diff
@@ -0,0 +1,148 @@
...
+		return num_deleted;
+	}
+
+	private async runDeletes() {
+		if (this.isStopped) {
+			return;
+		}
+
+		if (!this.db.isEnabled) {
+			// TODO: remove this once the server is intelligent enough to wait for init
```
MaxLeiter, line 82: Can this be removed now that there's `await this.initDone.promise` in `deleteMessages`?

MaxLeiter (changes requested):

server/storageCleaner.ts
```diff
@@ -0,0 +1,148 @@
...
+			return;
+		}
+
+		clearTimeout(this.ticker);
+	}
+}
+
+function assertNoBadPolicy(_: never): never {
+	throw new Error(
+		`Invalid deletion policy "${Config.values.storagePolicy.deletionPolicy}", fix your config.`
```
MaxLeiter, line 146:
```suggestion
		`Invalid deletion policy "${Config.values.storagePolicy.deletionPolicy}". Fix under \`storagePolicy\` in your config.`
```

server/command-line/storage.ts
brunnre8: yes. The command defaults to all users, you can however run a specific one if you need to for whatever reason.

Usually the admin wants to migrate all users at the same time... so defaulting to that makes sense no?

server/storageCleaner.ts
brunnre8: then you are piling on to the blocked calls already in flight... is there a reason for doing so?
The whole initDone thing is a gross hack, because TL is using it when it should be waiting for init to complete.
Unsure when you start running into problems due to node having too much garbage on the even loop. Do you know?

server/command-line/storage.ts
```diff
@@ -65,4 +78,46 @@ function isUserLogEnabled(manager: ClientManager, user: string): boolean {
 	return conf.log;
 }
 
+async function runCleaning(user: string) {
```
MaxLeiter, line 81: let's make user optional then

```suggestion
async function runCleaning(user?: string) {
```

server/command-line/storage.ts
MaxLeiter: See suggestion above

server/command-line/storage.ts
brunnre8: agree on the param name, but we do it for a user, not a username. So let's meet in the middle and only convert the first, is that ok?

MaxLeiter: fixed

server/storageCleaner.ts
brunnre8: `Invalid deletion policy "${Config.values.storagePolicy.deletionPolicy}" in the \`storagePolicy\` object, fix your config.` ?

Sounds a bit more natural that way to me than yours. Point taken that we should point users into the right direction though.

server/storageCleaner.ts
MaxLeiter: No reason for doing so, I just saw the promise and saw the TODO and was making sure it wasn't old.

brunnre8: no, I meant the generally broken behavior of TL in regards to activating the storage backend and then immediately queuing inserts etc when the backend didn't had the chance to init.
It needs to buffer those internally until the storage is ready, then start using it, not before.

server/storageCleaner.ts
MaxLeiter: Yeah I’m fine with whatever. i just think we should provide some actionable item instead of “fix this”.

MaxLeiter (approved):

PR #4806, opened 2023-12-09
Update webmanifest to include screenshots

ghost:
I was not able to reproduce the issue mentioned locally but these changes should fix the problems mentioned. Please let me know if you approve of these changes.


client/thelounge.webmanifest
```diff
@@ -10,44 +10,59 @@
...
 			"type": "image/svg+xml",
-			"purpose": "maskable any",
-			"sizes": "513x513"
+			"purpose": "maskable",
+			"sizes": "150x150"
+		}
+	],
+	"screenshots": [
+		{
+			"src": "./screenshots/homescreen.png",
```
xPaw, line 55: Missed img/?

xPaw:
See https://web.dev/patterns/web-apps/richer-install-ui for more information. In chrome devtools, you can see warnings by going to Application > Manifest tab.

As for screenshots, I've found that on desktop they scale very poorly, so it needs to be zoomed in a lot. As well as private information such as IPs from your join message, and the user list should not include random users (for their sake). Perhaps a few messages should be sent to show link previews etc.

client/thelounge.webmanifest
jstjnsn: Ah yes, thank you, i will change it.

brunnre8:
Considering that the screenshots are meant to showcase the app a bit, showing the server window is probably not quite the right thing to do.
Let's synthesize some interesting content in some test / dummy chan instead.

Link to https://thelounge.chat, hold a convo with some dummy users or whatever.
Does this make sense?

Essentially the goal is to be as realistic as possible, showcasing the features, without putting anyone's privacy at risk that didn't give their consent.

Feel free to use #thelounge-test, that's just "wendy" and I, both are non secret so feel free to bug wendy with commands / expose our IPs etc.

both are joined from hosts I control.

jstjnsn:
@brunnre8 I'm not sure I understand, I tried to join thelounge-test channel and chat, but I am not getting any responses. Is it good enough if I just add the following?
- screenshot of the homepage login window
- screenshot of the thelounge-test channel with a message I sent
anything else?

brunnre8:
Do you understand *why* the linter yells at us?
As in, can you follow the reasoning as to why we need the pictures?

I don't understand why you'd want to show a login screen? It's not interesting.

You want to have pictures which are appealing, showcasing what the app can do.
Having a channel with one message doesn't do the trick.

https://thelounge.chat/img/thelounge-screenshot.png

^ this is an example from our webpage, see it showcases some fun things you can't do in other clients... audio, link preview etc.
Multiple users... Those things.

THAT is what this issue is about, not just putting some random images into our repo to shut the linker up.

PR #4818, opened 2024-01-18
Add web+ prefix to linkified URI schemes

SoniEx2:
Briefly discussed on the IRC. For tokodon/tuba/feditext and future web-based protocols.

(Also they are called schemes, as per RFC 3986 and whatnot. Whether to be consistent with the RFC or with the linkify library tho, :shrug: )

### DIFF
--- a/shared/linkify.ts
+++ b/shared/linkify.ts
@@ -11,6 +11,8 @@ export type LinkPart = {
 	link: string;
 };
 
+const webplus_scheme_chars = /^[a-z]+:/i;
+
 LinkifyIt.prototype.normalize = function normalize(match: NoSchemaMatch) {
 	match.noschema = false;
 
@@ -24,6 +26,10 @@ LinkifyIt.prototype.normalize = function normalize(match: NoSchemaMatch) {
 		match.schema = "http:";
 		match.url = "http:" + match.url;
 		match.noschema = true;
+	}
+
+	if (match.schema === "web+") {
+		match.schema = "web+" + match.url.slice(4).match(webplus_scheme_chars)![0];
 	}
 
 	if (match.schema === "mailto:" && !/^mailto:/i.test(match.url)) {
@@ -54,6 +60,25 @@ const commonSchemes = [
 for (const schema of commonSchemes) {
 	linkify.add(schema + ":", "http:");
 }
+
+linkify.add("web+", {
+	validate(text: string, pos: number, self: LinkifyIt.LinkifyIt) {
+		const tail = text.slice(pos);
+
+		if (webplus_scheme_chars.test(tail)) {
+			const offset = tail.match(webplus_scheme_chars)![0].length;
+			const result = self.testSchemaAt(text, "http:", pos + offset);
+
+			if (result === 0) {
+				return 0;
+			}
+
+			return offset + result;
+		}
+
+		return 0;
+	},
+});
 
 export function findLinks(text: string) {
 	const matches = linkify.match(text) as NoSchemaMatch[];
--- a/test/shared/findLinks.ts
+++ b/test/shared/findLinks.ts
@@ -372,4 +372,34 @@ describe("findLinks", () => {
 
 		expect(actual).to.deep.equal(expected);
 	});
+
+	it("should find web+ap urls", () => {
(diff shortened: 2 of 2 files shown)


shared/linkify.ts
```diff
@@ -11,6 +11,8 @@ export type LinkPart = {
 	link: string;
 };
 
+const webplus_scheme_chars = /^[a-z]+:/i;
```
brunnre8, line 14: this is an invalid regex... Will match web+SpOnGeBoB: which is not valid

> The custom scheme has only lowercase [ASCII](https://developer.mozilla.org/en-US/docs/Glossary/ASCII) letters in its name.

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler#permitted_schemes

shared/linkify.ts
```diff
@@ -26,6 +28,10 @@ LinkifyIt.prototype.normalize = function normalize(match: NoSchemaMatch) {
 		match.noschema = true;
 	}
 
+	if (match.schema === "web+") {
+		match.schema = "web+" + match.url.slice(4).match(webplus_scheme_chars)![0];
+	}
+
```
brunnre8, line 34: This doesn't need to be in the global function... you can add a normalize function to the protocol registration

shared/linkify.ts
```diff
@@ -55,6 +61,25 @@ for (const schema of commonSchemes) {
 	linkify.add(schema + ":", "http:");
 }
 
+linkify.add("web+", {
+	validate(text: string, pos: number, self: LinkifyIt.LinkifyIt) {
+		const tail = text.slice(pos);
+
+		if (webplus_scheme_chars.test(tail)) {
+			const offset = tail.match(webplus_scheme_chars)![0].length;
```
brunnre8, line 69: that assertion isn't valid...

```js
const webplus_scheme_chars = /^[a-z]+:/i;
console.log("web+://invalid.example");
console.log("://invalid.example".match(webplus_scheme_chars));
// web+://invalid.example
// null

```

brunnre8 (changes requested):

shared/linkify.ts
SoniEx2: that documentation is for the call to registerProtocolHandler, not for the `href`s/links themselves (which get normalized by the browser before being passed onto the OS/the handler).

shared/linkify.ts
SoniEx2: that causes an unfixable type error.

shared/linkify.ts
SoniEx2: the assertion is valid because it is wrapped in a conditional of `webplus_scheme_chars.test(tail)`

brunnre8: Ah, right. Sorry

shared/linkify.ts
brunnre8: what do you mean?

shared/linkify.ts
brunnre8: well, https://html.spec.whatwg.org/multipage/iana.html#web+-scheme-prefix agrees, so while my link was poorly chosen, the statement is still true I think

shared/linkify.ts
SoniEx2: due to the `noschema` property. the protocol registration provides an entire replacement to the global function, so you'd have to set `noschema` in the protocol registration normalize function too.

shared/linkify.ts
SoniEx2: we can refer to RFC 3986, section 3.1: https://www.rfc-editor.org/rfc/rfc3986#section-3.1

> Scheme names consist of a sequence of characters beginning with a letter and followed by any combination of letters, digits, plus ("+"), period ("."), or hyphen ("-").  Although schemes are case- insensitive, the canonical form is lowercase and **documents that specify schemes must do so with lowercase letters**.

(emphasis added)

there is also:

> An implementation should accept uppercase letters as equivalent to lowercase in scheme names (e.g., allow "HTTP" as well as "http") for the sake of robustness but should only produce lowercase scheme names for consistency.

and in practice this is what browsers do, for `a href` they normalize the scheme name before processing it further.

PR #4821, opened 2024-01-18
Add link previews for fedilinks

SoniEx2:
Depends on/includes #4818 

Briefly mentioned on the IRC.

### DIFF
--- a/server/plugins/irc-events/link.ts
+++ b/server/plugins/irc-events/link.ts
@@ -506,7 +506,31 @@ function fetch(uri: string, headers: Record<string, string>) {
 
 function normalizeURL(link: string, baseLink?: string, disallowHttp = false) {
 	try {
-		const url = new URL(link, baseLink);
+		let url = new URL(link, baseLink);
+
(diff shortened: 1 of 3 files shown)


brunnre8:
I don't think we should implement this protocol until it got some more widespread discussion.

web+ is a namespace available to everyone: https://html.spec.whatwg.org/multipage/iana.html#web+-scheme-prefix

The implemented redirect to '/.well-known/protocol-handler' has not been part of any discussion within the whatwg group as far as I can tell.

Considering that this hasn't happened and that the namespace that's grabbed is huge, I NACK this PR.
web+coffee might not want to have this redirect semantics, if some lower level web+ap for instance wants to agree on doing so they can.

Alternatively this should get discussed and agreed upon in the proper fora, by a wide variety of stakeholders.

SoniEx2:
we still maintain that this *link preview protocol* only affects *participating* hosts and therefore it's okay.

if a host doesn't want the link previews they can just not implement the protocol.

note that this is not a redirect. the link previews still link to the original web+ uri, the protocol is only used for retrieving the preview cards.

PR #4824, opened 2024-01-21
linkify: simplify noscheme detection logic

brunnre8:
Overriding the built in is poor form, as this prevents adding a new type handler with its own normalize handler.

We only ever want to override protocol-less URLs to http, so we just do so explicitly in the "//" schema normalizer.

This also means that we don't need all that type conversion dance, we simply set the schema to null when we patch it and filter on the schema directly

Take 2, as the first approach wasn't working.

### DIFF
--- a/shared/linkify.ts
+++ b/shared/linkify.ts
@@ -1,34 +1,10 @@
 import LinkifyIt, {Match} from "linkify-it";
 import tlds from "tlds";
 
-export type NoSchemaMatch = Match & {
-	noschema: boolean;
-};
-
 export type LinkPart = {
 	start: number;
 	end: number;
 	link: string;
-};
-
-LinkifyIt.prototype.normalize = function normalize(match: NoSchemaMatch) {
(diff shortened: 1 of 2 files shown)


brunnre8:
"code scanning" errors are bogus, it's used as input to linkify.match, which doesn't treat it as a regex.

shared/linkify.ts
```diff
@@ -73,12 +49,25 @@ linkify.add("web+", {
...
+});
+
+// we must rewrite protocol less urls to http, else if TL is hosted
+// on https, this would incorrectly use https for the remote link.
+// See https://github.com/thelounge/thelounge/issues/2525
+//
+// We take the validation logic from linkify and just add our own
+// normalizer.
+linkify.add("//", {
+	validate: (linkify as any).__schemas__["//"].validate,
```
brunnre8, line 62: nasty? Yes. This is using a private implementation detail.
But our tests will catch it, should this every blow up.

PR #4828, opened 2024-01-27
Replace yarn with pnpm

MaxLeiter:
It's time to move on from yarn 1.

Not that I contribute enough to care much, but it was frustrating to use yarn while setting my local repo up again. Happy to go back to npm / upgrade to a newer version of yarn too.


brunnre8:
Hrm...
This also needs changes for the package logic within TL where yarn is used as well.

This also needs corresponding PRs to the deb package generation as well.

As is this is incomplete and not mergable I think

brunnre8:
Fwiw I have 0 reasons not to change from yarn.

Don't think npm is much better though, both are shitty (or at least were historically).

What made ypu go forn pnpm?

brunnre8 (changes requested):
As already mentioned, this needs some additional work.
Marking as such so that the UI is easier to filter

PR #4834, opened 2024-02-01
server: remove version from CTCP response

### LINKED ISSUE #1344: Allow hiding version number from CTCP VERSION response.

### LINKED ISSUE #4785: CTCP VERSION response handling
We have a bug open that wants to hide the TL version from ctcp: https://github.com/thelounge/thelounge/issues/1344
As well as a corresponding PR implementing that. The PR also allows sending any garbage response of the admins choice.

However, the reasoning given in the PR while this seems worthwhile is questionable to me.

If one is concerned that people may figure out a security bug in TL by seeing the version, just sending TL without a version doesn't help whatsoever. An attacker can simply play back any known CVE since the beginning of time until one sticks.

Tracking users seems to be a non concern, considering that most will be using the latest release and that's a pretty common string therefore.

Adding garbage there defeats the whole point of ctcp version in the first place and I don't think we should humor that. If people want to do so, they should patch the code.

Now, my personal preference is to remove the ctcp version response altogether.

Responding to external probing without the user giving their consent is not a thing I want any software to do.
I've turned off any ctcp responses but action on both libera and oftc and it works just fine.
All of emersions clients (gamja, soju) don't carry ctcp version and no one complained (and that's the bouncer sr.ht uses with quite the user base).

We stopped doing [finger](https://en.wikipedia.org/wiki/Finger_(protocol)) too and nobody died :wink: people will survive without it.

@MaxLeiter @itsjohncs what is your opinion here?

Essentially here's our options:
1) Keep it as is, ctcp version is useful
2) Allow hiding the version string, stops being useful for any sensible use of the string (security scanning of net ops, support chan)
3) allow full override, as implemented by the PR or something similar.
   The response is now useless for everything it was intended. Can't do scanning as netops, can't use it for support.
4) Stop responding to anything but ctcp action. Trivial to do, solves all the security concerns, principle of least surprise as one can't be probed by random people.

flotwig:
Closes #1344 
Closes #4785 (?)

Instead of making this configurable as in #4780, just remove the version number altogether. See discussion: https://github.com/thelounge/thelounge/issues/4785#issuecomment-1825342142

cc @MaxLeiter @brunnre8

### DIFF
--- a/server/plugins/irc-events/ctcp.ts
+++ b/server/plugins/irc-events/ctcp.ts
@@ -12,7 +12,7 @@ const ctcpResponses = {
 			.join(" "),
 	PING: ({message}: {message: string}) => message.substring(5),
 	SOURCE: () => pkg.repository.url,
-	VERSION: () => pkg.name + " " + Helper.getVersion() + " -- " + pkg.homepage,
+	VERSION: () => pkg.name + " -- " + pkg.homepage,
 };
 
 export default <IrcEventHandler>function (irc, network) {


brunnre8:
Thanks for tolerating the "we don't really know or agree what we want" dance...
Appreciate your contribution.

PR #4861, opened 2024-04-21
Untangle client and server

brunnre8:
Our project was quite confused as to the boundaries between client and server code.
This false sharing meant that it was quite hard to tell what was actually sent to the client and what was uniquely scoped to either side.

Further, this meant that our compilation and build pipelines were very confused and pulled in files they should not have.

This commit series tries to untangle the two. This also entails fixing quite some typing issues.
It's hard to make this in sane, small,  commits that still build at each step (it's impossible, as fixing one type error / any type immediately lead to further errors in a game of whack a mole).

So you'll get my actual progress in small commits that can each be reviewed, however the earlier ones are in fact sometimes wrong and get cleaned up later once the picture is a bit clearer.
If you have a better idea of how to do it that isn't just squash it all to a single commit, I'd be all ears


server/models/chan.ts
```diff
@@ -206,38 +184,54 @@ class Chan {
...
+			num_users: this.num_users,
+		};
+		// TODO: funny array mutation below might need to be reproduced
+		// static optionalProperties = ["userAway", "special", "data", "closed", "num_users"];
+		// return Object.keys(this).reduce((newChannel, prop) => {
+		// 	if (Chan.optionalProperties.includes(prop)) {
+		// 		if (this[prop] !== undefined || (Array.isArray(this[prop]) && this[prop].length)) {
+		// 			newChannel[prop] = this[prop];
+		// 		}
+		// 	}
```
brunnre8, line 232: Is there any reason that we need to do that dance from the old code there?
I couldn't figure out why it was done that way originally.

MaxLeiter: No idea, its been there a long time. I think

MaxLeiter (approved):

PR #4869, opened 2024-05-04
Ts update

brunnre8:
Bump compiler to the latest, needs updated eslint as well, which lead to a bunch of errors that needed to be addressed.


.eslintrc.cjs
```diff
@@ -93,6 +93,7 @@ const tsRules = defineConfig({
 		// note you must disable the base rule as it can report incorrect errors
 		"no-shadow": "off",
 		"@typescript-eslint/no-shadow": ["error"],
+		"@typescript-eslint/no-redundant-type-constituents": "off",
```
brunnre8, line 96: don't really see the point in making our types worse, even though it's technically correct that just having `undefined` as a type encompasses everything, but documentation wise that's poor.

client/js/commands/index.ts
```diff
@@ -1,19 +1,11 @@
...
+import {input as expand} from "./expand";
+import {input as join} from "./join";
+import {input as search} from "./search";
+
+export const commands = {
+	collapse: collapse,
+	expand: expand,
+	join: join,
+	search: search,
+};
```
brunnre8, line 11: This we probably also need to do with the other things that do the dynamic import...
While it does safe some typing, this is preventing us from moving to modules (we need top level imports or wait for the promise if we do it dynamically), is bad for tree shaking and simply put not needed as all our stuff is available at build time, we don't have plugins or such

client/js/commands/index.ts
```diff
@@ -1,19 +1,11 @@
...
-	acc[command] = commands(path).default;
-
-	return acc;
-}, {});
+import {input as collapse} from "./collapse";
+import {input as expand} from "./expand";
+import {input as join} from "./join";
+import {input as search} from "./search";
+
+export const commands = {
```
MaxLeiter, line 6: Much simpler

MaxLeiter (approved):

PR #4872, opened 2024-05-09
Identd: fix various issues

brunnre8:
There's a bunch of sub optimal behavior from our ident server.
For one, it allows user enumeration which we don't really want and it doesn't clean up connections that don't send any data.

Fix that

### DIFF
--- a/server/identification.ts
+++ b/server/identification.ts
@@ -66,6 +66,14 @@ class Identification {
 
 	serverConnection(socket: Socket) {
 		socket.on("error", (err: string) => log.error(`Identd socket error: ${err}`));
+		socket.setTimeout(5000, () => {
+			log.warn(
+				`identd: no data received, closing connection to ${
+					socket.remoteAddress || "undefined"
+				}`
+			);
+			socket.destroy();
+		});
 		socket.once("data", (data) => {
 			this.respondToIdent(socket, data);
 			socket.end();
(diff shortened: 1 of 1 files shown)


brunnre8:
@Juerd I've added you as co-authored-by, hope that's ok but if not ping me and I'll amend the commit

server/identification.ts
```diff
@@ -66,31 +66,56 @@ class Identification {
...
-			if (connection.socket.remotePort === fport && connection.socket.localPort === lport) {
-				return socket.write(
-					`${lport}, ${fport} : USERID : TheLounge : ${connection.user}\r\n`
-				);
+			// we only want to respond if all the ip,port tuples match, to avoid user enumeration
+			if (
+				connection.socket.remotePort === fport &&
+				connection.socket.localPort === lport &&
+				socket.remoteAddress === connection.socket.remoteAddress &&
+				socket.localAddress === connection.socket.localAddress
```
brunnre8, line 107: This ignores ipv4 in ipv6 handling...
Meaning if for some reason our ident server listens on ipv6 (dual stack) but the IRC connection is in ipv4 we will wrongly send an not found response.

Considering that you have to try rather hard to get to that edge case, I chose not to care.

Juerd:
Reto skribis 2024-05-09 10:06 (-0700):
>@Juerd I've added you as co-authored-by, hope that's ok but if not ping me and I'll amend the commit

That is fine, thanks for confirming this with me.

MaxLeiter (approved):

PR #4876, opened 2024-05-19
Switch from `sqlite3` to `node-sqlite3-wasm`

espadolini:
This PR changes the sqlite message storage implementation from `sqlite3` to `node-sqlite3-wasm`, which provides a prebuilt wasi module that doesn't require any post-install download or compilation step, unlike `sqlite3`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "lodash": "4.17.21",
     "mime-types": "2.1.34",
     "node-forge": "1.3.0",
+    "node-sqlite3-wasm": "0.8.14",
     "package-json": "7.0.0",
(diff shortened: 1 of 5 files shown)


brunnre8:
While I appreciate the effort and see the benefit of having a single binary blob that's arch agnostic, I don't think we can merge this.

The dependency you pick here is a one man show with no corporate backing... a hobby project it seems.
We have too many questionably dependencies already and I don't think we can pull that in.

The sqlite project has an official endorsed wasm port, however so far only browser based.
I think we should let the ecosystem stabilize a bit here.

PR #4882, opened 2024-06-19
Update contextMenu.ts

TehPeGaSuS:
Most of IRC clients use `Give op` and `Take op`.
IMHO, TheLounge should follow the same schema.


brunnre8:
I don't think this has any benefit.
Really, both are English words. Both have their own merrit.

Let's not bikeshed this to death and simply keep what we already have.

TehPeGaSuS:
I was just trying to put evrything logical.

We have `Give op` but `Revoke op´.

But I do understand the reason to not change.

Closing

PR #4887, opened 2024-07-16
Add label to network list search input

### LINKED ISSUE #4863: Textarea for entering messages lacks label
```
<textarea id="input" dir="auto" class="mousetrap" enterkeyhint="send" placeholder="" aria-label=""></textarea>
```

Ideally a label is associated to the control using the HTML `label` element and the `for` attribute so that all users can identify the purpsoe of the control and assistive technologies can identify it programmatically.

### LINKED ISSUE #4864: Input to search channels lacks label
```
<input placeholder="Jump to..." type="search" class="search input mousetrap" aria-label="Search among the channel list" tabindex="-1">
```

Ideally a label is associated to the control using the HTML `label` element and the `for` attribute so that all users can identify the purpsoe of the control and assistive technologies can identify it programmatically.

mven:
resolves #4863
resolves #4864

### DIFF
--- a/client/components/ChannelWrapper.vue
+++ b/client/components/ChannelWrapper.vue
@@ -22,7 +22,7 @@
 		:data-name="channel.name"
 		:data-type="channel.type"
 		:aria-controls="'#chan-' + channel.id"
-		:aria-selected="active"
+		:aria-selected="true"
 		:style="channel.closed ? {transition: 'none', opacity: 0.4} : undefined"
 		role="tab"
 		@click="click"
--- a/client/components/ChatInput.vue
+++ b/client/components/ChatInput.vue
@@ -2,19 +2,20 @@
 	<form id="form" method="post" action="" @submit.prevent="onSubmit">
 		<span id="upload-progressbar" />
 		<span id="nick">{{ network.nick }}</span>
-		<textarea
-			id="input"
-			ref="input"
-			dir="auto"
-			class="mousetrap"
-			enterkeyhint="send"
-			:value="channel.pendingMessage"
-			:placeholder="getInputPlaceholder(channel)"
-			:aria-label="getInputPlaceholder(channel)"
-			@input="setPendingMessage"
-			@keypress.enter.exact.prevent="onSubmit"
-			@blur="onBlur"
-		/>
+		<label :aria-label="getInputPlaceholder(channel)">
+			<textarea
+				id="input"
+				ref="input"
+				dir="auto"
+				class="mousetrap"
+				enterkeyhint="send"
+				:value="channel.pendingMessage"
+				:placeholder="getInputPlaceholder(channel)"
+				@input="setPendingMessage"
+				@keypress.enter.exact.prevent="onSubmit"
+				@blur="onBlur"
+			/>
+		</label>
 		<span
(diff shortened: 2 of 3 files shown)


client/components/ChannelWrapper.vue
```diff
@@ -22,7 +22,7 @@
 		:data-name="channel.name"
 		:data-type="channel.type"
 		:aria-controls="'#chan-' + channel.id"
-		:aria-selected="active"
+		:aria-selected="true"
```
mven, line 25: Drive-by fix here. `active` is not a valid value. It's either `true`, `false`, or undefined (blank)

client/components/ChatInput.vue
```diff
@@ -2,19 +2,20 @@
...
-			class="mousetrap"
-			enterkeyhint="send"
-			:value="channel.pendingMessage"
-			:placeholder="getInputPlaceholder(channel)"
-			:aria-label="getInputPlaceholder(channel)"
-			@input="setPendingMessage"
-			@keypress.enter.exact.prevent="onSubmit"
-			@blur="onBlur"
-		/>
+		<label :aria-label="getInputPlaceholder(channel)">
```
mven, line 5: added parent `label` and moved `aria-label` to it, same as below.

client/components/ChannelWrapper.vue
brunnre8: NACK, active is a prop, passed via the element by vue.

mven: Ah right. I'm new to Vue and I missed that this was a dynamic attribute. I'll revert this change shortly.

brunnre8:
Can you explain to me how this fixes https://github.com/thelounge/thelounge/issues/4863?
You populate the label with the same content as is already there, so that doesn't seem to do anything at a glance for screen readers?

As for https://github.com/thelounge/thelounge/issues/4864 this has the aria-label already on there as well, normal users don't need a label as the placeholder text serves the same purpose, so what's the rational for doing it?

brunnre8:
as is this also borks the CSS as far as I can tell, for both input boxes but most notably for the chat input:

image

mven:
> Can you explain to me how this fixes #4863? You populate the label with the same content as is already there, so that doesn't seem to do anything at a glance for screen readers?

One distinction is that the `label` associated with the `input` provides more of a semantic structure than what's currently there. Though, I've run a few tests on my end and realized that this code change isn't best practice taking the intentional lack of a visual label into consideration.

It might not cover all cases and there's a fallback such as wrapping the `label` around the `input` much like this code change, but 
What's currently in the source of these corresponding changes currently meets WCAG 2.1 AA guidelines, so I think I'll close this PR.

mven:
> As for #4864 this has the aria-label already on there as well, normal users don't need a label as the placeholder text serves the same purpose, so what's the rational for doing it?

`placeholder` text isn't sufficient since they're mostly inaccessible and folks with vision impairments can have problems with it. An improvement could be made to add `aria-placeholder`, but this should only be used as a supplement to an actual label (or `aria-label`).

mven:
Closing out for now! Thanks for the review @brunnre8

PR #4891, opened 2024-07-25
pull

MobiusCraftFlip:


brunnre8:
patching out the dev banner is not a good idea... build your fork instead in production mode.

There's a bunch of horrendous things that you definitely don't want in prod enabled in dev mode

PR #4896, opened 2024-08-07
Support safe area insets to improve UX on mobile browsers.

rval:
Adds support for [safe area insets](https://developer.mozilla.org/en-US/docs/Web/CSS/env#values) to improve the user experience on mobile browsers.

This prevents UI elements (e.g., message text input, sidebar controls) from being obscured by the iPhone's home indicator and notch, as shown in the attached screenshots.

This tries to merge in some of the work from @mhajder's [pull request](https://github.com/thelounge/thelounge/pull/4516).

# Details

* I've checked to ensure the mentions panel is still dismissible and performs the same way it does in the main branch.
* This MR tries to follow Apple's recommendations (and the web standard) as closely as possible. That is, it uses the `safe-area-insets` as provided by the OS, without tweaks for aesthetics. The padding is consistent with the behavior of Apple's native apps on the iPhone.

# Before

Notice that the iPhone's home indicator obscures the message input box and sidebar controls, making them difficult to tap.

Channel in Portrait (Before)
Sidebar in Portrait (Before)
Channel in Landscape with Keyboard (Before)
Channel in Landscape (Before)
Sidebar in Landscape (Before)

# After

The iPhone's home indicator doesn't obscure the message input box or sidebar controls, and it extends into the background (at least in portrait), which is more consistent with Apple's native iPhone apps. 

Channel Portrait (Without Chrome)
Sidebar Portrait (With Chrome)
Channel in Landscape
Channel in Landscape with Keyboard

# Comparison With Native Apps

Note that the padding is consistent with native apps on the iPhone (although it looks more pronounced because the Lounge's text field doesn't have borders):

Messages
Channel Portrait (Without Chrome)

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -673,7 +673,7 @@ p {
 	display: none;
 	flex-direction: column;
 	width: 220px;
-	max-height: 100vh;
+	max-height: 100%;
 	will-change: transform;
 	color: #b7c5d1; /* same as .channel-list-item color */
 }


client/index.html.tpl
```diff
@@ -3,7 +3,7 @@
 	<head>
 
 	<meta charset="utf-8">
-	<meta name="viewport" content="width=device-width, user-scalable=no">
+	<meta name="viewport" content="width=device-width, user-scalable=no, viewport-fit=cover">
```
rval, line 6: Note that `viewport-fit=cover` is required to making the safe area insets work.

brunnre8:
Thanks for your effort.

There's a similar PR in https://github.com/thelounge/thelounge/pull/4516 that I can't really test as I don't own an iOS device and on Android those things work a bit differently, least on my device.

Can you compare which of the two works better for you? There was some feedback on the earlier PR, not that I like the games that are played by reducing the inset areas... that probably has to go away.
Because if devices really say what they need this would again break things.

Let people complain to apple if the margins are insane.

But there seems to have been an issue with the notification popup for the simple solution, can you evaluate which of the approaches work better for you? The older PR or yours?

client/css/style.css
```diff
@@ -2285,6 +2290,10 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
 	background: transparent;
 }
 
+#mentions-popup-container {
+	margin-top: env(safe-area-inset-top);
+}
```
rval, line 2295: Ensures the mentions popup container can still be dismissed.

rval:
@brunnre8 I wasn't aware of that pull request, thanks for letting me know. 

I think either one would be an improvement, but I tried to integrate the changes from @mhajder's PR into this one with the following changes:

* I avoided aesthetic tweaks. This PR is purely focused on fixing the usability issues. 
* I wasn't able to get safe area insets working properly in landscape without breaking portrait orientation. Since landscape was relatively usable, I focused on just getting a fix for portrait mode in place.
* I did make sure the mentions popup continues to work as it does in the main branch (at least as far as I can tell).

Let me know what you think!

brunnre8:
as said, can't test it myself so I do rely on feedback from iOS folks like yourself.

All I'll do is test that you don't break it on my devices 😉 

Thanks a bunch, will have a look tomorrow

rval:
@brunnre8 I've updated my pull request to address some of the remaining issues and updated the description with new images. I think this PR does improve on the originals now.

Let me know if there's anything else I can do to help get this over the finish line. I took some video of the interactions but Github's rejecting the file sizes.

client/css/style.css
```diff
@@ -107,6 +108,13 @@ body {
 	overflow: hidden; /* iOS Safari requires overflow rather than overflow-y */
 }
 
+@media screen and (orientation: portrait) {
+	/* Allow the main chat interface to merge into the background. */
+	body {
+		background: white;
```
brunnre8, line 114: please remove that again.
I don't see why you need to set the color there for a padding issue.

People use dark themes you know and having some flash white is seldom a good idea :)

client/css/style.css
```diff
@@ -2211,6 +2219,18 @@ part/quit messages where we don't load previews (adds a blank line otherwise) */
...
 }
 
+@media screen and (orientation: landscape) {
+	/* Pad the form with safe area insets. */
+	#form {
+		padding-bottom: env(safe-area-inset-bottom)
+	}
+
+	/* When the form becomes focused, remove the safe area insets padding. A bit of a hack, but seems to work. */
+	#form:focus-within {
```
brunnre8, line 2229: why exactly is this needed now? Your first version didn't have it?

client/themes/default.css
```diff
@@ -15,6 +15,12 @@
 	box-shadow: 0 0 25px rgb(0 0 0 / 50%);
 }
 
+@media screen and (orientation: portrait) {
+	.window {
+		box-shadow: 0 0 0;
```
brunnre8, line 20: why are you messing with the shadows here?

rval: It was an attempt to make it look a bit more at home on iOS. Here it is with and without these rules, respectively:

Channel in Portrait
Channel Portrait (Without Chrome)

But probably better left to CSS overrides or a separate theme maybe, so I left it out. I'll update the screenshots in the merge request to reflect the new appearance.

client/css/style.css
rval: Without this, the keyboard keeps the safe area padding when it's focused, which looks weird and reduces the limited amount of vertical space. 

Here's what it looks like with this CSS (notice the extra padding in the input box):

Padded Form

And here it is with the CSS (notice the padding is not there, and there's more room for the channel content):

Unpadded Form

I'm not sure why it wasn't present in my earlier version. I tried a bunch of different ways to address this but this was the best one I landed on.

client/css/style.css
rval: This was to get the effect mentioned in my other comment, but definitely not necessary. Removed as requested.

client/css/style.css
rval: @brunnre8 if this is a blocker (or we just want to kick this particular decision down the road) I could make one of the following changes:

1. I can remove the `#form:focus-within` hack, and we can just let the padding exist as shown in the screenshot above.
2. I can descope this pull request to support the safe area insets _only_ for portrait orientation, and leave landscape orientation as is (it can always be tackled later since portrait mode is the bigger issue, I think).

Let me know if you have a preference.

client/themes/default.css
JipFr: I definitely think that the background should be the same color as the text input box, perhaps there is a different way to enforce that?

ivanjx:
what is still blocking the merge? even though it might not be perfect this is still way better than the current ux

FlyGoat:
+1, this improved iOS PWA experience a lot.

macmanx2:
Especially after iOS 26, this would _dramatically_ improve The Lounge as a PWA app. Right now, the iPhone's home indicator takes up the bottom 1/4 of the text box, meaning that you'll need to be very precise when tapping in to type, or you might launch Siri.

macmanx2:
While all of the above is the right way to fix this, until/if it's all merged (and until/if your admin updates), I "fixed" it by adding this to my custom stylesheet (adjust as desired).

`#form {
    padding-bottom: 16px;
}`

benyafai:
> While all of the above is the right way to fix this, until/if it's all merged (and until/if your admin updates), I "fixed" it by adding this to my custom stylesheet (adjust as desired).
> 
> 
> 
> `#form {
> 
>     padding-bottom: 16px;
> 
> }`



I've had this in my instance for years; 

```
#form {
  margin-bottom: 30px;
}
textarea#user-specified-css-input {
  height: 600px;
}
```

balgerion:
Any chance to merge this PR ? we are now in iOS 26 and rly need this to have working PWA

PR #4901, opened 2024-08-19
whois: support actual_username field

### LINKED ISSUE #4900: Show the real username in the WHOIS output
### Feature Description

Some servers (e.g. InspIRCd v4) support the concept of a real username, i.e. the username that was originally sent as distinct from their virtual username set by services. This is most commonly used to prevent ban evasion using services from users on shared hosts (e.g. IRCCloud).

This value is sent in `RPL_WHOISACTUALLY` and is already parsed by irc-framework into the `actual_username` field. Unfortunately unlike `actual_hostname` The Lounge does not currently show this in the WHOIS output. Would it be possible for you to add this?

brunnre8:
Some servers (e.g. InspIRCd v4) support the concept of a real username, i.e. the username that was originally sent as distinct from their virtual username set by services.

Closes: https://github.com/thelounge/thelounge/issues/4900

### DIFF
--- a/client/components/MessageTypes/whois.vue
+++ b/client/components/MessageTypes/whois.vue
@@ -34,6 +34,11 @@
 				</dd>
 			</template>
 
+			<template v-if="message.whois.actual_username">
+				<dt>Actual username:</dt>
+				<dd>{{ message.whois.actual_username }}</dd>
+			</template>
+
 			<template v-if="message.whois.real_name">
 				<dt>Real name:</dt>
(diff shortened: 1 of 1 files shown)


SadieCat:
Unfortunately it seems like irc-framework hasn't had a release in over a year so the `actual_username` field hasn't made it into a release yet. However, after updating irc-framework from Git this works as expected.

Screenshot from 2024-08-19 11-25-58

SadieCat:
There's a new irc-framework release so this should work now.

https://github.com/kiwiirc/irc-framework/releases/tag/v4.14.0

PR #4907, opened 2024-09-03
Fix scoped package install & theme loading

arminius-smh:
Hello, there has been this issue for quite some time now that breaks `thelounge install` for scoped npm packages: https://github.com/thelounge/thelounge/issues/4295

This pr fixes this by checking whether the package is scoped or not, and then performing the same split operations so now every one of these options can be installed through `thelounge install`:
```
thelounge install @scope/package (defaults to latest)
thelounge install @scope/package@2.0.0
thelounge install package (defaults to latest)
thelounge install package@2.0.0
```

The other problem in the linked issue is that themes from scoped packages don't get loaded when they are selected, because of the `/` in their name, I fixed this by encoding the URI at the two relevant sections. Now themes from scoped packages load when they are selected in the appearance menu, and also stay loaded when reloading the website.

### DIFF
--- a/server/command-line/install.ts
+++ b/server/command-line/install.ts
@@ -47,9 +47,15 @@ program
 				.readFile(path.join(packageName.substring("file:".length), "package.json"), "utf-8")
 				.then((data) => JSON.parse(data) as typeof packageJson);
 		} else {
-			const split = packageName.split("@");
-			packageName = split[0];
-			const packageVersion = split[1] || "latest";
+			// properly split scoped and non-scoped npm packages
+			// into their name and version
+			let packageVersion = "latest";
+			const atIndex = packageName.indexOf("@", 1);
+
+			if (atIndex !== -1) {
+				packageVersion = packageName.slice(atIndex + 1);
+				packageName = packageName.slice(0, atIndex);
+			}
 
 			readFile = packageJson.default(packageName, {
 				fullMetadata: true,
--- a/server/plugins/packages/themes.ts
+++ b/server/plugins/packages/themes.ts
@@ -85,7 +85,7 @@ function makePackageThemeObject(
 	return {
 		displayName: module.name || moduleName,
 		filename: path.join(modulePath, module.css),
-		name: moduleName,
+		name: encodeURIComponent(moduleName),
 		themeColor: themeColor,
 	};
 }
--- a/server/server.ts
+++ b/server/server.ts
@@ -107,7 +107,7 @@ export default async function (
 	// handler. Remember this if you make changes to this function, serving of
 	// local themes will not get those changes.
 	app.get("/themes/:theme.css", (req, res) => {
-		const themeName = req.params.theme;
+		const themeName = encodeURIComponent(req.params.theme);
 		const theme = themes.getByName(themeName);
 
 		if (theme === undefined || theme.filename === undefined) {


server/plugins/packages/themes.ts
```diff
@@ -82,10 +82,11 @@ function makePackageThemeObject(
 
 	const themeColor = /^#[0-9A-F]{6}$/i.test(module.themeColor) ? module.themeColor : null;
 	const modulePath = Config.getPackageModulePath(moduleName);
+
```
brunnre8, line 85: try not to make unrelated changes that have nothing to do with your PR ;)

server/command-line/install.ts
```diff
@@ -47,9 +47,20 @@ program
...
+
+				if (atIndex > 0) {
+					packageVersion = packageName.slice(atIndex + 1) || "latest";
+					packageName = packageName.slice(0, atIndex);
+				}
+			} else {
+				const split = packageName.split("@");
+				packageName = split[0];
+				packageVersion = split[1] || "latest";
+			}
```
brunnre8, line 63: This is more complicated than need be I think...
You don't have to special case anything here, just start looking at the 2nd char

```ts
let packageVersion = "latest"
const atIndex = packageName.lastIndexOf("@", 1); // make sure scoped @org/something works
if(atIndex != -1) {
    packageName = packageName.slice(0,atIndex);
    packageVersion = packageName.slice(atIndex+1);
}
```
Or such would work and remove most of the conditionals

brunnre8 (changes requested):
Lovely, thanks a lot for the effort.
You can simplify the version calc I think, other than that lgtm

server/command-line/install.ts
brunnre8: (copy pasta leftover from your code fixed... mea culpa)

arminius-smh: Thanks! I guess I was overthinking this a little, I just had to change packageName and packageVersion around, so that the sliced packageName doesn't get used for the Version

server/plugins/packages/themes.ts
arminius-smh: Ahh my git was configured to hide only whitespace changes, I was confused as I was pretty sure I checked my diff and didn't saw something like that, now it won't EVER happen again!

server/command-line/install.ts
```diff
@@ -47,9 +47,13 @@ program
 				.readFile(path.join(packageName.substring("file:".length), "package.json"), "utf-8")
 				.then((data) => JSON.parse(data) as typeof packageJson);
 		} else {
-			const split = packageName.split("@");
-			packageName = split[0];
-			const packageVersion = split[1] || "latest";
+			let packageVersion = "latest";
+			const atIndex = packageName.indexOf("@", 1);
+
+			if (atIndex !== -1) {
```
brunnre8, line 53: Can you add a human friendly comment saying why you do that dance? It seems obvious now but give yourself a month or two ;)

arminius-smh: Something like how it is now?

brunnre8: should work yes. I'd have given an example personally but that's bike shedding it to death.
Thanks.

brunnre8 (approved):

PR #4928, opened 2024-10-10
client: suppress "***" in joins/parts/etc. for screen readers

kfranqueiro:
I noticed while testing with screen readers that events such as joins and parts have a non-printed `***` that is given the class `only-copy` but currently not `aria-hidden="true"`, resulting in screen readers announcing "star star star" before every one of these events. I suspect this was an oversight, as most `only-copy` elements do have `aria-hidden="true"`.


brunnre8:
thank you

PR #4929, opened 2024-10-14
Fix showing multiple certificate fingerprints on InspIRCd v4.

sadiepowell:
InspIRCd v4 will send multiple certificate fingerprints in WHOIS if a server is migrating from an old fingerprint (e.g. md5) to a new one (e.g. sha256). Currently The Lounge will only show the last (latest release) or first (latest git with the updated irc-fw) of these fingerprints.

Example of this from the InspIRCd testnet at testnet.inspircd.org with a sha3-256 and sha256 hash:

Screenshot from 2024-10-14 18-07-59

### DIFF
--- a/client/components/MessageTypes/whois.vue
+++ b/client/components/MessageTypes/whois.vue
@@ -91,9 +91,11 @@
 				<dd>Yes</dd>
 			</template>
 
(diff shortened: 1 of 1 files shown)


brunnre8:
thank you.
Sorry for the late merge

PR #4932, opened 2024-10-16
client: Indicate active context menu item for screen readers

kfranqueiro:
The app's context menus include keyboard navigation support via arrow keys, but currently the active item is only indicated via a CSS class, which means screen readers have nothing to announce when the active item changes.

This PR hooks up `aria-activedescendant` to child IDs, which screen readers can announce.


brunnre8:
Thanks a lot for your contribution.
Sorry for the late merge, been busy lately.

Issue #4311, opened 2021-09-15
Lack of semantics to communicate structure


daniel-montalvo:
Elements such as headings and page regions are missing. They communicate the structure of the page or application to assistive technologies, and help users navigate efficiently to the different sections. Some examples of such regions are:

- List of networks
- List of channels
- Dates in a message list
- New messages

We recommend appropriate use of  headings and page regions to communicate the different sections of the page.

brunnre8:
Would you mind elaborating?
What is missing concretely?

At least for the channel list the structure is annotated

![x](https://labrat.space/irc/be22cc611871d1f3/2021-09-15_15-01-10.png)

daniel-montalvo:
Thanks @brunnre8 for getting back to this so quickly

> Would you mind elaborating?
> What is missing concretely?

I am missing headings that identify the different sections of the page. Screen reader users navigate using specific screen readers functionality that relies on headings.

I would have at least the following:

- h1 with the title of the current network
- h2 before the list of available networks (if the user has several of them available)
- h2 before the list of channels
- h2 before each change in the date of the messages, so that users can jump quickly through the different dates in a long list of messages

For references on how to communicate the structure of a page, see 
https://www.w3.org/WAI/tutorials/page-structure/

> 
> At least for the channel list the structure is annotated
> 
> ![x](https://camo.githubusercontent.com/5b7285c54f899b60fa55ce988a5202b9d3125074475cb6e1d7b697452ff4389b/68747470733a2f2f6c61627261742e73706163652f6972632f626532326363363131383731643166332f323032312d30392d31355f31352d30312d31302e706e67)

I am sorry, I do not understand what you are pointing me too. I am a completely blind person and this seems to me like a screen capture.

From what I am seeing, I have also suggestions for the channels list that I would probably put in a different issue.

brunnre8:
yeah in hindsight that was a stupid thing to do... sorry for that.

the channel list's html contains aria labels and roles which should have been picked up for you, but apparently didn't.

The corresponding elements are divs though

daniel-montalvo:
> yeah in hindsight that was a stupid thing to do... sorry for that.

No worries :-) Just letting you know.
> 
> the channel list's html contains aria labels and roles which should have been picked up for you, but apparently didn't.
> 
> The corresponding elements are divs though

[...]

PR #4936, opened 2024-11-12
client: Add headings for easier navigation by AT

kfranqueiro:
Refs #4311 (this should be a step in the right direction, though it does not fully implement every idea presented in that thread; Daniel has tested my changes and noted they're an improvement)

This adds headings to the following:

- Channel name above main chat area (added aria roles to title span to set heading level 2)
- User list (added h3 with `sr-only` class)
- Date markers (converted to h4; simplified markup to move label into innerHTML, out of aria-label)
- Unread marker (converted to h4; simplified markup to move label into innerHTML, out of CSS)
- Network and channel list (added h2 with `sr-only` class; moved content from `aria-label` in parent element)
- The Lounge title (added h1 with `sr-only`; relocated content from logo image alt text)

Anywhere that I changed an existing element's tag, I tested and changed CSS accordingly to preserve styles. I resorted to ARIA attributes for the channel name because it seemed like preserving the styles would've been less straightforward there.

If any of these seem like they'd be problematic for theme compatibility, maybe I can come up with alternatives.

### DIFF
--- a/client/components/Sidebar.vue
+++ b/client/components/Sidebar.vue
@@ -2,17 +2,16 @@
 	<aside id="sidebar" ref="sidebar">
 		<div class="scrollable-area">
 			<div class="logo-container">
+				<h1 class="sr-only">The Lounge</h1>
 				<img
 					:src="`img/logo-${isPublic() ? 'horizontal-' : ''}transparent-bg.svg`"
 					class="logo"
-					alt="The Lounge"
-					role="presentation"
+					alt=""
 				/>
 				<img
 					:src="`img/logo-${isPublic() ? 'horizontal-' : ''}transparent-bg-inverted.svg`"
 					class="logo-inverted"
-					alt="The Lounge"
-					role="presentation"
+					alt=""
 				/>
 				<span
 					v-if="isDevelopment"
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -18,9 +18,10 @@
 			>
 				<div class="header">
 					<SidebarToggle />
-					<span class="title" :aria-label="'Currently open ' + channel.type">{{
-						channel.name
-					}}</span>
+					<span class="title" role="heading" aria-level="2">
+						<span class="sr-only">Currently open {{ channel.type }}:&nbsp;</span>
+						{{ channel.name }}
+					</span>
 					<div v-if="channel.editTopic === true" class="topic-container">
 						<input
 							ref="topicInput"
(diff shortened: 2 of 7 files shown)


client/components/Sidebar.vue
```diff
@@ -2,17 +2,16 @@
 	<aside id="sidebar" ref="sidebar">
 		<div class="scrollable-area">
 			<div class="logo-container">
+				<h1 class="sr-only">The Lounge</h1>
```
vivienlacourba, line 5: My impression is that before the `alt` attributing was detailing the image logo, while moving it to an `<h1>` heading probably changes its meaning to be the service's name. So it should maybe be moved outside of the `<div class="logo-container">` and maybe even outside of  `<aside id="sidebar" ref="sidebar">`

Another problem is that on the connect page there is already an `<h1>` heading.
Example of the one used on [the demo instance](https://demo.thelounge.chat/#/connect):
`<h1 class="title">Connect to Libera.Chat</h1>`

Also it would be useful if that service name could be configurable in particular for people that might be connected to multiple The Lounge instances to more easily differentiate them.
I suppose that last comment might actually best addressed in a separate issue and PR as it also affect the html page's title which is currently set to `<title>The Lounge</title>`.

vivienlacourba (changes requested):

client/components/Sidebar.vue
vivienlacourba: > Another problem is that on the connect page there is already an `<h1>` heading.

The edit network page also uses an `<h1>` for its heading.

kfranqueiro: Oof, thanks for catching this; I hadn't realized that some of the forms have h1's on them. I'm kind of inclined to just back out my logo/h1 change for now.

However, this may still be problematic on a further level, because the h1 present on various forms also appears later in the document than the h2 I added for Network and Channel List, which seems not ideal.

It seems as if the forms might be the only places where headings already existed prior to this PR - almost as if headings were _only_ considered for the forms and not for the rest of the UI that is in use the majority of the time. This may make adding headings to the rest of the UI less straightforward unless they were to be re-leveled within forms?

iadawn: I don't see a huge problem with multiple `<h1>`s. I know there are some who would say this is a "bad thing" but I am not one of those. Ultimately this is about exposing the structure of the page to make it easier for assistive technology users to navigate around.

vivienlacourba: I had thought that having a double `<h1>` could bring accessibility issues, but if it doesn't then I'm fine with this part of the changes.

PR #4951, opened 2025-01-10
server: decode link-preview text using document/transport-specified encoding

sideshowbarker:
This change detects the specified encoding of any document posted as a link, and uses that encoding to display the preview text for the link.

Otherwise, without this change, the link-preview text for documents with legacy encodings (such as Shift_JIS) doesn’t get displayed as expected, but instead gets garbled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "ua-parser-js": "1.0.39",
     "uuid": "8.3.2",
     "web-push": "3.4.5",
+    "whatwg-mimetype": "4.0.0",
     "yarn": "1.22.22"
   },
   "optionalDependencies": {
@@ -96,7 +97,6 @@
     "@textcomplete/textarea": "0.1.13",
     "@types/bcryptjs": "2.4.6",
     "@types/chai": "4.3.5",
-    "@types/cheerio": "0.22.35",
     "@types/content-disposition": "0.5.8",
     "@types/express": "4.17.21",
     "@types/is-utf8": "0.2.3",
@@ -116,6 +116,7 @@
     "@types/web-push": "3.3.2",
     "@types/webpack-env": "1.16.4",
     "@types/webpack-hot-middleware": "2.25.6",
+    "@types/whatwg-mimetype": "3.0.2",
     "@types/ws": "8.5.12",
     "@typescript-eslint/eslint-plugin": "7.8.0",
     "@typescript-eslint/parser": "7.8.0",
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "tlds": "1.228.0",
     "ua-parser-js": "1.0.39",
     "web-push-neo": "0.1.2",
+    "whatwg-mimetype": "4.0.0",
     "yarn": "1.22.22"
   },
   "devDependencies": {
@@ -84,7 +85,6 @@
     "@textcomplete/core": "0.1.10",
     "@textcomplete/textarea": "0.1.10",
     "@types/bcryptjs": "2.4.6",
-    "@types/cheerio": "0.22.35",
     "@types/content-disposition": "0.5.8",
     "@types/express": "4.17.21",
     "@types/ldapjs": "2.2.5",
@@ -97,6 +97,7 @@
     "@types/semver": "7.3.9",
     "@types/sortablejs": "1.15.8",
     "@types/ua-parser-js": "0.7.39",
+    "@types/whatwg-mimetype": "3.0.2",
     "@types/ws": "8.5.12",
     "@typescript-eslint/eslint-plugin": "7.8.0",
     "@typescript-eslint/parser": "7.8.0",
--- a/server/plugins/irc-events/link.ts
+++ b/server/plugins/irc-events/link.ts
@@ -1,7 +1,9 @@
 import * as cheerio from "cheerio";
+import {CheerioAPI} from "cheerio";
 import got from "got";
 import {URL} from "url";
 import mime from "mime-types";
+import MIMEType from "whatwg-mimetype";
 
 import log from "../../log";
 import Config from "../../config";
@@ -15,6 +17,7 @@ import Msg from "../../models/msg";
 type FetchRequest = {
 	data: Buffer;
 	type: string;
+	charset: string | undefined;
 	size: number;
 };
 const currentFetchPromises = new Map<string, Promise<FetchRequest>>();
@@ -78,8 +81,9 @@ function parseHtml(preview, res, client: Client) {
 	// TODO:
 	// eslint-disable-next-line @typescript-eslint/no-misused-promises
 	return new Promise((resolve: (preview: FetchRequest | null) => void) => {
-		const $ = cheerio.load(res.data);
-
+		const $ = cheerio.loadBuffer(res.data, {
+			encoding: {transportLayerEncodingLabel: res.charset},
+		});
 		return parseHtmlMedia($, preview, client)
 			.then((newRes) => resolve(newRes))
 			.catch(() => {
@@ -140,7 +144,7 @@ function parseHtml(preview, res, client: Client) {
 }
 
 // TODO: type $
-function parseHtmlMedia($: any, preview, client: Client): Promise<FetchRequest> {
+function parseHtmlMedia($: CheerioAPI, preview, client: Client): Promise<FetchRequest> {
 	return new Promise((resolve, reject) => {
 		if (Config.values.disableMediaPreview) {
 			reject();
@@ -167,7 +171,7 @@ function parseHtmlMedia($: any, preview, client: Client): Promise<FetchRequest> 
 				return;
 			}
 
-			$(`meta[property="og:${type}:type"]`).each(function (this: cheerio.Element, i: number) {
+			$(`meta[property="og:${type}:type"]`).each(function (i) {
 				const mimeType = $(this).attr("content");
 
 				if (!mimeType) {
@@ -464,15 +468,22 @@ function fetch(uri: string, headers: Record<string, string>) {
 				.on("end", () => gotStream.destroy())
 				.on("close", () => {
 					let type = "";
+					let charset;
 
 					// If we downloaded more data then specified in Content-Length, use real data size
 					const size = contentLength > buffer.length ? contentLength : buffer.length;
 
 					if (contentType) {
-						type = contentType.split(/ *; */).shift() || "";
+						try {
+							const mimeType = new MIMEType(contentType);
+							type = mimeType.essence;
+							charset = getTransportLayerEncodingLabel(mimeType);
+						} catch {
+							// the Content-Type value isn't valid; ignore it
+						}
 					}
 
-					resolve({data: buffer, type, size});
+					resolve({data: buffer, type, charset, size});
 				});
 		} catch (e: any) {
 			return reject(e);
@@ -486,6 +497,14 @@ function fetch(uri: string, headers: Record<string, string>) {
 	currentFetchPromises.set(cacheKey, promise);
 
 	return promise;
+}
+
+function getTransportLayerEncodingLabel(mimeType: MIMEType): string | undefined {
+	try {
+		return mimeType.parameters.get("charset");
+	} catch (error) {
+		return undefined;
+	}
 }
 
 function normalizeURL(link: string, baseLink?: string, disallowHttp = false) {
--- a/server/plugins/irc-events/link.ts
+++ b/server/plugins/irc-events/link.ts
@@ -3,6 +3,7 @@ import type {Element} from "domhandler";
 import got from "got";
 import {URL} from "url";
 import mime from "mime-types";
+import MIMEType from "whatwg-mimetype";
 
 import log from "../../log";
 import Config from "../../config";
@@ -16,6 +17,7 @@ import Msg from "../../models/msg";
 type FetchRequest = {
 	data: Buffer;
 	type: string;
+	charset: string | undefined;
 	size: number;
 };
 const currentFetchPromises = new Map<string, Promise<FetchRequest>>();
@@ -79,8 +81,9 @@ function parseHtml(preview, res, client: Client) {
 	// TODO:
 	// eslint-disable-next-line @typescript-eslint/no-misused-promises
 	return new Promise((resolve: (preview: FetchRequest | null) => void) => {
-		const $ = cheerio.load(res.data);
-
+		const $ = cheerio.loadBuffer(res.data, {
+			encoding: {transportLayerEncodingLabel: res.charset},
+		});
 		return parseHtmlMedia($, preview, client)
 			.then((newRes) => resolve(newRes))
 			.catch(() => {
@@ -140,8 +143,7 @@ function parseHtml(preview, res, client: Client) {
 	});
 }
 
-// TODO: type $
-function parseHtmlMedia($: any, preview, client: Client): Promise<FetchRequest> {
+function parseHtmlMedia($: cheerio.CheerioAPI, preview, client: Client): Promise<FetchRequest> {
 	return new Promise((resolve, reject) => {
 		if (Config.values.disableMediaPreview) {
 			reject();
@@ -467,15 +469,22 @@ function fetch(uri: string, headers: Record<string, string>) {
 				.on("end", () => gotStream.destroy())
 				.on("close", () => {
 					let type = "";
+					let charset;
 
 					// If we downloaded more data then specified in Content-Length, use real data size
 					const size = contentLength > buffer.length ? contentLength : buffer.length;
 
 					if (contentType) {
-						type = contentType.split(/ *; */).shift() || "";
+						try {
+							const mimeType = new MIMEType(contentType);
+							type = mimeType.essence;
+							charset = getTransportLayerEncodingLabel(mimeType);
+						} catch {
+							// the Content-Type value isn't valid; ignore it
+						}
 					}
 
-					resolve({data: buffer, type, size});
+					resolve({data: buffer, type, charset, size});
 				});
 		} catch (e: any) {
 			return reject(e);
@@ -489,6 +498,14 @@ function fetch(uri: string, headers: Record<string, string>) {
 	currentFetchPromises.set(cacheKey, promise);
 
 	return promise;
+}
+
+function getTransportLayerEncodingLabel(mimeType: MIMEType): string | undefined {
+	try {
+		return mimeType.parameters.get("charset");
+	} catch (error) {
+		return undefined;
+	}
 }
 
 function normalizeURL(link: string, baseLink?: string, disallowHttp = false) {
--- a/yarn.lock
+++ b/yarn.lock
@@ -1501,13 +1501,6 @@
   resolved "https://registry.yarnpkg.com/@types/chai/-/chai-4.3.5.tgz#ae69bcbb1bebb68c4ac0b11e9d8ed04526b3562b"
   integrity sha512-mEo1sAde+UCE6b2hxn332f1g1E8WfYRu6p5SvTKr2ZKC1f7gFJXk4h5PyGP9Dt6gCaG8y8XhwnXWC6Iy2cmBng==
 
-"@types/cheerio@0.22.35":
-  version "0.22.35"
-  resolved "https://registry.yarnpkg.com/@types/cheerio/-/cheerio-0.22.35.tgz#0d16dc1f24d426231c181b9c31847f673867595f"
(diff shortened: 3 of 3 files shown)


server/plugins/irc-events/link.ts
```diff
@@ -78,7 +80,8 @@ function parseHtml(preview, res, client: Client) {
 	// TODO:
 	// eslint-disable-next-line @typescript-eslint/no-misused-promises
 	return new Promise((resolve: (preview: FetchRequest | null) => void) => {
-		const $ = cheerio.load(res.data);
+		const encoding = chardet.detect(res.data) || "utf-8";
```
brunnre8, line 83: I don't really want to guess.
We have in most cases a transport that knows what the encoding is.

Guessing leads to quirky encoding issues and is problematic for various security reasons, if a malicious payload is loaded.

brunnre8 (changes requested):
Thank you for the contribution.
I do thing you should tackle this from another approach though.

Code paths that do support it, namely the html one, shouldn't guess but use the encoding that's specified in the transport layer.

server/plugins/irc-events/link.ts
sideshowbarker: > I don't really want to guess. We have in most cases a transport that knows what the encoding is.
> 
> Guessing leads to quirky encoding issues and is problematic for various security reasons, if a malicious payload is loaded.

👍 Thanks much — that makes more sense; I’ll rework the patch to make it do that.

server/plugins/irc-events/link.ts
```diff
@@ -78,8 +81,22 @@ function parseHtml(preview, res, client: Client) {
...
+			if (!charset) {
+				const mimeType = new MIMEType(
+					$('meta[http-equiv="Content-Type" i]').attr("content")
+				);
+				charset = mimeType.parameters.get("charset");
+			}
+			if (charset && !/^utf-?8$/i.test(charset)) {
+				$ = cheerio.load(Iconv.decode(res.data, charset));
+			}
+		}
```
brunnre8, line 99: You are overcomplicating this...

```js
import * as cheerio from "cheerio"
import MIMEType from "whatwg-mimetype";

function get_encoding(resp){
	const fallback = 'utf-8'
	const ct = resp.headers.get('content-type')
	if (!ct){
		return fallback
	}
	const mimeType = new MIMEType(ct);
	const transport = mimeType.parameters.get("charset")
	if (!transport){
		return fallback
	}
	return transport
}

const resp = await fetch("https://www.w3.org/2025/tests/shift_jis-encoding.html")
const body = await resp.bytes()
const $ = cheerio.loadBuffer(body, {encoding: {transportLayerEncodingLabel: get_encoding(resp)}})
console.log($('body').text())

```

brunnre8: note: needs a guard around `new MIMEType` in your actual version... it blows up on invalid input.

sideshowbarker: OK thanks much — I do see now that the `encoding-sniffer` code should handle what needs to be done (both checking for the charset in the `Content-Type` header and in any `meta` element in the document itself).

And I had earlier tried calling `cheerio.loadBuffer` — but when I do in my local environment, I get a compile error:

> server/plugins/irc-events/link.ts:98:45 - error TS2339: Property 'loadBuffer' does not exist on type '{ default: CheerioAPI; version: string; load(html: string | Buffer, options?: CheerioParserOptions | undefined): Root; load(element: Element | Element[], options?: CheerioParserOptions | undefined): Root; ... 4 more ...; xml(dom?: string | ... 2 more ... | undefined): string; }'.

```js
98                                 $ = cheerio.loadBuffer(body, {encoding: {transportLayerEncodingLabel: get_encoding(res)}});
                                               ~~~~~~~~~~
```
And my LSP integration in my editor gives me the same error; the only methods and fields it shows exposed from the `cheerio` object are `load`, `xml`, `html`, `root`, `default`, `version`, `contains`, and `parseHTML`.

So I guess there’s something I’m missing about what else I need to do in my environment in order to be able to call that `loadBuffer` method.

brunnre8: `yarn remove @types/cheerio`

The external types are completely bogus.

sideshowbarker: OK, I’m finding that `cheerio.loadBuffer` doesn’t seem to work as expected in the case where there’s no encoding  declared in the transport but instead only declared in a `meta` element in the document.

As far as I understand from https://cheerio.js.org/docs/api/functions/loadBuffer and from code inspection of the `encoding-sniffer/dist/commonjs/sniffer.js` code, that code _should_ be able to get the encoding from the `meta` element.

The original document I initially began testing with is at `https://www.ntt-west.co.jp/news/1504/150417a.html`, and it doesn’t have an encoding declared in the `Content-Type` header — instead, it just has a `<meta charset="shift_jis">` element.

When I try the script in https://github.com/thelounge/thelounge/pull/4951#discussion_r1912443652 with that document, it doesn’t appear to decode that document correctly. That is, this:

```js
import * as cheerio from "cheerio"
import MIMEType from "whatwg-mimetype";

function get_encoding(resp){
	const fallback = 'utf-8'
	const ct = resp.headers.get('content-type')
	if (!ct){
		return fallback
	}
	const mimeType = new MIMEType(ct);
	const transport = mimeType.parameters.get("charset")
	if (!transport){
		return fallback
	}
	return transport
}

const resp = await fetch("https://www.ntt-west.co.jp/news/1504/150417a.html")
const body = await resp.bytes()
const $ = cheerio.loadBuffer(body, {encoding: {transportLayerEncodingLabel: get_encoding(resp)}})
console.log($('body').text())
```

It _does_ work as expected for the `https://www.w3.org/2025/tests/shift_jis-encoding.html` document (which declares the encoding in its `Content-Type` header)

But it doesn’t work for the `https://www.ntt-west.co.jp/news/1504/150417a.html` document.

Calling encoding-sniffer’s `getEncoding` with the bytes from that document returns `windows-1252`.

brunnre8: meta charset=shift_jis is out of spec and hence broken

>The charset attribute specifies the [character encoding](https://encoding.spec.whatwg.org/#encoding) used by the document. This is a [character encoding declaration](https://html.spec.whatwg.org/multipage/semantics.html#character-encoding-declaration). If the attribute is present, its value must be an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "utf-8".

In other words, try to fix the webpage or drop the site owner an email

sideshowbarker: Thanks — after doing some further investigation, I found that using the alternative label `shift-jis` (with a dash) actually works. See https://www.w3.org/2025/tests/shift-dash-jis-in-meta.html

In fact, I found that _every_ label I tried that’s for an encoding supported by iconv-lite (which is what `cheerio.loadBuffer` ends up calling into, by way of encoding-sniffer) also works — except for one: `koi8_r`.

Here’s a script that tests some of the encoding labels:

```js
const { getEncoding } = require("encoding-sniffer");

const log_encoding = async (label, url) => {
	const response = await fetch(url);
	const document = await response.bytes()
	console.log(`${label.padEnd(12, " ")} ${getEncoding(document).padEnd(15, " ")} ${url}`);
}

(async () => {
	await log_encoding("shift_jis", "https://www.w3.org/2025/tests/shift-underscore-jis-in-meta.html");
	await log_encoding("shift-jis", "https://www.w3.org/2025/tests/shift-dash-jis-in-meta.html");
	await log_encoding("koi8_r", "https://www.w3.org/2025/tests/koi8-underscore-r-in-meta.html");
	await log_encoding("koi8-r", "https://www.w3.org/2025/tests/koi8-dash-r-in-meta.html");
	await log_encoding("EUC-JP", "https://www.w3.org/2025/tests/euc-jp-encoding-in-meta.html");
	await log_encoding("EUC-KR", "https://www.w3.org/2025/tests/euc-kr-encoding-in-meta.html");
	await log_encoding("GBK", "https://www.w3.org/2025/tests/gbk-encoding-in-meta.html");
	await log_encoding("Big5", "https://www.w3.org/2025/tests/big5-encoding-in-meta.html");
})();
```

Output:

```
shift_jis    windows-1252    https://www.w3.org/2025/tests/shift-underscore-jis-in-meta.html
shift-jis    Shift_JIS       https://www.w3.org/2025/tests/shift-dash-jis-in-meta.html
koi8_r       windows-1252    https://www.w3.org/2025/tests/koi8-underscore-r-in-meta.html
koi8-r       KOI8-R          https://www.w3.org/2025/tests/koi8-dash-r-in-meta.html
EUC-JP       EUC-JP          https://www.w3.org/2025/tests/euc-jp-encoding-in-meta.html
EUC-KR       EUC-KR          https://www.w3.org/2025/tests/euc-kr-encoding-in-meta.html
GBK          GBK             https://www.w3.org/2025/tests/gbk-encoding-in-meta.html
Big5         Big5            https://www.w3.org/2025/tests/big5-encoding-in-meta.html
```

So it seems maybe encoding-sniffer has a bug in its handling of encoding labels with underscores. So I’ve raised an issue at https://github.com/fb55/encoding-sniffer/issues/927, and I’m looking into the code there myself too, to see if I can isolate the cause.

Given that, I’d like to go ahead with consideration of this PR — so I’ve updated the patch to align with the approach given in https://github.com/thelounge/thelounge/pull/4951#discussion_r1912443652.

Then if/when https://github.com/fb55/encoding-sniffer/issues/927 gets resolved and into a release upstream, we could update encoding-sniffer to that release version to get the fix.

sideshowbarker: OK, found the cause of the encoding-sniffer bug and wrote a patch https://github.com/fb55/encoding-sniffer/pull/928/files.

(It was lowercasing encoding names/labels using a `c | 0x20` bitwise operation, which has the unintended effect of turning underscores into `0x7f` ␡ characters.)

server/plugins/irc-events/link.ts
```diff
@@ -464,15 +472,25 @@ function fetch(uri: string, headers: Record<string, string>) {
...
+						try {
+							const mimeType = new MIMEType(contentType);
+							type = mimeType.essence;
+							const charsetParam = mimeType.parameters.get("charset");
+							if (charsetParam) {
+								charset = charsetParam;
+							}
+						} catch {
+							// the Content-Type value isn't valid; ignore it
+						}
```
brunnre8, line 490: use a function same as I did last n my example

server/plugins/irc-events/link.ts
```diff
@@ -167,7 +175,7 @@ function parseHtmlMedia($: any, preview, client: Client): Promise<FetchRequest>
 				return;
 			}
 
-			$(`meta[property="og:${type}:type"]`).each(function (this: cheerio.Element, i: number) {
+			$(`meta[property="og:${type}:type"]`).each(function (this: any, i: number) {
```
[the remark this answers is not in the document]
sideshowbarker: > any isn't a type I want to add anywhere

Changed it to just `Element`

server/plugins/irc-events/link.ts
sideshowbarker: > use a function, same as I did in my example

OK, I made it a `getTransportLayerEncodingLabel(contentType)` function — since the existing `fetch` function is already doing a `res.headers["content-type"]` to get the `contentType`.

If that’s not what you had in mind, please let me know, and I’ll take another crack at it.

server/plugins/irc-events/link.ts
```diff
@@ -78,8 +80,14 @@ function parseHtml(preview, res, client: Client) {
 	// TODO:
 	// eslint-disable-next-line @typescript-eslint/no-misused-promises
 	return new Promise((resolve: (preview: FetchRequest | null) => void) => {
-		const $ = cheerio.load(res.data);
-
+		let $;
```
brunnre8, line 83: this probably doesn't need the conditional. Chances are you can pass undefined there as a transport encoding and be fine.

server/plugins/irc-events/link.ts
```diff
@@ -167,7 +175,7 @@ function parseHtmlMedia($: any, preview, client: Client): Promise<FetchRequest>
 				return;
 			}
 
-			$(`meta[property="og:${type}:type"]`).each(function (this: cheerio.Element, i: number) {
+			$(`meta[property="og:${type}:type"]`).each(function (this: Element, i: number) {
```
brunnre8, line 178: nack, stop lying to the compiler.

brunnre8 (changes requested):

brunnre8:
the CI will yell at you too, fix those up as well

package.json
```diff
@@ -96,7 +97,6 @@
     "@textcomplete/textarea": "0.1.13",
     "@types/bcryptjs": "2.4.6",
     "@types/chai": "4.3.5",
-    "@types/cheerio": "0.22.35",
```
brunnre8, line 99: pretty sure that both of those changes made changes to the lockfile, no?

brunnre8: ah and exact version specs please for our direct deps

server/plugins/irc-events/link.ts
sideshowbarker: > this probably doesn't need the conditional. Chances are you can pass undefined there as a transport encoding and be fine.

OK, makes sense, will try that

server/plugins/irc-events/link.ts
sideshowbarker: > nack, stop lying to the compiler.

Well, I gotta confess: I don’t know that the actual type needs to be. The problem with `cheerio.Element` is that it relies on `@types/cheerio`, right? And if we have `@types/cheerio`, then we can’t use `cheerio.loadBuffer`?

server/plugins/irc-events/link.ts
```diff
@@ -488,6 +503,16 @@ function fetch(uri: string, headers: Record<string, string>) {
 	return promise;
 }
 
+function getTransportLayerEncodingLabel(contentType) {
```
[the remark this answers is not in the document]
sideshowbarker: > this needs types. Catch the error here and return undefined if you don't know

Will do

package.json
sideshowbarker: > pretty sure that both of those changes made changes to the lockfile, no?

So, about that: In my environment, for some reason `yarn install` causes _every single line_ of the `yarn.lock` file to be changed. For example, the following:

```diff
-"@ampproject/remapping@^2.1.0":
-  version "2.2.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
-  integrity sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==
-  dependencies:
-    "@jridgewell/gen-mapping" "^0.1.0"
-    "@jridgewell/trace-mapping" "^0.3.9"
-
```

…becomes this:

```diff
+"@ampproject/remapping@npm:^2.1.0":
+  version: 2.2.0
+  resolution: "@ampproject/remapping@npm:2.2.0"
+  dependencies:
+    "@jridgewell/gen-mapping": "npm:^0.1.0"
+    "@jridgewell/trace-mapping": "npm:^0.3.9"
+  checksum: 10c0/d267d8def81d75976bed4f1f81418a234a75338963ed0b8565342ef3918b07e9043806eb3a1736df7ac0774edb98e2890f880bba42817f800495e4ae3fac995e
+  languageName: node
+  linkType: hard
```
…for 13117 lines.

Any guidance on how I can avoid having that happen? It make the diff for `yarn.lock` pretty useless, as far as identifying what actual substantive change got made.

server/plugins/irc-events/link.ts
brunnre8: remove the types from the .each() function.
Type the $ argument passed into the current scope (the top level function) as to what it is.

Cheerio has built in types, it'll figure it out correctly

package.json
sideshowbarker: > sadly, no. I'm fighting yarn every time I use it as well

👍 I hear yah. OK, committed and pushed

brunnre8: I'm not gonna apply an obvious lock file insanity, you *do* have to debug it on your end

package.json
```diff
@@ -3,9 +3,7 @@
   "description": "The self-hosted Web IRC client",
   "version": "4.4.3",
   "preferGlobal": true,
-  "bin": {
-    "thelounge": "index.js"
-  },
+  "bin": "index.js",
```
brunnre8, line 6: ?

package.json
sideshowbarker: Understood. Will get it figured out

package.json
sideshowbarker: Ah that's some other unwelcome jackassery that yarn is performing. Hadn't intended to commit it (been doing  add -p to avoid it). Will back out out

sideshowbarker:
OK, I think I’ve resolved all the review comments. Lemme know if I got anything wrong

package.json
brunnre8: this is still there

server/plugins/irc-events/link.ts
```diff
@@ -167,7 +171,7 @@ function parseHtmlMedia($: any, preview, client: Client): Promise<FetchRequest>
 				return;
 			}
 
-			$(`meta[property="og:${type}:type"]`).each(function (this: cheerio.Element, i: number) {
+			$(`meta[property="og:${type}:type"]`).each(function (this, i) {
```
brunnre8, line 174: get rid of the this param

server/plugins/irc-events/link.ts
```diff
@@ -488,6 +499,15 @@ function fetch(uri: string, headers: Record<string, string>) {
 	return promise;
 }
 
+function getTransportLayerEncodingLabel(contentType: string): string | undefined {
+	const mimeType = new MIMEType(contentType);
```
brunnre8, line 503: again, this can fail, guard against it here

brunnre8 (changes requested):

sideshowbarker:
OK, thanks — just now pushed an update https://github.com/thelounge/thelounge/compare/812f65af800efc14f71cd6063dd13a2f97dde0fb..7eba8edfbd5d489cfdfe09a6eb7b097fb7802c52

server/plugins/irc-events/link.ts
```diff
@@ -488,6 +499,15 @@ function fetch(uri: string, headers: Record<string, string>) {
 	return promise;
 }
 
+function getTransportLayerEncodingLabel(contentType: string): string | undefined {
+	try {
+		const mimeType = new MIMEType(contentType);
+		return (mimeType.parameters.get("charset") as string) || undefined;
```
brunnre8, line 505: absolutely not, we don't introduce new cast like that.
They are a maintenance hassle and prevent the type checks from working.
Why did you do that? The code was compiling no?

sideshowbarker: > Why did you do that? The code was compiling no?

Yeah, the code was compiling fine, but I got notified about this workflow-run eslint failure:
https://github.com/thelounge/thelounge/actions/runs/12826124441/job/36982976340
> /home/runner/work/thelounge/thelounge/server/plugins/irc-events/link.ts
> 505:3  error  Unsafe return of an `any` typed value  @typescript-eslint/no-unsafe-return

So the update I pushed was a (ham-fisted) attempt to fix that.

sideshowbarker:
@brunnre8 Felix did a new release of `encoding-sniffer` that includes my fix for the labels-with-underscores bug: https://github.com/fb55/encoding-sniffer/releases/tag/v0.2.1

server/plugins/irc-events/link.ts
```diff
@@ -1,7 +1,9 @@
 import * as cheerio from "cheerio";
+import {CheerioAPI} from "cheerio";
 import got from "got";
 import {URL} from "url";
 import mime from "mime-types";
+import MIMEType from "whatwg-mimetype";
```
MaxLeiter, line 6: we should declare this in our dependencies and not rely on it being present transitively

MaxLeiter (changes requested):

server/plugins/irc-events/link.ts
sideshowbarker: Made it so.

sideshowbarker: @MaxLeiter I’m going ahead and marking this resolved. Lemme know if you see any other problems with this.

sideshowbarker:
Rebased to resolve the merge conflicts.

PR #4952, opened 2025-01-13
server: disconnect the socket when authentication fails

jkbzh:
Fixes a security issue where thelounge is open to bulk password guessing attacks: 

Without the patch, thelounge client opens a websocket connection to the server using HTTP GET Connection: Upgrade and then submits the authentication form using that websocket. If authentication fails, the next submission happens on the same open websocket. Although the authentication failure is logged, thelounge server doesn't take any other action. As common firewalls, such as cloudflare's IP Lists and fail2ban/iptables, apply to the opening of a connection, this behavior allows to bypass them and do bulk password guessing attacks.

With the patch, thelounge server will close the websocket when authentication fails, thus forcing the client to open a new websocket connection using HTTP GET Connection: Upgrade, allowing firewalls to block that request if given conditions are triggered.

### DIFF
--- a/client/js/socket-events/auth.ts
+++ b/client/js/socket-events/auth.ts
@@ -12,12 +12,14 @@ declare global {
 }
 
 socket.on("auth:success", function () {
+	store.commit("isAuthFailure", false);
 	store.commit("currentUserVisibleError", "Loading messages…");
 	updateLoadingMessage();
 });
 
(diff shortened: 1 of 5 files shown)


jkbzh:
Closing this PR in favor of a forthcoming one from @kfranqueiro which approaches the bulk guessing attacks in a more elegant way.

PR #4970, opened 2025-04-09
Trim username in sign-in form

sfan5:
with e.g. autocompletion it's easy to get spaces in there which results in bad UX


brunnre8:
Thank you for the effort, sorry for the delay in merging...

PR #4973, opened 2025-05-12
Fix state machine to prevent overwriting username

jessielw:
This PR adds support to lock the users nick, disallowing the random numbered fall back via a configurable checkbox in the UI and config. **The old behavior is retained when the box is not ticked**. 

It works by remembering the desired nick and attempting to reconnect with that nick every **30 seconds** _(interval is configurable via the user config, but not in the front end)_. So, when enabled, it will only ever reconnect with the users nick and never attempt to append random numbers to it.

image

https://github.com/thelounge/thelounge/issues/4816
https://github.com/thelounge/thelounge/issues/4350


Xiseous:
I see this was posted in May, good feature, especially when connecting over VPNs that have random disconnects and no one has looked at this yet.

jessielw:
> Thanks for your work. However, I don't think that this is the right solution. We shouldn't be delaying the connection, rather we should fix the state machine so that TL will just re-grab the nick as expected (and stop overwriting the user prefs).

Thanks, that _does_ sound like a better way to handle it. Unfortunately, I have some other things I have to take care of atm. I will try to look at this again in a few days or so, but if anyone is interested in taking over that wants it in sooner I have no objections!

jessielw:
I'll take a swing at fixing this *this week* if time allows

jessielw:
@brunnre8 Change are complete with the keep nick changes reverted

brunnre8:
the diff contains random changes and there's no tests that proof that this works?

jessielw:
> the diff contains random changes and there's no tests that proof that this works?

Not sure what happened. For some reason my commits got corrupted and an automatic format plugin messed everything up in my IDE. I reset everything and applied only the needed changes to fix the state machine. 

------

I attempted to do a mock test, but it wasn't super trivial to mock the IRC server that easily in node. I did a mock up of an IRC server in python to do the test with. I wish I could have done a _proper_ test for the repo, but it was very minimal code changes needed to actually fix the problem and I couldn't figure out a simple way to unit test it. 

Python server terminal (script will be attached in this post below):
[test-nick-fallback.py](https://github.com/user-attachments/files/24997414/test-nick-fallback.py)
```
python .\test-nick-fallback.py         
   
Starting IRC Test Server...
Type 'help' for commands


============================================================
IRC Test Server Running on 127.0.0.1:6667
============================================================

[TAKEN] Nick 'testnick' is already in use

TEST INSTRUCTIONS:
1. Start The Lounge: npm start
2. Connect to localhost:6667
3. Use nick: testnick
4. You should get a fallback nick (e.g., testnick1)
5. WAIT 60 seconds - nick will auto-release
6. The Lounge should automatically reclaim 'testnick'

OR type 'release' to free the nick immediately

============================================================

[CONNECT] New connection from ('127.0.0.1', 51872)

[RECV] CAP LS 302
[SEND] CAP * LS :
[RECV] NICK testnick
[SEND] 433 Nick in use: testnick
[INFO] Client should try fallback...
[RECV] USER testnick 0 * testnick
[USER] Set username to: testnick
[RECV] CAP END
[RECV] NICK testnick3
[ACCEPTED] Nick set to: testnick3
[REGISTERED] testnick3 connected successfully
[RECV] JOIN #test
[JOIN] testnick3 joined #test
release
[RELEASED] Nick 'testnick' is now available!
[BROADCAST] Sending QUIT to all clients...
[SENT] QUIT message to testnick3
[INFO] The Lounge should now reclaim it automatically
[RECV] NICK testnick
[ACCEPTED] Nick set to: testnick
[SEND] NICK change: testnick3 -> testnick
[SUCCESS] ✓ Nick reclaimed!
[RECV] PING 1769989177401

============================================================
[AUTO-RELEASE] 60 seconds elapsed!
[RELEASED] Nick 'testnick' is now available!
[BROADCAST] Sending QUIT to all clients...
[SENT] QUIT message to testnick
[INFO] The Lounge should now reclaim it automatically
============================================================
```

Here is output from the lounge maintaining the proper state and reclaiming the username
image

brunnre8:
The problem is that without tests, you can for one easily regress.
Then, you've run a single scenario, but what about the rest of the state machine? Did you now get all the transitions correct? Did your patch introduce problems somewhere else?

You don't know, you can't test all possible scenarios from the top of your head all the time, the person that has to do that might not even be you personally but someone else that doesn't have all possible variants in their head.

server/plugins/irc-events/welcome.ts
```diff
@@ -6,7 +6,19 @@ export default <IrcEventHandler>function (irc, network) {
...
+		if (data.nick === network.nick) {
+			// We got our preferred nick, clear keepNick if it was set
+			if (network.keepNick === data.nick) {
+				network.keepNick = null;
+			}
+		} else if (data.nick !== network.nick) {
+			// We registered with a fallback, don't call setNick which would overwrite the preference
+			// Just update the IRC session nick
+			irc.user.nick = data.nick;
+		}
```
brunnre8: That whole block shouldn't be inline in the registration part, you're adding to the spaghetti code.

The keep nick field is owned by the network (for some reason) and hence should ONLY be modified from methods on the network. Part of why the state machine is broken is exactly because stuff gets modified in tons of places and no one really looks for that when modifying code.

Extract the stuff into its own class, then testing is trivial. You don't mock the ircd, you extract the functionality into a NickKeeper class and feed *that* the state transitions.

It becomes trivial to test, the rest is just calling the update methods in the proper places which is much easier to review to be correct.

brunnre8 (changes requested):

server/plugins/irc-events/welcome.ts
jessielw: I completely understand why it's important but I was attempting to do a quick patch for code that was already spaghetti...

So with the latest commit I've fixed the base code and implemented the requested feature. I've also added testing for said issue and ran all tests in the project. 

```
 npm run test:mocha

> thelounge@4.4.3 test:mocha
> webpack --mode=development && cross-env NODE_ENV=test TS_NODE_PROJECT='./test/tsconfig.json' mocha --config=test/.mocharc.yml 'test/**/*.ts'

Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme


  ...........................................................................................
  ...........................................................................................
  .........................................................................

  255 passing (5s)
  ```
  
  Let me know if you need anything else before merging this in.

**Comment by jessielw** [comment]
> The problem is that without tests, you can for one easily regress. Then, you've run a single scenario, but what about the rest of the state machine? Did you now get all the transitions correct? Did your patch introduce problems somewhere else?
> 
> You don't know, you can't test all possible scenarios from the top of your head all the time, the person that has to do that might not even be you personally but someone else that doesn't have all possible variants in their head.

The changes went from +18 to +228 changes, but everything is now passing with the new NickKeeper class.

**Tests**
```
 npm run test:mocha

> thelounge@4.4.3 test:mocha
> webpack --mode=development && cross-env NODE_ENV=test TS_NODE_PROJECT='./test/tsconfig.json' mocha --config=test/.mocharc.yml 'test/**/*.ts'

  ...........................................................................................
  ...........................................................................................
  .........................................................................

  255 passing (5s)
  ```

opeeba:
@brunnre8 and @jessielw,
Of all the projects I follow, this has been the most civil and fun to watch conversation. Thank you for your efforts and I look forward to the implementation of this feature whenever and however it comes.

jessielw:
> @brunnre8 and @jessielw, Of all the projects I follow, this has been the most civil and fun to watch conversation. Thank you for your efforts and I look forward to the implementation of this feature whenever and however it comes.

No problem! This issue has been a pain in the butt for quite some time, especially when updating my docker container at least a handful of the services will modify my nick name.

Yapping7409:
Thanks for all the work @jessielw. I hope this gets merged soon.

jessielw:
@brunnre8 have you had the time to take a look at this?

server/models/network.ts
```diff
@@ -121,6 +122,7 @@ class Network {
 	chanCache!: Chan[];
 	ignoreList!: IgnoreList;
 	keepNick!: string | null;
+	private nickKeeper?: NickKeeper;
```
brunnre8: this should not be optional, do the initialization in the network constructor

brunnre8:
So you're now signaling through the nick keeper, which seems much cleaner.
But why didn't you just push the state into it?
As is, the functionality is still spread out all over the place with relatively magic callbacks where you have the network ask the nick keeper what to do and then still all the actual logic is done within the network.

brunnre8:
To be a bit more concrete, what I had in mind (very fuzzy, take with a huge grain of salt) is to just have two string in the thing, desired nick and current nick.
You init the keeper and then just send it events (nick left, you got a new current nick (registration) etc).
Then invoke a callback with "change nick" that may very well be in the style that you just already have there, but it should be driven by the nick keeper. The network would know nothing about the current nick, it would just forward it directly to the thing that cares.

If you really want to you can keep the current nick a property of the net, but given that you need to signal ANYWAY when it changes or becomes active during connection registration, you might as well keep track of it already and safe yourself a relatively pointless indirection.

server/plugins/irc-events/error.ts
```diff
@@ -57,20 +56,16 @@ export default <IrcEventHandler>function (irc, network) {
...
+				// Only change the IRC session nick (irc.user.nick), not the user's preference (network.nick)
+				// This allows the quit handler to reclaim the preferred nick when it becomes available
 				irc.changeNick(random);
 			}
 		}
-
-		client.emit("nick", {
-			network: network.uuid,
-			nick: irc.user.nick,
-		});
```
brunnre8, line 73: guess you rely on the irc.on("nick") callback to then tell the client once the retry succeeds?

jessielw:
@brunnre8 it's been rewritten with your suggestions and tests

- Removed the old keepNick owner-adapter pattern from Network.
- NickKeeper now owns desired/current/pending nick state and exposes event-style methods.
- IRC handlers now forward events into network.nickKeeper and execute callbacks for side effects.
- The fallback retry path still does not emit "nick" early; successful nick changes continue through irc.on("nick").
- Rewrote test/tests/nickKeeper.ts around state transitions and added a Network constructor test.

jessielw:
@brunnre8 want to get this in?

MaxLeiter:
has a conflict now but when fixed we can get this in for 4.6! thanks!

jessielw:
> has a conflict now but when fixed we can get this in for 4.6! thanks!

Thanks, this is now resolved!

server/models/network.ts
```diff
@@ -408,7 +410,7 @@ class Network {
 		const oldNick = this.nick;
 		const oldRealname = this.realname;
 
-		this.keepNick = null;
+		this.nickKeeper.cancelPendingNick();
```
brunnre8: Don't do this, you are violating the layers this way.
The whole point of pushing the state machine into the nick keeper is that you DON'T have to care from the client / network bits.

The client network bits should push transition signals into the nick keeper, it should not call what should be a private helper method.

The user either changed their nick that they want, at which point you get a signal, or they did not then you don't from the keeper perspective.

server/models/nickKeeper.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export type RegisteredCallbacks = {
+	setPreferredNick(nick: string): void;
+	setCurrentNick(nick: string): void;
+};
+
+export default class NickKeeper {
+	private desiredNick: string;
+	private currentNick: string;
+	private pendingNick: string | null = null;
```
brunnre8: Why do we need 3 nick fields?

server/models/nickKeeper.ts
```diff
@@ -0,0 +1,84 @@
...
+		this.pendingNick = null;
+	}
+
+	onNickInUse(context: NickInUseContext) {
+		if (!context.registered && !context.isPublic) {
+			this.pendingNick = this.desiredNick;
+		}
+	}
+
+	onRegistered(registeredNick: string, callbacks: RegisteredCallbacks) {
```
brunnre8: This starts to be a bit funny, You have events above from the state of the keeper (nick changed / nick in use) that's ok.
Now you're talking about an IRC upstream thing suddenly, which should be isolated from in this component

Further down you suddenly start talking about network events (onSocketClose) which seems rather odd.

Like, sure, I get it where you get those things from when porting the existing code. But somehow this seems funny to me.

Again, the point is isolation (and insulation) so the events should be on the perspective of what the keeper cares about, not about the protocols / logic of the consumers

Style wise, don't call things onSomething if it isn't just registering a callback.
You are mixing the callback registration and the state transitions in one method, which I don't think is very good (particularly in onNickInUse, that isn't a callback).
You should register the callbacks once (they don't change, there's only one way to change the nick in the upstream protocol after all)

server/plugins/inputs/nick.ts
```diff
@@ -42,8 +42,7 @@ const input: PluginInputHandler = function (network, chan, cmd, args) {
 		return;
 	}
 
-	// If we were trying to keep a nick and user changes nick, stop trying to keep the old one
-	network.keepNick = null;
+	network.nickKeeper.cancelPendingNick();
```
brunnre8: layering violation, call of what should be a private method

test/tests/nickKeeper.ts
```diff
@@ -0,0 +1,184 @@
...
+			},
+			setCurrentNick(nick: string) {
+				calls.currentNick = nick;
+			},
+		},
+	};
+}
+
+describe("NickKeeper", function () {
+	it("stores desired nick when nick in use on connect", function () {
```
brunnre8: what do you mean by "store" in the test descriptions?
As far as I can tell you are testing if the keeper changes the nick if whatever holds the nick quits, there's no store involved there.

Same applies to tests below

test/tests/nickKeeper.ts
```diff
@@ -0,0 +1,184 @@
...
+		const keeper = new NickKeeper("preferred");
+		let reclaimedNick: string | null = null;
+
+		keeper.onNickInUse({registered: false, isPublic: true});
+		keeper.onQuit("preferred", (nick) => {
+			reclaimedNick = nick;
+		});
+
+		expect(reclaimedNick).to.equal(null);
+	});
```
brunnre8: Should it do that?
Is this simply due to the existing code or where is this from?

server/models/nickKeeper.ts
```diff
@@ -0,0 +1,84 @@
+export type NickInUseContext = {
+	registered: boolean;
+	isPublic: boolean;
+};
+
+export type RegisteredCallbacks = {
+	setPreferredNick(nick: string): void;
```
brunnre8: This callback shouldn't exist in my view.
Push the state into the keeper, don't also keep it in the network.
It's really a one way street, all the nick handling is now in the keeper, send it signals what happened.
Don't make this a triangular affair, there's no reason for the keeper to report back what its internal state is.
The network should not have any concept of a desired nick, pushing that into the keeper is the point

brunnre8 (changes requested):
Thanks for updating the PR (and sorry for not responding for a while, university got in the way).

I think we can still improve this a bit, but it certainly looks better than the last time I've reviewed things.

server/models/nickKeeper.ts
jessielw: Gone, both callbacks and the RegisteredCallbacks type with them. The keeper takes one callback in the constructor now, (nick) => this.irc?.changeNick(nick), and handlers only push in: wantNick, nickConfirmed, nickRefused, nickReleased. welcome.ts went from the callbacks object down to one setCurrentNick(data.nick) call.

One thing I kept, so you don't have to find it: export()/exportForEdit() and the socket close handler read nickKeeper.desiredNick. That's a read for persistence and the reconnect, not the keeper driving logic. I couldn't push it through the callback because Network has no client reference, so it can't emit the nick event. This would need re-worked if you wanted that.

server/models/nickKeeper.ts
jessielw: Down to two. pendingNick was only recording "we're waiting for a nick", which is just current !== desired. That's also what fixed the two layering violations you flagged. Cancelling isn't an operation anymore, it falls out of wantNick(), so there's no cancelPendingNick() left to call from the wrong layer.

server/models/nickKeeper.ts
jessielw: Callbacks are registered once in the constructor and nothing is named onX anymore.

onRegistered is gone - registration is just "the server confirmed this nick", the same signal a confirmed NICK gives, so welcome.ts and nick.ts both call nickConfirmed. onSocketClose is gone, the keeper has no idea sockets exist; connection.ts restores the local nick itself. onNickInUse is gone too: registered is now the handler choosing which signal to send, and isPublic is a constructor option.

server/models/network.ts
jessielw: Removed. It was dead code anyway.  _edit() calls validate() which calls setNick(), so the intent was already getting pushed and the cancel did nothing. edit() now pushes wantNick(args.nick) and compares against nickKeeper.desiredNick._

server/plugins/inputs/nick.ts
jessielw: Now network.setNick(newNick), moved above the connected/disconnected branch so the intent is recorded either way. The reason a cancel was there at all is that the connected branch returns early and never reached setNick.

test/tests/nickKeeper.ts
jessielw: Fair! Nothing was stored, they were testing the reclaim. Renamed them all to say what they actually assert and grouped by method: "asks for the desired nick once its holder releases it", "ignores a released nick that is not the one we want", "keeps the desired nick when the server renames us against our will", and so on.

test/tests/nickKeeper.ts
jessielw: It's inherited, not something I added. Comes from 600115b8

jessielw:
> Thanks for updating the PR (and sorry for not responding for a while, university got in the way).
> 
> I think we can still improve this a bit, but it certainly looks better than the last time I've reviewed things.

I spent a few hours re-working parts of this based on your replies. I pushed two commits - the second one (the invalid nick change) isn't required by your review so you can drop it if you disagree with it. 

Two behavior notes worth looking at:
- A /nick refused while already connected stops the chase. So, only connect time collisions setup for a reclaim. Same as today, without it you'd get renamed hours later by a nick you typo'd. 
- A forced rename setups the reclaim. Since current `current !== desired` doesn't care how they diverged. 

Note: I also left nick comparison **case-sensitive** since this is how the old code was. I.e **MyNick** quitting won't release **mynick**. 

Let me know if you think there are more changes/improvements that we need to make

jessielw:
A couple quick things I noticed, but not necessarily in the scope of this particular PR _(though it could be added)_:

1. TL's self-nick detection works by accident of irc-framework's emit ordering

nick.ts does `const self = data.nick === irc.user.nick`. That only works because of irc-framework/src/commands/handler.js:88-92:

```ts
original_emit.apply(this, ['all'].concat(args));  // 'all' fires first
original_emit.apply(this, args);                  // specific event second
```

'all' is what proxies through to client.emit('nick', ...) and runs TL's handler, so it runs before irc-framework's own commands.on('nick') listener updates client.user.nick. Swap those two lines upstream and self-detection silently inverts - every own-nick change would be treated as someone else's.

2. irc-framework refuses to track a nick starting with a digit

client.js:271-277:

```ts
if (client.user.nick === event.nick) {
    // nicks starting with numbers are reserved for uuids
    if (event.new_nick.match(/^\d/)) {
        return;
    }
    client.user.nick = event.new_nick;
}
```

If a server renames you to something numeric, irc.user.nick goes permanently stale, which then breaks the self-detection in point 1 above for every later nick event. Not reachable through TL's own fallback (nick + digit keeps the leading character), but some ircds do hand out numeric nicks on collision.

server/models/network.ts
```diff
@@ -120,7 +122,7 @@ class Network {
 
 	chanCache!: Chan[];
 	ignoreList!: IgnoreList;
-	keepNick!: string | null;
+	nickKeeper!: NickKeeper;
```
brunnre8: Don't assert that thing, it's set in the constructor.
The assertions all over Network are baggage from the JS days, they shouldn't exist at all anywhere

server/models/network.ts
```diff
@@ -89,6 +90,7 @@ export type NetworkConfig = {
 };
 
 class Network {
+	// The nick we currently have; the one the user asked for is in nickKeeper
 	nick!: string;
```
brunnre8: Why do you still have a nick prop?
That should go away, keep the state once only server side

brunnre8:
>TL's self-nick detection works by accident of irc-framework's emit ordering

This is slightly concerning... But at least currently the code also messes with private internals of the framework which is much worse so... I guess we have bigger problems.

brunnre8:
>A /nick refused while already connected stops the chase. So, only connect time collisions setup for a reclaim. Same as today, without it you'd get renamed hours later by a nick you typo'd.

Keep the behavior as it is now, any changes can be made in follow up PRs. It's complicated enough to just replicate what's there I think.

>A forced rename setups the reclaim. Since current current !== desired doesn't care how they diverged.

IIUC this is fine, as long as we don't start looping. But you only trigger when the other nick leaves, so unless they join spam you should be ok. (Say if someone has guard enabled via chanserv or ghosts in the connect, you'll retry once they leave but otherwise don't right?)

server/models/network.ts
jessielw: Fixed

server/models/network.ts
jessielw: Fixed

jessielw:
> > A /nick refused while already connected stops the chase. So, only connect time collisions setup for a reclaim. Same as today, without it you'd get renamed hours later by a nick you typo'd.
> 
> Keep the behavior as it is now, any changes can be made in follow up PRs. It's complicated enough to just replicate what's there I think.

Agreed, it does what it does today. I also put back the client.emit("nick", …) at the end of the nick in use handler that I'd dropped.

The one behaviour change I did keep is the nick invalid line, and only because without it the split of setNick/setCurrentNick would change behaviour: today the forced change to the default nick lands in network.nick via welcome, so it gets persisted and reused on reconnect. With this, welcome only sets the current nick, so the invalid one stays as the desired nick and we'd retry it on every connect.

> IIUC this is fine, as long as we don't start looping. But you only trigger when the other nick leaves, so unless they join spam you should be ok. (Say if someone has guard enabled via chanserv or ghosts in the connect, you'll retry once they leave but otherwise don't right?)

Right. The only thing that ever sends a NICK is nickReleased, and that's called from exactly two places: a QUIT, and a NICK from someone who isn't us. Our own nick changes go down the if (self) branch, so we can't trigger ourselves (that's our loop guard)

ChanServ guard specifically: it renames us off the nick rather than someone taking it, so nobody ever releases it and we don't retry at all. Ghost on connect: one NICK when the ghost's QUIT comes in, then we're done (either the server confirms and we're level, or it 433s and nickRefused stops it).

Worst case is someone camping the nick with quit/join spam, and that's one NICK per quit. One difference from today: today keepNick is cleared the moment we send the NICK, so it only ever retries once. Here the chase ends when the server confirms, not when we ask (that falls out of the condition being derived from current !== desired once pendingNick went away).

One gap I should flag: if the reclaim itself gets 433 because someone grabbed the nick in the race, nickRefused treats it like a refused /nick and sets desired back to the fallback, which then gets persisted. Same as today, so no regression, but it's the one case where this PR doesn't fix the original bug. Telling those two apart needs state I just deleted per your request.

server/models/network.ts
```diff
@@ -496,8 +505,30 @@ class Network {
 		return this.ignoreList.some((entry) => Helper.compareHostmask(entry, data));
 	}
 
+	// The nick the server has us on; the one asked for is nickKeeper.desiredNick
+	get nick() {
+		return this.nickKeeper.currentNick;
+	}
```
brunnre8, line 515: this is too cute I think... we don't really use getters in this code base, just let the function call look like a function call. You already have setNick, so getNick won't kill you 😉

brunnre8:
>One gap I should flag: if the reclaim itself gets 433 because someone grabbed the nick in the race, nickRefused treats it like a refused /nick and sets desired back to the fallback,

I don't follow, you should never set the desired nick to anything unless it's coming from a direct user action (either network settings or a manual /nick).
Why is the server involved here? 433 is nick in use aye? So that should never nuke the user's direct choice.

Even if the server is unhappy about the desired nick choice of a user, that shouldn't matter, it's not your job to clean up bad input from the user, the user should clean this up. TL should just bubble up the error so that they are aware of it no?

What am I missing?

server/models/network.ts
jessielw: Oops, got this fixed

Bigger change than it sounds: with nick gone from the class, _.defaults was putting it straight back on the instance from the saved config, so that's _.omit(attr, "nick") and the constructor takes a Partial<Network> & {nick?: string}. Config.Defaults was picking nick off Network too, and inputs/mode.ts was destructuring it. Should be sound now!

jessielw:
> > One gap I should flag: if the reclaim itself gets 433 because someone grabbed the nick in the race, nickRefused treats it like a refused /nick and sets desired back to the fallback,
> 
> I don't follow, you should never set the desired nick to anything unless it's coming from a direct user action (either network settings or a manual /nick). Why is the server involved here? 433 is nick in use aye? So that should never nuke the user's direct choice.
> 
> Even if the server is unhappy about the desired nick choice of a user, that shouldn't matter, it's not your job to clean up bad input from the user, the user should clean this up. TL should just bubble up the error so that they are aware of it no?
> 
> What am I missing?

You're not missing anything, you're right and I had it backwards. I only added nickRefused because today keepNick is cleared on a registered-time 433, and I was trying not to change behaviour. But that only existed because keepNick was a one-shot (there's nothing to cancel here).

It's gone. The only things that write the desired nick now are validate(), edit() and /nick, all direct user actions.

- /nick taken while connected now sticks. You get the error, and TL keeps waiting for taken and persists it, so a restart still asks for it. That's the feature working, but it does differ from today, where a refused /nick is forgotten.
- I moved the "an attempt to use it will be made when this nick quits" line out of the registered === false branch, since it's now true either way and you said TL should surface it.

By the same rule I reverted the nick invalid commit _432_ was the server picking a nick for the user too. Error goes to the lobby and the config is left alone.

PR #4978, opened 2025-07-26
Add proxy protocol support to the identd server

ananthb:


brunnre8:
The patch is whitespace damaged and you didn't provide any sort of explanation what you are aiming for...

ananthb:
I'm going to re-open this when it's ready and when I have more time to work on it.

ananthb:
Of course. The PR wasn't quite ready yet as I'd mentioned.

I'm running thelounge on fly.io which doesn't let one directly accept TCP connections. I'm planning on running the identd server which means I'll need the server to understand the PROXY protocol. That's the gist of it.

PR #4984, opened 2025-08-23
Precise user configuration reloading.

kevincox:
Previously when any user config file was changed all users would be reloaded. This could be very expensive on installations with thousands of users. Since this triggered when any file was changed it would trigger when any user connected which can be quite frequent.

As a side-effect this removes the debouncing. This means that user changes take effect instantly rather than after 1s.

Since there is no longer a 1s delay it is extra important that files are written safely. To this end the `thelounge add <user>` command was updated to write the user file atomically. (The update path already did this.)

### DIFF
--- a/server/clientManager.ts
+++ b/server/clientManager.ts
@@ -86,39 +86,28 @@ class ClientManager {
 	}
 
 	autoloadUsers() {
-		fs.watch(
-			Config.getUsersPath(),
-			_.debounce(
-				() => {
-					const loaded = this.clients.map((c) => c.name);
-					const updatedUsers = this.getUsers();
+		fs.watch(Config.getUsersPath(), (_eventType, file) => {
+			if (!file.endsWith(".json")) {
+				return;
+			}
 
-					if (updatedUsers.length === 0) {
-						log.info(
-							`There are currently no users. Create one with ${colors.bold(
-								"thelounge add <name>"
-							)}.`
-						);
-					}
+			const name = file.slice(0, -5);
 
-					// Reload all users. Existing users will only have their passwords reloaded.
-					updatedUsers.forEach((name) => this.loadUser(name));
+			const userPath = Config.getUserConfigPath(name);
 
-					// Existing users removed since last time users were loaded
-					_.difference(loaded, updatedUsers).forEach((name) => {
-						const client = _.find(this.clients, {name});
+			if (fs.existsSync(userPath)) {
+				this.loadUser(name);
+				return;
+			}
 
-						if (client) {
-							client.quit(true);
-							this.clients = _.without(this.clients, client);
-							log.info(`User ${colors.bold(name)} disconnected and removed.`);
-						}
-					});
-				},
-				1000,
-				{maxWait: 10000}
-			)
-		);
+			const client = _.find(this.clients, {name});
+
+			if (client) {
+				client.quit(true);
+				this.clients = _.without(this.clients, client);
+				log.info(`User ${colors.bold(name)} disconnected and removed.`);
+			}
+		});
 	}
 
 	loadUser(name: string) {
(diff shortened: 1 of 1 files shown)


server/clientManager.ts
```diff
@@ -86,39 +86,27 @@ class ClientManager {
...
+			if (!file.endsWith(".json")) {
+				return;
+			}
+
+			const name = file.slice(0, -5);
+
+			const userPath = Config.getUserConfigPath(name);
+
+			if (fs.existsSync(userPath)) {
+				this.loadUser(name);
```
brunnre8, line 99: return here, you're done

server/clientManager.ts
```diff
@@ -86,39 +86,27 @@ class ClientManager {
...
+			} else {
```
brunnre8, line 100: then this doesn't need an else block

brunnre8 (commented):
Thanks for your contribution.

fs.watch can drop events if you don't read them fast enough, least on Linux.
When you say "thousands of users", did you test it with that amount?

Because if you dropped events with the old code, you'd be fine, as long as you read one event.
Now you'd bork the system state until you restart TL (or try again to modify the same user)

kevincox:
> fs.watch can drop events if you don't read them fast enough ... Because if you dropped events with the old code, you'd be fine, as long as you read one event.

Yes, I considered this but the limit is typically >10k events. If your process is so backed up to 10k events you probably have bigger problems so I figured it was fine not to worry about it. The node docs also aren't particularly clear on how this is handled. inotify will generate a `IN_Q_OVERFLOW` event on any drop but node doesn't say how this is handled. There is an `error` event on the listener but it isn't clear that this overflow is considered an error. It is also very hard to test this case.

[`fsPromises.watch`](https://nodejs.org/api/fs.html#fspromiseswatchfilename-options) does say that it can put into `throw` mode on overflow. This would be ideal if it works as expected as the fallback could do a full reload then return to watching. However the docs seem to imply that this is just the internal buffer in the node runtime, not if this includes overflows in the underlying inotify queue.

So given that it isn't clear if there is a good way to solve it the current solution seemed like the best tradeoff. I don't like making imperfect code but sometimes it is the best option. Especially since this will only occur when the machine (or at least the nodejs process) is grinding to a halt it seemed much better to drop a few events then to waste precious cycles reloading all of the config files every time any user connects.

> When you say "thousands of users", did you test it with that amount?

We have been testing for a bit over a day and it seems to be working well. But it is a bit early to say. Maybe after a week or so I can provide a better update. We also have automation to create users by dropping in files and updating passwords so it will test the corners.

kevincox:
We have been running this for a while now and the change is dramatic. Honestly far more than I expected (I think I underestimated just how often these user files are modified). We went from regularly capping out the CPU and occasionally crashing to ~1% CPU usage and rock solid. The user interface went from being very slow (many users were complaining about significant delays when sending messages and doing other basic interactions) to very snappy. It feels like a whole different product after shipping just this patch on top of the same release we were running previously (we backported this patch to v4.4.3 where it also applies cleanly). We have seen no issues with regards to user loading despite regularly creating new users and updating their passwords.

Its safe to say that we are going to continue running this patch whether or not it is accepted. This change flipped us from "we probably need to replace this, but maybe we can make some quick improvements" to "this is fine and we don't need anything else". We have cancelled our plans to try and prune less active users and are considering raising history retention (which is currently cranked down to the minimum to keep The Lounge from dying).

I still think that the approach this patch takes is the best available option. The very minor correctness risk will almost certainly only trigger in cases where The Lounge is running unusably slow anyways. NodeJS doesn't appear to provide a good cross-platform API to solve this more reliably so this patch is probably the best set of tradeoffs.

On a higher level a better design would have been to avoid filesystem watching as a critical function in the first place and have a first-class API for adding and updating users, but that ship has probably sailed. Maybe a user upsert API and an option to disable filesystem watching could be added, but that would not be a drop-in improvement for existing installations.

brunnre8:
Thanks for the feedback. Let's try and see if someone runs into issues.
Please address the style nit.

kevincox:
Great. I have updated to your preferred style.

brunnre8:
Thank you

PR #4987, opened 2025-09-18
fix broken manual channel ordering

Jay2k1:
This fixes a small regression from #4861 (specifically, commit 0067c30) that resulted in manually reordering channels not being received by the server and hence not saved or synced.


brunnre8:
thanks a bunch

PR #4991, opened 2025-09-28
feat: Add option to not send mobile push notifications if there is any active session

xpkmb:
# Summary

This PR introduces a new option to not send mobile push notifications if there is already an active session in any other device. When enabled, push notifications will only be sent to mobile devices if there is no active session (socket connection) in any other device.

Example:

1. Lounge is closed on phone AND desktop
Result: I get a **push** notification on phone

You will not get a mobile push notification in any other case.

I believe this contributes to https://github.com/thelounge/thelounge/issues/1441

image

### Notes:

This setting is disabled by default, so users will need to manually enable this setting to notice any changes.

### DIFF
--- a/client/components/Settings/Notifications.vue
+++ b/client/components/Settings/Notifications.vue
@@ -30,6 +30,16 @@
 					<span>Push notifications are not supported by your browser.</span>
 				</div>
 			</div>
+			<div>
+				<label class="opt">
+					<input
+						:checked="store.state.settings.onlyPushWhenInactive"
+						type="checkbox"
+						name="onlyPushWhenInactive"
+					/>
+					Only send push notifications when all sessions are inactive
+				</label>
+			</div>
 		</template>
 
 		<h2>Browser Notifications</h2>
@@ -87,17 +97,6 @@
 					name="notifyAllMessages"
 				/>
 				Enable notification for all messages
-			</label>
-		</div>
-
-		<div v-if="!store.state.serverConfiguration?.public">
-			<label class="opt">
-				<input
-					:checked="store.state.settings.onlyPushWhenInactive"
-					type="checkbox"
-					name="onlyPushWhenInactive"
-				/>
-				Only send push notifications when all sessions are inactive
 			</label>
 		</div>
 
--- a/client/components/Settings/Notifications.vue
+++ b/client/components/Settings/Notifications.vue
@@ -33,9 +33,9 @@
 			<div>
 				<label class="opt">
 					<input
-						:checked="store.state.settings.onlyPushWhenInactive"
+						:checked="store.state.settings.notifications.onlyPushWhenAllSocketsClosed"
 						type="checkbox"
-						name="onlyPushWhenInactive"
+						name="notifications.onlyPushWhenAllSocketsClosed"
 					/>
 					Only send push notifications when all sessions are inactive
 				</label>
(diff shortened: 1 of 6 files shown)


brunnre8:
I don't think we should be testing for desktop vs mobile at all, that check is brittle in the best case.
Having a toggle that stops emitting the notification if a user is "active" is fine, but that should be completely device agnostic. Doesn't matter what screen you are on, either you've seen the message and the notification is redundant or you have not.

Now, what's "active"? Just having a socket open isn't enough, it might be for you if you close out TL when you are not using it, but to give you a different viewpoint I've 3 open TL windows, in at least 2 different cities at almost all times. That's simply because some systems aren't normally shut down for various reasons and the tab simply stays put in the background.

Meaning your check for an active socket doesn't do what you think it does.

So when you want to implement this (and I guess we can) make the check simpler, and maybe think about how we should measure "activity"

Jay2k1:
As someone who has patched his own thelounge instance in the past to not send mobile push notifications when a client was connected, I'd welcome a built-in solution that does this in one way or another.

However, reading your post, I'm not so sure it does what I'd like it to, so I'd like to clarify a few things:

1. For me, when I'm using thelounge on at least one device, I don't need notifications anywhere, as it emits the pop sound and changes favicon and message indicator. But when no client can do this because no client is connected, then I'd like to get push notified, because I'd otherwise miss the query/highlight. 
Your demo cases are a bit misleading I think, because you mix mobile (push) notifications and browser notifications. One might choose to disable browser notifications, but use mobile push notifications. Obviously, when you have your client (browser) open somewhere and have enabled browser notifications, you will get them. So for this PR it would be more interesting to just look at the push situation.
So to me, only the first of your first demo cases was interesting in terms of changing the current situation, because the other three just get notified wherever thelounge is open. Maybe do the examples again, but with disabled browser notifications.
Also, the tooltip in the screenshot clearly says "mobile push notifications will be skipped if you have an active desktop session", but in example 3 (open on both), you wrote that both get notified, you probably see how this is confusing? :)
2. You've put the option in the "Browser notifications" settings group. Wouldn't it belong in the "Push notifications" settings group if it changes when *push notifications* are being sent?

My personal goal would be that I get push notifications when no client is connected, so I do not miss "important" things. 
If I have it open on a device, as said, I'm gonna hear a sound and see something - changed favicon and message indicators within the app. Maybe on Windows, as a PWA, the taskbar icon will start flashing, and on macOS there might be a badge on the dock icon. I happen to use it as PWA on a Mac, but as a browser tab on Windows, and I might be afk when a highlight comes in, so to not miss out I could always enable browser notifications.

xpkmb:
I have simplified my entire approach and updated the description of the pull request to reflect this. 

The new approach will only send push notifications if all sessions (from any device) are inactive. This means that we don't need to check if the session is from a desktop or mobile device anymore.

I also moved the setting to the appropiate section and updated its description for accuracy: `Only send push notifications when all sessions are inactive`

Thank you @Jay2k1 for the feedback. I am very satisfied now with the end result.

@brunnre8 I believe this solves one of your concerns regarding the user agents. The issue with defining an active session as "open socket" still remains, but I hope that is something that can be improved upon in the future if someone with more experience chimes in.

MaxLeiter (commented):
This LGTM but I will let @brunnre8 decide if its worth the merge or not.

MaxLeiter (approved):

PR #5001, opened 2025-11-04
Modernize: update compatible dependencies

marmitar:
Updates dependencies in `package.json` to the latest versions that are still compatible with Node 18 and CommonJS, with minimal source changes. The PR is split into small commits so it's easier to review each change, cherry-pick individual updates, or revert them later if needed. If you prefer, this can also be squashed or split into multiple PRs.

Besides the version bumps, this also unpins versions in `package.json` and lets `yarn.lock` handle exact versions instead. The lockfile is a better place for that, since it also pins transitive dependencies and keeps installs reproducible.

I did a quick manual smoke-test with Node 18 and didn't see any obvious regressions. I can do more targeted testing depending on the direction you want to take with this PR.


marmitar:
PR updated to re-sync dependencies. Also had to update the workaround for debug files in `socket.io-parser@4.2.5` (because of socketio/socket.io@7594aa4).

server/command-line/install.ts
```diff
@@ -1,16 +1,15 @@
...
 import {Command} from "commander";
-import {FullMetadata} from "package-json";
+import {FullVersion} from "package-json";
 
-type CustomMetadata = FullMetadata & {
-	thelounge: {
-		supports: string;
+type CustomMetadata = FullVersion & {
+	thelounge?: {
+		supports?: string;
```
brunnre8: what's that to do with dep updates?

server/command-line/install.ts
```diff
@@ -63,13 +61,13 @@ program
 			});
 		}
 
-		if (!readFile) {
+		if (readFile === null) {
```
brunnre8: ?

marmitar: TypeScript ESLint flags this with [no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises/) after it discovers that `readFile` is a promise.

> 64:8  error  Expected non-Promise value in a boolean conditional  @typescript-eslint/no-misused-promises

server/command-line/install.ts
marmitar: This is so the `readFile: Promise<CustomMetadata> | null` matches both initialization paths, either via `readFile(...).then((data) => JSON.parse(data))` or `packageJson({ fullMetadata: true, version: ... })`.

But indeed, not truly required. It's possible to either force cast `packageJson.default` back to this type or restore the previous `any` type. I just assumed this would be something you might be interested because of the `// TODO: type` comment.

marmitar:
Fixed the commit message for Webpack and package-json updates, and re-synced the dependencies.

server/command-line/install.ts
marmitar: At the bare minimum, the update only needs:

```patch
diff --git a/server/command-line/install.ts b/server/command-line/install.ts
--- a/server/command-line/install.ts
+++ b/server/command-line/install.ts
@@ -1 +0,0 @@
-/* eslint-disable @typescript-eslint/restrict-template-expressions */
@@ -9 +9 @@ import {Command} from "commander";
-import {FullMetadata} from "package-json";
+import {FullVersion} from "package-json";
@@ -11 +11 @@ import {FullMetadata} from "package-json";
-type CustomMetadata = FullMetadata & {
+type CustomMetadata = FullVersion & 
```

And then we can declare the missing type:

```patch
diff --git a/server/command-line/install.ts b/server/command-line/install.ts
--- a/server/command-line/install.ts
+++ b/server/command-line/install.ts
@@ -36,2 +36 @@ program
-		// TODO: type
-		let readFile: any = null;
+		let readFile: Promise<CustomMetadata> | null = null;
@@ -47 +46 @@ program
-				.then((data) => JSON.parse(data) as typeof packageJson);
+				.then((data) => JSON.parse(data) as CustomMetadata);
@@ -62 +61 @@ program
-			});
+			}).then((data) => data as unknown as CustomMetadata);
@@ -71 +70 @@ program
-			.then((json: CustomMetadata) => {
+			.then((json) => {
```

But avoid casts from that:

```patch
diff --git a/server/command-line/install.ts b/server/command-line/install.ts
--- a/server/command-line/install.ts
+++ b/server/command-line/install.ts
@@ -11,2 +11,2 @@ type CustomMetadata = FullVersion & {
-	thelounge: {
-		supports: string;
+	thelounge?: {
+		supports?: string;
@@ -61 +61 @@ program
-			}).then((data) => data as unknown as CustomMetadata);
+			});
@@ -80 +80 @@ program
-					json.thelounge.supports &&
+					json.thelounge?.supports &&
```

It made sense to me to do these changes together because the `CustomMetadata` type and the `readFile` variable are tightly linked to one another, even though that's not explicit in the code.  

I can revert to the minimal patch if you want, though.

marmitar: Reverted to the minimal patch.

marmitar:
I'm closing this in favor of #5042. I'll leave the branch up if any specific commit may be useful later on.

PR #5002, opened 2025-11-04
Modernize: Node 22

marmitar:
> [!IMPORTANT]
> Implemented on top of #5001.

Bumps the minimum supported version to Node 22 (current LTS) and updates the dependencies that become available with that version, while still keeping source changes minimal. I skipped Node 20, since it only has about six months left before End-of-Life in April 2026, so the project can focus on versions 22 and 24 instead.

Most "production" dependencies are updated to their latest versions in this PR. The only exceptions are:

- `got` v14, which [doesn't work with `"module": "commonjs"` in TypeScript](https://github.com/sindresorhus/got/releases/tag/v13.0.0).
- `ldapjs` v3, which has quite a few changes from v2 and [has been decommissioned](https://github.com/ldapjs/node-ldapjs), so an alternative might be preferable.

As with #5001, the PR is split into small commits for easier review.

Tested manually with Node 22. I'm also daily-driving this branch on Node 25 in my own setup for more than a week now.


ahyattdev:
@marmitar  `ldapjs` is being removed by #5008, where `passport-ldapauth` is used instead. Unfortunately `passport-ldapauth` also uses `ldapjs` v2, but least it wouldn't be a direct dependency.

marmitar:
I'm closing this in favor of #5041. I'll leave the branch up if any specific commit may be useful later on.

PR #5004, opened 2025-11-06
Show STATUSMSG indication for actions as well

supertassu:
It is possible to send actions (/me et al.) to a STATUSMSG target (+#channel etc), even though thelounge does not currently support sending that. Show the statusmsg indicator on such messages.

### DIFF
--- a/client/components/StatusmsgMarker.vue
+++ b/client/components/StatusmsgMarker.vue
@@ -1,0 +1,19 @@
+<template>
+	<span
+		v-if="group"
+		:aria-label="`This message was only shown to users with ${group} mode`"
+		class="msg-statusmsg tooltipped tooltipped-e"
+		><span>{{ group }}</span></span
+	>
+</template>
+
+<script lang="ts">
+import {defineComponent} from "vue";
+
+export default defineComponent({
+	name: "StatusmsgMarker",
+	props: {
+		group: {type: String, required: false, default: undefined},
+	},
+});
+</script>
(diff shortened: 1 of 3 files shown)


brunnre8 (changes requested):
Thanks, can you reduce the dependency of the thing? There's no reason that the new component takes a message

client/components/StatusmsgMarker.vue
```diff
@@ -0,0 +1,21 @@
...
+
+<script lang="ts">
+import {defineComponent, PropType} from "vue";
+
+import type {ClientMessage} from "../js/types";
+
+export default defineComponent({
+	name: "StatusmsgMarker",
+	props: {
+		message: {type: Object as PropType<ClientMessage>, required: true},
```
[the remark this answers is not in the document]
supertassu: good point, done!

client/components/StatusmsgMarker.vue
```diff
@@ -0,0 +1,19 @@
...
+	>
+</template>
+
+<script lang="ts">
+import {defineComponent} from "vue";
+
+export default defineComponent({
+	name: "StatusmsgMarker",
+	props: {
+		group: {type: String, required: true},
```
MaxLeiter:
```suggestion
		group: {type: String, required: false, default: undefined},
```

I believe this is More Correct, as `message.statusmsgGroup` can be `undefined` when its not a STATUSMSG.

supertassu: Oh, indeed. Fixed, and fixed the typing of the `Msg` where I originally got this from.

MaxLeiter (approved):

PR #5008, opened 2025-11-17
Use passportjs for auth, enabling OpenID Connect support

ahyattdev:
This builds on previous attempts, using passportjs where possible. It now authenticates all WebSocket sessions before the socket is even opened. The uploader API is also now requires the user to be authenticated.

Session tokens are stored on disk in a new sqlite database. The creation an application-wide database accessed through Sequelize enable the following future improvements:
- Easily add Postgres support
- Migrate user configuration into a single table (rather than files on disk).
- Store messages in one database, rather than per-user databases.

Another change is that the new session secret and VAPID secret can now be loaded from either an environmental variable or a custom path on disk. This enables secure scenarios where secrets as passed to TheLounge via in-memory secrets mounted on the filesystem (such as is common in Kubernetes).

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "got": "11.8.6",
     "irc-framework": "4.14.0",
     "is-utf8": "0.2.1",
-    "ldapjs": "2.3.3",
     "linkify-it": "3.0.3",
     "lodash": "4.17.21",
     "mime-types": "2.1.35",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -56,26 +56,35 @@
   ],
   "dependencies": {
     "@fastify/busboy": "1.0.0",
+    "@sequelize/core": "7.0.0-alpha.47",
+    "@sequelize/sqlite3": "7.0.0-alpha.47",
```
marmitar: Another SQLite dependency? Maybe it's possible to replace `sqlite3` with this one.

PR #5009, opened 2025-11-21
Add minimal self-registration feature

### LINKED ISSUE #914: Registration UI

micr0-dev:
closes #914

adds basic self-registration support in private mode. opt-in via config, disabled by default.

## what it does
- new signup form (username + password + confirm)
- creates user account immediately 
- config flag `selfRegister: false` to enable/disable

## what it doesn't do (intentionally minimal)
- no email verification
- no admin approval workflow
- no built-in rate limiting (handle via reverse proxy)
- no captcha

## why minimal
keeping scope tight to get something merged. discussed approach in #914 avoiding the scope creep that killed #641. security concerns like rate limiting should be handled at the infrastructure level (nginx/cloudflare) similar to how public mode works.

## can be enhanced later with
- email verification system
- admin approval workflow  
- password strength requirements
- rate limiting

## testing
tested locally with:
- successful registration
- duplicate username handling
- password mismatch validation
- config toggle working
- user can sign in after registering

## screenshots
image
image

image
image

https://github.com/user-attachments/assets/9e49a196-2f8f-4671-93cb-ca0279322eec

ready for review. happy to iterate on feedback but trying to keep scope focused.

### DIFF
--- a/client/components/Windows/SignIn.vue
+++ b/client/components/Windows/SignIn.vue
@@ -49,6 +49,11 @@
 			<div v-if="errorShown" class="error">Authentication failed.</div>
 
 			<button :disabled="inFlight" type="submit" class="btn">Sign in</button>
+
+			<p v-if="selfRegisterEnabled" class="sign-up-link">
+				Don't have an account?
+				<router-link to="/sign-up">Sign up</router-link>
+			</p>
 		</form>
 	</div>
 </template>
@@ -57,7 +62,8 @@
 import storage from "../../js/localStorage";
 import socket from "../../js/socket";
 import RevealPassword from "../RevealPassword.vue";
-import {defineComponent, onBeforeUnmount, onMounted, ref} from "vue";
+import {computed, defineComponent, onBeforeUnmount, onMounted, ref} from "vue";
+import {useStore} from "../../js/store";
 
 export default defineComponent({
 	name: "SignIn",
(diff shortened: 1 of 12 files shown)


caiot5:
Woot, that's a hot one... Can't wait to have it merged

client/components/Windows/SignIn.vue
```diff
@@ -57,20 +62,24 @@
...
 	},
 	setup() {
+		const store = useStore();
 		const inFlight = ref(false);
 		const errorShown = ref(false);
 
 		const username = ref(storage.get("user") || "");
 		const password = ref("");
 
+		const selfRegisterEnabled = computed(() => store.state.selfRegister);
```
MaxLeiter, line 81: this won't change during runtime so I don't believe we need to make it reactive with computed

defaults/config.js
```diff
@@ -18,6 +18,17 @@ module.exports = {
 	// This value is set to `false` by default.
 	public: false,
 
+	// ### `selfRegister`
+	//
+	// When set to `true`, users can create their own accounts via a sign-up form
+	// on the login page. This only applies in private mode (when `public` is
+	// `false`). The server admin is responsible for rate limiting via reverse
+	// proxy (nginx, Apache, Cloudflare, fail2ban, etc.) since The Lounge does
+	// not implement built-in protection against registration abuse.
```
MaxLeiter, line 27: we may want to include a line about no built-in password recovery

MaxLeiter (changes requested):

PR #5019, opened 2026-01-16
Add smart encoding detection for legacy IRC clients

### LINKED ISSUE #3746: Detect wrong encoding
### Feature Description

Some users may not write with Unicode encoding. We can detect it because it sometimes does not match any standard character but known characters in another encoding. (I'm pretty sure some libraries already do the work)

What do you think about implementing it inside thelounge?

 :+1: to agree :-1: to disagree, please vote as team leader :) .

rollecode:
Many older IRC clients (like mIRC) use ISO-8859-1 or ISO-8859-15 encoding instead of UTF-8, causing messages with non-ASCII characters (like Finnish ä, ö, å) to display as garbled text (e.g., `��` instead of `ää`).

This PR adds automatic encoding detection that properly handles both UTF-8 and legacy ISO-8859-1/15 encoded messages.

## How it works

1. **Receive as latin1** - Raw bytes from the IRC server are preserved (latin1 maps bytes 0-255 directly to Unicode codepoints)
2. **Smart detection** - For each incoming message, try to decode the bytes as UTF-8. If valid, use UTF-8; otherwise keep as ISO-8859-1
3. **Send as UTF-8** - Outgoing messages are always sent as UTF-8 for compatibility with modern clients

## Changes

- New `encoding.ts` module with `decodeSmartEncoding()` function
- Modified `network.ts` to use latin1 encoding for receiving and patch the transport to send as UTF-8
- Applied encoding detection to all text-containing IRC events:
  - Messages (PRIVMSG, NOTICE, ACTION, WALLOPS)
  - Topics
  - Quit/Part/Kick messages
  - Away messages
  - CTCP messages
  - MOTD
  - WHOIS realname and away fields
  - JOIN gecos field

## Test plan

- [x] UTF-8 client sends message with ääöö → displays correctly
- [x] ISO-8859-1 client (old mIRC) sends message with ääöö → displays correctly
- [x] The Lounge user sends ääöö → other UTF-8 clients see it correctly
- [x] The Lounge user sends ääöö → other ISO-8859-1 clients see it correctly
- [x] All existing tests pass (246 + 9 new encoding tests)

Fixes #3746

---

> **Note:** Part of this code was developed with AI assistance. Additional review from maintainers with IRC protocol expertise would be appreciated.

### DIFF
--- a/server/models/network.ts
+++ b/server/models/network.ts
@@ -302,6 +302,12 @@ class Network {
 			// it will keep trying for well over an hour (plus the timeouts)
 			auto_reconnect_max_retries: 30,
 
+			// Use latin1 encoding to preserve raw bytes from the IRC server.
+			// This allows smart encoding detection in the message handler to
+			// properly handle both UTF-8 and ISO-8859-1/15 encoded messages.
+			// Many older IRC clients (like mIRC) use ISO-8859-1/15 instead of UTF-8.
+			encoding: "latin1",
+
 			// TODO: this type should be set after setIrcFrameworkOptions
 		}) as NetworkWithIrcFramework["irc"];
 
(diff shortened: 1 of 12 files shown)


rollecode:
Tested with users on our channel:

image

brunnre8:
This isn't the right approach in my view. The world has moved on, utf8 is the standard and I don't think the complexity is worth it.

Any reasonably modern IRC client can emit utf8, those that can't aren't worth catering for.

rollecode:
> This isn't the right approach in my view. The world has moved on, utf8 is the standard and I don't think the complexity is worth it.
> 
> Any reasonably modern IRC client can emit utf8, those that can't aren't worth catering for.

I understand your point, and that's fair. However, IRC is one of the oldest messaging platforms, and we still have users running clients that are over 20 years old and don't have that charset support. I'm just tired of telling everyone to update their clients instead of fixing my open-source client. Reminds me of this:

image

Anyway, I can keep using my fork. Thanks for taking the time to review this PR 👍

PR #5023, opened 2026-01-31
fix: call stack size exceeded

### LINKED ISSUE #5022: Failed to load messages: Maximum call stack size exceeded
- _Node version:_  v25.3.0
- _Browser version:_ Mozilla LibreWolf 147.0.1-3
- _Device, operating system:_ Desktop, Arch Linux
- _The Lounge version:_ v4.4.3 + #5003 

---

Sometime ago I increased `maxHistory` to 131072 and noticed my history was not being loaded on startup. After taking a look at it today, I noticed these logs:

```log
2026-01-31 21:15:22 [ERROR] Failed to load messages for marmis: RangeError: Maximum call stack size exceeded
2026-01-31 21:15:23 [ERROR] Failed to load messages for marmis: RangeError: Maximum call stack size exceeded
2026-01-31 21:15:24 [ERROR] Failed to load messages for marmis: RangeError: Maximum call stack size exceeded
```

Going by the error message, this spread operator is probably the culprit:

https://github.com/thelounge/thelounge/blob/b2e3112806c4fb615b04c8981ac49a7308dedf58/server/models/chan.ts#L302

Likewise, there's a similar `unshift` in the frontend:

https://github.com/thelounge/thelounge/blob/b2e3112806c4fb615b04c8981ac49a7308dedf58/client/js/socket-events/more.ts#L25

I believe they could be replaced by a `.concat()` call, not only allow more messages, but to avoid copying the contents to the call stack and then to another messages array again.

marmitar:
The spread operator will place the arguments, which can reach the call stack limit if too many messages are being sent. This fix uses `.concat()` to avoid the spread operator.

Fixes #5022

Tested with `maxHistory: 131072`.

### DIFF
--- a/client/js/socket-events/more.ts
+++ b/client/js/socket-events/more.ts
@@ -22,7 +22,7 @@ socket.on("more", async (data) => {
 	);
 	channel.moreHistoryAvailable =
 		data.totalMessages > channel.messages.length + data.messages.length;
-	channel.messages.unshift(...data.messages);
+	channel.messages = data.messages.concat(channel.messages);
 
 	await nextTick();
 	channel.historyLoading = false;
--- a/server/models/chan.ts
+++ b/server/models/chan.ts
@@ -299,7 +299,7 @@ class Chan {
 					return;
 				}
 
-				this.messages.unshift(...messages);
+				this.messages = messages.concat(this.messages);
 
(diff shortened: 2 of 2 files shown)


MaxLeiter (approved):
thanks!

PR #5029, opened 2026-02-09
Fix mobile viewport

### LINKED ISSUE #4752: keyboard push up content by mistake
on android 10 using microsoft swiftkey keyboard

when focusing the typing input, the keyboard appears and pushes whole the page to up resulting in hiding the Top bar

df88a047-08e0-42b5-b06e-5e5384c33d7f
1e44758c-1fd8-4251-87c3-aedfdf5395a9

Jay2k1:
Make content resize again when the keyboard is slid out.

Fixes #4752 (Chromium) and also Firefox's different behaviour of not moving the content at all, making the keyboard overlay the bottom half of the app.

### DIFF
--- a/client/index.html.tpl
+++ b/client/index.html.tpl
@@ -3,7 +3,7 @@
 	<head>
 
 	<meta charset="utf-8">
-	<meta name="viewport" content="width=device-width, user-scalable=no">
+	<meta name="viewport" content="width=device-width, user-scalable=no, interactive-widget=resizes-content">
 
 	<link rel="preload" as="script" href="js/loading-error-handlers.js?v=<%- cacheBust %>">
 	<link rel="preload" as="script" href="js/bundle.vendor.js?v=<%- cacheBust %>">


NeatNit:
Important fix.

I brought up this issue on #thelounge a few days ago. Particularly in recent versions of Firefox, it made The Lounge nearly unusable, as the keyboard covers up the input box meaning I can't see what I'm typing. I had to use clever/clumsy workarounds such as split-screen apps just to chat.

It just so happens that I use the instance of The Lounge that @Jay2k1 hosts, so the fix in this PR is already available to me and it fixes the issue completely.

brunnre8:
looks like that won't help us on iOS... any ideas?

Jay2k1:
There's a bug that's now closed at https://bugs.webkit.org/show_bug.cgi?id=229876 with a link to a [test page](https://matan-hershberg-nuance.github.io/ios15-beta-visualviewport-bug/), but neither @NeatNit nor I have an iDevice to test with.
If Safari still doesn't support the meta tag way, but `virtualViewport.height` changes, we'd have to go the JS eventhandler way described [here](https://www.bram.us/2021/09/13/prevent-items-from-being-hidden-underneath-the-virtual-keyboard-by-means-of-the-virtualkeyboard-api/) I guess. However, that article also says that like Chrome, Safari pushes the whole app out of the viewport, so at least you'll still be able to see the text input, unlike what Firefox did:

image

I'm all in for the "if we fix it, let's do it properly and for all browsers", but if it turns out there's no good fix for Safari, I'd opt for at least fixing FF and Chromium-based browsers, the emphasis being on FF because on that, it's currently outright unusable.

NeatNit:
Had my sister test it on her iPhone via [this test page](https://viewport-resize-behavior.netlify.app/interactive-widget/resizes-content/), and this is the result:

IMG-20260211-WA0006

So the current form of this PR won't change anything in iOS, but a JavaScript approach should be possible.

sfan5:
Can confirm that this fix works on Firefox Mobile. Much appreciated.

NeatNit:
Is there any reason not to merge this? The change is trivial and fixes a major issue on Android. The iOS fix can come later, as it's more complex and must be implemented in JS, where bugs are much more likely.

It can also be argued that iOS should simply support this standard; it seems this is already implemented but disabled in the Safari configuration. https://github.com/mdn/browser-compat-data/issues/29011#issuecomment-3916534434

Either way there is no reason to delay the Android fix.

MaxLeiter (approved):

Issue #5025, opened 2026-02-08
RFC: AuthProvider Interface for Decoupled Auth


deevus:
Hi everyone. I've been looking into continuing #4676 and have a working version that addresses all of the feedback given there except for one thing:

>If a auth backend needs special logic, then pass control to the auth backend and let it return the data. Then strongly type it.
>
>The client should get an event and already know what it is, not some grab bag that contains maybe a hash, maybe openid connect, maybe $next_huge_auth_thing etc.
>
> — [@brunnre8, PR #4676](https://github.com/thelounge/thelounge/pull/4676#discussion_r1089711297)

After some consideration I have come up with the following proposal. The proposal assumes the existence of the not yet merged OpenID provider that I have been [working on](https://github.com/deevus/thelounge/tree/openid). I have used Claude to format the below, but the design decisions are my own. 

Since this is a fairly fundamental change I wanted to stop here before going any further. 

## Goal

Decouple server and auth from their implementations by introducing a strongly-typed `AuthProvider` interface. This replaces the current grab-bag pattern that uses `as any` and optional chaining for provider-specific methods.

---

### Problem

The current auth system has tight coupling between `server.ts` and specific auth providers:

1. **Provider-specific checks scattered in server.ts** - Direct `Config.values.openid.enable` and `Config.values.ldap.enable` checks appear in 6+ locations
2. **Grab-bag event payloads** - `auth:start` sends `openidEnabled`, `openidInit` fields that expose implementation details to the client
3. **Type safety abandoned** - The auth facade uses `as any` and dynamically copies properties from the active plugin

### Solution

Introduce a unified `AuthProvider` interface that all auth methods implement:

- **Factory pattern** - A `createAuthProvider()` function reads config and returns the appropriate provider
- **Strongly typed** - All methods have explicit parameter and return types
- **Capability-based** - Instead of `ldapEnabled`/`openidEnabled`, expose capabilities like `canChangePassword` and `authUrl`
- **Provider-agnostic server.ts** - No provider-specific code in the server; all logic delegated to the provider

### Key Design Decisions

1. **Parameter objects** - All methods receive named parameter objects:
   ```typescript
   authenticate(params: AuthenticateParams): Promise<AuthResult>
   ```

2. **Optional methods with sensible defaults** - `authStart?`, `logout?`, `disconnect?` are optional. Caller provides defaults:
   ```typescript
   auth.authStart?.({ socketId }) ?? {}
   ```

[...]

PR #5031, opened 2026-02-11
Decoupling auth: AuthProvider implementations for local and ldap

deevus:
RFC: #5025 

First of potentially a few PRs to refactor auth into a decoupled model. 

1. Adds `AuthProvider` and supporting types. 
2. Adds `LocalAuthProvider` class
3. Adds `SimpleLDAPAuthProvider`
4. Adds `AdvancedLDAPAuthProvider` (maybe just call this `LDAPAuthProvider`?)

I split the "simple" and "advanced" ldap implementations since it makes each much easier to understand. I'll be adding a factory method that given a `Config` instance can return an `AuthProvider`. Not sure if that should go in this PR or not. 

Also, I could pull the classes out into their own files. I opted to inline them alongside the legacy implementation for now. 

I've set this as a draft so that I can get early feedback. 🙏

### DIFF
--- a/server/plugins/auth/ldap.ts
+++ b/server/plugins/auth/ldap.ts
@@ -2,7 +2,7 @@ import ldap, {SearchOptions} from "ldapjs";
 import colors from "chalk";
 
 import log from "../../log";
-import Config from "../../config";
+import Config, {ConfigType} from "../../config";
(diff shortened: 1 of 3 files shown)


deevus:
Been away for a bit. Will be getting back to this soon.

PR #5038, opened 2026-03-30
feat: Add click-to-scroll mentions with full history browsing

rollecode:
## Click-to-scroll for mentions

Clicking a mention in the mentions popup now navigates to the channel and scrolls to the highlighted message. Works regardless of how old the mention is, if the message isn't in the currently loaded buffer, it fetches messages around it from the server.

### What's new

- **Click any mention** in the popup to jump directly to that message in the channel
- **Old mentions work too**: a new `messages:around` socket event fetches messages from server memory when the message isn't in the client's loaded buffer
- **Bidirectional history browsing**: after jumping to an old mention, you can scroll both up (existing "Show older messages") and down (new "Show newer messages") to browse the full history seamlessly
- **Hover state** on clickable mentions for visual feedback

### Changes

**Client:**
- `Mentions.vue`: click handler navigates to channel, scrolls to message, falls back to server fetch
- `MessageList.vue`: "Show newer messages" button + IntersectionObserver for auto-loading
- `messages_around.ts`: handles server response, replaces message buffer, resets unread marker
- `more_newer.ts`: handles newer message batches, restores live mode when caught up
- `chan.ts` / `types.d.ts`: added `moreNewerAvailable` flag

**Server:**
- `client.ts`: `messagesAround()` returns ~100 messages centered on target ID, `moreNewer()` returns 100 messages after a given ID
- `server.ts`: socket handlers for `messages:around` and `more:newer`
- `socket-events.d.ts`: type definitions for new events

### DIFF
--- a/client/components/Mentions.vue
+++ b/client/components/Mentions.vue
@@ -21,7 +21,10 @@
 				<p v-else>You have no recent mentions.</p>
 			</template>
 			<template v-for="message in resolvedMessages" v-else :key="message.msgId">
-				<div :class="['msg', message.type]">
+				<div
+					:class="['msg', message.type, {'mention-clickable': !!message.channel}]"
+					@click="goToMention(message)"
+				>
 					<div class="mentions-info">
 						<div>
 							<span class="from">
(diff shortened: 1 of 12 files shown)


spin-drift:
Not to bandwagon onto this (much-needed!) PR, but is there a way this could be adapted to also work with search? It suffers from the same usability issues as mentions.

Issue #4997, opened 2025-10-24
yarn.lock update needed for NodeJS 25 compatibility


ChrisTX:
theLounge indirectly (via `jws` and `jwa`) has `buffer-equal-constant-time` being pulled in and loaded, and due to [the removal of the deprecated SlowBuffer API](https://nodejs.org/api/deprecations.html#DEP0030) with NodeJS 25, this breaks.

jwa is locked at 2.0.0, but 2.0.1 fixed the compatibility with NodeJS 25 by not using that package on recent enough Node, c.f. [here](https://github.com/auth0/node-jwa/releases/tag/v2.0.1).

This affects both the current Git trunk and the 4.4.3 release package.

marmitar:
PR #4857 fixes it.

marmitar:
Master now works with Node 25 and 26 after 84963db2.

PR #5041, opened 2026-04-10
actions and package.json: remove Node 18 (EOL), bump sqlite3

MaxLeiter:
Node 18 is EOL (https://nodejs.org/en/about/previous-releases). We can use 22+ instead.

~Note that bumping `engines` is a breaking change, so we may need a semver major here?~
from irc: We use minor bumps for node bumps


test/fixtures/env.ts
```diff
@@ -1,5 +1,5 @@
 import * as path from "path";
-const home = path.join(__dirname, ".thelounge");
+const home = path.join(process.cwd(), "test", "fixtures", ".thelounge");
```
brunnre8, line 2: This doesn't do the same thing though no?
The current working directory of the test runner shouldn't have an impact on __dirname, it does now.

Does our test runner enforce the cwd?

MaxLeiter: `__dirname` doesn't work with ESM modules, and newer versions of nodes are more strict: https://stackoverflow.com/questions/46745014/alternative-for-dirname-in-node-js-when-using-es6-modules

package.json
marmitar, line 1: You might need to set `"type": "commonjs"` to avoid the ES modules in Node 22.

package.json
```diff
@@ -43,7 +43,7 @@
   ],
   "license": "MIT",
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=20.0.0"
```
marmitar: Node 20 is reaching EOL in [20 days](https://github.com/nodejs/release#release-schedule). Might be better to skip it.

package.json
marmitar, line 1: Also need to update mocha to a newer version, e.g. `11.7.5`.

test/fixtures/env.ts
marmitar: NPM specifically always runs scripts from the `package.json` directory. I would assume yarn does the same, so it shouldn't matter too much.

Still, might be better to keep `__dirname` for now, and replace it with `import.meta.dirname` during ESM migration.

package.json
brunnre8: will also allow us to ditch the unmaintained sqlite module, so absolutely

MaxLeiter: will do sqlite in a followup PR after this ships

.github/workflows/build.yml
```diff
@@ -12,19 +12,13 @@ jobs:
...
+            node_version: 22.x
           - os: ubuntu-latest
-            node_version: 18.x
-          # EOL: April 2026
-          - os: ubuntu-latest
-            node_version: 20.x
-          # EOL: April June 2024
-          - os: ubuntu-latest
-            node_version: 21.x
+            node_version: 22.x
```
marmitar, line 21: Maybe test on Node 24 and ~25 too~ (not working)?

```yaml
          # EOL: April 2028
          - os: ubuntu-latest
            node_version: 24.x
```

~Keep in mind that Node 25 and soon 26 have some important changes regarding Buffers (e.g. #4997).~

marmitar (approved):

.github/workflows/build.yml
marmitar: Never mind. Node 25 is still not working.

test/fixtures/env.ts
MaxLeiter: missed this comment - i was hitting build issues with `__dirname` but maybe the `type: commonjs` resolved that. we can revert if needed but its just tests and they still pass

PR #5043, opened 2026-04-10
ircv3: add bot mode support and [bot] userlist indicator

MaxLeiter:
This PR adds support for ircv3's bot mode: https://ircv3.net/specs/extensions/bot-mode

irc-framework already exposes it in message tags, but we don't do anything with it at the moment. This adds it to our `User` model and adds an indicator in the userlist.

image


server/models/user.ts
```diff
@@ -8,13 +8,15 @@ class User {
 	away!: string;
 	nick!: string;
 	lastMessage!: number;
+	bot!: boolean;
```
brunnre8: isBot maybe... bot sounds like an object which it is not

server/plugins/irc-events/message.ts
```diff
@@ -123,6 +124,24 @@ export default <IrcEventHandler>function (irc, network) {
...
+				if (user && !user.bot) {
+					user.bot = true;
+				}
+			}
+
+			client.emit("users", {
+				chan: chan.id,
+			});
+		}
+
```
brunnre8, line 127: This doesn't seem right.
We have 2 choices, flag a message (at which points it's a property of the message itself) or flag the user, but if we flag the user you'd get that in the nicklist from the RPL_WHOISBOT numeric no?

But the msg callback should definitely not randomly start emitting user events.
Also, that function is already bad, can we start adding functions to it please ;) ?

MaxLeiter: Much simpler.

PR #5047, opened 2026-04-10
 ircv3: support standard-replies 

MaxLeiter:
https://ircv3.net/specs/extensions/standard-replies

irc-framework already supports it (behind a flag but unpublished to npm) so its trivial for us to support and unblocks more ircv3 features


MaxLeiter:
@copilot fix conflicts

PR #5048, opened 2026-04-10
security: fix potential SSRF in link previews

MaxLeiter:
warning: claude generated

we should not embed private IPs. this is unlikely to be a super serious issue because people self-host their own lounge instances, but if you are on e.g. a corporate network or cloud VPC with IAM this could be a bypass if TL has privilege.

### DIFF
--- a/server/plugins/irc-events/link.ts
+++ b/server/plugins/irc-events/link.ts
@@ -52,20 +52,25 @@ function safeDnsLookup(
 
 	lookup(hostname, opts, (err: any, address: string, family: number) => {
 		if (err) {
-			return cb(err, address, family);
+			cb(err, address, family);
+
+			return;
 		}
 
 		if (!_testing.disableSSRFProtection && isPrivateIP(address)) {
-			return cb(
+			cb(
 				new Error(`Blocked request to private IP address: ${address}`),
 				address,
 				family
 			);
+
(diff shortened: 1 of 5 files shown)


MaxLeiter:
@copilot  fix the tests

server/plugins/irc-events/link.ts
```diff
@@ -506,6 +552,11 @@ function normalizeURL(link: string, baseLink?: string, disallowHttp = false) {
 			return undefined;
 		}
 
+		// Block IP literals pointing to private/reserved ranges (defense-in-depth for SSRF)
+		if (!_testing.disableSSRFProtection && isIP(url.hostname) && isPrivateIP(url.hostname)) {
```
supertassu: the regex-based approach is not very effective with user-supplied hostnames, since IP addresses can be written in a variety of formats (e.g. `http://10.20.10.181` can also be written as `http://169085621`, just as a single example) which would bypass the current filtering

brunnre8:
we also shouldn't go out of our way to do this.
I don't want to ship a bloody IP parser (that's probably buggy) just because someone has a NON AUTHENTICATED get request that's somewhat "sensitive"

If it was indeed sensitive, it shouldn't even be accessible by TL, let alone without auth

PR #5050, opened 2026-04-10
ircv3: +typing support

### LINKED ISSUE #2419: Add typing indicator (BitlBee integration)

MaxLeiter:
https://ircv3.net/specs/client-tags/typing

https://libera.chat/news/new-and-upcoming-features-3#client-tags

Very open to style feedback: I took some tips from discord:
- bump padding on the bottom to the chat to fit it (20px, not very noticeable)
- when scrolled up, a gradient is applied
- I dont love adding the padding but idk a better solution and IMO its not noticeable

https://github.com/user-attachments/assets/a911994e-1564-45bf-8475-37742ad8b4f5

image
image

Closes https://github.com/thelounge/thelounge/issues/2419
Supersedes/closes #4611


Jay2k1:
While testing this PR, I've noticed that the newly added bottom padding of `#chat .messages` is not applied when on mobile. This is caused by a media query in the style.css:

image

client/js/settings.ts
```diff
@@ -60,6 +60,9 @@ const defaultConfig = {
 		default: "",
 		sync: "always",
 	},
+	typing: {
+		default: "on",
```
MaxLeiter, line 64: I could be persuaded to default to `off`

MaxLeiter: maybe a `defaultTypingSetting` option on the server? I hate to add another option there, but I can imagine teams/events wanting it on for everyone.

brunnre8: this should really default to off

PR #5051, opened 2026-04-10
chore: refactor Mentions, add isIgnoredUser util

MaxLeiter:
- Mentions were doing a bunch of mutations; not just uses derived state
- no need for a separate `ClientMention` 
- added `extractInputHistory` for shared logic
- tests made by claude

### DIFF
--- a/client/components/Mentions.vue
+++ b/client/components/Mentions.vue
@@ -154,7 +154,13 @@ import dayjs from "dayjs";
 import relativeTime from "dayjs/plugin/relativeTime";
 import {computed, watch, defineComponent, ref, onMounted, onUnmounted} from "vue";
 import {useStore} from "../js/store";
-import type {ResolvedMention} from "../js/types";
+import type {SharedMention} from "../../shared/types/mention";
+import type {NetChan} from "../js/types";
+
+type MentionWithContext = SharedMention & {
+	localetime: string;
+	channel: NetChan | null;
+};
 
 dayjs.extend(relativeTime);
 
@@ -191,7 +197,7 @@ export default defineComponent({
 			return dayjs(time).fromNow();
 		};
 
-		const dismissMention = (message: ResolvedMention) => {
+		const dismissMention = (message: MentionWithContext) => {
 			store.state.mentions.splice(
 				store.state.mentions.findIndex((m) => m.msgId === message.msgId),
 				1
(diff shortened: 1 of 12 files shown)


client/components/Mentions.vue
```diff
@@ -169,14 +169,15 @@ export default defineComponent({
...
-
-			return messages.filter((message) => !message.channel?.channel.muted);
+			return store.state.mentions
+				.slice()
+				.reverse()
+				.map((message) => ({
+					...message,
+					localetime: localetime(message.time),
+					channel: store.getters.findChannel(message.chanId),
+				}))
```
brunnre8, line 185: this doesn't really improve the readability? It's subjective I guess but the change seems a bit funny

client/js/helpers/inputHistory.ts
```diff
@@ -0,0 +1,9 @@
+import {MessageType, SharedMsg} from "../../../shared/types/msg";
+
+export function extractInputHistory(messages: SharedMsg[], limit: number): string[] {
+	return messages
+		.filter((m) => m.self && m.text && m.type === MessageType.MESSAGE)
+		.map((m) => m.text!)
```
brunnre8: keep the comment around as to why this is done

client/js/types.d.ts
```diff
@@ -50,8 +50,10 @@ type NetChan = {
 	network: ClientNetwork;
 };
 
-type ClientMention = SharedMention & {
-	localetime: string; // TODO: this needs to go the way of the dodo, nothing but a single component uses it
+type ClientMention = SharedMention;
+
+type ResolvedMention = SharedMention & {
```
brunnre8: Resolved what?
Yes, if you read the code you understand that this was resolving the date to localetime, but when you see it in code this makes no real sense, least not to me

brunnre8 (changes requested):
Bunch of comments. Generally looks good though, thanks 😃 

Can you please not merge unrelated changes in a single PR though? It's hard to review that way, same for the commits, would be nice to clean them up pre-review so that I can walk through them, rather than just seeing the full diff at once

client/components/Mentions.vue
MaxLeiter: you shouldn't mutate the store in-place like that.

client/js/types.d.ts
MaxLeiter: changed it to `DisplayMention` which should be more clear.

MaxLeiter:
image

PR #5055, opened 2026-04-10
Remove sqlite3 dependency; use built-in node:sqlite

### LINKED ISSUE #5033: Build failing on Node.js v25.3.0: napi_get_uv_event_loop undeclared identifier in sqlite3

MaxLeiter:
https://www.npmjs.com/package/sqlite3 is deprecated and Node 22 (#5041) lets us use the builtin `node:sqlite` package (although it is experimental)

most changes are a result of the native module being synchronous

relies on #5041 (now merged)

Closes https://github.com/thelounge/thelounge/issues/5033


server/plugins/messageStorage/sqlite.ts
```diff
@@ -126,12 +113,13 @@ class SqliteMessageStorage implements SearchableMessageStorage {
 		this.initDone = new Deferred();
 	}
 
-	async _enable(connection_string: string) {
-		this.database = new sqlite3.Database(connection_string);
+	_enable(connection_string: string) {
+		this.database = new DatabaseSync(connection_string, {
+			enableForeignKeyConstraints: true,
```
brunnre8: this is the default

brunnre8 (changes requested):
Given that node doesn't cache prepared statements, we should use the tagstore I think https://nodejs.org/api/sqlite.html#class-sqltagstore

Given that you anyway flipped everything to not being a promise and the fact that finally sqlite stops being optional, we should also get rid of the enable hack that I wrote to cope with the issues

MaxLeiter:
@brunnre8 Node 22 doesn't support the Tag Store AFAICT. Seems like it first appeared (in the docs at least) with node 24.

brunnre8:
sad... oh well

server/storageCleaner.ts
```diff
@@ -102,7 +101,6 @@ export class StorageCleaner {
 			}
 		}
 
-		// need to recheck here as the field may have changed since the await
 		if (this.isStopped) {
 			return;
 		}
```
brunnre8, line 107: I mean... if you delete the comment you should delete the conditional 😉

brunnre8 (approved):
Other than the nit this looks good to me, thanks a lot 🙂

brunnre8:
mind cleaning up the commits, so that we have a sane log?

MaxLeiter:
@brunnre8 thats why I enabled squash merges; it will squash into one commit on merge.

PR #5059, opened 2026-04-10
fix: preserve network name in sidebar after editing in public mode (fixes #4922)

### LINKED ISSUE #4922: While on public mode, editing a network makes it name disappear from the sidebar
- _Node version:_ v20.18.0
- _Browser version:_ Google Chrome Version 129.0.6668.70 (Official Build) (64-bit)
- _Device, operating system:_ Desktop, Ubuntu 24.04.1 LTS
- _The Lounge version:_ Latest stable (4.4.3), installed via yarn

---
While on public mode, if we: `Right click network name > Edit this network > Make some edits > Click "Save Network"`, the network name disappears from the sidebar and only a page refresh makes it reappear.

Screenshot:
TL_Network_Disappear

MaxLeiter:
We should only emit `network:name` to clients after we validate the name, otherwise their state can get wonky 

Fixes #4922

### DIFF
--- a/client/components/Windows/NetworkEdit.vue
+++ b/client/components/Windows/NetworkEdit.vue
@@ -40,12 +40,7 @@ export default defineComponent({
 			const network = store.getters.findNetwork(data.uuid);
 
 			if (network) {
-				// In public mode with lockNetwork, the name field is not in the form,
-				// so data.name will be undefined. Preserve the existing name in that case.
-				if (data.name) {
-					network.name = network.channels[0].name = data.name;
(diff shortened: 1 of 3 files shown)


brunnre8 (commented):
That's one way of doing it.
But do you know *why* the name isn't in the form in public mode?
I don't understand why this is done, the server seems to know the name just fine so where's the problem in sending that to the client?

MaxLeiter:
@brunnre8 fixed and updated PR desc

PR #5060, opened 2026-04-10
fix: send credentials when using PWA installation behind Cloudflare Access

### LINKED ISSUE #5032: Can't install as PWA behind Cloudflare Access
Using exposed/public cloudflare application PWA installation works fine. Maybe related: https://github.com/danny-avila/LibreChat/discussions/5154

MaxLeiter:
Add crossorigin="use-credentials" to the manifest link tag so the browser includes authentication cookies when fetching the web manifest. Without this, Cloudflare Access redirect to the credentialless manifest request to a login page, which blocks PWA installation.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/crossorigin#use-credentials

Fixes #5032

### DIFF
--- a/client/index.html.tpl
+++ b/client/index.html.tpl
@@ -24,7 +24,7 @@
 	<!-- Safari pinned tab icon -->
 	<link rel="mask-icon" href="img/icon-black-transparent-bg.svg" color="#415364">
 
-	<link rel="manifest" href="thelounge.webmanifest">
+	<link rel="manifest" href="thelounge.webmanifest" crossorigin="use-credentials">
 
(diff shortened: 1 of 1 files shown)


brunnre8 (dismissed):
This doesn't seem right.
We don't do a cross origin request.
If some proxy messes with the fetching, it's up to that proxy to unbork what they broke.

We should not be degrading the security properties just because some tool is doing funny things.
It's also not the only request to the TL host itself, it's very random to do that here for that single request

xPaw:
@brunnre8 I suspect this is also an issue if you use http basic authentication, so the fix looks correct to me. https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest#deploying_a_manifest

I don't see how this degrades security, all the other assets should already be fetched with credentials.

ryukinix:
Thank you @MaxLeiter @brunnre8

PR #5061, opened 2026-04-11
replace web-push with web-push-neo

### LINKED ISSUE #4997: yarn.lock update needed for NodeJS 25 compatibility

MaxLeiter:
https://github.com/web-push-libs/web-push relies on outdated libraries that use `SlowBuffer` and thus are incompatible with Node 25

`web-push-neo` (https://github.com/ryoppippi/web-push-neo) is new and not battle tested, but is single-dependency and uses modern web standards. It also unblocks node 25.

closes https://github.com/thelounge/thelounge/issues/4997

one todo is that `web-push-neo` is ESM only so we need to dynamic `import` it -- this can be fixed once we switch to ESM


brunnre8:
hm I'm generally reluctant to pull in some random dependency if you don't know how good that is / will be maintained etc.

One would naively assume that given the prevalence of webpush there's some standard lib that's maintained by a few people  / orgs?

MaxLeiter:
@brunnre8 while I tend to agree, I couldn’t find an alternative. The project seems feature complete and I’m familiar with the author as a developer / user of their work’s product. But also fine to hold off and wait for a better alternative. 

cc @ryoppippi, in case you have anything to add (sorry for the ping!)

ryoppippi:
omg, thank you for using it. i wanted to use it in my internal project. i think it works fine, but i cannot guarantee. if you find something let me know anytime!

xPaw:
Web push is essentially just sending some data in a specific format to a url, so the dependency here is not that big.

PR #5062, opened 2026-04-11
 ircv3: add +reply support 

MaxLeiter:
Relies on https://github.com/kiwiirc/irc-framework/pull/411

https://ircv3.net/specs/client-tags/reply.html

Design feedback very welcome

Screenshot 2026-04-11 at 21-00-22 #thelounge — The Lounge

1. we lack a 'thread' view (irccloud, slack, etc). can be a followup if we do want it
2. need to test scrolling to messages outside of the current buffer
3. due to lack of client support, i opted to auto prefill the input with \`<targetNick>: \` so people can follow along even without reply support
4. CSS can definitely be improved (once a design is settled on)


Jay2k1:
so while testing this PR, I've noticed that I can reply and then that's rendered for me, but if someone else replies to my message, i don't see that it is a reply.

image

PR #5063, opened 2026-04-11
fix: hide Account settings tab on public instance, sometimes hide General

MaxLeiter:
on public TL instances like the demo, this is what the account and general tabs look like in Firefox:
 
| General | Account |
|--------|--------|
| image | image | 

1. we should just always hide account if mode is public, its useless
2. if file uploads are off, there's nothing to show in General except the native app install, which i figure is fine/not really necessary for public instances (and isn't supported by Firefox, hence the blankness)

### DIFF
--- a/client/components/Settings/Navigation.vue
+++ b/client/components/Settings/Navigation.vue
@@ -3,10 +3,10 @@
 	<aside class="settings-menu">
 		<h2>Settings</h2>
 		<ul role="navigation" aria-label="Settings tabs">
-			<SettingTabItem name="General" class-name="general" to="" />
+			<SettingTabItem v-if="showGeneral" name="General" class-name="general" to="" />
 			<SettingTabItem name="Appearance" class-name="appearance" to="appearance" />
 			<SettingTabItem name="Notifications" class-name="notifications" to="notifications" />
-			<SettingTabItem name="Account" class-name="account" to="account" />
+			<SettingTabItem v-if="!isPublic" name="Account" class-name="account" to="account" />
 		</ul>
 	</aside>
 </template>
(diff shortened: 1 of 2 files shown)


client/components/Settings/Navigation.vue
```diff
@@ -93,11 +93,18 @@
...
 
 export default defineComponent({
 	name: "SettingsTabs",
 	components: {
 		SettingTabItem,
 	},
+	setup() {
+		const store = useStore();
+		const isPublic = store.state.serverConfiguration?.public;
+		const showGeneral = !isPublic || store.state.serverConfiguration?.fileUpload;
```
brunnre8, line 106: we have more in general than just the file upload, this doesn't work I believe

client/js/router.ts
```diff
@@ -48,6 +48,16 @@ const router = createRouter({
...
+					beforeEnter(_to, _from, next) {
+						const config = store.state.serverConfiguration;
+
+						if (config?.public && !config?.fileUpload) {
+							next({name: "Appearance"});
+							return;
+						}
+
+						next();
+					},
```
brunnre8: Why is this needed? We shouldn't even be showing the button if it doesn't apply

brunnre8 (changes requested):

client/components/Settings/Navigation.vue
MaxLeiter: but the other two sections are only if `!isPublic` and see description for how I feel about PWA installer

client/js/router.ts
MaxLeiter: in case people have it in their e.g. browser history  + so `/general` works

brunnre8 (approved):
Thank you

PR #5064, opened 2026-04-11
remove webpack, babel, mocha; switch to vite/vitest

MaxLeiter:
```
Webpack vs Vite Build Comparison

  ┌────────────┬──────────────────────┬────────────────────┬───────────────┐
  │   Metric   │      Webpack 5       │ Vite 8 (Rolldown)  │    Change     │
  ├────────────┼──────────────────────┼────────────────────┼───────────────┤
  │ Build time │ 1,961ms (2.96s wall) │ 612ms (0.82s wall) │ 3.2x faster   │
  ├────────────┼──────────────────────┼────────────────────┼───────────────┤
  │ CPU time   │ 11.0s user           │ 1.3s user          │ 8.5x less CPU │
  └────────────┴──────────────────────┴────────────────────┴───────────────┘

  Bundle sizes (gzipped)

  ┌────────────────┬─────────┬────────┬────────┐
  │     Asset      │ Webpack │  Vite  │ Change │
  ├────────────────┼─────────┼────────┼────────┤
  │ App JS         │ 152.6K  │ 104.9K │ -31%   │
  ├────────────────┼─────────┼────────┼────────┤
  │ Vendor JS      │ 251.1K  │ 81.6K  │ -68%   │
  ├────────────────┼─────────┼────────┼────────┤
  │ Runtime JS     │ —       │ 0.5K   │ new    │
  ├────────────────┼─────────┼────────┼────────┤
  │ Total JS       │ 403.7K  │ 187.0K │ -54%   │
  ├────────────────┼─────────┼────────┼────────┤
  │ CSS            │ 12.5K   │ 11.9K  │ -5%    │
  ├────────────────┼─────────┼────────┼────────┤
  │ Total transfer │ 416.2K  │ 198.9K │ -52%   │
  └────────────────┴─────────┴────────┴────────┘

  Raw (uncompressed)

  ┌───────────────┬─────────┬──────┬────────┐
  │     Asset     │ Webpack │ Vite │ Change │
  ├───────────────┼─────────┼──────┼────────┤
  │ App JS        │ 1.3M    │ 304K │ -77%   │
  ├───────────────┼─────────┼──────┼────────┤
  │ Vendor JS     │ 1.1M    │ 236K │ -79%   │
  ├───────────────┼─────────┼──────┼────────┤
  │ CSS           │ 64K     │ 56K  │ -13%   │
  ├───────────────┼─────────┼──────┼────────┤
  │ Total public/ │ 5.1M    │ 3.4M │ -33%   │
  └───────────────┴─────────┴──────┴────────┘
```


test/public/testclient.js
```diff
@@ -0,0 +1,72 @@
+/*
```
brunnre8: this should probably not be in the repo anymore I guess?

MaxLeiter: nice

scripts/copy-fonts.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+
+const src = path.join("node_modules", "@fortawesome", "fontawesome-free", "webfonts");
+const dest = path.join("client", "public", "fonts");
```
xPaw: Does vite not support importing files from node modules without having to copy them manually and have them properly managed as assets?

client/index.html.tpl
```diff
@@ -6,10 +6,10 @@
 	<meta name="viewport" content="width=device-width, user-scalable=no, interactive-widget=resizes-content">
 
 	<link rel="preload" as="script" href="js/loading-error-handlers.js?v=<%- cacheBust %>">
-	<link rel="preload" as="script" href="js/bundle.vendor.js?v=<%- cacheBust %>">
-	<link rel="preload" as="script" href="js/bundle.js?v=<%- cacheBust %>">
 
-	<link rel="stylesheet" href="css/style.css?v=<%- cacheBust %>">
+	<% _.forEach(cssFiles, function(css) { %>
```
xPaw: I wonder, do we actually need a .tpl that is rendered by the server at runtime? Is it only for the server theme and packages? Because I think the actual built assets could be managed by vite as normal, and only the packages need to be injected.

Which would also mean cacheBust can be removed as the files should be hashed correctly.

scripts/copy-fonts.js
MaxLeiter: yep, `url()` seems to work fine 👍 good call.

xPaw (approved):

PR #5066, opened 2026-04-11
workflows: don't run on push and PR for every branch

MaxLeiter:
image

Right now we run identical workflows twice for every PR. instead, we should run on `push` on main and on `pull_request` for everything else.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,11 @@ name: Build
 permissions:
   contents: read
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   build:


brunnre8:
does this then run if someone does follow up commit to a PR after the reviews?

MaxLeiter:
@brunnre8: yes:

> More than one activity type triggers this event. For information about each activity type, see [Webhook events and payloads](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request). By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. To trigger workflows by different activity types, use the types keyword. For more information, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onevent_nametypes).

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

should've included that in the PR description, my bad

PR #5085, opened 2026-04-18
tests/coverage: ignore test/client and test/public, fix running coverage

MaxLeiter:
we need to ignore webpack's build output / bundles in `public/`

fixes `yarn coverage`

Issue introduced in https://github.com/thelounge/thelounge/commit/a12ddc75


astorije (approved):
😉

PR #5087, opened 2026-04-18
fix order of messages with identical timestamps

Jay2k1:
When sending a message that is too long, thelounge will split it up into multiple PRIVMSGs. These will be saved to the sqlite database with identical timestamps. Thelounge retrieves messages from the DB using `ORDER BY time DESC`, which under certain circumstances causes these messages to be loaded and displayed in the wrong order. This PR addresses that by also ordering by id (only affects messages with identical timestamps).

Before:
image

After:
image

### DIFF
--- a/server/plugins/messageStorage/sqlite.ts
+++ b/server/plugins/messageStorage/sqlite.ts
@@ -381,7 +381,7 @@ class SqliteMessageStorage implements SearchableMessageStorage {
 
 		const rows = this.database
 			.prepare(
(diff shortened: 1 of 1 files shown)


MaxLeiter:
We have the same bug in `search` if you can fix that as well.

Jay2k1:
Done. I could've searched the whole file for `time DESC` at the start 🤦‍♂️

MaxLeiter (approved):

PR #5089, opened 2026-04-19
router: add `beforeEnter` for `/settings` to avoid rendering empty pages

MaxLeiter:
https://github.com/thelounge/thelounge/pull/5063 hid the general tab if it had nothing to show, but we removed the route guard. turns out, we need it to avoid taking you to an empty settings page (reproducible on https://demo.thelounge.chat/) 

Screenshot 2026-04-18 at 5 16 21 PM

### DIFF
--- a/client/js/helpers/settingsTabs.ts
+++ b/client/js/helpers/settingsTabs.ts
@@ -1,0 +1,6 @@
+import {store} from "../store";
+
+export function showGeneralSettings() {
+	const config = store.state.serverConfiguration;
+	return !config?.public || !!config?.fileUpload;
+}
--- a/client/js/helpers/settingsTabs.ts
+++ b/client/js/helpers/settingsTabs.ts
@@ -1,6 +1,6 @@
 import {store} from "../store";
 
-export function showGeneralSettings() {
+export function shouldShowGeneralSettings() {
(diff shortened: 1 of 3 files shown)


client/js/helpers/settingsTabs.ts
```diff
@@ -0,0 +1,6 @@
+import {store} from "../store";
+
+export function showGeneralSettings() {
```
brunnre8: This probably shouldn't be called like a verb, it doesn't show the settings.
It asks if the settings should be shown

client/js/router.ts
```diff
@@ -48,6 +49,14 @@ const router = createRouter({
...
 					path: "",
 					component: GeneralSettings,
+					beforeEnter(to, from, next) {
+						if (!showGeneralSettings()) {
+							next({name: "Appearance"});
+							return;
+						}
+
+						next();
+					},
```
brunnre8, line 59: I still think this is the wrong place of doing that, tying the router to a particular behavior of the view component seems strange.

I also don't have an immediate other idea though, so do as you see fit.

MaxLeiter:
@brunnre8 pushed a rename for the function. not sure on a better solution for redirecting than the router, sadly.

PR #5098, opened 2026-05-12
uploader: change order mime type translation

brunnre8:
The upload maps some mime types onto more well-known ones. However it does so after the check whether it should set the inline disposition.
We need to do the translation first so that it gets matched against the mime type we actually send.

### DIFF
--- a/server/plugins/uploader.ts
+++ b/server/plugins/uploader.ts
@@ -11,16 +11,25 @@ import contentDisposition from "content-disposition";
 import type {Socket} from "socket.io";
 import {Request, Response} from "express";
 
-// Map of allowed mime types to their respecive default filenames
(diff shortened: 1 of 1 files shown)


xPaw:
Doesn't inlineContentDispositionTypes contain the pre-translated types though? Only video/quicktime seemed to be missing.

brunnre8:
You are right.
That code seems to be very confused...
Testing the inline disposition pre-translation seems wrong, so rather than playing the game, just change the lookup to be post map.

Better?

brunnre8:
also... TS is getting confused I think... there's a null check a few lines above. Any idea what the problem is?

MaxLeiter:
@brunnre8 I think you need to mark `mimeAliases` with `as const` so TS knows (well, believes) its not going to change at runtime

PR #5102, opened 2026-05-21
ident: normalize ipv4-mapped addresses when comparing

### LINKED ISSUE #5101: identd forwarding broken with 4.5.0 release
- _Node version:_ v24.15.0
- _Browser version:_ Brave Stable
- _Device, operating system:_ Linux
- _The Lounge version:_ 4.5.0

---

After todays update identd forwarding is broken to oidentd, same config works on 4.4.3.

MaxLeiter:
Reported by AlexC on IRC:

> after upgrade [to 4.5.0] identd forwarding is broken

4.5.0 included 0e48014d (`ident: only respond if the ip,port tuples match`). It correctly requires the addresses to match but the comparison compares the IPs as strings, but an IP can have multiple variations.

A default `bind` unset ident server will look something like `::ffff:127.0.0.1` while outgoing will be `127.0.0.1.` and those obviously do not match in a string equality check.

closes #5101

### DIFF
--- a/server/identification.ts
+++ b/server/identification.ts
@@ -5,6 +5,19 @@ import Helper from "./helper";
 import Config from "./config";
 import log from "./log";
 
+// Node reports an IPv4 peer accepted on a dual-stack (`::`) listener as an
+// IPv4-mapped IPv6 address (e.g. `::ffff:127.0.0.1`)
+// See "IPv4-Mapped IPv6 Address": https://www.rfc-editor.org/rfc/rfc4291.html#section-2.5.5.2
+function normalizeAddress(address: string | undefined): string | undefined {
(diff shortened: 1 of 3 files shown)


xPaw:
I feel like if you're going through the trouble of normalizing IPs, they should be compared as binary because addresses can be truncated with trailing `::` for example, but maybe that's not a problem with sockets.

AlexanderCurl:
Can confirm this patch is working as intended. Tested today, looks fine.

MaxLeiter:
@AlexanderCurl thanks!

@xPaw both come from the same node lib so they're the same format but in general you're right

xPaw (approved):

PR #5106, opened 2026-05-31
Show SASL authentication in public mode, remove redundant server pass…

TehPeGaSuS:
…word field

In public mode with `lockNetwork` enabled, the connect form showed an "I have a password" checkbox that exposed the IRC server password field. This is redundant — operators running a locked public instance are expected to set `defaults.password` in the server config if a server password is needed. Surfacing it in the UI is confusing and inconsistent with how private mode handles authentication.

At the same time, the SASL authentication section was hidden entirely in public mode behind a `v-else`, meaning users on a public locked instance had no way to authenticate with IRC services (e.g. NickServ via SASL PLAIN) at connect time.

This commit:
- Removes the "I have a password" / server password block from the public + lockNetwork branch
- Moves the Authentication section (SASL radio buttons + account/password fields) outside the public/private conditional so it renders in all modes
- Keeps SASL EXTERNAL restricted to private mode + TLS, as client certificate management is not meaningful for ephemeral public sessions
- Removes the now-unused `displayPasswordField` ref and its associated `watch` from the component script


brunnre8:
>This is redundant — operators running a locked public instance are expected to set defaults.password in the server config if a server password is needed

This is wrong... You can have a locked network but don't want to allow just anyone to connect that ahs access to the website. It's a different paradigm.

TehPeGaSuS:
> > This is redundant — operators running a locked public instance are expected to set defaults.password in the server config if a server password is needed
> 
> This is wrong... You can have a locked network but don't want to allow just anyone to connect that ahs access to the website. It's a different paradigm.

This should address it, I hope.

PR #5108, opened 2026-06-02
Attempt to clean up some loose `any` types

deltamualpha:
As briefly discussed on IRC, I decided to take a swing at cleaning up some usage of explicit `any` in the typescript. I found a few places where the types were relatively straightforward to either manually derive or that they just needed to be imported from a library's types file. Happy to break this into further smaller PRs if desired.

One big weak spot right now in the code related to `any` usage is in `catch` statements, which are kind of all over the map with how they treat the `Error` type. TS best practices appears to be to never assert in the catch definition that a thrown value is an Error (since you can literally `throw` any value in javascript, including `null` and `undefined`), and instead always do explicit `typeof` in the catch body. Standardizing that across the codebase feels like a bigger undertaking that should have more input from the existing contributor core.

### DIFF
--- a/client/js/helpers/ircmessageparser/merge.ts
+++ b/client/js/helpers/ircmessageparser/merge.ts
@@ -72,12 +72,12 @@ function merge(
 
 	// Distribute the style fragments within the text parts
 	return allParts.map((part: Part) => {
-		const fragmentedPart = part as PartWithFragments;
-		fragmentedPart.fragments = styleFragments
-			.filter((fragment) => anyIntersection(fragmentedPart, fragment))
-			.map((fragment) => assign(fragmentedPart, fragment));
-
-		return fragmentedPart;
+		return {
+			...part,
+			fragments: styleFragments
+				.filter((fragment) => anyIntersection(part, fragment))
+				.map((fragment) => assign(part, fragment)),
+		};
 	});
 }
 
--- a/client/js/helpers/ircmessageparser/merge.ts
+++ b/client/js/helpers/ircmessageparser/merge.ts
@@ -71,12 +71,13 @@ function merge(
 	const allParts: MergedParts = [...parts, ...filled].sort(sortParts); // Sort all parts identified based on their position in the original text
 
 	// Distribute the style fragments within the text parts
-	return allParts.map((part: any) => {
-		part.fragments = styleFragments
-			.filter((fragment) => anyIntersection(part, fragment))
-			.map((fragment) => assign(part, fragment));
+	return allParts.map((part: Part) => {
+		const fragmentedPart = part as PartWithFragments;
+		fragmentedPart.fragments = styleFragments
+			.filter((fragment) => anyIntersection(fragmentedPart, fragment))
+			.map((fragment) => assign(fragmentedPart, fragment));
 
-		return part as PartWithFragments;
+		return fragmentedPart;
 	});
 }
 
--- a/client/js/settings.ts
+++ b/client/js/settings.ts
@@ -162,7 +162,7 @@ export function createState() {
 	return state;
 }
 
-function normalizeConfig(obj: Record<string, typeof defaultConfig[keyof typeof defaultConfig]>) {
+function normalizeConfig(obj: typeof defaultConfig) {
 	const newConfig: Partial<typeof defaultConfig> = {};
 
 	for (const settingName in obj) {
--- a/client/js/settings.ts
+++ b/client/js/settings.ts
@@ -162,7 +162,7 @@ export function createState() {
 	return state;
 }
 
-function normalizeConfig(obj: any) {
+function normalizeConfig(obj: Record<string, typeof defaultConfig[keyof typeof defaultConfig]>) {
 	const newConfig: Partial<typeof defaultConfig> = {};
 
 	for (const settingName in obj) {
--- a/server/client.ts
+++ b/server/client.ts
@@ -460,7 +460,7 @@ class Client {
 
 	input(data: {target: number; text: string}) {
 		const client = this;
-		data.text.split("\n").forEach((line: string) => {
+		data.text.split("\n").forEach((line) => {
 			data.text = line;
 			client.inputLine(data);
 		});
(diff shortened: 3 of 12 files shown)


server/client.ts
```diff
@@ -338,7 +341,13 @@ class Client {
 		});
 	}
 
-	connectToNetwork(args: Record<string, any>, isStartup = false) {
+	connectToNetwork(
+		args:
+			| Record<keyof NetworkConfig, NetworkConfig[keyof NetworkConfig]>
+			| NetworkConfig
+			| {host: string; port: number; tls: boolean},
```
xPaw, line 350: whoever calls this should probably use NetworkConfig too. Ideally the args would only accept NetworkConfig as a type, without Record (why does it need that?)

marmitar: The `Record` is there to match this usage in `network:new`:

https://github.com/thelounge/thelounge/blob/0fe011235e305348a3083feab24a0a0b420d2ea2/server/server.ts#L480

The alternative is moving `NetworkConfig` (with more optional fields) to `shared/types` and declaring `ClientToServerEvents` with:

```ts
	"network:edit": (data: NetworkConfig) => void;
	"network:new": (data: NetworkConfig) => void;
```

That would also need an update on the client code.

server/server.ts
```diff
@@ -1056,7 +1056,7 @@ function performAuthentication(this: Socket, data: AuthPerformData) {
 		return;
 	}
 
-	Auth.initialize().then(() => {
+	void Auth.initialize().then(() => {
```
marmitar: This usage of `void` to mask floating promises can be really easy to miss. I think just adding an explict `// eslint-disable-next-line @typescript-eslint/no-floating-promises` is better for maintenance in the long run.

server/client.ts
```diff
@@ -447,15 +456,15 @@ class Client {
 		});
 	}
 
-	input(data) {
+	input(data: {target: number; text: string}) {
 		const client = this;
-		data.text.split("\n").forEach((line) => {
+		data.text.split("\n").forEach((line: string) => {
```
marmitar: Since `data.text` is known to be a string now, this `line: string` declaration is not really needed.

server/plugins/auth.ts
```diff
@@ -53,7 +67,7 @@ const toExport = {
 			log.error("None of the auth plugins is enabled");
 		}
 	},
-} as any;
+} as Module;
```
marmitar: Casts can hide bugs from time to time. It's better to just declare the type in the variable, when possible.

```ts
const toExport: Module = {
```

server/plugins/irc-events/cap.ts
```diff
@@ -20,11 +20,11 @@ export default <IrcEventHandler>function (irc, network) {
 		}
 
 		const isSecure = irc.connection.transport.socket.encrypted;
-		const values = {} as any;
+		const values = {} as Record<string, string>;
```
marmitar: No need for the cast anymore.

```suggestion
		const values: Record<string, string> = {};
```

client/js/settings.ts
```diff
@@ -159,7 +159,7 @@ export function createState() {
 	return state;
 }
 
-function normalizeConfig(obj: any) {
+function normalizeConfig(obj: Record<string, typeof defaultConfig[keyof typeof defaultConfig]>) {
```
marmitar: Isn't `typeof defaultConfig` enough here?

```suggestion
function normalizeConfig(obj: typeof defaultConfig) {
```

client/js/helpers/ircmessageparser/merge.ts
```diff
@@ -71,12 +71,13 @@ function merge(
...
-			.filter((fragment) => anyIntersection(part, fragment))
-			.map((fragment) => assign(part, fragment));
+	return allParts.map((part: Part) => {
+		const fragmentedPart = part as PartWithFragments;
+		fragmentedPart.fragments = styleFragments
+			.filter((fragment) => anyIntersection(fragmentedPart, fragment))
+			.map((fragment) => assign(fragmentedPart, fragment));
 
-		return part as PartWithFragments;
+		return fragmentedPart;
```
marmitar: More casts to avoid here:

```ts
	return allParts.map((part) => {
		return {
			...part,
			fragments: styleFragments
				.filter((fragment) => anyIntersection(part, fragment))
				.map((fragment) => assign(part, fragment)),
		};
	});
```

client/js/settings.ts
deltamualpha: I'll check. At the time it seemed to cause checking issues, but perhaps I was just holding it wrong.

server/plugins/auth.ts
deltamualpha: very good point

deltamualpha:
Thanks for the detailed feedback, as you probably guessed, I'm not all that experienced with TS, so pointing out some of the nits is helpful.

server/server.ts
deltamualpha: I can't tell if this is actually a _bug_ or just behavior that works because it's undefined what's supposed to happen and node just does the right thing. There's a couple of these in the codebase.

marmitar: It's not a bug. Linters are commonly used to flag code patterns that *may* lead to bugs at another code or time. In particular, [no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/) flags `Promise`s with a `.catch()` call, to avoid unexpected runtime errors like [`'unhandledRejection'`](https://nodejs.org/api/process.html#event-unhandledrejection). It would be better to handle it, but it's not completely necessary.

server/client.ts
deltamualpha: Alternatively, this also makes the compiler somewhat happier and also lets us remove the Record type from `client.ts`.

```ts
client.connectToNetwork(data as NetworkConfig);
```

deltamualpha: Do we want to leave the kinda messy `Record` or do an `as`? I feel like the messier type is both more accurate to what's coming into the function as well as an obvious thing that someone will feel compelled to clean up in the future, whereas `as` just hides the complexity.

deltamualpha: @marmitar @xPaw any further thoughts here? I think this is the last point before this can be merged and I can take a swing at some more types elsewhere in the code.

marmitar: On this change specifically, no, I don't have anything to add.

Just bear in mind that that this *might* take a bit longer to be merged. I'm not associated with The Lounge in any capacity, so it's up to the maintainers of this project to actually review it.

PR #5112, opened 2026-06-23
Make the actual host mask in whois more like the host mask.

sadiepowell:
The current way the real hostname and IP are shown is annoying for opers as when we want to ban a user where the username is significant (e.g. IRCCloud) we can't just copy the user@host mask. Some servers (e.g. Libera.Chat) also just return an actual IP and no actual hostname.

This pull request makes the following changes:

1. Moves the IP address to its own field.
2. Always shows a `user@host` mask when the actual hostname numeric is sent even on IRCds that do not send a real username alongside the hostname in `RPL_WHOISACTUALLY`.
3. Renames the whois field to "Actual host mask" to match the "Host mask" field.

Before (4.5.0):
Screenshot from 2026-06-23 12-22-02

After (on a server that sends a real username, e.g. `testnet.inspircd.org`):
Screenshot from 2026-06-23 12-33-50

After on a server that only sends a real hostname, e.g. `irc.libera.chat`):
Screenshot from 2026-06-23 12-24-14

### DIFF
--- a/client/components/MessageTypes/whois.vue
+++ b/client/components/MessageTypes/whois.vue
@@ -20,23 +20,31 @@
 			</dd>
 
 			<template v-if="message.whois.actual_hostname">
-				<dt>Actual host:</dt>
+				<dt>Actual host mask:</dt>
 				<dd class="hostmask">
+					<ParsedMessage
+						v-if="message.whois.actual_username"
+						:network="network"
+						:text="message.whois.actual_username + '@' + message.whois.actual_hostname"
+					/>
+					<ParsedMessage
+						v-else
+						:network="network"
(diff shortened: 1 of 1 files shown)


sadiepowell:
Note: there's a bug in irc-fw with parsing this numeric on some servers that also affects current stable versions of The Lounge (whois yourself on Libera.Chat for an example of this). I'll be sending a pull request upstream to fix that in a bit.

Edit: https://github.com/kiwiirc/irc-framework/pull/415

MaxLeiter (approved):

MaxLeiter:
thanks! ill merge the kiwiirc PR into the branch we use (due to no new releases) and then merge this.

sadiepowell:
The irc-framework fix I mentioned upthread has now been merged.

PR #5118, opened 2026-06-29
Config reload fix for BSD

Sweeistaken:
Commit c20e9d2 implements user reloading by relying on filename from fs.watch, but this breaks compatibility for BSD because Node.js only reports filenames on Windows, Linux, MacOS, and AIX.

This patch fixes it by doing a full reload again if the filename is null.

### DIFF
--- a/server/clientManager.ts
+++ b/server/clientManager.ts
@@ -87,26 +87,61 @@ class ClientManager {
 	}
 
 	autoloadUsers() {
+		const fullReload = _.debounce(
+			() => {
+				const loaded = this.clients.map((c) => c.name);
+				const updatedUsers = this.getUsers();
+
+				if (updatedUsers.length === 0) {
+					log.info(
+						`There are currently no users. Create one with ${colors.bold(
+							"thelounge add <name>"
+						)}.`
+					);
+				}
+
+				// Reload all users. Existing users will only have their passwords reloaded.
+				updatedUsers.forEach((name) => this.loadUser(name));
+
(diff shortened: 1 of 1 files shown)


server/clientManager.ts
```diff
@@ -87,25 +87,59 @@ class ClientManager {
...
-
-			const client = _.find(this.clients, {name});
-
-			if (client) {
-				client.quit(true);
-				this.clients = _.without(this.clients, client);
-				log.info(`User ${colors.bold(name)} disconnected and removed.`);
+			// If the filename is null, the platform (e.g. BSD) does not report filenames in fs.watch
+			if (file === null) {
+				_.debounce(
```
MaxLeiter: this debounce isnt doing much; every time `watch` is fired a new debounced function is made, which kind of defeats the purpose

```tsx
const reloadAllUsers = _.debounce(...);
fs.watch(Config.getUsersPath(), (_eventType, file) => {
   // code with reloadAllUsers()
});
```

MaxLeiter (changes requested):

server/clientManager.ts
Sweeistaken: then i don't see why it did so on the commit prior to that update, it doesn't seem to require any args from watch so i could properly define a debounce for the function somewhere as shown in your snippet.

Sweeistaken: added changes and used `git commit --amend`

MaxLeiter (approved):
thanks!

PR #5121, opened 2026-07-11
Fix Unicode nickname matching

### LINKED ISSUE #4930: Incorrect nickname highlighting when Polish diacritics are present
[There is a bot on Quakenet called "D"](https://www.quakenet.org/help/d/what-is-d). When he is on the channel, TL starts treating the letter "D" or "d" strangely. It seems that app sees calling of D-nick in simple words, provided that the the "d" must be followed by a specific letter from the Polish alphabet. Below are some examples.

obraz
obraz
obraz
obraz

- Firefox 128.3.1esr
- Windows, Android
- TL 4.4.3

---

JamBalaya56562:
## What changed

- Match nickname tokens with Unicode letters and numbers while preserving IRC nickname characters.
- Use Unicode-aware word boundaries for nickname highlights.
- Add regression coverage for Polish text, Unicode nicknames, and existing IRC nickname characters.

## Why

The previous regular expressions used ASCII-only character classes. This caused a one-character nickname such as `D` to be detected inside words containing Polish diacritics, for example `dzięki`.

## Before / after

| Before (v4.5.2) | After (this PR) |
|:---:|:---:|
| Before | After |

| Scenario | Before | After |
|---|---|---|
| Nick `Ądam42` appears in a message | Not recognized as a nick — rendered as plain text (no color, not clickable) | Recognized and colored like any other nick |
| `Ćadam` / `adamć` with your nick `adam` | Falsely highlighted as a mention of you | Not highlighted |
| Genuine mention `hey adam` | Highlighted | Highlighted (no regression) |

<sub>Screenshots are a faithful reproduction rendered with TheLounge's real <code>client/css/style.css</code> and the actual old/new regexes (not a live IRC capture).</sub>

## Validation

- Direct Unicode nickname extraction and boundary checks with Node.js 24
- Full Vitest/lint suite not run because repository dependencies were unavailable and dependency installation stalled in the local environment

Fixes #4930

### DIFF
--- a/client/js/helpers/ircmessageparser/findNames.ts
+++ b/client/js/helpers/ircmessageparser/findNames.ts
@@ -1,6 +1,6 @@
 import {Part} from "./merge";
 
-const nickRegExp = /([\w[\]\\`^{|}-]+)/g;
+const nickRegExp = /([\p{Letter}\p{Number}_[\]\\`^{|}-]+)/gu;
 
 export type NamePart = Part & {
 	nick: string;
--- a/server/models/network.ts
+++ b/server/models/network.ts
@@ -485,15 +485,15 @@ class Network {
 	setNick(this: Network, nick: string) {
 		this.nick = nick;
 		this.highlightRegex = new RegExp(
-			// Do not match characters and numbers (unless IRC color)
-			"(?:^|[^a-z0-9]|\x03[0-9]{1,2})" +
+			// Do not match letters and numbers (unless IRC color)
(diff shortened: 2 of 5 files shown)


MaxLeiter:
Thanks! The change seems good but lint is failing

JamBalaya56562:
Thanks for the review! You were right — `lint:prettier` was failing because the `nickRegExp` assignment was split across two lines. Prettier wants it on a single line, so I've fixed the formatting and pushed the update. CI should be green now.

MaxLeiter (approved):

PR #5122, opened 2026-07-15
Match custom highlights as a substring for scripts without word boundaries

### LINKED ISSUE #4164: option for custom highlight expressions matches just containing string, not for word.
Hi, thank you for awesome IRC Client!

### Feature Description

Custom highlight of the Lounge matches for word and special case like mention `@...`. (is it correct?)
It is very useful.
But words in Japanese (CJK?) sentence connect without whitespace.
eg. "今日はいい天気です" = "今日" "は" "いい" "天気" です" (means "it is nice weather today")
So I cannot get highlight/notification for Japanese word. I can get highlight/notification with full Japanese sentence.

I request the feature that adding option for custom highlight, to check custom highlight expressions matches just containing string (in the message).

JamBalaya56562:
## What changed

- Custom highlights **and** highlight exceptions now match keywords written in scripts **without word boundaries** — CJK (Han / Hiragana / Katakana / Hangul), Thai, Lao, Khmer, Myanmar — as a **plain substring**, automatically and **per keyword**.
- Latin / ASCII keywords keep the existing **word-boundary** behavior, so `foo` still does **not** match inside `football`.
- **No new setting.** Whether a keyword is matched as a substring is decided by the keyword's own script, so a mixed list like `foo, 天気` does the right thing for each keyword.
- Add regression tests covering CJK substring matching, Latin word-boundary safety, and highlight exceptions.

## Why

Custom highlights currently only match on **word boundaries** — the compiled regex wraps the tokens in explicit delimiter groups (`(?:^|[ .,+!?|/:<>(){}'"@&~-]) … (?:$|[ .,+!?|/:<>(){}'"-])`). Languages that do not delimit words with spaces or punctuation — Japanese and other CJK scripts — can never trigger a custom highlight for a keyword that sits in the middle of a sentence (e.g. `天気` inside `今日はいい天気ですね`).

The first version of this PR added an opt-in `highlightSubstring` setting that turned word boundaries off **globally**. As @deltamualpha [pointed out in review](https://github.com/thelounge/thelounge/pull/5122#issuecomment-4987587037), a global toggle does the wrong thing for a mixed keyword list: with `天気` and `foo`, turning substring matching on would also make `football` highlight — presumably not desired.

This version drops the setting and instead decides **per keyword**: a keyword is matched as a substring only when it contains characters from a script that has no word boundaries. So `天気` matches anywhere, while `foo` still requires boundaries — no configuration to get wrong, and no over-matching for Latin keywords.

## Before / after

| Before (word-boundary only) | After (this PR) |
|:---:|:---:|
| Before | After |

Custom highlight keyword: `天気`

| Message | Before | After |
|---|---|---|
| `今日はいい天気ですね` (keyword embedded, no spaces) | Not highlighted | **Highlighted** |
| `明日の 天気 はどうかな` (keyword with spaces) | Highlighted | Highlighted (no regression) |
| `おはよう、みなさん` (no keyword) | Not highlighted | Not highlighted |

### Latin keywords are unaffected — no over-matching

With a mixed list `foo, 天気`:

| Message | Result |
|---|---|
| `football` | Not highlighted (`foo` still needs a word boundary) |
| `hey foo!` | **Highlighted** |
| `今日はいい天気ですね` | **Highlighted** (`天気` matched as a substring) |

<sub>Screenshots are a faithful reproduction rendered with TheLounge's real <code>client/css/style.css</code> and the actual <code>compileCustomHighlights</code> logic (not a live IRC capture).</sub>

## Validation

- `npx vitest run test/tests/customhighlights.ts` — 11/11 pass (existing word-boundary cases, new CJK substring cases, Latin word-boundary safety, and exception cases).
- eslint / prettier clean on changed files; `tsc -p server/tsconfig.json` clean.

Fixes #4164

### DIFF
--- a/server/client.ts
+++ b/server/client.ts
@@ -544,6 +544,13 @@ class Client {
 	}
 
 	compileCustomHighlights() {
+		// Scripts that do not delimit words with spaces or punctuation (CJK, Thai,
+		// Lao, Khmer, Myanmar, ...). Tokens containing such characters can appear in
+		// the middle of a run of text with no boundary around them, so they are
+		// matched as a plain substring instead of being wrapped in word boundaries.
+		const noWordBoundary =
+			/[\p{Script=Han}\p{Script=Hiragana}\p{Script=Katakana}\p{Script=Hangul}\p{Script=Thai}\p{Script=Lao}\p{Script=Khmer}\p{Script=Myanmar}]/u;
+
 		function compileHighlightRegex(customHighlightString: string) {
 			if (typeof customHighlightString !== "string") {
 				return null;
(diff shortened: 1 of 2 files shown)


deltamualpha:
Seems like this setting is *only* useful if you have custom highlights that are entirely in languages without word boundaries? in any other case, turning that setting on would do the wrong thing (i.e. if my list is `天気` and `foo`, turning it on would make `football` highlight now, which is presumably not desired behavior).

JamBalaya56562:
Good point — you're right that a global toggle does the wrong thing for a mixed list. With `天気` and `foo`, turning it on would also highlight `football`, which nobody wants.

I've reworked this to drop the setting entirely and decide **per keyword** instead. A keyword is matched as a substring only when it contains characters from a script that has no word boundaries (Han, Hiragana, Katakana, Hangul, Thai, Lao, Khmer, Myanmar); everything else keeps the existing word-boundary behavior. So:

- `天気` matches anywhere (e.g. inside `今日はいい天気ですね`)
- `foo` still requires a boundary and does **not** match inside `football`

That way a mixed list like `foo, 天気` does the right thing for each keyword with no configuration to get wrong. Implementation just splits the tokens into two alternations at compile time in `compileCustomHighlights`.

Added regression tests for the CJK substring cases, the Latin word-boundary safety (`foo` ≠ `football`), and highlight exceptions — `test/tests/customhighlights.ts` passes 11/11. The description and screenshots are updated to match.

PR #5123, opened 2026-07-15
Always wrap emoji with the emoji class (inside links, channels and styles)

### LINKED ISSUE #1783: Messages containing emoji and colors are not fully optimized
See https://github.com/thelounge/lounge/pull/1770#discussion_r154022888

When parsing `Super \x034💚 green!`, result should be:

```html
Super <span class="emoji irc-fg4">💚</span><span class="irc-fg4"> green!</span>
```

It is instead:
```html
Super <span class="emoji"><span class="irc-fg4">💚</span></span><span class="irc-fg4"> green!</span>
```

### LINKED ISSUE #1784: Emoji in channels and links are not given the correct class
Emoji parsed in message does not get parsed correctly. It can be seen in the app because they appear smaller:

screen shot 2017-11-30 at 20 14 26

See https://github.com/thelounge/lounge/pull/1770#discussion_r154023071 for relevant discussion.

JamBalaya56562:
## What changed

Emoji were only wrapped in `<span class="emoji">` at the **part** level, so an emoji
that fell inside a link, channel name, or nick lost the `emoji` class and rendered at the
normal font size instead of the enlarged emoji size. Emoji combined with an IRC color also
produced nested spans.

This moves emoji wrapping to the **fragment (text-run) level** in `client/js/helpers/parse.ts`:
each styled fragment is split into emoji and non-emoji runs, and emoji runs are wrapped with
the `emoji` class merged with any surrounding style. As a result:

- Emoji keep the `emoji` class inside links, channels and nicks (**#1784**).
- A link's `href` is unaffected — the emoji span only goes into the visible text.
- Styled emoji collapse from nested spans into a single `<span class="emoji irc-fgN">` (**#1783**).

## Why

`.emoji` bumps emoji to `font-size: 1.4em`. Without the class, emoji inside a channel/link
stayed at `1em` and looked noticeably smaller than emoji elsewhere in the message.

## Before / after

| Before | After |
|:---:|:---:|
| Before | After |

| Case | Before | After |
|---|---|---|
| Emoji in a channel `#i❤️thelounge` | `❤️` unwrapped, renders at 1em (small) | `❤️` wrapped in `.emoji`, 1.4em |
| Emoji in a link `https://i.❤️.thelounge.chat` | `❤️` unwrapped in link text | `❤️` wrapped in link text; `href` unchanged |
| Emoji + color `\x034💚` | `<span class="emoji"><span class="irc-fg4">💚</span></span>` | `<span class="emoji irc-fg4">💚</span>` |

<sub>Screenshots are a faithful reproduction rendered with TheLounge's real <code>client/css/style.css</code> and the exact HTML emitted by <code>parse()</code> before/after the fix (not a live IRC capture).</sub>

## Tests / validation

- Updated the three FIXME-marked cases in `test/client/js/helpers/parse.ts` to the corrected
  output for #1784 and #1783.
- Verified the actual rendered HTML of `parse()` (via `ParsedMessage.vue`) for emoji inside
  links, channels, styled text, multiple emoji per link, and standalone emoji.

> Note: the `test/client/js/helpers/**` parser tests are not currently matched by the vitest
> `include` globs (only `*Test.ts` / `test/tests/**` are), so this suite is not run by CI today.
> The expected values are updated to reflect correct behavior regardless. Wiring these tests
> back into CI is a separate change and out of scope here.

Fixes #1784
Fixes #1783

### DIFF
--- a/client/js/helpers/parse.ts
+++ b/client/js/helpers/parse.ts
@@ -35,8 +35,11 @@ type StyledFragment = Fragment & {
 	strikethrough?: boolean;
 };
 
-// Create an HTML `span` with styling information for a given fragment
-function createFragment(fragment: StyledFragment): VNode | string | undefined {
+// Build the class list and inline style shared by styled and emoji fragments
+function fragmentStyle(fragment: StyledFragment): {
+	classes: string[];
+	style?: Record<string, string>;
+} {
 	const classes: string[] = [];
 
 	if (fragment.bold) {
@@ -67,13 +70,29 @@ function createFragment(fragment: StyledFragment): VNode | string | undefined {
 		classes.push("irc-monospace");
 	}
 
+	let style: Record<string, string> | undefined;
+
+	if (fragment.hexColor) {
+		style = {
+			color: `#${fragment.hexColor}`,
+		};
+
+		if (fragment.hexBgColor) {
+			style["background-color"] = `#${fragment.hexBgColor}`;
+		}
+	}
+
+	return {classes, style};
+}
+
+// Create an HTML `span` with styling information for a given fragment
+function createFragment(fragment: StyledFragment): VNode | string | undefined {
+	const {classes, style} = fragmentStyle(fragment);
+
 	const data: {
 		class?: string[];
 		style?: Record<string, string>;
-	} = {
-		class: undefined,
-		style: undefined,
-	};
+	} = {};
 
 	let hasData = false;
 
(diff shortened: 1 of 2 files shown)


deltamualpha:
Honestly, the Before images in that table look better to me. Emoji that are inline with some other meaningful text string like that (A URL or a channel name) should match the size of the surrounding text, not be oversized in relation to it.

JamBalaya56562:
Thanks, that's a fair point and I think you're right — when an emoji sits inline inside a meaningful string like a URL or a channel name, blowing it up to `1.4em` looks out of place next to the surrounding text.

So to make sure I'm heading in the right direction before I rework this: the desired behavior is that the `.emoji` enlargement should **only** apply to emoji in regular message text, and emoji embedded inside a link / channel / nick should stay at the surrounding text size (i.e. the "Before" rendering is actually correct for those cases). Is that right?

If so, I'll flip the approach: keep standalone emoji enlarged as today, and specifically **not** enlarge emoji when they're part of a link/channel/nick, rather than adding the class everywhere.

One thing I'd like to keep regardless of the sizing decision is the #1783 part — collapsing the nested `<span class="emoji"><span class="irc-fgN">💚</span></span>` into a single span. That's just redundant-markup cleanup and is independent of how large the emoji renders. Happy to split that out into its own PR if you'd prefer to keep the two concerns separate.

deltamualpha:
Did your user just point Claude at thelounge repo and say "fix issues"? These are the sorts of PRs that need to be discussed before just throwing them over the wall at maintainers.

JamBalaya56562:
Fair, and I hear you — I don't want to pile review load on you that you didn't ask for.

To be upfront: yes, I'm working with AI assistance, but this isn't a fire-and-forget "go fix issues" run — I'm driving and reviewing the changes myself. I opened a PR mainly because I find it much easier to reason about what the *correct* behavior should be against a concrete change and a screenshot than in the abstract, so I meant it as a starting point for that discussion rather than a "please merge this" drop.

That said, this is your call on the design. I don't have hard proof that the current behavior is a bug rather than intended — so if you consider it correct/by-design, that's completely fine and I'll close the PR. And if you'd prefer to settle the direction in the issue first, I'm happy to move the discussion there. I just wanted to surface it and help work toward a fix together, not throw noise over the wall.

PR #5124, opened 2026-07-16
Compare locked-network host to default case-insensitively

### LINKED ISSUE #4733: server adresses with uppercase letters in it will cause problems when reconnecting to a server after restarting The Lounge
- _Node version:_ v20.0.0
- _Browser version:_ Chrome, Edge, Firefox
- _Device, operating system:_ Windows, Android
- _The Lounge version:_ v4.3.1 - 2022-04-11

if there is an uppercase letter in the hostname of the server (e.g. irc.eXample.net) in the defaults block, there will be an error message saying: "The hostname you specified (irc.example.net) is not allowed." when reconnecting after restarting The Lounge

JamBalaya56562:
## What changed

`Network.validate()` now compares a locked network's host against the configured
`defaults.host` **case-insensitively**.

## Why

In `server/models/network.ts`, `validate()` lowercases the incoming host (hostnames are
case-insensitive):

```ts
this.host = cleanString(this.host).toLowerCase();
```

but the `lockNetwork` guard then compared that lowercased value against the **raw** configured
default with a strict `!==`:

```ts
this.host !== Config.values.defaults.host
```

`Config.values.defaults.host` is never normalized on load, so when an admin's `defaults.host`
contains uppercase letters (e.g. `irc.eXample.net`), a persisted network whose stored host had
been lowercased (`irc.example.net`) no longer matched the default and was rejected with:

> The hostname you specified (irc.example.net) is not allowed.

The block is guarded by `this.host.length > 0`, so it doesn't fire on the first connect (locked
networks send an empty host there). It only bites **after a restart**, when the persisted network
JSON carries the stored host — which is exactly the symptom reported in #4733: the network never
reconnects.

## Fix

Compare against `Config.values.defaults.host.toLowerCase()`. One line; no behavior change for
hosts that already matched.

## Tests

Added a regression test to `test/models/network.ts` (`#validate()`), which runs in CI. It sets a
mixed-case `defaults.host` under `lockNetwork` + private mode and asserts a network with the same
mixed-case host validates successfully. The test fails without the fix (`validate()` returns
`false`) and passes with it.

- `yarn vitest run test/models/network.ts` — 24/24 pass.
- `tsc -p server/tsconfig.json`, eslint and prettier clean on the changed files.

Fixes #4733

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -231,6 +231,15 @@ class Config {
 			this.merge(userConfig);
 		}
 
+		// Hostnames are case-insensitive. Network.validate() lowercases the host of
+		// every incoming/persisted network, so normalize the configured default to
+		// the same canonical form once here at load. This keeps the lockNetwork
+		// allow-list comparison in Network.validate() correct no matter what casing
+		// an admin used for defaults.host in config.js (see #4733).
+		if (this.values.defaults.host) {
+			this.values.defaults.host = this.values.defaults.host.toLowerCase();
+		}
+
 		if (this.values.fileUpload.baseUrl) {
 			try {
 				new URL("test/file.png", this.values.fileUpload.baseUrl);
--- a/server/models/network.ts
+++ b/server/models/network.ts
@@ -252,8 +252,9 @@ class Network {
 				!Config.values.public &&
 				this.host &&
 				this.host.length > 0 &&
-				// hostnames are case-insensitive; this.host is already lowercased above
-				this.host !== Config.values.defaults.host.toLowerCase()
+				// hostnames are case-insensitive; this.host is lowercased above and
+				// defaults.host is normalized to lowercase at config load (server/config.ts)
+				this.host !== Config.values.defaults.host
 			) {
 				error(this, `The hostname you specified (${this.host}) is not allowed.`);
 				return false;
--- a/server/models/network.ts
+++ b/server/models/network.ts
@@ -252,7 +252,8 @@ class Network {
 				!Config.values.public &&
 				this.host &&
 				this.host.length > 0 &&
-				this.host !== Config.values.defaults.host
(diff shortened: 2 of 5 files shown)


deltamualpha:
Wouldn't a more general fix here to be to validate the default hostnames in the configuration at load, and store them back in the configuration json in the correct uniform lowercase format?

JamBalaya56562:
Good call — I've pushed a follow-up commit (e9230aba7049bdc469c506303842efcc17dbd6a7) that does exactly this at load time.

`Config.setHome()` now lowercases `defaults.host` right after the config merge, so every reader — not just this one comparison — sees the canonical form. I also dropped the now-redundant inline `.toLowerCase()` in `Network.validate()`'s `lockNetwork` guard.

One note on "store them back in the configuration json": the server config is a hand-edited JS module (`config.js`) with no write-back path, so the normalization is applied to the in-memory `Config.values` at load rather than persisted to disk. Happy to adjust if you had something else in mind.

Tests: the fixture now uses a mixed-case `defaults.host` to exercise the load-time normalization, there's an assertion in `mergeConfig`, and the #4733 regression test is updated to match.

MaxLeiter (approved):

MaxLeiter:
Thanks! In the future @JamBalaya56562 please do not use AI to respond to PR comments/reviews/etc. I will be publishing more formal guidelines soon

JamBalaya56562:
Understood, and sorry about that — I won't do it again.

One thing I'd like to mention: English is not my first language,
so I sometimes write in my native language and translate it into English before posting.
I hope that's still acceptable under the guidelines.

Thanks for the heads-up, and I'll watch for the formal guidelines.

MaxLeiter:
Yea, the guidelines specify you should use google translate or something translation specific so it preserves your voice. Thanks for the contributions!

Issue #4606, opened 2022-07-10
Support logging to PostgreSQL (or similar)


cyberia-ng:
### Feature Description

It already can read/write logs to an SQLite file; it would be nice if you could use an external db server for this too, so you can e.g. run analytics, scrub logs periodically to comply with GDPR, etc.

nodadyoushutup:
Bumping this request. Would LOVE to see a feature like this. Due to my new k3s build all my apps are being migrated. And I am trying to get every application on postgres the best I can. Would be super interested in seeing what we could do to facilitate this request.

Issue #3398, opened 2019-09-11
Log channels in public mode


primozc:
### Feature Description
It would be great if you could log channels in public mode.  Currently is working great in private mode.

pdurbin:
I love this idea! Would the public logs be visible before you sign in? Below is an example of public logging (anonymous read access) from Gitter. Note how at the bottom there's a button that says, "Sign in to start talking."

<img width="1503" alt="Screen Shot 2019-09-11 at 5 45 08 AM" src="https://user-images.githubusercontent.com/21006/64686885-6ef05c00-d457-11e9-9861-41cdb849e5aa.png">

xPaw:
This would require a separate user to always be in a channel for it to be logged fully. How would you tackle that when public users can join any channel on the server.

The solution here is chathistory in ircv3 which means IRC server can send the client their history.

pdurbin:
What if I have my own IRC logging bot and do a one time population of the public logs in The Lounge based on this TSV file? http://dvn-vm1.hmdc.harvard.edu/data/mysql/irclog/irclog.tsv

And then going forward maybe I could teach my logging bot to write to the logging format for The Lounge? (And continue to write to a database like it does now.) This is the bot I use: https://github.com/pdurbin/philbot

primozc:
> This would require a separate user to always be in a channel for it to be logged fully. How would you tackle that when public users can join any channel on the server.
> 
> The solution here is chathistory in ircv3 which means IRC server can send the client their history.

Then we can wait for IRCv3 to be implemented. Temporary solution could be simple bot which connects and track specific / pinned channels?

xPaw:
Discussing logging bots quickly gets out of scope for TL.

chathistory cap is what we would want to support ideally in core. And stuff like logging bots and idling in specific channels would be something a plugin could eventually be able to do.

pdurbin:
@xPaw do you (or anyone reading this) know if freenode or any other popular IRC networks have at least experimental support for IRCv3 chathistory? I'm aware that https://github.com/ircv3/ircv3-specifications/pull/393 hasn't been merged yet, which seems to be the latest version of the (work in progress) spec.

McInkay:
I do not know of any popular networks having it.

If you look at the issue, it has the label "Implementation needed", which means that no one has properly implemented it. That, combined with the fact that there is no one commenting that they have implemented it, implies it isn't actually implemented anywhere yet.

[...]

PR #5138, opened 2026-08-07
Add jump to message navigation

### LINKED ISSUE #4101: Mentinos window: click to jump to context
### Feature Description

Often a channel where a mention happened has a lot of traffic, so scrolling to the context of a mention can be tedious.
It would be nice if the mentions window would allow me to click on a mention to jump to that context.

n.b.: i searched thru existing issues, but I'm not sure if https://github.com/thelounge/thelounge/issues/3943 is related

gimletlove:
Supersedes #5038. Closes #4101.

### Description

Messages can now be seamlessly opened directly by clicking (or keyboard activation) the message / text from:

- Search results
- Mentions
- Browser notifications

The new storageId field uses the existing SQLite message row ID to allow the server to retrieve messages that are no longer loaded in client or server memory. 

The client loads a window around the selected message, scrolls to it, and highlights it. History can be further be browsed in both directions as normal, while the existing down arrow/jump to end button returns to the actual live end of the channel.

While viewing history, Incoming messages are not appended to prevent gaps and duplicates.

Additionally fixed the hidden “Show older messages” button in search, which prevented users from loading more messages in search. Older matches now also load automatically when scrolling upward, matching the existing behavior in chats.

### Testing

- Jumped to messages from search results, mentions, and browser notifications using my own database with channels and private messages going back years.
- Loaded and scrolled through older and newer history after jumping.
- Returned to the latest message using the down arrow/jump to last button.
- Scrolled through large search result sets and automatically loaded older matches.
- Unit tests for: storage IDs being assigned before messages are emitted,  loading and paging through history around a selected message.

### DIFF
--- a/client/components/Chat.vue
+++ b/client/components/Chat.vue
@@ -95,7 +95,7 @@
 							{'scroll-down-shown': !channel.scrolledToBottom},
 						]"
 						aria-label="Jump to recent messages"
-						@click="messageList?.jumpToBottom()"
+						@click="messageList?.jumpToLatest()"
 					>
 						<div class="scroll-down-arrow" />
 					</button>
@@ -105,6 +105,7 @@
 						:network="network"
 						:channel="channel"
 						:focused="focused"
+						:focused-storage-id="focusedStorageId"
 					/>
 				</div>
 			</div>
@@ -156,6 +157,7 @@ export default defineComponent({
 		network: {type: Object as PropType<ClientNetwork>, required: true},
 		channel: {type: Object as PropType<ClientChan>, required: true},
 		focused: Number,
+		focusedStorageId: Number,
 	},
 	emits: ["channel-changed"],
 	setup(props, {emit}) {
--- a/client/components/Mentions.vue
+++ b/client/components/Mentions.vue
@@ -21,7 +21,14 @@
 				<p v-else>You have no recent mentions.</p>
 			</template>
 			<template v-for="message in resolvedMessages" v-else :key="message.msgId">
-				<div :class="['msg', message.type]">
+				<div
+					:class="['msg', message.type, {clickable: message.channel}]"
+					:role="message.channel ? 'button' : undefined"
+					:tabindex="message.channel ? 0 : undefined"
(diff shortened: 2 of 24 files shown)


spin-drift:
THIS IS HUGE

spin-drift:
It would be so incredible if this could be included in 4.6.0.

ColinHebert:
Thanks for this, jumping from search results and mentions is a big usability win.

One suggestion on `storageId`, while it's still cheap to change: could it be an **opaque token owned by the storage provider** instead of the SQLite row id as a number?

Right now the row id escapes the storage layer and becomes part of the client protocol and of URLs that persist outside the app:

- `storageId?: number` in `shared/types/msg.ts`, `shared/types/mention.ts` and the `more` / `history:*` socket events
- `isMessageId()` in `server/client.ts` and the service worker only accept positive safe integers
- notifications open `#/…?focusedStorageId=<rowid>`

That's fine for SQLite. But not every history source has a stable integer key. The obvious one in this project is IRCv3 `chathistory` (#3398): it addresses messages by `msgid=` or `timestamp=`, not row ids. The same applies to an external store (#4606). If the client contract is "an integer row id", every future provider has to fake one or change the protocol again.

A small change would keep the feature exactly as it is:

1. Type it as `storageId?: string` in the shared types and socket events.
2. SQLite encodes it as `String(row.id)` in `parseStoredRow()` and after the insert in `index()`, and parses it back in `getMessagesAround()` / `getStoredMessage()`. That can be a regex check on the decoded value, so bad input still returns `null` as it does now.
3. `isMessageId()` for `storageId` (server and service worker) becomes a "non-empty string, bounded length" check. The provider does the real validation.
4. Declare `getMessagesAround()` on the storage interface in `server/plugins/messageStorage/types.d.ts`, not only on `SqliteMessageStorage`, so it's part of the provider contract next to `getMessages()` and `search()`.

With SQLite as the only provider, behaviour is identical. It just stops the storage row id from leaking into the client and URLs.

Happy to send a follow-up PR for this after merge if you'd prefer to keep this one focused.

Issue #2831, opened 2018-10-06
TypeError: Cannot set property 'openChannel' of undefined on 2.7.1


utopicdev:
Using a stable version (2.7.1 @ Node.js 8.12.0 on linux x64), I get TheLounge constantly crashing when there are >100 simultaneous client connections.

> /usr/lib/node_modules/thelounge/src/client.js:466
>         this.attachedClients[socketId].openChannel = target.chan.id;
>                                                    ^
> 
> TypeError: Cannot set property 'openChannel' of undefined
>     at Client.open (/usr/lib/node_modules/thelounge/src/client.js:466:45)
>     at Socket.<anonymous> (/usr/lib/node_modules/thelounge/src/server.js:353:11)
>     at emitOne (events.js:116:13)
>     at Socket.emit (events.js:211:7)
>     at /usr/lib/node_modules/thelounge/node_modules/socket.io/lib/socket.js:513:12
>     at _combinedTickCallback (internal/process/next_tick.js:132:7)
>     at process._tickCallback (internal/process/next_tick.js:181:9)
>

xPaw:
This is a duplicate of #1995, and I think it's fixed on 3.0, try using the pre-release.

Issue #1995, opened 2018-01-20
Socket.io still fires event after disconnect


crosma:
https://github.com/socketio/socket.io/issues/3095

> <+xPaw> get socketio-wildcard, in disconnected event set disconnected[socketid] = true, with wildcard, add "*" event, check against disconnected array

Use this to log when disconnected sockets receive events and stop them. Also log if socketId ever existed or if it's somehow new.

MaxLeiter:
Possibly related:
```
/root/lounge/src/client.js:465
        this.attachedClients[socketId].openChannel = target.chan.id;
                                                   ^

TypeError: Cannot set property 'openChannel' of undefined
    at Client.open (/root/lounge/src/client.js:465:45)
    at Socket.<anonymous> (/root/lounge/src/server.js:353:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at /root/lounge/node_modules/socket.io/lib/socket.js:513:12
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)
```

xPaw:
You can most likely get more debug info without modifying the code, run lounge with `env DEBUG="*"`, that should give a lot of socket.io info leading to the crash.

xPaw:
Does this still happen? Maybe this issue was related to `uws` module which we explicitly disabled.

PR #5141, opened 2026-08-24
Don't crash when a socket detaches during token generation

Ian-Costa18:
generateToken is asynchronous, so clientDetach can delete the attached client before its callback runs. Assigning the token then throws from a threadpool callback, which nothing catches, and the process exits.

Client.open() already guards this (#1995, #2831), this is the same bug at a second site which the guard never got applied to.


MaxLeiter (approved):

MaxLeiter:
thanks!

PR #5142, opened 2026-09-02
Add crossorigin attribute to manifest link so it work with CF tunnels

Macleykun:
Reason: https://github.com/cloudflare/cloudflare-docs/blob/production/src/content/docs/cloudflare-one/access-controls/applications/http-apps/authorization-cookie/cors.mdx example: https://github.com/open-webui/open-webui/issues/1538

So with the help of chatgpt i saw 
image

And it's fixed with this small change:
image

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -23,7 +23,7 @@
 		<!-- Safari pinned tab icon -->
 		<link rel="mask-icon" href="img/icon-black-transparent-bg.svg" color="#415364" />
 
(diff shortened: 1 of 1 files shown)


MaxLeiter:
Thanks!

This was originally fixed in https://github.com/thelounge/thelounge/pull/5060 but #5064 lost it (due to a bad merge, most likely)

MaxLeiter (approved):

